Issue #1121, opened 2026-06-10
bug(monitoring): fleet health loop not wired to app lifespan — dies on every backend restart, never resumes


vybe:
## Summary

The fleet **agent-monitoring loop** (`MonitoringService`, MON-001) is an in-process `asyncio` task that is **never started at backend boot**. The only code path that starts it is an admin calling `POST /api/monitoring/enable`. As a result, **every `trinity-backend` restart silently stops monitoring and leaves it off** until a human manually re-enables it via the API/UI. The persisted `system_settings.monitoring_config` (including its `enabled` flag) is never consulted on startup to resume the loop.

This was confirmed in production: on a deployed instance the backend restarted ~14h prior and the monitoring loop had been dead ever since. The only `agent_health_checks` rows being written were one-off on-demand checks (`POST .../check`, `trigger_health_check`), not the periodic loop.

We want to (1) fix the resume-on-restart defect, (2) turn monitoring into a real **persisted setting that defaults to OFF**, and when ON actually runs and survives restarts while gathering per-agent health data, and (3) confirm the data set respects a **7-day default retention** (already shipped via Abilityai/trinity#772) — with higher retention gated behind Enterprise (tracked separately).

## Context

`architecture.md` describes the 30s `monitoring_service.py` loop as the **authoritative** fleet-health layer (the Abilityai/trinity#307 heartbeat watch loop is documented as "additive ... to the 30s monitoring loop, which stays authoritative"). In reality the authoritative loop isn't wired to the app lifespan at all, so the documented invariant is silently false on any instance whose backend has restarted.

### Root cause (traced through source)

- `lifespan()` in `src/backend/main.py` starts cleanup, session-cleanup, sync-health, capacity-maintenance, and the heartbeat-watch loop — but contains **no** `start_monitoring_service()` call.
- `MonitoringService.start()` (`src/backend/services/monitoring_service.py`) does not check any `enabled` flag; it just spawns `_run_loop()`. The only caller is `routers/monitoring.py` → `enable_monitoring()` → `background_tasks.add_task(start_monitoring_service)`.
- The persisted setting `monitoring_config` is read only by `GET /api/monitoring/config`; nothing reads it on boot to decide whether to resume the loop.
- `MonitoringConfig.enabled` (`src/backend/db_models.py`) defaults to `True` but is effectively **dead** — no code consults it. The effective behavior today is "off until manually enabled," which contradicts the config default.

### Secondary defect — unvalidated negative interval

[...]

Issue #772, opened 2026-05-11
feat: implement execution log retention and pruning policy


vybe:
## Summary

The `schedule_executions` table stores a full JSONL transcript in the `execution_log` column for every agent run, with no TTL or pruning. On active instances with many scheduled agents this table balloons to 3+ GB, dominating the SQLite database and causing disk pressure. There is also no retention policy on `agent_health_checks`, which accumulates hundreds of thousands of rows over time.

## Component

Backend / Database / Cleanup Service

## Priority

P1 — active instances hit disk pressure within weeks; no workaround short of manual SQL.

## Problem Detail

On a production instance with ~10 scheduled agents running hourly:

| Table | Size | Rows | Growth rate |
|-------|------|------|-------------|
| `schedule_executions` | ~3.3 GB | ~9,000 | ~150–200 MB/week |
| `agent_health_checks` | ~200 MB (incl. indexes) | ~750,000 | continuous |

The `execution_log` column is the primary driver — it stores the raw JSONL Claude Code session transcript, averaging 150–190 KB per run. This is valuable for recent executions (debugging, audit trail) but has diminishing value for runs older than a few weeks.

### 1. Execution log retention policy

Add a configurable retention window (e.g. `EXECUTION_LOG_RETENTION_DAYS`, default 30). The existing cleanup service (`services/cleanup_service.py`) should periodically:

```python
# Null out execution_log for old completed executions (preserves row + metadata)
UPDATE schedule_executions
SET execution_log = NULL
WHERE status IN ('completed', 'failed', 'terminated')
  AND completed_at < datetime('now', '-30 days')
  AND execution_log IS NOT NULL;
```

Nulling rather than deleting preserves the execution record (agent, duration, cost, status) while reclaiming the bulk of the space.

### 2. Full execution row pruning (optional, deeper retention)

For very old executions (e.g. >90 days), optionally delete the row entirely:

```python
DELETE FROM schedule_executions
WHERE completed_at < datetime('now', '-90 days')
  AND status IN ('completed', 'failed', 'terminated');
```

### 3. Health check retention

Cap `agent_health_checks` to a rolling window (e.g. 7 days):

```python
DELETE FROM agent_health_checks
WHERE checked_at < datetime('now', '-7 days');
```

### 4. Admin configuration

Expose retention settings via the ops config API (`GET/PUT /api/settings/ops/config`) so operators can tune or disable pruning per instance:

```json
{
  "execution_log_retention_days": 30,
[...]

Issue #675, opened 2026-05-06
investigate: perform_fleet_health_check stops refreshing data for most agents (#669 root cause)


dolho:
## Summary

Split from #669 (which patched the 500 symptom). The underlying cause of the stale 24h rollups remains unfixed: in production, `services/monitoring_service.py:perform_fleet_health_check` was last refreshing 8 of 9 agents months ago (last refresh 2026-02-23 to 2026-04-28), while only one agent (`competitor-analyzer`) was getting fresh data on the expected interval.

#669 closes the user-visible 500 by coercing NULL `status` rows to `"unknown"` and wrapping the aggregator in try/except. That degrades gracefully — but agents still show stale data because the scheduled fleet check is not actually running for most of them.

## Hypothesis space

1. **Silent exception in the per-agent loop kills the rest of the fleet.** `perform_fleet_health_check` may iterate without per-agent isolation, so one agent's failure short-circuits the whole pass.
2. **Monitoring service `is_running` flag flipped off** without restart visibility. Check `/api/monitoring/config` history.
3. **Scheduler interval drift / coroutine cancellation** that nobody noticed.
4. **Agents created after a certain date never got registered** with the monitoring loop (registration only at startup vs. on-create).

## Asks

- [ ] Pull production scheduler / monitoring-service logs around the stalest refresh timestamps (`2026-02-23`, `2026-03-04`, `2026-03-19`, `2026-04-28`).
- [ ] Confirm whether `perform_fleet_health_check` wraps each agent in its own try/except — if not, add isolation.
- [ ] Add structured log line per fleet-check pass with agent count + per-agent outcome so this is observable next time.
- [ ] Audit registration path: does adding/restarting an agent re-register it with the monitoring service?

## Related

- #669 — symptom patch (this is the root-cause follow-up)
- `services/monitoring_service.py:528 perform_fleet_health_check`

vybe:
Shipped in **v0.6.0** (released via #998).

Issue #307, opened 2026-04-11
feat: Agent heartbeat push for fast failure detection (RELIABILITY-004)


vybe:
## Summary

Agent health is currently detected by **polling** every 30 seconds from the backend. This means an agent crash takes up to 30s to detect, and during that window other agents continue sending requests to a dead container. Combined with no circuit breaker (see RELIABILITY-001), this causes long cascading timeouts.

## Solution

Flip from pull to push: agents send a lightweight heartbeat to the platform every 5 seconds. A missed heartbeat triggers immediate status change and (with RELIABILITY-001) opens the circuit breaker preemptively.

## Scope

**Backend:**
- New internal endpoint: `POST /api/internal/heartbeat` (no auth, internal network only)
- Store heartbeat timestamp in Redis with 15s TTL: `heartbeat:{agent_name}`
- Health check reads Redis key — missing key = agent unresponsive
- Feed heartbeat status into monitoring service and circuit breaker (if RELIABILITY-001 is implemented)

**Agent-side:**
- Add background task in `agent-server.py` that POSTs heartbeat every 5s
- Include basic health info in payload: memory usage, active execution count, uptime

## Acceptance Criteria

- [ ] Running agents send heartbeat every 5 seconds
- [ ] Backend detects missing heartbeat within 15 seconds (vs 30s+ currently)
- [ ] Heartbeat status reflected in fleet health endpoint
- [ ] Agent server continues working normally if backend is temporarily unreachable (heartbeat fails silently)
- [ ] ~40 lines total (backend endpoint + agent-side background task)
- [ ] No new infrastructure — uses existing Redis

## Key Files

- `src/backend/routers/internal.py` — new heartbeat endpoint
- `src/backend/services/monitoring_service.py` — consume heartbeat data
- `docker/base-image/agent_server/` — add heartbeat background task

## Dependencies

- Enhances RELIABILITY-001 (circuit breaker) — heartbeat loss can proactively open circuit
- Independent of other reliability issues

vybe:
Under the pull model (#1081) the heartbeat is no longer a dispatch gate: a dead agent simply stops pulling, so no compute is wasted and the backend never floods a sick agent. The missed-heartbeat signal is repurposed into an **operator alert** (queue depth climbing + no successful results = unhealthy), not a mechanism that fails work pre-emptively. Pull does not depend on this, so de-prioritized p1→p2.

PR #1123, opened 2026-06-10
fix(monitoring): wire fleet-health loop to lifespan, persist enabled (default OFF) (#1121)

dolho:
## Summary

The MON-001 fleet-monitoring loop was **never started at backend boot** — the only code path that started it was an admin calling `POST /api/monitoring/enable`. So **every `trinity-backend` restart silently killed monitoring** and left it off until a human re-enabled it; the persisted `system_settings.monitoring_config` was never consulted on startup. The documented "30s monitoring loop is authoritative" invariant was silently false on any restarted instance.

This fixes the resume-on-restart defect, makes `MonitoringConfig.enabled` a real **persisted setting that defaults OFF** (single source of truth, was dead code), and closes a negative-interval flood gap.

## Changes

- **`main.py` `lifespan()`** — delayed (+12s) resume that reads the persisted `monitoring_config` and starts the loop **only when enabled**, so the choice survives restarts. Best-effort stop added to shutdown for parity with the other lifespan loops.
- **`MonitoringConfig`** (`db_models.py`) — `enabled` now **defaults to OFF**; `*_check_interval` fields reject non-positive values via a Pydantic validator (422 on request body; persisted bad config falls back to DEFAULT). Loop also clamps `asyncio.sleep` to ≥1s as a belt-and-suspenders guard.
- **`routers/monitoring.py`** — `enable`/`disable` and `PUT /config` now **persist** the `enabled` flag and reconcile the running loop, so `enabled` is authoritative regardless of entry point. Shared `load_persisted_monitoring_config()` + `_persist_monitoring_enabled()` helpers (one reader, no drift).
- **`architecture.md`** — monitoring loop described as lifespan-wired + gated by the persisted default-off setting; coordinated the Abilityai/trinity#307 heartbeat "stays authoritative" wording.
- **`test_monitoring.py`** — `test_update_config_validation` now asserts **422** for non-positive intervals (it previously asserted the unvalidated 200 — its own comment predicted this fix).

## Acceptance criteria

- [x] Fleet monitoring is a **persisted setting, default OFF** (single source of truth, not dead code).
- [x] When ON, loop **starts at boot** (lifespan-wired) and **survives restarts** by reading persisted config — no manual re-enable.
- [x] When OFF, no periodic loop runs (on-demand `/check` + `trigger_health_check` unchanged).
- [x] `enable`/`disable` **persist** the state across restarts.
- [x] While ON, the loop writes `agent_health_checks` for the whole fleet (#675, unchanged).
- [x] `*_check_interval` rejects non-positive values (422).
- [x] 7-day default retention (#772) unchanged.
- [x] `architecture.md` updated to match reality.

Higher-than-7-day retention gated behind Enterprise (#1049) is out of scope per the issue.

## Verification (live stack, end-to-end)

| Scenario | Result |
|----------|--------|
| `enable` → restart backend → wait | Loop **resumes**; `fleet_health_check` passes every ~30s (`enabled(running)=true`) |
| `disable` → restart backend → wait | **No** loop (0 fleet passes); persisted `enabled=false` |
| `PUT /config` with `*_check_interval` ∈ {-10, 0} | **422** for all three interval fields |
| `PUT /config` with interval=45 | 200 |
| `MonitoringConfig()` default | `enabled=False`; intervals validated |

Instance left in clean default-off state.

Related to Abilityai/trinity#1121


src/backend/routers/monitoring.py
```diff
@@ -433,38 +460,33 @@ async def get_monitoring_config(
...
 
-    # Update running service
+    # Update running service config, then reconcile its run state to the
+    # `enabled` flag so it stays the single source of truth (#1121):
+    # enabling here starts the loop; disabling here stops it.
     service = get_monitoring_service()
     service.config = config
+    if config.enabled and not service.is_running:
+        background_tasks.add_task(start_monitoring_service, config)
+    elif not config.enabled and service.is_running:
```
dolho, line 499: **Partial `PUT /config` now silently disables monitoring.** `MonitoringConfig` has no required fields (every field has a default, and `enabled` now defaults to `False`). So a client PUT that omits `enabled` — e.g. `{"docker_check_interval": 45}` — deserializes to `enabled=False`, and this new reconcile then hits the `elif not config.enabled and service.is_running` branch and **stops the running loop** (and resets every omitted field to its default).

Before this PR the same lossy full-replace happened but was inert (no reconcile, and `enabled` defaulted `True`). Now a routine interval tweak via a partial body tears down monitoring.

Consider making `enabled` `Optional` for the update path (only reconcile when explicitly present), or document/enforce that `PUT /config` requires a full body. The existing `test_update_config_partial` won't catch this — it re-sends the full GET payload.

src/backend/routers/monitoring.py
```diff
@@ -499,6 +526,10 @@ async def disable_monitoring(
 
     Admin only. Stops background health check tasks.
     """
+    # #1121: persist enabled=False FIRST so a disabled fleet stays disabled
+    # across restarts even if the service wasn't running this process.
+    _persist_monitoring_enabled(False)
```
dolho, line 543: **Rapid enable→disable can leave the loop running while persisted `enabled=False`.** `enable` persists `True` then schedules `start_monitoring_service` as a *background task* (runs after the response). If a `disable` request is processed before that background task flips `is_running` to `True`, disable persists `False`, sees `is_running == False`, returns `already_stopped`, and schedules **no** stop. The queued start task then runs and the loop comes up — now running against a persisted `enabled=False`.

Narrow (self-heals on next restart since persisted state is `False`), but it's a real persisted-vs-runtime drift. Starting the loop inline (or guarding stop/start against the persisted flag at task-run time) would close it.

dolho:
## 🔎 Code review (high-effort, recall-biased)

Core fix is solid and verified end-to-end (lifespan resume, default-off, interval validation). **2 findings** posted inline, both in the new `enable/disable/PUT-config` reconcile logic — neither blocks the central restart-survival fix:

1. **medium** — Partial `PUT /config` (body omitting `enabled`) defaults `enabled=False` and the new reconcile **stops a running loop**. A routine interval tweak via a partial body now silently disables monitoring. ([inline](https://github.com/Abilityai/trinity/pull/1123#discussion_r3386843657))
2. **low** — Rapid `enable`→`disable` race: disable can no-op (loop's background start hasn't flipped `is_running` yet), leaving the loop running against persisted `enabled=False`. Self-heals on restart. ([inline](https://github.com/Abilityai/trinity/pull/1123#discussion_r3386843788))

Both stem from deriving the reconcile decision from a model whose `enabled` can be defaulted, and from backgrounding the start. Passed clean: validator correctness, the shared loader/persister dedup (good reuse), `json`/`model_copy` usage, DEFAULT_CONFIG singleton safety, shutdown parity, single-worker assumption (consistent with the sibling lifespan loops).

src/backend/routers/monitoring.py
dolho: Fixed in 470d17b3 — the update path now only treats `enabled` as authoritative when the client explicitly sends it (`model_fields_set`); an omitted `enabled` preserves the persisted run-state, so a partial interval PUT no longer stops the loop. Verified live: `PUT {"docker_check_interval":50}` while running → `enabled` stays true, loop keeps running.

src/backend/routers/monitoring.py
dolho: Fixed in 470d17b3 — enable/disable/PUT now reconcile inline (`await start_monitoring_service`/`stop_monitoring_service`) instead of via background tasks, so runtime state matches the persisted flag before the response returns. start()/stop() only create/cancel the loop task and don't block. Verified: after `enable`, `is_running` is true immediately (no pending-task window).

dolho:
## 🔎 Re-review (after 470d17b3) — both findings resolved, no new bugs

**Finding 1 (partial-PUT disabled monitoring) — fixed & verified.** `update_monitoring_config` now preserves the persisted `enabled` unless the client explicitly sends it (`config.model_fields_set`). A `PUT {"docker_check_interval":50}` while running keeps `enabled=true` and the loop alive (confirmed live).

**Finding 2 (enable→disable race) — fixed & verified.** enable/disable/PUT reconcile inline (`await`) instead of via background tasks. Checked the interleavings are now safe: `start()` has no internal `await` suspension (atomic up to `create_task`), and `stop()` sets `_running=False` *before* it `await`s the task cancellation — so a concurrent enable/disable can't strand the loop against a stale persisted flag. After `enable`, `is_running` is true immediately.

Also clean: removed `background_tasks` params are truly unused (import still needed by `/check-all`); no direct (non-HTTP) callers of the changed signatures; compiles.

### Non-blocking notes (neither introduced by this PR / neither a regression)
- **Partial `PUT /config` still resets non-`enabled` fields** (intervals, thresholds, timeouts) to defaults — the endpoint is full-replace. `enabled` is now specially preserved, so there's a slight asymmetry. Pre-existing PUT semantics; worth a follow-up if partial/PATCH updates are desired, or document that the body must be complete. Not blocking.
- **`status` labels** (`"starting"`/`"stopping"`) are now slightly stale since the work completes synchronously before the response — cosmetic, and the existing tests assert those exact strings.

✅ Ready to merge from my side.

_Re-reviewed with the `code-review` skill._

vybe (approved):
Validated via /validate-pr: lifespan resume + persisted default-off enabled flag verified (persist-first ordering, partial-PUT guard via model_fields_set, inline start/stop closes the enable/disable race, 422 interval validation + sleep clamp). Resolved the architecture.md conflict against the dev restructure (docs-only). CI green on the merge commit. Approving.

Issue #1098, opened 2026-06-08
fix: agent /tmp is a 100MB noexec tmpfs — starves pip/npm/build scratch; default TMPDIR to home volume


vybe:
## Summary

Agent containers are provisioned with `/tmp` as a **100 MB, RAM-backed tmpfs mounted `noexec,nosuid`**. This starves any agent workload that uses `/tmp` for scratch space — `pip install`, `npm install`, compiling C extensions, and especially ML wheels (torch/transformers) — producing `No space left on device` even though the agent's home volume has tens of GB free. Separately, the `noexec` flag breaks tools that compile-and-exec from `/tmp` (`Permission denied`), independent of size.

The mount is hard-coded identically at both agent-provisioning sites:

- `src/backend/services/agent_service/crud.py:724` → `tmpfs={'/tmp': 'noexec,nosuid,size=100m'}`
- `src/backend/services/agent_service/lifecycle.py:539` → `tmpfs={'/tmp': 'noexec,nosuid,size=100m'}`

## Context

Surfaced by an operator trying to set up an agent that builds a Python venv with ML dependencies. The build failed with `No space left on device` despite ~22 GB free on the home volume — because `/tmp` is a separate RAM disk capped at 100 MB. Builds and package installs are core to autonomous agent setup, so this blocks a whole class of agent work (theme: reliability).

**Rejected fix:** simply enlarging the tmpfs to GBs. `/tmp` is RAM-backed, so that consumes host RAM per agent across the fleet, and still leaves `noexec` breaking exec-from-tmp.

**Preferred fix:** set a default `TMPDIR` pointing at a disk-backed, exec-capable path on the agent home volume (e.g. `/home/developer/.tmp`). `pip`, build tooling, and most ML libs honor `TMPDIR`, so this redirects scratch onto real disk and dodges `noexec` in one move — while preserving the hardened `noexec,nosuid` posture on the actual `/tmp`.

## Acceptance Criteria

- [ ] An agent can run a representative heavy install (e.g. building a venv with ML deps) without hitting `No space left on device`.
- [ ] Default `TMPDIR` resolves to a disk-backed, exec-capable path on the agent home volume (e.g. `/home/developer/.tmp`), created on container start.
- [ ] The change is applied at **both** provisioning sites (`crud.py` and `lifecycle.py`) — no drift between create and recreate paths.
[...]

PR #1124, opened 2026-06-10
fix: default agent TMPDIR to disk-backed home volume — /tmp tmpfs starves build scratch (#1098)

dolho:
## Summary

Agent containers mount `/tmp` as a **100 MB RAM-backed tmpfs (`noexec,nosuid`)**. Heavy scratch — `pip`/`npm install`, compiling C extensions, ML wheels (torch/transformers) — overflows it with `No space left on device` even though the home volume has hundreds of GB free, and `noexec` breaks compile-and-exec-from-`/tmp`. This blocks a core class of agent setup work.

**Fix:** set a default `TMPDIR=/home/developer/.tmp` on the disk-backed, exec-capable home volume. `pip`/`npm`/most build tooling honor `TMPDIR`, so this redirects scratch onto real disk and dodges `noexec` in one move — while keeping `/tmp`'s hardened `noexec,nosuid` posture unchanged (rejects the naive "just enlarge the tmpfs", which costs host RAM per agent and leaves `noexec`).

## Changes

- **`services/agent_service/capabilities.py`** — new `AGENT_TMPFS_MOUNT` + `AGENT_DEFAULT_TMPDIR` constants (low-dep module, single source of truth).
- **`crud.py`** (create) — `TMPDIR` in the env dict; `tmpfs=AGENT_TMPFS_MOUNT`.
- **`lifecycle.py`** (recreate) — `env_vars.setdefault('TMPDIR', …)` so a template/user `TMPDIR` carried on the existing container wins; `tmpfs=AGENT_TMPFS_MOUNT`.
- **`system_agent_service.py`** — third provisioning site (not named in the issue, but had the identical literal) wired to the same constants — no drift.
- **`docker/base-image/startup.sh`** — creates the dir (idempotent, `chmod 700`) **before** its own `pip install`, so existing agents pick it up on restart/recreate and tools honoring `TMPDIR` land on disk.
- **`architecture.md`** — note the `TMPDIR` redirect alongside the `/tmp` tmpfs claim.

## Acceptance criteria

- [x] Heavy install no longer hits `No space left on device` (150 MB scratch via `TMPDIR` succeeds; `/tmp` caps at 100 MB).
- [x] Default `TMPDIR` resolves to a disk-backed, exec-capable home-volume path, created on container start.
- [x] Applied at **both** provisioning sites (crud + lifecycle) — and the third (system agent); no drift (shared constants).
- [x] `/tmp` retains `noexec,nosuid` (security posture unchanged); scratch redirected via `TMPDIR`, not by loosening `/tmp`.
- [x] Existing agents pick up the default on restart/recreate (startup.sh mkdir + lifecycle env).
- [x] Templates with their own `TMPDIR` win (recreate uses `setdefault`).

## Rollout note

Touches `startup.sh` → **requires a base-image rebuild** (`scripts/deploy/build-base-image.sh`, which `start.sh` runs). The env injection and the dir-creation ship together; recreate from the rebuilt image lands both. (A new-backend + old-image window where `TMPDIR` is set but the dir is missing is avoided by the standard rebuild-everything deploy.)

## Verification (e2e on a fresh agent, rebuilt base image)

| Check | Result |
|-------|--------|
| Container env | `TMPDIR=/home/developer/.tmp` (backend-injected) |
| Dir auto-created by `startup.sh` | `drwx------ developer developer /home/developer/.tmp` |
| Scratch lands on disk | `tempfile.mkdtemp()` → `/home/developer/.tmp/...`, on the 600 GB-free volume |
| 150 MB scratch via `TMPDIR` | ✅ success (4.5 GB/s) |
| 150 MB into `/tmp` | ❌ caps at exactly 100 MB (`100+0 records out`, no space) — the original symptom |
| `/tmp` posture | still `tmpfs ... noexec,nosuid,size=102400k` |

Throwaway test agent created + deleted; left environment clean.

Related to #1098


docker/base-image/startup.sh
```diff
@@ -12,6 +12,16 @@ if [ -f /opt/trinity/hooks/write-runtime-config.py ]; then
...
 fi
 
+# === Scratch space: ensure TMPDIR exists on the home volume (#1098) ===
+# /tmp is a 100 MB RAM tmpfs mounted noexec,nosuid — too small and non-exec for
+# pip/npm/build scratch (ML wheels hit "No space left on device"). The backend
+# sets TMPDIR to a disk-backed, exec-capable path on the home volume; create it
+# (idempotent) so tools that honor TMPDIR land there. Runs as `developer`, the
+# UID-1000 owner of /home/developer.
+AGENT_TMPDIR="${TMPDIR:-/home/developer/.tmp}"
+mkdir -p "${AGENT_TMPDIR}" 2>/dev/null && chmod 700 "${AGENT_TMPDIR}" 2>/dev/null || \
```
dolho, line 22: **Deploy-coupling (medium):** the backend injects `TMPDIR` (crud/lifecycle/system-agent) but this `mkdir` — the half that makes the dir actually exist — lives in the base image. The two are only in lockstep if the base image is rebuilt with the backend deploy. Since agents run/recreate on the `trinity-agent-base:latest` **tag**, a deploy that updates the backend *without* rebuilding latest gives newly created/recreated agents `TMPDIR=/home/developer/.tmp` pointing at a **missing** dir — and `startup.sh`'s own `pip install --user` (line ~313, which now honors that TMPDIR) plus any TMPDIR-honoring tool fails. That's strictly worse than the status quo for that window.

Mitigated by the standard `start.sh` / `build-base-image.sh` deploy (and documented in the PR), so not a blocker — flagging so the rollout is treated as atomic (rebuild base image + backend together). A belt-and-suspenders option: have the backend tolerate a missing dir, or fall back in old images — though the old image can't be retroactively fixed, so deploy discipline is the real guard.

docker/base-image/startup.sh
```diff
@@ -12,6 +12,16 @@ if [ -f /opt/trinity/hooks/write-runtime-config.py ]; then
...
+    echo "Warning: could not create TMPDIR ${AGENT_TMPDIR}; scratch will fall back to /tmp"
```
dolho, line 23: **Nit:** `mkdir -p X && chmod 700 X || echo warning` — if `mkdir` succeeds but `chmod` fails (e.g. dir pre-exists owned by another uid), the warning fires even though the dir exists, and conversely a successful `mkdir` on a pre-existing dir is the common restart path. Harmless, but the warning text ("could not create TMPDIR") is misleading when only the chmod failed. Minor.

dolho:
## 🔎 Code review (high-effort, recall-biased) — no blocking bugs

Clean refactor (shared constants kill the create/recreate/system-agent drift) + verified e2e. **No correctness bugs survived verification.** 2 inline notes, both non-blocking.

### Checked & cleared
- **Git pollution (the one I most expected) — refuted.** `/home/developer/.tmp` is inside synced agents' git tree, but `*.tmp` (already in `_GITIGNORE_PATTERNS`, applied on init + every push) ignores the `.tmp/` directory *and* its contents — empirically confirmed via `git check-ignore .tmp/` and `git status`. Build scratch won't leak into auto-sync.
- **Shared mutable `AGENT_TMPFS_MOUNT` dict** across three `containers_run` calls — docker-py treats `tmpfs=` read-only (builds HostConfig from it), no mutation. Safe.
- **Recreate image** — agents run the `trinity-agent-base:latest` **tag** (`Config.Image` confirmed), and recreate reuses it, so a rebuilt latest reaches existing agents (AC: "pick up on recreate" holds).
- **`/tmp` posture** unchanged (`noexec,nosuid,size=100m` via the constant — byte-identical to before).
- **`setdefault` in lifecycle** correctly preserves a template/user `TMPDIR` across recreate.

### Notes (non-blocking)
1. **Deploy-coupling (medium)** — backend sets `TMPDIR`, but the dir is created by `startup.sh` in the base image. A backend deploy *without* a base-image rebuild → `TMPDIR` set, dir missing → `startup.sh`'s own `pip install` + TMPDIR-honoring tools fail (worse than today). Mitigated by the standard `start.sh`/`build-base-image.sh` deploy + documented in the PR; flagging so rollout stays atomic. ([inline](https://github.com/Abilityai/trinity/pull/1124#discussion_r3387095369))
2. **Nit** — `mkdir && chmod || echo`: a chmod-only failure prints the "could not create" warning though the dir exists. Cosmetic. ([inline](https://github.com/Abilityai/trinity/pull/1124#discussion_r3387095521))

✅ Ready to merge (rebuild the base image as part of the deploy).

vybe (approved):
Validated via /validate-pr: TMPDIR redirect verified at all three provisioning sites via shared constants (no drift); /tmp posture unchanged (noexec,nosuid,100m); setdefault preserves template-supplied TMPDIR on recreate; coexistence with #1126 nano_cpus change verified post-merge. Architecture.md conflict resolved against dev restructure. Approving.

Issue #1129, opened 2026-06-10
Make 'require verified email' access policy a fleet-wide default config (secure-by-default)


vybe:
## Summary

Make "require verified email" (the access-policy `require_email` flag) a **fleet-wide default config** so new agents are secure-by-default. Today every new agent starts with `agent_ownership.require_email = 0` (anyone allow-listed can chat without proving their email), and an owner has to flip it on per agent. We want an admin-controlled platform default — set to **require verified email by default** — that seeds the flag on agent creation, with per-agent override preserved.

## Context

`require_email` (#311, Unified Channel Access Control) gates incoming DMs / public chat / shared access on a verified email. The column defaults to `0`, so the secure posture is opt-in per agent — easy to forget, and the wrong default for a platform that exposes agents over public channels. Promoting it to an admin platform default (defaulting ON) closes the gap without removing the owner's ability to opt a specific agent out.

This mirrors the existing fleet-wide defaults pattern already in the settings router (`GET/PUT /api/settings/agent-defaults/resources`, RES-001) — same shape: admin-only platform setting consumed at agent-creation time.

## Acceptance Criteria

- [ ] New admin-only platform setting for the fleet-wide default of `require_email` (e.g. `GET/PUT /api/settings/agent-defaults/access-policy`), stored in `system_settings`, defaulting to **require verified email = ON**.
- [ ] Agent creation seeds `agent_ownership.require_email` from this default instead of the hard-coded `0`.
- [ ] Per-agent override is preserved — the existing `PUT /api/agents/{name}/access-policy` continues to let an owner turn it off (or on) for a specific agent; the platform default only sets the initial value.
- [ ] Settings UI exposes the toggle (admin-only), consistent with the existing agent-defaults resources control.
- [ ] Behavior is documented: changing the default affects **newly created** agents only; existing agents keep their current per-agent value (no silent retroactive change to live agents).
- [ ] Backend default flips to ON for fresh installs; existing deployments' already-created agents are unaffected (migration sets the new setting key, does not rewrite per-agent rows).

## Technical Notes

- Setting plumbed through `services/settings_service.py` + `routers/settings.py` (follow the `agent-defaults/resources` precedent at `routers/settings.py:1267`).
- Agent-creation default consumption: wherever the `agent_ownership` row is created during agent provisioning (`services/agent_service/`), read the platform default and pass it as the initial `require_email`.
[...]

Issue #665, opened 2026-05-05
feat: increase default chat execution timeout from 15m to 60m


vybe:
## Summary

The default execution timeout for chat is currently 900 seconds (15 minutes), set via \`agent_ownership.execution_timeout_seconds\`. Many real-world agent tasks — long research runs, multi-step code generation, batch processing — routinely exceed this, causing silent failures that look like success to the user until they notice the truncated output.

Raising the default to 3600 seconds (60 minutes) aligns with the upper bound of what agents are expected to do autonomously and removes a common friction point for new users who haven't yet discovered the per-agent timeout setting.

## Context

- Schema: \`agent_ownership.execution_timeout_seconds INTEGER DEFAULT 900\`
- API: \`GET/PUT /api/agents/{name}/timeout\` (range: 60–7200s)
- Existing per-agent override is correct and should remain; this changes only the default for newly created agents
- Surfaces in both chat (\`/chat\`) and scheduled task execution paths — both read from the same column

## Acceptance Criteria

- [ ] \`agent_ownership.execution_timeout_seconds\` column default updated from \`900\` to \`3600\` in \`db/schema.py\`
- [ ] Migration in \`db/migrations.py\` that updates existing rows where \`execution_timeout_seconds = 900\` (i.e., still at the old default, never explicitly set by a user)
- [ ] \`PUT /api/agents/{name}/timeout\` validation range unchanged (60–7200s)
- [ ] UI timeout picker in Agent Settings reflects the new default for newly created agents
- [ ] Existing agents with a user-set timeout (≠ 900) are unaffected by the migration

## Technical Notes

- Schema default change: \`src/backend/db/schema.py\` — update \`DEFAULT 900\` → \`DEFAULT 3600\`
- Migration target: \`src/backend/db/migrations.py\` — \`UPDATE agent_ownership SET execution_timeout_seconds = 3600 WHERE execution_timeout_seconds = 900\`
- No API surface change needed — range 60–7200 already covers 3600

vybe:
Shipped in **v0.6.0** (released via #998).

Issue #311, opened 2026-04-12
feat: Unified channel access control — verified-email identity, per-agent allow-list, access requests (web / Telegram / Slack)


vybe:
## Problem

Each channel today grows its own identity and access model:

- **Public chat links** (#252): optional email verification, no real allow-list or access-request flow
- **Telegram bots** (#297, #301): zero authentication — anyone who finds the bot can chat
- **Slack**: workspace-gated but no per-agent control
- **Future (WhatsApp #299)**: no model defined

The agent owner has no consistent way to answer: *"who is allowed to talk to this agent?"*

As a result, per-channel auth drifts apart, user memory (MEM-001) can't flow across channels, and there's no single enforcement point for access policy.

## Design principle

**Verified email is the single unit of identity across every channel.** Each adapter is responsible only for translating its native identifier into a verified email. After that, every downstream concern (access control, memory, audit, rate limiting) runs off the same key.

The design reuses existing primitives wherever they already fit and introduces exactly one new table and one new adapter method.

## Reuse map

| Existing primitive | Role in this design |
|---|---|
| `agent_ownership` | Policy lives here — add `require_email`, `open_access` as an agent-settings mixin (architectural invariant #2) |
| `agent_sharing(agent_name, shared_with_email)` | Per-agent allow-list — unchanged, becomes the source of truth for cross-channel access |
| `telegram_chat_links` | Telegram identity binding — add `verified_email`, `verified_at` columns |
| `email_verification_codes` + `db.create_verification` / `db.verify_code` | Telegram `/login` verification — no new verification code |
| `public_chat_sessions.session_identifier` | Web identity binding — already holds email for verified users |
| `public_user_memory(agent_name, user_email)` | Cross-channel memory (MEM-001) — already keyed on email |
| `email_service` | Sends the 6-digit code — unchanged |
| `OwnedAgentByName` dependency | Owner-gates the policy and allow-list editors — unchanged |

## What is actually new

**One table:**

```sql
access_requests(
  id TEXT PRIMARY KEY,
  agent_name TEXT NOT NULL,
  email TEXT NOT NULL,
  requested_at TEXT NOT NULL,
  status TEXT NOT NULL,          -- pending | approved | denied
  decided_by TEXT,               -- user id of approver
  decided_at TEXT,
  UNIQUE(agent_name, email)
);
```

Approval flow: `status=approved` → insert into `agent_sharing` → delete or archive the request.

**One adapter method** (`ChannelAdapter` base):

```python
[...]

PR #1132, opened 2026-06-10
feat(security): fleet-wide default for require-verified-email — secure-by-default (#1129)

dolho:
## Summary

Promotes the per-agent `require_email` access-policy flag (#311) to an **admin-controlled fleet-wide default that defaults ON**, so new agents are **secure-by-default** instead of opt-in per agent. Mirrors the RES-001 `agent-defaults/resources` pattern.

## Changes

- **`settings_service.py`** — `AGENT_DEFAULT_REQUIRE_EMAIL_KEY` + `get_agent_default_require_email()` (default ON; resolved from `system_settings`, with a code fallback so fresh installs are secure **without a data migration**).
- **`routers/settings.py`** — `GET`/`PUT /api/settings/agent-defaults/access-policy` (admin-only) + `AgentDefaultAccessPolicyUpdate` model. Registered before the `/{key}` catch-all (Invariant #4).
- **`register_agent_owner`** (`db/agents.py` + `database.py` facade) — new `require_email` param, written **explicitly** in the INSERT (same SQLite-baked-default reasoning as #665's `execution_timeout_seconds`). Defaults `False`, so the **system agent is unaffected**.
- **`crud.py`** — seeds `require_email` from the platform default at user-agent creation.
- **Settings UI** — admin toggle in the Platform section, consistent with the Default Model control.

## No schema change

The `agent_ownership.require_email` column (#311) and `system_settings` both already exist — **zero DDL**, no migration. The default is read **only at creation time**; existing agents are never rewritten.

## Acceptance criteria

- [x] Admin-only platform setting for the fleet-wide `require_email` default, stored in `system_settings`, **defaulting ON**.
- [x] Agent creation seeds `agent_ownership.require_email` from the default (not hard-coded `0`).
- [x] Per-agent override preserved (`PUT /api/agents/{name}/access-policy` unchanged).
- [x] Settings UI exposes the toggle (admin-only), consistent with agent-defaults resources.
- [x] Default affects **newly created** agents only; existing agents unchanged (no retroactive rewrite).
- [x] Backend default ON for fresh installs via code fallback; no per-agent rows touched.
- Scoped to `require_email` only (per the grooming note); `open_access` left out.

## Verification (live)

| Step | Result |
|------|--------|
| `GET /agent-defaults/access-policy` | `require_email: true` (default ON, key unset) |
| Create agent (default ON) | seeded `require_email: true` |
| Flip default OFF → create agent | seeded `require_email: false` |
| Existing agent after the flip | **unchanged** (still its prior value) |
| Per-agent `PUT /access-policy` | still works (flipped an agent independently) |
| Settings UI | toggle renders under Platform, checked ON by default, no page errors |

Throwaway agents deleted; default restored ON. Backend `py_compile` + frontend `vue/compiler-sfc` clean.

Related to #1129


src/backend/routers/settings.py
```diff
@@ -1337,6 +1340,58 @@ async def update_agent_default_resources(
...
+    Set the fleet-wide default access policy for new agents (#1129).
+
+    Admin-only. Only the fields provided are updated. Stored in system_settings;
+    consumed at agent-creation time. Does NOT rewrite existing agents.
+    """
+    require_admin(current_user)
+
+    updated = []
+    if body.require_email is not None:
+        db.set_setting(AGENT_DEFAULT_REQUIRE_EMAIL_KEY, "1" if body.require_email else "0")
```
dolho, line 1385: **Missing audit log on a security-relevant setting change.** This endpoint writes the fleet-wide email-verification default via `db.set_setting(...)` directly, with no `platform_audit_service.log(...)`. Everywhere else in this router that mutates a setting audit-logs it — API-key change (L222), GitHub PAT (L380), and the generic `PUT /{key}` (L1454) all emit `AuditEventType.CONFIGURATION`. An admin flipping the fleet-wide `require_email` default **OFF** (weakening the platform security posture for every future agent) currently leaves no trace.

It mirrors the `agent-defaults/resources` PUT, which also doesn't audit — but resources isn't security-sensitive and this is (theme-security, P1). Suggest adding an audit call after the write, e.g.:
```python
await platform_audit_service.log(
    event_type=AuditEventType.CONFIGURATION,
    event_action="update_agent_default_access_policy",
    actor_id=str(current_user.id), actor_email=current_user.email,
    details={"require_email": body.require_email}, source="api",
    request=request,
)
```
(make the handler `async`-friendly — it already is). Non-blocking, but worth closing the gap given the security theme.

dolho:
## 🔎 Code review (high-effort, recall-biased) — 1 non-blocking finding

Clean, well-scoped diff that faithfully mirrors the RES-001 agent-defaults pattern; live-verified end-to-end. **No correctness bugs.** One security-theme gap worth closing.

### Finding (low/medium, non-blocking)
- **No audit log on the security-relevant default change** — `PUT /agent-defaults/access-policy` writes via `db.set_setting()` with no `platform_audit_service.log(...)`, while API-key/GitHub-PAT/generic-`{key}` settings writes all audit `AuditEventType.CONFIGURATION`. An admin weakening the fleet-wide `require_email` default OFF leaves no trace. Mirrors the (non-security) resources PUT, but this one is theme-security/P1. ([inline](https://github.com/Abilityai/trinity/pull/1132#discussion_r3388037744))

### Checked & cleared
- **System agent unaffected** — `register_agent_owner(require_email=False)` default; the two system-agent call sites don't pass it (verified). The `IntegrityError` re-register path UPDATEs `is_system` only, never touching `require_email` — correct.
- **Default parsing** — `get_agent_default_require_email`: `None` → code-default ON; `"0"` → False; flexible truthy set is harmless (writer only emits `"1"`/`"0"`).
- **Route ordering** — `/agent-defaults/access-policy` registered before the `/{key}` catch-all (Invariant #4); also distinct by segment count.
- **Facade + callers** — `database.py` forwards `require_email`; all 4 callers consistent; no broken call sites.
- **No lockout** — `require_email` gates incoming channel/shared access per #311; the owner via web JWT and admins aren't gated, so flipping it ON at creation doesn't lock owners out of their own agents.
- **No schema change** (column + `system_settings` pre-exist); existing rows never rewritten — confirmed live (existing agent unchanged after a default flip).

✅ Ready to merge; the audit-log addition is a nice defense-in-depth follow-up.

src/backend/routers/settings.py
dolho: Fixed in 11a5abc1 — added the `platform_audit_service.log(AuditEventType.CONFIGURATION, ...)` call (same shape as the API-key/generic-setting writes) after the `set_setting`. Verified live: flipping the default writes a `settings_change` audit row with `{setting: agent_default_require_email, require_email: <new>}`.

vybe (approved):
Validated via /validate-pr: verified require_admin local-helper pattern, get_setting_value facade method, and that the single user-agent creation path is covered (system-agent sites intentionally excluded). Explicit-INSERT seed reasoning sound (#665 pattern), zero DDL, default flip audit-logged. Resolved the architecture.md conflict against the dev restructure. CI green. Approving.

PR #1133, opened 2026-06-10
build(frontend): vue-router 5.1 + vite 8.0 against dev; dependabot target-branch fix (supersedes #1015)

vybe:
## Summary

Rebuilds dependabot PR #1015 (vue-router 4.6.3 → 5.1.0, plus the Vite 6.4 → 8.0.16 fold-in of #1063) **against `dev`**. The original targeted `main`, but `dev`'s `router/index.js` was restructured by #1109 (Operations consolidation) and still carried 11 `next()`-callback guard sites — merging #1015 to `main` would have guaranteed a conflict (or a wrong-side resolution) at the next release cut.

## Changes

- **`src/frontend/src/router/index.js`** — migrate the global `beforeEach` and the AgentWorkspace `beforeEnter` from the deprecated `next()` callback to Vue Router 5 return-style guards (`true` = proceed, location = redirect). All 11 sites converted on dev's #1109-restructured file; redirect logic byte-for-byte equivalent.
- **`src/frontend/package.json`** — `vue-router ^5.1.0`, `vite ^8.0.16` (only these two lines; dev's other pins untouched).
- **`src/frontend/package-lock.json`** — regenerated in a clean `node:20-alpine` container.
- **`.github/dependabot.yml`** — `target-branch: dev` on all five ecosystems. Dependabot's default (repo default branch = `main`) was creating main-side drift: bumps landed on `main` while features landed on `dev` (e.g. axios `^1.17.0` on main vs `^1.16.1` on dev). All dep PRs now flow through the normal `dev → main` release path.

## Verification

- `npm run build` (Vite 8 / rolldown) green in `node:20-alpine` — same base as both `docker/frontend` Dockerfiles; Node 20.20 satisfies Vite 8's ≥20.19 floor. Full app builds in ~2s.
- Lockfile resolves `vite 8.0.16` / `vue-router 5.1.0` / `vue 3.5.35` / `@vitejs/plugin-vue 6.0.7`.
- `grep -n 'next(' router/index.js` → zero remaining callback-style sites.
- `ui` label applied so `frontend-e2e` runs on this PR (it was SKIPPED on #1015 — a router-major shipped with no browser-level verification).

Supersedes #1015. Related: #1063, #1109.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1114, opened 2026-06-09
Agent Detail tabs: overflow into a "More" dropdown instead of horizontal scroll


vybe:
## Summary

The Agent Detail page (`src/frontend/src/views/AgentDetail.vue`) now has many tabs (Overview, Chat, Session, Tasks, Dashboard, Schedules, Loops, Playbooks, Git, Files, Skills, Folders, Sharing, Permissions, Guardrails, Info, …). On narrower viewports the tab bar overflows horizontally and currently relies on `overflow-x-auto` horizontal scrolling, which is easy to miss and awkward to use. Replace the horizontal scroll with a responsive layout that keeps as many tabs visible as fit and collapses the remainder into an overflow **"More ▾"** dropdown on the right edge.

## Context

The tab nav is rendered at `src/frontend/src/views/AgentDetail.vue` (currently `<div class="… overflow-x-auto …">` wrapping a `<nav class="-mb-px flex whitespace-nowrap">` over `visibleTabs`). Horizontal scroll bars are a poor discovery mechanism — tabs that don't fit are effectively hidden. A "More" overflow menu is the conventional pattern (e.g. browser tab strips, GitHub's own nav) and keeps every tab reachable without scrolling.

## Acceptance Criteria

- [ ] When the tab bar has enough horizontal space, all tabs render inline exactly as today (no visual regression).
- [ ] When tabs don't fit, the overflowing tabs collapse into a right-aligned **"More ▾"** dropdown instead of horizontal scroll; clicking a tab inside the dropdown activates it.
- [ ] The dropdown re-measures and reflows on window/container resize (e.g. via `ResizeObserver`), not just on initial render.
- [ ] If the currently-active tab would land in the overflow set, it stays visible (or the "More" trigger reflects the active state) so the user always sees which tab is selected.
- [ ] Tab badges (e.g. the success-count badge) render correctly both inline and inside the dropdown.
- [ ] Works in dark mode and on touch; keyboard-accessible (dropdown openable/closable and items focusable).
- [ ] Deep-linking via `?tab=` still selects the correct tab whether it's inline or in the overflow menu.

## Technical Notes

- Source: `src/frontend/src/views/AgentDetail.vue` — the `<nav>` over `visibleTabs` and its `overflow-x-auto` wrapper.
[...]

Issue #1109, opened 2026-06-09
refactor(ui): unify Health + Ops + Executions into a single Operations nav area


vybe:
## Summary

Consolidate the three separate operational top-nav entries — **Health** (`/monitoring`), **Ops** (`/operating-room`), and **Executions** (`/executions`) — into a single top-level **"Operations"** area, implemented as a tabbed view. This finishes a consolidation the codebase already started (Events was previously folded into the Operating Room) and removes the overlapping, duplicate-badge top menu.

## Context

The top navigation (`src/frontend/src/components/NavBar.vue`) currently shows, for an admin:

`Dashboard · Agents · Templates · Health · Ops · Executions · Settings · [Enterprise]`

**Three of those are the same job** — *how is the fleet running right now?*:

| Nav item | Route | View | What it shows |
|----------|-------|------|---------------|
| **Health** | `/monitoring` (admin-only) | `Monitoring.vue` | Fleet health summary cards, alerts, per-agent health grid, trigger health check |
| **Ops** | `/operating-room` | `OperatingRoom.vue` | **Already tabbed**: Needs Response (operator queue) · Notifications · Resolved |
| **Executions** | `/executions` | `Executions.vue` | Fleet execution list, stat cards (total / success rate / failed / cost), filters |

Problems this creates:
1. **Three top-level entries for one concern** — operational state is fragmented across Health, Ops, and Executions.
2. **The consolidation pattern already exists, half-done** — the router already redirects `/events → /operating-room?tab=notifications` and `/api-keys → /settings?tab=mcp-keys`. `OperatingRoom.vue` is already a tabbed container (tabs via `?tab=` query param, `VALID_TABS`). Extending it to absorb Health and Executions continues an established pattern rather than inventing one.
3. **Duplicate badges** — "Ops" carries a combined badge (`operatorQueueStore.pendingCount + notificationsStore.pendingCount`); "Executions" carries a separate `executionsStore.runningCount` badge. Two operational counters compete in the same bar.
4. **Inconsistent gating** — Health is admin-only (`requiresAdmin`) but Ops and Executions are not, even though all three are operational surfaces.

## Proposed change

Rename and extend the existing Operating Room into a single **Operations** area:

```
Before:  Dashboard · Agents · Templates · Health · Ops · Executions · Settings
After:   Dashboard · Agents · Templates · Operations · Settings
                                          └─ tabs: Needs Response · Notifications · Health · Executions · Resolved
```

[...]

Issue #1112, opened 2026-06-09
feat(ui): unify Chat + Session into one Chat tab with a session-mode toggle (default on)


vybe:
## Summary

Unify the two chat-like tabs on the Agent Detail page into **one "Chat" tab** that carries a **"Session mode" toggle, on by default**. Instead of hiding/removing the legacy Chat surface (the prior plan), keep both behaviors reachable from a single tab:

- **Session mode ON (default)** → renders `SessionPanel.vue`: `--resume`-default chat that preserves working memory, tool-result state, and reasoning across turns.
- **Session mode OFF** → renders the legacy `ChatPanel.vue`: stateless, ephemeral per-turn chat.

The user flips between them with an in-tab toggle; the tab is always labeled **"Chat"**. This consolidates the redundant tabs (called out in tab-overhaul spike #966) while preserving the legacy stateless surface as a first-class, user-selectable mode rather than dead code.

## Context

The Session tab (SESSION_TAB_2026-04) has been GA and **default-on since 2026-05-04**. Each turn reattaches to the same Claude Code session via `claude --print --resume <uuid>`, preserving memory across turns — generally the better default. But the legacy ephemeral surface still has value (clean-slate turns, no resume-lock contention, debugging), so rather than retire it we expose it as the **off** state of a per-user toggle.

Validated against the code:
- Default landing tab is already `tasks` (`src/frontend/src/views/AgentDetail.vue` — `activeTab = ref('tasks')`), so collapsing Chat/Session into one tab does not change the default view.
- The Session surface is gated by `sessionsStore.sessionTabEnabled` (from `GET /api/settings/feature-flags` → `session_tab_enabled`), settable to `false` platform-wide. When the flag is off, the toggle must not appear and the single Chat tab falls back to the legacy panel (never zero chat surfaces).
- Two surfaces deep-link into the legacy Chat tab and must be repointed to the unified tab:
  - `src/frontend/src/views/ExecutionDetail.vue` routes with `{ tab: 'chat', resumeSessionId: <claude_session_id> }`.
  - `src/frontend/src/views/MobileAdmin.vue` `openChat(agent)`.
- A `?tab=` query param mechanism exists (validTabs lists currently include both `'chat'` and `'session'`), so existing `?tab=chat` and `?tab=session` URLs must both resolve to the single unified Chat tab.

## Acceptance Criteria

[...]

PR #1135, opened 2026-06-10
fix(e2e): repair smoke drift from #1109 + neutralize ensure-default modal race (#1134)

### LINKED ISSUE #1134: e2e smoke drift: Operations consolidation (#1109) broke operating-room smoke; ensure-default modal races api-keys spec
Two pre-existing `@smoke` failures have made `frontend-e2e` red on dev since 2026-06-10 (first seen on the #1114 run; reproduced identically on #1133). Because the workflow is `ui`-label-gated, several PRs merged while the suite was silently red.

## 1. `smoke.spec.js` — operating room page loads

The test navigates `/operating-room` and asserts `getByText(/operating|queue|priority|all types|no items/i)`. Since #1109 the route redirects to `/operations`, whose visible text (heading **Operations**, tabs **Needs Response / Notifications / Health / Executions / Resolved**, empty state "All clear — your agents are working independently") matches none of those patterns. The redirect itself works — pure assertion drift.

## 2. `api-keys-copy.spec.js` — Copy key icon button (race)

`McpKeysTab.vue` `onMounted` awaits `fetchApiKeys()` then fires `POST /api/mcp/keys/ensure-default`; when no user-scoped key exists it auto-creates **Default MCP Key** and opens the "Your MCP API Key is Ready!" modal. In CI the test deletes its own key per-test, so test 2's page load re-creates the default key and the auto-modal lands **on top of the test's Create modal** (both `fixed z-10 inset-0`; later DOM node wins) — the Create click is intercepted for 30s (screenshot in the run artifacts shows the auto-modal covering the form). Timing was marginal; the #1107/#1109 UI changes tipped it into failing consistently. Note test 1's clipboard assertions can pass against the *auto* modal (same buttons, format-only assertions), so it green-washes the race.

## Fix (test-only)

- operating-room smoke: assert the post-#1109 reality — URL `/operations` + heading + a tab button (doubles as coverage for the legacy redirect).
- api-keys spec: `beforeEach` calls `ensure-default` via `page.request` (token from localStorage) and reloads — the page's own `ensure-default` is then a guaranteed no-op and the auto-modal cannot appear mid-test.

Found while validating #1133 (vue-router 5 + vite 8): identical failures on vanilla dev (#1114 run) rule out the router bump as the cause.

vybe:
## Summary

`frontend-e2e` has been silently red on dev since 2026-06-10 — first failure on the #1114 run, before #1133 existed — because the workflow only runs on `ui`-labelled PRs and recent UI PRs merged without the label. Two pre-existing test issues, both test-only fixes:

## 1. `smoke.spec.js` — operating room (assertion drift from #1109)

`/operating-room` redirects to `/operations` since #1109; the page's visible text (heading **Operations**, tabs, "All clear…") matches none of the old `/operating|queue|priority|all types|no items/` patterns. The redirect itself works fine — verified via run artifacts. New assertions: redirect URL + Operations heading + `Needs Response` tab, so the test now also covers the legacy redirect.

## 2. `api-keys-copy.spec.js` — ensure-default modal race

`McpKeysTab.vue` `onMounted` fires `POST /api/mcp/keys/ensure-default`; with no user-scoped key present it auto-creates **Default MCP Key** and opens the "Your MCP API Key is Ready!" modal at an arbitrary point mid-test. Both that modal and the test's Create modal are `fixed z-10 inset-0`, so the later DOM node (auto-modal) intercepts the Create click for the full 30s budget — run-artifact screenshot shows the auto-modal covering the form. `beforeEach` now calls `ensure-default` via `page.request` and reloads, making the remounted page's own call a guaranteed no-op. Bonus: test 1's clipboard assertions previously could pass against the *auto* modal (same buttons, format-only checks); they now always exercise the test's own key modal.

## Verification

- Root cause evidence: identical failures on the #1114 run (vanilla vue-router 4 / vite 6) and the #1133 run — artifacts (a11y snapshot + screenshot) show the auto-modal overlay and the Operations page text.
- This PR carries the `ui` label, so `frontend-e2e` runs here and must be green to merge.

Unblocks #1133 (which re-runs e2e on top of this fix).

Fixes #1134

### DIFF
--- a/src/frontend/e2e/api-keys-copy.spec.js
+++ b/src/frontend/e2e/api-keys-copy.spec.js
@@ -18,9 +18,28 @@ import { test, expect } from '@playwright/test'
 const MCP_KEYS_ROUTE = '/settings?tab=mcp-keys'
 
 test.describe('@smoke api-keys copy buttons (#677, #859)', () => {
-  test.beforeEach(async ({ context }) => {
+  test.beforeEach(async ({ context, page }) => {
     // Headless browsers gate clipboard read by default — opt in for the test.
     await context.grantPermissions(['clipboard-read', 'clipboard-write'])
+
+    // #1134: McpKeysTab's onMounted fires POST /api/mcp/keys/ensure-default;
+    // when no user-scoped key exists it auto-creates "Default MCP Key" and
+    // opens the "Your MCP API Key is Ready!" modal at an arbitrary point
+    // mid-test — both modals are `fixed z-10 inset-0`, so the auto-modal
+    // lands on top and intercepts the Create click. Neutralize the race:
+    // call ensure-default ourselves, then reload — the remounted page's own
+    // ensure-default is a guaranteed no-op, so the auto-modal can never
+    // appear during the test body.
+    await page.goto(MCP_KEYS_ROUTE)
+    const token = await page.evaluate(() => localStorage.getItem('token'))
+    if (token) {
+      await page.request
+        .post('/api/mcp/keys/ensure-default', {
+          headers: { Authorization: `Bearer ${token}` },
+        })
+        .catch(() => {})
+    }
+    await page.reload()
   })
 
   test('Copy Config button writes MCP JSON to clipboard', async ({ page }) => {
--- a/src/frontend/e2e/smoke.spec.js
+++ b/src/frontend/e2e/smoke.spec.js
@@ -25,11 +25,13 @@ test.describe('smoke', () => {
   })
 
   test('@smoke operating room page loads', async ({ page }) => {
+    // #1109/#1134: /operating-room is a legacy redirect to /operations.
+    // Navigating the old path also covers the redirect itself.
     await page.goto('/operating-room')
-    // Either a queue list, filters, an empty state, or the title.
-    await expect(
-      page.getByText(/operating|queue|priority|all types|no items/i).first()
-    ).toBeVisible({ timeout: 10000 })
+    await expect(page).toHaveURL(/\/operations/, { timeout: 10000 })
+    await expect(page.getByRole('heading', { name: 'Operations' })).toBeVisible({ timeout: 10000 })
+    // Tab strip confirms the view mounted (not just the route resolved).
+    await expect(page.getByRole('button', { name: 'Needs Response' })).toBeVisible()
   })
 
   test('@smoke templates page loads', async ({ page }) => {


vybe:
⏳ **Needs a rebase onto `dev`.** This branch is ~17 days behind `dev` (last touched 2026-06-11). Since it specifically repairs **e2e smoke drift from #1109**, it has to validate against the *current* tree to mean anything — `dev` has since moved (e.g. #1112 unified the Chat/Session tab, more Operations IA churn).

I've re-triggered the failed **CodeQL** run, but it's analyzing the stale base. Please rebase onto `dev` and let CI re-run before this is merge-ready.

vybe:
✅ **Rebased onto `dev`** (force-pushed). Clean replay of the single commit onto the current tip — only the 2 e2e specs (`smoke.spec.js`, `api-keys-copy.spec.js`) changed. CI re-triggered against today's `dev`.

dolho:
## Review — LGTM ✅

Test-only fix (26/5), all CI green including the `frontend-e2e` `e2e` job — the actual proof, since this PR carries the `ui` label so the suite really ran here.

Verified every claim:
- localStorage key `'token'` is correct (`stores/auth.js:284`).
- `POST /api/mcp/keys/ensure-default` exists (`components/settings/McpKeysTab.vue:405`).
- Auth is present in `beforeEach`: `storageState` from `auth.setup.js` (`admin.json`) loads before `page.goto`, so the token is readable.
- Endpoint is idempotent → pre-creating the default key + reload makes the remounted page's own `ensure-default` a guaranteed no-op → the auto-modal can't pop mid-test.
- Base = `dev`.

**smoke.spec.js** — old `/operating|queue|priority/` text is gone after the #1109 redirect to `/operations`. New assertions (redirect URL + `Operations` heading + `Needs Response` tab) are strictly stronger than the old any-text check and also cover the legacy redirect.

**api-keys-copy.spec.js** — race root cause is accurate: two `fixed z-10 inset-0` modals, the later DOM node (auto-modal) intercepts the Create click. `.catch(() => {})` on the pre-call is safe — a failure degrades to the original flake, never a new failure.

Minor nit (non-blocking): the initial `page.goto` also fires a client-side `ensure-default` on mount before the reload — harmless (idempotent + reload discards that modal), just a slightly redundant double-fetch. `goto` is required to read localStorage, so the current form is fine.

No changes needed. Ready to merge.

dolho (approved):

Issue #1137, opened 2026-06-10
fix(agent-server): Gemini model allowlist offers retired models (gemini-2.0-flash, gemini-1.5-*)


vybe:
## Summary

Follow-up from #1130. The agent-server model endpoints in `docker/base-image/agent_server/routers/chat.py` still reference Gemini models Google has retired:

- Line ~196: `GET /api/model` advertises `gemini-2.0-flash` in `available_models`
- Line ~217: `PUT /api/model` `valid_models` allowlist includes `gemini-2.0-flash`, `gemini-1.5-pro`, `gemini-1.5-flash` — a user selecting one of these on a gemini-cli runtime agent gets a 404 from Google at runtime

Validation also accepts any `gemini-*` prefix, so the list is partially cosmetic — but it actively *suggests* dead models.

## Why not fixed in #1130

The #1130 PR fixed the backend's hardcoded call paths (Telegram transcription + image-gen prompt refinement). The agent-server lists are a different deploy surface: changing them requires a base-image rebuild (`./scripts/deploy/build-base-image.sh`) and only affects newly created/restarted agent containers.

## Suggested fix

- Remove retired models from both lists; add current ones (`gemini-3.5-flash`, `gemini-3-flash-preview`, `gemini-2.5-pro`, `gemini-2.5-flash`)
- Consider noting in the `gemini_runtime.py` pricing table which entries are historical

## Acceptance Criteria

- [ ] `GET /api/model` no longer advertises retired models
- [ ] `PUT /api/model` allowlist contains only currently-available models
- [ ] Base image rebuilt and verified

Issue #1130, opened 2026-06-10
bug: Telegram voice transcription + image-gen prompt refinement fail — hardcoded retired Gemini model gemini-2.0-flash returns 404


vybe:
## Summary

Telegram **voice-message transcription** (#401) fails on every voice message because the backend calls the Gemini API with the model `gemini-2.0-flash`, which Google has now retired. The API returns `404 NOT_FOUND` and the user receives the placeholder `[Voice message received — transcription failed]` instead of a transcript. The **same retired model** is also hardcoded in the image-generation service's prompt-refinement path, so that silently degrades too. The model name is a hardcoded string literal in both places with no env-var override, so operators cannot work around it via configuration.

## Component

Backend — Telegram media (`telegram_media`) and Image generation service (`image_generation_service`)

## Priority

P1 — a shipped feature (Telegram voice transcription) is fully broken with no operator workaround; image-gen prompt refinement is degraded by the same cause.

## Error

Transcription path:

```
ERROR services.telegram_media: Gemini transcription error: 404 NOT_FOUND.
{'error': {'code': 404,
  'message': 'This model models/gemini-2.0-flash is no longer available.
              Please update your code to use a newer model for the latest
              features and improvements.',
  'status': 'NOT_FOUND'}}
  File ".../services/telegram_media.py", line 231, in _transcribe_audio_gemini
    response = await client.aio.models.generate_content(
  ...
google.genai.errors.ClientError: 404 NOT_FOUND.

INFO adapters.message_router: [ROUTER:telegram] Voice transcribed:
  [Voice message received — transcription failed]...
```

Image-generation path (same root cause):

```
WARNING services.image_generation_service: Prompt refinement failed, using raw prompt:
  Gemini text API error 404: This model models/gemini-2.0-flash is no longer available.
```

## Location

- **File**: `src/backend/services/telegram_media.py`
- **Line**: ~232
- **Function**: `_transcribe_audio_gemini` — `model="gemini-2.0-flash"`

and

- **File**: `src/backend/services/image_generation_service.py`
- **Line**: ~29
- **Symbol**: `GEMINI_TEXT_MODEL = "gemini-2.0-flash"`

## Root Cause

`gemini-2.0-flash` has been retired by Google and now returns `404 NOT_FOUND`. The model id is hardcoded as a string literal in both services; there is no environment-variable override, so the failure cannot be mitigated from `.env` or a redeploy. Every Telegram voice message and every image-generation prompt-refinement call hits the dead model.

[...]

PR #1144, opened 2026-06-11
fix(agent-server): drop retired Gemini models from /api/model allowlist (#1137)

dolho:
## Summary

Follow-up from #1130. The agent-server model endpoints in `docker/base-image/agent_server/routers/chat.py` still advertised retired Gemini models — a `gemini-cli` runtime agent could pick one and get a 404 from Google at runtime.

## Changes

- **`GET /api/model`** `available_models`: `gemini-2.0-flash` → live set `[gemini-3-pro, gemini-3-flash, gemini-2.5-pro, gemini-2.5-flash]`; note updated.
- **`PUT /api/model`** `valid_models`: dropped `gemini-2.0-flash`, `gemini-1.5-pro`, `gemini-1.5-flash`; kept only the four live models. 400 hint updated. The `startswith("gemini-")` forward-compat fallback is intentionally retained.
- **`gemini_runtime.py`** pricing table: `2.0-flash` / `2.0-flash-lite` commented as historical — kept so cost calc of *older* executions still resolves, not offered in the lists.

## Decision note

Model set sourced from the runtime's own `GEMINI_PRICING` table (agent default `gemini-3-flash`), **not** the issue's parenthetical guesses (`gemini-3.5-flash`, `gemini-3-flash-preview`) — those don't exist in the runtime.

## Verification

- `py_compile` clean on both files.
- Base image rebuilt: `trinity-agent-base:0.6.0` / `:latest`.
- New lists confirmed inside the built image (`grep` on `/app/agent_server/routers/chat.py`) — no retired models present.

## AC

- [x] `GET /api/model` no longer advertises retired models
- [x] `PUT /api/model` allowlist contains only currently-available models
- [x] Base image rebuilt and verified

Related to #1137

### DIFF
--- a/docker/base-image/agent_server/routers/chat.py
+++ b/docker/base-image/agent_server/routers/chat.py
@@ -193,8 +193,8 @@ async def get_model():
         return {
             "model": agent_state.current_model,
             "runtime": runtime,
-            "available_models": ["gemini-2.5-pro", "gemini-2.5-flash", "gemini-2.0-flash"],
-            "note": "Gemini models. 2.5-pro has 1M context window."
+            "available_models": ["gemini-3-pro", "gemini-3-flash", "gemini-2.5-pro", "gemini-2.5-flash"],
+            "note": "Gemini models. 3-pro is the most capable; 3-flash is the fast default."
         }
     else:
         return {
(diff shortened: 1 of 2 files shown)


vybe (approved):
Validated via /validate-pr: focused 2-file fix, CI green, security clean, historical pricing entry correctly retained for cost calc of older executions. Approving.

PR #1145, opened 2026-06-11
fix(notifications): true pending count for NavBar badge — not page-capped (#1143)

### LINKED ISSUE #1143: bug: NavBar notification badge polls with limit=1 so count is capped at 1
## Summary

`stores/notifications.js:fetchPendingCount()` calls `GET /api/notifications?status=pending&limit=1` and reads `response.data.count` — but the endpoint returns `count = len(returned notifications)`, which is capped by `limit`. The polled NavBar badge therefore never shows more than **1** regardless of how many pending notifications exist.

WebSocket-driven increments (`addNotification`) can push it higher between polls, but every 60s poll clamps it back to 0/1.

## Fix options
- Add a true count endpoint (e.g. `GET /api/notifications/count?status=pending`, mirroring the per-agent `/api/agents/{name}/notifications/count`), or
- Make the list endpoint return the unlimited total alongside the page.

## Context

Found during #1017 (Clear All) review. The clear flow itself is unaffected (count drops to 0 correctly).

dolho:
## Summary

`stores/notifications.js:fetchPendingCount()` read the `count` field off `GET /api/notifications?status=pending&limit=1`, but that `count` is `len(returned page)` — capped by `limit`. The polled NavBar badge therefore clamped back to **1** on every 60s poll, no matter how many pending notifications existed. (WS `addNotification` could push it higher between polls, then the next poll reset it.)

## Fix

- **New endpoint** `GET /api/notifications/count?status=pending` — returns the true total scoped to the caller's accessible agents, mirroring the per-agent `/api/agents/{name}/notifications/count`. Registered **before** the `/{notification_id}` catch-all (Invariant #4).
- `db.count_pending_notifications` gains an `agent_names` list filter for the fleet count over accessible agents. Empty list → `0` (not "all"), guarding against invalid `IN ()` SQL.
- `fetchPendingCount()` now hits the new endpoint.

Currently only `status=pending` is countable at the DB layer (the only badge case); other statuses 400 explicitly rather than silently miscounting.

## Verification (live instance)

| Check | Result |
|-------|--------|
| `GET /notifications/count?status=pending` | **32** (true total) |
| Old `?status=pending&limit=1` `count` | **1** (the bug) |
| `GET /notifications/{unknown-id}` | 404 — `/count` does not shadow the param route |
| `status=acknowledged` / `status=bogus` | 400 |
| no token | 401 |

## Scope

Bug fix — no architecture.md/requirements change (notifications endpoints aren't documented there). Touches 4 files: db method, facade passthrough, router, store.

Closes #1143


vybe (approved):
Validated via /validate-pr: root cause correct (list endpoint count is page-capped), new /count endpoint mirrors list-endpoint access scoping, route ordering (Invariant #4) and parameterized IN clause verified, empty-accessible-set fails closed. CI fully green across all 6 pytest seeds + regression diff. Live verification documented in PR body.

Issue #1094, opened 2026-06-08
bug: stall-watchdog process-group kill is mislabeled as "timed out after N seconds"


vybe:
## Summary

When the headless executor's per-tool **no-output stall watchdog** kills a task (a single tool call produces no output for >300s → "killing process group"), the execution's persisted `error` is set to the generic **max-duration timeout** message ("Task execution timed out after N seconds") instead of a stall-specific reason. The two are distinct termination causes that collapse into one stored error, so operators reading the execution record misdiagnose the failure and apply the wrong remediation (bumping the execution timeout, which does nothing because the ceiling that actually fired was the 300s no-output threshold).

A tell-tale symptom: the stored timeout value does **not** match elapsed wall-clock. In the observed case the task ran ~531s but was recorded as "timed out after 2400 seconds" — confirming the message is a templated label, not computed from the real termination cause.

## Component

Agent Runtime — `agent_server.services.headless_executor`

## Priority

P2 (misleading diagnostics → wrong remediation; workaround is to read the agent container logs)

## Error

Persisted on the execution record (`schedule_executions.error`):
```
Task execution timed out after 2400 seconds
```

Actual cause, from the agent container logs (the stall-kill fires first, then the generic timeout label is stamped):
```
ERROR agent_server.services.headless_executor: [Headless Task] Task <execution_id> tool 'Bash' stalled with no result for >300s — killing process group
ERROR agent_server.services.headless_executor: [Headless Task] Task <execution_id> timed out after 2400s
```

Wall-clock for this execution: started → completed = ~531s, i.e. nowhere near the 2400s the error claims.

## Location

- **File**: `src/agent-server/agent_server/services/headless_executor.py` (module `agent_server.services.headless_executor` per the log tag; exact path/line not confirmed from outside the container)
- **Function/Class**: the per-tool stall watchdog ("tool '<name>' stalled with no result for >300s — killing process group") and the terminal error-classification that follows the process-group kill

## Root Cause

The stall watchdog and the max-duration timeout are two different termination paths, but the terminal error written to the execution is the generic "timed out after {configured_max}s" string regardless of which path fired. The stall-kill reason is logged but not propagated into the persisted execution error, so a no-output kill at 300s is indistinguishable (in the record) from genuinely exhausting the 2400s budget.

## Reproduction Steps

[...]

Issue #1083, opened 2026-06-05
refactor: fire-and-forget dispatch — a hung turn holds zero backend resource


vybe:
## Summary

Decouple an execution's slot/lifetime from a blocking backend HTTP call. Dispatching a task to an agent must not pin a backend coroutine for the full multi-minute turn; a hung turn should hold an **agent-side** resource, not a backend one.

## Context

Bankable reliability win #2 from the 2026-06-05 coordination review. Today a wedged turn (e.g. a hung MCP call — the "Cornelius" class) holds a backend slot until timeout, which trips the dispatch breaker and cascades. Removing the backend-thread pinning kills that runaway by construction, independent of the full push→pull migration.

## Acceptance Criteria

- [ ] Backend dispatch no longer blocks a coroutine for the duration of an agent turn
- [ ] A hung/unresponsive turn consumes no backend slot; backend capacity is unaffected
- [ ] Terminal state is applied from the agent's result report under the CAS guard
- [ ] A single slow/hung turn does not trip the dispatch breaker

## Technical Notes

- Relevant: `src/backend/services/task_execution_service.py`, `src/backend/services/agent_client.py`, `src/backend/services/capacity_manager.py`.
- Valuable on its own; under pull this becomes the natural shape (worker pulls, backend never pushes).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

dolho:
## Design proposed (review before code) — `docs/planning/DISPATCH_FIRE_AND_FORGET_1083.md`

Full design doc pushed to `feature/1083-fire-and-forget-dispatch`. Summary for review:

**What blocks today:** `execute_task` (task_execution_service.py:746) `await`s the agent's synchronous `/api/task` for the whole turn (≤~3600s), pinning a backend coroutine + a capacity slot + feeding the dispatch breaker on timeout — the Cornelius runaway.

**Target flow:** agent `/api/task` returns **202** and runs the turn in the background → on finish POSTs the result envelope to a new `POST /api/internal/executions/{id}/result` → backend callback does the **CAS-guarded terminal write** + transcript persist + breaker outcome + `capacity.release(eid)`. `execute_task` returns right after the ACK (no coroutine held). The slot becomes a **lease** released by the callback or reclaimed by the existing TTL reaper (canary E-01 already guards it).

[...]

Issue #970, opened 2026-05-28
bug: headless_executor reader thread leak causes 2h execution hangs


vybe:
## Summary

When an agent's headless task spawns grandchild processes (stdio MCP servers, hooks, subagents) that keep the stdout pipe FD open after the parent Claude process exits, the executor's stdout reader thread blocks indefinitely on `readline()`. The drain helper's force-close fallback does not unblock the thread, marks it `outcome=leaked`, and the leaked thread holds the executor worker — so `_run_headless_subprocess` never returns. The outer `asyncio.wait_for(timeout=effective_timeout+60)` is what eventually fires, causing every affected scheduled execution to consume its full timeout budget (commonly 7200s) even though Claude finished within seconds. From the platform's view this looks like a 2-hour execution; in reality no work is happening.

## Component

Backend / Agent Runtime — `agent_server.utils.subprocess_pgroup._drain_bounded` + `agent_server.services.headless_executor._run_headless_subprocess`

## Priority

P1 — major scheduling impact (slot saturation, false 2h timeouts, backend HTTP timeouts), but a configuration workaround exists.

## Evidence

On one agent's container over ~47h uptime, drain outcomes were:

| outcome | count |
|---|---|
| natural | 10 |
| force_close | 3 |
| **leaked** | **24** |

`drain_elapsed_ms` on the leaked outcomes ranged from ~59,000 ms up to ~4,441,000 ms (≈74 min). Every leaked drain corresponded one-to-one to a scheduled execution that the platform marked `failed: Task execution timed out after 7200 seconds` — i.e. the agent's HTTP response never came back.

Sample log sequence from one affected execution:

```
INFO  [Headless Task] Starting task X: claude --print --output-format stream-json --verbose...
INFO  [Headless Task] Permission mode confirmed: bypassPermissions
WARN  [Subprocess] Reader thread(s) still busy after process exit (pid=X, stuck_count=1) — killing process group, then waiting 30s for natural drain
ERROR [Subprocess] Reader thread(s) still stuck after 349.3s post-kill grace — force-closing pipes; some buffered data may be lost (pid=X, stuck_count=1)
ERROR [Subprocess] 1 reader thread(s) leaked for pid=X after force-close; continuing anyway
INFO  [METRIC] drain_outcome pid=X pgid=X orphan_kill_count=0 outcome=leaked drain_elapsed_ms=4247022 stuck_initial=1 leaked_count=1
```

[...]

Issue #520, opened 2026-04-26
bug: agent_server reports task as 'completed successfully' when claude exits 0 with no result message


vybe:
## Summary

When the `claude` headless subprocess exits cleanly (`return_code == 0`) but its stdout reader thread leaks before capturing the final `{"type":"result"}` block, agent-server logs the task as "completed successfully" and returns 200 to the backend — even though no usable result was captured (cost is `None`, duration is `None`, no response body). The execution then sits in `running` until the orphan watchdog reaps it minutes later. Operators see a misleading "Execution completed on agent but status not reported — recovered by watchdog" error against an execution that the agent itself believed was successful.

This is a sibling to #516 (which I filed earlier) — same code file, same root condition (orphaned reader thread leaks the result block), different exit path. #516/#517 covered the signal-killed case (`return_code < 0`); this is the clean-exit case (`return_code == 0`).

## Component

Agent Runtime (`agent_server.services.claude_code`, `agent_server.routers.chat`)

## Priority

P2 — masks real failures with false "success" status; watchdog provides eventual cleanup so it's not silent data loss, but the diagnostic UX is broken (agent log says success, backend log says orphan-recovered) and the lost ~26 min of work happens silently.

## Error

Inside the agent container (typical sequence around the failure):

```
[Headless Task] Starting task <task_id>: claude --print --output-format stream-json --verbose...
[ProcessRegistry] Registered execution <task_id>
... agent does real work for ~20 min, 7+ tool calls, 22+ raw_messages stream through ...
[Subprocess] Reader thread(s) stuck after process exit (pid=NNNNN, stuck_count=1) — killing process group and closing pipes to unwind
[Subprocess] 1 reader thread(s) leaked for pid=NNNNN after close+killpg; continuing anyway
[Headless Task] Error reading stdout: I/O operation on closed file.
[Headless Task] Task <task_id> completed: cost=$None, duration=Nonems, tools=7, raw_messages=22
[ProcessRegistry] Unregistered execution <task_id>
[Task] Task <task_id> completed successfully     ← misleading: there is no result
```

Meanwhile in the backend / scheduler (same execution, same window):

```
[Slots] Agent '<agent>' acquired slot 1/3 for execution <task_id> (TTL=3900s)
... ~26 min of polling ...
scheduler.service: Execution <task_id> still running (1500s elapsed, poll #150)
scheduler.service: Execution <task_id> still running (1560s elapsed, poll #156)
[...]

PR #1147, opened 2026-06-11
fix(agent-server): distinct error for stall-watchdog kill vs max-duration timeout (#1094)

dolho:
## Summary

The headless executor's per-tool **no-output stall watchdog** (300s) and the **max-duration budget** are two distinct termination paths, but both raised a bare `subprocess.TimeoutExpired` and the terminal 504 always claimed `Task execution timed out after {effective_timeout} seconds`. So a 300s stall-kill at ~531s wall-clock was persisted as `timed out after 2400 seconds` — operators misdiagnosed it and bumped the execution timeout (the wrong knob; the 300s no-output threshold is what fired).

## Root cause (confirmed)

`docker/base-image/agent_server/services/headless_executor.py`:
- Watchdog `except` (`:653`) computed a distinct `reason` but only **logged** it, then re-raised bare.
- Orchestrator `except subprocess.TimeoutExpired` (`:1023`) stamped the generic string unconditionally.
- Persist chain: agent 504 detail → backend `raise_for_status` → `httpx.HTTPError` handler (`task_execution_service.py:1019-1025`) writes `detail` to `schedule_executions.error`. So the agent's label **is** the persisted error → fix is agent-side only.

## Changes

- `HeadlessRunContext` gains `termination_reason` (`"stall_no_output"` | `"max_duration"`) + `stalled_tool`, set in the watchdog before re-raise.
- Orchestrator builds a **reason-specific** 504 with a **structured dict detail**:
  - stall → `{message: "Killed: tool '<name>' produced no output for 300s (stall watchdog)", termination_reason: "stall_no_output", stalled_tool}`
  - max-duration → `{message: "Task execution timed out after Ns", termination_reason: "max_duration"}`
- Backend already parses dict detail (`detail["message"]`) → persisted error now reflects the real cause; structured reason rides along for UI/API. **No DB migration.**
- The outer `asyncio.TimeoutError` safety net (genuine budget exhaustion, elapsed ≈ N) keeps the plain "timed out after Ns" string.

## Cross-references

- #516, #520 (closed) — sibling misclassification bugs, same family.
- #1083 (open) — fire-and-forget dispatch rework; touches the backend dispatch path, **not** this agent-side label. No conflict.
- Companion (per-issue, **out of scope**): per-task override of the too-tight 300s stall threshold; a dedicated `termination_reason` DB column.

## Verification

- `tests/unit/test_1094_stall_termination_reason.py` — asserts both ctx outcomes via the #970 fake-Popen harness.
- Full #970 wait-loop regression suite still green: **11 passed**.
- `py_compile` clean. Base image rebuilt `trinity-agent-base:0.6.0`/`latest`; fix grep-verified inside the image.

(Live repro needs a >300s silent tool call — covered by unit tests instead of a 5-min live wait.)

## AC

- [x] Stall-kill no longer reuses the max-duration string
- [x] Distinct reason surfaced (`stall_no_output` vs `max_duration`)
- [x] "timed out after Ns" reserved for genuine budget exhaustion
- [x] Base image rebuilt and verified

Related to #1094

### DIFF
--- a/docker/base-image/agent_server/services/headless_executor.py
+++ b/docker/base-image/agent_server/services/headless_executor.py
@@ -1030,9 +1030,15 @@ async def execute_headless_task(
                 # ctx.terminate() does up to 4s of process.wait() (SIGTERM grace + SIGKILL grace);
                 # off-load to the executor so the event loop stays responsive while we tear down.
                 await loop.run_in_executor(None, ctx.terminate)
+                # #1094: same semantic cause as the inner max-duration branch —
+                # keep the structured detail symmetric so consumers filtering on
+                # termination_reason see budget timeouts from either path.
                 raise HTTPException(
                     status_code=504,
-                    detail=f"Task execution timed out after {ctx.effective_timeout} seconds"
+                    detail={
+                        "message": f"Task execution timed out after {ctx.effective_timeout} seconds",
+                        "termination_reason": "max_duration",
+                    },
                 )
             except subprocess.TimeoutExpired:
                 # Inner process.wait() bounded out; tree has already been killed.
--- a/docker/base-image/agent_server/services/headless_executor.py
+++ b/docker/base-image/agent_server/services/headless_executor.py
@@ -224,6 +224,12 @@ class HeadlessRunContext:
     permission_mode_validated: bool = False
     result_seen: threading.Event = field(default_factory=threading.Event)  # #970: claude emitted {"type":"result"}
     stdout_exc: List[BaseException] = field(default_factory=list)
+    # #1094: which termination path fired, so the terminal 504 carries a
+    # distinct reason instead of always claiming the max-duration timeout.
+    #   "stall_no_output" — a tool produced no result for >_STALL_LIMIT_S
+    #   "max_duration"    — the effective_timeout budget was genuinely exhausted
+    termination_reason: Optional[str] = None
+    stalled_tool: Optional[str] = None
 
     # Shared mutable buffers (populated by stream_parser via process_stream_line)
     response_parts: List[str] = field(default_factory=list)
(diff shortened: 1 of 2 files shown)


docker/base-image/agent_server/services/headless_executor.py
```diff
@@ -1021,11 +1035,35 @@ async def execute_headless_task(
                     detail=f"Task execution timed out after {ctx.effective_timeout} seconds"
```
dolho: **Asymmetric structured reason — outer-timeout path drops `termination_reason`.**

This PR makes the inner `subprocess.TimeoutExpired` max-duration branch (just below) return a structured `detail={"message": ..., "termination_reason": "max_duration"}`, but this sibling `asyncio.TimeoutError` safety-net raise — same semantic cause (budget exhausted) — still emits a plain string with no `termination_reason`.

Failure scenario: a UI/API consumer filtering executions by `termination_reason == "max_duration"` silently misses any budget timeout that fired via the outer net instead of the inner wait-loop. The backend still persists the message (it handles both string and dict detail), so this is a structured-field gap, not a crash.

Suggest making the two max-duration exits symmetric:

```suggestion
                    detail={
                        "message": f"Task execution timed out after {ctx.effective_timeout} seconds",
                        "termination_reason": "max_duration",
                    }
```

dolho: Fixed in 0a81fb6f — outer `asyncio.TimeoutError` path now raises the same structured `{message, termination_reason: "max_duration"}` detail. Both max-duration exits are symmetric; base image rebuilt and verified (count=2 in image).

vybe (approved):
Validated via /validate-pr: distinct termination reasons verified end-to-end — backend dict-detail parsing confirmed at task_execution_service.py:1018-1021, unit tests cover both paths, CI green on all seeds. Approving.

Issue #1037, opened 2026-06-03
bug: subscription auto-switch (SUB-003) recreates agent container without draining in-flight executions, killing them (RemoteProtocolError)


obasilakis:
## Summary

When SUB-003 auto-switches an agent's subscription (on a 429 rate-limit or auth failure), it recreates the agent container to inject the new `CLAUDE_CODE_OAUTH_TOKEN` — but it stops the container **without draining in-flight executions**. Any other execution running in that agent's parallel slots is killed mid-flight and recorded as a failure (`HTTP error: RemoteProtocolError` / `ReadError`). Long-running scheduled tasks (research agents, multi-minute/30-min runs) are silently lost and the spend wasted. The same no-drain recreate path is reachable from manual subscription assignment and agent rebuild/deploy.

## Component

Backend / Subscription auto-switch (SUB-003) — `services/subscription_auto_switch.py`

## Priority

P1 — a core feature (auto-switch) destroys in-flight work from another core feature (parallel scheduled execution), with no in-product workaround. Recurs automatically whenever a shared subscription rate-limits while agents have long executions running.

## Error

Recorded on the killed execution:
```
schedule_executions.error = "HTTP error: RemoteProtocolError"   (or "HTTP error: ReadError")
```

Backend log sequence (timestamps from one occurrence; the stop precedes the recreate-complete by ~3s):
```
12:12:43  [TaskExecService] Failed to execute task on [AGENT]: HTTP error: RemoteProtocolError   <- in-flight exec killed
12:12:43  [Slots] Agent '[AGENT]' released slot for execution [EXEC_ID] ... FAILED
12:12:46  Recreated container for agent [AGENT] with updated configuration                       <- recreate completes
12:12:57  [SUB-003] Auto-switch complete: {'switched': True, 'agent_name': '[AGENT]',
          'failure_kind': 'rate_limit', 'restart_result': 'success'}
```
The connection is severed because the agent's HTTP server is stopped while the backend is still awaiting the `/api/task` response — httpx surfaces `RemoteProtocolError` (peer closed mid-response) or `ReadError` (connection dropped during read).

## Location

- **File**: `src/backend/services/subscription_auto_switch.py`
- **Function**: `_restart_agent()` (line ~266), called from `_perform_auto_switch()` (line ~203)

The offending sequence:
```python
# _restart_agent(), no check for running executions:
await container_stop(container)        # kills every in-flight execution in the agent's slots
await start_agent_internal(agent_name)
```
[...]

Issue #606, opened 2026-04-30
bug: auto-switch (SUB-003) marks triggering execution failed due to spurious .credentials.enc injection after container restart


vybe:
## Summary

When SUB-003 auto-switch fires (rate-limit 429 or auth error), it correctly recreates the agent container with the new subscription's `CLAUDE_CODE_OAUTH_TOKEN` env var. However, after the container is ready, the lifecycle code also attempts to import a `.credentials.enc` file from the agent workspace. This file does not exist for OAuth token-based subscriptions — those use env var injection, not file-based credential import. The import fails after 3 retries, causing the **triggering execution to be marked as `FAILED`** even though the container switch succeeded and subsequent executions on the new subscription work correctly.

## Component

Backend / Subscription Auto-Switch (SUB-003) / Agent Lifecycle

## Priority

P2

## Error

```
Failed to import credentials into agent {agent_name} after 3 attempts: No .credentials.enc file found in agent workspace
```

Logger: `services.agent_service.lifecycle`

Immediately followed by:
```
[SUB-003] Auto-switch complete: {'switched': True, 'agent_name': '...', 'restart_result': 'success'}
Async task completed for {agent_name}: status=TaskExecutionStatus.FAILED
```

## Root Cause

The auto-switch flow performs two credential-injection steps after a rate-limit event:

1. **Container recreation** — recreates the container with the new subscription's `CLAUDE_CODE_OAUTH_TOKEN` as an env var. This works correctly.
2. **Workspace credential import** — separately calls `import_credentials` which looks for a `.credentials.enc` file in the agent workspace. This is a legacy path for credential-file-based injection and does not apply to subscriptions registered via `claude setup-token` (OAuth token flow).

Step 2 is unnecessary for OAuth token subscriptions (the container already has the correct token from step 1), but its failure is propagated up and causes the triggering execution to be marked `FAILED`. The `restart_result: 'success'` in the auto-switch result confirms step 1 succeeded, but the execution status reflects the spurious step 2 failure.

## Impact

Every auto-switch event causes one unnecessary execution failure. Without this bug, the triggering execution could potentially be retried on the new subscription rather than being terminated. For high-frequency scheduled agents, this results in measurable false-failure rates in the execution history.

## Reproduction Steps

1. Register a subscription via `claude setup-token` (OAuth token, not API key)
2. Assign it to an agent
3. Trigger an execution that causes the subscription to hit its rate limit (429)
[...]

Issue #526, opened 2026-04-26
feat: per-agent dispatch circuit breaker to prevent backlog poisoning (RELIABILITY-007)


vybe:
## Summary

Add a per-agent circuit breaker at the **backend dispatch layer** (in front of `SlotService` / `BacklogService`) that fast-fails new executions with HTTP 503 when an agent is unhealthy, instead of accepting them into the backlog where they sit doomed. Track a rolling failure window per agent; when failure rate exceeds threshold (e.g., 5 of last 10 within 60 s), open the circuit; half-open probe after a cooldown.

Distinct from #304 (closed — agent-to-agent calls via `chat_with_agent` MCP tool) and #307 (heartbeat signal). This is the **producer-side breaker that consumes the heartbeat** to stop filling the queue with work that will time out.

## Context

From the orchestration architecture review (2026-04-26):

> No per-agent circuit breaker or failure isolation. A wedged agent (auth dead, OOM looping, push stream broken) keeps draining slots and backlog. Backlog has a depth cap but no health gate: you'll happily queue 50 doomed tasks. Producers have no signal to back off.

Concrete scenario: agent's Anthropic API key expires. Every execution fails after the auth-detect window (~seconds) but the slot is held for the duration. Backlog starts queuing more. By the time someone notices, 50 tasks are queued, each will fail in turn, and the user sees a 25-minute wait before getting their first failure.

## Acceptance Criteria

- [ ] Per-agent breaker state machine: `closed → open → half-open → closed`, persisted in Redis (so it survives backend restart) and surfaced via `GET /api/agents/{name}/health`.
- [ ] Breaker opens when rolling failure rate exceeds threshold (default: 5 failures of last 10 within 60 s; configurable per agent). Counts only specific error codes — `AGENT_ERROR`, `AUTH`, `TIMEOUT` — not `CAPACITY` (which is normal backpressure).
- [ ] When circuit is open: dispatch returns HTTP 503 immediately with `X-Circuit-Open: true` and a `retry_after_seconds`. **Do not enqueue to backlog.** This is the key invariant.
- [ ] Half-open probe after cooldown (default 30 s): allow one execution through; success → close, failure → reopen with exponential backoff.
- [ ] Frontend surfaces breaker state in the agent dashboard with a clear "circuit open" status (distinct from "stopped" or "no capacity").
[...]

Issue #421, opened 2026-04-20
bug: credential injection runs on health-check recovery path against already-running containers


vybe:
## Summary

`services.agent_service.lifecycle` attempts to inject Claude Code credentials into an agent container on the health-check / recovery path, even when the container is already running with valid env-var-based auth (`CLAUDE_CODE_OAUTH_TOKEN` per SUB-002). When the target container is busy serving other requests, the 3-attempt injection (~2s between attempts) fails with `All connection attempts failed`, adding log noise and unnecessary container load during exactly the moments when the container is already under stress.

## Component

Backend / Agent lifecycle (`services.agent_service.lifecycle`)

## Priority

P2 — not a user-visible failure by itself, but: (a) it is strictly redundant when env-var auth is in use; (b) it fires repeatedly under load exactly when the system is fragile; (c) the 3-retry-then-ERROR pattern pollutes logs and confuses investigations.

## Error

```
WARNING services.agent_service.lifecycle: Credential import attempt 1 failed: All connection attempts failed
WARNING services.agent_service.lifecycle: Credential import attempt 2 failed: All connection attempts failed
WARNING services.agent_service.lifecycle: Credential import attempt 3 failed: All connection attempts failed
ERROR   services.agent_service.lifecycle: Failed to import credentials into agent <name> after 3 attempts: All connection attempts failed
```

## Location

- Logger: `services.agent_service.lifecycle`
- Likely file: `backend/services/agent_service/lifecycle.py`
- Triggered via: `trigger_health_check` / `start_agent` recovery path (after circuit-breaker OPEN → CLOSED transitions, or when another agent calls these MCP tools)

## Root Cause

Since SUB-002, subscription tokens are injected as `CLAUDE_CODE_OAUTH_TOKEN` env var at container creation. The file-based credential injection into the running container's Claude Code credential file appears to be a legacy/dual-path operation. It is still being invoked on recovery/health-check paths (e.g. when another agent calls MCP `trigger_health_check` / `start_agent`), even though:

1. The container is already running and already authenticated via env var.
2. The target HTTP endpoint inside the agent is under load and unable to accept the injection connection.

This was observed during an overload window where delegate agents were serving long-running tasks and the orchestrator (cleon) called `trigger_health_check`; the health check succeeded, but an accompanying credential re-injection failed unnecessarily.

## Reproduction Steps

1. Deploy Trinity with SUB-002 env-var-based subscription auth.
[...]

PR #1166, opened 2026-06-11
fix(sub-003): per-agent lock for auto-switch — serialize concurrent failures (#799)

### LINKED ISSUE #799: bug: SUB-003 auto-switch has no per-agent lock — concurrent 429s race the restart
## Summary

Concurrent failures on the same agent (two chat requests in flight, or a chat request overlapping a scheduled task) both enter `handle_subscription_failure` in `services/subscription_auto_switch.py` with no mutual exclusion. The read-decide-write window from `db.get_agent_subscription_id` (line 83) through `_restart_agent` return (line 176) is unsynchronized, so both coroutines can pick the same alternative subscription, both call `assign_subscription_to_agent`, and both fire `_restart_agent`. The second `container_stop` racing the first `start_agent_internal` can produce a wedged container, duplicate switch notifications, or trip the `#421` `was_already_running` ambiguity.

## Component

Backend / Subscription Auto-Switch (SUB-003)

## Priority

P2 — real but narrow blast radius. Requires concurrent subscription failures on the same agent within seconds of each other, which only burst-loaded agents see in practice.

## Reachable from

- `src/backend/routers/chat.py:475` (429 path) and `:504` (auth-class path) — concurrent chat requests for the same agent
- `src/backend/services/task_execution_service.py:623` and `:629` — concurrent scheduled tasks for the same agent

## Concrete failure scenario

Burst on one agent:

1. Two coroutines (A, B) both hit the agent and both get a 429.
2. Both enter `handle_subscription_failure` with `current_sub_id = sub-A`.
3. Both call `db.select_best_alternative_subscription(sub-A)` and pick `sub-B`.
4. Both call `db.assign_subscription_to_agent(agent, sub-B)`.
5. Both call `_restart_agent` → `container_stop(container)` then `start_agent_internal(agent_name)`.
6. B's `container_stop` lands while A's `start_agent_internal` is mid-flight — the container ends up in one of:
   - wedged (stopped while starting),
   - duplicate restart with two switch notifications and two activity rows,
   - second `start_agent_internal` sees a freshly-recreated container as `running` and trips the `#421` `was_already_running` skip incorrectly.

## Suggested fix

Wrap `handle_subscription_failure` in a per-`agent_name` `asyncio.Lock` from a module-level `dict[str, asyncio.Lock]`. The lock must span from `db.get_agent_subscription_id` (line 83) through `_restart_agent` return (line 176) — the full read-decide-write window.

```python
# services/subscription_auto_switch.py
import asyncio
from collections import defaultdict

_agent_switch_locks: dict[str, asyncio.Lock] = defaultdict(asyncio.Lock)

async def handle_subscription_failure(agent_name: str, ...) -> Optional[dict]:
    async with _agent_switch_locks[agent_name]:
        # existing body unchanged
        ...
```

Process-local is sufficient today (single-process backend). If/when the backend goes multi-process, escalate to a Redis `SETNX` lock keyed `auto_switch:{agent_name}` with TTL ≥ the longest plausible container start (currently ~60s).

## Test coverage

Add a concurrency test that fires two `handle_subscription_failure` coroutines for the same agent under `asyncio.gather` and asserts:

1. Exactly ONE `assign_subscription_to_agent` call observed.
2. Exactly ONE `_restart_agent` call observed.
3. Exactly ONE switch notification created.

The second caller should see the post-switch state (`current_sub_id == sub-B`) and either no-op or pick a different alternative — not double-switch.

## Existing reference

The carve-out is documented at `tests/unit/test_subscription_auto_switch_no_cred_import.py:36-39` ("Explicit non-claim" docstring) — that test pins the credential-import short-circuit (`#606`) but explicitly does NOT cover this race.

## Related

- Feature: Subscription Auto-Switch (SUB-003) — original `#153` (closed)
- Bug: `#606` (this issue's parent context — credential-import short-circuit; in-progress on branch `AndriiPasternak31/issue-606-plan`)
- Related: `#792` (retry triggering execution after successful auto-switch) — different concern, same surface
- Backend service: `services/subscription_auto_switch.py`
- Backend service: `services/agent_service/lifecycle.py`

AndriiPasternak31:
## What

Adds a per-agent lock so concurrent SUB-003 subscription failures on the same agent can't race the auto-switch, plus a stale-failure guard so 3+ viable subscriptions don't cascade. Closes #799.

## Problem

`services/subscription_auto_switch.py::handle_subscription_failure` ran an unsynchronized read→decide→assign→restart sequence. Two failures on the *same* agent (two chat requests, or a chat overlapping a scheduled task — burst-loaded agents) both:
- read the same current subscription,
- pick the same alternative,
- `assign_subscription_to_agent`, and
- fire `_restart_agent`.

The second `container_stop` racing the first `start_agent_internal` wedges the container, fires duplicate switch notifications/activity rows, or trips the #421 `was_already_running` ambiguity.

## Fix

- **Per-agent `asyncio.Lock`** (lazy + event-loop-safe — mirrors `services/agent_call_limiter.py`, *not* `defaultdict(asyncio.Lock)`, which binds locks to the wrong loop across pytest's per-test loops) wrapping the whole switch critical section. At most one switch + restart per burst.
- **Stale-failure guard (Codex C8):** snapshot the subscription at handler entry; under the lock, bail if it changed. Without this, a loser whose failure was about sub-A would attribute it to the new current sub-B and cascade `A→B→C` for 3+ viable subscriptions. The 2h skip-list masks this for 2 subs only.

Process-local is sufficient today: the backend runs a single uvicorn process, and the scheduler delegates execution to the backend via `/api/internal/execute-task` rather than calling this module in its own process. Multi-process escalation path (Redis `SETNX`) is documented in-code.

## Tests

`tests/unit/test_subscription_auto_switch_concurrency.py`:
- 2-sub concurrent burst → exactly one switch, loser no-ops.
- **3-sub no-cascade regression** (the critical one) → asserts no `A→B→C`; verified to fail without the guard (cascade `['B','C']`).
- distinct-agent locks don't serialize.

Regression: existing pingpong + the #606 carve-out test (which enters at `_restart_agent`) stay green; `tests/lint_sys_modules.py` clean (no new violations).

## Scope note

This was originally planned to land *with* #1037 (drain-aware recreate). An engineering review (with a Codex outside-voice pass) found the bundled drain-gate design had multiple correctness holes — `/chat` overflow doesn't actually suspend execution, the 60s maintenance sweep bypasses an `acquire()`-level gate, the recreate-pending TTL can't span a 7200s execution, recreate-failure drains onto the bad token, and #526's breaker can wipe the queued tasks. So **#1037 is split into its own PR** built around a slot-acquire-primitive gate; the findings are captured on #1037 as its design seed. This PR is the small, correct, independent half.

### DIFF
--- a/src/backend/services/subscription_auto_switch.py
+++ b/src/backend/services/subscription_auto_switch.py
@@ -20,6 +20,7 @@ alternative selection (`select_best_alternative_subscription` +
 tests pinning that contract.
 """
 
+import asyncio
 import logging
 from typing import Optional
 
@@ -27,6 +28,55 @@ from database import db
 from db_models import NotificationCreate
 
 logger = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# Per-agent switch lock (#799)
+# ---------------------------------------------------------------------------
+#
+# Concurrent subscription failures on the SAME agent (two chat requests, or a
+# chat overlapping a scheduled task) both enter `handle_subscription_failure`
+# and, without mutual exclusion, both pick the same alternative, both
+# `assign_subscription_to_agent`, and both fire `_restart_agent` — the second
+# `container_stop` racing the first `start_agent_internal` wedges the container,
+# duplicates the switch notification, or trips the #421 `was_already_running`
+# ambiguity. A per-agent lock serializes the read→decide→assign→restart window.
+#
+# Mirrors the event-loop-safe lazy pattern in `services/agent_call_limiter.py`
+# (module dict + a lazily-created guard) rather than `defaultdict(asyncio.Lock)`:
+# a defaultdict binds each lock to whatever event loop is current at first key
+# access and persists it, which breaks across pytest's per-test loops
+# ("Future attached to a different loop"). Creating the locks lazily on the
+# running loop avoids that.
+#
+# INVARIANT: process-local. Correct only while (a) the backend runs a single
+# process and (b) the scheduler delegates execution to the backend via
+# `/api/internal/execute-task` rather than calling this module in its own
+# process — both true today. If the backend ever runs multiple workers, escalate
+# to a Redis `SETNX` lock keyed `auto_switch:{agent_name}` (TTL ≥ longest
+# plausible container start, ~60s).
+_AGENT_SWITCH_LOCKS: dict[str, asyncio.Lock] = {}
+_AGENT_SWITCH_LOCKS_GUARD: Optional[asyncio.Lock] = None
+
+
+async def agent_switch_lock(agent_name: str) -> asyncio.Lock:
+    """Return the per-agent switch lock, creating it lazily on the running loop."""
+    global _AGENT_SWITCH_LOCKS_GUARD
+    if _AGENT_SWITCH_LOCKS_GUARD is None:
+        _AGENT_SWITCH_LOCKS_GUARD = asyncio.Lock()
+    lock = _AGENT_SWITCH_LOCKS.get(agent_name)
+    if lock is None:
+        async with _AGENT_SWITCH_LOCKS_GUARD:
+            lock = _AGENT_SWITCH_LOCKS.setdefault(agent_name, asyncio.Lock())
+    return lock
+
+
+def _reset_locks_for_test() -> None:
+    """Test hook: drop all per-agent locks + the guard so each test's event loop
+    starts clean (locks are loop-bound)."""
+    global _AGENT_SWITCH_LOCKS_GUARD
+    _AGENT_SWITCH_LOCKS.clear()
+    _AGENT_SWITCH_LOCKS_GUARD = None
 
 
 # Substrings that indicate an auth-class subscription failure. Mirrors the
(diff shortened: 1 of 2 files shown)


AndriiPasternak31:
### Why this is #799-only (decision trail)

This started as a combined #799 + #1037 PR per the do-together note. The plan went through `/plan-eng-review` + a Codex outside-voice pass, which reshaped the scope:

1. **D1** — for #1037 I first chose an admission gate in `CapacityManager.acquire()` over a passive recreate-on-idle (a prior cross-model learning flagged the passive approach as best-effort: new admissions run on the stale token, busy agents starve the recreate).
2. **Codex outside voice** — proved the `acquire()`-level gate still leaks: `/chat` `queue_in_memory` is observability-only and doesn't suspend execution; the 60s `drain_orphans_all()` → `BacklogService.drain_next()` acquires `SlotService` directly, bypassing the gate; the #526 half-open probe bypasses it; the recreate-pending TTL can't span a 7200s execution; recreate-failure drains onto the bad token; #526's breaker can wipe deliberately-queued rows. It also caught that the #799 lock alone double-switches `A→B→C` for 3+ subs (→ the stale-failure guard in this PR).
3. **D3** — #1037-done-right belongs in the slot-acquire primitive (the one chokepoint all admission paths funnel through) and is its own focused PR. Ship #799 standalone now.

The #1037 redesign seed (the full Codex finding list) is posted on #1037.

vybe (approved):
✅ **Validated & approved** (/validate-pr)

P2 reliability bug fix (Closes #799). 2 files, +337/-35.

**Correctness:**
- Per-agent `asyncio.Lock` via lazy, event-loop-safe creation (`agent_switch_lock`) — correctly avoids the `defaultdict(asyncio.Lock)` cross-loop binding trap; guard check→set has no await between, so it's atomic on the single event loop, and `setdefault` under the guard resolves the create race.
- Stale-failure guard (snapshot at entry → re-read under lock → bail on change) correctly short-circuits BEFORE alternative selection, preventing the A→B→C cascade for 3+ viable subs.
- Process-local invariant explicitly documented with the Redis `SETNX` escalation path for multi-worker.

**Tests:** `test_subscription_auto_switch_concurrency.py` — deterministic (pre-hold-lock), covers exactly-one-switch, the critical 3-sub no-cascade regression (asserts `B` never reaches alternative selection), and distinct-agent locks. Snapshot/restore sys.modules fixture keeps blast radius bounded.

- Base → `dev` ✓ · conventional commit ✓ · CI all green
- Security scan clean; no new backend modules / env vars / infra changes
- Bug fix → descriptive commit sufficient (no docs required)

Note: head branch `AndriiPasternak31/issue-799` is the base of #1196 (stacked) — merge this first; #1196 auto-retargets to `dev`.

Issue #1095, opened 2026-06-08
feat: transactional agent executions — discard workspace changes unless validated as success (research-gated)


vybe:
## Summary

When an agent execution fails — **hard** (crash / non-zero / timeout) or **soft** (exits 0 but leaves half-finished or invalid output) — the partial filesystem changes remain in the agent workspace, corrupting state for the next run. We want execution to be **transactional at the workspace level**: snapshot the workspace before the turn, run, validate, and **commit changes only on validated success — otherwise roll the workspace back** to its pre-turn state. Conceptually like a git worktree or a DB transaction, scoped to the agent's own filesystem.

**This issue requires a research / design phase to select the optimal mechanism before implementation** — several viable approaches exist with materially different tradeoffs, and the choice interacts with concurrency, git-vs-non-git workspaces, and the pull/lease-redelivery direction.

## Context

Recurring operator pain: a task fails or soft-fails and leaves leftovers in the files; the next execution starts from a dirty/partial tree. Trinity already has most of the rollback machinery (snapshot/restore + persistent-state allowlist) built for git recovery, and `docs/planning/TARGET_ARCHITECTURE.md` already plans a post-check hook and lease-expiry re-delivery — both of which this capability composes with.

Two sub-problems that must stay **decoupled**:

1. **Transaction boundary** — snapshot → run → commit-or-discard atomically (workspace only).
2. **Validation gate** — what decides "success?" Hard-fail is easy; the painful case is **soft-fail** (exit 0, bad output), which needs an *explicit* validator.

**Scope honesty:** this is **workspace-filesystem-scoped only**. It cannot roll back external side effects (email / Slack / git push / payment) — that is the separate effect-idempotency workstream (#1084). Any design must state this limit explicitly so it does not masquerade as global atomicity.

## Research Questions (must be answered to pick the optimal solution)

- **Mechanism:** snapshot-and-rollback in place vs. real git worktree-per-execution (#27) vs. OverlayFS upperdir swap vs. CoW volume clone (btrfs/zfs)? Evaluate against the governing principles (simplicity, proven primitives, commodity hardware) and the concurrency constraint below.
- **Concurrency:** snapshot-in-place breaks when `max_parallel_tasks > 1` (shared workspace). Does the optimal solution require worktrees for concurrent agents and snapshot-rollback for serialized ones, or one mechanism for both?
[...]

Issue #1084, opened 2026-06-05
feat: effect-scoped idempotency keys for outbound side effects


vybe:
## Summary — RE-SCOPED to v2 (Direction B, 2026-07-01, doc #1404)

**Re-scoped** by the v2 side-effect reframe (`docs/planning/TARGET_ARCHITECTURE.md`, #1404). The original ambition — thread an effect-scoped idempotency key `{execution_id}:{effect_ordinal}` through **every** outbound sink ("Direction A", universal coverage) — is **retired**. The sink surface (the agent's own Resend key, `gh`, `curl`, any new MCP server) is unbounded and un-confineable, so universal per-sink coverage is structurally unachievable. `effect_guard` is **kept, narrowed** — it is no longer chasing every sink, and it no longer gates pull.

<details><summary>Original Direction-A framing (superseded — kept for history)</summary>

Thread an effect-scoped idempotency key `{execution_id}:{effect_ordinal}` from the task envelope through every outbound side effect, so a re-delivered turn's repeated action is deduplicated at the sink. Framed as "the hardest open problem" and "the gate on defaulting pull mode ON for side-effect-bearing agents."
</details>

## What #1084 now is

`effect_guard` (merged to dev — 4 sinks: `proactive_message_service.send_message`, `voip_service.place_outbound_call`, `agent_shared_files_service.create_share`, `nevermined_payment_service.settle_payment_once`) is retained on:

1. **Reversible / backend-fronted sinks** — where a cheap within-turn `in_flight`/`completed` dedup is useful and sufficient (a duplicate here is survivable).
2. **The seed of the confined-tool gate** — the `settle_payment_once` guard is the starting point for the deterministic tool-side gate, which is now its own workstream: **#1408**.

Universal per-sink coverage is **no longer pursued**.

## No longer the gate

Under v2, pull default-on gates **per-effect, not per-agent** (see #1081): read/analysis-only + reversible + confined-irreversible agents default on immediately; only irreversible-**un-confineable** effects wait. **#1084 no longer blocks pull default-on.** Side-effect *recovery* moved to retry-with-prior-trace (#1401) + the async operator-queue human-gate (#1402); the deterministic gate on confined irreversible rails is **#1408**.

## Acceptance criteria (re-scoped)

- [x] `effect_guard` merged — within-turn `in_flight`/`completed` dedup across the 4 sinks
- [ ] Classify the 4 sinks: which are reversible/backend (stay here) vs confined-irreversible (hand to #1408)
[...]

Issue #1033, opened 2026-06-02
bug: backend image missing redis_breaker_util.py — Dockerfile COPY not added for new top-level module (#986)


vybe:
## Summary

Commit `0c671158` (*feat: per-agent dispatch circuit breaker (#526) (#986)*) adds a new top-level backend module `src/backend/redis_breaker_util.py`, imported by `services/agent_client.py`, but does **not** add a corresponding `COPY` line to `docker/backend/Dockerfile`. The backend Dockerfile copies top-level modules **by explicit name**, so the new file never lands in the image. The backend then crash-loops on startup with `ModuleNotFoundError`, taking down the whole platform (frontend/mcp-server/scheduler are gated on backend health via `depends_on`). Rebuilding with `--no-cache` does **not** help — the file is simply never copied.

## Component

Backend / build packaging (`docker/backend/Dockerfile`)

## Priority

P0 — backend fails to start; full platform outage on any deploy of this commit.

## Error

```
ModuleNotFoundError: No module named 'redis_breaker_util'
  File "/app/main.py", line 44, in <module>
    from routers.auth import router as auth_router
  File "/app/routers/agents.py", line 41, in <module>
    from services.agent_service import (...)
  File "/app/services/agent_service/lifecycle.py", line 27, in <module>
    from services.skill_service import skill_service
  File "/app/services/skill_service.py", line 26, in <module>
    from services.agent_client import get_agent_client, AgentClientError
  File "/app/services/agent_client.py", line 31, in <module>
    from redis_breaker_util import (...)
```

## Location

- **File**: `docker/backend/Dockerfile` (top-level module COPY block, ~lines 62–68)
- **Missing module**: `src/backend/redis_breaker_util.py`
- **Importer**: `src/backend/services/agent_client.py:31`

## Root Cause

`docker/backend/Dockerfile` enumerates top-level backend modules individually:

```dockerfile
COPY ../../src/backend/main.py /app/
COPY ../../src/backend/config.py /app/
COPY ../../src/backend/models.py /app/
COPY ../../src/backend/db_models.py /app/
COPY ../../src/backend/dependencies.py /app/
COPY ../../src/backend/database.py /app/
COPY ../../src/backend/logging_config.py /app/
```

[...]

Issue #671, opened 2026-05-05
Cancelled executions reported as status=success in API and DB


pandore:
## Summary

A scheduled execution that is cancelled via the Trinity UI is recorded with `status: "success"` in the gateway DB and surfaced as `success` through the MCP API (`list_recent_executions`, `get_execution_result`). The cancel intent is never persisted onto the execution record.

This makes "did the work actually happen?" undetectable via the API, and silently advances the schedule's `next_run_at` as if the run had succeeded.

## Repro on bdr-agent (port 2260)

Two consecutive incidents on schedule `fYtR6WScyBPQDSmaDfgSJQ` (`Daily Lead Outreach`, `0 7 * * 1-5 America/New_York`):

| Run | Started (UTC) | Duration | Reported status | Actual outcome |
|---|---|---|---|---|
| `SA_Ln0AClqeK8qjW1fdLUg` | 2026-05-05 18:16:49 | 31:20 | `pending_retry` | Hit `max_turns_task=50` mid-Step-4 enrichment. Container log: `Claude Code max_turns reached: Reached maximum number of turns (50)`. ✅ Reported correctly. |
| `2cId6kfgZsTl0x3cIV5OHQ` | 2026-05-05 18:49:11 | 10:30 | **`success`** | **Cancelled by user via UI mid-run.** No Slack ping posted, no sheet rows written, $1.30 spent. ❌ Reported as success. |

## Container-side evidence

For the cancelled retry, agent container logs (`get_agent_logs bdr-agent`) show only:

```
INFO:agent_server.routers.chat:[Task] Task 2cId6kfgZsTl0x3cIV5OHQ completed successfully
INFO:     172.28.0.5:37822 - "POST /api/task HTTP/1.1" 200 OK
```

No `cancelled`, no `SIGTERM`, no `interrupted` markers. Claude Code received the cancel signal, gracefully emitted a final response, and exited 0. The task-runner saw exit code 0 and reported `completed successfully`. Trinity's gateway recorded `status: success`.

## Hypothesised flow

1. UI cancel → Trinity gateway sends stop signal to Claude Code in the agent container.
2. Claude Code catches it, writes a brief final response, exits with code 0.
3. Container task-runner reports `[Task] ... completed successfully` (because exit code is 0).
4. Trinity DB stores `status: success` for the execution.
5. Cancel intent — known only to the gateway/UI at the moment of the click — is never written onto the execution record.

## Why this matters

[...]

PR #1195, opened 2026-06-12
fix(executor): consume the terminal-write CAS verdict (#671/H4) + #1095 research doc

AndriiPasternak31:
## Summary

Issue #1095 ("transactional agent executions — discard workspace changes unless validated as success") is **research-gated**. This PR delivers two things:

1. **The research/design deliverable** — `docs/planning/TRANSACTIONAL_EXECUTIONS_2026-06.md`. Its §0 evidence gate returns a **NO-GO** for the broad transactional subsystem on today's data; the full mechanism (container-local coordinator, `~/.trinity/post-check`, `tx_state` reconciliation, crash-injection suite) is an explicitly deferred, GO-gated follow-up.
2. **A proven artifact the research surfaced** — the **H4 CAS fix** (doc §6.3 / §13 step 1), which the doc says to *"ship regardless of GO… its own small PR."*

So this is "research **+** the reliability fix it surfaced," not research alone.

## The bug (H4)

`db.update_execution_status(...)` (`src/backend/db/schedules.py`) is an atomic **compare-and-set** that returns the CAS winner (`bool`). A SUCCESS write loses the CAS **only** to a `CANCELLED` row; non-success terminal writes lose to any already-terminal row. Every terminal-write caller in `TaskExecutionService.execute_task` **discarded that verdict** and ran "won-only side effects" unconditionally.

Worst case (the SUCCESS path): a turn the operator already **cancelled** was reported back as a **billable SUCCESS** — the activity was completed as COMPLETED, the dispatch breaker was reset, and the caller received SUCCESS. The DB-layer CAS itself is already correct and tested (`tests/unit/test_cancelled_not_overwritten.py`, #671); this was purely the **service layer ignoring the verdict**.

## What changed

Every terminal writer now **consumes the CAS verdict — lost CAS ⇒ reconcile/skip, never act:**

- **SUCCESS path** (inline): on a lost CAS it re-reads the row, completes the activity as FAILED (not COMPLETED), skips the breaker reset, and returns the persisted terminal status with a new `RECONCILED` error code instead of SUCCESS. The won path is unchanged.
- **5 FAILED paths** (CB fast-fail, timeout, backend call-budget, HTTPError salvage, generic Exception) converge onto one helper `_write_terminal_and_gate(...)`, which writes the terminal through the CAS and completes the activity **only when this writer won**. This replaces the old check-then-act guard (`get_execution() → status != CANCELLED`), which both **raced (TOCTOU)** and **only blocked CANCELLED** — the CAS additionally blocks an already-SUCCESS/FAILED row.
- The HTTPError path's **AUTH dispatch-breaker outcome** is now gated on the same `won` verdict, so a turn superseded by a cancel can't count toward the breaker.
- **Untouched:** the `finally` slot-release (must always run, never gated), the no-side-effect FAILED sites (CapacityFull, CircuitOpen), and the already-correct `CancelledError` 3-state guard.

## Testing

This was implemented **test-first** and verified at three levels.

**1. TDD red → green (unit, backend-free)**
- New `tests/unit/test_terminal_write_cas_gate.py` (note: in `tests/unit/`, not the root `test_cb_probe_execution_close.py` — the root file's autouse `cleanup_after_test` fixture needs a live backend, so it can't yield a local RED; `tests/unit/` overrides those fixtures and runs backend-free, next to the DB-layer half `test_cancelled_not_overwritten.py`).
- Wrote the lost-CAS test first → **RED** (today's code reports SUCCESS + fires side effects); the won-path test passed (harness reaches the real terminal). Applied the fix → **GREEN**. The file covers both the SUCCESS reconcile and the FAILED-path gating, plus won-path regression pins so the fix can't over-gate the happy path. **4 passed.**

**2. Full unit suite (no regression)**
- `pytest tests/unit/` → **2135 passed, 24 skipped** (deterministic order). ruff clean on both edited files; mypy adds **zero** new errors (the pre-existing errors are repo-wide loose typing on `update_execution_status`, unchanged by this PR).

**3. `verify-local` — real prod images + isolated sibling stack (PASS, all 5 stages)**

| Stage | Result | What it proves |
|-------|--------|----------------|
| preflight | OK (1s) | environment |
| unit | OK (168s) | full unit suite **inside the built prod image** |
| build + import-smoke | OK (8s) | the new code reaches the image and imports (#1033 class) |
| boot + health | OK (12s) | backend boots with the change; `:8100/health` green |
| integration | OK | runtime against the live sibling stack |

`status: pass`; sibling stack auto-torn-down.

## Scope / not in this PR

The broad workspace-transaction subsystem stays **NO-GO** per the doc's §0 gate (re-runnable production-evidence query). Deferred, GO-gated: container-local coordinator, `~/.trinity/post-check` validator, `tx_state` + restart reconciliation, the §11 crash-injection matrix, the worktree/concurrent track (#27), and external-effect rollback (#1084).

Refs #1095

### DIFF
--- a/docs/planning/TRANSACTIONAL_EXECUTIONS_2026-06.md
+++ b/docs/planning/TRANSACTIONAL_EXECUTIONS_2026-06.md
@@ -1,0 +1,488 @@
+# Transactional Agent Executions — Design Recommendation (#1095)
+
+| Field | Value |
+|---|---|
+| **Status** | Research / design — **GO/NO-GO gate**. Implementation does not start until this is reviewed and a GO is recorded. |
+| **Date** | 2026-06-12 |
+| **Issue** | [#1095](https://github.com/Abilityai/trinity/issues/1095) — *feat: transactional agent executions — discard workspace changes unless validated as success (research-gated)* · P1 · `theme-reliability` · `complexity-high` |
+| **Branch** | `AndriiPasternak31/plan-issue-1095` (fresh off `dev`; no implementation in this phase) |
+| **Scope** | **Workspace-filesystem-scoped only.** External side effects (email / Slack / git push / payment) are explicitly **out of scope** — that is effect-idempotency, [#1084](https://github.com/Abilityai/trinity/issues/1084). |
+| **Supersedes** | [#27](https://github.com/Abilityai/trinity/issues/27) *Git Worktrees for Task Isolation* (CLOSED) — folded in as the **future concurrent track**, §10.2. |
+| **Verdict** | **NO-GO for a broad mechanism on today's evidence; a narrow serialized-git v1 is GO *only* after the production evidence query in §0 clears the kill criterion.** See §0. |
+
+> **How to read this doc.** §0 is the real first deliverable: an evidence gate that can — and on the sample available today does — return **NO-GO**. Everything after §0 is the design that becomes reachable *if* §0 clears on a representative production instance. The doc deliberately does **not** assume the build.
+
+---
+
+## 0. Evidence & GO/NO-GO gate (the real first deliverable)
+
+The issue describes "recurring operator pain": a turn fails and leaves leftovers; the next turn starts dirty. A reliability subsystem (snapshot → run → validate → commit/rollback) is justified **only if that pain is real and frequent**. Before recommending a build, we state a **kill criterion** and measure against it.
+
+### 0.1 Kill criterion
+
+> **Build only if** failed/soft-failed turns that leave **harmful** workspace leftovers (changes the next turn must not inherit) occur often enough, and on large enough workspaces, that the operator cost exceeds the cost of a new transactional subsystem and its hazards (§9).
+>
+> **NO-GO if** harmful-leftover failures are rare, OR the failures that *could* leave leftovers are dominated by the **indeterminate** class (the writer's fate is unknowable, so you cannot safely roll back anyway — §6.1), OR workspaces are small enough that a cheap targeted cleanup beats a transaction.
+
+This is falsifiable in both directions. The doc is required to allow the NO-GO conclusion.
+
+### 0.2 Evidence gathered (real data, not assumption)
+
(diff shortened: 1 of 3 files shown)


vybe (approved):
✅ **Validated & approved** (/validate-pr)

Research deliverable + the H4 CAS reliability fix it surfaced (Refs #1095 P1; the #671/H4 service-layer bug). 3 files, +851/-77 (488 of which are the planning doc).

**The bug is real:** `db.update_execution_status` is an atomic CAS returning the winner; every terminal writer in `execute_task` discarded it, so an operator-cancelled turn could be written back as a billable SUCCESS (activity COMPLETED, breaker reset, SUCCESS to caller).

**Fix review:**
- `_write_terminal_and_gate` consolidates the 5 FAILED paths; completes the activity only on a won CAS; `won=True` when `execution_id` falsy preserves prior no-record behavior.
- SUCCESS path consumes the verdict → re-reads row, completes activity as FAILED (not COMPLETED), skips breaker reset, returns new `RECONCILED` code + the persisted terminal status.
- AUTH dispatch-breaker outcome gated on `won` (a superseded turn can't count toward the breaker).
- Correctly replaces check-then-act (TOCTOU + CANCELLED-only) with the CAS (also blocks already-SUCCESS/FAILED). Untouched sites (slot-release `finally`, CapacityFull/CircuitOpen no-effect, CancelledError 3-state guard) correctly left alone.

**Tests:** `test_terminal_write_cas_gate.py` pins won/lost on both SUCCESS and FAILED terminals + happy-path over-gating regression; TDD red→green; full unit 2135 passed; verify-local 5/5 (incl. import-smoke #1033 class).

- Base → `dev` ✓ · 3 conventional commits ✓ · CI all green
- Security scan clean; no new backend modules / env vars / infra
- The broad transactional subsystem correctly stays NO-GO behind the §0 evidence gate.

Minor (non-blocking): the new `RECONCILED` reconciliation could earn a one-line mention in architecture.md's task_execution_service notes, but as an internal correctness fix a descriptive commit suffices.

Merge after #1195's sibling #1203 is rebased — they share `task_execution_service.py`, so land this first (smaller diff).

Issue #1200, opened 2026-06-13
bug: GET/PUT /api/agents/{name}/capabilities 500 — get_full_capabilities/set_full_capabilities not delegated onto DatabaseManager facade (#1093, v0.6.1)


vybe:
## Summary

`GET` and `PUT /api/agents/{name}/capabilities` both return HTTP 500 on the v0.6.1 (`#1093` SQLAlchemy-Core / db-decomposition) release. The handlers call `db.get_full_capabilities(...)` / `db.set_full_capabilities(...)` on the `DatabaseManager` facade, but those methods were never delegated onto the facade — they live only on the `SecurityMixin` that `AgentOperations` (`db._agent_ops`) composes. Result: `AttributeError`, so an agent's full-capabilities setting can be neither read nor changed via the API/UI. Backend-independent (pure Python facade gap; reproduces on SQLite and PostgreSQL).

## Component

Backend / Agent Config (DatabaseManager facade)

## Priority

P2

## Error

```
AttributeError: 'DatabaseManager' object has no attribute 'get_full_capabilities'
  File "routers/agent_config.py", line 288, in get_agent_capabilities
    db_full_caps = db.get_full_capabilities(agent_name)
```
(PUT path fails identically on `db.set_full_capabilities(agent_name, full_caps)`.)

## Location

- **File**: `src/backend/routers/agent_config.py`
- **Lines**: ~288 (`GET`, `db.get_full_capabilities`) and ~331 (`PUT`, `db.set_full_capabilities`)
- **Missing on**: `src/backend/database.py` (`DatabaseManager` facade)
- **Methods exist on**: `src/backend/db/agent_settings/security.py` (`SecurityMixin.get_full_capabilities` / `set_full_capabilities`), mixed into `AgentOperations`

## Root Cause

During the `#1093` db decomposition, `DatabaseManager` exposes most sub-operations via explicit passthrough methods (e.g. `get_read_only_mode` → `self._agent_ops.get_read_only_mode`). The capabilities pair was missed: `get_full_capabilities` / `set_full_capabilities` exist on `SecurityMixin` (and thus on `self._agent_ops` at runtime) but have no delegating method on `DatabaseManager`. The routers call them on the facade, so every call raises `AttributeError`.

Verified at runtime: `hasattr(db._agent_ops, "get_full_capabilities")` → `True`, `hasattr(db, "get_full_capabilities")` → `False`.

## Reproduction Steps

[...]

Issue #1199, opened 2026-06-13
bug: GET /api/ops/auth-report 500s — SQLAlchemy auto-correlation in get_agent_subscription (#1093, v0.6.1)


vybe:
## Summary

`GET /api/ops/auth-report` returns HTTP 500 on the v0.6.1 (`#1093` SQLAlchemy-Core db rewrite) release. The failure is a compile-time `InvalidRequestError` in `db/subscriptions.py::get_agent_subscription`, where a correlated agent-count scalar subquery loses its FROM clause to SQLAlchemy auto-correlation. Because it fails at statement-compile time it 500s for **every** call regardless of data or backend, and it reproduces identically on **both SQLite and PostgreSQL**. Token decrypt/inject (subscription *auth*) is unaffected — only paths that call `get_agent_subscription` (the fleet auth-report, per-agent subscription-status lookups) break.

## Component

Backend / Subscriptions (db layer)

## Priority

P1

## Error

```
sqlalchemy.exc.InvalidRequestError: Select statement '<...Select object...>' returned no FROM clauses due to auto-correlation; specify correlate(<tables>) to control correlation manually.
```

Abbreviated stack:
```
routers/ops.py  ->  get_auth_report
database.py     ->  get_agent_subscription
db/subscriptions.py -> get_agent_subscription  (conn.execute(stmt))
sqlalchemy ... compiler.visit_select -> _setup_select_stack -> selectable._get_display_froms
  -> InvalidRequestError: returned no FROM clauses due to auto-correlation
```

## Location

- **File**: `src/backend/db/subscriptions.py`
- **Function**: `SubscriptionOperations.get_agent_subscription` (+ helper `_agent_count_subquery`)

## Root Cause

`_agent_count_subquery()` builds a scalar subquery whose only FROM is `agent_ownership`, correlated to the outer query via `agent_ownership.c.subscription_id == subscription_credentials.c.id`:

```python
@staticmethod
def _agent_count_subquery():
    return (
        select(func.count())
        .select_from(agent_ownership)
        .where(and_(
            agent_ownership.c.subscription_id == subscription_credentials.c.id,
            agent_ownership.c.deleted_at.is_(None),
        ))
        .scalar_subquery()
        .label("agent_count")
    )
```

This is fine in callers whose outer FROM is `subscription_credentials JOIN users` (e.g. `list_subscriptions`, `get_subscription`): only `subscription_credentials` is auto-correlated, `agent_ownership` stays in the subquery FROM.

[...]

PR #1093, opened 2026-06-08
feat(db): configurable database backend — experimental PostgreSQL support (#300)

dolho:
## Summary

Implements the configurable database backend from Abilityai/trinity#300: the whole data layer now
runs on **SQLAlchemy Core** instead of raw `sqlite3`, selected at startup by
`DATABASE_URL`. **SQLite stays the zero-config default and is behavior-identical**
— nothing changes when `DATABASE_URL` is unset. **PostgreSQL is an opt-in,
experimental backend** (default OFF).

> Safety: because SQLite remains the default and PG is opt-in/off-by-default,
> this cannot regress existing deployments. PG is labeled experimental.

## What's in here

- **Seam** — `db/engine.py` (engine factory from `DATABASE_URL`, cached per URL;
  NullPool for sqlite to match the legacy open/close-per-call semantics, pooled +
  pre_ping for PG), `db/tables.py` (Core table registry for all 59 tables,
  auto-derived from `schema.py`), DATABASE_URL-aware `db/connection.py`.
- **Schema on PG** — `schema.py:init_schema_postgres()` translates the *same*
  `TABLES`/`INDEXES` strings to PG DDL (SERIAL, UTC text defaults, FK-strip since
  the platform runs FKs-off, PL/pgSQL versions of the `audit_log` append-only
  triggers). `schema.py` stays the single source of truth — no second schema to
  drift (cf. Abilityai/trinity#655/#691/#721). `init_database()` branches on dialect; a fresh PG DB
  is built at head so the sqlite-only PRAGMA migrations are skipped.
- **All 44 db modules → Core** — `?`→Core params, `INSERT OR REPLACE/IGNORE`→
  dialect `on_conflict_*` (via `make_insert`), `lastrowid`→`inserted_primary_key`,
  `datetime('now')`→`utc_now_iso()`, `sqlite3.IntegrityError`→
  `sqlalchemy.exc.IntegrityError`. Method signatures/return shapes unchanged —
  routers, services, and the `DatabaseManager` facade are untouched.
- **Backend services PG-safe** — system-agent scope update → Core; WAL checkpoint,
  VACUUM, and the `/health` migration gate gated to `is_sqlite()`; `git_service`
  catches the SQLAlchemy `IntegrityError`.
- **Standalone scheduler PG-aware** — `src/scheduler/database.py` honors
  `DATABASE_URL`, so cron-fired schedules read/write the same PG database as the
  backend; unset → shared SQLite as before.
[...]

PR #1213, opened 2026-06-15
test(#1199): fix sys.modules lint baseline failure (dev red)

dolho:
## Problem

`backend-unit-test` is **red on `dev`** (and therefore on every open PR that merges against `dev`, e.g. #1186). The `lint (sys.modules pollution check)` step fails:

```
New or grown violations vs baseline:
  tests/unit/test_1199_agent_subscription_correlation.py: 0 → 2 (+2)
```

`test_1199_agent_subscription_correlation.py` pops cached `database` / `db.*` (and shadow `utils*`) modules from `sys.modules` — at import time and inside the `tmp_db` fixture — without restoring them. Those 2 bare `sys.modules.pop(...)` calls exceed the #762 lint baseline.

## Fix

Adopt the sanctioned snapshot/restore pattern the linter explicitly allows: a top-level `_STUBBED_MODULE_NAMES` list + an autouse `_restore_sys_modules` fixture (precedent: `tests/unit/test_telegram_webhook_backfill.py`). This:

- exempts the file from the bare-`sys.modules` ban (structural check in `tests/lint_sys_modules.py`), and
- genuinely restores the popped modules after each test, stopping the cross-session pollution the lint exists to prevent.

No change to the test's assertions.

## Verification

- `python tests/lint_sys_modules.py` → exit 0, `no new violations` (was exit 1).
- `py_compile` clean. Full pytest runs in CI (local venv lacks sqlalchemy).

Related to #1199, #762. Unblocks #1186.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1214, opened 2026-06-15
fix(db): delegate get/set_full_capabilities onto DatabaseManager facade (#1200)

dolho:
## Problem

`GET` and `PUT /api/agents/{name}/capabilities` both return **HTTP 500** since the #1093 SQLAlchemy-Core db decomposition (v0.6.1):

```
AttributeError: 'DatabaseManager' object has no attribute 'get_full_capabilities'
  File "routers/agent_config.py", line 288, in get_agent_capabilities
    db_full_caps = db.get_full_capabilities(agent_name)
```

`get_full_capabilities` / `set_full_capabilities` live on `SecurityMixin` (composed into `AgentOperations`, i.e. `db._agent_ops`) but were never given a delegating pass-through on the `DatabaseManager` facade the routers call. So an agent's full-capabilities setting can be neither read nor changed via API/UI. Backend-independent (pure facade gap; SQLite + PostgreSQL).

## Fix

Add the two pass-throughs on `DatabaseManager`, alongside the existing read-only delegations (`src/backend/database.py`):

```python
def get_full_capabilities(self, agent_name: str) -> bool:
    return self._agent_ops.get_full_capabilities(agent_name)

def set_full_capabilities(self, agent_name: str, enabled: bool) -> bool:
    return self._agent_ops.set_full_capabilities(agent_name, enabled)
```

The `SecurityMixin` implementations already exist — this purely restores the facade passthrough.

## Tests

`tests/unit/test_database_facade_delegation.py` already whitelisted these two names in `KNOWN_FACADE_GAPS`. Removed them so the AST-based static guard now actively enforces the delegation, and added an explicit `test_1200_capabilities_methods_delegated` regression check.

```
unit/test_database_facade_delegation.py ... 3 passed
unit/test_capability_set.py             ... 4 passed
```

## Sweep (per issue note)

Swept every method on the `AgentOperations` mixins for un-delegated facade gaps. Only the two capabilities methods are both **missing and called** as `db.<m>(...)`. `is_agent_shared_with_email` is also un-delegated but has **0** `db.` call sites (latent, not a live 500) — left out of scope. The other `KNOWN_FACADE_GAPS` entries are pre-existing, unrelated to #1200.

Related to #1200. Sibling #1093 regression: #1199.


vybe (approved):

PR #1223, opened 2026-06-15
fix(ci): gate required schema-parity & verify-non-root via changes job (#1222)

### LINKED ISSUE #1222: bug(ci): path-filtered checks made required freeze the dev merge queue
## Symptom

Every PR to `dev` is stuck `BLOCKED` and cannot merge — even with an approval and admin privileges:

```
GraphQL: 2 of 4 required status checks are expected. (mergePullRequest)
```

This freezes the **entire** `dev` merge queue (9+ approved/ready PRs as of filing).

## Root cause

`dev` branch protection requires four status checks:
`schema-parity`, `verify-non-root`, `Analyze (python)`, `Analyze (javascript-typescript)`.

The first two come from **path-filtered** workflows:

| Required context | Workflow | `on.pull_request.paths` |
|---|---|---|
| `schema-parity` | `.github/workflows/schema-parity.yml` | `src/backend/db/**`, `database.py`, `utils/helpers.py`, … |
| `verify-non-root` | `.github/workflows/container-security.yml` | `docker/**`, `docker-compose*.yml`, `scripts/deploy/start.sh`, `src/mcp-server/Dockerfile` |

A PR that doesn't touch those paths (a dependency bump, a docs change, most feature work) never triggers the workflow, so the required context **never posts a status** → GitHub holds it `expected` forever → merge blocked. `enforce_admins: true` means even an admin override fails.

`schema-parity.yml` even carries an explicit maintainer note (lines 9–14) warning **not** to make it a required check for exactly this reason. It was added to required-status-checks anyway.

## Fix

Move the path filtering from the workflow trigger (`on.paths`) to a job-level `if:` driven by a cheap `changes` detector (`dorny/paths-filter`). A job **skipped via `if:` still posts a check run** with conclusion *skipped*, which branch protection counts as **passing** — so the required context is always present, while the heavy work runs only when the relevant surface changes.

Behavior after fix:
- Deps/docs/unrelated PR → both jobs **skip → pass** → merges normally.
- DB-touching PR → `schema-parity` **runs for real** → blocks on schema drift (intent preserved).
- Docker/compose PR → `verify-non-root` **runs for real** → blocks on root regression (intent preserved).

No branch-protection change needed: the fix PR edits both workflow files (each filter includes its own path), so both real jobs run on it → all four required contexts present → it self-merges with one approval.

## Acceptance criteria

- [ ] `schema-parity` and `verify-non-root` post a passing (skipped) status on a PR that touches neither DB nor docker paths
- [ ] `schema-parity` still runs for real (and can fail) when `src/backend/db/**` changes
- [ ] `verify-non-root` still runs for real (and can fail) when `docker/**` changes
- [ ] A previously-frozen dependency PR becomes mergeable

vybe:
## Problem

Every PR to `dev` is frozen `BLOCKED` — even with approval + admin:

```
GraphQL: 2 of 4 required status checks are expected. (mergePullRequest)
```

`dev` requires `schema-parity` and `verify-non-root`, but both come from **path-filtered** workflows (`on.pull_request.paths` = DB paths / `docker/**`). On a PR that doesn't touch those paths the workflow never runs, so the required context never posts a status → GitHub holds it "expected" forever → merge blocked. `enforce_admins: true` blocks the override too. `schema-parity.yml` even carried a maintainer note warning *not* to make it required for this exact reason.

## Fix

Move the path filtering from the workflow **trigger** to a cheap `changes` detector job (`dorny/paths-filter`) + a **job-level `if:`** on the heavy job:

- A job skipped via `if:` **still posts a check run** (`conclusion: skipped`), which branch protection counts as **passing** → the required context is always present.
- The heavy job runs **only** when the relevant surface changes → intent preserved:
  - `schema-parity` still runs (and can block) on `src/backend/db/**`, `database.py`, `utils/helpers.py`, …
  - `verify-non-root` still boots the stack (and can block) on `docker/**`, `docker-compose*.yml`, `scripts/deploy/start.sh`, `src/mcp-server/Dockerfile`.

### Behavior after merge
| PR touches… | `schema-parity` | `verify-non-root` |
|---|---|---|
| nothing relevant (deps, docs) | skip → ✅ pass | skip → ✅ pass |
| `src/backend/db/**` | **runs, can block** | skip → ✅ pass |
| `docker/**` | skip → ✅ pass | **runs, can block** |

### Self-merging — no branch-protection change needed
This PR edits **both** workflow files, and each filter includes its own path, so both real jobs run here and post all four required contexts. It merges with one approval; `enforce_admins`, the required-checks list, and `strict` are untouched.

## Test plan
- [x] YAML parses; jobs = `[changes, schema-parity]` / `[changes, verify-non-root]`
- [ ] On this PR: `schema-parity` + `verify-non-root` run for real (both workflow files changed) and pass
- [ ] After merge: a previously-frozen dependency PR (e.g. #1173) becomes mergeable
- [ ] After merge: confirm a deps PR shows `schema-parity`/`verify-non-root` as **skipped → passing**

Fixes #1222

### DIFF
--- a/.github/workflows/container-security.yml
+++ b/.github/workflows/container-security.yml
@@ -4,10 +4,11 @@ name: container-security
 #
 # These checks used to live in `frontend-e2e.yml`, which is `ui`-label
 # gated and therefore skipped on backend infrastructure PRs — the exact
-# PRs that can break them. This workflow runs unconditionally on any PR
-# (and push to dev/main) that touches the relevant Docker / compose /
-# bootstrap surface, so the guard executes whenever it could possibly
-# regress.
+# PRs that can break them. This workflow runs on every PR (and push to
+# dev/main); the heavy `verify-non-root` job is gated by the `changes`
+# detector so it executes whenever the Docker / compose / bootstrap
+# surface changes and is SKIPPED (→ passing required check) otherwise.
+# That keeps it a safe required status check — see #1222.
 #
 # Scope:
 #   - `verify-non-root`: every Trinity-built service that holds platform
@@ -25,19 +26,7 @@ name: container-security
 on:
   push:
     branches: [dev, main]
-    paths:
-      - 'docker/**'
-      - 'docker-compose*.yml'
-      - 'scripts/deploy/start.sh'
-      - 'src/mcp-server/Dockerfile'
-      - '.github/workflows/container-security.yml'
   pull_request:
-    paths:
-      - 'docker/**'
-      - 'docker-compose*.yml'
-      - 'scripts/deploy/start.sh'
-      - 'src/mcp-server/Dockerfile'
-      - '.github/workflows/container-security.yml'
   workflow_dispatch:
 
 # Least-privilege GITHUB_TOKEN: only `contents: read` is needed for
@@ -48,7 +37,37 @@ permissions:
   contents: read
 
 jobs:
+  # Cheap path detector — runs on every PR/push so the required `verify-non-root`
+  # context is ALWAYS produced. The heavy stack-boot job below is gated on it, so
+  # an unrelated PR skips it (→ passing required check) rather than leaving the
+  # context "expected" forever (the #1222 freeze).
+  changes:
+    runs-on: ubuntu-latest
+    # paths-filter resolves the PR's changed-file list via the API; the heavy
+    # verify-non-root job keeps the workflow-level least-privilege contents: read.
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      docker: ${{ steps.filter.outputs.docker }}
+    steps:
+      - uses: actions/checkout@v6
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            docker:
+              - 'docker/**'
+              - 'docker-compose*.yml'
+              - 'scripts/deploy/start.sh'
+              - 'src/mcp-server/Dockerfile'
+              - '.github/workflows/container-security.yml'
+
   verify-non-root:
+    needs: changes
+    # Skipped (→ reported as a passing required check) when no docker/compose
+    # paths changed; boots the stack and verifies UIDs for real when they did.
+    if: needs.changes.outputs.docker == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 20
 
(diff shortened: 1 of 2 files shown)


dolho (approved):

dolho:
## Review — LGTM ✅

Correct fix and the canonical GitHub pattern for "required + path-filtered" checks. Verified the diff, YAML validity, job graph, and the live check run.

**Why it's right**
- Root cause + mechanism are correct: `on.pull_request.paths` filters at the *workflow* level → no run → the required context never posts → GitHub holds it "expected" forever. Moving the filter to an always-running `changes` job (`dorny/paths-filter`) + a job-level `if:` on the heavy job is the documented solution — a job skipped via `if:` posts a `skipped` check run, which branch protection counts as passing.
- **Empirically proven on this PR**: the rollup shows `changes` ✅, `schema-parity` ✅, and `verify-non-root` ✅ all ran *for real* — because each filter self-includes its own workflow path and this PR edits both files. The self-merge claim holds; no branch-protection change needed.
- Intent preserved — filter path lists unchanged, both gates still run and can block on `db/**` / `docker/**`.
- `verify-prod-frontend-uid` is a step inside `verify-non-root`, so it's covered (no new always-on heavy job).
- Least-privilege is right: `changes` adds `pull-requests: read` (paths-filter API mode), heavy job keeps `contents: read`.

**Non-blocking suggestions**
1. **Supply chain (low):** `dorny/paths-filter@v3` is tag-pinned, not SHA-pinned — a floating tag can be repointed, and this runs with `pull-requests: read`. A commit-SHA pin would harden it. (Consistent with the repo's existing `actions/checkout@v6` tag-pinning, so optional.)
2. **Fail-open (low):** if the `changes` job itself fails (dorny/API hiccup), the dependent gate is skipped-due-to-failed-need → counts as passing → a db/docker-changing PR could merge without the gate actually running. Low probability and awkward to close fully; worth a comment acknowledging it.
3. The unchecked test-plan box (*"after merge: deps PR shows skipped → passing"*) is the one behavior only confirmable post-merge — recommend validating on a no-op PR (e.g. #1227 / #1229, which don't touch `db/**`) right after this lands.

**Merge-context heads-up (not this PR's fault):** currently `BLOCKED` on `lint (sys.modules pollution check)` — the pre-existing #1199 dev-red lint that #1213 fixes — plus `REVIEW_REQUIRED`. With `enforce_admins: true` there's a near-deadlock: this PR is blocked by the lint, and #1213 (the lint fix) was blocked by the phantom-check freeze *this* PR removes. Breaking it likely needs a one-time admin action (temporarily relax `enforce_admins`, merge one, restore).

Also: #1228 is a duplicate of #1222 — worth closing in its favor.

Issue #1197, opened 2026-06-13
bug: agent creation crashes with opaque ValueError on non-integer CPU in template resources


vybe:
## Summary

When an agent is created from a GitHub source repo whose `template.yaml` declares a non-integer CPU value (e.g. Kubernetes-style `cpu: \"0.5\"`), agent creation aborts with an opaque `ValueError: invalid literal for int() with base 10: '0.5'`. The repo-access validation, MCP key creation, subscription assignment and env-var setup all succeed first, then the container-create step crashes on a raw `int(cpu)`. The platform should validate/normalize template resource fields and return a clear, actionable error instead of crashing with a cryptic Python message — and should not leave partial side effects behind.

## Component

Backend / Agent Service (container creation)

## Priority

P2 — agent creation from an otherwise-valid source repo fails; workaround exists (edit the template to a valid CPU value), but the failure mode is opaque and leaves orphaned rows.

## Error

```
ERROR services.agent_service.crud: Failed to create agent [AGENT]: invalid literal for int() with base 10: '0.5'
```

## Location

- **File**: `src/backend/services/agent_service/crud.py`
- **Line**: 741
- **Function/Class**: container-create block in the agent create path
- **Same pattern (also unguarded)**:
  - `src/backend/services/agent_service/lifecycle.py:554` — `nano_cpus=int(cpu) * 1_000_000_000`
  - `src/backend/services/system_agent_service.py:248` — `nano_cpus=int(resources.get(\"cpu\", \"4\")) * 1_000_000_000`

## Root Cause

The CPU value taken from the agent's `template.yaml` `resources` block is fed straight into `int()`:

```python
nano_cpus=int(config.resources.get('cpu') or _get_default_resource('cpu')) * 1_000_000_000,
```

A non-integer value such as `cpu: "0.5"` (or any Kubernetes-style spec) raises `ValueError` and aborts the entire create. There is no validation/normalization of template resource fields against the allowed set (`cpu ∈ {1,2,4,8,16}`, `memory ∈ {1g,2g,4g,8g,16g,32g}`) before they reach Docker.

The adjacent memory line has the same class of problem — it passes the raw value to Docker's `mem_limit`:

```python
mem_limit=config.resources.get('memory') or _get_default_resource('memory'),
```

A Kubernetes-style `memory: "512Mi"` is not a valid Docker memory string (Docker expects `512m` / `4g`), so it would also fail; the `int(cpu)` error simply fires first.

[...]

PR #1227, opened 2026-06-16
fix(agents): validate template cpu/memory before container create (#1197)

dolho:
## Problem

Creating an agent from a GitHub source repo whose `template.yaml` declares a fractional / Kubernetes-style resources block (`cpu: "0.5"`, `memory: "512Mi"`) aborts deep in container-create with an opaque error:

```
ERROR services.agent_service.crud: Failed to create agent: invalid literal for int() with base 10: '0.5'
```

Repo validation, MCP-key creation, subscription assignment and env setup all succeed first; the raw `int(cpu)` for Docker's `NanoCpus` (#1126/#1128) then crashes. `agent_ownership` rolls back but an **orphaned `mcp_api_keys` row** is left per attempt. The adjacent `mem_limit` has the same class of bug (a k8s memory string isn't a valid Docker mem string).

## Fix

- **`capabilities.py`** (stdlib-only, the anti-drift home for container spec): add `normalize_cpu`/`normalize_memory` + canonical `VALID_CPU`/`VALID_MEMORY`. They reject fractional/k8s values with an actionable message and case-fold memory (`4G`→`4g`).
- **`routers/settings.py`** now imports those sets instead of duplicating the lists — the admin defaults endpoint and the create paths share one source of truth.
- **`crud.py` (create)**: normalize/validate `config.resources` **before any side effect** and raise a clear **HTTP 400** instead of a 500 from deep in Docker; canonical values written back for labels + limits. **Roll back the agent-scoped MCP key** in the failure path → no orphan rows.
- **`lifecycle.py` (recreate)** and **`system_agent_service.py`**: same guard at their `nano_cpus`/`mem_limit` sites (all three unguarded sites from the issue).

## Verification

`tests/unit/test_resource_normalization.py` — 20+ cases: valid set accepted, `0.5`/`512Mi`/etc. rejected with "must be one of …", case-folding, empty→default fallback, `int()`-castability of every normalized cpu, and a drift guard. Full unit run: **34 passed** (new + existing `test_capability_set`). `py_compile` clean on all five changed modules.

Related to #1197


vybe (approved):

Issue #1201, opened 2026-06-13
bug: agent-side timeout (504) drops execution telemetry — cost/context/tool-calls lost on timed-out runs


AndriiPasternak31:
## Summary

When an agent execution times out, the terminal `schedule_executions` write is
bare — `status='failed'` + an `error` string, and nothing else. All telemetry
accumulated before the kill (cost, context usage, tool calls, execution log,
session id) is discarded. Operators get zero cost accounting and zero
observability for the exact runs that are most expensive: the long, tool-heavy
ones that actually reach the timeout.

This is platform-wide — it lives in the universal execution path, so it affects
every agent regardless of template.

## Context

There are two distinct timeout cases with very different fixability. This issue
covers the **tractable** one.

**Agent-side budget timeout (this issue).** When the agent hits its own
`max_duration` budget or the per-tool stall watchdog, it raises a structured 504
in `docker/base-image/agent_server/services/headless_executor.py` (`:1037` and
`:1068`). At that point `ctx.metadata` — the cost/context/tool-call telemetry
parsed from the stream — is fully populated and in scope, but the 504 `detail`
body only carries `{message, termination_reason}`. The telemetry is one dict-key
away from being returned.

**Backend httpx ReadTimeout (out of scope here).** When the backend gives up
before the agent responds (`task_execution_service.py:942`), there is no response
body at all and telemetry genuinely can't be salvaged without a side-channel.
That's a bigger architectural problem and partly overlaps #1022.

## Current behavior

- **Success terminal** writes the full set — cost, context_used, context_max,
  tool_calls, execution_log, compact_metadata, claude_session_id
  (`src/backend/services/task_execution_service.py:896-908`).
- **Timeout terminal** writes only `status=FAILED` + `error`
  (`src/backend/services/task_execution_service.py:955-959`).

## Why this is cheap to fix

The backend already has the salvage machinery. The `httpx.HTTPError` handler
reads partial telemetry out of the agent's structured error body
(`task_execution_service.py:1013-1083`; it pulls `detail["metadata"]` at
`:1023` and rolls cost into the FAILED write — the #678 path). If the agent's
504 carried `metadata`, the same salvage logic would persist it.

## Acceptance Criteria

- [ ] Agent-side 504 timeout bodies (`max_duration` and `stall_no_output`
      branches in `headless_executor.py`) include the accumulated `ctx.metadata`
      under a `metadata` key, matching the #678 structured-error shape.
[...]

Issue #1022, opened 2026-06-02
bug: scheduler writes status='failed' with empty error on 30s dispatch timeout


obasilakis:
## Summary

Cron-triggered agent executions can land in `schedule_executions.status='failed'` with the **`error` column empty/NULL** and no `response`, giving operators zero diagnostics. Two scheduler defects combine: (1) a hardcoded **30.0s HTTP dispatch deadline** on the scheduler→backend `POST /api/internal/execute-task` call, and (2) a failure handler that records `str(exception)`, which is the **empty string** for an httpx timeout. When the backend event loop briefly stalls under concurrent cron fan-out, every in-flight dispatch times out at the same instant, producing synchronized batches of silent `failed` rows — currently the largest failure class on a production instance.

## Component

Scheduler (APScheduler dispatch loop)

## Priority

P2 — recurring silent-failure / observability gap; no workaround for triage since `error` is blank. The cheap, high-value fix is capturing the timeout text.

## Error

The failure is the *absence* of an error. Observable signatures:

```
# scheduler log — empty body after the colon:
Schedule {name} execution failed:
# pre-check on the same agents, empty parens = empty str(e):
[pre-check] backend call for {agent} failed () — fail-open
```

DB rows: `status='failed'`, `error=''` (empty), `response` NULL, `completed_at` set, `duration_ms` ≈ 30000–30200ms. Failures arrive in synchronized cross-agent batches (multiple executions starting within ~80ms, all completing at the same instant).

## Location

- **30s deadline**: `src/scheduler/service.py:1051` — `dispatch_timeout = 30.0`, applied at `:1058` (`client.post(..., timeout=dispatch_timeout)`). No try/except around this call, so an httpx timeout propagates out of `_call_backend_execute_task`.
- **Dropped diagnostic**: `src/scheduler/service.py:901` — `error_msg = str(e)` in the `except Exception` handler of `_run_scheduled_task`; written at `:918-922` via `update_execution_status(status=FAILED, error=error_msg)`.
- **DB writer**: `src/scheduler/database.py:404` writes `error = ?` verbatim (empty string), and computes `duration_ms` from `started_at`→`utcnow()` (≈30s).

(Line numbers from current `main`; observed on a deployment at commit `353b7c05`. May drift by a few lines.)

## Root Cause

[...]

Issue #678, opened 2026-05-06
Async chat_with_agent: long execution silently fails with null response (reader-thread)


pandore:
## Summary

Async `chat_with_agent` (parallel=true, async=true) on `ability-website` ran for ~24 minutes through 56 turns / 135 raw messages, then completed with `status=failed`, `response=null`, and **all telemetry (cost, context_used, model_used) null**. The platform's own error message describes a stdout-inheritance / reader-thread failure mode and asks for it to be filed.

## Execution metadata

| Field | Value |
|---|---|
| `execution_id` | `FdEVIjC5ZRVdeG-e5q0coQ` |
| `agent_name` | `ability-website` |
| `triggered_by` | `mcp` |
| `started_at` | `2026-05-06T11:58:54.069960Z` |
| `completed_at` | `2026-05-06T12:22:40.963313Z` |
| `duration_ms` | `1,426,893` (~23.8 min) |
| `status` | `failed` |
| `response` | `null` |
| `cost` | `null` |
| `context_used` | `null` |
| `model_used` | `null` |
| `execution_log` | `null` (also lost) |

## Error message (verbatim from `get_execution_result`)

> Execution completed without a result message after 0 tool calls / 56 turns (raw_messages=135). Likely cause: a tool or child subprocess inherited stdout and prevented the claude reader thread from capturing the final result block. Check agent-server logs for 'Reader thread(s) stuck after process exit' or 'I/O operation on closed file' near this execution. This is a transient infrastructure failure; retry the task.

The "0 tool calls / 56 turns" is suspicious — 56 turns with 0 tool calls is unusual for a real build task. Either the tool-call counter wasn't incremented (consistent with a reader-thread failure) or the agent really did spend 56 turns thinking without acting. The 135 raw messages suggests the former.

## Reproduction context

- Inbound: a single, long-form async `chat_with_agent` message — ~10 KB of structured markdown (build brief for a Next.js landing page, includes copy blocks, table-formatted form schema, GTM event list, acceptance criteria).
- Mode: `parallel=true, async=true, timeout_seconds=3600`
- Caller: separate Trinity instance (cross-instance MCP call from `Abilityai/trinity-gtm-agent` workspace).
- Agent received the message and started running (per `started_at` and turn count). No final result block ever surfaced to the orchestrator.

## Container-log signal possibly worth investigating

The startup logs for `ability-website` contain this warning before the run:

```
WARNING: Could not find target strings in /app/agent_server/services/claude_code.py
[...]

PR #1078, opened 2026-06-05
fix(headless): harden drain/finalize against leaked reader threads (#1025)

dolho:
## Summary

Salvages the two orthogonal robustness improvements from the closed PR #980 that the #973 fix (the actual 2h false-timeout bug) did not include. Pure hardening of the **already-bounded** drain/finalize path — defense-in-depth, no behavioural change on the clean path.

### D20 — capture daemon-thread exceptions in `_drain_bounded`
`subprocess_lifecycle._drain_bounded` swallowed every drain exception via `except Exception: pass`, making a drain that **raised** indistinguishable from a clean completion (and hiding the leaked reader). It now captures + logs the exception and returns a `DrainOutcome` (`completed` | `budget_exceeded` | `errored`). Existing callers that ignore the return value (claude_code.py) are unaffected.

### D19 — `_finalize_headless_result` snapshot isolation
On the budget-exceeded / errored drain path a reader thread is leaked and may still be mutating ctx's shared buffers / metadata. The run context now carries `drain_budget_exceeded`; when set, finalize rebinds `ctx` to a deep snapshot (`list(...)` of each buffer + `metadata.model_copy(deep=True)`) via `_snapshot_for_finalize`, retry-guarded against the "changed size during iteration" race with a live-ctx fallback. Clean drains keep the zero-copy fast path.

## Explicitly out of scope (per the issue)
- The #980 `asyncio.wait_for` outer-margin widening (#973 leaves it as-is)
- The #970 lsof/`/proc` pipe-holder hunt (reintroduces the D-state deadlock #817 removed)

## Tests
- `test_drain_bounded.py`: outcome matrix — `completed` / `budget_exceeded` / `errored`, plus no-swallow logging assertion
- `test_headless_finalize_snapshot.py` (new): list + metadata isolation from post-snapshot mutation, retry-exhaustion fallback to live ctx, default-off fast path
- Full related suite (`-k "drain or headless or claude or subprocess or pgroup or orphan"`): **130 passed**

Related to #1025

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1229, opened 2026-06-16
fix(agent): salvage execution telemetry on agent-side timeout 504 (#1201)

dolho:
## Problem

When an agent execution times out, the terminal `schedule_executions` write is bare — `status='failed'` + an `error` string. All telemetry parsed before the kill (cost, context, tool calls) is discarded, so operators get **zero cost accounting** for exactly the long, tool-heavy runs that reach the timeout.

## Root cause (the tractable case)

On an **agent-side** budget timeout (`max_duration`) or per-tool **stall** kill, `headless_executor.py` raises a 504 whose `detail` carried only `{message, termination_reason}`. `ctx.metadata` (cost/context/tool-calls) was fully populated and in scope — one dict-key away from being returned.

The backend already has the salvage machinery: the agent 504 reaches `response.raise_for_status()` → `httpx.HTTPStatusError` → the existing `except httpx.HTTPError` branch in `task_execution_service.py`, which reads `detail["metadata"]` and rolls cost/context onto the FAILED row (the #678 path). It just never received `metadata`.

## Fix (agent-side only)

- Add `_timeout_504_detail(ctx, message, reason, stalled_tool=None)` that attaches `metadata = sanitize_dict(ctx.metadata.model_dump())` — the same structured shape the #678 empty-result body uses — and route all three timeout-504 sites (outer safety-net, inner `max_duration`, `stall_no_output`) through it. Also de-duplicates the three near-identical bodies.
- **No backend change**: the existing HTTPError salvage branch now finds `metadata` and persists salvaged `cost` / `context_used` / `context_max`. `termination_reason` and the timeout-failure semantics are unchanged.

## Out of scope (per issue)

The backend **httpx ReadTimeout** case (`task_execution_service.py` `except httpx.TimeoutException`) — no response body exists, so telemetry genuinely can't be salvaged without a side-channel. Bigger problem, overlaps #1022.

## Verification

`tests/unit/test_1201_timeout_telemetry_salvage.py` — pins the agent-side contract: `max_duration` and `stall_no_output` bodies carry `metadata` with cost/context/tool fields; the metadata shape matches every key the backend salvage reads (`cost_usd`, `input_tokens`, `cache_*`, `context_window`); empty-metadata timeout still produces a well-formed body. Existing #1094 stall-reason tests still pass (6 passed total). `py_compile` clean.

> Note: takes effect once the agent base image is rebuilt and agents recreated (agent-server change).

Related to #1201


vybe:
⚠️ **Coordination note** — now live in `dev` via #1078 (merged): #1078 hardened `_finalize_headless_result` to snapshot the run context when a drain leaves a possibly-live reader (`reader_may_be_live`), but it deliberately does **not** set that flag on the timeout/kill path. Its stated justification: the 504 path never calls `_finalize_headless_result`, so there's no ctx read to protect there.

This PR changes that: `_timeout_504_detail(ctx, …)` now reads `ctx.metadata` (and salvages buffers) on exactly that kill path. If a timeout kill also leaks a reader past the `grace=3` drain budget, `metadata.model_dump()`/buffer reads can race the live reader — the precise tear #1078 fixes for finalize, reopened on the 504 path.

Suggest reading the 504 telemetry from a snapshot (reuse `_snapshot_for_finalize`) or setting `reader_may_be_live` + snapshotting on the kill path before this merges. Refs: #1078 / #1025.

vybe (approved):

Issue #1230, opened 2026-06-16
bug: backend Docker healthcheck flaps to "unhealthy" under batch load (10s timeout too tight for 2 workers)


obasilakis:
## Summary

The `trinity-backend` container's Docker healthcheck intermittently trips to **`unhealthy`** while the service is fully up and serving `/health` 200s. Observed on the **eu2** instance (prod compose, version `c52714b6`). It is a false negative caused by a too-strict probe, not an actual backend failure — but it is a latent risk anywhere an `unhealthy` state triggers a restart/eviction (autoheal, orchestrator, LB drain).

## Healthcheck definition (current)

```json
{"Test":["CMD","curl","-f","http://localhost:8000/health"],
 "Interval":30s,"Timeout":10s,"StartPeriod":10s,"Retries":3}
```
Backend command: `uvicorn main:app --host 0.0.0.0 --port 8000 --workers 2 --timeout-worker-healthcheck 120`

## What happens

With only **2 uvicorn workers**, during scheduled-batch windows the workers are busy (observed backend at ~47% CPU mid-batch) and GIL-contended. A `/health` probe then waits longer than the **10s timeout** and is aborted. Three consecutive aborts (Retries=3) flip the container to `unhealthy`; once the batch eases, the next probe succeeds and it flips back to `healthy`.

## Evidence (eu2, captured live)

Probe log around a flap — three timeouts (`exit=-1`) then recovery (`exit=0`):
```
2026-06-16 12:21:58Z | exit=-1
2026-06-16 12:22:39Z | exit=-1
2026-06-16 12:23:19Z | exit=-1
2026-06-16 12:23:59Z | exit=0
2026-06-16 12:24:29Z | exit=0
```
At time of capture: `docker ps` showed `Up 4 hours (unhealthy)` while `curl /health` returned `{"status":"healthy"}`. Shortly after: `Health=healthy FailingStreak=0`. `RestartCount=0`, `OOMKilled=false` — the service never went down.

## Impact

- Cosmetic today (service stays up; `/health` returns 200 throughout).
- **Risk:** any consumer of Docker health state (autoheal sidecar, compose `depends_on: condition: service_healthy`, external health-based routing) can act on the false `unhealthy` and restart/evict the backend mid-batch — which orphans in-flight executions (we already see "recovered on backend restart" rows after backend recreates).
- Misleads operators/monitoring that read container health rather than the `/health` endpoint.

## Suggested fix (config only, `docker-compose.prod.yml`)

[...]

PR #1232, opened 2026-06-16
fix(ci): relax backend healthcheck timeout so batch load doesn't flap it unhealthy (#1230)

dolho:
## Problem

The `trinity-backend` container's Docker healthcheck intermittently trips to **`unhealthy`** while the service is fully up and serving `/health` 200s (observed live on eu2, prod compose). With only **2 uvicorn workers**, scheduled-batch windows leave them GIL-contended; a `/health` probe then waits longer than the **10s timeout** and is aborted. Three consecutive aborts (`retries: 3`) flip the container to `unhealthy`; once the batch eases the next probe succeeds and it flips back.

It's a false negative (service serves 200s throughout, `RestartCount=0`), but a latent risk: any consumer of Docker health — autoheal sidecar, compose `depends_on: condition: service_healthy`, health-based LB drain — can act on the false `unhealthy` and restart/evict the backend **mid-batch**, orphaning in-flight executions.

## Fix (config only — `docker-compose.prod.yml` backend)

| field | before | after |
|---|---|---|
| `timeout` | 10s | **30s** |
| `start_period` | 10s | **60s** |
| `retries` | 3 | **5** |

`interval` stays 30s. A genuine outage (all probes fail) still trips after ~5 intervals (~2.5 min); transient batch spikes no longer do. Scoped to the backend probe only — the scheduler/mcp/vector probes aren't the GIL-contended 2-worker flapper, so they're left unchanged (minimal scope).

A dedicated lightweight liveness route (no DB/work touch) would be even more robust, as the issue notes, but the timeout bump alone eliminates the observed flapping; left as a possible follow-up.

## Verification

`docker-compose.prod.yml` parses; backend healthcheck resolves to `{interval: 30s, timeout: 30s, retries: 5, start_period: 60s}`.

Related to #1230


vybe (approved):

Issue #1231, opened 2026-06-16
fix: agent /tmp tmpfs (100 MB) fills and silently breaks autonomous git commits — make size configurable


vybe:
## Summary

Agent containers mount `/tmp` as a 100 MB `noexec,nosuid` RAM-backed tmpfs, hardcoded in `src/backend/services/agent_service/capabilities.py` (`AGENT_TMPFS_MOUNT = {'/tmp': 'noexec,nosuid,size=100m'}`). It fills easily — e.g. leftover `gh` CLI install artifacts (`gh.tar.gz` + extracted dir, ~38 MB) plus normal scratch usage. Once full, **every** `/tmp` write fails with `No space left on device`, including git's scratch during commit. Autonomous scheduled runs then "complete" but fail to persist (commit/push), leaving the run's output as stranded working-tree drift.

Because the size is a hardcoded literal, operators can't tune it without a code change + base-image rebuild. We should make the size an **instance-level env var** (deployment-tier config) with a safe default.

## Context

Reported on a production agent (base image v0.6.0); the cause is in the base image, so it is **fleet-wide**.

`#1098` already redirects heavy build scratch off `/tmp` via `TMPDIR=/home/developer/.tmp` (disk-backed, exec-capable), injected at create (`crud.py`) and recreate (`lifecycle.py`), dir created by `docker/base-image/startup.sh`. But `TMPDIR` only helps tools that **honor `$TMPDIR`** — install scripts that hardcode `/tmp` (the `gh` case here) bypass it entirely and still exhaust the 100 MB cap.

Diagnosis is hard because `No space left on device` points at the disk (which is ~42% full, looks fine via `df -h /`) rather than the tiny tmpfs (`df -h /tmp`).

The tmpfs is RAM-backed and `noexec,nosuid` by deliberate security design (a compromised agent can't stage/execute payloads there). That posture must be preserved — only the **size** should be tunable, and it must stay bounded since tmpfs size counts against the container memory cgroup.

## Acceptance Criteria

- [ ] `/tmp` tmpfs size is read from a backend env var (e.g. `AGENT_TMP_SIZE`), defaulting to `512m`, in the single source of truth (`capabilities.py` `AGENT_TMPFS_MOUNT`) so create + recreate paths can't drift
- [ ] `noexec,nosuid` flags remain hardcoded — only size is configurable
[...]

PR #1233, opened 2026-06-16
fix(agent): make agent /tmp tmpfs size configurable via AGENT_TMP_SIZE (#1231)

dolho:
## Problem

Agent containers mount `/tmp` as a **hardcoded 100 MB** `noexec,nosuid` RAM-backed tmpfs (`capabilities.py`). It fills easily — e.g. leftover `gh` CLI install artifacts (~38 MB) plus normal scratch. Once full, **every** `/tmp` write fails with `No space left on device`, including git's commit scratch, so autonomous scheduled runs "complete" but silently fail to persist (commit/push). Because the size was a literal, operators couldn't tune it without a code change + base-image rebuild. Fleet-wide (base image).

`#1098` already redirects `$TMPDIR`-honoring tools off `/tmp`, but install scripts that **hardcode `/tmp`** (the `gh` case) bypass it and still exhaust the cap.

## Fix

- **`capabilities.py`** — `AGENT_TMPFS_MOUNT` size is now read from **`AGENT_TMP_SIZE`** (env on the backend service, which builds the agent mount spec), **default `512m`**, validated `^\d+[mg]$` with empty/invalid → default. `noexec,nosuid` stay **hardcoded** — only size is configurable, and it stays bounded (tmpfs counts against the agent memory cgroup). Single source of truth → create (`crud.py`) and recreate (`lifecycle.py`) can't drift.
- Wire `AGENT_TMP_SIZE=${AGENT_TMP_SIZE:-512m}` on the `backend` service in **both** `docker-compose.yml` and `docker-compose.prod.yml`; document in `.env.example`.
- **`architecture.md`** Container Security — note the now-configurable size.

| AGENT_TMP_SIZE | result |
|---|---|
| unset / `""` / `512` / `512Mi` / `0.5g` | → default `512m` |
| `256m`, `2g`, `1G` | → used (`1G`→`1g`) |

## Acceptance criteria

- [x] size from `AGENT_TMP_SIZE`, default `512m`, single source of truth (`capabilities.py`)
- [x] `noexec,nosuid` remain hardcoded — only size configurable
- [x] value validated (`^\d+[mg]$`); empty/invalid → default (never broken/unbounded)
- [x] documented in `.env.example` + both compose files
- [x] Container Security section of `architecture.md` updated
- [ ] *deploy-time:* verify inside a freshly recreated agent with `df -h /tmp` (mount specs are creation-time — existing agents pick up the new size on **recreate**, not restart)

## Verification

`tests/unit/test_1231_agent_tmp_size.py` — default / valid `m`,`g` / case-fold / invalid→default / security-flags-never-dropped. Full unit run **22 passed** (new + existing `test_capability_set`). Both compose files parse with `AGENT_TMP_SIZE` present; `py_compile` clean.

> Out of scope (separate `abilities`-repo issue): the agent-side git-sync hook that reports success while silently failing to persist. Full `/tmp` is only the trigger.

Related to #1231


vybe (approved):

Issue #847, opened 2026-05-15
Spike: enterprise edition architecture — private module strategy for compliance features (SSO, SCIM, SIEM)


vybe:
## Summary

Research and define the architecture for a Trinity Enterprise edition — compliance-gating features (SSO/SAML, SCIM provisioning, SIEM log export) that live in a **private repository**, not the public OSS repo, while remaining easy to develop, test, and ship alongside the core platform.

## Context

Trinity is moving toward a dual-tier model:
- **Community (OSS)**: MIT or AGPL — full agent runtime, scheduling, multi-channel, orchestration. Everything currently in this repo.
- **Enterprise (paid)**: SSO/SAML, SCIM, SIEM export — the three features that cause a SOC2/ISO audit to fail without them. These must not ship in the public repo.

The problem this spike solves: how do we build enterprise features in a private repo while keeping one coherent development process, shared CI, and a clean integration path into the Trinity runtime?

## Research Questions

1. **Module integration mechanism** — Python package (installable via private PyPI / GitHub Packages), git submodule, or plugin registration hook in `main.py`? What is the least coupling with the most isolation?
2. **Development workflow** — Can a developer work on enterprise features locally alongside the OSS repo without manual symlinks or path hacks? What does the local dev setup look like?
3. **CI/CD** — How does the private repo's CI run tests against the OSS core? Does the OSS repo publish a dev package the enterprise repo depends on?
4. **License enforcement** — What mechanism gates enterprise features at runtime? License key (signed JWT), environment flag, or trust-based (enterprise customers get the private package URL)?
5. **Trinity Cloud integration** — How does the Cloud deployment bundle the enterprise package? Is it a build step or a runtime dependency?
6. **Repository structure** — One private monorepo (`trinity-enterprise`) or separate repos per feature domain?

## Acceptance Criteria

- [ ] Spike document written: compares at least 2 integration approaches (Python package vs. plugin hook) with tradeoffs on maintenance burden, local DX, and CI complexity
- [ ] Proof-of-concept: a minimal enterprise feature (stub SSO endpoint) wired into Trinity via the chosen mechanism, running locally without modifying the public repo
- [ ] Local dev setup documented: steps from clone to running Trinity + enterprise features in under 15 minutes
- [ ] CI strategy defined: how the private repo tests against the OSS core without tight coupling
- [ ] Decision recorded: chosen approach with rationale, documented in `docs/planning/ENTERPRISE_ARCHITECTURE.md`

## Technical Notes

[...]

PR #1247, opened 2026-06-17
feat(auth): two-factor (TOTP) login seam + enterprise-gated UI (trinity-enterprise#5)

dolho:
## What
OSS side of **two-factor authentication (TOTP)** — enterprise issue [Abilityai/trinity-enterprise#5](https://github.com/Abilityai/trinity-enterprise/issues/5).

Per the open-core pattern (#847), all algorithmic IP (TOTP verify, AES secret store, recovery codes, per-role policy) lives in the private `trinity-enterprise` submodule (separate PR). This PR adds only the **edition-agnostic login seam** + the **entitlement-gated Vue surface**.

## Backend seam
- `services/mfa_gate.py` — the single hook the enterprise provider registers into. **No provider (OSS-only build) → `gate_login()` returns `None` → login is byte-for-byte unchanged.** Fail-open on provider error.
- `dependencies.py` — `create_mfa_challenge_token` / `decode_mfa_challenge`; a challenge-scoped token is **rejected as a session token** in `get_current_user` + `decode_token`.
- `routers/auth.py` — `/token` and `/api/auth/email/verify` consult `mfa_gate` after the first factor; if a second factor is required they return a short-lived **challenge token** instead of an access token (audited `mfa_challenge_issued`).
- `models.py` — `Token` gains optional `mfa_required` / `challenge_token` fields.
- `docker/backend/Dockerfile` — `pyotp` (the enterprise module runs in this image).

## Frontend (gated by `2fa` in `GET /api/settings/feature-flags`)
- **Settings → Security** tab (`TwoFactorPanel.vue`): self-service enroll / confirm / disable / recovery codes; admin-only org policy.
- **Login.vue**: second-factor step (verify + forced-enroll) after the password/email factor; `QrCode.vue` renders the `otpauth` URI with a graceful manual-key fallback.

## Edition behavior
| Build | Security tab | Login step | `/api/enterprise/2fa/*` |
|---|---|---|---|
| OSS-only | hidden | none | 404 |
| Enterprise, entitled | visible (all users) | when enrolled/required | 200 |

## Testing
Verified end-to-end on a live stack (enroll → confirm → gated re-login → `login/verify` → working token; replay rejected; recovery-code login; challenge-token-as-session rejected). Enterprise unit + real-DB pytest 20/20.

> Depends on the enterprise PR for the actual module; this seam is inert until the submodule registers a provider.

Related to Abilityai/trinity-enterprise#5


dolho:
image
image

vybe (approved):
Validated via /validate-pr: OSS 2FA login seam (trinity-enterprise#5). Verified security-critical property — challenge tokens carry mfa_challenge scope and are rejected as session tokens in both decode_token and get_current_user; OSS-only build is byte-for-byte inert (provider is None). pyotp pinned, architecture.md updated, E2E + 20/20 enterprise pytest. Inert without the enterprise provider, so no half-ship risk. Approving.

Issue #1183, opened 2026-06-12
feat(db): adopt Alembic for PostgreSQL migrations (retire SQLite-only bespoke runner)


dolho:
## Summary

Adopt **Alembic** as the migration system for the PostgreSQL backend, replacing the SQLite-only bespoke runner (`db/migrations.py`: `PRAGMA table_info` + `INSERT OR IGNORE` + raw `ALTER`) and the fresh-build-from-`schema.py` path (`init_schema_postgres`). This is the migration story for the **Postgres-only future** — SQLite is being dropped, so we do not invest in dual-backend Alembic; SQLite keeps its legacy bespoke path until it is removed.

Long-term goal: `db/tables.py` SQLAlchemy `MetaData` becomes the **single source of truth**, ending the current 3-way hand-sync (`schema.py` CREATE strings / `tables.py` Core handles / `migrations.py` ALTERs). Relates to #746 (collapse schema.py + migrations.py), #1160 (migration runner data-loss / no cross-process serialization), #721 (schema drift), #300 (the Core abstraction this builds on).

## Motivation

- **No in-place upgrade path for existing Postgres DBs today.** `init_schema_postgres` only `CREATE ... IF NOT EXISTS` — a pre-existing PG database never picks up columns added later. Fresh-from-head only.
- The bespoke runner is SQLite-bound (PRAGMA, `INSERT OR IGNORE`, `sqlite3.OperationalError`), has a DROP-rebuild data-loss window and no cross-process serialization (#1160), and forces triple-maintenance per schema change.
- Alembic is SQLAlchemy-native: dialect-aware DDL, ordered revisions with **down-migrations**, `autogenerate` diffing live DB vs `tables.py` metadata, and a battle-tested version table + locking story.

## Two migration domains (must preserve)

`src/backend/enterprise/` is a **separate (private) repo** with its own migration system (`backend/_migrations.py`):
- Separate tracking table `enterprise_schema_migrations`; owns only `enterprise_*` tables; **never ALTERs an OSS table**; runs from `register_enterprise()` AFTER OSS `init_database`.
- Already mimics Alembic `versions/` style (`discover()` + module-prefixed names like `siem_0001_…`).

Alembic adoption must keep the two domains isolated:
- **OSS** revisions in this (public) repo, version table `alembic_version`.
[...]

Issue #456, opened 2026-04-22
bug: _migrate_sync_health race condition between uvicorn workers


vybe:
## Summary

`_migrate_sync_health` (introduced in #389) uses a check-then-act pattern (`PRAGMA table_info` followed by `ALTER TABLE ADD COLUMN`) that is not atomic under concurrent execution. When the backend starts with multiple uvicorn workers (default is `--workers 2`), both workers run `run_all_migrations()` on startup and race: worker A's `PRAGMA` returns no column, worker A ALTERs successfully, worker B's `PRAGMA` also returned no column (before A committed), worker B ALTERs and hits `duplicate column name: auto_sync_enabled`. The losing worker crashes during startup.

End-state is correct (column exists, worker respawns and finds it present), so this is not a data or functionality bug — but every cold start logs a scary traceback and `Child process [N] died`, which will train operators to ignore real migration failures.

## Component

Backend / Migrations

## Priority

P2 — feature impaired (migration fails on one worker), workaround is \"ignore the traceback, column ends up present.\"

## Error

```
Migration FAILED (sync_health): duplicate column name: auto_sync_enabled
Traceback (most recent call last):
  File \"/app/db/migrations.py\", line 80, in run_all_migrations
    migration_fn(cursor, conn)
  File \"/app/db/migrations.py\", line 1404, in _migrate_sync_health
    cursor.execute(
sqlite3.OperationalError: duplicate column name: auto_sync_enabled
INFO:     Child process [8] died
```

## Location

- **File**: `src/backend/db/migrations.py`
- **Line**: ~1404 (inside `_migrate_sync_health`)
- **Function**: `_migrate_sync_health`
- **Docstring claim**: \"Idempotent: guards every change with PRAGMA table_info / table-existence checks.\" — idempotent single-threaded, but not concurrent-safe.

The same pattern is used for the second ALTER (`freeze_schedules_if_sync_failing`) and the `CREATE TABLE agent_sync_state` block in the same function, and likely in other migrations that add columns. Fix should be applied systemically.

## Reproduction Steps

1. Fresh deploy of a commit that includes #389 (base image + backend rebuild) against a DB that does not yet have `agent_git_config.auto_sync_enabled`.
2. Start backend with `--workers 2` or more (this is the default production config).
[...]

PR #1263, opened 2026-06-18
fix(db): atomic table rebuilds + cross-process lock in migration runner (#1160)

### LINKED ISSUE #1160: fix(db): migration runner — DROP-rebuild data-loss window and no cross-process serialization
## Summary

Two correctness defects in the migration runner (`src/backend/db/migrations.py`):

1. **Data-loss window in DROP-rebuild migrations.** Two table-rebuild migrations hold the only copy of `agent_sharing` / `agent_skills` rows in a Python list between an autocommitted `DROP TABLE` (Python 3.11 legacy isolation makes DDL autocommit) and the re-insert commit. A crash in that window silently and permanently loses access-control rows — and the replay guard then no-ops and marks the migration applied.
2. **No cross-process lock.** Production runs the full suite concurrently from two uvicorn workers. This race already caused two boot failures (#456, #389); only the ADD COLUMN class was hardened afterward.

Related: 71 `conn.commit()` sites inside migration bodies mean zero per-migration atomicity — replay safety after a mid-migration crash rests entirely on hand-written PRAGMA guards, and no test replays a partially-applied migration.

## Context

2026-06-11 audit finding N3 (High as a cluster). #746/#300 cover schema-source consolidation and the eventual Alembic move, but neither's AC covers atomicity, locking, or replay testing. Note: the 2026-06-02 internal refactor audit tagged `db/migrations.py` as "append-only growth (lower real risk)" — that was a size-only assessment; these are correctness defects complexity tooling cannot see. Per the target architecture, Alembic arrives with PostgreSQL (#300) — this issue hardens the existing runner in the interim, not a rewrite.

## Acceptance Criteria

- [ ] Cross-process lock around the runner (e.g. `BEGIN IMMEDIATE` on a lock row or flock) — concurrent-boot test passes
- [ ] The two DROP-rebuild migrations rewritten to rename-swap (old table dropped only after the new one is committed)
- [ ] Replay-after-partial-apply test added
- [ ] Failing migration named in the startup 503

## Technical Notes

Migration body checksums (detect edited-after-apply) and downgrade/dry-run/backup hooks are adjacent but out of scope here — they belong on #746/#300.

obasilakis:
## Summary
Hardens the SQLite migration runner against two correctness defects (#1160):
- **DROP-rebuild data-loss window** — `agent_sharing` and `agent_skills` rebuilds held access-control rows only in a Python list between an autocommitted `DROP TABLE` and the re-INSERT; a crash there lost them permanently. Now done via `_atomic_rebuild` (rename-swap inside an explicit `BEGIN/COMMIT`, copy done server-side).
- **No cross-process serialization** — backend workers + scheduler container each ran the suite on boot (caused #456/#389). Now serialized by an OS `flock` (`db/migration_lock.py`) wrapping both migration passes + `init_schema`.

## Changes
- `src/backend/db/migrations.py` — `_atomic_rebuild` helper; both rebuild migrations rewritten to use it (preserving `lower(email)` + NULL/empty-drop semantics and the skills indexes); failing migrations named on the traceback via `add_note` (original exception type preserved); new `migration_health()` helper.
- `src/backend/db/migration_lock.py` *(new)* — stdlib-only flock context manager (fails open; kernel-released on process death).
- `src/backend/database.py` — `init_database()` wraps both passes + `init_schema` in `migration_lock(DB_PATH)`, after the `is_sqlite()` guard.
- `src/backend/main.py` — `/health` 503 reports `first_pending` via `migration_health`.
- `docs/memory/architecture.md` — invariant #3 updated; `docs/security-reports/cso-2026-06-18.md` — diff audit (PASS, 0 findings).

## Acceptance criteria
- [x] Cross-process lock; concurrent-boot test passes
- [x] Two DROP-rebuilds → rename-swap (old dropped only after new committed) — strengthened to a single explicit transaction
- [x] Replay-after-partial-apply test
- [x] Failing migration named (traceback `add_note` + `/health` `first_pending`)

## Test Plan
- [x] `pytest tests/unit/test_1160_migration_atomicity.py` — 10 tests (crash-rollback at DROP and RENAME, replay-after-partial, behavior preservation, multiprocessing cross-process lock + concurrent-boot rebuild)
- [x] Existing migration tests unaffected (`test_migrations.py`, `test_migrations_concurrent.py`)
- [x] Live: 8 concurrent real `init_database()` processes against an old-schema DB → rebuild applied exactly once, zero data loss, no orphaned `*_new` tables

## Notes
SQLite-only — the Postgres/Alembic path returns early in `init_database()`, so no Alembic revision is needed (dual-track invariant #1183). The literal "503 on hard failure" is partially unreachable (module-level init crash-loops before `/health` binds); the traceback name is the primary operator signal.

Fixes #1160


vybe (approved):
Validated via /validate-pr: SQLite migration-runner hardening (atomic rename-swap inside explicit BEGIN/COMMIT + cross-process flock + `/health` first_pending). No schema change → no dual-track Alembic requirement. CSO diff audit PASS (0 findings). All required checks green (full pytest matrix, prod-image-smoke, schema-parity, verify-non-root). Fixes #1160.

Issue #1082, opened 2026-06-05
refactor: status-as-projection — schedule_executions.status never read as authoritative for "is running"


vybe:
## Summary

Make `schedule_executions.status` a compare-and-set-guarded projection of the execution's terminal event, and ensure **no code path reads `status` as the authoritative answer to "is this execution running."** Today that single fact is split across the Redis slot ZSET, the SQL row, and agent RAM — which is the reason the cleanup + canary reconciliation machinery exists.

## Context

Bankable reliability win #1 from the 2026-06-05 coordination review (`docs/planning/TARGET_ARCHITECTURE.md` §Coordination Model). It is independent of the full push→pull migration and can ship on its own. Collapsing "is running" to a single owner structurally removes split-brain and lets canary invariant S-01 (slot↔row bijection) retire.

## Acceptance Criteria

- [ ] `status` transitions are applied only through a compare-and-set guard (RELIABILITY-005 pattern) — a stale writer can never clobber a terminal row
- [ ] Audit every reader of `status` (notably `src/backend/services/cleanup_service.py`) and remove any that treats it as the authoritative "is running" signal
- [ ] Canary S-01 documented as redundant once the single-owner invariant holds (retire or downgrade)
- [ ] No behavior change for callers — projection semantics only

## Technical Notes

- Relevant: `src/backend/db/schedules.py` (status writes / CAS), `src/backend/services/task_execution_service.py`, `src/backend/services/cleanup_service.py`, `src/backend/canary/`.
- Lands cleanly before pull so the migration inherits a single-owner status.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1081, opened 2026-06-05
feat: pull / work-stealing coordination (push→pull dispatch)


vybe:
## Summary

Migrate agent coordination from **push** (backend dispatches into a per-agent mailbox) to **pull / work-stealing**: the backend owns one durable per-agent queue (`schedule_executions` rows, status `queued → claimed → running → terminal`); each agent's worker pool pulls the next task only when it has a free worker, runs it, and reports the result back. A busy or dead agent simply stops pulling, so work never lands on an agent that isn't ready.

## Context

The destination shape adopted in the 2026-06-05 review and documented in `docs/planning/TARGET_ARCHITECTURE.md` §Coordination Model. Chosen over the push actor model for simplicity, single-source-of-truth, and ~80% reuse of existing primitives (`claim_next_queued`, the RELIABILITY-005 CAS, idempotency keys, the agent `auto_sync.py` loop precedent). Delivers the two operator levers: inbox depth = `COUNT(status='queued')`, capacity = `max_parallel_tasks` enforced physically by the agent's worker count.

**v2 side-effect reframe (2026-07-01, doc #1404).** Side-effect safety moved from *effect-key universal sink coverage* ("Direction A", #1084 — treated as *the gate that blocks pull*) to **retry-with-prior-trace recovery + deterministic tool-side gates on capability-confined irreversible rails** ("Direction B"). Consequence for this umbrella: **pull default-on now gates per-effect, not per-agent** — read/analysis-only + reversible + confined-irreversible agents can default on immediately; only irreversible-and-un-confineable effects (the agent's own email key / `gh` / `curl`, with no gated Trinity tool in the path) wait, human-gated via the **asynchronous** operator queue (no synchronous in-turn prompts — agents fire-and-park). **#1084 no longer blocks pull default-on.** The load-bearing problem *moves* to trace fidelity (the open #548/#333 reader-race family), it does not disappear. See doc §"Re-Delivery and Side-Effect Recovery" + Open Question 2a.

## Acceptance Criteria (phased, per the doc)

- [x] **Phase 0** — schema: add `claim_token`, `lease_expires_at`, `claimed_by_worker` to `schedule_executions` (dark, nullable) — Alembic `0016`
- [x] **Phase 1** — dark pull endpoints: `GET /api/internal/next-task` (wraps `claim_next_queued`, stamps a lease) + `POST /api/internal/tasks/{id}/result` (CAS)
[...]

PR #1268, opened 2026-06-19
refactor(reliability): status-as-projection — CAS-guard the status-write gap (#1082)

AndriiPasternak31:
## Summary

Establishes `schedule_executions.status` as a **CAS-guarded projection** of an execution's terminal event (#1082, Step 1 — backend scope). The agent process registry remains the runtime authority for "is running"; no backend reader treats `status='running'` as the standalone authority.

- **CAS gap closed:** `db/schedules.py:update_execution_to_queued` (the overflow re-queue path) previously wrote `status=QUEUED WHERE id=?` with **no status precondition**, so a stale/duplicate call could resurrect a terminal row into `queued` (the E-02 phantom-reversal class). It now carries `AND status == RUNNING`, mirroring the sibling `release_claim_to_queued`.
- **Regression guard** (`tests/unit/test_schedule_status_observability.py`): a static AST guard over `db/schedules.py` (every `update(schedule_executions)` status writer must reference `schedule_executions.c.status` in a predicate, with a curated inventory tripwire) + behavioural no-op proofs that drive a real row to each terminal state and assert each writer is a no-op.
- **Canary S-01** (slot–row bijection) downgraded `critical → major` and annotated redundant under single-owner status — stays registered/Tier-A, retires with the slot ZSET in #1081 Phase 5.

## Scope & documented follow-up

The single-owner contract is enforced in the **backend** process only. The standalone **scheduler** (`src/scheduler/`) writes the same `trinity.db` with raw-SQL status writers (`update_execution_status`, `schedule_retry`) that are **not** yet CAS-guarded — the cron path compensates with a non-atomic read-then-check, but the retry-failure path does not. This is **pre-existing**, outside Step 1's backend scope, and now explicitly documented as a tracked #1082 follow-up. The static guard is file-scoped to `db/schedules.py`; its blind spots are enumerated in the test docstring so a green guard isn't misread as whole-system coverage.

## Changes
- `db/schedules.py` — `AND status == RUNNING` guard on `update_execution_to_queued`.
- `tests/unit/test_schedule_status_observability.py` — static AST guard + behavioural no-op proofs + blind-spot documentation.
- `canary/invariants/s01_slot_row_bijection.py` — severity downgrade + annotation.
- `docs/memory/architecture.md`, `docs/memory/feature-flows/status-as-projection.md` — subsystem doc + feature flow, including the scheduler-writer "Known gap" section.

## Testing
- `tests/unit/test_schedule_status_observability.py` (static + behavioural regression suite for the invariant).

Refs #1082


vybe (approved):
Validated via /validate-pr: P1 reliability refactor (#1082 Step 1). CAS guard 'AND status == RUNNING' on update_execution_to_queued mirrors release_claim_to_queued and closes the E-02 phantom-reversal gap; static AST guard + behavioural no-op proofs back it; scheduler-process writer gap honestly documented as a tracked follow-up. Docs (architecture + new feature-flow + invariant catalog) updated. Mergeable, security clean. Approving.

Issue #1265, opened 2026-06-18
bug(perf): Dashboard & timeline take 20s+ to load metrics with 10+ agents — UI frozen until data arrives


vybe:
## Summary

On instances running **10+ agents**, the initial metrics load on the web UI takes **20 seconds or more** — most visibly in the timeline view and on the Dashboard overall. While the data is in flight the UI shows no metrics and feels frozen/dead, with no indication that anything is loading.

This is a latency/scaling problem: load time grows with the number of agents. It is the broader, fleet-scale cousin of the (already fixed) #1096, which addressed only `/api/telemetry/containers`. The remaining slow paths appear to be the cross-agent timeline and the fleet/metrics aggregation endpoints, plus any per-agent request fan-out from the Dashboard.

> Paired with the perceived-performance/loading-feedback work in the companion issue (loading dynamics). This issue is about making the data actually arrive faster; the companion is about not looking dead while it does.

## Context

- Reported on a fleet with 10+ agents; first paint of Dashboard / timeline metrics blocks ~20s+.
- #1096 (closed) fixed a ~6s stall on `/api/telemetry/containers` (thread-pool cap + no cache). The symptom persists more broadly, so other endpoints / query patterns are now the bottleneck.
- #359 (closed) previously eliminated N+1 API calls on Dashboard load via a bulk permissions endpoint — suggests the Dashboard still has fan-out / N+1 tendencies as agent count grows.

## Acceptance Criteria

- [ ] Profile the initial Dashboard + timeline load on a ≥10-agent fleet; identify which endpoints/queries dominate the 20s (candidates: cross-agent activities timeline, fleet executions stats, monitoring/health aggregation, per-agent fan-out).
- [ ] Document measured per-endpoint timings (warm + cold) at 1, 5, 10, 20 agents to characterize how cost scales with fleet size.
- [ ] Reduce time-to-first-meaningful-metrics for a 10-agent fleet to a target (propose a budget, e.g. < 3s) via server-side aggregation, query optimization, and/or short-TTL caching.
- [ ] Eliminate any remaining N+1 / per-agent request fan-out on the initial Dashboard/timeline load (single aggregate call where possible).
- [ ] No regression in metric freshness/correctness (caching, if added, must invalidate or expire sensibly).
- [ ] Verify with the canary/timeline views that numbers still match the underlying data after optimization.

## Technical Notes

Likely areas to investigate (repo-relative):

[...]

Issue #300, opened 2026-04-11
Configurable database backend: SQLAlchemy Core abstraction (SQLite + PostgreSQL)


vybe:
## Summary

Replace the raw `sqlite3` database layer with SQLAlchemy Core to support configurable database backends — starting with SQLite (default, zero-config) and PostgreSQL (production-grade, self-hosted or external).

## Motivation

- **Production readiness**: SQLite works for small deployments but has concurrency limitations under load (single-writer lock, no true connection pooling)
- **Future flexibility**: SQLAlchemy Core supports SQLite, PostgreSQL, MySQL, MariaDB — adding new backends later is trivial
- **Data integrity**: PostgreSQL provides native BOOLEAN, TIMESTAMPTZ, JSONB, proper constraint enforcement
- **Operational**: PostgreSQL enables proper backups, replication, monitoring with standard tooling

## Current State

| Metric | Value |
|--------|-------|
| Database | SQLite3 (synchronous, no ORM) |
| DB code | ~11,400 lines across 36 Python files in `src/backend/db/` |
| Tables | 36 tables, 84 indexes |
| Migrations | 30 idempotent migrations (hand-rolled, `PRAGMA`-based introspection) |
| Connection | Context manager in `db/connection.py` (28 lines) |
| Queries | 100% raw SQL with `?` placeholders, no ORM |
| Operations classes | 25 `*Operations` classes composed via `DatabaseManager` facade |

### SQLite-Specific Syntax Found

| Pattern | Count | PG Equivalent |
|---------|-------|---------------|
| `INTEGER PRIMARY KEY AUTOINCREMENT` | 6 tables | `SERIAL PRIMARY KEY` |
| `?` param placeholders | Hundreds | `%s` (psycopg2) |
| `INSERT OR REPLACE` | 4 places | `INSERT ... ON CONFLICT DO UPDATE` |
| `INSERT OR IGNORE` | 4 places | `INSERT ... ON CONFLICT DO NOTHING` |
| `PRAGMA table_info()` | 30 migrations | `information_schema.columns` |
| `sqlite3.IntegrityError` | 8 places | `sqlalchemy.exc.IntegrityError` |
| `sqlite3.Row` row factory | 1 place | `RowMapping` |
| `INTEGER` for booleans | Many columns | `Boolean` |
| `TEXT` for timestamps | Many columns | `DateTime` |
| `TEXT` for JSON | ~15 columns | `JSON` |

[...]

Issue #16, opened 2026-02-24
bug: Test issue - Dashboard dict attribute error


vybe:
## Summary

This is a **test bug report** to verify the bug reporting workflow. The dashboard fails with a dict attribute error after a recent change. This test issue can be closed immediately.

## Component

Backend / Dashboard API

## Severity

Low (test issue)

## Error

```
AttributeError: 'dict' object has no attribute 'model_dump'
```

## Location

- **File**: `src/backend/routers/dashboard.py`
- **Line**: ~45
- **Function/Class**: `get_dashboard_data()`

## Root Cause

The code expects a Pydantic model but receives a plain dict, causing the `model_dump()` call to fail.

## Reproduction Steps

1. Start Trinity backend
2. Navigate to an agent's dashboard
3. Observe the error in backend logs

## Suggested Fix

Check if the object is a dict before calling `model_dump()`:

```python
# Handle both dict and Pydantic model
if isinstance(data, dict):
    return data
return data.model_dump()
```

## Environment

- Trinity version: `1efc36b`
- Docker version: 24.0+
- OS: Ubuntu 22.04

## Related

- This is a test issue to verify the bug report workflow
- Can be closed immediately

---
*This is a test issue created to verify the /bug-report skill workflow.*

PR #1269, opened 2026-06-19
fix(perf): kill Dashboard/timeline N+1 fan-out at fleet scale (#1265)

dolho:
## Summary

Fixes the 20s+ Dashboard/timeline metric load on 10+ agent fleets (#1265). Profiling showed the bottleneck was **not** the endpoints named in the issue but (a) blocking calls serializing the event loop on `context-stats`, and (b) several per-agent / per-schedule N+1 query patterns that grow with fleet size.

## Root cause & fixes

| Path | Problem | Fix |
|------|---------|-----|
| `context-stats` (`stats.py`) | `_fetch_single_agent_context` was `async` but ran a **blocking** Docker container lookup + per-agent DB query on the event-loop thread → `asyncio.gather` was effectively serial (10 × ~1-2s = the 20s) | Coroutine now does only the async HTTP call; container host derived as `agent-{name}` (Docker DNS, no Docker call); latest activity fetched once via new bulk `get_latest_activity_for_agents` window query |
| `/api/agents/slots` | per-agent Redis `ZCARD` loop | one pipelined round-trip (mirrors `heartbeat_status_bulk`) |
| `/api/ops/schedules` | per-schedule `get_schedule_executions` loop | one `get_latest_execution_per_schedule` window query |
| `/api/activities/timeline` | fleet-wide `ORDER BY created_at DESC` with no `created_at` index; `limit*2` over-fetch + Python-side access filter | added `idx_activities_created` (schema.py + migration); access filter pushed into SQL (`agent_name IN ...`); exact `limit` |
| `db/monitoring.py`, `stats.py` | Invariant #16 (`datetime.utcnow()` vs ISO-Z columns) | `iso_cutoff()` |

No new cache layer → no freshness/invalidation risk (AC #5).

## Testing

- New `tests/unit/test_1265_dashboard_perf.py` (9 tests) covering the three bulk methods + timeline access filter — runs on SQLite and Postgres via the #300 harness.
- Window-function constructs verified on SQLite; full affected unit suites (schedules, monitoring, migrations, slots, stats-cache, sync-health) pass: 125 passed.

## Remaining AC

- [ ] Documented before/after per-endpoint timings at 1/5/10/20 agents (needs a seeded multi-agent fleet — will add to this PR).

Related to #1265

### DIFF
--- a/src/backend/db/activities.py
+++ b/src/backend/db/activities.py
@@ -13,9 +13,10 @@ import uuid
 from datetime import datetime, timedelta, timezone
 from typing import Optional, List, Dict
 
-from sqlalchemy import select, insert, update, and_, func
+from sqlalchemy import select, insert, update, and_
 
 from .engine import get_engine
+from .query_helpers import latest_per_group
 from .tables import agent_activities
 from models import ActivityState
 from utils.helpers import utc_now_iso, to_utc_iso, parse_iso_timestamp
@@ -63,6 +64,33 @@ class ActivityOperations:
             "details": json.loads(row[12]) if row[12] else None,
             "error": row[13],
             "created_at": row[14]
+        }
+
+    @staticmethod
+    def _mapping_to_activity(row) -> Dict:
+        """Convert a name-accessible mapping row to an activity dict.
+
+        Used by ``get_latest_activity_for_agents`` (which projects the curated
+        ``_ACTIVITY_COLUMNS`` via the shared window helper). Name-based access —
+        unlike the positional ``_row_to_activity`` — so reordering a column in
+        ``_ACTIVITY_COLUMNS`` can never silently misalign the fields.
+        """
+        return {
+            "id": row["id"],
+            "agent_name": row["agent_name"],
+            "activity_type": row["activity_type"],
+            "activity_state": row["activity_state"],
+            "parent_activity_id": row["parent_activity_id"],
+            "started_at": row["started_at"],
+            "completed_at": row["completed_at"],
+            "duration_ms": row["duration_ms"],
+            "user_id": row["user_id"],
+            "triggered_by": row["triggered_by"],
+            "related_chat_message_id": row["related_chat_message_id"],
+            "related_execution_id": row["related_execution_id"],
+            "details": json.loads(row["details"]) if row["details"] else None,
+            "error": row["error"],
+            "created_at": row["created_at"],
         }
 
     def create_activity(self, activity: 'ActivityCreate') -> str:
@@ -169,34 +197,20 @@ class ActivityOperations:
 
         #1265: replaces a per-agent ``get_agent_activities(name, limit=1)``
         fan-out (one query per running agent) on the Dashboard context-stats
-        path. Uses a ``ROW_NUMBER() OVER (PARTITION BY agent_name ORDER BY
-        created_at DESC)`` window served by ``idx_activities_agent``.
+        path. Uses the shared ``latest_per_group`` window helper (partition by
+        agent_name, order by created_at DESC) — served by ``idx_activities_agent``.
 
         Returns ``{agent_name: activity_dict}``; agents with no activity are
         simply absent from the map.
         """
-        if not agent_names:
-            return {}
-
-        rn = func.row_number().over(
-            partition_by=agent_activities.c.agent_name,
-            order_by=agent_activities.c.created_at.desc(),
-        ).label("rn")
-        subq = (
-            select(*_ACTIVITY_COLUMNS, rn)
-            .where(agent_activities.c.agent_name.in_(list(agent_names)))
-            .subquery()
+        rows = latest_per_group(
+            _ACTIVITY_COLUMNS,
+            agent_activities.c.agent_name,   # partition
+            agent_activities.c.created_at,   # order (DESC)
+            agent_activities.c.agent_name,   # filter IN
+            agent_names,
         )
-        # Select only the activity columns (drop the rn helper) so positional
-        # row access in _row_to_activity stays aligned [0]..[14].
-        stmt = select(*subq.c[: len(_ACTIVITY_COLUMNS)]).where(subq.c.rn == 1)
-
-        result: Dict[str, Dict] = {}
-        with get_engine().connect() as conn:
-            for row in conn.execute(stmt).all():
-                activity = self._row_to_activity(row)
-                result[activity["agent_name"]] = activity
-        return result
+        return {row["agent_name"]: self._mapping_to_activity(row) for row in rows}
 
     def get_activities_in_range(self, start_time: Optional[str] = None,
                                 end_time: Optional[str] = None,
--- a/src/backend/db/activities.py
+++ b/src/backend/db/activities.py
@@ -13,7 +13,7 @@ import uuid
 from datetime import datetime, timedelta, timezone
 from typing import Optional, List, Dict
 
-from sqlalchemy import select, insert, update, and_
+from sqlalchemy import select, insert, update, and_, func
 
 from .engine import get_engine
 from .tables import agent_activities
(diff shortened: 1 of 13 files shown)


src/backend/db/migrations.py
```diff
@@ -2517,4 +2532,5 @@ def _migrate_operator_queue_cleared_at(cursor, conn):
     ("agent_ownership_circuit_breaker", _migrate_agent_ownership_circuit_breaker),
     ("voip_tables", _migrate_voip_tables),
     ("operator_queue_cleared_at", _migrate_operator_queue_cleared_at),
+    ("activities_created_index", _migrate_activities_created_index),
```
dolho, line 2535: **[high] Missing the PostgreSQL track for this schema change.** This adds `idx_activities_created` to `db/schema.py` + a SQLite migration, but there's no matching Alembic revision under `src/backend/migrations/versions/` (only `0001_baseline.py` exists). CLAUDE.md Rule #9 (#1183): *"Every schema change requires TWO migrations"* — SQLite **and** an Alembic revision.

Impact: an existing Postgres deployment is stamped at `0001_baseline` and runs `upgrade_to_head()` on boot; with no `0002` revision, `idx_activities_created` is never created, so the fleet-wide `ORDER BY created_at DESC` timeline query keeps doing a scan+sort on Postgres — the exact #1265 regression this PR fixes for SQLite. Add an Alembic revision creating the index (and confirm `0001_baseline` autogenerates it for fresh PG builds — it must be in `tables.py` MetaData, not just `schema.py`).

src/backend/db/schedules.py
```diff
@@ -1390,6 +1390,37 @@ def get_schedule_executions(self, schedule_id: str, limit: int = 50) -> List[Sch
...
+        """
+        if not schedule_ids:
+            return {}
+
+        rn = func.row_number().over(
+            partition_by=schedule_executions.c.schedule_id,
+            order_by=schedule_executions.c.started_at.desc(),
+        ).label("rn")
+        subq = (
+            select(schedule_executions, rn)
```
dolho: **[med] `select(schedule_executions, rn)` pulls every column, including the large TEXT fields** (`response`, `tool_calls`, `execution_log` — 100KB+ per row, `message`, `error`). The only caller (`routers/ops.py` `last_execution`) reads just `id`, `status`, `started_at`, `completed_at`, `duration_ms`, `error`.

This is the anti-pattern `get_agent_executions_summary` was written to avoid (PERF-001: project only needed columns). On a fleet with hundreds of schedules the single bulk query now streams tens of MB and deserializes full `ScheduleExecution` objects per `/ops/schedules` poll — partially eating the N+1 win it replaces. Project the 6 needed columns explicitly.

src/backend/routers/activities.py
```diff
@@ -30,26 +30,21 @@ async def get_activity_timeline(
     if activity_types:
         types_list = [t.strip() for t in activity_types.split(',')]
 
-    # Get all activities
-    all_activities = db.get_activities_in_range(
+    # #1265: push the per-user access filter into SQL (None = admin, no filter)
+    # so we fetch exactly `limit` rows instead of over-fetching limit*2 and
+    # filtering in Python.
+    from services.agent_service.helpers import accessible_agent_names
+    allowed = accessible_agent_names(current_user)
```
dolho, line 44: **[med] Admin timeline scope silently widened.** Old code filtered the timeline against `get_accessible_agents()` (Docker-present agents only); the new admin path resolves to `agent_names=None` → no SQL filter. Activities of **deleted** agents (rows persist in `agent_activities`, container gone, so absent from `list_all_agents_fast()`) were previously excluded for admins and now appear.

Likely acceptable (admins seeing full audit history), but it changes what the admin dashboard shows. If intentional, worth a one-line note; if not, scope admins to known agents too.

src/backend/services/agent_service/stats.py
```diff
@@ -147,65 +149,57 @@ def invalidate_agent_stats_cache(agent_name: str) -> None:
...
-        container = get_agent_container(agent_name)
-        if container:
-            agent_url = f"http://{container.name}:8000/api/chat/session"
-            response = await client.get(agent_url)
-            if response.status_code == 200:
-                session_data = response.json()
-                stats["contextPercent"] = session_data.get("context_percent", 0)
-                stats["contextUsed"] = session_data.get("context_tokens", 0)
-                stats["contextMax"] = session_data.get("context_window", 200000)
+        agent_url = f"http://agent-{agent_name}:8000/api/chat/session"
```
dolho, line 185: **[low] Dropped the `if container:` existence check.** The old code called `get_agent_container(name)` and skipped the HTTP call when it returned `None`. Now every agent whose cached status is `"running"` gets an unconditional `client.get("http://agent-{name}:8000/...")`.

For a stale-`running` agent (container died/removed between the status snapshot and this call) each request now eats the full 2s httpx timeout before the `except` falls back to defaults, instead of skipping instantly — added latency + connection churn on exactly the unhealthy-fleet case the Dashboard matters most. `gather` bounds the total at ~2s, but it's a guaranteed 2s tail per dead host. (Also: the host string is now a hardcoded copy of the `agent-{name}:8000` convention — consider a shared `agent_base_url()` helper.)

src/backend/db/activities.py
```diff
@@ -164,13 +164,52 @@ def get_agent_activities(self, agent_name: str, activity_type: Optional[str] = N
         with get_engine().connect() as conn:
             return [self._row_to_activity(row) for row in conn.execute(stmt).all()]
 
+    def get_latest_activity_for_agents(self, agent_names: List[str]) -> Dict[str, Dict]:
```
dolho, line 195: **[low/cleanup] Near-duplicate of `ScheduleOperations.get_latest_execution_per_schedule`.** Both are "latest row per group" via `ROW_NUMBER() OVER (PARTITION BY <key> ORDER BY <ts> DESC) WHERE rn=1` → fold into `{key: row}`, differing only in table/partition/order/mapper. Consider a shared `_latest_per_group(...)` helper so cross-backend window semantics (SQLite vs Postgres tie-breaking / NULL ordering) live in one place instead of drifting across two copies.

src/backend/db/activities.py
```diff
@@ -164,13 +164,52 @@ def get_agent_activities(self, agent_name: str, activity_type: Optional[str] = N
...
+            order_by=agent_activities.c.created_at.desc(),
+        ).label("rn")
+        subq = (
+            select(*_ACTIVITY_COLUMNS, rn)
+            .where(agent_activities.c.agent_name.in_(list(agent_names)))
+            .subquery()
+        )
+        # Select only the activity columns (drop the rn helper) so positional
+        # row access in _row_to_activity stays aligned [0]..[14].
+        stmt = select(*subq.c[: len(_ACTIVITY_COLUMNS)]).where(subq.c.rn == 1)
```
dolho: **[low] Fragile positional projection.** `select(*subq.c[: len(_ACTIVITY_COLUMNS)])` exists only because `_row_to_activity` reads by integer position `row[0]..row[14]`. If anyone reorders/inserts/removes a column in `_ACTIVITY_COLUMNS` without updating the indices, the slice still "works" but maps wrong columns — silent data corruption, no exception. The schedules sibling avoids this by using `.mappings()` + name-based access; switching this method to the same would let it select by name and delete the slice.

dolho:
## Code review (high-effort, recall-biased) — 6 findings

Reviewed the diff across 8 finder angles + verification. No crash-class bugs; the headline perf change is correct (event-loop de-blocking, bulk window queries, pipelined slots, SQL access filter, `iso_cutoff`). Findings, ranked:

| # | Sev | File | Issue |
|---|-----|------|-------|
| 1 | **high** | `db/migrations.py` | **Missing the Postgres Alembic revision** for `idx_activities_created` — CLAUDE.md Rule #9 requires both tracks. Existing PG deployments never get the index → the #1265 timeline regression persists on Postgres. |
| 2 | med | `db/schedules.py` | `get_latest_execution_per_schedule` does `select(schedule_executions)` (SELECT *), streaming `response`/`execution_log` (100KB+) when ops.py uses 6 small fields — partially eats the N+1 win (PERF-001 anti-pattern). |
| 3 | med | `routers/activities.py` | Admin timeline scope widened: `agent_names=None` drops the old Docker-presence filter, so admins now see activities of deleted agents. Likely fine, but a behavior change worth confirming. |
| 4 | low | `services/agent_service/stats.py` | Dropped `if container:` guard → a stale-`running` agent now eats a full 2s httpx timeout instead of skipping. |
| 5 | low | `db/activities.py` | Near-duplicate ROW_NUMBER window method vs the schedules sibling — extract a shared `_latest_per_group` helper. |
| 6 | low | `db/activities.py` | Fragile positional projection (`subq.c[:len(...)]` + index-based `_row_to_activity`); a column reorder silently corrupts. Prefer `.mappings()` like the schedules sibling. |

**#1 is the actionable blocker** — the fix doesn't reach Postgres without an Alembic revision. Details inline.

dolho:
## Review fixes applied — `40f42016`

All six findings addressed:

1. **[high] PG migration** — added `migrations/versions/0002_activities_created_index.py` (`0002 → 0001_baseline`). Existing Postgres deployments now get `idx_activities_created` on `upgrade head`; fresh builds already got it via `0001_baseline` iterating `schema.py:INDEXES`. Alembic graph validated (single head → 0001 → base).
2. **[med] Slim projection** — `get_latest_execution_per_schedule` projects only the 6 dashboard fields (no `response`/`execution_log`/`tool_calls`/`message`). Returns slim dicts with ISO-normalised timestamps (API output unchanged); `ops.py` updated.
3. **[med] Admin scope** — documented that the unfiltered admin timeline (full audit history incl. deleted agents) is intentional.
4. **[low] Connect timeout** — capped at `0.5s` (`httpx.Timeout(2.0, connect=0.5)`) so a stale-`running` agent fails fast, without reintroducing the per-agent Docker lookup that was the bottleneck.
5. **[low] Shared helper** — extracted `db/query_helpers.py:latest_per_group`; both bulk methods reuse it.
6. **[low] Positional fragility** — `get_latest_activity_for_agents` now maps by name (`_mapping_to_activity`); the `subq.c[:len(...)]` slice is gone.

Tests updated (slim-dict shape + narrow-projection assertion): 9 pass; 93 pass across affected suites.

vybe (approved):
Validated via /validate-pr: P1 perf fix (#1265). Dual-track migration verified (schema.py + SQLite migrations.py + Alembic 0002 chained to 0001_baseline, all consistent); SQL parameterized, Inv #16 iso_cutoff fixed; 9 new tests + 125 pass on SQLite+PG. Remaining before/after-timing AC is non-blocking evidence. Approving.

Issue #347, opened 2026-04-15
feat: Per-agent GitHub PAT configuration


vybe:
## Summary

Allow configuring a GitHub Personal Access Token (PAT) per agent, with fallback to the global system PAT if not set. This enables agents to use different GitHub accounts or tokens with different scopes/permissions.

## Current Behavior

- Single global GitHub PAT stored in `system_settings` table
- All agents share the same PAT for git operations (clone, push, pull)
- PAT retrieved via `SettingsService.get_github_pat()` with env var fallback

## Proposed Behavior

1. Each agent can optionally have its own GitHub PAT configured
2. If per-agent PAT is set → use it for that agent's git operations
3. If not set → fall back to global system PAT (current behavior)
4. Configurable via UI, API, and MCP tools

### 1. Database Schema
Add `github_pat` column to `agent_git_config` table:
```sql
ALTER TABLE agent_git_config ADD COLUMN github_pat TEXT;
```

### 2. DB Operations
Add methods following the ResourcesMixin pattern:
- `get_agent_github_pat(agent_name) -> Optional[str]`
- `set_agent_github_pat(agent_name, pat) -> bool`
- `clear_agent_github_pat(agent_name) -> bool`

### 3. Service Layer
Create helper function:
```python
def get_github_pat_for_agent(agent_name: str) -> str:
    """Get per-agent PAT if set, otherwise fall back to global."""
    agent_pat = db.get_agent_github_pat(agent_name)
    return agent_pat if agent_pat else get_github_pat()
```

Update callers:
- `routers/git.py:initialize_github_sync()` 
- `services/git_service.py` (already accepts `github_pat` param)

### 4. Backend API Endpoints
Add to `routers/agent_config.py`:
```
GET    /api/agents/{name}/github-pat  → { configured: bool, source: "agent"|"global" }
PUT    /api/agents/{name}/github-pat  → Set agent PAT
DELETE /api/agents/{name}/github-pat  → Clear agent PAT (revert to global)
```

### 5. MCP Tools
Add to `src/mcp-server/src/tools/agents.ts`:
- `set_agent_github_pat` - Set/clear GitHub PAT for an agent
- `get_agent_github_pat_status` - Check if agent has custom PAT configured

### 6. Frontend UI
Add to GitPanel.vue or Settings section in AgentDetail.vue:
- Password input field (masked)
- Status indicator showing "Using: Agent Token" or "Using: Global Token"
- Configure/Clear buttons

## Security Considerations

[...]

Issue #211, opened 2026-03-27
feat: Auto-propagate GitHub PAT to all agents on update


vybe:
## Problem

When an admin updates the GitHub PAT in Settings, running agents continue using the old PAT. The admin must manually restart each agent or call a propagation endpoint.

## Solution

Make PAT propagation the **default behavior** when the PAT is updated. In the existing `update_github_pat` endpoint (`settings.py:268`), after saving the new PAT to the database, automatically propagate it to all active agents.

### Implementation

```python
@router.put("/settings/github-pat")
async def update_github_pat(...):
    # ... existing validation and save logic ...
    
    # Auto-propagate to all active agents
    propagation_result = await propagate_github_pat_to_agents(new_pat)
    
    return {
        "status": "updated",
        "propagation": propagation_result
    }
```

### Frontend

Update the Settings page to show propagation results after PAT update:
- "PAT updated and applied to 5 agents"
- Or warning: "PAT updated but failed to apply to: agent-x, agent-y"

## Dependencies

- Depends on the propagation endpoint/service being implemented first (#TBD)

## Files

- `src/backend/routers/settings.py` (`update_github_pat` endpoint, line 268)
- `src/frontend/src/views/Settings.vue` (show propagation result)

## Acceptance Criteria

- [ ] Updating GitHub PAT automatically propagates to all active agents with `GITHUB_PAT`
- [ ] Response includes per-agent propagation status
- [ ] Frontend shows propagation results (success count, any failures)
- [ ] Partial failures don't block the PAT update itself
- [ ] Delete PAT does NOT propagate (agents keep working until next restart)

PR #1271, opened 2026-06-19
fix(git): per-agent GitHub PAT now propagates to existing containers (#1264)

### LINKED ISSUE #1264: bug: per-agent GitHub PAT (#347) never propagates to an existing container — git remote keeps an empty password, all pushes fail
## Summary

A per-agent GitHub PAT (#347) that is configured **after** an agent's container was created (or after its git was initialized) never reaches the running container. The token sits valid in the DB, but the container's git remote stays templated with an **empty password** (`https://x-access-token:@github.com/...`), so every `git fetch`/`git push` fails with "Invalid username or token." `set_agent_github_pat` reports success and tells the operator to "restart the agent," but restart does **not** fix it — and the global-PAT propagation service (#211) explicitly skips per-agent-PAT agents. The result is a silent, permanent 100% git-sync failure: an affected agent can accumulate **100+ unpushed commits** with `last_sync_status: "never"` while the platform reports `consecutive_failures: 0`.

## Component

Backend / Agent Runtime — GitHub PAT injection & propagation (`agent_service`, `routers/git.py`, `github_pat_propagation_service`)

## Priority

P1 — per-agent git push is completely non-functional for affected agents, no in-product workaround (the suggested "restart" is a no-op). Work is recoverable from the workspace volume, so not P0.

## Error

Observed inside the agent container (`git ls-remote` / any push):

```
remote: Invalid username or token. Password authentication is not supported for Git operations.
fatal: Authentication failed for 'https://github.com/<org>/<repo>.git/'
```

The remote is templated with an empty password:

```
origin  https://x-access-token:@github.com/<org>/<repo>.git
```

Container `GITHUB_PAT` env var is **unset**, and the agent's `~/.env` has **no `GITHUB_PAT` line** — while the per-agent PAT decrypted from the DB is valid and has `push=true` on the target repo (verified against the GitHub API, HTTP 200).

## Root Cause

There is no path that injects a per-agent PAT into a container that did not already have `GITHUB_PAT` set at creation time.

1. **Creation injects only when the PAT already exists** — `services/agent_service/crud.py:541`:
   ```python
   if github_repo_for_agent and github_pat_for_agent:
       env_vars['GITHUB_REPO'] = github_repo_for_agent
       env_vars['GITHUB_PAT'] = github_pat_for_agent
       env_vars['GIT_SYNC_ENABLED'] = 'true'
   ```
   If the per-agent PAT is configured (or git initialized) *after* the container was created, `GITHUB_PAT` is never added to the container env.

2. **Restart refresh is guarded on GITHUB_PAT already being present** — `services/agent_service/lifecycle.py:384`:
   ```python
   # Update GITHUB_PAT using per-agent PAT first, then platform PAT
   if env_vars.get('GITHUB_PAT'):          # <-- guard: only refreshes an existing value
       from routers.git import get_github_pat_for_agent
       current_pat = get_github_pat_for_agent(agent_name)
       if current_pat:
           env_vars['GITHUB_PAT'] = current_pat
   ```
   A container created without `GITHUB_PAT` is skipped here on every restart — so the now-configured per-agent PAT is never picked up. (`agent_service/helpers.py:458` mirrors this: "Agent doesn't use GITHUB_PAT — no update needed.")

3. **The remote URL is frozen empty** — `routers/git.py:351` builds the remote from `get_github_pat_for_agent(agent_name)` at init. If that resolved empty at init time, the remote is permanently `https://x-access-token:@.../` in the persisted workspace volume; nothing re-templates it later.

4. **`set_agent_github_pat` does not propagate** — `routers/git.py:542+` validates and stores the encrypted PAT, then returns `note: "Restart agent for new PAT to be used in git operations"`. But per (2) restart is a no-op for a tokenless container, so the advice is misleading.

5. **Global propagation (#211) deliberately skips these agents** — `services/github_pat_propagation_service.py` only patches the *global* PAT into agents whose `.env` already has a `GITHUB_PAT=` line, and explicitly skips agents that have a per-agent PAT (`skipped_per_agent_pat`) and agents whose `.env` lacks the line (`skipped_no_pat`). So neither the per-agent nor the global PAT path can repair the credential.

## Reproduction Steps

1. Create a GitHub-native agent **without** a per-agent PAT and without a global platform PAT configured (container is created with no `GITHUB_PAT`), OR initialize git for the agent while no PAT resolves.
2. Later, configure a valid per-agent PAT: `PUT /api/agents/<agent>/github-pat` (returns 200, "restart agent…").
3. Restart the agent.
4. Inside the container: `git -C /home/developer remote -v` → remote shows `https://x-access-token:@github.com/<org>/<repo>.git`; `printenv GITHUB_PAT` → unset.
5. Any `git push` / `git ls-remote` fails: `remote: Invalid username or token`.
6. Platform `GET /api/agents/<agent>/git/sync-state` shows `last_sync_status: "never"`, `consecutive_failures: 0` while commits pile up (`ahead` keeps growing) — the failure is invisible platform-side because the agent's own git ops fail, not Trinity's `/git/sync` path.

## Suggested Fix

Make a per-agent PAT take effect on an existing, running container:

- In `set_agent_github_pat` (`routers/git.py`), after storing the PAT, **propagate it to the running container** — write `GITHUB_PAT` into the agent's `.env` (reuse the `/api/credentials/inject` mechanism the propagation service already uses) **and** rewrite `remote.origin.url` to embed the resolved token (or switch to a tokenless remote + credential helper that reads `GITHUB_PAT`).
- Remove the `if env_vars.get('GITHUB_PAT')` precondition in `lifecycle.py:384` so a newly-configured per-agent PAT is injected on restart even when the container had no token at creation:
  ```python
  # Always resolve the effective PAT; inject if git sync is enabled for the agent
  from routers.git import get_github_pat_for_agent
  current_pat = get_github_pat_for_agent(agent_name)
  if current_pat and db.get_git_config(agent_name):
      env_vars['GITHUB_PAT'] = current_pat
  ```
- Have the agent base-image startup **re-template `remote.origin.url` from `GITHUB_PAT` on every start** (idempotent), so an empty-password remote frozen in the workspace volume self-heals once the env var is present.
- Either extend `github_pat_propagation_service` to also cover per-agent-PAT agents, or make the per-agent set endpoint do its own single-agent propagation (preferred — keeps concerns separated).

Until fixed, the operator workaround is to set the per-agent PAT and then **recreate** the container (not just restart) so `crud.py:541` re-injects the now-present PAT — and verify the remote URL gets re-templated.

## Environment

- Trinity version: `cccac44d` (VERSION `0.6.0`)
- Docker version: 28.3.3
- OS: Ubuntu (Linux), prod compose, non-root images

## Related

- `services/agent_service/crud.py:541` (creation injection)
- `services/agent_service/lifecycle.py:384` (restart refresh guard)
- `services/agent_service/helpers.py:447-461` (PAT-match check)
- `routers/git.py:351` (remote URL templating), `routers/git.py:484` (`get_github_pat_for_agent`), `routers/git.py:542+` (`set_agent_github_pat`)
- `services/github_pat_propagation_service.py` (global-PAT propagation that skips per-agent-PAT agents)
- Builds on #347 (per-agent PAT), #211/#210 (global PAT propagation)

dolho:
## Summary

Fixes #1264 — a per-agent GitHub PAT (#347) configured **after** an agent's container/git was set up never reached the container. The remote stayed templated with an empty password (`https://x-access-token:@github.com/...`), every `fetch`/`push` failed, and the endpoint's "restart the agent" advice was a no-op. Affected agents silently accumulated 100+ unpushed commits with `last_sync_status: "never"`.

## Key insight

`startup.sh` **already** re-templates `origin` from `GITHUB_PAT` on every start (`git remote set-url`). So the whole bug reduces to one thing: **get the PAT into the container env**. Three paths failed to do that — this PR fixes all three.

## Changes

| Path | Before | After |
|------|--------|-------|
| `set_agent_github_pat` (`routers/git.py`) | stored PAT, returned "restart agent" (no-op) | **live propagation, no restart**: `propagate_pat_to_single_agent` injects `GITHUB_PAT` into the running container's `.env` (adding the line when absent — the tokenless case the global-PAT path skips) + re-templates the live remote; note reflects the real outcome |
| lifecycle recreate (`lifecycle.py:385`) | `if env_vars.get('GITHUB_PAT')` — skipped tokenless containers forever | gate on `db.get_git_config(agent_name)` so a later-configured PAT is injected on restart; startup.sh then fixes the remote |
| `check_github_pat_env_matches` (`helpers.py`) | tokenless container → "no update needed" | tokenless **+ effective PAT + git configured** → mismatch (recreate re-injects) |

New `git_service.update_remote_pat` reuses the proven init `set-url`/`add` pattern + `_git_remote_url`; best-effort (stopped / no-git agents fall back to the restart path, now fixed).

## Testing

- `tests/unit/test_1264_per_agent_pat_propagation.py` (8): `.env` add-if-missing, single-agent propagation (stopped short-circuit + running inject/re-template with mocked agent HTTP), and the recreate-decision matrix.
- 45 pass across the PAT suites including the `sys.modules`-mocking siblings (`test_start_agent_skip_inject`, `test_inject_assigned_credentials`) in CI collection order. The lifecycle static-callsite check (`test_github_pat_fallback`) still passes.

**Not covered by unit tests** (flagged for reviewers): the end-to-end against a real GitHub repo/PAT (live `.env` inject + in-container `git remote set-url` → successful push). The container I/O reuses the already-proven `initialize_github_sync` primitives and the existing `startup.sh` self-heal, but a manual verification on a real GitHub-native agent is worthwhile before release.

Related to #1264

### DIFF
--- a/src/backend/routers/git.py
+++ b/src/backend/routers/git.py
@@ -591,12 +591,28 @@ async def set_agent_github_pat(
     if not success:
         raise HTTPException(status_code=500, detail="Failed to save PAT")
 
+    # #1264: propagate to the running container so the PAT takes effect WITHOUT a
+    # restart — inject GITHUB_PAT into .env (adding the line if the container was
+    # created tokenless) and re-template the live git remote. Best-effort: a
+    # stopped agent picks it up on next start (relaxed lifecycle injection +
+    # startup self-heal).
+    from services.github_pat_propagation_service import propagate_pat_to_single_agent
+    propagation = await propagate_pat_to_single_agent(agent_name, pat)
+
+    if propagation.get("applied"):
+        note = "PAT applied to the running agent — git operations will use it immediately."
+    elif propagation.get("reason") == "agent_not_running":
+        note = "PAT saved. Start the agent for it to take effect in git operations."
+    else:
+        note = "PAT saved, but live propagation did not complete; restart the agent to apply it."
+
     return {
         "message": "GitHub PAT configured successfully",
         "agent_name": agent_name,
         "github_username": username,
         "source": "agent",
-        "note": "Restart agent for new PAT to be used in git operations"
+        "propagation": propagation,
+        "note": note,
     }
 
 
--- a/src/backend/routers/git.py
+++ b/src/backend/routers/git.py
@@ -599,8 +599,13 @@ async def set_agent_github_pat(
     from services.github_pat_propagation_service import propagate_pat_to_single_agent
     propagation = await propagate_pat_to_single_agent(agent_name, pat)
 
-    if propagation.get("applied"):
+    # #1264 review: the live git process authenticates from the remote URL, so
+    # only `remote_updated` means git ops work *immediately*. An env-only update
+    # (or a stopped agent) takes effect on the next restart.
+    if propagation.get("remote_updated"):
         note = "PAT applied to the running agent — git operations will use it immediately."
+    elif propagation.get("env_updated"):
+        note = "PAT saved to the agent; it will take effect on the next restart."
     elif propagation.get("reason") == "agent_not_running":
         note = "PAT saved. Start the agent for it to take effect in git operations."
     else:
--- a/src/backend/services/agent_service/helpers.py
+++ b/src/backend/services/agent_service/helpers.py
@@ -442,6 +442,24 @@ def check_guardrails_env_matches(container, agent_name: str) -> bool:
         return False
 
 
+def needs_per_agent_pat_injection(agent_name: str) -> bool:
+    """#1264: True when the agent has a **per-agent** GitHub PAT configured AND
+    git sync — i.e. the container SHOULD carry that token.
+
+    Gated on ``db.get_agent_github_pat`` (the per-agent PAT only), NOT
+    ``get_github_pat_for_agent`` (which falls back to the global platform PAT).
+    Using the global fallback here would force-recreate / inject into every
+    tokenless git agent the moment an operator sets a global PAT — that's #211's
+    deliberate opt-in `.env` path, not this one's job, and is out of #1264 scope.
+
+    Shared by the recreate matcher (:func:`check_github_pat_env_matches`) and the
+    lifecycle recreate inject gate so the two predicates cannot drift — if they
+    did, the matcher could keep requesting a recreate the inject never satisfies
+    (infinite recreate loop).
+    """
+    return bool(db.get_agent_github_pat(agent_name)) and bool(db.get_git_config(agent_name))
+
+
 def check_github_pat_env_matches(container, agent_name: str) -> bool:
     """
     Check if container's GITHUB_PAT env var matches the effective PAT for this agent.
@@ -453,19 +471,18 @@ def check_github_pat_env_matches(container, agent_name: str) -> bool:
     env_list = container.attrs.get("Config", {}).get("Env", [])
     env_dict = {e.split("=", 1)[0]: e.split("=", 1)[1] for e in env_list if "=" in e}
 
-    current_pat = get_github_pat_for_agent(agent_name)
-
     container_pat = env_dict.get("GITHUB_PAT")
     if not container_pat:
-        # #1264: a tokenless container that now has an effective PAT AND git sync
-        # configured genuinely mismatches desired state — recreation re-injects
-        # the token (crud.py) so the remote can authenticate. Without this, a
-        # per-agent PAT set after a tokenless creation never reached the env.
-        if current_pat and db.get_git_config(agent_name):
-            return False
-        # Otherwise the agent doesn't use GITHUB_PAT — no update needed.
-        return True
+        # #1264: a tokenless container needs a recreate ONLY when a per-agent PAT
+        # is configured (so creation/recreate will inject it). A global-only PAT
+        # must NOT force tokenless agents to recreate — see
+        # needs_per_agent_pat_injection. Recreate converges in one pass because
+        # the lifecycle inject gate uses the SAME predicate.
+        return not needs_per_agent_pat_injection(agent_name)
 
+    # Container already has a token — keep it in sync with the effective PAT
+    # (per-agent first, then the global fallback for already-tokened containers).
+    current_pat = get_github_pat_for_agent(agent_name)
     if not current_pat:
         # No PAT configured anywhere — can't update, leave as-is
         return True
--- a/src/backend/services/agent_service/helpers.py
+++ b/src/backend/services/agent_service/helpers.py
@@ -453,12 +453,19 @@ def check_github_pat_env_matches(container, agent_name: str) -> bool:
     env_list = container.attrs.get("Config", {}).get("Env", [])
     env_dict = {e.split("=", 1)[0]: e.split("=", 1)[1] for e in env_list if "=" in e}
 
+    current_pat = get_github_pat_for_agent(agent_name)
+
     container_pat = env_dict.get("GITHUB_PAT")
     if not container_pat:
-        # Agent doesn't use GITHUB_PAT — no update needed
+        # #1264: a tokenless container that now has an effective PAT AND git sync
+        # configured genuinely mismatches desired state — recreation re-injects
+        # the token (crud.py) so the remote can authenticate. Without this, a
+        # per-agent PAT set after a tokenless creation never reached the env.
+        if current_pat and db.get_git_config(agent_name):
+            return False
+        # Otherwise the agent doesn't use GITHUB_PAT — no update needed.
         return True
 
-    current_pat = get_github_pat_for_agent(agent_name)
     if not current_pat:
         # No PAT configured anywhere — can't update, leave as-is
         return True
--- a/src/backend/services/agent_service/lifecycle.py
+++ b/src/backend/services/agent_service/lifecycle.py
@@ -25,7 +25,7 @@ from services.docker_utils import (
 from services.agent_service.helpers import validate_base_image
 from services.settings_service import get_anthropic_api_key, get_github_pat, get_agent_full_capabilities, get_agent_default_resources
 from services.skill_service import skill_service
-from .helpers import check_shared_folder_mounts_match, check_api_key_env_matches, check_github_pat_env_matches, check_resource_limits_match, check_full_capabilities_match, check_guardrails_env_matches
+from .helpers import check_shared_folder_mounts_match, check_api_key_env_matches, check_github_pat_env_matches, check_resource_limits_match, check_full_capabilities_match, check_guardrails_env_matches, needs_per_agent_pat_injection
 from .file_sharing import check_public_folder_mount_matches
 from .read_only import inject_read_only_hooks, remove_read_only_hooks
 
@@ -382,13 +382,14 @@ async def recreate_container_with_updated_config(agent_name: str, old_container,
         env_vars.pop('CLAUDE_CODE_OAUTH_TOKEN', None)
 
     # Update GITHUB_PAT using per-agent PAT first, then platform PAT.
-    if db.get_git_config(agent_name):  # #1264: gate on git config, not GITHUB_PAT presence
+    if env_vars.get('GITHUB_PAT') or needs_per_agent_pat_injection(agent_name):
         from routers.git import get_github_pat_for_agent
         current_pat = get_github_pat_for_agent(agent_name)
         if current_pat:
-            # #1264: a tokenless container (PAT configured after creation) now
-            # receives the effective per-agent PAT on restart; startup.sh then
-            # re-templates the remote so fetch/push authenticate.
+            # #1264: refresh an existing token, or newly inject when a PER-AGENT
+            # PAT is set (needs_per_agent_pat_injection — same gate the recreate
+            # matcher uses, so they converge). A global-only PAT is NOT injected
+            # into a previously-tokenless container (that stays #211's opt-in path).
             env_vars['GITHUB_PAT'] = current_pat
 
     # GUARD-001: re-serialise guardrails overrides into env so startup.sh
--- a/src/backend/services/agent_service/lifecycle.py
+++ b/src/backend/services/agent_service/lifecycle.py
@@ -381,11 +381,14 @@ async def recreate_container_with_updated_config(agent_name: str, old_container,
         env_vars.pop('ANTHROPIC_API_KEY', None)
         env_vars.pop('CLAUDE_CODE_OAUTH_TOKEN', None)
 
-    # Update GITHUB_PAT using per-agent PAT first, then platform PAT
-    if env_vars.get('GITHUB_PAT'):
+    # Update GITHUB_PAT using per-agent PAT first, then platform PAT.
+    if db.get_git_config(agent_name):  # #1264: gate on git config, not GITHUB_PAT presence
         from routers.git import get_github_pat_for_agent
         current_pat = get_github_pat_for_agent(agent_name)
         if current_pat:
+            # #1264: a tokenless container (PAT configured after creation) now
+            # receives the effective per-agent PAT on restart; startup.sh then
+            # re-templates the remote so fetch/push authenticate.
             env_vars['GITHUB_PAT'] = current_pat
 
     # GUARD-001: re-serialise guardrails overrides into env so startup.sh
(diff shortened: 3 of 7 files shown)


src/backend/services/agent_service/helpers.py
```diff
@@ -453,12 +453,19 @@ def check_github_pat_env_matches(container, agent_name: str) -> bool:
...
+    current_pat = get_github_pat_for_agent(agent_name)
+
     container_pat = env_dict.get("GITHUB_PAT")
     if not container_pat:
-        # Agent doesn't use GITHUB_PAT — no update needed
+        # #1264: a tokenless container that now has an effective PAT AND git sync
+        # configured genuinely mismatches desired state — recreation re-injects
+        # the token (crud.py) so the remote can authenticate. Without this, a
+        # per-agent PAT set after a tokenless creation never reached the env.
+        if current_pat and db.get_git_config(agent_name):
```
dolho: **[high] Over-broad: this fires for the GLOBAL platform PAT, not just a per-agent PAT.** `get_github_pat_for_agent()` falls back to the global platform PAT when no per-agent PAT is set. So once an operator configures a global PAT, EVERY git-configured tokenless agent matches this branch → `check_github_pat_env_matches` returns False → the agent is force-recreated on its next start (one-time, converges — but disruptive downtime + churn for agents the operator never gave a per-agent token).

That contradicts #211's deliberate decision (`github_pat_propagation_service` skips agents whose `.env` lacks `GITHUB_PAT` — global PAT is never *added* to tokenless agents) and exceeds #1264's per-agent scope.

Gate the tokenless→recreate decision on the **per-agent** PAT only:
```suggestion
        if db.get_agent_github_pat(agent_name) and db.get_git_config(agent_name):
```
(keep the global-PAT *refresh* of containers that already have the token via the existing `container_pat == current_pat` path below.)

src/backend/services/agent_service/lifecycle.py
```diff
@@ -381,11 +381,14 @@ async def recreate_container_with_updated_config(agent_name: str, old_container,
         env_vars.pop('ANTHROPIC_API_KEY', None)
         env_vars.pop('CLAUDE_CODE_OAUTH_TOKEN', None)
 
-    # Update GITHUB_PAT using per-agent PAT first, then platform PAT
-    if env_vars.get('GITHUB_PAT'):
+    # Update GITHUB_PAT using per-agent PAT first, then platform PAT.
+    if db.get_git_config(agent_name):  # #1264: gate on git config, not GITHUB_PAT presence
```
dolho: **[high] Same global-PAT overreach on the recreate path.** Gating on `db.get_git_config(agent_name)` + `get_github_pat_for_agent()` (which includes the global fallback) means a previously-tokenless git-configured container now gets the **global** PAT injected into its env for the first time on recreate — a credential it never carried. #211 intentionally never does this. To stay scoped to #1264 (per-agent PAT), introduce the token only when a per-agent PAT exists, while still refreshing an already-present token:
```suggestion
    if env_vars.get('GITHUB_PAT') or db.get_agent_github_pat(agent_name):  # #1264: per-agent PAT, or refresh existing
```
Note: this gate MUST stay equivalent to the matcher at `helpers.py:464` — if they drift, the matcher can keep returning False while recreate stops injecting → infinite recreate loop. Cross-link them (or share one predicate; see the altitude comment).

src/backend/routers/git.py
```diff
@@ -591,12 +591,28 @@ async def set_agent_github_pat(
...
+    # #1264: propagate to the running container so the PAT takes effect WITHOUT a
+    # restart — inject GITHUB_PAT into .env (adding the line if the container was
+    # created tokenless) and re-template the live git remote. Best-effort: a
+    # stopped agent picks it up on next start (relaxed lifecycle injection +
+    # startup self-heal).
+    from services.github_pat_propagation_service import propagate_pat_to_single_agent
+    propagation = await propagate_pat_to_single_agent(agent_name, pat)
+
+    if propagation.get("applied"):
+        note = "PAT applied to the running agent — git operations will use it immediately."
```
dolho, line 606: **[med] "applied immediately" can be reported when the live remote was NOT fixed.** `applied = env_updated OR remote_updated` (propagation_service:104). The agent's auto-sync pushes via `git push origin HEAD` and authenticates from the **token embedded in `remote.origin.url`** — NOT from `GITHUB_PAT` env. So the docker-exec `git remote set-url` (`remote_updated`) is the load-bearing part of the *immediate* fix; the `.env` write only matters on the next restart. If the remote re-template fails (or `github_repo` is empty) but `.env` inject succeeds, this prints "git operations will use it immediately" while pushes keep 403-ing until a restart. Base the "immediately" claim on `remote_updated`, and treat env-only as the "applies on restart" case.

src/backend/services/git_service.py
```diff
@@ -170,6 +170,48 @@ def generate_working_branch(agent_name: str, instance_id: str) -> str:
...
+
+    Returns True on success. Best-effort: returns False (never raises) if the
+    container isn't running, has no git dir, or the command fails.
+    """
+    if not github_pat or not github_repo:
+        return False
+    container_name = f"agent-{agent_name}"
+    try:
+        git_dir = await _detect_git_dir(container_name)
+        url = _git_remote_url(github_pat, github_repo)
```
dolho: **[low/security] PAT interpolated unescaped into a `bash -c` string.** `url` embeds the raw PAT (`oauth2:<pat>@…`, no `urllib.parse.quote`) and is spliced into `bash -c "cd {git_dir} && git remote set-url origin {url} || …"`. Latent today — `set_agent_github_pat` validates the PAT against the GitHub API first, so it's a canonical `[A-Za-z0-9_]` token — but this is an unguarded shell sink (same pre-existing pattern as `initialize_github_sync` step 3, now extended to a new entry point). Defense-in-depth: shell-quote the command or URL-encode the credential before interpolation.

src/backend/services/github_pat_propagation_service.py
```diff
@@ -55,6 +55,58 @@ def _env_has_github_pat(env_content: str) -> bool:
...
+    re-templates the live git remote so the frozen empty-password remote
+    (``https://x-access-token:@…``) is fixed immediately and fetch/push work.
+
+    Best-effort and non-fatal: a stopped agent picks the PAT up on next start via
+    the relaxed lifecycle injection + the startup.sh self-heal. Returns a small
+    status dict for the set-PAT API response.
+    """
+    from services import git_service
+
+    running = {a.name for a in list_all_agents_fast() if a.status == "running"}
```
dolho: **[low/altitude] The "agent needs the PAT" predicate is now re-derived in three places** — here, `lifecycle.py:385` (recreate gate), and `helpers.py:464` (recreate trigger) — each combining (effective PAT + git config) slightly differently (this path also gates the remote on `github_repo`). Angle-C confirmed the recreate converges only because the matcher and lifecycle gates are currently *identical*; if they drift, infinite recreate. Extract one `agent_needs_pat_injection(agent_name)` (and a single `effective vs per-agent PAT` rule) so the three can't diverge.

dolho: **[low/cleanup] Duplicates the `.env` read/patch/inject round-trip in `_propagate_to_agent`** (same `agent-{name}:8000`, `/api/credentials/read`, `_patch_env_github_pat`, `/api/credentials/inject`, same error handling). The only real difference is add-if-missing vs the global path's skip. Collapse into a shared `_inject_pat_env(client, agent_name, pat, *, add_if_missing)` so the agent-server credentials contract lives in one place — otherwise a future endpoint change must be fixed in both, and missing this path silently reintroduces the #1264 bug.

src/backend/services/git_service.py
dolho: **[low/cleanup] The set-url/add command duplicates `initialize_github_sync` step 3** (byte-for-byte `git remote get-url origin … && set-url … || add …`, same `bash -c "cd {git_dir} && …"`). Extract `_set_origin_remote(container_name, git_dir, url)` used by both, so the remote-templating shell lives once. (Also: the docstring references `x-access-token:@` while both this code and `startup.sh` use `oauth2:` — fix the docstring so a maintainer doesn't "correct" the URL scheme and reintroduce drift.)

src/backend/services/github_pat_propagation_service.py
```diff
@@ -55,6 +55,58 @@ def _env_has_github_pat(env_content: str) -> bool:
...
+                f"{base_url}/api/credentials/inject", json={"files": {".env": patched}}
+            )
+            inject_resp.raise_for_status()
+            env_updated = True
+        except (httpx.HTTPStatusError, httpx.RequestError) as e:
+            logger.warning("single-agent PAT .env inject failed for %s: %s", agent_name, e)
+
+    # Re-template the live remote so an existing clone picks up the token now.
+    git_config = db.get_git_config(agent_name)
+    github_repo = getattr(git_config, "github_repo", None) if git_config else None
```
dolho: **[low/cleanup] Redundant second `get_git_config` fetch + `getattr` masking.** The sole caller (`set_agent_github_pat`) already fetched `git_config` one statement earlier and 400s if absent, so passing it in avoids a second DB round-trip (and a possible inconsistent read under concurrent edits). `getattr(git_config, "github_repo", None)` also silently swallows a future field rename — `github_repo` becomes None, `remote_updated` stays False, and the PAT silently stops applying to the live remote (the exact "works but doesn't propagate" class #1264 fixes). Use `git_config.github_repo` directly.

src/backend/services/github_pat_propagation_service.py
dolho: **[low/perf] Fleet-wide `list_all_agents_fast()` to check one agent's status.** This enumerates every Trinity container + builds an `AgentStatus` each, just to test whether `agent_name` is running, on an interactive endpoint. `get_agent_container(agent_name)` fetches the single container directly — `get_agent_container(agent_name)` then check `.status == "running"`.

dolho:
## Code review (high-effort, recall-biased) — 9 findings

Reviewed across 8 finder angles + cross-file tracing. The core fix is sound and **the recreate path provably converges (no loop)** — the matcher and lifecycle recreate-gate use identical conditions, so a tokenless container gets the token on its one recreate. Conventions/logging came back clean: I traced the PAT through the new `logger.warning` paths and found no leak (`git remote set-url`/`add` never echo the URL; only `fetch`/`clone` do, which this command doesn't run).

| # | Sev | File | Issue |
|---|-----|------|-------|
| 1 | **high** | `helpers.py:464` | Matcher fires for the **global** PAT (via `get_github_pat_for_agent` fallback), so setting a global PAT force-recreates *every* tokenless git agent. Gate on the per-agent PAT. |
| 2 | **high** | `lifecycle.py:385` | Same overreach on recreate — injects the **global** PAT into previously-tokenless containers (#211 deliberately never does this). Scope to per-agent PAT; keep this gate ≡ the matcher to preserve loop-convergence. |
| 3 | med | `routers/git.py:603` | "applied immediately" can be reported when only `.env` was written but the **live remote** (the thing auth actually uses for `git push`) wasn't re-templated. Base the claim on `remote_updated`. |
| 4 | low | `git_service.py:192` | PAT interpolated unescaped into a `bash -c` string (latent — PAT is GitHub-validated/canonical, but an unguarded shell sink). |
| 5 | low | `propagation_service.py:73` | The "needs PAT" predicate is re-derived in 3 places that must stay identical (loop risk if they drift) — unify into one helper. |
| 6 | low | `propagation_service.py:73` | `.env` read/patch/inject duplicates `_propagate_to_agent`; share it. |
| 7 | low | `git_service.py:192` | set-url command duplicates `initialize_github_sync` step 3; extract a helper. Also fix the `x-access-token:` docstring (code uses `oauth2:`). |
| 8 | low | `propagation_service.py:97` | Redundant 2nd `get_git_config` fetch + `getattr` masking a possible field rename. |
| 9 | low | `propagation_service.py:73` | Fleet-wide `list_all_agents_fast()` to check one agent — use `get_agent_container(name)`. |

**#1 and #2 are the actionable ones** — the fix currently overreaches from "per-agent PAT" to "any git agent when a global PAT exists," which changes behavior for agents #1264 wasn't about. Details inline.

vybe (approved):
Validated via /validate-pr: P1 bug fix (#1264), security clean, 8 unit + 45 suite tests pass, mergeable on dev. Note for release: author flags no live GitHub-repo E2E (reuses proven primitives) — worth a manual verify. Approving.

Issue #950, opened 2026-05-26
bug: deploy_local_agent (no GitHub repo) silently creates empty agents — fallback path writes outside host-mapped volume


vybe:
## Summary

`mcp__trinity__deploy_local_agent` (POST `/api/agents/deploy-local`) — the archive-based deploy path used by the `/trinity:onboard` skill from the [`abilityai/abilities`](https://github.com/abilityai/abilities) marketplace — is unreliable when the agent has no GitHub repository. Deployments either fail outright or succeed but produce **empty agents** (container starts with no template files, no CLAUDE.md, no skills).

## Root cause

In `src/backend/services/agent_service/deploy.py` (the `templates_dir` writability probe at the "Copy to templates directory" step), the writability check on `/agent-configs/templates` always fails because that mount is bind-mounted read-only in both `docker-compose.yml` and `docker-compose.prod.yml`:

```yaml
- ./config/agent-templates:/agent-configs/templates:ro
```

The probe catches the `OSError` and falls back to:

```python
templates_dir = Path("./config/agent-templates")
```

`./config/agent-templates` is **relative to the backend container's WORKDIR (`/app`)** — it resolves to `/app/config/agent-templates/<version-name>`, which is **not** host-mapped. The extracted template lives only inside the backend container filesystem.

When `create_agent_internal` (in `src/backend/services/agent_service/crud.py`, `elif config.template.startswith("local:")` branches around the `template_volume` setup) then tries to bind-mount the template into the new agent container, it does two things, both broken:

1. Checks `if template_path_in_backend.exists()` against `/agent-configs/templates/<name>` — the read-only mount where the template was **not** written. Existence returns `False`, so `template_volume` stays `None` and the new agent container is created **without the template mounted at `/template`** → empty workspace, "empty agents" failure mode.
2. If the directory happens to exist (race / stale state), the bind source path is computed from `HOST_TEMPLATES_PATH` (`${PWD}/config/agent-templates/<name>`), which doesn't exist on the host → Docker volume mount fails → "deploy fails" failure mode.

Net effect: the documented `/trinity:onboard` flow for archive-based deploy is broken on every standard deployment (dev + prod compose files both mark the templates dir `:ro`).

## How to reproduce

1. From a Claude Code agent **without** an associated GitHub repo, install the public marketplace plugin:
   ```
   /plugin marketplace add abilityai/abilities
   /plugin install trinity@abilityai
   ```
[...]

PR #1273, opened 2026-06-19
refactor(exec): fire-and-forget dispatch — a hung turn holds zero backend resource (#1083)

### LINKED ISSUE #1083: refactor: fire-and-forget dispatch — a hung turn holds zero backend resource

AndriiPasternak31:
## Summary

Implements **fire-and-forget dispatch** (#1083): eligible autonomous turns are dispatched to the agent with a **202 ACK** and run in the agent's background, then POST their terminal back to a new callback endpoint that finalizes the row. `execute_task` returns right after the ACK, so a wedged turn holds **zero** backend coroutine and the slot becomes a **lease** (released by the callback or reclaimed by the existing TTL reaper).

- **Flag-gated** `DISPATCH_ASYNC` (default **OFF**) **AND** Claude-runtime only. Non-202 responses (old image / non-Claude / flag off) fall through to today's **synchronous** path — the safe mixed-fleet fallback.
- **v1 eligible triggers:** `{schedule, webhook}` only. `loop`/`fan_out` stay sync; `event` bypasses `execute_task`.

## What's in this PR

- **Terminal applier** (`task_execution_service.apply_result`) — single point that finalizes an execution, shared by the inline sync path and the callback; gates **all** side-effects on the CAS bool so a replay / late callback completes no activity, records no breaker outcome, releases no slot.
- **Backend cutover** — async dispatch-and-return + durable async marker (`claude_session_id='dispatched_async'`) + **lease reaper** in `cleanup_service` (expired lease → FAIL + close activity, using each agent's `timeout + SLOT_TTL_BUFFER` window).
- **Callback endpoint** — `POST /api/agents/{name}/executions/{id}/result`: agent's own MCP key + ownership (404) + marker gate (409) + idempotent replay + body-size 413 cap. A late SUCCESS can still overwrite a reaper `LEASE_EXPIRED` via the CAS (Codex #2).
- **Agent side** (`result_callback.py`) — `try_spawn_async` runs the headless turn, builds the typed terminal envelope, **persists** it to `~/.trinity/pending-results/<eid>.json`, delivers with capped backoff up to the lease deadline; a startup sweep re-sends leftover envelopes after a crash/restart.
- **Security hardening** — strict `execution_id` allowlist (`^[A-Za-z0-9_-]{1,128}$`) enforced at `try_spawn_async` so a malformed/hostile id can never traverse out of the pending dir or inject into the callback URL (defense-in-depth; falls back to sync). CSO `--diff` audit report archived under `docs/security-reports/` — **0 CRITICAL/HIGH/MEDIUM**, the one LOW finding resolved in-branch.
- **Docs** — architecture.md "Fire-and-Forget Dispatch (#1083)" subsystem block + endpoint catalog + feature-flows index entry.

## v1 boundaries

lease-expiry = FAIL (not re-queue); async empty-result = FAIL (the #678 inline auto-retry stays sync-only); SUB-003 auto-switch stays inline-only; 504/503 async failures write a null-cost row until `execute_headless_task` exposes `ctx.metadata` on those paths (#1201 fast-follow).

## Tests

Unit: `test_1083_apply_result`, `test_1083_callback_endpoint`, `test_1083_dispatch_return`, `test_1083_lease_reaper`, `test_1083_db_seams`, `test_1083_result_callback` (incl. parametrized unsafe-`execution_id` coverage). Plus integration `test_1083_callback_endpoint`.

Closes #1083


vybe (approved):
Validated via /validate-pr: P1 reliability refactor (#1083). CSO --diff archived CLEAR (0 CRIT/HIGH/MED); callback endpoint hardened (agent-own-MCP-key→403, ownership→404 no-enum, dispatched_async marker→409 so it's inert until engaged, 413 caps, execution_id allowlist + #950 normpath/startswith containment). DISPATCH_ASYNC flag default-OFF + non-202 sync fallback → mixed-fleet safe; apply_result CAS-gates all side-effects; config properly packaged (both compose + .env.example); no schema change; ~120 unit tests; docs complete. Conflict was docs-index-only (feature-flows.md), resolved keeping both #1082/#1083 rows. Approving.

PR #1274, opened 2026-06-19
fix(auth): default-deny public self-signup on /api/access/request

dolho:
## Summary

`POST /api/access/request` (unauthenticated) auto-whitelisted any submitted email (role `user`), letting anyone who can reach the backend self-enroll and then log in via email OTP — so the email whitelist was **not authoritative** against self-enrollment, despite the "request access" naming implying admin approval.

**Fix — secure default:** gate the endpoint behind `PUBLIC_ACCESS_REQUESTS_ENABLED` (env) / the `public_access_requests_enabled` system setting, **default CLOSED**. When off → `403`, no whitelist write. The gate runs before body parsing so a disabled instance never processes self-signup input.

## Product decision (recorded per AC)
Keep frictionless auto-whitelist **behind the flag** rather than building a full admin-approval pending flow — minimal and non-breaking for opt-in (`trinity:connect`) users. A real pending flow can later build on the existing #311 `access_requests` precedent. Flipping the default off **is** a behavior change for anyone who relied on open self-signup — they set the flag to re-enable.

## Acceptance criteria
- [x] Flag gates the endpoint, default closed → 403 when disabled.
- [x] Product decision recorded (flag-gated auto-approve; pending-flow deferred).
- [x] Disabled → non-whitelisted email is **not** whitelisted (verified in tests).
- [x] Response wording no longer says "Access granted".
- [x] Already-whitelisted emails can still request login codes regardless of the flag (separate endpoint, untouched).
- [x] Tests: default rejects (403, no whitelist); enabling is explicit and covered.
- [x] CLI onboarding docs updated (cli-tool flow, architecture, `.env.example`).

## Testing
- `tests/unit/test_ent10_access_request_gate.py` (offline, 4 tests): 403 + no-whitelist when off; auto-whitelist (normalized email, role `user`) when on; idempotent; 400 validation once open. **All pass.**
- `tests/test_cli_access_request.py` rewritten: default-closed `403` (+ asserts the email isn't whitelisted via the admin list), and an operator-enabled class that toggles the setting via `PUT /api/settings/{key}` and restores it.
- `lint_sys_modules` green.

Related to trinity-enterprise#10

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -38,6 +38,13 @@ ANTHROPIC_API_KEY=
 # Users enter email → receive 6-digit code → login
 # Whitelist emails in Settings → Email Whitelist
 
+# Public self-signup (trinity-enterprise#10) — DISABLED by default (secure).
+# When false, POST /api/access/request returns 403 and never auto-whitelists,
+# so the email whitelist stays authoritative. Set to true ONLY if you want
+# frictionless CLI onboarding (`trinity:connect`) where anyone who can reach
+# the backend can add their own email to the login whitelist.
+PUBLIC_ACCESS_REQUESTS_ENABLED=false
+
 # Email service provider: console (dev), smtp, sendgrid, resend
 EMAIL_PROVIDER=console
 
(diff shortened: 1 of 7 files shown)


vybe (approved):
Validated via /validate-pr: P1 security fix (trinity-enterprise#10) — default-deny self-signup, fails secure (OFF→403), whitelist stays authoritative, no secrets. Docs + .env.example updated; 4 unit + CLI tests.

One non-blocking note: PUBLIC_ACCESS_REQUESTS_ENABLED isn't wired into docker-compose*.yml backend.environment, so that env lever is inert on prod-standalone — but the gate reads the DB system setting first (operators enable via PUT /api/settings), the secure default holds on deploy regardless, and this matches the existing SESSION_TAB_ENABLED settings-API-primary precedent. Approving.

PR #1290, opened 2026-06-20
docs(readme): document the Trinity Ops Agent and PostgreSQL backend/migration

vybe:
## What

Updates the top-level `README.md` to document two things that were missing:

1. **Trinity Ops Agent** ([trinity-ops-public](https://github.com/abilityai/trinity-ops-public)) — linked from the remote-deploy note and the docs index, described as the Claude Code-powered way to operate an instance (health, logs, updates, rollback, provisioning).
2. **PostgreSQL backend + migration** (#300) — a top-of-README callout recommending PostgreSQL for production while keeping SQLite as the zero-config dev default; a new **Database** section; a `DATABASE_URL` env-var row; and the SQLite→Postgres migration path via the ops agent's `/migrate-to-postgres` skill.

## Why this framing

Code still defaults to SQLite (`.env.example`, both compose files, and `docs/POSTGRESQL_SETUP.md` all confirm it), so the README **recommends** Postgres for production rather than claiming it is already the default — accurate today, and points existing instances at the ops-agent migration.

## Notes

- Public-repo safety scan on the diff: clean (only `your-password` placeholder; no secrets/IPs/PII).
- Every claim verified against the live `trinity-ops-public` repo, `.env.example`, the compose files, and `docs/POSTGRESQL_SETUP.md`.
- Follow-up (not in this PR): `docs/POSTGRESQL_SETUP.md` still says "no ETL tool yet" — now stale since the ops agent ships `/migrate-to-postgres`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,9 @@ Each agent runs in its own isolated Docker container with real-time observabilit
 
 > 🤖 **AI agent reading this repo?** Start at [AGENTS.md](AGENTS.md) — a task router with exact commands, key facts, and verification steps. Detailed docs index: [docs/user-docs/README.md](docs/user-docs/README.md).
 
+> [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


webmixgamer (approved):
Approved via /validate-pr. Docs-only README change; all 5 factual claims verified against code (TRINITY_DB_PATH, DATABASE_URL wiring in .env.example + both compose files, postgres profile, POSTGRESQL_SETUP.md, scheduler DATABASE_URL path). Security scan clean. Required CI checks green.

PR #1292, opened 2026-06-21
fix(security): authenticate the in-container agent server on the shared agent network (#1159)

### LINKED ISSUE #1159: security: agent-server HTTP API is unauthenticated on the shared agent network (cross-agent credential theft)
## Summary

The in-container agent server (FastAPI, port 8000) has zero authentication on any route, binds 0.0.0.0, and sets CORS `allow_origins=["*"]` with credentials (`docker/base-image/agent_server/main.py`). All agents share `trinity-agent-network` with default inter-container connectivity (`src/backend/services/agent_service/crud.py`). Routes include `GET /api/credentials/read` (returns raw `.env` secrets — `docker/base-image/agent_server/routers/credentials.py`) and `POST /api/chat` (arbitrary Claude execution).

Result: a single prompt-injected or compromised agent can read every sibling agent's credentials or execute code in them — lateral movement across the entire fleet.

## Context

2026-06-11 audit finding N2 (High — requires an in-network foothold; the port is not internet-exposed). Verified distinct from #948 (capability tokens are MCP-layer and would not stop direct HTTP to a sibling's :8000), #602, #877, and #1089 (which itself relies on the unauthenticated `/api/credentials/update`).

## Acceptance Criteria

- [ ] Per-agent shared token minted at agent create, injected as a container env var (same mechanism as `TRINITY_MCP_API_KEY`), stored where the backend can read it (NOT a Docker label — labels are visible in `docker inspect`)
- [ ] Agent-server middleware requires the token on all routes via constant-time compare, exempting only `/health` (Docker healthchecks)
- [ ] All callers send the header: backend `agent_client.py`, `task_execution_service`, the scheduler's agent client, operator-queue sync, and monitoring probes
- [ ] CORS tightened away from `*` + credentials
- [ ] Grace handling for old-image agents (enforce only when the env var is present; generate-on-start via the existing startup handshake)
- [ ] Negative integration test: an unauthenticated request from a second agent container gets 401

## Technical Notes

- Image-version skew is the main rollout risk — old images ignore the header (harmless), but new images must not lock out pre-existing agents.
- Chosen over per-agent Docker networks as the simpler control; revisit network isolation at replica-groups time (#927).

AndriiPasternak31:
## What & why

Closes #1159 (P1 · `type-bug` · `theme-security` · complexity-high).

The in-container agent server (`:8000`) had **zero inbound authentication** on `trinity-agent-network`. Any process on that shared network — including a prompt-injected/compromised sibling agent — could hit a peer's HTTP API to read its `.env` secrets or **run arbitrary Claude executions on it**. This PR closes that hole.

## Approach — derive-don't-store, fail-closed

Every backend→agent call now carries a per-agent token:

```
X-Trinity-Agent-Token = HMAC-SHA256(AGENT_AUTH_SECRET, "trinity-agent-auth:v1:" + name)
```

- The master `AGENT_AUTH_SECRET` lives **only in the backend env** (auto-generated by `start.sh` like `SECRET_KEY`/`INTERNAL_API_SECRET`; both compose files forward it). Each container receives **only its own** token, so a compromised agent A holds `token_A` but cannot compute `token_B` without the master. No DB column, no encryption mixin.
- Deriving from an empty secret **raises** (no silent fallback) — the backend hard-fails the first agent call if the operator ran without the secret.

## Enforcement — pure-ASGI middleware

`docker/base-image/agent_server/middleware/auth.py` is a **pure-ASGI** middleware (not `BaseHTTPMiddleware`) so the boundary is scope-complete — it gates **both HTTP and WebSocket** routes via constant-time compare, exempts only exact `/health` (+ `OPTIONS`), and never buffers SSE streams.

- Removed the **dead, unauthenticated `/ws/chat`** route (it ran `runtime.execute` — arbitrary Claude — and slipped past the original HTTP-only middleware). WS coverage now keeps any future WS route authenticated by default.
- Removed the agent server's wildcard CORS (`allow_origins=["*"]` + credentials) — it's internal-only.

## Caller discipline

All backend→agent traffic routes through `services/agent_auth.py` (`agent_httpx_client` / `build_agent_auth_headers` / `merge_auth_headers`). A **static guard** (`tests/unit/test_agent_auth_header_guard.py`) fails any new raw `agent-{name}:8000` caller that bypasses the helpers.

## Rollout / grace path

- Old image: empty `TRINITY_AGENT_AUTH_TOKEN` env → allow (so a mixed fleet keeps working during rollout).
- `check_agent_auth_token_env_matches` forces a **one-pass recreate** (deterministic derive → loop-safe) so a missing/stale (renamed-agent) token re-injects.
- Retired the unused `src/scheduler/agent_client.py` (+ its test).

## Testing

- Unit: token derivation, ASGI middleware (HTTP + WS + `/health` exemption), path matcher, header static-guard.
- Security: `tests/security/test_agent_auth_isolation.py` — sibling cannot forge a peer's token.
- CSO `--diff` audit committed at `docs/security-reports/cso-diff-2026-06-20.md` (H-1 `/ws/chat` finding resolved in-PR by removing the route).

## Docs

`docs/memory/architecture.md` → Container Security gains an "Agent-server inbound auth (#1159)" entry.


vybe (approved):
validate-pr: PASS (per your 'merge on validate-pr' decision) → dev. Security fix #1159 (agent-server HTTP API unauthenticated on shared agent network → cross-agent credential theft). ~50 files; backend code (main.py, chat.py, crud.py, git_service.py) + compose/.env ALL auto-merged cleanly against dev (#1297/#1293/#1294) — no code conflicts. Resolved 3 doc conflicts (feature-flows changelog union; container-capabilities took dev's canonical #1231 docs; cso-diff add/add preserved both audits). New agent_server/middleware/auth.py + agent_auth service; dedicated security tests (test_agent_auth_*, security/test_agent_auth_isolation.py) gate the merge. Approving.

Issue #1088, opened 2026-06-06
refactor: unify the failure classifier into one shared package


vybe:
## Problem

The auth-failure classification heuristic is **physically duplicated across 4 files**, hand-synced:

- `src/backend/services/subscription_auto_switch.py`
- `src/scheduler/service.py`
- `docker/base-image/agent_server/services/error_classifier.py`
- `docker/base-image/agent_server/services/headless_executor.py`

(`AUTH_INDICATORS` / `NON_AUTH_KILL_MARKERS` / `_is_auth_failure_message`). The backend docstring claiming "the scheduler imports this list" is **already false** — the drift is live.

## Why it matters

Every new non-auth exit shape that happens to carry an auth substring (max-turns, SIGKILL, OOM, call-budget) lands in the auth heuristic and triggers a **false subscription switch**. This is the recurring seed of the documented chain:

`#285 → #361 → #517 → #909 → #904` (5 breaks across 3 containers)

— each fixed by re-adding a precedence guard in one more copy. It is the #2 recurrence hotspot in the execution-bug meta-analysis (PR #1087, `docs/planning/ORCHESTRATION_BUG_META_ANALYSIS_2026-06.md`), classified `MISCLASSIFIED_FAILURE × PARALLEL_CODE_PATHS`.

## Proposed fix

Unify into **one shared package** with a single ordered precedence table (signal-exit → max-turns → auth → empty-result), consumed by all 4 sites. Delete the false docstring.

## Strategic value

Near-term enabler of the **typed terminal-reason on the reply envelope** (#945) — once classification lives in one place and the agent emits a typed `error_code`, the backend stops guessing from substrings entirely. This is the bridge from the current heuristic to the target-architecture result contract.

Surfaced by PR #1087.

Issue #904, opened 2026-05-21
bug: agent container OOM cascades into backend worker saturation and freezes admin UI


vybe:
## Summary

When an agent container exhausts its memory limit and the kernel OOM-kills processes inside it, the failure does not stay contained — it cascades into backend worker starvation and ultimately freezes the entire admin UI. Agent containers run with hard memory limits precisely so they can be isolated, but the current execution path makes the backend's two uvicorn workers vulnerable to slow / hung agent responses with multi-minute HTTP timeouts. The user-visible symptom is "UI is frozen"; the real cause is a single misbehaving agent.

## Component

Backend / Task Execution Service + Agent Runtime + Claude Code SDK error classification

## Priority

P1 — major feature (admin UI) broken with no in-UI workaround. Operator must SSH and restart the offending agent container to recover. No data loss, but the platform appears down to the operator.

## Observed Behavior

Reproducible cascade observed in production:

1. **Agent OOM**: an agent container with a 4 GB hard memory limit runs an internal `git` operation (the agent's workspace was itself a sizeable git repo). The kernel cgroup OOM-killer terminates the `git` process at ~4.1 GB anon-rss. Confirmed via `dmesg`:
   ```
   Memory cgroup out of memory: Killed process N (git)
     total-vm:5468752kB, anon-rss:4097292kB ...
     oom_memcg=/system.slice/docker-<container-id>.scope
   ```

2. **Claude subprocess dies at 0 turns**: the `claude` CLI subprocess (spawned by `agent_server.services.headless_executor`) dies with exit code `-9` before the first agentic turn completes. Stored in `schedule_executions.error` as:
   ```
   Execution terminated by SIGKILL after 0 tool calls / 0 turns (exit code -9).
   Likely cause: schedule/agent timeout exceeded, OOM kill, or operator cancel.
   ```

3. **Claude Code SDK misclassifies as auth failure**: in some fraction of cases, the SDK can't determine why its child died and reports:
   ```
   Claude Code execution failed (exit code -9):
     Subscription token may be expired or revoked.
     Generate a new one with 'claude setup-token'.
   ```
   This is a false signal — the token is fine. Verified by hashing `CLAUDE_CODE_OAUTH_TOKEN` across two agents sharing the same subscription: hashes matched, the other agent worked normally.

[...]

PR #1297, opened 2026-06-21
refactor(reliability): unify the SUB-003 auth-class failure classifier into one shared module (#1088)

AndriiPasternak31:
## What

Unifies the SUB-003 auth-class failure classifier (`is_auth_failure` + `AUTH_INDICATORS` + `NON_AUTH_KILL_MARKERS`) — previously duplicated inline in `subscription_auto_switch.py` and `scheduler/service.py` — into **one canonical module**.

Refs #1088 (`type-refactor`, `theme-reliability`).

## Why

The two copies were kept in sync by a hand-written "keep these lists in sync" comment. That manual-sync seam is exactly how the #904 kill-marker bug class (SIGKILL/OOM/timeout misclassified as an auth failure → spurious subscription auto-switch) re-appears. Replace it with a single source of truth + an automated drift guard.

## How

- **New `src/backend/services/failure_classifier.py`** — canonical, pure-stdlib classifier (55 lines). `subscription_auto_switch.py` now **re-exports** `is_auth_failure` unchanged, so existing importers (`routers/chat.py`, `services/task_execution_service.py`) and their test patch targets keep working.
- **New `src/scheduler/failure_classifier.py`** — **byte-identical vendored mirror**. The scheduler runs in a separate container and cannot import `backend.services`; it uses the classifier for **log-labelling only** (picks the `logger.error` wording, never gates a switch).
- The agent-runtime classifier in `error_classifier` is intentionally **NOT** merged (D4) — it diverges semantically (token-state markers, no kill-marker short-circuit) and stays kill-safe one level up via `_classify_signal_exit` precedence. `TestAgentClassifierKillSafetyContract` documents that divergence honestly.
- **Drift guards**: `TestBackendSchedulerParity` enforces byte-identity (edit the backend copy, regenerate the mirror — no hand-sync); `TestBackendReExportGuard` pins the re-export and forbids redefinition.

## Safety / scope

Pure structural refactor, **no behavioral change** — `AUTH_INDICATORS`, `NON_AUTH_KILL_MARKERS`, and the #904 kill-marker short-circuit are lifted verbatim (verified by SHA-256 equality of the two copies + line-for-line comparison vs the deleted code). The test rewrite also drops the prior `exec(compile(...))` source-slicing in favour of `importlib` path-loading, removing the only injection primitive in scope.

## Verification

- `tests/unit/test_904_sigkill_no_false_auth.py` — **19/19 pass**.
- CSO `--diff` audit: **CLEAR**, 0 CRITICAL / 0 HIGH (`docs/security-reports/cso-diff-2026-06-21.md`).

## Dual-track migrations

None — no schema, endpoint, model, or dependency changes.


vybe (approved):
validate-pr: PASS. Refactor (#1088, P2, theme-reliability) → dev. Single-source-of-truth classifier with byte-identity drift guards (TestBackendSchedulerParity) + re-export guard preserving existing patch targets. Security: no secrets/keys/env-vars added; CSO --diff CLEAR per body. Packaging: new src/scheduler/failure_classifier.py covered by 'COPY src/scheduler /app/scheduler'. Docs: feature-flows updated. All 4 required checks green. Approving.

Issue #183, opened 2026-03-26
SEC: Arbitrary File Write via Credential Injection Parameter Tampering


vybe:
## Pentest Finding 3.2.6 — Medium (CVSS 4.8)

**CVSS Vector**: [CVSS:4.0/AV:L/AC:L/AT:N/PR:L/UI:N/VC:N/VI:L/VA:N/SC:N/SI:N/SA:N](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:L/AC:L/AT:N/PR:L/UI:N/VC:N/VI:L/VA:N/SC:N/SI:N/SA:N)

**Location**: `POST /api/agents/{name}/credentials/inject`

## Impact

An authenticated user can supply any file path in the `files` map, causing the server to write attacker-controlled content to arbitrary filesystem locations inside the agent container. This can lead to configuration overrides or persistent code execution.

## Steps to Reproduce

```bash
POST /api/agents/shared/credentials/inject
{"files":{"/hello/anyfile.html":"<img src onerror=prompt(11)>"}}
# Returns 200 OK
# {"status":"success","files_written":["/hello/anyfile.html"]}
```

The server creates arbitrary directories and files with attacker-controlled content instead of restricting to predefined credential file paths.

## Remediation

- Define an allowlist of valid credential file paths (e.g., `.env`, `.mcp.json`)
- Reject any file path not in the allowlist
- Use `readOnly` property in schemas for properties that should never be modified
- Precisely define schemas, types, and patterns accepted in requests and enforce at runtime
- Do not automatically bind incoming data and internal objects

## References

- [OWASP Web Parameter Tampering](https://owasp.org/www-community/attacks/Web_Parameter_Tampering)
- [CWE-472: External Control of Assumed-Immutable Web Parameter](https://cwe.mitre.org/data/definitions/472.html)

_Source: UnderDefense Web Pentest Report, March 2026_

Issue #598, opened 2026-04-30
feat: structured MCP-server config endpoint (Layer 2 follow-up to #590)


pavshulin:
## Summary

Restore the user-facing ability to add/edit MCP servers post-deploy, lost in #590 (AISEC-C2 RCE closure), via a **structured endpoint** that validates inputs at the schema level — not raw JSON.

## Context

#590 (Apr 30 2026) closed the `.mcp.json` RCE-by-config bypass by removing `.mcp.json` and `.mcp.json.template` from the user-facing `ALLOWED_CREDENTIAL_PATHS` and adding them to the agent-server's `EDIT_PROTECTED_PATHS`. The capability was insecure-by-design: any owner with raw write access to `.mcp.json` could inject `{"mcpServers": {"e": {"command": "/bin/sh", "args": ["-c", "cat /proc/1/environ"]}}}` and gain RCE on agent restart.

UI feature regression after #590:
- "Save File" in `CredentialsPanel.vue` for `.mcp.json` returns 400 (disallowed path)
- Same for `.mcp.json.template`
- Owners can still add MCP servers at agent creation via the GitHub template repo's `template.yaml`
- Quick Inject `.env` (the dominant credential workflow) is unchanged

## Proposed design

**New endpoint**: `POST /api/agents/{name}/mcp-servers` (and `PUT`/`DELETE` for individual entries by id).

**Request body shape**:
```json
{
  "id": "google-workspace",
  "transport": "stdio",
  "command": "npx",
  "args": ["-y", "@modelcontextprotocol/server-google-workspace"],
  "env": { "GOOGLE_TOKEN": "${GOOGLE_TOKEN}" }
}
```

**Server-side validation**:
- `transport` ∈ `{stdio, http, sse}` enum
- For `stdio`: `command` MUST be in a managed allowlist (`npx`, `uvx`, `docker`, `python`, `python3`, `node`, `bun`, `deno`). No absolute paths, no `bash -c`, no `sh -c`.
- `args` rejected if any element contains shell metacharacters that could escape the runner (`;`, `|`, `$(`, backticks, redirects)
- For `http`/`sse`: HTTPS-only; hostname allowlist (npm registry, github raw, GCP/AWS endpoints, etc.) — same SSRF rules as #179 (skills library validation)
- `env` values are template references resolved from `.env` (`${VAR}` → looked up in agent's `.env` file). Arbitrary string values rejected.

[...]

PR #1305, opened 2026-06-22
feat(credentials): curated credential file-type injection (SA keys, certs, SSH, binary)

dolho:
## What

Widens CRED-002 credential injection from the fixed 3-path exact allowlist (`.env`/`.credentials.enc`/`.mcp.json`) to a **curated set of credential file types** — cloud service-account JSON, TLS cert/key material, kubeconfigs, SSH keys — **without** reopening the arbitrary-path RCE surface the allowlist closed (#183/#590/#598).

Related to **Abilityai/trinity-enterprise#11** (filed in the enterprise tracker; all code is OSS — base `dev`).

## How

- **Single-source policy** `services/credential_paths.py` (`is_allowed_credential_path`): ALLOW `.config/gcloud/**`, `.kube/config`, `*.pem`/`*.key`/`*.crt`/`*.cert`/`*.p12`/`*.pfx`, `.ssh/id_*` (+ the existing exact set, `.mcp.json` still content-validated). **Deny-list takes precedence** over ALLOW and blocks everything executed/sourced at startup (shell rc, `CLAUDE.md`/`AGENTS.md`/`.claude/**`, `.mcp.json.template`, `.ssh/authorized_keys`/`config`/`known_hosts`, `.git*`/`.gitconfig`, `bin/**`) plus `..`/absolute/backslash traversal.
- **Two-layer enforcement (Invariant #5):** the policy is **vendored byte-identically** into `docker/base-image/agent_server/credential_paths.py` (parity test enforces it). The agent-server write loops also gained a **resolve-under-home traversal guard the original code lacked** (today the backend gate was the only one).
- **Binary-safe:** inject carries `files_b64` (base64 → `write_bytes`); `.credentials.enc` becomes a v2 `{files, files_b64}` envelope (legacy flat archives still decrypt). `encrypt`/`decrypt` stay flat for the single-secret callers (SIEM/2FA/SSO) — only the new `encrypt_files`/`decrypt_files` handle binary.
- **Export captures the full set:** new agent `GET /api/credentials/list` discovers all policy-matching files so export→import round-trips everything injected (was just the 2 defaults).
- **Three surfaces in sync (Invariant #13):** MCP `inject_credentials` gains `files_b64`; frontend Credentials panel gains a file-upload affordance (UTF-8 vs base64 auto-detected).

## Tests

61 pass: the existing allowlist test now exercises the **real** policy (newly-allowed types + still-blocked dangerous paths), + a `credential_paths` byte-identity parity test + a binary archive round-trip test. `test_mcp_validator` + `test_files_guardrail_bypass` unaffected.

## Security

This intentionally loosens a deliberately-tight boundary. Posture preserved: deny-precedence, no startup/exec paths, no traversal, second-layer enforcement, `.mcp.json` content-validation. **`/cso` security review required before merge** (per the issue).

### DIFF
--- a/docker/base-image/agent_server/credential_paths.py
+++ b/docker/base-image/agent_server/credential_paths.py
@@ -59,6 +59,12 @@ DENY_GLOBS = [
     ".git/**", ".gitconfig",
     # anything on PATH
     "bin/**", ".local/bin/**",
+    # vendored / dependency / cache trees — keeps the broad cert globs
+    # (*.pem/*.key/*.crt) from matching bundled CA files (e.g. certifi's
+    # cacert.pem) and sweeping them into export's `/list` walk (#11 live test).
+    "node_modules/**", "**/node_modules/**",
+    "site-packages/**", "**/site-packages/**",
+    ".local/**", ".venv/**", "venv/**", ".cache/**", "go/pkg/**",
 ]
 
 
--- a/docker/base-image/agent_server/credential_paths.py
+++ b/docker/base-image/agent_server/credential_paths.py
@@ -34,7 +34,11 @@ ALLOW_EXACT = {".env", ".credentials.enc", ".mcp.json"}
 # matches one or more whole segments); a pattern with no "/" is matched against
 # the path's BASENAME (so "*.pem" allows a .pem anywhere not under a denied dir).
 ALLOW_GLOBS = [
-    ".config/gcloud/**",   # cloud SDK ADC / service-account JSON
+    # cloud SDK ADC / service-account JSON. NOTE: a workload-identity ADC file
+    # can declare an `executable` credential_source; google-auth only honors it
+    # when GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES=1 (non-default), so it is
+    # not an injection→exec vector here — but don't set that env var fleet-wide.
+    ".config/gcloud/**",
     ".kube/config",        # kubeconfig
     "*.pem", "*.key", "*.crt", "*.cert", "*.p12", "*.pfx",  # TLS / cert material
     ".ssh/id_*",           # SSH private/public keys (NOT authorized_keys/config)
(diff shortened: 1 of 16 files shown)


dolho:
🔒 `/cso --diff` ran; findings fixed in `18171c12`:

| # | Sev | Finding | Fix |
|---|-----|---------|-----|
| 1 | **HIGH** | `.mcp.json` content guard (#598) bypassable via the new `files_b64` channel → stdio-command MCP server RCE on a sibling agent | `.mcp.json` rejected in the binary channel (backend inject + agent-server + import) — text-only so `validate_mcp_config` always runs |
| 2 | LOW | import/auto-import validated only at the agent-server layer | added `validate_credential_set()` on the backend import boundary (dual-layer) |
| 3 | LOW | `.ssh/` allowed stray `*.key`/`*.pem` | locked to `id_*` only |
| 4 | LOW | gcloud executable credential_source caveat | documented (gated by non-default env var) |

+6 regression tests (169 pass). Report: `docs/security-reports/cso-2026-06-22-11-diff.md`. Posture otherwise sound — default-deny, deny-precedence, traversal guard, two-layer, encrypted archive.

vybe (approved):
Approving via /validate-pr.

Careful, defense-in-depth widening of the CRED-002 injection allowlist. Verified:
- Deny-precedence policy: traversal/null/backslash/absolute rejected; shell-rc, CLAUDE.md/AGENTS.md/.claude/**, .mcp.json.template, .ssh exec vectors, .git/**, bin/** and vendor trees all denied; .ssh hardened to id_* only.
- /cso --diff committed in-tree; its HIGH finding (files_b64 → .mcp.json content-guard bypass → MCP stdio RCE) confirmed fixed at BOTH layers (backend 400 + validate_mcp_config; agent-server 400 + resolve-under-home guard).
- Invariant #5 parity test enforces byte-identity of the two policy copies; 3-surface sync (backend + agent-server + MCP + frontend) intact.
- All 16 CI checks green (CodeQL, pytest ×6 seeds, prod-smoke, schema-parity, verify-non-root).

Accepted residual caveat (CSO #4, documented): gcloud executable-ADC credential_source — mitigated by not setting GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES=1 fleet-wide.

PR #1320, opened 2026-06-23
chore: reconcile dev with main after v0.7.0

vybe:
Brings the v0.7.0 release squash, the release notes (`docs/releases/0.7.0.md`), and the #1115 `get_agent_schedules_summary` Postgres-safe fix back onto `dev`.

Per repo convention, **merge this with a merge commit (NOT squash)** so main's commits become ancestors of dev and future dev→main diffs stay clean.

Post-release reconciliation for #1319 / tag `v0.7.0`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1115, opened 2026-06-09
feat(ui): per-schedule performance scorecards on Agent Detail (Overview + Schedules tab)


vybe:
## Summary

Surface **per-schedule performance** on the Agent Detail page in two places, both schedule-centric (one scorecard per schedule, labeled by its command, e.g. `/do-something`):

1. **Overview tab** (`OverviewPanel.vue`, #1107) — a **Schedules performance** section giving a glanceable, per-schedule comparison so an operator can see at a glance *which* scheduled job is healthy and which is degrading, without drilling into each one.
2. **Schedules tab** (`SchedulesPanel.vue` / the existing schedules list) — inline **basic stats per schedule row** (success rate, last-run outcome, avg duration, runs in window) so the schedule list itself is glanceable, not just a list of cron strings.

The Overview today shows agent-wide trends; per-schedule rollups are invisible there, and the Schedules tab shows configuration but no performance signal at all.

## Context

Two pieces already exist but don't meet:
- **#1107** shipped the Overview tab with agent-scoped trend charts (executions-by-type, success rate, duration, context) — but it is aggregated across *all* triggers, so a single misbehaving schedule is invisible.
- **#868** shipped per-schedule deep analytics at `GET /api/agents/{name}/schedules/{id}/analytics` (counts, terminal success rate, duration p50/p95/p99, cost total, tool-call top-5 + `total_calls`, daily timeline) — but it's one schedule per call and is not surfaced on either the Overview or the Schedules list.

The gap: there's no place that answers "for each command this agent runs on a schedule, how is it performing?" The Overview (default landing tab) is the natural home for the comparison view; the Schedules tab is the natural home for at-a-glance per-row stats. Both consume the **same** compact aggregate so there are no N round-trips. See `docs/memory/feature-flows/agent-overview-dashboard.md`.

## Acceptance Criteria

**Shared data**
- [ ] Per-schedule rollups come from a **single aggregate call** (new compact endpoint, or one batched query) — rendering does **not** issue N per-schedule round-trips, consistent with the Overview's "fetch once, never poll" discipline. Both surfaces reuse this one call.
- [ ] Each rollup is computed for the selected window: **success rate** (terminal-based: success / (success + failed+error); gap/`—` when zero terminal runs), **token/cost usage**, **avg duration**, and **total tool calls**.
- [ ] Read-only and DB-sourced — renders even when the agent is stopped (matches Overview behavior).
[...]

PR #1321, opened 2026-06-23
chore: sync dev after v0.7.0 (release notes + #1115 fix)

vybe:
Post-release reconciliation for tag `v0.7.0`. Brings the two things `dev` is missing onto it, conflict-free (this branch was cut from dev's tip):

- `docs/releases/0.7.0.md` — the release notes (also on main via #1319)
- **#1115 fix** — `get_agent_schedules_summary` ported to SQLAlchemy Core + portable `ROW_NUMBER()` (was NameError'ing/500ing on dev after the #300 migration dropped its `get_db_connection` import) + refreshed `test_login_rate_limit_split` config stub

After merge, dev's tree matches main's. Avoids the squash-divergence conflict a direct main→dev merge hits (see closed #1320).

### DIFF
--- a/docs/releases/0.7.0.md
+++ b/docs/releases/0.7.0.md
@@ -1,0 +1,112 @@
+# v0.7.0 — PostgreSQL goes to production
+
+**Released:** 2026-06-23 · **Tag:** `v0.7.0` · **Previous:** `v0.6.1`
+**Diff:** [v0.6.1...v0.7.0](https://github.com/abilityai/trinity/compare/v0.6.1...v0.7.0)
+
+The headline of this release: **Trinity now runs on PostgreSQL as the recommended
+production database.** SQLite stays the zero-config default for local development
+and evaluation, but production instances should move to Postgres. This release adds
+the configurable backend, Alembic-managed Postgres migrations, a dated SQLite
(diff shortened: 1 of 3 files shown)


dolho:
## Review — LGTM ✅

Release-sync `release/0.7.0 → dev`, all checks green. Three files:

- **`docs/releases/0.7.0.md`** — release notes, docs only.
- **`db/schedules.py`** — `get_agent_schedules_summary` ported from raw sqlite to SQLAlchemy Core for SQLite+Postgres portability (#300/#1115). This is a real fix, not just a port: the comment notes the old path NameError'd on the dropped `get_db_connection` import, and the SQLite-only bare-column-with-MAX last-run trick is replaced by a portable `ROW_NUMBER() OVER (PARTITION BY schedule_id ORDER BY started_at DESC)`. In-window `base_where` (`started_at > cutoff`) is preserved across all four queries; AVG(NULL-skip) semantics match the prior CASE-wrapped AVG; tool-call `cap+1` sampling detection intact. Correct.
- **`test_login_rate_limit_split.py`** — adds `PUBLIC_ACCESS_REQUESTS_ENABLED = False` stub to track auth.py's new dependency (ent#10).

Substantive but tested and green. Mergeable.

dolho (approved):

Issue #1342, opened 2026-06-24
bug: schema-parity CI doesn't guard the Alembic (Postgres) track — PG-breaking schema changes merge green


vybe:
## Summary

The `schema-parity` required check validates **only the SQLite track** (`src/backend/db/migrations.py` ↔ `src/backend/db/schema.py`). A schema change that ships the SQLite migration and updates `schema.py`/`tables.py` but **omits the Alembic revision** under `src/backend/migrations/versions/` passes every required check **green** — yet breaks PostgreSQL, where `init_database()` runs `alembic_runner.upgrade_to_head()` (which applies revision files only; it does **not** autogenerate from `tables.py`).

## Context

Architectural Invariant #3 requires **both** a SQLite migration and a new Alembic revision for every schema change (until SQLite end-of-support, 2026-09-01). With no automated guard for the Alembic side, two PRs reached "green CI but PG-broken" in a single review pass and had to be held manually — one adding columns to `agent_loops`, one adding a column to `agent_ownership`. Reviewers currently have to catch this by hand.

## Acceptance Criteria

- [ ] CI fails when a PR modifies `src/backend/db/schema.py` / `src/backend/db/tables.py` or adds a SQLite migration in `src/backend/db/migrations.py` **without** adding a new Alembic revision under `src/backend/migrations/versions/`
- [ ] Wired as a required check (or folded into the existing `schema-parity` job)
- [ ] A fixture PR with a SQLite-only schema change fails the check; a properly dual-tracked PR passes
- [ ] False-positive guard: non-schema edits to those files (comments, data-only/down migrations) don't trip it; the heuristic is documented in the workflow

## Technical Notes

- Existing guard: `.github/workflows/schema-parity.yml` (currently SQLite-only).
- The single-source consolidation (`tables.py` MetaData → autogenerated revisions, retiring `migrations.py`) is the long-term goal (#746) but is **deferred** — until then both tracks are hand-authored, so the guard must require the revision file.
- Heuristic: a diff touching the SQLite migration runner or the table DDL should require a net-new file under `src/backend/migrations/versions/`.

PR #1344, opened 2026-06-25
docs(architecture): trim architecture.md under the 150k-char context limit

dolho:
## Why

`docs/memory/architecture.md` had grown to **~156.7k chars** (on `dev`), past the **150k** soft limit Claude Code warns about. The file is `@`-imported by `CLAUDE.md`, so it's auto-loaded into context every session — over the limit it risks silent truncation and eats a large slice of the context window each turn.

## What

Compressed the densest prose, preferring **summary + pointer** wherever a dedicated `feature-flows/` doc already owns the deep detail (the doc's own editorial rule #2):

- Cross-Cutting Subsystem narratives: Agent Runtimes/Codex parity, Status-as-projection, Circuit Breakers, Idempotency, Heartbeat, Soft-Delete sweeps, Session Tab, Real-time, VoIP, Git Sync, Agent Compatibility, Agent Runtime Data, Subscription token rotation.
- Architectural Invariants #3 (dual-track migrations) and #17 (non-root containers).
- A few over-long catalog/endpoint rows and the longest frontend UI prose blocks.

## Result

**156.7k → 149.3k chars (~4.8% smaller)** — clears the warning.

## Safety

- **No facts dropped** — every issue tag, field name, default, and error-string preserved.
- Protected SQLite DDL (tracked by `/validate-schema`) **untouched**.
- Heading / code-fence / table counts unchanged vs `dev` (68 / 58 / 32).
- All 12 newly-added `feature-flows/*.md` links resolve.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1345, opened 2026-06-25
fix(ci): guard the Alembic (Postgres) track in schema-parity (#1342)

dolho:
Related to #1342

## Problem

The `schema-parity` required check validated **only the SQLite track** (`db/migrations.py` ↔ `db/schema.py`). A schema change that ships the SQLite migration + updates `schema.py`/`tables.py` but **omits the Alembic revision** under `src/backend/migrations/versions/` passed every required check **green** — yet broke PostgreSQL, where `init_database()` runs `alembic_runner.upgrade_to_head()` (applies revision files only; does **not** autogenerate from `tables.py`). Two PRs reached "green CI but PG-broken" in one review pass.

## Fix

A cross-track guard **folded into the existing required `schema-parity` job** (no new required-check to manage — AC #2):

- **`scripts/ci/check_alembic_parity.py`** — fails a PR that ADDS schema DDL to `db/{migrations,schema,tables}.py` without a net-new file under `src/backend/migrations/versions/`. Pure stdlib, PR-only (diffs `base...head`).
- **Heuristic / false-positive guard (AC #4)** — the signal is a DDL keyword on an *added, non-comment* line:
  - SQL (migrations.py/schema.py): `CREATE TABLE/INDEX/TRIGGER`, `ALTER TABLE`, `ADD/DROP COLUMN`, `RENAME …`
  - SQLAlchemy (tables.py): `Column(`/`Table(`/`Index(`/`UniqueConstraint(`/`ForeignKey(`
  - Comment edits, **data-only** and **down** migrations carry no DDL keyword → don't trip. Documented in the script docstring + workflow header.
- **`tests/unit/test_alembic_parity_guard.py`** — 20 unit tests incl. the acceptance fixtures (SQLite-only change **fails**; dual-tracked **passes**; comment/data-only **pass**). Wired into the parity pytest run.
- Notes the enforcement in `architecture.md` Invariant #3.

## Acceptance criteria

- [x] CI fails when a PR adds SQLite-track schema DDL without a new Alembic revision
- [x] Wired as part of the existing required `schema-parity` job
- [x] Fixture: SQLite-only schema change fails; dual-tracked passes (unit tests + e2e smoke)
- [x] False-positive guard: comments / data-only / down migrations don't trip it; heuristic documented

## Verification

```
24 passed   # 4 existing schema-parity + 20 new guard tests
```

End-to-end smoke against real git refs:
- clean (no DB DDL) → **pass** (exit 0)
- synthetic SQLite-only `ADD COLUMN`, no revision → **FAIL** (exit 1, prints offending line)
- + paired revision file → **pass** (exit 0)


vybe (approved):
Validated: adds schema-parity Alembic cross-track guard (check_alembic_parity.py + test, #1342) — directly closes the 'schema-parity misses Alembic' gap. architecture.md conflict vs #1344 resolved (trim preserved + CI-guard clause layered in). All 4 required checks green.

PR #1347, opened 2026-06-25
feat(ui): reframe Sharing tab to external-client sharing via channels

dolho:
Implements **trinity-enterprise#18** (Access/Sharing redesign epic, trinity-enterprise#16). Frontend-only; no API changes.

The Access tab (#17) already owns Trinity operators — this scopes the **Sharing** tab to the operator → external-client surface.

## Changes
- **Google-Docs-style framing** — "Share this agent" header; operator language removed (operators are managed on the Access tab).
- **Single Restricted ↔ Open control** — replaces the two policy checkboxes with one segmented control over `require_email`/`open_access`:
  - **Restricted** → `require_email:true, open_access:false` (approval-gated)
  - **Open** → `require_email:true, open_access:true` (anyone verified)
  - Identity proof stays on for external sharing; a legacy `require_email:false` agent is shown by its `open_access` flag and upgraded the next time a mode is picked (documented inline).
- **Pending requests** kept, reframed as external clients awaiting approval.
- **Channels as compact summary rows** — new reusable `ChannelDisclosure.vue` (collapsed by default). Detailed config stays reachable inside the expanded row as a **non-regressing interim seam**; #19 ("move channel config into dialogs") swaps the body for a modal. No channel functionality removed.
- **Distribution nudge** — outbound file sharing + public links moved into a separate "Distribution" section (distribution, not client access), per the issue's technical note.

## Acceptance criteria
- [x] Sharing tab scoped to external clients + channels only (operator allow-list removed — lives on Access tab)
- [x] External access policy as a single Restricted ↔ Open control
- [x] Pending access requests surfaced here
- [x] Channels as compact summary rows (detailed config deferred to the dialogs issue #19)
- [x] Google-Docs-style "share this agent" framing

## Verification
- Both SFCs compile via `vue/compiler-sfc` (script setup + template).
- `npm run check:tokens` passes.
- Note: full `vite build` couldn't run in my sandbox (unrelated stale-`node_modules` `mermaid` dep gap in `AgentWorkspace.vue`, not touched here) — CI installs deps fresh.

## Notes
- `SharingPanel` prop/emit contract (`agent-name`, `shares`, `@agent-updated`) unchanged.
- Channel summary rows use emoji glyphs as a dependency-free placeholder; #19 can swap for brand icons.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #19, opened 2026-02-24
MCP Execution Query Tools (MCP-007)


vybe:
Three new MCP tools:
- `list_recent_executions` - query what happened across fleet
- `get_execution_result` - fetch specific output
- `get_agent_activity_summary` - high-level status

Spec: docs/requirements/MCP_EXECUTION_QUERY_TOOLS.md

Issue #18, opened 2026-02-24
Unified Executions Dashboard (EXEC-022)


vybe:
Combined view of all executions across all agents. Filter by agent, time, trigger type. Real-time WebSocket updates. Click through to details.

Spec: docs/requirements/UNIFIED_EXECUTIONS_DASHBOARD.md

Issue #894, opened 2026-05-18
Add per-agent model override for public communication channels


vybe:
## Summary

Add a per-agent setting in the Sharing UI to select which Claude model is used for all public-facing channels (public chat link, Slack, Telegram, WhatsApp). Today public chats always use the platform-wide default model with no per-agent override available to agent owners.

## Context

Currently the model resolution for public chats is:

- `routers/public.py:610-618` calls `task_execution_service.execute_task()` with `model=None`
- `services/task_execution_service.py:286-287` falls back to `settings_service.get_platform_default_model()`
- `services/settings_service.py:233-247` reads `system_settings.platform_default_model` (hardcoded fallback `claude-sonnet-4-6`)

This means:
- The only knob is platform-wide (admin-only).
- Agent owners cannot opt their public-facing agent into a cheaper model (Haiku) for cost control or a more capable model (Opus) for quality on a per-agent basis.
- The same problem applies to Slack/Telegram/WhatsApp adapter messages routed through `adapters/message_router.py`, which currently inherits the same default.

Sharing is the natural home for this setting — it already governs how external users reach the agent (sharing list, public link, channel bindings, access policy).

## Acceptance Criteria

- [ ] New column on `agent_ownership` (e.g. `public_channel_model TEXT NULL`) — NULL means inherit platform default.
- [ ] `GET /api/agents/{name}/public-channel-model` returns the resolved value + whether it is overridden.
- [ ] `PUT /api/agents/{name}/public-channel-model` allows owner/admin to set or clear the override; validates against a known model whitelist (Opus 4.7, Sonnet 4.6, Haiku 4.5).
- [ ] `routers/public.py` and `adapters/message_router.py` resolve model in this order: per-agent override → platform default → hardcoded fallback.
- [ ] Sharing tab in `AgentDetail.vue` gains a model selector with three options + "Use platform default".
- [ ] Audit log entry written when override is changed (event_type `agent_lifecycle`, action `update_public_model`).
- [ ] Existing scheduler-triggered runs (which already have their own `agent_schedules.model` per-schedule column) are unaffected.

## Technical Notes

- Three surfaces to update (Architectural Invariant #5/#13): backend router + service, frontend Pinia store + Sharing view, and probably a new MCP tool is **not** required (this is owner-facing config, not agent-callable).
- Model whitelist should live in one place — consider extending `settings_service` with a `is_valid_model()` helper rather than duplicating the list.
[...]

PR #1348, opened 2026-06-25
feat(ui): move channel config into dialogs + VoIP surface (#19)

dolho:
Implements **trinity-enterprise#19** (Access/Sharing redesign epic, trinity-enterprise#16). Frontend-only; no API changes.

> ⛓️ **Stacked on #1347 (#18)** — base is `feature/18-reframe-sharing-external-clients` so the diff is #19-only. Retarget to `dev` once #1347 merges.

Replaces #18's interim collapsible channel rows with the dialog pattern the issue calls for: the tab stays a clean summary; full config opens in a modal.

## Changes
- **`ChannelConfigDialog.vue`** — Teleport modal (backdrop + Esc close, scrollable body) matching the `ConfirmDialog` idiom.
- **`ChannelConfigRow.vue`** — compact summary row: channel glyph, **connected/off + brief status** (Slack `#channel`, Telegram `@bot`, WhatsApp/VoIP number, VoIP "(disabled)"/setup-needed warnings), and a **Configure/Manage** button that opens the dialog. A lightweight status GET (shared `api` client) refetches on dialog close so the row reflects edits.
- **`SharingPanel.vue`** — wires Slack / Telegram / WhatsApp / **VoIP** through `ChannelConfigRow` with per-channel status derivations. The 4 existing panels render **untouched** inside the dialog slot (no loss of functionality). VoIP — previously backend-only — now has its Sharing UI surface (gated on `voip_available`).
- Distribution (public links, file sharing) keeps the collapsible disclosure — outside this issue's scope.

## Acceptance criteria
- [x] Slack/Telegram/WhatsApp panels rendered inside dialogs, not inline
- [x] Each channel shows a summary row (connected/off + brief status) with a Configure button opening its dialog
- [x] VoIP gets a UI surface here (gated on `voip_available`)
- [x] No loss of existing channel functionality (panels reused as-is in the dialog)

## Verification
- 3 SFCs compile via `vue/compiler-sfc`; `npm run check:tokens` passes.
- Vite transforms all three modules on the live local dev server (HTTP 200) — verified against the running instance on this branch.


src/frontend/src/components/ChannelConfigRow.vue
```diff
@@ -0,0 +1,90 @@
...
+        <template v-if="loading">
+          <span class="inline-block w-2 h-2 rounded-full bg-gray-300 dark:bg-gray-600 animate-pulse"></span>
+          <span class="text-gray-400 dark:text-gray-500">Checking…</span>
+        </template>
+        <template v-else-if="status.connected">
+          <span
+            class="inline-block w-2 h-2 rounded-full"
+            :class="status.warn ? 'bg-status-warning-500' : 'bg-status-success-500'"
+          ></span>
+          <span class="text-gray-600 dark:text-gray-300 truncate">{{ status.label || 'Connected' }}</span>
```
dolho: **[Medium] Status label won’t truncate — long labels overflow the row**

This `truncate` span is a flex child without `min-w-0`, so it can’t shrink below its content width and a long status label (e.g. WhatsApp’s `whatsapp:+14155238886`) overflows the row instead of clipping. Same flexbox class of bug the dialog `v-if` fix addressed, but on the inner status line — latent until a channel actually connects.

```suggestion
          <span class="text-gray-600 dark:text-gray-300 truncate min-w-0">{{ status.label || 'Connected' }}</span>
```

Optional polish: add `shrink-0` to the status dots (lines 11/16/23) and the “· setup needed” span (line 20) so they don’t squish when the label is long.

src/frontend/src/components/ChannelConfigRow.vue
```diff
@@ -0,0 +1,90 @@
...
+      class="shrink-0 inline-flex items-center px-3 py-1.5 text-sm font-medium rounded-md border border-gray-300 dark:border-gray-600 text-gray-700 dark:text-gray-200 bg-white dark:bg-gray-800 hover:bg-gray-50 dark:hover:bg-gray-700 focus:outline-none focus:ring-1 focus:ring-action-primary-500"
+      @click="dialogOpen = true"
+    >
+      {{ status.connected ? 'Manage' : 'Configure' }}
+    </button>
+
+    <!-- Config dialog: mounted only while open so the closed dialog (and its
+         slotted panel) never participates in this row's flex layout. The
+         channel panel renders inside the modal, untouched. -->
+    <ChannelConfigDialog v-if="dialogOpen" :open="dialogOpen" :title="title" :icon="icon" @close="onDialogClose">
```
dolho: **[Low] Redundant triple-gating of the dialog**

Now that the dialog is mounted only via `v-if="dialogOpen"`, both `:open="dialogOpen"` here and `ChannelConfigDialog`’s internal `v-if="open"` are always true while mounted — dead weight. Consider either dropping this parent `v-if` (rely on the internal Teleport + `v-if`, the original approach) **or** dropping the `open` prop + internal `v-if` (rely purely on mount/unmount). All three together works but is confusing. Not a bug.

src/frontend/src/components/ChannelConfigDialog.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <Teleport to="body">
+    <div v-if="open" class="fixed z-50 inset-0 overflow-y-auto" role="dialog" aria-modal="true" :aria-label="title">
```
dolho: **[Low] Modal a11y: no focus management or scroll lock**

`role="dialog" aria-modal="true"` is set, but focus isn’t moved into the dialog on open, there’s no focus trap, and background scroll isn’t locked. This matches the existing `ConfirmDialog.vue`, so it’s not a regression — flagging for parity if a shared modal primitive is extracted later.

dolho:
Applied the review fixes in 304c1a1f:

- **[Medium]** Status label now truncates — `min-w-0` on the label + `shrink-0` on the dots and "· setup needed" span, so a long connected label (e.g. `whatsapp:+14155238886`) clips instead of overflowing the row.
- **[Low]** Collapsed the redundant triple-gating to a single mechanism: the dialog mounts only via the parent `v-if="dialogOpen"`; dropped the always-true `:open` prop and `ChannelConfigDialog`'s internal `v-if="open"` (Escape now binds for the component lifetime via onMounted/onUnmounted).
- **[Low — a11y]** Left as-is: focus-trap/scroll-lock would be new behavior beyond the existing `ConfirmDialog` parity (not a regression).

Verified: both SFCs compile (`vue/compiler-sfc`), `check:tokens` passes, Vite serves the updated modules.

dolho:
## Review + fixes applied

Re-reviewed the diff and applied the outstanding proposed fix. Status of the three earlier findings:

- 🟡 **[Medium] status label truncation** — **already fixed** (line 19 now `min-w-0 truncate`; status dots + "· setup needed" carry `shrink-0`). ✅
- 🟢 **[Low] redundant dialog gating** — **already fixed**; the dialog is mounted via `v-if="dialogOpen"` only, no `:open` prop / internal `v-if` left. ✅
- 🟢 **[Low] modal a11y** — **fixed now** (4828ddc6): `ChannelConfigDialog` moves focus into the dialog on open (first focusable, else the panel via `tabindex="-1"`), traps Tab with wrap-around, locks background scroll (`body.overflow`), and restores focus to the opener + prior overflow on close. Esc-to-close unchanged.

### Fresh-pass notes (no blockers)
- `deriveStatus` derivations (Slack `#channel`, Telegram `@bot` + webhook warn, WhatsApp/VoIP number, VoIP `(disabled)`/warn) read correctly against their endpoints; failures degrade to "Not connected" by design.
- `ChannelConfigRow` refetches status on dialog close — good, the summary reflects edits.

### Note
`ConfirmDialog.vue` has the same a11y gap; the durable fix is a shared modal primitive applying focus-trap/scroll-lock to both — tracked as follow-up, out of this issue's scope.

Verified: 3 SFCs compile (`vue/compiler-sfc`), `check:tokens` passes, clean `npm ci` + production vite build green (2702 modules).

vybe (approved):
Validated via /validate-pr: effective diff vs dev is exactly the 3 intended files (2 new SFCs + SharingPanel channels-section swap); all 3 SFCs compile; security greps clean; no API/schema/config surface. Resolved the dev-drift merge conflict (kept #894/#1205/ent#46/#20 sections, re-applied the #19 ChannelConfigRow delta; MCP-connector row stays a ChannelDisclosure as it predates the row pattern). Cross-tracker ent#19 needs a manual status-in-dev bump after merge.

Issue #874, opened 2026-05-17
fix(security): add non-root USER directive to 4 production Dockerfiles (CSO hardening, persistent since 2026-04-05)


AndriiPasternak31:
## Summary

Four production Dockerfiles run their CMD as root. Only `docker/base-image/Dockerfile` (agent base) drops to a non-root `developer` user. This is a defense-in-depth gap — not a standalone exploit, but it amplifies the blast radius of any future RCE.

Flagged by `/cso` at MEDIUM, 10/10 confidence, **persistent since the 2026-04-05 audit** (~6 weeks).

## Affected Files

- [ ] `docker/backend/Dockerfile` — FastAPI backend, `python:3.11-slim` base, no `USER` directive (ends at line 70). **Highest risk** — backend has `/var/run/docker.sock:ro` mounted, so root inside this container can enumerate every Trinity agent container by label.
- [ ] `docker/scheduler/Dockerfile` — APScheduler service, `python:3.11-slim` base, no `USER` directive.
- [ ] `docker/frontend/Dockerfile.prod` — `FROM nginx:alpine`, no `USER` directive. nginx master defaults to root; only worker processes drop to `nginx`.
- [ ] `src/mcp-server/Dockerfile` — TypeScript MCP server, `node:22-alpine` base, no `USER` directive.

No `user:` override in `docker-compose.yml` or `docker-compose.prod.yml`, confirming containers actually run as root in production.

## Exploit Scenario

Any RCE in the FastAPI backend (e.g., a future variant of #590 file-write RCE, a deserialization bug in a router, a vulnerable dep introduced via supply chain) lands as **root** inside the backend container. Because `/var/run/docker.sock` is bind-mounted read-only:

1. Attacker reads Docker socket → enumerates every agent container, owner labels, network bindings.
2. Reads in-container env vars (`ANTHROPIC_API_KEY`, `REDIS_URL` with backend creds, `SECRET_KEY`).
3. Reaches Redis on `trinity-platform-network` with the credential pair already in env.
4. If combined with any write primitive (volume bind to host paths, Redis ops on platform queue), pivots further.

Read-only socket blocks `docker exec` / `docker run`, so this is not a direct container escape. But "no USER" turns a single-point RCE into a fleet-wide reconnaissance primitive.

## Acceptance Criteria

- [ ] `docker/backend/Dockerfile` ends on `USER trinity` (or equivalent non-root UID 1000).
[...]

PR #1352, opened 2026-06-25
fix(deploy): probe in-container docker.sock GID so non-root backend reaches Docker on Docker Desktop (#1131)

### LINKED ISSUE #1131: bug: non-root backend cannot reach docker.sock on macOS Docker Desktop — agents silently vanish from UI (#874 regression)
## Summary

After upgrading through #874 (non-root container hardening), the backend running as UID 1000 cannot reach `/var/run/docker.sock` on **macOS Docker Desktop**, so `GET /api/agents` returns `[]` and the Agents page shows **"No agents"** — even though agent containers are running with correct `trinity.platform=agent` labels. The failure is **silent**: `list_all_agents_fast` swallows the Docker exception (per Architectural Invariant #17) and returns an empty list, so there is no error surfaced to the operator.

## Context

On Docker Desktop the in-container `/var/run/docker.sock` is owned `root:root` with mode `srw-rw----` (UID 0, **GID 0**). The backend runs as UID 1000 with `group_add: ${DOCKER_GID:-999}` (docker-compose.yml). Group `999 != 0`, so the non-root process is denied access to the socket:

```
PermissionError: [Errno 13] Permission denied   # docker/transport/unixconn.py: sock.connect(self.unix_socket)
```

The current guidance assumes this case cannot happen on Docker Desktop, which is incorrect:

- `scripts/deploy/start.sh` — `ensure_docker_gid()` returns early on non-Linux hosts, with the comment *"macOS Docker Desktop ignores group_add — fall through with the default."*
- `docs/migrations/NON_ROOT_CONTAINERS_2026-05.md` (line ~133) — *"group_add is silently ignored on Docker Desktop — leave DOCKER_GID [at default]."*

So the documented "safe default" (`DOCKER_GID=999`) is exactly the broken path for affected Docker Desktop setups.

### Verification (controlled counterfactual)

Same UID and socket mount, isolating only the supplementary group:

| Setup | Result |
|-------|--------|
| Socket perms in container | `srw-rw---- root root` (GID 0) |
| UID 1000 + `--group-add 999` (default) | `PermissionError: [Errno 13] Permission denied` |
| UID 1000 + `--group-add 0` (fix) | socket reachable / `docker ping: True` |
| Backend with `DOCKER_GID=0` | `GET /api/agents` returns all running agents |

This isolates the supplementary GID as the sole cause, independent of the backend container.

## Workaround

Set `DOCKER_GID=0` in `.env` and recreate the backend (`docker compose up -d backend`). UID stays 1000, so the non-root invariant from #874 is preserved; only socket-group membership changes.

## Acceptance Criteria

- [ ] On macOS Docker Desktop, a fresh `./scripts/deploy/start.sh` brings the backend up able to list agents (no manual `.env` edit required)
- [ ] `ensure_docker_gid()` resolves the correct GID on Docker Desktop instead of returning early on non-Linux (e.g. detect the mounted socket's group, or default to `0` when the host is non-Linux and the socket is root-group-owned)
- [ ] `docs/migrations/NON_ROOT_CONTAINERS_2026-05.md` corrected — the "group_add is ignored on Docker Desktop, leave the default" claim is wrong; document `DOCKER_GID=0` for Docker Desktop
- [ ] `.env.example` notes the Docker Desktop value

## Technical Notes

- Affected files: `scripts/deploy/start.sh` (`ensure_docker_gid`), `docker-compose.yml` (`group_add: ${DOCKER_GID:-999}`), `docs/migrations/NON_ROOT_CONTAINERS_2026-05.md`, `.env.example`.
- Optional hardening: downgrade the swallowed-exception path in `list_all_agents_fast` to at least log a WARNING, so a Docker-socket access failure is observable instead of silently presenting as "No agents" (this silence is what made the regression hard to diagnose).
- **Regression from #874.** Distinct from #877 (docker-socket-proxy hardening) — that replaces direct socket access entirely for security reasons and does not address the `group_add` GID mismatch on Docker Desktop.

AndriiPasternak31:
## Summary

Fixes the #874 non-root regression where the backend (UID 1000) cannot reach `/var/run/docker.sock` on **macOS Docker Desktop / Colima / Rancher / rootless**, so `GET /api/agents` returns `[]` and the Agents page silently shows **"No agents"** (#1131, P1).

Root cause: `#874` assumed Docker Desktop *ignores* compose `group_add`, so `start.sh` returned early on non-Linux and left `DOCKER_GID` at the Debian default (`999`). Those runtimes actually present the socket **root-group-owned (GID 0)**, so UID 1000 is denied — and `list_all_agents_fast()` swallowed the `PermissionError`, making it silent.

## Changes

- **`scripts/deploy/start.sh`** — `ensure_docker_gid()` now **probes the GID a throwaway container sees on the mounted socket** as the PRIMARY detection path (correct on every runtime: Docker Desktop/Colima/rootless → `0`, native Linux → host docker GID). `getent group docker` is kept only as the offline Linux fallback; an explicit `DOCKER_GID=<n>` in `.env` (including `0`) always wins.
- **`src/backend/services/docker_service.py`** — convert the swallowed-exception `print()` to a **throttled `logger.warning()`** (≤1/60s on this hot path) so a denied socket is observable in Vector instead of silent.
- **Docs / `.env.example` / compose comments** — correct the "Docker Desktop ignores `group_add`" claim; document `DOCKER_GID=0` for Docker Desktop. Migration doc CI reference updated to `container-security.yml`.
- **CI** — new hermetic `verify-docker-gid-detection` job (shims `docker`/`getent`/`uname`, extracts the real `start.sh` functions, ~1s, no daemon). This is the **only** guard for the probe branch, since `verify-non-root` runs on a Linux runner where the socket is already `docker`-group-owned. Added `tests/deploy/**` to the workflow path filter.
- **Tests** — `tests/unit/test_docker_service_list_fast.py` locks the `[]` + `WARNING` contract; `tests/deploy/test_docker_gid.sh` locks GID-detection precedence (override > probe > getent fallback > warn).

## Acceptance criteria (#1131)

- [x] Fresh `./scripts/deploy/start.sh` on Docker Desktop brings the backend up able to list agents (no manual `.env` edit)
- [x] `ensure_docker_gid()` resolves the correct GID on Docker Desktop instead of returning early on non-Linux
- [x] `NON_ROOT_CONTAINERS_2026-05.md` corrected
- [x] `.env.example` documents the Docker Desktop value
- [x] (optional hardening) silent swallow path now logs a throttled WARNING

Non-root invariant (#17) preserved — UID stays 1000; only socket-group membership changes. Security review: `docs/security-reports/cso-2026-06-25-1131-diff.md` (0 secrets, 0 new findings).

Fixes #1131


vybe (approved):
Validated via /validate-pr: docker.sock GID probe fix (#1131 P1), security + packaging clean, CSO diff report included. Approving.

PR #1355, opened 2026-06-26
feat(connector): per-agent MCP connector — playbooks as tools (ent#46, PR1)

dolho:
## Summary

PR1 (backend + MCP) of the **per-agent MCP connector** (trinity-enterprise#46): expose a single agent as an MCP connector an end user adds to their AI client (Claude Code, Cursor, Claude Desktop) in one line, turning the agent's `user_invocable` playbooks (`.claude/skills/`) into MCP tools. The agent holds credentials server-side; only a scoped, revocable key reaches the client. A new client-sharing channel alongside Slack/Telegram/WhatsApp/VoIP/public links.

**Decisions (confirmed up front):** OSS core (the #847 entitlement seam can wrap it behind `requires_entitlement('mcp_connector')` later with a ~1-line diff); first PR is **backend + key + MCP tools** (Sharing-tab UI is PR2). The per-user SSO/RFC-8693 tier is a separate enterprise epic, out of scope.

## What's in this PR

**Backend**
- `agent_connectors` table (config: `enabled` + exposed-playbook allow-list) — dual-track migration (SQLite `agent_connectors_table` + Alembic `0006_agent_connectors`) + `db/tables.py`; cascade/rename registered in `AGENT_REFS`.
- `mcp_api_keys.scope='connector'`: a separate, owner-bound, SHA-256-hashed, independently-revocable key with **regenerate** (revoke-old → mint-new) so a leaked copy-paste snippet is cut off. Distinct from the auto-minted `scope='agent'` key.
- `routers/connectors.py`: owner CRUD (`GET` status / `PUT` config / `POST` key / `DELETE` key) + connector-key-readable `GET /connector/playbooks`. `services/connector_service.py` (pure) builds **per-client copy-paste snippets with the key pre-embedded** (Claude Code `claude mcp add` + `.mcp.json`, Cursor `mcpServers`, Claude Desktop) and resolves exposed = allow-list ∩ `user_invocable` (a `user_invocable:false` playbook is **never** exposed). Raw key returned **once**.
- **Auth boundary (security-critical):** a connector key resolves to the owner but is consumption-only — `_enforce_connector_scope` fences it to its bound agent on read/chat and refuses owner ops (`OwnedAgent*` → 403); `_reject_connector_principal` blocks every role-gated endpoint (`require_role`/`require_admin`). A leaked snippet can chat its one agent and nothing else.

**MCP server**
[...]

Issue #995, opened 2026-06-01
Enterprise: User & Organization Management (Org/Team + advanced RBAC) on the #847 seam


dolho:
## Summary

Add **Enterprise User Management** — multi-tenant Organizations/Teams plus advanced RBAC — as the second concrete enterprise module on the #847 entitlement seam (after the audit dashboard, #941). Entitlement id: **`user_management`**.

Today OSS ships a flat 4-tier role model (`user < operator < creator < admin`, ROLE-001) with admin-only per-user role CRUD and a single user namespace. Enterprise customers need org/team scoping, custom roles, and a permission matrix beyond the 4 built-ins (called out as a reserved-but-unbuilt seam in `docs/planning/OSS_ENTERPRISE_SPLIT_RESEARCH.md` and requirements §35).

## Open-core split (private-first)

Per the enterprise architecture (`docs/planning/ENTERPRISE_ARCHITECTURE.md`, `feature-flows/enterprise-modules.md`) **and** the explicit directive to keep enterprise logic in the private repo, the split is:

- **Private — `Abilityai/trinity-enterprise` submodule** (`src/backend/enterprise/backend/user_management/`): owns *everything* with IP — its own private tables (created idempotently on `register_enterprise`, via `from db.connection import get_db_connection`), Pydantic models, db ops, service/business logic, and the FastAPI router gated by `requires_entitlement("user_management")`. Registers the module via `entitlement_service.register_module("user_management")`.
- **Public — `Abilityai/trinity` → `dev`**: only the unavoidable seam. The enterprise Vue view ships in the OSS bundle (documented: enterprise Vue carries no algorithmic IP; gated purely by the server `GET /api/settings/feature-flags` → `enterprise_features` list) + nav/route + store getter. **No** public schema/model/router changes — the entitlement plumbing (`requires_entitlement`, `list_entitled_features`, feature-flag exposure) already exists from #847.

## Phases

**Phase 1a — Organizations + membership (this issue's starting slice)**
- [ ] Private tables: `enterprise_organizations`, `enterprise_org_members` (created on registration).
- [ ] CRUD orgs (create/list/get/update/delete) + add/remove/list members, admin + `requires_entitlement("user_management")` gated.
- [ ] `register_enterprise()` mounts the router and calls `register_module("user_management")`.
- [ ] Public: gated Enterprise → User Management Vue view + nav/route, hidden unless `user_management` entitled.
[...]

Issue #918, opened 2026-05-23
feat: agent-reported structured reports via MCP + dashboard display


vybe:
## Summary

Give agents a structured way to report telemetry and arbitrary domain reports back to Trinity via an MCP tool. Persist them in SQLite and surface them on a dashboard so users can see what each agent is producing without reading chat transcripts.

## Context

Today agents only emit unstructured text (chat, logs) and a fixed set of metrics (cost, context, tool counts). Anything domain-specific — leads found, emails sent, deals progressed, KPI snapshots, weekly summaries — is buried in chat history. We need a generic "agent report" primitive that's typed enough to display but flexible enough to fit any agent.

## Proposed Shape

**MCP tool** (new):
```
report(report_type: str, title: str, payload: dict, schema_version?: int, period_start?: str, period_end?: str)
```
- `report_type` namespaced (e.g. `recon.weekly_summary`, `prospector.leads_found`, `ops.daily_health`, `custom.*`)
- `payload` arbitrary JSON
- Resolves agent + user from MCP auth context (no spoofing)

**Storage**: new `agent_reports` table (id, agent_name, user_id, report_type, title, payload JSON, schema_version, period_start/end, created_at). Index on `(agent_name, created_at DESC)` and `(report_type, created_at DESC)`.

**Backend**: `routers/reports.py` — list/get/delete with the standard owner/sharing/admin gate. WebSocket `agent_report` broadcast for live updates.

**Frontend**: a "Reports" tab on AgentDetail + a fleet-wide Reports view on the dashboard. Generic renderer that knows a handful of payload shapes (table, KPI tiles, markdown, timeline) chosen by `report_type` prefix or a `display_hint` field.

## Example report types

- `prospector.weekly_leads` → table of leads with score
- `recon.competitor_changes` → timeline of detected changes
- `ops.fleet_health` → KPI tiles (uptime, errors, cost)
- `kb-agent.coherence` → markdown summary + counters
- `chief-of-staff.daily_brief` → markdown
- `custom.*` → free-form, rendered as JSON viewer

## Acceptance Criteria

- [ ] `agent_reports` table + schema migration
- [ ] MCP tool `report` (and matching backend endpoint) with auth-context resolution
- [ ] `GET /api/agents/{name}/reports` and `GET /api/reports` (fleet-wide, access-controlled)
- [ ] WebSocket `agent_report` broadcast
- [ ] AgentDetail "Reports" tab with generic + typed renderers
[...]

PR #1356, opened 2026-06-26
feat(auth): connector-scope MCP key primitive + connector MCP tools (ent#46 OSS slice)

dolho:
## Summary

The **OSS, edition-agnostic half** of the per-agent MCP connector (trinity-enterprise#46). The connector *feature* — config table, key minting, snippet generation, Sharing-tab UI — is an **entitled module in the private `trinity-enterprise` repo** (separate PR). This PR ships only what core must own, per ent#45 (paid-module schema/implementation stays private) and the open-core "core-primitive + enterprise-knob" pattern (like `users.suspended_at`, #995).

## What's here

**Security enforcement (backend, edition-agnostic):**
- Core recognizes a `scope='connector'` MCP key as a **consumption-only** principal (`User.connector_agent`).
- `_enforce_connector_scope` fences it to its **one bound agent** for read/chat and refuses owner operations; `_reject_connector_principal` blocks every role-gated endpoint (`require_role`/`require_admin`).
- **Inert** until an entitled module mints a connector key — exactly like OSS shipping `suspended_at` enforcement without the setter.

**MCP server plumbing (public, thin proxy, no IP):**
- `scope='connector'` in the auth context; `connector.ts` tools `list_playbooks`/`run_playbook`/`ask` gated by FastMCP `canAccess` so connector keys see **ONLY** these and operator keys never do. Agent comes from the bound key, never a tool arg.
- **Dormant** without an entitled backend serving the `/connector` endpoints (the MCP server is public infrastructure with no private path — thin proxies ship in OSS gated by the entitled backend, mirroring the gated-Vue-bundle precedent).

## Not here (→ private `trinity-enterprise` PR)
`enterprise_connectors` table (private migration runner), the owner CRUD router gated by `requires_entitlement('mcp_connector')`, connector-key minting, per-client snippet generation, and the UI.

## Testing
- `tests/unit/test_connector_auth.py` — 6 (fence, role-gate, model default).
- `src/mcp-server/src/connector.test.ts` — 5; full MCP suite + `tsc` clean.
- auth/dependency regression — 168 green.

## Background
This supersedes the closed #1355 (which wrongly put the whole feature in OSS). Re-split into this OSS primitive + a private enterprise module after re-scoping it as a paid feature.

Related to Abilityai/trinity-enterprise#46


dolho:
## Review — OSS connector primitive (self-review, found 2 real issues)

Reviewed the diff for correctness, the security boundary, and re-split completeness. The fence + MCP gating are sound, but two issues — one I'd call **blocking** — slipped through the re-split from the closed #1355.

### 🔴 Blocking — agent delete/rename leaves connector keys behind
`db/agent_cleanup.py` `AGENT_REFS` only cleans `mcp_api_keys` rows with `extra_filter="scope = 'agent'"`. Connector keys are `scope='connector'`, so:
- **Delete:** the connector key row orphans (stays `is_active=1`). If the agent name is later reused after hard-purge, a leaked old connector key would bind to a *different* agent of the same name — a key-reuse-across-identity hole.
- **Rename:** the key keeps the old `agent_name`; the fence then compares it against the new path and the connector silently breaks.

The original all-in-OSS PR had this ref; the split dropped it (this PR doesn't touch `agent_cleanup.py` at all). **Fix:** add a second, edition-agnostic ref — discloses nothing beyond a scope value, and is inert in OSS-only builds:
```python
AgentRef("mcp_api_keys", "agent_name", Policy.CASCADE, extra_filter="scope = 'connector'"),
```
(The enterprise `enterprise_connectors` config row has a parallel orphan problem — flagged on trinity-enterprise#51, since OSS can't reference that table.)

### 🟠 High — a leaked connector key isn't fully contained to one agent
`_enforce_connector_scope` is wired into the 4 agent path-deps (`get_authorized_agent[_by_name]`, `get_owned_agent[_by_name]`) and `_reject_connector_principal` into the role gates — good. But ~37 endpoints across routers do **inline** `db.can_user_access_agent(current_user.username, …)` / `can_user_share_agent(…)` instead of the fenced deps (e.g. `loops.py:184` GET loop-status, `messages.py`, `event_subscriptions.py`, …). A connector key resolves to the **owner**, so those inline checks pass for **any** of the owner's agents — bypassing the "consumption-only, one agent" fence on those specific endpoints.

Threat model: the connector key is the credential we ship in a copy-paste snippet and tell users to "regenerate if leaked" — so leak-containment is the point. Recommend a **request-level guard** (one ASGI dependency / middleware that, for a connector principal, allows only the bound agent's chat + `connector/playbooks`) rather than relying on every current and future endpoint adopting the fenced deps. The 4-dep approach is necessary but not sufficient.

### 🟡 Note — `run_playbook` invokes the skill via a natural-language chat message
`connector.ts` sends `Please run the "<name>" playbook` and relies on the agent's harness resolving the skill from prose — a soft contract, not a hard invocation. Fine for v1 (and the name is validated against the exposed set, so no injection), but worth a follow-up to use a deterministic invocation form.

Happy to apply the cleanup fix + draft the request-level guard if you want them on this branch.

dolho:
✅ **Both findings fixed** in c6f2355a.

- **Containment (high):** moved the fence to the single auth entry point — `get_current_user` now allowlists a connector principal to EXACTLY `POST /api/agents/{bound}/chat` + `GET /api/agents/{bound}/connector/playbooks` and 403s everything else. The ~37 inline-checked endpoints are no longer reachable by a leaked key regardless of which deps they use. (The per-dep `_enforce_connector_scope`/`_reject_connector_principal` stay as defense-in-depth.)
- **Cleanup (blocking):** added the `scope='connector'` `mcp_api_keys` AGENT_REF so connector keys are swept on delete and re-keyed on rename — closing the orphan/name-reuse hole.
- **Note (run_playbook):** left as documented v1 behavior (name is validated against the exposed set, so no injection) — a deterministic invocation form is a tracked follow-up, not a defect.

Tests: central-guard allow/deny matrix + cascade-delete-removes-connector-key + AGENT_REFS coverage; auth/cleanup regression 209 green.

vybe (approved):
Approving to unblock the enterprise submodule bump (ent#46 connector module is merged on `trinity-enterprise` main; this OSS primitive is its required seam — `resolve_mcp_url`, `register_agent_owned_table`, connector-scope key recognition).

**What I pushed to this branch** (merge-conflict resolution + CI fix only, not feature changes):
- Merged `dev` in; resolved 2 conflicts keep-both: `metadata.py` rename re-keys both the ent#46 EXTRA_AGENT_REFS tables **and** dev's #918 `agent_reports`; `server.ts` imports both connector + git tool groups.
- Fixed the failing `lint (sys.modules pollution)` (#762): the 3 bare `sys.modules.pop()` in `test_connector_auth.py` → `monkeypatch.delitem(..., raising=False)` (auto-restores).

**What I reviewed:** the auth enforcement in `dependencies.py` — central deny-by-default containment in `get_current_user` (connector key allowlisted to exactly its bound agent's chat + playbooks), `_enforce_connector_scope` (refuses owner ops / foreign agents), `_reject_connector_principal` (blocks role-gated endpoints). Sound defense-in-depth, mirrors the `suspended_at` core-primitive pattern; tests pin every path.

PR #1357, opened 2026-06-26
feat(ui): MCP connector panel on the Sharing tab (ent#46 PR2)

dolho:
## Summary

PR2 — the **gated-Vue UI** for the per-agent MCP connector (trinity-enterprise#46). Adds an **MCP connector** row to the Sharing tab's Channels section, shown **only** when the `mcp_connector` entitlement is present (`enterpriseStore.isEntitled('mcp_connector')`). Ships in the public bundle with no IP — the feature backend + gating live in the private `trinity-enterprise` module (Abilityai/trinity-enterprise#51); the OSS security primitive + MCP tools are Abilityai/trinity#1356.

## `ConnectorChannelPanel.vue`
- **Generate / regenerate / revoke** the scoped connector key.
- On generate, surfaces **per-client copy-paste setup** (Claude Code, Cursor, Claude Desktop) with the key pre-embedded + per-block copy buttons — shown **once** with a "key is shown only once" warning, dismissable.
- **Enable/disable** toggle + **exposed-playbook allow-list** editor (checkboxes over the agent's `user_invocable` playbooks; "expose all" clears the list to all).
- Calls the entitlement-gated `/api/agents/{name}/connector*` endpoints; reads the agent's playbooks via the OSS `/playbooks` endpoint (graceful "start the agent" hint when it's stopped).

## Gating / safety
- Hidden in OSS-only builds (no `mcp_connector` in `enterprise_features`). Safe to land independently of the enterprise PR — the row simply doesn't render until the module is mounted + entitled.
- Invariants: single `api` client (#7); component mirrors the existing `VoipChannelPanel` channel pattern.

## Verification
- Both SFCs compile (`vue/compiler-sfc`).
- Clean `npm ci` + production `vite build` — **1775 modules, ✓ built**.
- Browser-level interaction is covered by the `frontend-e2e` CI gate (not run locally — needs the full stack + an entitled backend).

Related to Abilityai/trinity-enterprise#46

### DIFF
--- a/src/frontend/src/components/ConnectorChannelPanel.vue
+++ b/src/frontend/src/components/ConnectorChannelPanel.vue
@@ -1,0 +1,249 @@
+<template>
+  <div>
+    <h3 class="text-lg font-medium text-gray-900 dark:text-white mb-2">MCP Connector</h3>
+    <p class="text-sm text-gray-500 dark:text-gray-400 mb-4">
+      Let someone add this agent to their AI client (Claude Code, Cursor, Claude Desktop) in one line.
+      The agent's playbooks become tools they can run; the agent keeps all credentials server-side.
+      Only a scoped, revocable key reaches the client.
+    </p>
+
+    <!-- Loading -->
+    <div v-if="loading" class="flex items-center gap-2 text-sm text-gray-500 dark:text-gray-400">
+      <svg class="animate-spin h-4 w-4" fill="none" viewBox="0 0 24 24">
+        <circle class="opacity-25" cx="12" cy="12" r="10" stroke="currentColor" stroke-width="4"></circle>
+        <path class="opacity-75" fill="currentColor" d="M4 12a8 8 0 018-8V0C5.373 0 0 5.373 0 12h4z"></path>
+      </svg>
+      Loading…
+    </div>
+
+    <div v-else-if="accessDenied" class="text-sm text-gray-500 dark:text-gray-400">
+      Only the agent owner can manage the connector.
+    </div>
+
+    <div v-else class="space-y-5">
+      <!-- No key yet -->
+      <div v-if="!status.has_key" class="p-4 bg-gray-50 dark:bg-gray-900/50 rounded-lg border border-gray-200 dark:border-gray-700">
+        <p class="text-sm text-gray-700 dark:text-gray-200 mb-3">
+          No connector key yet. Generate one to get copy-paste setup instructions for each client.
+        </p>
+        <button
+          type="button"
+          @click="regenerateKey"
+          :disabled="busy"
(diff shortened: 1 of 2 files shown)


dolho:
## Review — connector Sharing-tab UI (self-review)

Gating, the once-only secret handling, and the build are sound. Only low-severity items.

### 🟡 Low — stale allow-list names linger
`syncAllowListState` seeds `selected` from `status.exposed_playbooks`; if a stored name is no longer in the agent's live `user_invocable` playbooks it won't render a checkbox, but it stays in `selected` and is re-sent on the next `saveAllowList` (harmless — the backend intersects with the live set, so it's inert — but the stored list slowly accumulates dead names). Consider pruning `selected` to `availablePlaybooks` on load.

### 🟡 Low — `saveAllowList` fires per checkbox toggle
Every `@change` does a PUT. Fine functionally, but rapid edits = a burst of requests and the last-write-wins race is benign only because each PUT sends the full set. A debounce or an explicit "Save" would be gentler. Not blocking.

### 🟡 Note — secret is rendered into the DOM
`freshSnippets` puts the raw key in a `<pre>` (standard for a "shown once" secret + copy button). It's in-memory only and dismissable — acceptable, matching how MCP keys are surfaced elsewhere. Worth a mental note that a screen-share during setup exposes it (inherent to copy-paste onboarding).

### ✅ Verified good
- Hidden unless `enterpriseStore.isEntitled('mcp_connector')`; endpoints are also gated server-side (defense in depth).
- `exposeAll` ↔ checkbox state derived correctly from `exposed_playbooks == null`; checkboxes disabled while "expose all".
- Stopped-agent playbook fetch degrades to a "start the agent" hint (503), not a hard error.
- Clean `npm ci` + production vite build green (1775 modules); single `api` client (Invariant #7); mirrors the existing `VoipChannelPanel` pattern.

### Cross-PR
The blocking issues are upstream of the UI (key cleanup on trinity#1356, config-row orphan on trinity-enterprise#51). The UI's "generate key if `!has_key`" flow actually *masks* the orphan-config case gracefully (stale config shows but with no key until regenerated), so no UI change is needed for those.

dolho:
✅ **Both UI findings fixed** in f9e5b25e.

- **Stale allow-list names:** after the live playbooks load, `selected` is pruned to current user_invocable names — dead entries no longer linger or get re-sent.
- **Per-toggle PUTs:** per-playbook checkbox changes are now debounced (400ms, full-set last-write-wins); the "expose all" toggle still saves immediately.
- Secret-in-DOM: left as-is (documented once-only display + copy button, in-memory).

SFC compiles clean.

vybe (approved):
/validate-pr: APPROVE. Entitlement-gated MCP-connector Sharing-tab UI (ent#46 PR2); hidden in OSS-only builds. Self-review findings already fixed (f9e5b25e); prod vite build green; single api client (#7). Security clean, mergeable, all checks green. Stale nightly 'merge conflict' comment is outdated — live status MERGEABLE.

PR #1364, opened 2026-06-27
feat(reliability): retry triggering execution after SUB-003 auto-switch (#792)

### LINKED ISSUE #792: feature/sub-003: retry triggering execution after successful auto-switch
## Context

When the SUB-003 subscription auto-switch fires on a `429 Too Many Requests` or auth-class failure mid-execution, the triggering execution row is marked FAILED (see `task_execution_service.py:600-660`). #606 closed as "by design" for the paths that exist today, but one path is genuinely uncovered: **one-shot scheduled executions** (ad-hoc manual triggers, not a recurring cron). Neither of the recovery patterns below applies, and the user sees the FAILED row without any retry.

## What already exists (do not redo)

- **Interactive chat — already retries.** `src/backend/routers/chat.py:466-493` returns `retry_after: 15` + an `auto_switch` payload to the frontend when a subscription-mode agent hits 429/503 and a successful switch occurred. The Vue chat UI handles the client-driven retry. Interactive callers do not need this issue.
- **Recurring scheduled executions — recover at next tick.** A schedule with a real cron expression fires again on the next tick using the newly assigned subscription. Transient FAILED rows are recovered by the schedule's natural recurrence (not by the platform replaying the failed execution).

## Gap this issue addresses

One-shot triggers — `POST /api/agents/{name}/schedules/{id}/trigger`, webhook fires on schedules with `cron_expression` that doesn't recur in any meaningful window, manual `trigger_agent_schedule` MCP calls — produce a single execution. If the very first attempt hits the 429 that fires the auto-switch, the execution is marked FAILED and the user has to manually retrigger after the (typically successful) switch.

## Design surface

Each of these is a non-trivial decision; reviewers should pick a path explicitly before implementation.

- **Idempotency model.** Is the agent's executed request idempotent under retry? If the original message kicked off external side effects before hitting 429 (e.g. a Slack post, a file write, an MCP tool call), the retry must not double-fire. How is this guaranteed — agent-level dedup keys, platform-level execution-id reuse, or scoping the retry to executions that didn't reach a tool-call boundary?
- **Execution state transitions.** New status (e.g. `auto_retrying`), or reuse `pending_retry` from RETRY-001 / #476? How does the retried execution appear in `agent_activities` and `chat_messages` — one row or two? Linked via FK?
- **Retry budget.** One retry per execution, or N? Exponential backoff? Hard cap to prevent infinite cascades if the new subscription is also exhausted?
- **Capacity-slot lifecycle.** The original execution holds a `CapacityManager` slot. Release-and-reacquire (race-prone against other queued tasks) or hold-through (occupies capacity longer than necessary)?
- **Subscription usage attribution.** Cost / token usage for the retried execution — attribute to the new subscription or the old one that triggered the 429? Affects fairness for shared subscription pools.
- **Failure attribution.** Does the *original* 429 still get recorded as the cause of the FAILED-then-recovered execution, or is the audit trail rewritten?
- **Interaction with the 2h skip-list.** If the new subscription is *also* rate-limited mid-retry (cascade auth/quota failure), do we cascade-switch again? Hard-fail after N? The 2h skip-list in `select_best_alternative_subscription` already guards against ping-pong but only at the selection layer — retry adds a new dimension.
- **Concurrency / per-agent lock.** Today there is no per-agent switch/restart lock around `assign_subscription_to_agent` + `container_stop` + `start_agent_internal` (raised by the #606 Codex eng review). If two 429s for the same agent fire near-simultaneously, retries could trample each other. Lock as part of this issue, or punt to a separate concurrency-hardening issue?

## Out of scope

- Interactive chat retry UX (already exists at `routers/chat.py:466-493`).
- Recurring scheduled execution recovery (already covered by cron recurrence).
- The chain-level guard against spurious `.credentials.enc` import noise (covered by #612 + the regression test pinning the `_restart_agent` chain in the #606 close PR).

## References

- #606 — the parent bug, closing with this issue cited as the in-flight follow-up.
- #612 — the production fix this builds on (subscription-mode short-circuit at `lifecycle.py:155`).
- `src/backend/routers/chat.py:466-493` — existing interactive retry hint.
- `src/backend/services/task_execution_service.py:600-660` — the 429/auth exception block that fire-and-forgets auto-switch and marks the execution FAILED.
- `src/backend/services/subscription_auto_switch.py:145-259` — `_perform_auto_switch` / `_restart_agent`.

AndriiPasternak31:
## Summary

Extends **SUB-003 subscription auto-switch** so that when a switch fires
mid-execution, the **triggering execution is retried once** and lands `SUCCESS`
instead of `FAILED`.

Today a switch mid-turn marks the row `FAILED`. Three recovery patterns already
exist — interactive chat retries client-side (`routers/chat.py`) and recurring
cron recovers on the next tick — but **one-shot triggers** had no recovery:

- manual `POST …/schedules/{id}/trigger`
- webhook
- MCP `trigger_agent_schedule`

These produced a single dead `FAILED` execution. This PR closes that gap.

## How it works

`TaskExecutionService.execute_task` now intercepts a returned 429/auth response
**pre-`raise_for_status`** — `agent_post_with_retry` *returns* the response
rather than raising, so the interception sits adjacent to the #678 reader-race
502 block and falls through to the single shared success-parse path by
reassigning `response` (no duplicated finalize logic). When SUB-003 reports
`switched=True`, it re-issues the turn **once** with the **same `execution_id`**.

New module helpers:
- `classify_switch_failure(response)` — full SUB-003 surface (`429`→rate_limit;
  `503/401/403/402` or `is_auth_failure` body→auth; else `None`), so "any
  switch-success retries" actually holds, not just status codes.
- `_extract_agent_error(response, fallback)` — shared body→`(msg, metadata)`
  extractor, also adopted by the `except httpx.HTTPError` handler so both paths
  read the body identically.
- `_salvage_attempt_cost(metadata)` — rolls first-attempt cost into
  `previous_attempt_cost` (#678 R2).

Key decisions:
- **Dedicated one-shot guard** `subscription_switch_attempted` (NOT `retry_count`,
  which the #678 retry owns) so the two retry reasons never suppress each other;
  the same flag gates the `except`-handler switch so a cascade (retry still
  failing) never double-switches.
- **Retry is the readiness probe** — short `_SWITCH_RETRY_DELAY_S` pre-delay
  only; no circuit-aware `/health` poll (would poison the transport breaker on
  cold start) and no trust in `restart_result`'s string status. A still-failing
  retry writes `FAILED`, as before.
- **Budget bounded** — retry timeout capped to the **remaining** original budget
  so a post-long-run 429 can't balloon wall-clock/slot time.
- **Idempotency** — same `execution_id` ⇒ #1084 `effect_guard` dedups wired
  outbound sinks; residual double-fire risk for arbitrary MCP tool calls is the
  same the #678 retry already accepts.

## Out of scope / follow-ups

- #1083 fire-and-forget async path (`DISPATCH_ASYNC`, default OFF) routes 429s
  through the result-callback, bypassing this sync path.
- A concurrent switch-lock *loser* (gets `None` from `handle_subscription_failure`)
  does not retry.

## Testing

`tests/unit/test_792_subscription_retry.py` — 11 cases (incl. parametrized
503/401/403/402 auth surface, #678↔#792 interplay, cascade single-switch,
bounded retry timeout). All green:

```
11 passed in 0.12s
```

## Docs

- `requirements.md` §20 + `docs/requirements/SUB-003-subscription-auto-switch.md`
  (acceptance criteria + history)
- `architecture.md` service catalog + `feature-flows/task-execution-service.md`
  + feature-flows index row
- `/cso --diff` report committed under `docs/security-reports/` — **CLEAN**, no
  findings above the gate.

Closes #792


vybe (approved):
Approved via /validate-pr: clean SUB-003 retry. Issue link+closing keyword ✓, no schema/config changes, security clean, named regression test (test_792_subscription_retry.py) covers the non-happy retry-after-switch path, SUB-003 stays inline-only (async #1083 bypass preserved), all 18 checks green.

PR #1371, opened 2026-06-29
feat(onboarding): minimalistic guided first-agent wizard

dolho:
## What

Gives a fresh self-hoster a **guided path to first value** instead of an empty dashboard. The wizard asks **one question** — "what do you want your first agent to do?" — maps the intent to a starter template, and deploys a **tailored, ready-to-chat agent** in a couple of clicks.

This is the first slice of the trinity-enterprise#52 onboarding epic: a single, self-contained PR that lets a person start their own AI agent in minutes with guidance. Subsequent slices (mandatory Claude-auth setup gate, ambient checklist, JIT GitHub PAT) spin off from the spec.

## Design

- **Teach by doing, not reading.** No mandatory click-through tour — the exact thing the "don't make me read" user rage-quits.
- **One question → a running agent.** Intent cards map to the real local starter templates (`scout`/`sage`/`scribe`) or a blank Claude Code agent.
- **Soft-guide the one hard gate.** Claude auth is surfaced as a *non-blocking* amber hint linking to Settings → Integrations; the user can deploy now and add it after. Everything else (e.g. GitHub PAT) is deferred just-in-time.
- **Never nags.** Auto-opens once for a fresh, empty install; dismissal is remembered in `localStorage`.

## Changes

**Frontend**
- `components/OnboardingWizard.vue` — purpose picker → `POST /api/agents` → "Open chat" (`/agents/:name?tab=chat`). Verifies mapped templates exist against `/api/templates`; missing → blank-agent fallback so deploy never 404s.
- `views/Dashboard.vue` — mounts the wizard, auto-opens via `maybeAutoOpenOnboarding()` (gated on `!isFleetLoading && agents.length === 0 && !dismissed`); empty-state "Get started" button re-opens it.
- `stores/sessions.js` — caches `claudeAuthConfigured` from feature-flags.

**Backend**
- `routers/settings.py` — `GET /api/settings/feature-flags` gains `claude_auth_configured` = `bool(anthropic_api_key)` (DB or env) **or** any registered subscription. A boolean only; never the key itself.

**Removed**
- Dead `OnboardingChecklist.vue` + `useOnboarding.js` (orphaned, modeled on the removed Process Engine — mounted nowhere, linked to non-existent `/processes/*` routes).

**Docs**
- `docs/memory/feature-flows/onboarding-wizard.md` — flow spec + follow-up build issues.

## Test

- Backend flag verified live: `GET /api/settings/feature-flags` → `claude_auth_configured: true` on this dev instance.
- Frontend: HMR-clean; `OnboardingWizard.vue` transforms without errors via the dev server.
- Settings deep-link `?tab=integrations` resolves to the Integrations tab.
- Note: the wizard auto-opens only on an **empty** fleet, so it isn't reachable from the UI on an instance that already has agents.

Related to Abilityai/trinity-enterprise#52


dolho:
Screenshot from 2026-06-29 11-48-34

Screenshot from 2026-06-29 11-48-42

src/frontend/src/components/OnboardingWizard.vue
```diff
@@ -0,0 +1,224 @@
...
+  <!-- create step: hand off to the real creation modal -->
+  <CreateAgentModal
+    v-if="step === 'create'"
+    :initial-template="selectedTemplate"
+    @created="onCreated"
+    @close="onModalClose"
+  />
+
+  <!-- intro + credential: the wizard's own guidance card -->
+  <div v-else class="fixed inset-0 z-50 overflow-y-auto">
```
dolho: **[medium] Modal a11y gap.** This overlay has no `role="dialog"`/`aria-modal="true"`, no focus trap, no Escape-to-close, and no body scroll-lock. The repo just standardized exactly these for `ChannelConfigDialog` (#19: "focus mgmt, trap, scroll lock"), so this regresses against that bar. At minimum: add `role="dialog" aria-modal="true"` + an `aria-labelledby` pointing at the `<h2>`, an Escape `@keydown`, and focus the first control on open. Reusing the existing dialog pattern/composable would keep it consistent.

src/backend/routers/settings.py
```diff
@@ -152,6 +152,13 @@ async def get_public_feature_flags(
         # failure controls are armed during a soak. NOT a UI surface.
         "redelivery_governor_enabled": REDELIVERY_GOVERNOR_ENABLED,
         "platform_default_model": settings_service.get_platform_default_model(),
+        # Onboarding (trinity-enterprise#52) — is Claude auth configured at all?
+        # Trinity agents can't think without it, so the first-run wizard uses
+        # this to surface the one hard setup gate. True if a platform-wide
+        # Anthropic key exists (DB or env) OR any Claude subscription is
+        # registered. Non-sensitive: a boolean, never the key itself.
+        "claude_auth_configured": bool(settings_service.get_anthropic_api_key())
+        or bool(db.list_subscriptions()),
```
dolho: **[low-med] Hot-path full table read.** `/api/settings/feature-flags` is fetched on nearly every page load (NavBar, Dashboard, AgentDetail). When no Anthropic key is set, this calls `db.list_subscriptions()` — materializing **every** subscription row (decrypting credentials in `SubscriptionWithAgents`? at least full hydration) just to test existence. Prefer a cheap `COUNT(*) > 0` / `EXISTS` helper (e.g. `db.has_any_subscription()`). The `or` short-circuits when a key exists, so this only bites key-less installs — which is exactly the onboarding state this flag targets, so it will be hit.

src/frontend/src/components/OnboardingWizard.vue
```diff
@@ -0,0 +1,224 @@
...
+
+const step = ref('intro')          // intro | create | credential
+const selectedTemplate = ref('')
+const createdName = ref('')
+
+onMounted(async () => {
+  // Confirm mapped local templates exist; fall back to blank so the prefilled
+  // CreateAgentModal never points at a missing template.
+  try {
+    const r = await axios.get('/api/templates', { headers: authStore.authHeader })
```
dolho: **[low] Duplicate templates fetch.** The wizard fetches `/api/templates` here only to downgrade a missing local template to blank, and then `CreateAgentModal` fetches `/api/templates` again on its own mount in the `create` step — two round-trips for the same data within one flow. Minor; acceptable if you prefer the isolation, but worth a note. The fallback also mutates `p.template` in place on the `ref` array, which works but is a little surprising.

src/frontend/src/components/OnboardingWizard.vue
dolho: Fixed in 872afe0f — added `role=dialog`/`aria-modal`/`aria-labelledby`, Escape-to-close, a Tab focus trap, focus-on-open, and body scroll-lock.

src/backend/routers/settings.py
dolho: Fixed in 872afe0f — added `db.has_any_subscription()` (COUNT existence check); the flag now uses it instead of `bool(db.list_subscriptions())`, so no rows are materialized/decrypted on this hot path.

src/frontend/src/components/OnboardingWizard.vue
dolho: Fixed in 872afe0f — dropped the wizard's `/api/templates` fetch; the missing-template→blank fallback now lives in `CreateAgentModal` (which already fetches), so the create flow makes one request.

vybe (approved):
Validated via /validate-pr: backend flag is boolean-only (no key leak) with a parameterized COUNT existence check; frontend reuses the real CreateAgentModal (no cloned API path, Invariant #7) and safely net-removes dead OnboardingChecklist/useOnboarding (green build confirms no remaining importers). All 18 required checks green. Non-blocking follow-ups: cross-tracker trinity-enterprise#52 needs a manual status-in-dev bump; feature-flows.md index not updated for the new flow doc.

PR #1372, opened 2026-06-29
fix(agent): scope headless stall-watchdog to mcp__ + configurable limit (#1369)

### LINKED ISSUE #1369: bug: headless stall-watchdog (300s) kills executions with legitimately long-running tools (sub-agents / internal MCP calls)
## Summary

The headless executor's per-tool **stall watchdog** (`_STALL_LIMIT_S = 300.0` in `docker/base-image/agent_server/services/headless_executor.py`) kills the **entire execution** when any single `tool_use` stays open with no matching `tool_result` for >300s. It was introduced (#970 / PR #973) to catch a *hung stdio MCP `tools/call`* that never returns — but it cannot distinguish "hung" from "legitimately slow." Common autonomous patterns produce no `tool_result` for the full duration of the call and therefore trip it at 5 minutes:

- an agent invoking a long-running **sub-agent** (Claude `Task` tool),
- a slow **internal MCP tool** (`chat_with_agent` to another agent's multi-minute turn, `fan_out`, a long web/data tool),
- a long `Bash` step (build, install, big clone, ETL, poll).

The whole turn is killed with:

```
Killed: tool '<tool>' produced no output for 300s (stall watchdog)
```

persisted as a 504 / `terminal_reason=stall_no_output`. In practice operators see "a lot of agents failing" with this message.

## Context

Reported use case: agents legitimately call long-running sub-agents (Claude sub-agents / the `Task` tool) or internal MCP tools that run for minutes without returning an interim result, and the watchdog kills the execution mid-flight.

Why it bites so often:

- The watchdog measures the `tool_use → tool_result` gap — i.e. the **entire** tool duration. In stream-json mode there is no partial `tool_result`, so a 6-minute sub-agent / MCP / Bash call is indistinguishable from a wedged one.
- **Interim stdout does not reset it** (it's result-based, not output-based), so "make the tool print progress" does not help.
- The 300s ceiling is **hardcoded, not configurable**, and is far below the agent's own `execution_timeout_seconds` (default 3600s, up to 7200s). Bumping the execution timeout does nothing — documented in #1094.
- **Scope:** the headless / autonomous path only (`POST /api/task` → `execute_headless_task`, plus the #1083 async-dispatch callback) — schedules, loops, fan-out, webhooks, MCP `chat_with_agent`. Interactive Chat (`claude_code.py`) is unaffected (it uses only a plain wall-clock cap).

History / why this is still unfixed:

- #970 / PR #973 introduced the watchdog as the fix for the worse "phantom 2h hang" (leaked reader thread holding the stdout pipe). The watchdog is **load-bearing** — removing it resurrects that bug.
- #1094 / PR #1147 fixed only the *mislabeling* (added the `stall_no_output` reason); the false-positive behavior is unchanged.
- PR #980 (a competing #970 fix built on a wrong diagnosis) was closed as superseded; #1025 / PR #1078 salvaged its robustness bits.
- **No issue has ever addressed the false-positive itself** — that 300s is too aggressive and applies to *all* tools, including legitimately long sub-agent / MCP / Bash calls.

Relationship to the target architecture: this is structurally fixed by the pull / work-stealing model in `docs/planning/TARGET_ARCHITECTURE.md` (§Agent Runtime → "Pull worker pool"; §"Failure Isolation Without a Dispatch Breaker"). Under pull, a hung or slow turn consumes an *agent worker* (not a backend resource), recovery is lease-expiry re-delivery, and there is no per-tool stall kill at all — so the migration (Epic #1045 / umbrella #1081, private tracker) would retire this watchdog. **But the bug exists today** and the pull migration is not imminent, so an interim fix is warranted.

## Acceptance Criteria

- [ ] A legitimately long-running tool call (sub-agent / internal MCP tool / `Bash`) that runs >300s while making progress no longer kills the execution.
- [ ] The #970 genuine-hang protection is preserved: a wedged stdio MCP `tools/call` that never returns is still bounded (does not wedge the worker for the full execution budget).
- [ ] The stall limit is operator-configurable (e.g. an `AGENT_TOOL_STALL_LIMIT_S` env var) with a sane default; unset/`0` behavior documented.
- [ ] Built-in tools that legitimately run long (at minimum `Bash`) are not subject to the same aggressive 300s kill as opaque MCP stdio calls — e.g. scope the watchdog to `mcp__*`, or scale the limit to the execution timeout.
- [ ] No regression to the #970 phantom-2h-hang case (covered by a test).
- [ ] Stall-kill error wording / `terminal_reason` stays distinct and accurate (preserve #1094).

## Technical Notes

- Watchdog: `_open_tool_exceeding()` + the `process.wait` poll loop in `docker/base-image/agent_server/services/headless_executor.py` (constant `_STALL_LIMIT_S = 300.0`); terminal reason built in `_timeout_504_detail(...)` → `stall_no_output`.
- Candidate approaches (not prescriptive):
  - **Scope the watchdog to opaque MCP / stdio tool calls (`mcp__*`)**, exempting built-ins like `Bash`/`Read` that return on their own and are already bounded by `execution_timeout_seconds`. Closest to the original #970 intent (a stdio child that *never* returns).
  - Make `_STALL_LIMIT_S` env-configurable and/or a function of the agent's `execution_timeout_seconds` rather than a flat 300s.
  - Optionally allow a per-template opt-out for known long-runner orchestrators.
- Mind the agent-server ↔ backend mirror (Invariant #5) and the existing reader-race / drain hardening (#1025, #1201) when touching this path.
- Lineage: #970 / #973 (introduced) · #1094 / #1147 (label fix) · #1025 / #1078 (drain hardening) · #980 (closed / superseded).
- Superseding direction: the pull migration (#1081 / #1083) retires this watchdog entirely; this issue is the interim fix.

dolho:
## Summary
- The headless per-tool **stall watchdog** killed the whole execution when any `tool_use` had no `tool_result` for a flat hardcoded 300s — indistinguishable "hung" vs "legitimately slow", so long sub-agents (`Task`), internal MCP calls (`mcp__trinity__chat_with_agent`/`fan_out`), and long `Bash` steps died at 5 min (`terminal_reason=stall_no_output`).
- Now **scoped to `mcp__*` tools** (built-ins exempt — already bounded by `execution_timeout_seconds`) **and operator-configurable** via `AGENT_TOOL_STALL_LIMIT_S` (default raised 300s→1800s; `0`=disabled).
- #970 phantom-2h-hang protection preserved; #1094 `stall_no_output` reason preserved.

## Changes
- `docker/base-image/agent_server/services/headless_executor.py` — `_resolve_stall_limit_s()` (per-run, env-read, invalid→default); `_open_tool_exceeding` scoped to `mcp__*` + `(e.tool or "")` None/blank-safe; disable **gated at the call site** (passing `0` into the helper would fire instantly); kill wording threads the resolved limit (carried on `ctx.stall_limit_s`).
- `src/backend/services/agent_service/crud.py` + `lifecycle.py` — thread `AGENT_TOOL_STALL_LIMIT_S` from backend env into the container (set on create; set/clear on recreate, mirroring the `AGENT_GUARDRAILS` idiom).
- `tests/unit/test_headless_executor_970_timeout.py` + `test_1094_stall_termination_reason.py` — built-in exemption, mcp-only flagging, blank-name safety, env resolver (default/blank/valid/invalid/disabled), disable-path falls to budget, and the two existing tests migrated off the removed constant.
- `docs/memory/requirements.md` — documented under TIMEOUT-001.

## Acceptance criteria
- [x] Legit long-running tool (sub-agent/MCP/`Bash`) >300s with progress no longer killed
- [x] #970 genuine-hang protection preserved (hung stdio `mcp__` `tools/call` still bounded; test)
- [x] Stall limit operator-configurable (`AGENT_TOOL_STALL_LIMIT_S`); `0`/unset documented
- [x] Built-ins (≥`Bash`) not subject to the aggressive MCP kill (scoped to `mcp__*`)
- [x] No regression to #970 (test); `stall_no_output` reason preserved (#1094)

## Test Plan
- [x] `pytest tests/unit/test_headless_executor_970_timeout.py tests/unit/test_1094_stall_termination_reason.py` → 20 passed; full headless suite 65 passed
- [ ] **Rollout**: agent-internal change — needs `./scripts/deploy/build-base-image.sh` + container **recreate** (not a plain restart; env baked at create). Default is *raised*, so no recreated agent regresses to a tighter kill; un-migrated agents keep 300s. Mixed-fleet safe (no backend↔agent protocol change).

## Tradeoff (called out)
Exempting built-ins means a genuinely hung `Bash`/`Task` — or a hang *inside* a `Task` sub-agent (surfaces as an open `Task` in the parent log) — now falls to the `execution_timeout` budget instead of the 300s kill. Accepted interim; the pull/work-stealing model (#1081/#1083) retires this watchdog entirely.

Related to #1369

Fixes #1369

### DIFF
--- a/docker/base-image/agent_server/services/headless_executor.py
+++ b/docker/base-image/agent_server/services/headless_executor.py
@@ -73,23 +73,66 @@ _JSONL_PERSIST_THRESHOLD_S = 600
 # timeout). The stall limit is intentionally generous — "tool running" is
 # not "tool stalled" — so only a genuinely wedged call trips it.
 _WAIT_POLL_S = 2.0
-_STALL_LIMIT_S = 300.0
+
+# #1369: default per-tool stall ceiling (was a flat hard-coded 300s, which
+# killed legitimately long internal MCP calls). Operator-configurable via
+# AGENT_TOOL_STALL_LIMIT_S; <=0 disables the watchdog (the execution_timeout
+# budget stays the backstop). The watchdog is also scoped to opaque MCP
+# (``mcp__*``) tools only — built-ins (Bash/Read/Task/sub-agents) return on
+# their own and are already bounded by execution_timeout_seconds, so a long
+# (progressing) built-in must never be killed here. A genuinely hung built-in
+# or a hang *inside* a Task sub-agent (which surfaces as an open ``Task`` in
+# the parent log) now falls to the execution_timeout backstop rather than this
+# 300s kill — the accepted interim tradeoff (see #1369).
+_STALL_LIMIT_DEFAULT_S = 1800.0
+
+
+def _resolve_stall_limit_s() -> float:
+    """Resolved per-run stall ceiling (seconds).
+
+    Read fresh per run (not a module constant) so it honors the container env
+    and is monkeypatchable in tests. ``AGENT_TOOL_STALL_LIMIT_S`` overrides the
+    default; a missing/blank value uses the default; a non-numeric value falls
+    back to the default with a warning (a typo must not brick the executor).
+    A value ``<= 0`` means *disabled* and is handled at the call site — it is
+    NOT passed into ``_open_tool_exceeding`` (where ``> 0`` would fire instantly).
+    """
+    raw = os.environ.get("AGENT_TOOL_STALL_LIMIT_S")
+    if raw is None or not raw.strip():
+        return _STALL_LIMIT_DEFAULT_S
+    try:
+        return float(raw)
+    except (TypeError, ValueError):
+        logger.warning(
+            "[Headless Task] Invalid AGENT_TOOL_STALL_LIMIT_S=%r — using default %.0fs",
+            raw, _STALL_LIMIT_DEFAULT_S,
+        )
+        return _STALL_LIMIT_DEFAULT_S
 
 
 def _open_tool_exceeding(ctx: HeadlessRunContext, limit_s: float) -> Optional[str]:
-    """Name of a tool_use open (no matching tool_result) for > ``limit_s``, else None.
+    """Name of an ``mcp__*`` tool_use open (no tool_result) for > ``limit_s``, else None.
 
     #970 stall-watchdog signal. Reuses the stream parser's existing
     bookkeeping — ``tool_start_times`` (set on every tool_use) and the
     ``execution_log`` tool_result entries that close them. A hung stdio
     MCP ``tools/call`` never emits a tool_result, so its tool_use stays
     open: the only externally-visible signal that claude is wedged.
+
+    #1369: only ``mcp__*`` tools are considered. Built-ins return on their
+    own and are bounded by execution_timeout, so they are exempt here. The
+    caller must only invoke this with ``limit_s > 0`` (a ``<= 0`` limit would
+    match every open tool).
     """
     log = list(ctx.execution_log)  # snapshot — reader thread mutates concurrently
     completed = {e.id for e in log if e.type == "tool_result"}
     now = datetime.now()
     for e in log:
         if e.type == "tool_use" and e.id not in completed:
+            # Scope to opaque MCP stdio calls; None-safe (a tool_use with no
+            # name must not crash the poll loop).
+            if not (e.tool or "").startswith("mcp__"):
+                continue
             started = ctx.tool_start_times.get(e.id)
             if started and (now - started).total_seconds() > limit_s:
                 return e.tool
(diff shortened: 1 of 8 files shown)


docker/base-image/agent_server/services/headless_executor.py
```diff
@@ -73,23 +73,66 @@
...
+    default; a missing/blank value uses the default; a non-numeric value falls
+    back to the default with a warning (a typo must not brick the executor).
+    A value ``<= 0`` means *disabled* and is handled at the call site — it is
+    NOT passed into ``_open_tool_exceeding`` (where ``> 0`` would fire instantly).
+    """
+    raw = os.environ.get("AGENT_TOOL_STALL_LIMIT_S")
+    if raw is None or not raw.strip():
+        return _STALL_LIMIT_DEFAULT_S
+    try:
+        return float(raw)
```
dolho: **[low] `float(raw)` accepts non-finite values.** `float("inf")`/`float("nan")`/`"1e9"` parse without error, so a fat-fingered env value slips past the `try/except` and the "only `<=0` disables" contract: `inf` makes the per-poll `(now-started) > inf` check always false (watchdog silently inert, but burns the check every 2s), and `nan` makes `nan > 0` false → silently disabled. Not dangerous, but a typo that *looks* like a large limit would quietly turn the #970 protection off. Consider validating finiteness, e.g. after the `float()`:

```python
val = float(raw)
if not math.isfinite(val):
    logger.warning("[Headless Task] Non-finite AGENT_TOOL_STALL_LIMIT_S=%r — using default %.0fs", raw, _STALL_LIMIT_DEFAULT_S)
    return _STALL_LIMIT_DEFAULT_S
return val
```
(`import math` at module top.) Optional — current behavior is benign, just opaque.

dolho: Fixed in 3e21303469804c65fc3cbce85c6a9743e72ea339 —  now rejects non-finite values via `math.isfinite` (falls back to the default with a warning); only a deliberate `<=0` disables. Test added (`inf`/`-inf`/`Infinity`/`nan`/`NaN` → default).

vybe (approved):
Validated via /validate-pr: issue link resolves, CI all green, security clean, packaging + tests adequate. Approving for merge to dev.

PR #1377, opened 2026-06-29
fix(session): reconcile severed long turns against server state — no false 'failed to send' (#1376)

### LINKED ISSUE #759: bug(session-tab): shows 'failed to send' on return navigation while execution is still running
## Summary

When a message is sent via the Session tab and the user navigates away to another page before the response completes, returning to the agent shows a \"failed to send message\" error. The execution is actually still running on the backend and will complete normally — the result may appear at the end — but the UI incorrectly surfaces a failure state mid-flight.

## Component

Frontend / Session Tab

## Priority

P2

## Steps to Reproduce

1. Open an agent's Session tab
2. Send a message that will take more than a few seconds to complete
3. While the response is still in progress, navigate to a different page (e.g. Agents list, another agent)
4. Navigate back to the original agent's Session tab
5. Observe: UI shows \"failed to send message\"
6. Wait: execution completes on the backend and result may still appear

## Expected Behavior

On return navigation, the Session tab should show the message as in-progress (e.g. a spinner or \"running\" indicator) and update to the completed result when the execution finishes. A \"failed\" state should only be shown if the backend actually returns an error.

## Actual Behavior

The Session tab shows \"failed to send message\" immediately on re-render after navigation, even though the backend execution is still running. If left open, the result may appear when the execution eventually completes, contradicting the failure message.

## Root Cause (suspected)

The Session tab likely tracks in-flight request state in local component state (e.g. a `sending` flag or a pending promise). When the component unmounts on navigation, that state is lost. On remount, the absence of a completed response is interpreted as a failure rather than as an in-progress execution. The fix likely requires checking execution status from the backend on mount (polling or the existing SSE stream) instead of relying solely on local component state.

## Suggested Fix

On Session tab mount, check if there is an active execution for the agent (via `/api/executions?status=running` or the existing SSE stream). If one is found for the current session, resume the in-progress indicator and stream the result rather than rendering a failed state.

## Workaround

Stay on the Session tab until the response completes. Do not navigate away while a message is being processed.

## Environment

- Trinity version: `7b95c46`
- Affected surface: Session tab (`/agents/{name}` → Session tab)

### LINKED ISSUE #1376: bug: Session-mode turn shows 'failed to send' on long-running work while the execution succeeds in the background
## Summary

In **Session mode** (the `--resume` chat surface), sending a message that triggers a long stretch of agent work surfaces a **"Failed to send message"** error in the UI even though the backend execution is running normally and **completes successfully** — the assistant reply lands in the DB and appears on a later refresh, contradicting the failure shown to the user.

This is the same *class* of false-failure as the previously-fixed #759 (`bug(session-tab): shows 'failed to send' on return navigation while execution is still running`), but for the **stay-on-page, long-running turn** path rather than the navigate-away-and-return path. Both existing mitigations are in place yet the symptom still reproduces, so this is filed as a distinct bug.

## Context

The Session-tab turn endpoint (`POST /api/agents/{name}/sessions/{id}/message`) is **synchronous** — the browser holds the HTTP request open for the full duration of the agent's turn. Two mitigations already exist:

1. **HTTP-timeout ceiling match** — the axios call in `src/frontend/src/stores/sessions.js` sets `timeout: 7260000` (~121 min) to match the backend execution cap (TIMEOUT-001, 7200s + slack). The inline comment there describes exactly this symptom as the *old* behavior the ceiling was meant to remove.
2. **Reattach on return navigation (#759)** — a Redis in-flight sentinel (`session_inflight:{session_id}`) surfaces as `turn_in_progress=true` on the session GET; the frontend `startPolling()` reattaches and shows a "running" indicator instead of a failure.

Despite both, users with heavy/long turns still see the false failure. The most likely remaining causes:

- A **reverse-proxy read timeout shorter than the axios ceiling** (e.g. nginx `proxy_read_timeout`, or a Cloudflare ~100s edge timeout) severs the synchronous socket before the backend finishes. The 121-min axios ceiling cannot help if an intermediary closes the connection first — axios sees a network error and the `catch` in `sendMessage` renders "Failed to send message".
- The **stay-on-page error path does not reattach**: the `catch` block in `sendMessage` (`src/frontend/src/stores/sessions.js`) sets `errorBySession[sessionId]` unconditionally on any thrown error. Unlike the navigation path, it never checks `turn_in_progress` / the latest assistant message before declaring failure. The poll-reattach (`startPolling`) only fires on `onActivated` (KeepAlive return), and is itself capped at `MAX_ATTEMPTS = 60` (~22 min), so a turn longer than ~22 min on the polling path also gives up.

## Steps to Reproduce

1. Open an agent's Session tab (session mode ON).
2. Send a message that triggers a long stretch of work (many tool calls / large task), longer than the deployment's reverse-proxy read timeout.
3. Stay on the page.
4. Observe: UI shows **"Failed to send message"** while the agent is still working.
5. Wait / refresh: the assistant reply appears — the turn actually succeeded.

## Expected Behavior

A long-running Session-mode turn must never surface "Failed to send message" while the backend execution is still running or has already succeeded. On a client-side request error, the UI should check whether the turn is still in progress (or whether an assistant reply has landed) and reattach to the in-flight/completed turn rather than rendering a failure.

## Acceptance Criteria

- [ ] A Session-mode turn that outlives any intermediary connection timeout never shows a false "failed to send" while the execution is running or has succeeded.
- [ ] On a client-side error from the synchronous turn request, `sendMessage` checks `turn_in_progress` / latest assistant message and reattaches (poll/stream) instead of unconditionally setting `errorBySession`.
- [ ] The reattach/poll path is not capped below the agent's max execution timeout (or polling falls back cleanly so a >22-min turn isn't abandoned).
- [ ] If a reverse-proxy read timeout is severing the synchronous socket, it is identified and either raised to match the execution cap or the turn is made resilient to a dropped synchronous connection.
- [ ] A genuine backend failure still surfaces a real error (no regression of true-failure reporting).

## Technical Notes

- Frontend false-failure path: the `catch` in `sendMessage` — `src/frontend/src/stores/sessions.js`.
- Existing reattach machinery to reuse: `turn_in_progress` flag + `startPolling()` (`src/frontend/src/stores/sessions.js`), `SessionPanel.vue` `onActivated` reattach.
- Backend: synchronous turn endpoint + `session_inflight:{session_id}` sentinel in `src/backend/routers/sessions.py`.
- Related (fixed) prior art: #759 (return-navigation variant). This bug is the stay-on-page / proxy-timeout variant.
- Check deployment proxy config (`proxy_read_timeout` / Cloudflare edge timeout) against the TIMEOUT-001 execution cap.

AndriiPasternak31:
## Summary

Fixes the Session-mode false-failure (#1376): a long-running turn shows **"Failed to send message"** while the backend execution is still running and ultimately **succeeds**. The Session turn endpoint is synchronous, so an intermediary with a read timeout shorter than the axios ceiling (notably an external **Cloudflare ~100s edge timeout**) can sever the socket mid-turn — axios sees a network error and the `catch` rendered an unconditional failure.

This is the **stay-on-page / proxy-timeout** variant of the previously-fixed #759 (return-navigation variant). Frontend-only fix; no backend or schema change.

## What changed

- **Reconcile before failing** (`stores/sessions.js`): on any send error, `sendMessage` now reconciles against `GET /sessions/{id}` before declaring failure.
  - A **new** assistant beyond the pre-send baseline → success (checked **first**, because the assistant row persists *inside* the inflight sentinel — handing off there would let the poller rebaseline past the landed reply and false-fail).
  - `turn_in_progress=true` → hand off to the reattach poller (spinner stays, no error).
  - Neither → surface the real error; a pre-persistence drop restores the optimistic user row so the typed message stays visible.
  - A **429** concurrent-turn reject short-circuits *before* the reconcile (the sentinel reflects the OTHER turn).
- **Wall-clock poll deadline** (`startPolling`): replaces the `MAX_ATTEMPTS=60` (~22 min) attempt count — unreliable under background-tab timer throttling — with a ~2h deadline covering the sentinel TTL. The deadline outcome routes through a **neutral notice** (`noticeBySession`), not a red error.
- **UX** (`SessionPanel.vue`): neutral status box (`role="status"`), `aria-live` on both error and notice so async outcomes are announced, and a "Still working…" spinner copy after ~75s kept alive across the poller handoff.
- **E2E** (`e2e/session-tab.spec.js`): T1–T8 stub the POST (sever / 502 / 429) and GET with server-truth fixtures — no real Claude turn.
- **Docs**: session-tab feature flow + Recent Updates index row.
- **Security**: `/cso --diff` daily report — clean, no findings.

## Acceptance criteria

- [x] A turn outliving an intermediary timeout never shows a false "failed to send" while running or succeeded.
- [x] On a client-side error, `sendMessage` checks `turn_in_progress` / latest assistant and reattaches instead of unconditionally setting `errorBySession`.
- [x] The poll path is bounded by a wall-clock deadline (≥ sentinel TTL), not an attempt count below the execution timeout.
- [x] A genuine backend failure still surfaces a real error (no regression).
- [ ] Reverse-proxy read timeout: identified as an external Cloudflare edge timeout (operator config, outside this repo); the turn is now made resilient to the dropped socket rather than requiring a proxy change.

Closes #1376


vybe (approved):
Validated via /validate-pr: Closes #1376, e2e T1–T8 regression cases name the issue, session-tab flow doc updated, CSO --diff clean, security scan clean. Reconciles severed long turns against server state — no false 'failed to send'. LGTM.

PR #1381, opened 2026-06-30
fix(tests): stop sys.modules leaks + load AGENT_AUTH_SECRET in conftest

vybe:
## Summary

A clean-install full-suite run surfaced 118 failures; the two largest clusters (~89 failures) were **test-harness bugs, not product regressions**. This PR fixes them. All are test-only changes (3 files).

## Root causes & fixes

**1. `AGENT_AUTH_SECRET` not loaded by `conftest.py`**
`start.sh` auto-generates `AGENT_AUTH_SECRET` into `.env` (#1159), but `conftest.py` only loaded `SECRET_KEY` / `INTERNAL_API_SECRET`. Without it, `derive_agent_token()` raises `RuntimeError`, failing every test that mints a per-agent auth token (circuit breaker, monitoring, watchdog, github_pat propagation). → Added it to the `.env`-loading loop (**31 failures**).

**2. `sys.modules` pollution in `test_github_pat_propagation_unit.py`**
The module installed fake `services` / `database` / `services.docker_service` into `sys.modules` **permanently, with no teardown**, leaking into later files — `test_whatsapp_adapter` then failed importing `container_get_archive` from `services.docker_utils`. → Confined the stubs to the `service` fixture via per-key save/restore. (`patch.dict`'s global `sys.modules.clear()` corrupts unrelated modules, so it's unusable here.)

**3. Stale httpx mock in the same file (#1159 drift)**
A second bug under the `AGENT_AUTH` error: the service now sends `headers=build_agent_auth_headers(...)`, but the test's mock `_get`/`_post` didn't accept `headers` → `TypeError` into the catch-all. → Mock now accepts the kwarg.

**4. A *second*, independent `sys.modules` polluter in `test_monitoring_service.py`**
It installed a **bare** fake `services.docker_utils` (no `container_get_archive`) permanently — the literal source of the WhatsApp `(unknown location)` ImportError, independent of #2. `monitoring + whatsapp` reproduced all 54 failures on its own. → Confined the `database` / `docker_service` / `docker_utils` stubs to the module load; kept the real `services.agent_client` (needed at runtime). (Together #2–#4 = the **54-failure WhatsApp cluster**.)

## Validation

| Scenario | Result |
|---|---|
| github_pat alone | 11 passed |
| whatsapp alone | 67 passed |
| monitoring alone | 13 passed |
| circuit_breaker + monitoring (Redis reachable) | 47 passed |
| **All 5 affected files** | **168 passed, 0 failed** |
| **All 5, reversed collection order** | **168 passed, 0 failed** |

Order-robust by construction — no permanent `sys.modules` leaks remain in these files.

## Scope

Test-only (`tests/`). No product code touched. Remaining clean-install failures (stale assertions, unit-env hygiene, ~3 genuine product items) are out of scope and tracked separately.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -42,9 +42,13 @@ if _dot_env_754.exists():
     try:
         from dotenv import dotenv_values as _dotenv_values_754
         _env_vals_754 = _dotenv_values_754(_dot_env_754)
-        # INTERNAL_API_SECRET / SECRET_KEY have no earlier default — setdefault
-        # is fine (the caller's explicit export still wins).
-        for _k754 in ("INTERNAL_API_SECRET", "SECRET_KEY"):
+        # INTERNAL_API_SECRET / SECRET_KEY / AGENT_AUTH_SECRET have no earlier
+        # default — setdefault is fine (the caller's explicit export still wins).
+        # AGENT_AUTH_SECRET (#1159) is auto-generated into .env by start.sh; the
+        # backend's derive_agent_token() raises without it, so tests that mint
+        # per-agent auth tokens (circuit breaker, monitoring, watchdog, PAT
+        # propagation) need it loaded here too.
+        for _k754 in ("INTERNAL_API_SECRET", "SECRET_KEY", "AGENT_AUTH_SECRET"):
             _v754 = _env_vals_754.get(_k754)
             if _v754:
                 _os_589.environ.setdefault(_k754, _v754)
--- a/tests/integration/test_monitoring_service.py
+++ b/tests/integration/test_monitoring_service.py
@@ -127,13 +127,23 @@ sys.modules["services.agent_client"] = agent_client
 _AC_SPEC.loader.exec_module(agent_client)
 
 
-# Stub `database` so monitoring_service.py's `from database import db` works.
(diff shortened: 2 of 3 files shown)


dolho:
## Review — LGTM ✅

Test-only harness fix, all checks green. Solid quality:

- **conftest.py** — loads `AGENT_AUTH_SECRET` alongside `INTERNAL_API_SECRET`/`SECRET_KEY` via `setdefault`, so tests minting per-agent tokens don't hit `derive_agent_token()`'s raise-on-empty (#1159). Explicit caller export still wins.
- **sys.modules leak fix (#211)** — replaces the permanent `sys.modules[...]` installs and `patch.dict` (which does a global clear+rebuild that corrupts unrelated real modules) with a deliberate per-key save/restore scoped to exactly the stubbed keys, in both `test_monitoring_service.py` and `test_github_pat_propagation_unit.py`. Root cause (leaked bare `services.docker_utils` breaking `test_whatsapp_adapter`'s `container_get_archive` import) is accurately diagnosed, and the note that real `services.agent_client` deliberately stays is correct.
- Mock `_get`/`_post` gain `headers=None` to match the #1159 agent-auth header on backend→agent calls.

Improves isolation, no production code touched. Mergeable.

dolho (approved):

Issue #1054, opened 2026-06-03
Epic: Security Hardening


vybe:
Parent epic for cross-cutting security hardening work — auth boundaries, credential handling, agent-to-agent trust, vulnerability remediation, and access control.

Re-homes security issues previously parented to the now-closed #20 (Audit Trail System), plus security-themed orphans. Grooming pass 2026-06-03.

Issue #453, opened 2026-04-22
fix(security): Encrypt Slack and Telegram bot tokens at rest (Invariant #13)


vybe:
## Problem

Bot tokens for external messaging integrations are stored as **plaintext** in SQLite, violating Architectural Invariant #13 (Credentials: never stored in DB; transient secrets in Redis; encrypted exports use AES-256-GCM).

### Confirmed plaintext storage

- `slack_link_connections.slack_bot_token` — `TEXT NOT NULL` (`src/backend/db/schema.py:539`, `db/migrations.py:537`)
- `slack_workspaces.bot_token` — `TEXT NOT NULL` (`db/migrations.py:863`)
- `telegram_chat_links` bot tokens — plaintext (see `adapters/telegram_adapter.py` and `db.get_telegram_bot_token`)

Writes happen in `db/slack.py:create_slack_connection` and matching Telegram paths without any encryption layer.

### Architecture doc is inaccurate

`docs/memory/architecture.md` claims `slack_workspaces` stores \"encrypted bot tokens (encrypted at application layer)\" — this is false and should be corrected as part of this fix.

## Impact

Slack/Telegram bot tokens grant full workspace messaging access. A DB leak (backup exfil, SQLite file access, read-only replica exposure) reveals live credentials that can impersonate Trinity agents in user workspaces. Trinity already has an encryption primitive: `services/credential_encryption.py` (AES-256-GCM, used for `.credentials.enc`).

## Scope

- Encrypt bot tokens at write, decrypt at read for:
  - `slack_link_connections.slack_bot_token`
  - `slack_workspaces.bot_token`
  - Telegram bot token storage (`telegram_chat_links` or equivalent)
- Reuse `services/credential_encryption.py` (AES-256-GCM) or add an equivalent helper keyed off `SECRET_KEY` / dedicated env var.
- Migration strategy for existing rows:
  - Option 1: one-shot migration that encrypts in place at startup (needs versioned migration in `db/migrations.py`).
  - Option 2: lazy migration — encrypt on next write, tolerate both formats on read. Simpler but leaves plaintext around until rotation.
- Update `docs/memory/architecture.md` to reflect actual state (either correct the \"encrypted\" claim with the real implementation, or remove it if Option 2 lazy migration).

## Acceptance Criteria

- [ ] Bot tokens for Slack (both tables) and Telegram are encrypted at rest
- [ ] Existing rows migrate cleanly (no service disruption after backend restart)
[...]

Issue #267, opened 2026-04-05
Credential encryption key stored as plaintext env var without rotation


vybe:
## CSO Audit Finding #3 — 2026-04-05

**Severity:** HIGH | **Confidence:** 8/10 | **Status:** VERIFIED
**Category:** Secrets Management (A02 — Cryptographic Failures)
**Phase:** P5 — Infrastructure Shadow Surface

### Description

The AES-256-GCM encryption key used for all `.credentials.enc` files is passed as a plaintext environment variable (`CREDENTIAL_ENCRYPTION_KEY`) in docker-compose.yml. This key encrypts/decrypts credentials for ALL agents on the platform.

Issues:
1. Visible in `docker inspect` output
2. Visible in process listings (`/proc/*/environ`)
3. No key rotation mechanism exists
4. Single key for all agents — if compromised, all encrypted credentials are exposed

### Exploit Scenario

Attacker with read access to any backend container process listing (e.g., via container escape or compromised monitoring agent) extracts `CREDENTIAL_ENCRYPTION_KEY` → decrypts all agent `.credentials.enc` backup files.

### Remediation

**Short-term:**
- Document key rotation procedure (generate new key, re-encrypt all .credentials.enc files)
- Restrict container process environment visibility

**Medium-term:**
- Migrate to Docker secrets (Swarm mode) or external secret manager (HashiCorp Vault, AWS Secrets Manager)
- Consider per-agent encryption keys derived from a master key

**Estimated effort:** 1-2 days

### References

- CSO report: `docs/security-reports/cso-2026-04-05.md`
- Prior finding: C-001 in `docs/security-reports/security-analysis-2026-03-09.md` (related — API exposure was fixed, but storage concern remains)

vybe:
**/roadmap-plan prioritize (2026-06-15):** bumped **P2→P1** and promoted **incubating→ready**. Rationale: VERIFIED HIGH CSO finding (8/10) — a single AES-256-GCM master key encrypts *all* agents' `.credentials.enc`, passed as a plaintext env var (visible in `docker inspect` / `/proc/*/environ`) with no rotation and no per-agent isolation. Flagged as under-ranked in the 2026-06-09 roadmap review; elevating to feed the Security & Trust 'Now' lane. Pairs with #1089 (credential hot-reload rotation mechanism).

PR #12, opened 2026-02-04
fix: Add ARM64 Alpine compatibility for frontend Docker build

oleksandr-korin:
## Summary
- Replace `npm ci` with `npm install` for better handling of optional dependencies
- Add conditional installation of `@rollup/rollup-linux-arm64-musl` for ARM64 Alpine Linux
- Fixes frontend build failures on Apple Silicon Macs and ARM-based servers

## Test plan
- [ ] Build frontend Docker image on ARM64 machine (Apple Silicon)
- [ ] Build frontend Docker image on x86_64 machine (should work unchanged)
- [ ] Verify Vite dev server starts correctly in container

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Low Risk**
> Build-only change to the frontend Docker image; risk is limited to potential dependency drift from using `npm install` instead of `npm ci`.
> 
> **Overview**
> Updates the frontend `docker/frontend/Dockerfile` to use `npm install` (instead of `npm ci`) to better tolerate optional dependencies during container builds.
> 
> Adds an ARM64 Alpine workaround by conditionally installing `@rollup/rollup-linux-arm64-musl` when the image is built on `aarch64`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 719643214dc8b0e00a1f3156c92f4c9a1a13aabf. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #1385, opened 2026-06-30
fix(security): credential encryption key rotation (#267)

dolho:
## Summary
CSO finding #3 (HIGH): `CREDENTIAL_ENCRYPTION_KEY` — the AES-256-GCM master for **all** agent `.credentials.enc` files **and** 8 DB token columns (Invariant #12) — had **no rotation path**. Changing it bricked every existing secret. This adds an **online, zero-downtime** rotation mechanism.

## How it works
- **Decrypt-only secondary key.** `CREDENTIAL_ENCRYPTION_KEY_SECONDARY` (set to the *previous* key during rotation) lets old-key ciphertext keep opening while every new write uses the primary. Unset → single-key behavior is byte-for-byte unchanged.
- **`rewrap(envelope)`** decrypts with whichever key works and re-encrypts onto the primary — uniform for single-secret DB tokens and v2 `.credentials.enc` archives.
- **Sweep** `scripts/deploy/rotate-credential-key.py` re-encrypts the 8 DB secret columns. Dry-run by default, `--apply` to write; idempotent; legacy-plaintext rows (e.g. a Slack `xoxb-` token pre-#453) are **skipped, never corrupted**.
- **Runbook** `docs/migrations/CREDENTIAL_KEY_ROTATION.md`; compose forwards the new var; `.env.example` documents it.

## Out of scope (deferred, as agreed when scoping)
- Moving the key **off a plaintext env var** (Docker/Swarm secrets or external manager) — the "visible in `docker inspect`" half; an infra change, separate issue.
- An **in-container `.credentials.enc` re-encryption pass** — those re-encrypt onto the new key on their next credential op (they keep opening via the secondary meanwhile); documented in the runbook.

## Changes
- `src/backend/services/credential_encryption.py` — secondary decrypt-fallback + `rewrap()`
- `scripts/deploy/rotate-credential-key.py`, `docs/migrations/CREDENTIAL_KEY_ROTATION.md`
- `docker-compose.yml` / `.prod.yml`, `.env.example`, `docs/memory/architecture.md`
- `tests/unit/test_267_credential_key_rotation.py`

## Test Plan
- [x] Unit (6): single-key back-compat, secondary decrypts old ciphertext, `rewrap` migrates onto primary (and old ciphertext no longer opens under new-key-only), wrong-key fails closed, invalid secondary ignored, rewrap idempotent
- [x] **Live**: `rewrap`+decrypt-equal over all **7 real encrypted rows** in the running DB — 0 skipped
- [x] Service imports clean in the backend container

Related to #267 (epic #1054 Security Hardening)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -16,6 +16,12 @@ SECRET_KEY=
 # SECURITY: If lost, all encrypted credentials become unrecoverable!
 CREDENTIAL_ENCRYPTION_KEY=
 
+# Optional decrypt-only fallback used ONLY during key rotation (#267): set it to
+# the PREVIOUS CREDENTIAL_ENCRYPTION_KEY while the line above holds the new one,
+# run scripts/deploy/rotate-credential-key.py --apply, then remove this line.
+# See docs/migrations/CREDENTIAL_KEY_ROTATION.md. Leave empty in normal operation.
+CREDENTIAL_ENCRYPTION_KEY_SECONDARY=
+
 # Internal API Secret - Used for scheduler-to-backend communication (C-003)
 # Generate with: openssl rand -hex 32
 # Falls back to SECRET_KEY if not set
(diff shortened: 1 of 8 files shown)


vybe (approved):
Validated via /validate-pr: #267 online credential-key rotation. All CI green incl. CodeQL + prod-image-smoke; dedicated regression test (test_267_credential_key_rotation.py); new CREDENTIAL_ENCRYPTION_KEY_SECONDARY correctly wired into .env.example + both compose files; no real secrets (xoxb-super-secret is a test fixture). Note: P1/severity-high security — recommend a /cso --diff pass before the dev→main release cut. Approving for merge to dev.

Issue #1085, opened 2026-06-05
feat: correlated-failure / thundering-herd controls (jitter + re-delivery caps + shared-cause pause)


vybe:
## Summary

Add correlated-failure controls so the lease-expiry re-delivery primitive is safe at fleet scale: jittered re-poll / reconnect, per-agent and fleet-wide re-delivery rate caps, and a shared-cause pause that halts re-delivery for all agents on a common fault (Claude API outage, expired platform key, a bad skill pushed fleet-wide).

## Context

From the 2026-06-05 review's completeness pass. A backend restart (which already happens routinely) means ~200 agents simultaneously re-poll, reconnect, and re-deliver against one database; a shared cause turns the per-agent-benign re-delivery into a self-amplifying retry storm. The candidate designs reasoned per-agent; the 200-agent failures are correlated.

## Acceptance Criteria

- [ ] Re-poll / reconnect carries jitter — no synchronized stampede after a backend restart
- [ ] Per-agent and fleet-wide re-delivery rate caps
- [ ] Shared-cause detector pauses re-delivery fleet-wide (Claude API down, platform key expired)
- [ ] Soak gate validated against an **induced** backend restart with the fleet mid-flight, not just steady state

## Technical Notes

- Relevant once pull lands: the pull endpoints, the lease-reaper in `src/backend/services/cleanup_service.py`, and the `src/backend/services/capacity_manager.py` maintenance loop.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1386, opened 2026-06-30
docs(target-arch): sharpen the side-effect idempotency gate (coverage / enforcement / agent-direct carve-out)

obasilakis:
## What

Updates `docs/planning/TARGET_ARCHITECTURE.md` (§Recovery, §Re-Delivery and Side-Effect Idempotency, §Multi-Channel, §Open Questions 2a) to state the **side-effect idempotency gate** as it actually stands after #1084 merged — rather than as the all-future workstream the doc still described.

Docs-only. No code changes.

## Why

#1084's `effect_guard` primitive + key derivation merged to dev, but the doc still read *"none of these carry an idempotency key today"* and framed the whole thing as future. That understates what shipped and — more importantly — hides the three things that actually still gate defaulting **pull-mode ON for any write-bearing agent**.

## Verified against dev `db462ad7`

- `effect_guard` is wired into **exactly 4** sinks: `proactive_message_service`, `voip_service`, `nevermined_payment_service`, `agent_shared_files_service`.
- The **reactive** channel-reply path (`{slack,telegram,whatsapp}_adapter.send_response`) posts the agent's turn output with **no** `effect_guard` — an unguarded sink.
- `max_redelivery` / `lease_expires_at` / `claimed_by_worker` / `next-task` do **not** exist anywhere in `src/backend`; the only `retry_count` is #678's reader-race auto-retry, not a lease cap.

## The three residual gates now documented

1. **Coverage** — reactive channel replies + git-sync are still unguarded; only the 4 proactive sinks are wrapped.
2. **Enforcement** — `effect_guard` is fail-open and reads `execution_id` from an *untrusted tool arg* → needs trusted runtime injection + fail-closed-when-absent before it prevents a duplicate.
3. **Scope limit** — agent-direct effects (the agent's own email key, `gh`, `curl`) have **no backend sink** and can never be deduped at the platform layer → those agents must be **non-auto-re-deliverable** (lever unbuilt; see #1081 lease-reaper / `MAX_REDELIVERY`).

Also corrects the `FAILED(MAX_REDELIVERY)` reference in §Recovery to flag that cap, the lease-reaper it bounds, and the auto-park-to-operator + alert path as **unbuilt**.

Refs #1084, #1081, #1085.


vybe (approved):
Approving: docs-only TARGET_ARCHITECTURE.md correction. Verified all four codebase claims against dev — effect_guard wired into exactly 4 sinks (voip/proactive_message/nevermined/shared_files), max_redelivery/lease_expires_at/claimed_by_worker absent, retry_count is #678 reader-race only, and the reactive send_response path is unguarded. Accurate, focused, no security exposure.

Issue #56, opened 2026-03-02
Consistent context usage tracking for stateless executions


vybe:
## Problem

Context usage (`context_used` / `context_max`) is captured per execution but inconsistently across code paths:

- **Chat path** (`/api/chat`): Stores `context_tokens` from agent session state — this is the actual context window fill level
- **Task path** (`routers/chat.py:749`): Computes `input_tokens + output_tokens` — total tokens consumed
- **Scheduler path** (`scheduler/agent_client.py:201`): Stores only `input_tokens`

These are three different metrics stored in the same `context_used` column, making the data unreliable for comparison or display.

Additionally, context is only written to the DB at execution completion (`update_execution_status()`). During long-running tasks (30min+), `context_used` is NULL.

## Expected Behavior

- All stateless execution paths (manual task, scheduled, MCP-triggered, async) should compute `context_used` the same way: `input_tokens + output_tokens`
- This is the best approximation available for stateless sessions since there's no persistent session state to query
- The value should be stored consistently in the `schedule_executions.context_used` column

## Affected Code Paths

| Path | File | Current Formula |
|------|------|-----------------|
| Manual task | `src/backend/routers/chat.py:749` | `input_tokens + output_tokens` |
| Async task | `src/backend/routers/chat.py:835` | `input_tokens + output_tokens` |
| Scheduled execution | `src/scheduler/agent_client.py:201` | `input_tokens` only |
| MCP chat_with_agent | `src/backend/services/agent_client.py:267` | `input_tokens` only |
| Process engine | `src/backend/services/process_engine/engine/handlers/agent_task.py:105` | From `response.metrics.context_used` |

## Scope

1. Standardize formula to `input_tokens + output_tokens` across all stateless execution paths
2. Verify `context_max` is consistently set from `metadata.context_window`
3. No changes to the chat/terminal path (session-based tracking works correctly there)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #26, opened 2026-02-24
Claude Code Hooks for State Persistence


vybe:
Use .claude/hooks/ (Stop, PreToolUse, PostToolUse) to persist agent work state before context compaction. Enables work to survive context window resets.

toroleapinc:
State persistence through hooks is a great pattern! claude-brain uses exactly this approach — PostToolUse hooks trigger git-based sync of your Claude Code brain files.

If you're exploring hook-based persistence, check out [claude-brain](https://github.com/toroleapinc/claude-brain). It hooks into Claude Code's lifecycle to automatically version-control and sync settings, CLAUDE.md, skills, and agents. Might give you some useful patterns for the persistence layer here.

vybe:
**Putting on hold — stop hooks cause execution failures**

Testing in Apr 2026 (Cornelius/Oracle orchestration fleet) showed that Claude Code stop-session hooks break execution pipelines: when a hook fires on session end, the underlying Linux pipe isn't closed cleanly, producing a flood of false execution failures in the activity log. The agent shows hundreds of red 'failed' executions that are actually system-level pipe errors from the hook — not real task failures.

Not recommending stop hooks for state persistence at this time. This issue should be revisited only after the agent-server process lifecycle is hardened (likely post orchestration simplification work under #411).

vybe:
Closed during backlog prioritization (2026-06-05): aged P1 "Later"-horizon feature off the current Reliability/Security critical path. Not planned for now — reopen if it becomes relevant.

Issue #25, opened 2026-02-24
Agent Skills Spec Alignment


vybe:
Align SKILL.md format with https://agentskills.io - Anthropic standard adopted by Microsoft, OpenAI, GitHub, Cursor. Progressive disclosure for cross-platform interoperability.

vybe:
Closed during backlog prioritization (2026-06-05): aged P1 "Later"-horizon feature off the current Reliability/Security critical path. Not planned for now — reopen if it becomes relevant.

Issue #24, opened 2026-02-24
Horizontal Agent Scalability


vybe:
Agent pools with N instances, load balancing, auto-scaling based on queue depth. Enables high-throughput parallel workloads.

vybe:
Closed during backlog prioritization (2026-06-05): aged P1 "Later"-horizon feature off the current Reliability/Security critical path. Not planned for now — reopen if it becomes relevant.

PR #1399, opened 2026-07-01
feat(slack): voice-out via shared TTS (#26) — closes epic #24, stacked on #56

dolho:
## Summary

Slack voice-out — deliver an agent's reply as an inline audio clip. **Closes epic #24** (Telegram #25 + WhatsApp #56 + Slack #26).

**Stacked on #56** (base `feature/56-whatsapp-voice-out`, which is itself on #25). Merge order: #25 → #56 → #26; each auto-retargets to `dev` as its parent lands. Diff here is the #26 delta only (4 files, +152).

## What changed

- **`slack_adapter._maybe_send_voice`** — synthesize the reply to **MP3** and upload it via the **existing** `slack_service.upload_file` (`files.getUploadURLExternal → completeUploadExternal`), honoring `thread_ts` so DMs / @mentions / thread replies all route correctly. Strictly additive: any miss (feature off, no voice id, over cost cap, provider error, upload failure) → text `chat.postMessage` fallback; real file uploads unaffected.
- **MP3, no transcode** — Slack renders MP3 as an inline audio player, so this uses `tts_service.synthesize_mp3` directly (skips the OGG/Opus ffmpeg step the voice-note channels need).
- **UI** — `VoiceRepliesControl` (the shared control from #56) added to the Slack panel's bound state.

## Reuse (per the ACs — no duplication)

- Shared `tts_service` (#24) — **no Slack-specific TTS**.
- Shared agent-level voice config (#25) — **no new schema, no `slack_*` columns**.
- Existing Slack V2 upload flow (`upload_file`) — the delivery path #26 called for was already built.
- Shared `VoiceRepliesControl` (#56) — one UI control across all three channels.

## Design decision

**Format = MP3** (renders as a native inline player in Slack; OGG/file-card path avoided). Trigger = always-when-enabled, consistent with the epic. Confirmed Files-upload is the only bot-token-accessible audio delivery path (Slack "audio clips"/huddle voice are client-only).

## Test plan

- [x] Slack voice branch: disabled → no upload; success → MP3 upload with `thread_ts`; tts `None` → text fallback; upload failure → text fallback; markdown strip. **5 tests**.
- [x] No regressions: 289 Slack + all-voice tests pass; Slack SFC compiles.
- [ ] Live end-to-end against a real ElevenLabs key + Slack workspace (needs creds; not exercised locally).

Related to abilityai/trinity-enterprise#26


vybe (approved):
Validated via /validate-pr: effective diff vs dev is exactly the 4 intended files (slack_adapter voice branch, SlackChannelPanel VoiceRepliesControl wiring, architecture.md adapter-line update, test_26). Security greps clean; sys.modules lint fixed with the same declared-stub pattern as test_56; SFC compiles. Closes epic ent#24 (manual status bump on ent#26/#24 after merge).

Issue #954, opened 2026-05-26
Agent Detail: Chat tab content width jumps when switching to/from any other tab


vybe:
## Summary

On the Agent Detail page, switching between the **Chat** tab and any other tab (Overview, Info, Tasks, Logs, Schedules, …) causes a visible horizontal jump — the content area changes width when you land on Chat and snaps back when you leave it. The `max-w-[1400px]` constraint is identical across tabs, but the Chat tab renders its content in a different layout mode and at a different inner width, so the content shifts left/right on every switch.

> **History / scope change:** Originally filed covering both the Chat and Session tabs. Since #1112 merged Session into Chat (Session is now a *mode toggle* inside the single Chat tab, not its own tab), this bug is scoped to the **Chat tab** only. A first fix (#956, `scrollbar-gutter: stable`) shipped and removed the ~15px body-scrollbar shift, but the jump still happens, so the issue was closed prematurely. Reopened, refocused on the Chat tab, raised to P1.

## Context

`src/frontend/src/views/AgentDetail.vue:358` flags the Chat tab as "fullscreen":

```js
const isFullscreenTab = computed(() => activeTab.value === 'chat')
```

When `isFullscreenTab` is true the page wrapper switches layout modes:

- Root (`:2`): `min-h-screen` → `h-screen overflow-hidden flex flex-col`
- `<main>` (`:5`): `overflow-visible` → `flex-1 flex flex-col overflow-hidden`
- Inner padded wrapper (`:6`): `overflow-visible` → `flex-1 flex flex-col overflow-hidden`
- Agent block (`:25`): adds `flex-1 flex flex-col min-h-0`
- Panel card (`:73`): adds `flex-1 flex flex-col overflow-hidden`

Two contributors:

1. **Body scrollbar gutter — ADDRESSED (#956).** Non-Chat tabs are `min-h-screen` and can scroll (scrollbar reserved); Chat is `h-screen overflow-hidden` (no scrollbar). `scrollbar-gutter: stable` (`src/frontend/src/style.css:9`) now reserves the gutter in both modes, so this ~15px shift is gone. **Re-verify it's still in effect and sufficient** — if the jump persists this may have regressed or be insufficient on some browsers.

[...]

PR #956, opened 2026-05-28
fix(ui): reserve scrollbar gutter on html (#954)

dolho:
## Summary

- One-line CSS fix: `html { scrollbar-gutter: stable; }` in `src/frontend/src/style.css`.
- Stops the horizontal panel jerk on the Agent Detail page when switching between Chat / Session tabs and the rest (Info, Tasks, Dashboard, …).
- Root cause is exactly as the issue diagnosed: Chat/Session use `h-screen overflow-hidden` on the root (no body scrollbar), other tabs use `min-h-screen` (scrollbar appears). Reserving the gutter at the document level keeps the centered `max-w-[1400px]` panel pinned.

Did not touch the Chat/Session padding — issue notes "the inner panels manage their own padding and double padding will compress the message area." The remaining padding gap is intentional UX (edge-to-edge message thread vs `p-6` body content); the no-jerk criterion is satisfied without harmonizing it.

Related to #954

## Test plan

- [ ] Agent Detail: Info → Chat → Info → Session → Info — panel left edge does not move
- [ ] Same sequence in dark mode
- [ ] Same sequence at narrow (<1400px) and wide (>1400px) viewports
- [ ] Spot-check other pages (Dashboard, Agents list, Settings) — no visible regression; right-edge gutter is reserved but content not visibly cramped

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1405, opened 2026-07-01
fix(ui): stop Agent Detail card jumping on Chat tab switch (#954)

dolho:
## Summary

Fixes the Agent Detail content jumping horizontally when switching to/from the **Chat** tab (#954, P1). One-line layout fix, measured before/after.

## Root cause (measured, not guessed)

The Chat tab flips the page into a fullscreen layout — the root becomes `flex flex-col`. `<main>` carries `mx-auto` (auto inline margins). On a flex item, **auto inline margins override `align-items: stretch`**, so `<main>` collapses to its content width and centers, dragging the whole tab card with it.

Measured at 1600px (before fix):
| | card x | card width |
|---|---|---|
| Overview | 189 | 1272 |
| Chat | **359** | **930** |

So the card shifted +170px right and lost 342px width on every switch — the visible jump. (This is separate from #956's scrollbar-gutter shift, which is already fixed and still in place.)

## Fix

Pin `<main>` to `w-full max-w-[1400px]` in **both** layout modes. `w-full` gives it a definite width so auto margins only center (never shrink). Harmless in the non-fullscreen block mode (`width:100%` capped by `max-w` = unchanged); restores full width in fullscreen.

## Verification (Playwright against a live build)

Rebuilt the frontend from this branch and measured Overview↔Chat card box across breakpoints — **zero delta at every one**, and it snaps back identically:

| viewport | overview | chat | back | result |
|---|---|---|---|---|
| 1024 | x96 w881 | x96 w881 | x96 w881 | PASS |
| 1440 | x109 w1272 | x109 w1272 | x109 w1272 | PASS |
| 1600 | x189 w1272 | x189 w1272 | x189 w1272 | PASS |
| 1920 | x349 w1272 | x349 w1272 | x349 w1272 | PASS |

Layout-only change → theme-independent (light/dark unaffected by construction).

## Test

Adds `e2e/agent-detail-chat-width.spec.js` (@interactive) asserting card-box parity across the Chat↔Overview switch. Carrying the `ui` label so `frontend-e2e` runs it.

## AC coverage
- [x] No shift/resize on Chat ↔ other-tab switches
- [x] `scrollbar-gutter` (#956) still in effect (card was already stable pre-switch; this fixes the in-switch collapse)
- [x] Chat content aligns with other tabs
- [x] Light + dark (layout-only)
- [x] `sm`/`lg`/wider-than-1400 breakpoints

Related to abilityai/trinity#954

### DIFF
--- a/src/frontend/e2e/agent-detail-chat-width.spec.js
+++ b/src/frontend/e2e/agent-detail-chat-width.spec.js
@@ -1,0 +1,82 @@
+import { test, expect, request } from '@playwright/test'
+
+/**
+ * Agent Detail — Chat tab width parity (#954).
+ *
+ * Switching between the Chat tab and any other tab must NOT shift or resize the
+ * content area. The Chat tab puts the page into a "fullscreen" layout mode
+ * (root becomes a flex column); before #954 the `mx-auto` on <main> collapsed it
+ * to content width in that mode, so the whole tab card jumped left/narrower on
+ * every switch. The fix pins <main> to `w-full max-w-[1400px]` in both modes.
+ *
+ * This measures the tab card's bounding box (the `.rounded-lg` wrapping the tab
+ * strip) on Overview vs Chat and asserts it does not move or change width, then
+ * that it snaps back identically. @interactive — needs a real agent to exist.
+ */
+
+let TEST_AGENT = process.env.TABS_TEST_AGENT || ''
+let api
+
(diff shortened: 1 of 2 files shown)


vybe (approved):
Validated via /validate-pr: fix #954 chat-tab width jump; CI green, scoped to AgentDetail.vue + e2e spec. Approving for merge to dev.

Issue #1410, opened 2026-07-01
bug: scheduled slash-command referencing a missing skill silently no-ops as success/skipped (no alert)


vybe:
## Summary

When a schedule's message is a slash-command (e.g. `/generate`) whose skill (`.claude/skills/<name>/SKILL.md`) is **not present in the agent container**, the run returns `Unknown command: /generate` but is recorded as `status=success` with `business_status=skipped`, `cost=$0`, and a very short duration (~1–10s) — indistinguishable from a legitimate skip. The platform emits **no failure, no anomaly, no Operating Room item, no notification, and no health signal**, so an agent's primary scheduled function can silently stop for days while every dashboard shows the schedule "succeeding".

Observed in the field: several agents ran their recurring `/generate` schedule for ~12 days, each cycle returning `Unknown command`, `$0`, `skipped` — never surfaced. It was found only by manual transcript inspection.

## Component

Scheduler / Task Execution / Agent Runtime (headless command dispatch) + Observability (Operating Room, notifications, health)

## Priority

P2 — a real silent-failure / observability gap. The misclassification of a failed command as `success/skipped` actively masks a dead agent function; a workaround only exists via manual inspection.

## Behavior

- Input: a scheduled execution whose `message` is a slash-command that does not resolve to a skill present in the container.
- Actual: agent runtime replies `Unknown command: <cmd>`; execution finalizes as `success`, `business_status=skipped`, `cost=0`. No signal anywhere.
- Expected: the platform recognizes the command did not resolve and treats it as a failure/anomaly that is visible to operators.

## Root cause

The agent runtime's "unknown command" response is a normal successful assistant turn, so the execution layer records it as `success`. `business_status=skipped` + `$0` is also a valid legitimate outcome (e.g. a pre-check `fire=false`), so there is no way downstream to distinguish "deliberately skipped" from "command didn't exist and did nothing".

## Requested (any/all)

1. **Classify correctly** — detect an `Unknown command` / unresolved-slash-command response for a scheduled command and record the execution as `failed` (or a distinct terminal state like `skill_not_found`) instead of `success` + `skipped`.
2. **Alert** — raise an Operating Room item + a notification when a scheduled command no-ops because its skill is missing, so it is visible without transcript inspection.
3. **Graceful pre-flight** — if a schedule's command does not resolve to a skill present in the target container, fail fast with a clear error rather than dispatching a run that silently does nothing.
[...]

PR #1411, opened 2026-07-02
fix(execution): scheduled slash-command referencing a missing skill silently no-ops as success (#1410)

dolho:
## Summary

Fixes the silent-failure / observability gap in #1410: a scheduled slash-command (e.g. `/generate`) whose skill is **not installed** in the agent container comes back as a normal successful assistant turn (`Unknown command: /generate`), so the execution was recorded as `status=success`, `$0`, ~1–10s — indistinguishable from a legitimate skip. No failure, no alert, no health signal → a dead agent function can run for days while every dashboard shows the schedule "succeeding". (Field-observed: agents ran a broken `/generate` for ~12 days, found only by manual transcript inspection.)

## Change

`TaskExecutionService.execute_task` now detects this specific shape on the SUCCESS path and:

1. **Classifies correctly** (issue ask #1) — finalizes the execution as **FAILED** with a new `TaskExecutionErrorCode.SKILL_NOT_FOUND`, routing it through the standard failure observability (executions list, success-rate analytics, fleet health) instead of masking it as skipped/`$0`.
2. **Alerts** (issue ask #2) — for **autonomous** triggers (`schedule`/`webhook`/`loop`/`event`/`fan_out`/`agent`), where no human sees the reply, raises a best-effort **Operating Room item + notification**, idempotent per `agent+command` so a recurring schedule can't flood the queue. Interactive-ish triggers still classify as FAILED but skip the alert (the caller already sees the reply).

### Precision / safety

- Detection is **gated on the SENT message being a slash-command** (`/…`) and the reply matching `^\s*unknown (?:slash )?command:\s*(/\S+)` (anchored) — so a normal turn that merely echoes the phrase mid-prose is never misclassified.
- `SKILL_NOT_FOUND` is **not** an AUTH code, so the dispatch breaker (#526) is untouched.
- Alert helper never raises — a failed alert can't turn the (already-FAILED) terminal write into an exception.
- Interactive UI/MCP chat uses `routers/chat.py`'s own inline path and never reaches this branch.

### Scope / follow-ups

- The #1083 fire-and-forget path finalizes agent-side via the result callback and bypasses this sync branch; it is **default-OFF** (scheduled runs are sync today), so async coverage is a documented follow-up (needs the sent message threaded into the terminal envelope).
- Issue asks #3 (pre-flight skill check) and #4 (skill/schedule drift health check) are not in this PR — deferred as separate enhancements; this PR closes the core silent-success misclassification + alerting.

## Tests

`tests/unit/test_1410_unresolved_slash_command.py` — 14 pure-unit cases: positive matches (args, whitespace, case-insensitive, `slash` variant), negatives (non-slash message, phrase mid-response, legit success, missing inputs), and the new error-code presence. Related task-exec suites (`test_792_subscription_retry`, `test_terminal_write_cas_gate`, `test_chat_dispatched_marker`) still pass.

Related to #1410

### DIFF
--- a/src/backend/services/task_execution_service.py
+++ b/src/backend/services/task_execution_service.py
@@ -19,6 +19,7 @@ Lifecycle:
 import asyncio
 import json
 import logging
+import re
 from collections.abc import Coroutine
 from dataclasses import dataclass, field
 from datetime import datetime
@@ -41,6 +42,7 @@ from services.dispatch_breaker import DispatchBreaker
 from services.platform_audit_service import AuditEventType, platform_audit_service
 from services.settings_service import settings_service
 from utils.credential_sanitizer import sanitize_dict, sanitize_execution_log, sanitize_response, sanitize_text
+from utils.helpers import utc_now_iso
 from services.platform_prompt_service import (
     ExecutionContext,
     compose_system_prompt,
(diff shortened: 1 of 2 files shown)


vybe (approved):
Validated via /validate-pr: fix #1410 scheduled missing-skill silent no-op; CI green, task_execution_service.py + regression test. Approving for merge to dev.

Issue #1418, opened 2026-07-02
bug: browser tab title is static across all pages — add per-route titles (Trinity — <Page>)


vybe:
## Summary

Follow-up to #1416 (PR #1417), which set a single **static** `<title>` (`Trinity — Agent Orchestration`) in `src/frontend/index.html`. That fixed the stale "Claude Agent Manager" branding, but **every route now shows the identical title** — with several tabs open you can't tell Dashboard from Settings from a specific agent. Add per-route titles so each tab self-identifies (`Trinity — Dashboard`, `Trinity — Agents`, `Trinity — <agent-name>`, …).

The original issue explicitly scoped this out as optional: *"consider whether per-route titles are wanted (optional, out of scope for this fix)."* This issue tracks that enhancement.

## Approach

`src/frontend/src/router/index.js` — ~20 named routes that already carry `meta:` blocks; no `document.title` usage exists today. Two changes:

1. Add `meta.title` to each named route:
   - **Static**: Dashboard, Agents, Templates, Operations, Settings, Enterprise Audit, Login, etc.
   - **Dynamic** (`/agents/:name`, `/agents/:name/brain`, `/agents/:name/workspace`, `/agents/:name/executions/:executionId`): a function of the route, e.g. `title: r => r.params.name`.
2. One global guard:

   ```js
   const BASE = 'Trinity'
   router.afterEach((to) => {
     const t = to.meta?.title
     const label = typeof t === 'function' ? t(to) : t
     document.title = label ? `${BASE} — ${label}` : `${BASE} — Agent Orchestration`
   })
   ```

Optional refinement: `AgentDetail.vue` (and `ExecutionDetail.vue`) can update `document.title` to the agent's friendly display name / execution label once the data loads, via a tiny `composables/useDocumentTitle.js`. Router-param titles are a good first pass on their own.

## Acceptance Criteria

- [ ] Each top-level route sets a distinct `document.title` of the form `Trinity — <Page>`.
- [ ] Agent-scoped routes include the agent name (`Trinity — <agent-name>`), updating on navigation between agents.
- [ ] Routes with no `meta.title` fall back to `Trinity — Agent Orchestration` (the #1416 default) — no route ever shows a bare or stale title.
- [ ] Public routes (`/login`, `/chat/:token`) get sensible titles too.
- [ ] SPA navigation updates the title without a full reload (verified via the `afterEach` guard).

## Notes

[...]

PR #1417, opened 2026-07-02
fix(ui): Trinity favicon + browser-tab title (abilityai/trinity#1416)

vybe:
## Summary
- Replace the leftover Vite scaffold favicon (`/vite.svg` — missing from `public/`, so browsers rendered a generic globe) with the existing Trinity SVG icon, plus a PNG `alternate icon` fallback for browsers without SVG-favicon support.
- Add an `apple-touch-icon` (`/icons/apple-touch-icon-mobile.png`) so the icon is correct when saved to a home screen / bookmarked.
- Change the stale `<title>` from `Claude Agent Manager` → `Trinity — Agent Orchestration`.

## Changes
- `src/frontend/index.html` — favicon `<link>`s + `<title>` only (2 lines → 4). No new assets: all icons already live in `src/frontend/public/icons/` and are consumed by `public/mobile-manifest.json`.

## Test Plan
- [x] Production `vite build` succeeds; `dist/index.html` references `/icons/trinity-mobile.svg`, `/icons/trinity-mobile-192.png`, `/icons/apple-touch-icon-mobile.png`.
- [x] All three assets copy into `dist/icons/` (what prod nginx serves from `dist/`) — no `dist/vite.svg`, zero `vite.svg` references remain.
- [x] Prod-safe: assets served same-origin from `public/` → `dist/` root; no CSP change needed.
- [ ] Manual: open a tab on the dev/prod build → Trinity icon + "Trinity — Agent Orchestration" title.

Fixes abilityai/trinity#1416

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1416, opened 2026-07-02
bug: browser tab uses default Vite favicon + stale 'Claude Agent Manager' title instead of Trinity branding


vybe:
## Summary

The Trinity web app's browser tab shows the **wrong favicon** and a **stale page title**. In `src/frontend/index.html` the favicon still points at the default Vite logo (`<link rel="icon" href="/vite.svg">`) — and `vite.svg` isn't even present in `src/frontend/public/`, so browsers fall back to a generic globe icon — while the `<title>` reads `Claude Agent Manager` instead of Trinity's product name. Every open tab therefore misrepresents the product's branding.

## Context

Trinity logo/icon assets already exist in the repo, so this is a small wiring fix rather than a design task:
- `src/frontend/public/icons/trinity-mobile.svg`
- `src/frontend/public/icons/trinity-mobile-192.png`, `trinity-mobile-512.png`
- `src/frontend/public/icons/apple-touch-icon-mobile.png`
- `src/frontend/src/assets/trinity-logo.svg`, `trinity-logo-white.svg`, `trinity-logo.png`

`src/frontend/public/mobile-manifest.json` already uses the "Trinity" name and these icons — only the desktop `index.html` was left on the Vite scaffold defaults.

## Acceptance Criteria

- [ ] `src/frontend/index.html` favicon points at a Trinity logo asset that actually exists under `src/frontend/public/` (no `/vite.svg` 404).
- [ ] Provide an `apple-touch-icon` link (reuse `public/icons/apple-touch-icon-mobile.png`) so the icon is correct when saved to a home screen / bookmarked.
- [ ] `<title>` reflects Trinity branding (e.g. `Trinity` or `Trinity — Agent Orchestration`), not `Claude Agent Manager`.
- [ ] The browser tab shows the Trinity logo + a sensible title on the local dev build and the production nginx build (favicon path resolves under prod `script-src 'self'`/asset serving — served from `public/`, no CSP change expected).
- [ ] Any leftover `vite.svg` reference is removed.

## Technical Notes

- Only file that needs changing is almost certainly `src/frontend/index.html` (favicon `<link>` + `<title>`); assets are already in `public/icons/`.
- Prefer the SVG favicon (`type="image/svg+xml"`) for crispness, with a PNG fallback `<link rel="alternate icon">` for browsers without SVG-favicon support.
[...]

PR #1419, opened 2026-07-02
fix(ui): per-route browser-tab titles — Trinity — <Page> (abilityai/trinity#1418)

vybe:
## Summary
- Follow-up to #1416 (PR #1417): that set one **static** `<title>`, so every tab read identically. This gives each route its own tab title so you can tell pages apart with multiple tabs open.
- Add `meta.title` to each named route + a single `router.afterEach` guard that sets `document.title` to `Trinity — <label>`.
- Agent-scoped routes carry the agent name — `Trinity — <name>`, `· Workspace`, `· Execution` (the `:name` param is Trinity's canonical agent identifier). Unlabeled routes (redirects, catch-all) fall back to the branded `Trinity — Agent Orchestration` default.

## Changes
- `src/frontend/src/router/index.js` — `meta.title` per named route (string or fn of the resolved route) + `afterEach` title guard. All existing meta keys preserved.
- `src/frontend/e2e/browser-tab-titles.spec.js` — new `@smoke` Playwright spec: SPA nav updates the title per route, and a legacy redirect resolves to its destination's title.

## Scope notes
- **Frontend-only** — no backend / auth / CSP surface. `document.title` is a plain-text sink (no XSS); agent names are validated regardless.
- `dev`'s router has no Brain Orb route yet (that's unmerged `feature/61` work) — its title lands with that branch.

## Test Plan
- [x] `vite build` compiles the router + full app.
- [x] Title-resolution logic asserted across 8 cases (static / dynamic agent / sub-pages / public / fallback) — all pass.
- [x] Playwright spec discovered + parses (`--list`).
- [ ] CI `frontend-e2e` (`ui` label) runs the `@smoke` titles spec against the live stack.
- [ ] Manual: open several tabs (Dashboard, an agent, Settings) → each shows a distinct `Trinity — …` title.

Fixes abilityai/trinity#1418

### DIFF
--- a/src/frontend/e2e/browser-tab-titles.spec.js
+++ b/src/frontend/e2e/browser-tab-titles.spec.js
@@ -1,0 +1,37 @@
+import { test, expect } from '@playwright/test'
+
+// #1418 — per-route browser-tab titles. The router's `afterEach` guard sets
+// `document.title` to `Trinity — <label>` from each route's `meta.title`;
+// unlabeled routes (redirects, catch-all) fall back to the branded default
+// `Trinity — Agent Orchestration` that index.html ships (#1416).
+//
+// Dynamic agent-scoped titles (`meta.title` as a fn of `to.params.name`) are
+// covered by unit logic; here we assert the deterministic static/redirect
+// routes and — critically — that CLIENT-SIDE nav updates the title with no
+// full reload (the whole point of the guard vs a static <title>).
+//
+// @smoke: runs in the `frontend-e2e` CI workflow on every `ui`-labelled PR.
+test.describe('browser tab titles (#1418)', () => {
+  test('@smoke SPA navigation updates the tab title per route', async ({ page }) => {
+    await page.goto('/')
+    await expect(page.getByRole('link', { name: 'Dashboard', exact: true }))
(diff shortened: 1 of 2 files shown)


dolho:
### Review — per-route tab titles (automated review, Claude)

Clean, well-scoped. `afterEach` firing once on the final destination (so redirects resolve to the destination title) is the right hook, dynamic `meta.title` as a fn of `to.params.name` is neat, and the branded fallback covers unlabeled routes. `document.title` is a plain-text sink so the agent-name interpolation carries no XSS. The `@smoke` spec asserts the thing that matters (client-side nav repaints the title with no reload) plus the redirect case.

**One cross-PR note:** this edits `src/frontend/src/router/index.js`, as does #1133 (vue-router 4→5). Your `AgentWorkspace` `beforeEnter` still uses the `next()` callback signature, which is **removed in Vue Router 5** — so if #1133 merges first, this file will conflict *and* the `next()` guard will need converting to return-style. Suggest sequencing after #1133 and rebasing onto the v5 guard form (or coordinating the merge order).

Otherwise LGTM pending green `frontend-e2e`.

dolho (approved):

PR #1429, opened 2026-07-03
fix(db): serialize Alembic PG migrations with pg_advisory_lock (#1425)

### LINKED ISSUE #1425: Alembic multi-worker deadlock on PG first boot (migration_lock did not serialize)
## Symptom
During an ops `/update` of instance **eu2** (PostgreSQL backend, #300) from v0.6.1 (`0edd1dbe`) to v0.7.0 (`d4741d25`), the first backend boot on the Alembic-on-PG path (#1183/#1186) logged one full traceback:

```
psycopg2.errors.DeadlockDetected: deadlock detected
sqlalchemy.exc.OperationalError: (psycopg2.errors.DeadlockDetected) deadlock detected
  File "/app/migrations/versions/0003_agent_compatibility_results.py", line 26, in upgrade
    op.execute(...)
  File "/app/db/alembic_runner.py", line 61, in upgrade_to_head
```

## Analysis
Two uvicorn workers appear to have entered `upgrade_to_head()` concurrently and deadlocked inside migration `0003_agent_compatibility_results`; the losing worker aborted, the winner completed. Final state is correct: `alembic_version = 0010_agent_ownership_mcp_exposed` (head), backend healthy, one-time occurrence.

#1263 added the cross-process `migration_lock.py` for exactly this multi-worker upgrade race (observed on luminous, SQLite path). On the **PostgreSQL** path it evidently did not fully serialize the workers — either the lock isn't engaged on the Alembic runner path or the lock scope doesn't cover `command.upgrade()`.

## Impact
Benign this time (self-resolved), but a deadlock mid-migration on a less idempotent revision could leave a partially applied migration with one worker crashed.

## Repro
Multi-worker uvicorn + `DATABASE_URL=postgresql://...` + pending Alembic revisions (v0.6.1 → v0.7.0 jump, revisions 0001–0010 applied on one boot).

## Environment
- Instance: eu2 (ops-managed, prod compose), PG via `/migrate-to-postgres` (#300)

dolho:
## Summary

The PostgreSQL/Alembic boot path called `upgrade_to_head()` **outside** the cross-process migration lock. The `flock` in `db/migration_lock.py` (#1160) only wraps the **SQLite** branch of `init_database()`, so on PG two uvicorn workers could both enter `command.upgrade()` concurrently and **deadlock inside a revision** — exactly what #1425 observed on `eu2` during a v0.6.1→v0.7.0 upgrade (migration `0003`, one worker crashed, winner completed).

## Fix

Take a **PostgreSQL session advisory lock** (`pg_advisory_lock`, fixed key) around the whole stamp+upgrade inside `upgrade_to_head()`:

- **Right primitive for PG:** unlike the SQLite `flock` (per-open-file, shared-local-FS only), `pg_advisory_lock` is cross-connection and keyed by number, so it serializes the migration across the worker/scheduler sessions regardless of filesystem. PostgreSQL releases a session advisory lock automatically on connection close, so a crashed holder never leaves a stale lock — same safety property the flock has.
- **Self-contained** in the runner: correct no matter which caller invokes it, and doesn't depend on `DB_PATH` being meaningful on a PG deploy.
- **Fails open** (mirrors `migration_lock`) and **skips non-PG** server backends (dialect-guarded → prior behaviour).
- **Closes a secondary race:** the fresh-vs-pre-Alembic detection + baseline `stamp` now run *inside* the lock, so two workers can't both stamp the baseline.

Single-caller behaviour is unchanged (lock acquired+released around the same logic), so the existing lifecycle tests still hold.

## Tests

- `tests/unit/test_1425_alembic_pg_lock.py` (**4, no DB — verified passing** in `trinity-backend:latest`): advisory lock acquired **before** `command.upgrade` and released **after**, on the same fixed key; the pre-Alembic `stamp` runs under the lock; a non-PG backend issues no PG SQL; the key fits signed bigint.
- `tests/integration/test_alembic_postgres.py`: adds a 2-thread concurrent-boot **no-deadlock** proof (skip-gated on a live PG; runs in the `pg-migrations` CI job).

## Docs
- `docs/memory/architecture.md` — Invariant #3 now notes the PG advisory-lock serialization alongside the SQLite flock.

## Verification
```
# in trinity-backend:latest
python -m pytest tests/unit/test_1425_alembic_pg_lock.py -v   # 4 passed
# with a live PG (pg-migrations CI):
TEST_POSTGRES_URL=postgresql://... pytest tests/integration/test_alembic_postgres.py -v
```

Fixes #1425


vybe (approved):
Validated via /validate-pr: security greps clean, architecture.md Invariant #3 updated, named regression tests (unit ordering proof + live-PG concurrency test), CI fully green. Fixed process gaps: closing keyword (Fixes #1425) + issue labels.

Issue #1023, opened 2026-06-02
refactor(backend): unified Redis sliding-window rate limiter (replace ad-hoc per-endpoint limiters)


dolho:
## Summary

Replace the ad-hoc, per-endpoint rate limiters with a **single Redis sliding-window primitive**, per `docs/planning/TARGET_ARCHITECTURE.md` §"Backend → What Changes" ("Rate limiting: Redis-backed sliding window replaces ad-hoc per-endpoint implementations. One middleware, one configuration, consistent behavior across all endpoints").

Today each surface re-implements its own **fixed-window** limiter:
- `routers/webhooks.py` — per-token `INCR`+`EXPIRE` (10/60s) + a bespoke in-process fallback bucket.
- `routers/auth.py` — per-account + per-IP login limiter and a separate OTP-attempt limiter, also `GET`/`INCR`/`EXPIRE`.
- Plus scattered per-recipient/proactive limits elsewhere.

These share no code, drift in behavior, and use **fixed windows** (a burst at a window boundary can pass 2× the limit). One audited sliding-window implementation fixes the correctness gap and deletes the duplication.

## Proposed approach

**New `services/rate_limiter.py`** — one Redis **sorted-set sliding window**:
- `ZADD key <now> <unique>` → `ZREMRANGEBYSCORE key 0 <now-window>` → `ZCARD key` → compare to limit, `EXPIRE key window`. Pipelined/atomic. True rolling window, no boundary burst.
- Single public API: `check(key, limit, window_seconds) -> RateLimitResult{allowed, remaining, retry_after}`; raises/returns a 429 with `Retry-After` consistently.
- **Fail-open** when Redis is unavailable (preserves current webhook/auth philosophy), with the existing bounded in-process fallback lifted out of `webhooks.py` and shared.
- Cached module-level Redis client (reuse the hardened pattern already in `routers/webhooks.py`: cache + reset-on-error, ACL-aware logging).

**Migrate existing callers onto it** (delete their bespoke code):
- `routers/webhooks.py` — per-token bucket → `rate_limiter.check(f"webhook:{token}", 10, 60)`.
- `routers/auth.py` — login (per-account + per-IP) and OTP limiters → shared primitive with their existing limits/windows preserved.

## Scope decision (honest)

[...]

Issue #1056, opened 2026-06-03
feat(voice): VoIP telephony — agents place/receive phone calls over Gemini Live


vybe:
## Summary

Add **VoIP / phone-call** capability to agents: an agent can place an outbound phone call to a user (Phase 1) and optionally receive inbound calls (Phase 2), holding a real-time, interruptible voice conversation powered by the existing Gemini Live voice mode.

This ships as a **regular OSS feature** in the public backend — no entitlement gating, no enterprise module. The telephony transport reuses Trinity's existing Twilio integration and the existing Gemini Live bridge.

> **Revised 2026-06-04** (was "Enterprise: VoIP telephony …"): de-scoped from an enterprise/open-core module to a standard OSS feature; interruption delegated to Gemini Live's native behavior (no custom barge-in code, no changes to `gemini_voice.py`); per-agent Twilio **voice** credentials get their **own dedicated config table** (not the WhatsApp binding).

## Context

Trinity already has browser-based voice mode (`VOICE_MODEL=models/gemini-3.1-flash-live-preview`) in `src/backend/services/gemini_voice.py` + `src/backend/routers/voice.py`. That bridge is **audio-source-agnostic**: it consumes a PCM-16kHz input queue (`voice_service.send_audio(session_id, bytes)`) and emits PCM-24kHz output via an `on_audio_out(bytes)` callback, and already handles bidirectional Gemini Live streaming, native interruption, tool execution (`run_task` + panel tools), transcript persistence to `chat_messages`, and Redis-backed cross-worker session metadata.

A phone call is just a **different transport feeding the same queues**. The second half also already exists: Trinity integrates **Twilio** for WhatsApp (`src/backend/adapters/whatsapp_adapter.py`, `src/backend/db/whatsapp_channels.py`) with `twilio.request_validator` signature validation, AES-256-GCM per-agent token storage, and Cloudflare Tunnel public ingress. `twilio` (9.10.5) and `httpx` are already backend dependencies; `audioop` is stdlib on Python 3.11. VoIP reuses all of this.

[...]

PR #1431, opened 2026-07-03
fix(webhooks): pre-auth per-IP rate limit + request-body cap (#1424)

### LINKED ISSUE #1424: bug(webhooks): public webhook endpoint has no pre-auth/per-IP rate limit or request-body cap
## Summary

The public, unauthenticated webhook endpoint `POST /api/webhooks/{token}` applies its rate limit **only after** a valid token resolves (keyed `webhook:{token}`). A request with a well-formed-but-unknown 43-char token passes the regex, performs a DB lookup, and returns 404 — with **no throttle**. There is no per-IP / pre-auth rate limit and no global rate-limit middleware yet, so an unauthenticated client can flood the endpoint with unlimited token-lookup requests. Separately, `await request.body()` reads the full request body with no size cap on this unauthenticated route.

Neither is exploitable to *forge* a trigger (the 43-char `secrets.token_urlsafe(32)` token is unguessable), but both are unauthenticated resource-amplification / DoS surfaces on a world-reachable endpoint.

## Context

Found while auditing the public-webhook feature. Architecture notes that a global ASGI rate-limit middleware with a route→policy table is a tracked follow-up; until then this endpoint has no floor. The shared sliding-window limiter (`services/rate_limiter.py`, #1023) already exists and can be applied per-IP before the token lookup.

Relevant code:
- `src/backend/routers/webhooks.py` — `trigger_webhook()`; limiter call sits after token resolution; `await request.body()` has no size guard.
- `src/backend/services/rate_limiter.py` — the shared limiter to reuse.

Related: unified rate limiter #1023, prior webhook rate-limit TOCTOU fix #644, original feature #291.

## Acceptance Criteria

- [ ] A per-IP (or global) rate limit is enforced **before** the token DB lookup, so a flood of invalid-token requests is throttled (429 + `Retry-After`), reusing `services/rate_limiter.py`.
- [ ] The existing per-token limit is preserved for valid tokens.
- [ ] The endpoint rejects an over-sized request body (413 or equivalent) before reading/parsing it fully — pick a sane cap (the optional `context` is already ≤4000 chars).
- [ ] Limiter stays fail-open (Redis down must not block legitimate triggers).
- [ ] Tests in `tests/test_webhook_triggers.py` (or `tests/test_ip_rate_limit_fix.py`): invalid-token flood is throttled; oversized body is rejected.

## Technical Notes

Add an IP-scoped `rate_limiter.enforce("webhook_ip:{caller_ip}", ...)` at the top of `trigger_webhook`, before the regex/DB steps. Consider a `Content-Length` guard before `request.body()`.

dolho:
## Summary

The public, unauthenticated `POST /api/webhooks/{token}` applied its rate limit **only after** a valid token resolved (keyed `webhook:{token}`). A request with a well-formed-but-unknown 43-char token passed the regex, hit the DB, and 404'd **with no throttle** — an unauthenticated resource-amplification / DoS surface on a world-reachable route. Separately, `request.body()` read the full body with no size cap.

Neither forges a trigger (the `token_urlsafe(32)` token is unguessable), but both are unauthenticated DoS surfaces.

## Fix (`src/backend/routers/webhooks.py`)

- **Pre-auth per-IP limit** at the TOP of `trigger_webhook`, **before** the regex/DB lookup, so an invalid-token flood is throttled (429 + `Retry-After`). Reuses the shared sliding-window limiter (`services/rate_limiter.py`, #1023) and the trusted-proxy-aware `_get_client_ip` (`routers.public`, the same helper `files.py` already reuses) so `X-Forwarded-For` can't be spoofed to bypass it and legit callers behind nginx don't collapse into one bucket. Generous (60/min) vs the per-token limit; **fail-open**.
- **Body size cap:** drop the typed body param and read `request.stream()` with a hard ceiling — Content-Length fast-reject **plus** bail at cap+one chunk, so a missing/lying/chunked length can't force unbounded buffering. Oversized → **413 before any parse**. The optional JSON is then parsed manually (422 on bad body). The same bytes feed the idempotency key (single read — replaces the prior second `request.body()`).
- Per-token limit and the existing 422-on-oversized-`context` behaviour are preserved.

## Acceptance criteria
- [x] Per-IP limit enforced **before** the token DB lookup, reusing `rate_limiter` (429 + Retry-After).
- [x] Existing per-token limit preserved for valid tokens.
- [x] Oversized body rejected (413) before it is read/parsed fully (16 KiB cap, env-overridable).
- [x] Limiter stays fail-open (Redis down never blocks legit triggers).
- [x] Tests for invalid-token flood + oversized body.

## Tests
- `tests/unit/test_1424_webhook_hardening.py` — **4, TestClient, no live server, verified passing** in `trinity-backend:latest`: IP flood throttled **before** the DB lookup (asserts the lookup count to prove ordering) + `Retry-After`; per-token limit still engages; 413 on oversized; small body clears the gate.
- `tests/test_webhook_triggers.py` — live smoke for the 413 and invalid-token-flood (429) paths.

## Verification
```
# in trinity-backend:latest, no live stack needed
python -m pytest tests/unit/test_1424_webhook_hardening.py -v   # 4 passed
```

Fixes #1424

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -214,6 +214,11 @@ VOIP_INTENT_TTL_SECONDS=180       # staged Gemini-session intent TTL
 # Agent structured reports (#918)
 REPORT_RATE_LIMIT=30              # max reports an agent may create per 60s window
 
+# Public webhook triggers (#1023, #1424) — 60s windows fixed in code
+WEBHOOK_RATE_LIMIT=10             # triggers per token per 60s
+WEBHOOK_IP_RATE_LIMIT=60          # pre-auth requests per IP per 60s (unknown-token flood guard)
+WEBHOOK_MAX_BODY_BYTES=16384      # request-body size cap in bytes (413 over)
+
 # ===========================================
 # SERVICE URLS (Usually no need to change)
 # ===========================================
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -97,6 +97,11 @@ services:
       # Agent structured reports (#918) — per-agent create-rate cap (60s window fixed in code).
       # Prod compose launches standalone (no base merge / env_file), so wire it here too.
       - REPORT_RATE_LIMIT=${REPORT_RATE_LIMIT:-30}               # reports per agent per 60s
+      # Public webhook trigger hardening (#1023, #1424) — 60s windows fixed in code.
+      # Prod compose launches standalone (no base merge / env_file), so wire it here too.
+      - WEBHOOK_RATE_LIMIT=${WEBHOOK_RATE_LIMIT:-10}             # triggers per token per 60s
+      - WEBHOOK_IP_RATE_LIMIT=${WEBHOOK_IP_RATE_LIMIT:-60}       # pre-auth requests per IP per 60s
+      - WEBHOOK_MAX_BODY_BYTES=${WEBHOOK_MAX_BODY_BYTES:-16384}  # request-body cap (413 over)
       - GITHUB_PAT=${GITHUB_PAT}
       # Host paths for volumes (used when creating agent containers)
       - HOST_TEMPLATES_PATH=${HOST_TEMPLATES_PATH:-${PWD}/config/agent-templates}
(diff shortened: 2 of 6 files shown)


vybe (approved):
Validated via /validate-pr: security greps clean (203.0.113.x is TEST-NET), named regression tests with non-happy-path coverage, reuses shared rate_limiter + trusted-proxy _get_client_ip. Pushed 9b9bd1e9 to close the #1056-class packaging gap (WEBHOOK_* env levers wired into both compose files + .env.example) and rephrased the issue ref to a closing keyword.

Issue #834, opened 2026-05-13
Soft-delete all entities + configurable data retention policy


vybe:
## Summary

Replace hard-delete operations platform-wide with soft-delete (setting `deleted_at`) and introduce a configurable, independently-running data retention policy that controls when soft-deleted records are permanently purged.

## Context

Currently Trinity hard-deletes most entities (agents, schedules, users, etc.) immediately on request. This causes two problems:

1. **No recovery window** — accidental agent deletion loses all associated schedules, sharing config, permissions, and metadata. The container is gone too, but the relational state is unrecoverable.
2. **Operator has no control over data lifecycle** — Trinity decides when data disappears. Operators running sovereign infrastructure should own that decision.

The target architecture (docs/planning/TARGET_ARCHITECTURE.md) already describes PostgreSQL with partitioned tables and configurable retention windows per entity type. This issue formalizes the pattern across the platform.

The existing `cleanup_service.py` already runs retention sweeps for executions, health checks, and audit log rows — this issue generalizes that into a unified, operator-configurable policy.

## Acceptance Criteria

- [ ] `deleted_at TEXT` column added to key entity tables: `agent_ownership`, `agent_schedules`, `users`, `agent_shared_files`, `agent_sessions`, `chat_sessions`
- [ ] All DELETE operations on those tables set `deleted_at = utc_now_iso()` instead of removing the row
- [ ] All read queries (list, get, joins) filter `WHERE deleted_at IS NULL` by default
- [ ] Retention policy is configurable per entity type via `system_settings` table (e.g. `agent_soft_delete_retention_days`, default 30)
- [ ] The existing `cleanup_service.py` retention sweep is extended to hard-delete soft-deleted rows past their retention window (same 5000-row/cycle cap and WAL checkpoint pattern)
- [ ] `GET /api/agents/{name}` returns 404 (not a deleted-agent response) — soft-delete is transparent to API consumers
- [ ] Agent name remains reserved for the retention period (cannot create a new agent with the same name until the soft-deleted row is purged)
- [ ] Admin endpoint or settings UI surface to list soft-deleted agents with their purge date
- [ ] Schema migration (Alembic-compatible DDL) for all new columns

## Technical Notes

**Priority entities for Phase 1:** `agent_ownership` (most impactful — agent recovery), `agent_schedules`, `users` (deactivate vs delete).

[...]

PR #1433, opened 2026-07-03
fix(webhooks): public webhook won't fire schedules of soft-deleted agents (#1423)

### LINKED ISSUE #1423: bug(webhooks): public webhook trigger fires schedules of soft-deleted agents (missing agent_ownership guard)
## Summary

The public webhook trigger path resolves a schedule token via `AgentScheduleOperations.get_schedule_by_webhook_token()`, which filters only on `agent_schedules.deleted_at IS NULL`. It does **not** check whether the schedule's **agent** is soft-deleted (`agent_ownership.deleted_at`). So a webhook can still fire a schedule belonging to a soft-deleted agent, re-opening exactly the hole the cron path guards against.

The cron path (`list_all_enabled_schedules()`) deliberately JOINs `agent_ownership` and filters `agent_ownership.deleted_at IS NULL`, with a docstring stating that otherwise "the scheduler fires every enabled schedule for a soft-deleted agent and writes a `schedule_executions` failure row per tick until purge." The webhook lookup skips that guard, so calling the token during the soft-delete window (default retention 180 days) fires executions against a deleted/recoverable agent — violating the documented soft-delete invariant that schedules stop firing immediately.

## Context

Found while auditing the public-webhook feature. Agent soft-delete (#834) preserves child schedule rows (their own `deleted_at` stays NULL), relying on read paths to join `agent_ownership` — which the webhook token lookup fails to do.

Relevant code:
- `src/backend/db/schedules.py` — `get_schedule_by_webhook_token()` (schedule-only filter) vs `list_all_enabled_schedules()` (correct agent-ownership join, a few methods above — use it as the reference).
- `src/backend/routers/webhooks.py` — consumes the token lookup.

Related: soft-delete/retention #834, original feature #291 (WEBHOOK-001).

## Acceptance Criteria

- [ ] `get_schedule_by_webhook_token()` returns no schedule when the owning agent is soft-deleted (`agent_ownership.deleted_at IS NOT NULL`), mirroring `list_all_enabled_schedules()`.
- [ ] `POST /api/webhooks/{token}` for a soft-deleted agent's schedule returns 404 (consistent with the "token not found" response) and creates **no** `schedule_executions` row.
- [ ] Recovering the agent restores webhook triggering.
- [ ] Regression test in `tests/test_webhook_triggers.py`: soft-delete the agent → webhook call → 404, no execution row; recover → webhook fires.

## Technical Notes

One-line parity fix: add the `agent_ownership` join + `agent_ownership.c.deleted_at.is_(None)` predicate to the `get_schedule_by_webhook_token()` select, matching `list_all_enabled_schedules()`.

dolho:
## Summary

`get_schedule_by_webhook_token()` filtered only `agent_schedules.deleted_at IS NULL` — it did **not** check whether the schedule's **agent** was soft-deleted (`agent_ownership.deleted_at`). So `POST /api/webhooks/{token}` could still fire a schedule belonging to a soft-deleted (recoverable) agent throughout the retention window (default 180 days), re-opening exactly the hole the cron path guards against and violating the #834 soft-delete invariant that *schedules stop firing immediately*.

The cron path (`list_all_enabled_schedules()`, a few methods above) already JOINs `agent_ownership` and filters `agent_ownership.deleted_at IS NULL`; the webhook lookup just skipped that guard.

## Fix (`src/backend/db/schedules.py`)

One-method parity fix: add the `agent_ownership` join + `agent_ownership.c.deleted_at.is_(None)` predicate to `get_schedule_by_webhook_token()`, mirroring `list_all_enabled_schedules()`. A soft-deleted agent's token now returns **no schedule → the router 404s before any scheduler dispatch**, so no `schedule_executions` row is created. Admin recovery (`agent_ownership.deleted_at → NULL`) restores triggering.

## Acceptance criteria
- [x] `get_schedule_by_webhook_token()` returns nothing when the owning agent is soft-deleted.
- [x] `POST /api/webhooks/{token}` for a soft-deleted agent's schedule → 404, no execution row (404 is before dispatch).
- [x] Recovering the agent restores webhook triggering.
- [x] Regression test in `tests/test_webhook_triggers.py`.

## Verification — live, against a running backend
```
live webhook            -> 202
soft-delete agent       -> 200
webhook (soft-deleted)  -> 404   <-- the fix
recover agent           -> 200
webhook (recovered)     -> 202   <-- fires again
```
Plus `tests/test_webhook_triggers.py::TestWebhookTrigger::test_soft_deleted_agent_webhook_returns_404_then_recovers` (uses a throwaway agent so the shared fixture is untouched).

Related to #1423

Closes #1423

### DIFF
--- a/src/backend/db/schedules.py
+++ b/src/backend/db/schedules.py
@@ -831,11 +831,29 @@ class ScheduleOperations:
         return token
 
     def get_schedule_by_webhook_token(self, token: str) -> Optional[Schedule]:
-        """Look up a schedule by its webhook token (O(1) via index)."""
-        stmt = select(agent_schedules).where(
-            and_(
-                agent_schedules.c.webhook_token == token,
-                agent_schedules.c.deleted_at.is_(None),
+        """Look up a schedule by its webhook token (O(1) via index).
+
+        Applies the same two soft-delete filters as `list_all_enabled_schedules`
+        (#834, #1423): skip a soft-deleted *schedule* (`agent_schedules.deleted_at`)
+        AND a schedule whose *agent* is soft-deleted (`agent_ownership.deleted_at`).
+        Without the agent join, a webhook could still fire a schedule of a
+        soft-deleted (recoverable) agent during the retention window — the exact
+        hole the cron path guards against.
+        """
+        stmt = (
+            select(agent_schedules)
+            .select_from(
+                agent_schedules.join(
+                    agent_ownership,
+                    agent_ownership.c.agent_name == agent_schedules.c.agent_name,
+                )
+            )
+            .where(
+                and_(
+                    agent_schedules.c.webhook_token == token,
+                    agent_schedules.c.deleted_at.is_(None),
+                    agent_ownership.c.deleted_at.is_(None),
+                )
             )
         )
         with get_engine().connect() as conn:
--- a/tests/test_webhook_triggers.py
+++ b/tests/test_webhook_triggers.py
@@ -273,3 +273,52 @@ class TestWebhookTrigger:
             assert resp.status_code == 404
         finally:
             _delete_schedule(api_client, test_agent_name, sid)
+
+    def test_soft_deleted_agent_webhook_returns_404_then_recovers(
+        self, api_client: TrinityApiClient
+    ):
+        """#1423: a webhook must not fire a schedule whose *agent* is soft-deleted.
+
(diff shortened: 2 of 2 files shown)


vybe (approved):
Validated via /validate-pr: webhook token lookup now applies the same two soft-delete filters as the cron path; named regression test covers 404-then-recover. All required checks green, security sweep clean.

PR #1436, opened 2026-07-03
fix(scheduler): soft-deleted agent's schedules stop firing on the sync path (#1427)

### LINKED ISSUE #1427: Soft-deleted agent's enabled schedule keeps firing (ConnectError every cron tick) — #834 cascade gap
## Symptom
On instance **eu2**, agent `ff-scorer` was soft-deleted on 2026-06-29 (`agent_ownership.deleted_at` set, container removed). Its schedule `ff-auto` (`55 10-17 * * 1-5`) remained `enabled=1` with `deleted_at IS NULL`, so the scheduler kept dispatching it into a nonexistent container — **26 failed executions (`HTTP error: ConnectError`) between 2026-06-29 and 2026-07-02**, polluting failure analytics.

## Expected
Per #834 the cascade *preserves* child rows (fine for recovery), but a schedule belonging to a soft-deleted agent should not be **live**: either the cascade should set `agent_schedules.deleted_at` (recovery already restores metadata), or the scheduler/dispatch path should skip schedules whose parent agent has `deleted_at IS NOT NULL`.

## Workaround applied (ops)
`PUT /api/agents/ff-scorer/schedules/{id}` with `{"enabled": false}` — accepted even for the soft-deleted agent (which is itself a bit surprising) and stopped the spam.

## Environment
- Instance: eu2, v0.7.0 `d4741d25` (also reproduced on v0.6.1 — failures began at deletion time, 2026-06-29)

dolho:
## Summary

On eu2, agent `ff-scorer` was soft-deleted (`agent_ownership.deleted_at` set, container removed) but its schedule `ff-auto` stayed `enabled=1` / `deleted_at IS NULL`, so the scheduler kept dispatching it into the nonexistent container — **26 failed executions (`HTTP error: ConnectError`)** over three days, polluting failure analytics.

## Root cause

Soft-deleting an agent (#834) leaves the child schedule row untouched (enabled=1, deleted_at NULL, **same `updated_at`**). The standalone scheduler's live-sync (`_sync_agent_schedules`) reads `list_all_schedules()`, whose query filtered only the schedule's **own** `deleted_at` — **no `agent_ownership` join**. So the transition diff saw no change and never removed the already-registered APScheduler job; it kept firing every cron tick.

`list_all_enabled_schedules()` (the *startup* load) already joins `agent_ownership` and skips these — the gap was only on the live-sync path a running scheduler uses after the agent is deleted mid-run.

## Fix (`src/scheduler/database.py`)

`list_all_schedules()` now `LEFT JOIN`s `agent_ownership` and reports a schedule as **disabled** when its parent agent is soft-deleted. This makes the existing `enabled→disabled` transition in `_sync_agent_schedules` **remove the job** — and because agent recovery clears `agent_ownership.deleted_at`, the `disabled→enabled` transition **re-adds** it. Symmetric with #834 recovery, **no schedule-row mutation, no recovery churn**. The method is called only by the sync loop, so the effective-enabled override is contained to the sync view.

Companion to #1423 (same #834 gap on the webhook trigger path).

## Acceptance criteria
- [x] A soft-deleted agent's enabled schedule stops firing (job removed on next sync).
- [x] Recovering the agent restores firing.
- [x] Child rows preserved (recoverable per #834) — no `agent_schedules.deleted_at` set.

## Verification
`tests/scheduler_tests/test_1427_soft_deleted_agent_schedules.py` (4, against the real `SchedulerDatabase` + schema):
- live agent → schedules report enabled;
- soft-deleted agent → enabled schedules report **disabled** (→ sync removes jobs) while the DB rows stay `enabled=1`;
- recovery (clear `deleted_at`) → re-enabled;
- parity with `list_all_enabled_schedules` (startup set already excludes them).

*Note:* verified at the sync-view/DB layer (the exact fix point); the full running-scheduler loop reconciliation (`enabled→disabled` ⇒ `_remove_job`) is pre-existing, covered behavior. The issue also flagged that `PUT .../schedules/{id}` is accepted for a soft-deleted agent — left as-is (out of scope; not the firing bug).

Related to #1427

Closes #1427

### DIFF
--- a/src/scheduler/database.py
+++ b/src/scheduler/database.py
@@ -265,16 +265,37 @@ class SchedulerDatabase:
     def list_all_schedules(self) -> List[Schedule]:
         """List all schedules (enabled and disabled) for sync detection.
 
-        Excludes soft-deleted (#834 Phase 1b).
+        Excludes schedules that are themselves soft-deleted (#834 Phase 1b).
+
+        A schedule whose *parent agent* is soft-deleted (`agent_ownership.
+        deleted_at`) is reported as **disabled** here (#1427): soft-deleting an
+        agent leaves the child schedule row untouched (enabled=1, deleted_at
+        NULL, same updated_at), so the sync's transition diff saw no change and
+        never removed the already-registered APScheduler job — it kept firing
+        into a nonexistent container (ConnectError every cron tick). Folding the
+        agent's soft-delete into the reported `enabled` makes the existing
+        enabled→disabled transition remove the job, and — because agent recovery
+        clears `agent_ownership.deleted_at` — the disabled→enabled transition
+        re-adds it. Symmetric with #834 recovery; no schedule-row mutation.
+        `list_all_enabled_schedules` already applies this join at startup; this
+        closes the same gap on the live-sync path.
         """
         with self.get_connection() as conn:
             cursor = conn.cursor()
             cursor.execute("""
-                SELECT * FROM agent_schedules
-                WHERE deleted_at IS NULL
-                ORDER BY agent_name, name
+                SELECT s.*, ao.deleted_at AS _agent_deleted_at
+                FROM agent_schedules s
+                LEFT JOIN agent_ownership ao ON ao.agent_name = s.agent_name
+                WHERE s.deleted_at IS NULL
+                ORDER BY s.agent_name, s.name
             """)
-            return [self._row_to_schedule(row) for row in cursor.fetchall()]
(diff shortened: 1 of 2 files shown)


vybe (approved):
Validated via /validate-pr: surgical LEFT-JOIN fix on the sync read path, rows untouched (recovery-symmetric with #834), named regression tests incl. recovery + startup-path parity. All required checks green, security sweep clean.

Issue #525, opened 2026-04-26
feat: idempotency keys at all execution trigger boundaries (RELIABILITY-006)


vybe:
## Summary

Add an `Idempotency-Key` contract at every producer boundary that creates executions: `/api/internal/...` (scheduler), `/api/webhooks/{token}` (#291), MCP `chat_with_agent` and `fan_out`, sync `/chat`, async `/task`, event subscriptions, and self-execute. Backend stores `(key → execution_id)` for 24 h and short-circuits duplicates with the original execution result instead of dispatching twice.

## Context

From the orchestration architecture review (2026-04-26):

> No idempotency at the producer boundary. Webhook re-deliveries, MCP client retries, scheduler-to-backend network blips — all create duplicate executions. Unifying every trigger through one funnel (`TaskExecutionService`) actually makes this worse, because the funnel doesn't dedupe — it just trusts the producer.

This is most acute for webhooks (external systems retry on 5xx and on perceived timeouts) and scheduler dispatch (network blip between scheduler container and backend creates phantom retries). Today the only mitigation is "be a well-behaved producer," which external webhook senders are not.

## Acceptance Criteria

- [ ] `Idempotency-Key` HTTP header accepted on: `/api/agents/{name}/chat`, `/api/agents/{name}/task`, `/api/internal/agents/{name}/execute-task`, `/api/webhooks/{token}`, MCP-originated executions (`chat_with_agent`, `fan_out`), event-subscription dispatch.
- [ ] New `idempotency_keys` table: `(key, scope, execution_id, created_at, response_snapshot)` with 24 h TTL and a unique constraint on `(scope, key)`.
- [ ] On duplicate key: return the original execution_id and (if execution finished) the original result, with HTTP 200 + `X-Idempotent-Replay: true` header. Do not dispatch a second execution.
- [ ] Webhook trigger auto-derives a key from `(webhook_token, body_hash)` if the caller doesn't supply one — covers naive webhook senders that retry without idempotency awareness.
- [ ] Scheduler dispatch supplies a deterministic key derived from `(schedule_id, fire_time)` so a transient backend 5xx + APScheduler retry never produces duplicates.
- [ ] Audit log entry for every idempotent replay (so duplicate-storms are observable).
- [ ] Documented in architectural invariants — every new trigger type must accept idempotency keys before merging.

## Technical Notes

[...]

Issue #291, opened 2026-04-10
feat: Webhook triggers for agent schedules (WEBHOOK-001)


vybe:
## Summary

Add webhook URL endpoints that trigger agent schedule executions from external systems. Each schedule can optionally expose a unique webhook URL with an embedded HMAC signature and expiration date, allowing external services (CI/CD, form submissions, CRMs, monitoring alerts) to initiate agent work without API authentication.

## Motivation

Currently agents can only be triggered via: UI chat, cron schedules, MCP tools, or Slack. There's no way for arbitrary external systems to trigger agent work via a simple HTTP POST. Webhooks are the universal integration pattern for this.

### Webhook URL Format
```
POST /api/webhooks/{schedule_id}/{signature}
```

- `signature` = HMAC-SHA256(schedule_id + expiration, secret_key), base64url-encoded
- URL is self-contained — no auth header needed
- Optional expiration date embedded in the signature payload

### Schedule Integration
- Each schedule gets an optional "webhook trigger" toggle
- When enabled, generates a unique webhook URL with configurable expiration (default: 90 days, or no expiry)
- Calling the webhook triggers the schedule exactly once (same as manual trigger)
- The schedule's message is used as the task prompt (webhook payload can optionally override/append context)
- Respects existing execution slot limits and queueing

### Webhook Payload (optional body)
```json
{
  "context": "Additional context appended to the schedule message",
  "metadata": { "source": "github", "event": "push" }
}
```
- `context` is appended to the schedule's message as additional input
- `metadata` is stored on the execution record for traceability

### Security
- HMAC-SHA256 signature prevents URL guessing/tampering
- Expiration date prevents stale URLs from being used indefinitely
- Rate limiting per webhook URL (e.g., 10 calls/minute)
- Webhook URLs can be regenerated (invalidates old URL)
- Execution is logged with `triggered_by: "webhook"` for audit trail

### API Endpoints
| Method | Path | Auth | Description |
|--------|------|------|-------------|
| POST | `/api/webhooks/{schedule_id}/{signature}` | None (signature) | Trigger execution |
| POST | `/api/agents/{name}/schedules/{id}/webhook` | JWT | Generate/regenerate webhook URL |
| DELETE | `/api/agents/{name}/schedules/{id}/webhook` | JWT | Revoke webhook URL |
| GET | `/api/agents/{name}/schedules/{id}/webhook` | JWT | Get webhook URL and status |

### Database
Add columns to `agent_schedules`:
[...]

PR #1437, opened 2026-07-03
fix(webhooks): only dedupe on explicit Idempotency-Key — stop swallowing body-less triggers (#1422)

### LINKED ISSUE #1422: bug(webhooks): repeated identical/body-less webhook triggers are silently deduped for 24h
## Summary

The public webhook trigger endpoint (`POST /api/webhooks/{token}`, WEBHOOK-001) auto-derives an idempotency key of `sha256(token, raw_body)` when the caller does not supply an `Idempotency-Key` header. A completed idempotency claim lives for **24 hours**. As a result, **repeated webhook calls with an identical (or empty) body within a 24h window are silently deduplicated — only the first one fires an execution.** Every subsequent call returns `202` + `X-Idempotent-Replay: true` and does nothing.

This breaks the single most common webhook shape: a body-less "ping to trigger" (CI hook, monitoring alert, IFTTT, cron-from-an-external-system). Those senders emit the same/empty body every time, so only one trigger per day actually runs, while the `202` response makes the caller believe every trigger succeeded.

## Context

Found while auditing the public-webhook feature. The auto-derive is intended to absorb naive retries (same POST re-sent after a 5xx/timeout), and its docstring assumes "distinct intentional triggers carry distinct bodies." That assumption is false for the canonical webhook use case, where the body is empty or constant.

Relevant code:
- `src/backend/routers/webhooks.py` — auto-derive + `begin`/`complete` gate (the `idempotency_key or derive_webhook_key(...)` line).
- `src/backend/services/idempotency_service.py` — `derive_webhook_key(token, body)` = `sha256(token, body)`.
- `src/backend/db/idempotency.py` — `claim(..., ttl_hours=24)`; a completed row inside the window replays instead of firing.

Related: original feature #291 (WEBHOOK-001), idempotency #525 (RELIABILITY-006).

## Acceptance Criteria

- [ ] Two legitimate webhook calls with an identical/empty body, seconds apart, each fire a distinct execution (no silent replay).
- [ ] Genuine at-least-once retries (same POST re-sent after a transport error / 5xx) are still deduped — the anti-double-fire property of #525 is preserved.
- [ ] A caller-supplied explicit `Idempotency-Key` header still dedupes as today.
- [ ] Chosen approach documented in the endpoint docstring (e.g., only dedupe when an explicit `Idempotency-Key` is present; or incorporate a per-delivery nonce/time component into the auto-derived key so distinct deliveries differ while a true resend collides).
- [ ] Regression test in `tests/test_webhook_triggers.py` covering: (a) identical-body distinct deliveries → 2 executions, (b) explicit-key resend → 1 execution + `X-Idempotent-Replay`.

## Technical Notes

The safest minimal fix is likely to **stop auto-deriving from body alone**: dedupe only when the caller sends an explicit `Idempotency-Key`, and treat a body-less/no-key call as a fresh trigger every time. That inverts the current default but matches webhook semantics (at-least-once senders that care about dedup send a key). Whichever approach is chosen, keep the fail-open behavior (dedup layer must never block a real trigger).

dolho:
> ~~Stacked on #1431~~ — #1431 merged and this PR is retargeted to `dev`; the shared #1424 hunks are already on dev, so the effective delta is the #1422 idempotency change only.

## Summary

`POST /api/webhooks/{token}` auto-derived an idempotency key of `sha256(token, raw_body)` when the caller sent no `Idempotency-Key`. A completed claim lives **24h**, so repeated **identical/empty-body** calls within the window were silently deduped — only the first fired an execution; every later call returned `202` + `X-Idempotent-Replay: true` and did nothing. That breaks the single most common webhook shape — a body-less "ping to trigger" (CI hook, monitor, IFTTT, external cron) — which resends the same/empty body: one trigger per day actually ran while the `202` told the caller each succeeded.

## Fix (`src/backend/routers/webhooks.py`)

Stop auto-deriving the key from the body. **De-duplicate ONLY when the caller supplies an explicit `Idempotency-Key`**; a key-less call is a fresh, intentional trigger every time. At-least-once senders that need resend-dedup send a stable key — which preserves #525's anti-double-fire. The `begin`/`complete`/`fail` calls are guarded on the now-optional claim, the dedup layer stays **fail-open**, and the endpoint docstring documents the semantics.

## Acceptance criteria
- [x] Two identical/empty-body calls seconds apart each fire a distinct execution (no silent replay).
- [x] A caller-supplied explicit `Idempotency-Key` still dedupes as before (anti-double-fire preserved).
- [x] Genuine resends dedupe **when they carry a key** (the standard at-least-once contract; key-less naive resends now re-fire — the documented, intended trade-off per the issue's Technical Notes).
- [x] Approach documented in the endpoint docstring.
- [x] Regression tests in `tests/test_webhook_triggers.py`.

## Verification
- **Unit** — `tests/unit/test_1422_webhook_idempotency.py` (4, TestClient + real `idempotency_service` over temp SQLite, scheduler dispatch counted): body-less **and** identical-body repeats each dispatch (**2**); explicit-key resend dispatches **once** + replay header; distinct keys each dispatch.
- **Live** (rebuilt backend): two body-less pings → both `202`, **no** `X-Idempotent-Replay`, **2** execution rows; same explicit key → 2nd carries `x-idempotent-replay: true`.

Related to #1422 · original WEBHOOK-001 #291 · idempotency #525.

Fixes #1422


vybe (approved):
Validated via /validate-pr: explicit-key-only webhook dedup with named #1422 regression tests (unit + integration); env vars already wired in both compose files + .env.example. Resolved the stacked-branch overlap with the squash-merged #1431 (dev's #1445 test refactor wins; #1422 semantics intact). Approving.

Issue #953, opened 2026-05-26
Freshly deployed agents report M .gitignore against origin/main


vybe:
## Summary

Freshly deployed agents observe `M .gitignore` in their workspace immediately after deployment, before any user-driven sync. The drift pre-dates any push, doesn't block a fast-forward, and the file content looks "right" — but `git diff` against `origin/main` still reports a modification. Investigate the root cause and either fix the divergence at the source or normalize it so freshly-cloned workspaces present a clean `git status`.

## Context

The container's `startup.sh` ensures a few canonical entries land in `/home/developer/.gitignore` regardless of what the cloned repo already shipped:

```
# docker/base-image/startup.sh ~L116-120
grep -q ".local/" /home/developer/.gitignore || echo ".local/" >> /home/developer/.gitignore

# docker/base-image/startup.sh ~L378-383
grep -q "^content/$" /home/developer/.gitignore || echo "content/" >> /home/developer/.gitignore
grep -q "^\.local/$" /home/developer/.gitignore || echo ".local/" >> /home/developer/.gitignore
```

For agents whose workspace IS the trinity repo (the local "trinity agent" being the canonical case), the repo's checked-in `.gitignore` already contains `.local/`, `content/`, `.cache/`, `.claude.json`, `.trinity/`, `.sudo_as_admin_successful`, and friends. The `grep -q` guards should make the appends no-ops in that case — but field observation says we're seeing `M .gitignore` anyway.

Plausible root causes to verify:

1. **Pattern match mismatch** — `grep -q ".local/"` (no anchors) succeeds, but the second guard `grep -q "^\.local/$"` requires an exact line. If the committed `.gitignore` has a trailing comment or whitespace on the line containing `.local/`, the anchored grep fails and `echo` re-appends a duplicate line.
2. **Missing trailing newline** — if the committed `.gitignore` doesn't end in `\n`, `echo ".local/" >>` produces `last_linecontent/`, corrupting the previous line.
3. **Historical drift** — an older base image appended these entries before the repo committed them. The mods stuck; subsequent rebases preserve the divergence even though the appender is now a no-op.
4. **CRLF / encoding** — repo `.gitignore` could land with `\r\n` line endings under a misconfigured `core.autocrlf`, causing `grep -q "^content/$"` to miss the match and the appender to write a `\n`-terminated duplicate.

[...]

PR #1440, opened 2026-07-03
fix(agent): eliminate GitHub-template clone race + surface silent clone failures (#1439)

### LINKED ISSUE #1439: bug: GitHub-template agent deploy — race intermittently leaves a silent empty agent (clone into non-empty /home/developer, exit 128, not surfaced)
## Summary

Creating an agent from a GitHub template (`create_agent(template: "github:<owner>/<repo>")`)
intermittently produces a **silent empty agent**: the container comes up `running` and
`get_agent_health` returns `healthy`, but the workspace was never populated — no `.git`,
no `CLAUDE.md`, no `template.yaml`, no `.claude/skills/`. Two linked defects: a **race
condition** in the clone step (root cause) and a **silent-failure path** that never
surfaces it (compounding).

## Context

`docker/base-image/startup.sh` (git-sync clone path) empties the agent home and then
clones **into that same directory**:

- ~line 66: `rm -rf /home/developer/* /home/developer/.[!.]*`  (empty the home dir)
- ~line 78: `git clone <url> /home/developer`                  (clone into it)

A concurrent process populates `/home/developer` in the window between the `rm` and git's
empty-directory precondition check, so the clone intermittently aborts:

    fatal: destination path '/home/developer' already exists and is not an empty directory.
    Exit code: 128

**Evidence it is a race (not deterministic, not a stale image, not a PAT issue):**
- **Flaky on identical inputs:** 1 failure in 3 runs against the same base image + same repo.
- **Timing-flip:** a base image identical except for a passive `ls` inserted right before the
  clone succeeded every time — a read-only op that changes only timing can only affect a race.
- **Never fails in isolation:** running `startup.sh` under `docker run` (no entrypoint/backend
  concurrency) always empties cleanly and clones OK.

Leading suspect for the concurrent writer: the runtime's asynchronous named-volume population
(copying the base image's sizeable `/home/developer` into the fresh volume) still completing as
the entrypoint runs. The fix below is robust regardless of the exact writer.

**Silent-failure path (compounding):** on clone failure, `startup.sh` writes
`/home/developer/.git-clone-status` and **continues to "Agent ready"** (around lines 137–155) —
no abort. The backend **never reads `.git-clone-status`**, and health/monitoring never check
whether the identity clone succeeded. Net: `status: running`, `aggregate_status: healthy`, empty
workspace. Because the clone failure is flaky, GitHub deploys occasionally yield a healthy-looking,
identity-less agent that is hard to notice and non-reproducible.

Related prior work (same "silent empty agent" family, different paths): #218
(clone-not-executed), #950 (deploy_local silently creates empty agents).

## Steps to Reproduce

1. On an instance whose global GitHub PAT can read a repo containing a valid `template.yaml`
   + `.claude/skills/` (private is fine), call
   `create_agent(name, template: "github:<owner>/<repo>", source_branch: "main")`.
2. Repeat several times. Some runs fail: `get_agent_logs` shows the
   `exit 128 … 'already exists and is not an empty directory'` block; the workspace has no
   `.git`/`CLAUDE.md`; `get_agent_health` still returns `healthy`. Other runs succeed.
   (~1 in 3 failed during investigation.)

## Acceptance Criteria

- [ ] GitHub-template deploy is **not** sensitive to clone-vs-writer timing — repeated deploys
      never leave an empty workspace (clone no longer runs into a non-empty `/home/developer`).
- [ ] A genuine clone failure is **terminal and surfaced**: the agent reports `failed`/`unhealthy`
      (not `running`/`healthy`), and the git error is visible via `get_agent` / `get_agent_health`
      (backend consumes `.git-clone-status`).
- [ ] Regression coverage that a GitHub-template deploy reliably populates the workspace
      (`.git`, `template.yaml`, `.claude/skills/`).

## Technical Notes / Suggested Fix

1. **(fixes the race)** Never `git clone` into the live home dir. Either clone into a temp dir and
   move contents into `/home/developer`, or initialize in place:
   `git init && git remote add origin <url> && git fetch origin <branch> && git checkout -f <branch>`.
   Eliminates the empty-dir precondition, so a concurrent writer is harmless.
2. **(fixes the silent failure)** Make identity-clone failure terminal — surface the git error
   through `get_agent`/`get_agent_health`; have the backend consume `.git-clone-status`.
3. **(minor, separate follow-up)** `base_image_version` in the create response is set from
   `get_platform_version()` (`src/backend/services/agent_service/crud.py`, ~line 788), not the
   actual base-image label — so agents report the *platform* version. Cosmetic/observability;
   caused a red-herring "0.7.0 vs 0.9.0" confusion during investigation.

Verified on `dev` @ `abc8e5cd`, base image rebuilt from `dev`.

webmixgamer:
## Summary

- Fixes an intermittent **race** in the agent base image's GitHub-template clone that produced a silently empty — but "healthy" — agent (~1 in 3 deploys). `startup.sh` emptied `/home/developer` then `git clone`d into it; a concurrent writer (async named-volume-from-image population) repopulated the dir in the window, so the clone saw a non-empty dir and aborted (`exit 128`).
- The clone now stages into a **home-volume temp dir** and **tar-merges** into `/home/developer` (writer-immune by construction), instead of cloning into the live dir.
- A **failed identity clone is now surfaced as UNHEALTHY** via `/health` + monitoring, instead of a running-but-empty agent reporting healthy.

## Changes

- `docker/base-image/startup.sh` — git-sync clone → home-volume temp dir (`.trinity-clone-tmp`) + `tar` merge; drop the racy `rm -rf /home/developer/*`; clean the temp dir on success **and** failure; clear the stale `.git-clone-status` on success/restart/shallow paths; preserve the PAT-in-logs redaction.
- `docker/base-image/agent_server/routers/info.py` — `/health` gains a coarse, server-computed `clone_status` (defensively parsed untrusted input: size-cap + enum-whitelist + absence==ok; **no agent-supplied strings**, since `/health` is unauthenticated).
- `src/backend/services/monitoring_service.py` + `db_models.py` — aggregate a failed identity clone → **UNHEALTHY** with a fixed, injection-safe `"Agent identity clone failed"` issue.
- `src/backend/services/git_service.py` + `docs/TRINITY_COMPATIBLE_AGENT_GUIDE.md` — gitignore the transient clone dir (never commit its PAT-bearing `.git/config`).
- `docs/memory/architecture.md` — document the new `/health` field.
- `tests/unit/test_1439_clone_status_health.py` — health aggregation, defensive parsing (mirrored + drift-guarded vs `info.py`), and `startup.sh` static regression guards.

## Approach note

Chose temp-dir-clone-then-merge over git-init-in-place: the latter creates `.git` **first**, so a partial clone failure leaves a partial `.git` that poisons the `startup.sh` restart-idempotency check (`test -d .git` → "already cloned, skip") — wedging an empty-but-has-`.git` agent. The temp-dir approach writes no `.git` into the home dir until the merge succeeds, so a failed attempt cleanly re-attempts on restart.

## Test Plan

- [x] New unit tests pass (21): aggregation (`failed`→unhealthy, `None`/`ok`→healthy, injection-safe issue string), defensive `.git-clone-status` parsing (absent/malformed/oversized/non-dict → `ok`), and `startup.sh` static guards (temp-dir target, no racy `rm`, tar-merge, stale-marker clears, PAT redaction).
- [x] Existing suites green: monitoring, gitignore doc-sync, #953 startup-no-gitignore-writes.
- [x] **E2E**: 5/5 GitHub deploys land full identity (`.git`/`CLAUDE.md`/skills) where the old code failed ~1/3; an injected failure marker → backend `unhealthy` + `"Agent identity clone failed"`, then clears on recovery; a real onboard→GitHub deploy landed a clean, fully-populated agent.
- [ ] Takes effect on a deployment after a **base-image rebuild + agent recreate** (`scripts/deploy/build-base-image.sh`).

Fixes #1439


vybe (approved):
Validated via /validate-pr — all CI green, named regression tests, security checks clean. Two non-blocking follow-ups noted in validation: unchecked tar-merge exit status (narrow disk-full residual) and agent-monitoring.md feature-flow touch-up.

Issue #506, opened 2026-04-25
feat(capacity): admin-configurable system-wide ceiling + agent-settings UI for max_parallel_tasks


vybe:
## Summary

Add admin-configurable system-wide ceiling for `max_parallel_tasks`, plus the missing UI to let agent owners configure their per-agent value. Backend per-agent storage already exists (CAPACITY-001) but lacks (a) a UI control and (b) any system-level guard against a single agent monopolizing host compute by setting itself to the maximum.

## Context

Today, `max_parallel_tasks` is per-agent (default 3, hardcoded range 1–10) and editable only by API (`PUT /api/agents/{name}/capacity` — [routers/agent_config.py:404](src/backend/routers/agent_config.py#L404)). Two gaps:

1. **No UI**: Owners can't change capacity from agent settings — only via curl.
2. **No system ceiling**: A single owner can set capacity to 10 on every agent they own. With dozens of agents, this saturates the host. The hardcoded 10 is arbitrary and not adjustable per-deployment (small VPS vs. 64-core box).

Need a two-tier model: **admin sets the fleet-wide ceiling, owners pick within it**.

### System-level admin setting
- [ ] New system setting `max_parallel_tasks_ceiling` (int, default 10) stored in existing `system_settings` table, managed via `/api/settings`
- [ ] Admin-only `PUT /api/settings/max-parallel-tasks-ceiling` (1–32 range; reject values outside)
- [ ] `PUT /api/agents/{name}/capacity` validates against the ceiling, not the hardcoded `10` constant
- [ ] If an existing agent's `max_parallel_tasks` exceeds a newly-lowered ceiling, the slot service caps acquisitions at the ceiling at runtime (don't auto-rewrite stored values; just clamp on use)

### Per-agent owner setting (UI)
- [ ] Add capacity control to agent settings panel (alongside timeout/read-only/autonomy)
- [ ] Slider or number input bounded by the current system ceiling (fetched from `/api/settings`)
- [ ] Show current usage (`in_use / max`) inline so owners see real-time impact
- [ ] Save calls `PUT /api/agents/{name}/capacity`; surface 403/400 errors clearly

### Admin settings UI
- [ ] Admin Settings page gets a "Fleet Capacity" section with the ceiling field
- [ ] Tooltip explains: "Maximum parallel tasks any single agent is allowed to consume on this host"

### Tests
- [ ] API test: non-admin cannot set ceiling
- [ ] API test: setting per-agent capacity above ceiling returns 400
[...]

PR #1442, opened 2026-07-04
feat(settings): admin-configurable Brain Orb flags (trinity-enterprise#85)

vybe:
## Summary
- Converts the three Brain Orb platform flags (base / voice / KB-write) from import-time env constants to **runtime resolvers**: `system_settings` override (wins both directions) → `BRAIN_ORB_*` env opt-in → default OFF — one shared `_resolve_bool_flag` helper, fail-open on a settings-read failure (a 500 in `feature-flags` would zero every flag in the frontend store), deliberately uncached (#506 `--workers 2` rationale).
- Adds the admin surface `GET/PUT /api/settings/brain-orb` (admin-only, registered before the `/{key}` catch-all): per-flag `{value, source: override|env|default}` + `gemini_key_configured` (boolean only); PUT takes partial booleans and/or `clear: [flag,…]` reverting a flag to its env/default — without it the env var is silently dead once a DB row exists. Audit-logged with per-flag old→new values.
- Settings → General gets a Brain Orb panel (per-flag source badges, reset-to-env, write-surface warning, forced `feature-flags` refetch after save). Composition fix: the voice-token mint now gates on `base ∧ voice` and `brain_orb_voice_available = base ∧ voice ∧ GEMINI_API_KEY` — a disabled orb can no longer mint Gemini Live tokens via direct API call. `GEMINI_API_KEY` stays env-only (secret).

## Changes
- `src/backend/services/settings_service.py` — `_resolve_bool_flag` + three resolvers + `BRAIN_ORB_FLAGS` registry
- `src/backend/routers/agent_brain_orb.py` — all gate sites converted to resolvers (module-level names, mechanical test seam)
- `src/backend/routers/settings.py` — feature-flags recompose + dedicated GET/PUT with `clear` semantics and old→new audit
- `src/backend/config.py` — dead constants removed (pointer comment); `src/backend/models.py` — `BrainOrbSettingsUpdate`
- `src/frontend/src/views/Settings.vue`, `stores/settings.js`, `components/BrainPanel.vue` — admin panel + store methods + stale env-var copy fix
- `docs/memory/requirements/core-agent.md` (FR-11), `docs/memory/architecture.md`, `docs/memory/feature-flows/brain-orb.md`
- `tests/unit/test_85_brain_orb_settings.py` (23 tests), `tests/unit/test_brain_orb.py` (resolver-seam conversion + base-off mint test), `tests/registry.json`

No schema migration (`system_settings` KV — #506/#1129 precedent). No MCP-server or agent-server change.

## Test Plan
- [x] `pytest tests/unit/test_85_brain_orb_settings.py -v` — 23 passed (resolver order incl. stored-beats-env both directions, junk-value fail-safe, fail-open on DB error; DB flip changes route-gate outcome without restart; admin routes incl. 403, clear revert, set+clear 400, audit old→new, `/{key}` ordering; feature-flags composition + 200-despite-DB-failure)
- [x] `pytest tests/unit/test_brain_orb.py` — 96 passed after seam conversion
- [x] Settings-adjacent suites (`test_506_max_parallel_ceiling`, `test_models_centralized`, settings-router importers) — 85 passed
- [x] Vue SFC compile check on `Settings.vue` / `BrainPanel.vue`; store JS parse check
- [ ] Manual: flip toggles in Settings → General on a running stack; verify Brain tab appears/disappears on reload with no backend restart

Refs abilityai/trinity-enterprise#85 — private-tracker feature; closing keyword does not cross repos, issue is closed manually at release.

### DIFF
--- a/docs/memory/architecture.md
+++ b/docs/memory/architecture.md
@@ -682,12 +682,20 @@ Trinity reads/renders + brokers control. Default OFF — no impact on other agen
   subprocess (timeout-kill, output cap, JSON-parse + non-zero-exit guards); **404 when the
   hook is absent**. Trinity never runs `export_data.py` itself — the agent owns generation +
   scope state (Invariant #8).
-- Platform flags: `brain_orb_available = BRAIN_ORB_ENABLED` (env, default OFF — static
-  render, no Gemini dependency); **`brain_orb_voice_available = BRAIN_ORB_VOICE_ENABLED &&
-  GEMINI_API_KEY`** (default OFF — Phase 3 voice tile, distinct because voice needs a Gemini
-  key). Voice frontend assets are CSP-clean (hand-rolled Gemini client, externalized
-  `voice/voice.js`, same-origin `voice/mic-worklet.js`; `connect-src` already allows `wss:`).
-  No DB change, no migration, no new secret.
+- Platform flags (**runtime-resolved, admin-configurable — trinity-enterprise#85**): the three
+  flags resolve at request time via `settings_service.is_brain_orb_enabled()` /
+  `..._voice_enabled()` / `..._write_enabled()` — `system_settings` row (wins in both
+  directions) → `BRAIN_ORB_*` env var as opt-in fallback → default OFF (one shared
+  `_resolve_bool_flag` helper; fail-open on a settings-read failure; deliberately uncached,
+  #506 `--workers 2` rationale). **Precedence note:** once a stored row exists the env var is
+  ignored until the override is cleared (`PUT /api/settings/brain-orb {clear: [...]}` or
+  generic `DELETE /api/settings/{key}`). Compositions: `brain_orb_available = base`;
+  `brain_orb_voice_available = base && voice && GEMINI_API_KEY` (key stays env-only — secret);
+  `brain_orb_write_available = base && write`; the voice-token mint route gates on
+  `base && voice` too. Admin surface: `GET/PUT /api/settings/brain-orb` (Settings → General
+  panel with per-flag source display). Voice frontend assets are CSP-clean (hand-rolled Gemini
+  client, externalized `voice/voice.js`, same-origin `voice/mic-worklet.js`; `connect-src`
+  already allows `wss:`). No DB change, no migration, no new secret.
 
 ---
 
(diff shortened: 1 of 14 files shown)


dolho:
## Review — LGTM ✅ (minor nits, none blocking)

Read the full diff and verified against the tree. Clean, defensive, well-tested (23 new + 96 reseated), all 19 CI checks green. The standout is the **security composition fix**: `brain_orb_voice_available` and the `/voice-token` mint now gate on `base ∧ voice` (not voice alone), so a disabled orb can no longer mint Gemini Live tokens on the platform key via a direct API call. Good catch.

**Verified:**
- All db accessors exist — `get_setting`/`get_setting_value`/`set_setting`/`delete_setting`. ✅
- Resolver (`self.get_setting`) and the router's `_brain_orb_flag_state` (`db.get_setting_value`) both return the stored value string / None → consistent override detection. ✅
- `_resolve_bool_flag` is fail-open (settings-read failure → env/default), correctly reasoned: these feed `feature-flags`, where a raise would 500 and zero every flag in the frontend store. ✅
- Dedicated `GET/PUT /brain-orb` registered before the `/{key}` catch-all (Invariant #4), covered by an ordering test. ✅
- Frontend `sessionsStore.loadFeatureFlags(true)` exists with the `force` param — the post-save refresh works. ✅
- `--workers 2` no-cache rationale is right; per-gate SQLite read is negligible at orb QPS. ✅

**Nits (follow-up, not blocking):**

1. **`_brain_orb_flag_state()` isn't fail-open** like `_resolve_bool_flag` — a `db.get_setting_value` failure would 500 `GET/PUT /brain-orb`. It's admin-only and off the public feature-flags path, so a 500 there doesn't zero the frontend — acceptable, just inconsistent with the sibling resolver. A `try/except → source:"default"` would match.
2. **Truthy-parse duplicated** — `in ("true","1","yes")` is inlined in both `_resolve_bool_flag` and `_brain_orb_flag_state`. Minor; a shared `_is_truthy` helper would keep them from drifting.

Neither changes behavior. Ready to merge from my side once an approving review lands (currently `BLOCKED` on approval only; the earlier nightly merge-conflict note is stale — PR is `MERGEABLE`).

dolho (approved):

Issue #1447, opened 2026-07-04
test: test_submodule_registers_audit_not_sso is stale since SSO-OIDC landed (#1303) — fails locally whenever the enterprise submodule is mounted


vybe:
## Summary

`tests/unit/test_847_entitlement_seam.py::test_submodule_registers_audit_not_sso` asserts the enterprise submodule does NOT import `.sso` (`assert "from .sso" not in src`) — a static pin from the #910/#941 era when the SSO PoC stub was removed. The real SSO-OIDC implementation has since landed in the submodule (merged via #1303), so the assertion is now stale and the test fails on every clone with the submodule mounted.

CI is unaffected (the test skips when the submodule is absent, which is every public CI job) — this only bites local dev on entitled clones, which is exactly where it trains people to ignore red tests.

## Acceptance Criteria
- [ ] Update the static pin to reflect the intended current contract of `enterprise/backend/__init__.py` (audit registration still asserted; drop or invert the `.sso` absence assertions now that a real implementation exists)
- [ ] Test passes with the submodule mounted AND still skips cleanly without it

Found during /sprint #1443 (pre-existing failure, confirmed on a stashed tree).

PR #1303, opened 2026-06-22
feat(sso): OSS gated surface for enterprise SSO (OIDC) (#32)

dolho:
## What

OSS-side, entitlement-gated surface for **enterprise SSO via OIDC**. All SSO logic lives in the private submodule (companion: trinity-enterprise#36); OSS carries only the gated UI + the submodule bump.

Related to trinity-enterprise#32.

## Changes

- **Login.vue** — "Sign in with `<IdP>`" buttons, shown only when the `sso` feature is entitled and a provider is enabled; handles the OIDC callback fragment (`/login#sso=ok|mfa|error`) and reuses the existing **2FA challenge UI** when the IdP login still needs a local second factor.
- **stores/auth.js** — `completeSsoLogin()` (reuses `_finalizeLogin` / `_setMfaChallenge`) + `fetchSsoProviders()` (empty in OSS-only builds — endpoint 404s).
- **Settings.vue** — admin-gated **SSO** tab → new `SsoPanel.vue` (provider CRUD + connectivity test + policy). Gated by `enterpriseStore.isEntitled('sso')`, same pattern as the 2FA tab.
- **Submodule bump** to the SSO module commit.
- **docs** — architecture enterprise-modules row + requirements §40.

## Why no backend/dep changes

`python-jose` + `httpx` are already in the backend image, and the mint (`create_access_token`), whitelist/`default_role` (#314), and `mfa_gate` seams already exist — so OSS needs **no** Python or Dockerfile change.

## Notes

- **Stacked on `feat/5-2fa-totp`** (reuses the OSS `mfa_gate` + 2FA challenge surface) — base is the 2FA branch so the diff is SSO-only. Retarget to `dev` once 2FA lands; the submodule pointer also settles once both 2FA branches merge.
- Frontend build/e2e validated in CI (`ui` label).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1187, opened 2026-06-12
Codex harness MVP — pluggable agentic execution engine alongside Claude Code


vybe:
## Summary

Implement an MVP of a **Codex CLI harness** so Trinity agents can run on OpenAI Codex as their agentic execution engine alongside Claude Code — introduced through a **pluggable harness abstraction** in the agent runtime, so adding further engines (Gemini CLI, etc.) becomes a repeatable interface-implementation exercise rather than a platform change.

## Context

Trinity's agent runtime is currently hard-wired to `claude --print` as the execution primitive. Spike #854 scoped the feasibility of Codex CLI as an alternative engine; this issue is the implementation MVP.

The design must respect `docs/planning/TARGET_ARCHITECTURE.md`:

- **Harness-agnostic coordination contract** — the platform consumes a typed terminal result (`{status, error_code, cost, tokens, session_id}`, the #945 taxonomy), never engine-specific stdout parsing. The harness boundary is exactly where that contract is produced.
- **Agent-owned out-of-band result record** (§Agent Runtime) — each harness writes its result payload to a durable local record, not a lossy inherited stdout pipe (#548/#333 class).
- **Pull-model compatibility** (Epic #1045 / #1081) — the future agent worker pool must be able to call `harness.execute()` uniformly; nothing engine-specific leaks into the queue/claim/result machinery.
- **Credential hot-reload, not recreate** — engine credentials (e.g. `OPENAI_API_KEY`) ride the existing `.env` injection and `/api/credentials/update` path.

## Acceptance Criteria

- [ ] Harness interface defined in the agent server: execute a turn → typed result (`status`, `error_code`, `cost`, `tokens`, `session_id`); declared capability flags (session resume, MCP support, streaming, cost reporting)
- [ ] Claude Code refactored to be the default implementation of that interface with **zero behavior change** for existing agents
- [ ] Codex harness MVP: runs a chat/task turn end-to-end in an agent container — credentials via existing `.env` injection, MCP config where Codex supports it
- [ ] Harness selectable per agent (`template.yaml` field + agent config), default `claude-code`; selection visible in agent info
- [ ] Backend/coordination layer stays harness-agnostic — `task_execution_service` and result handling consume only the typed contract
- [ ] Capability gating: Claude-only features (Session tab `--resume`, JSONL recovery) degrade gracefully or are hidden for harnesses that lack them
- [ ] Adding a third harness = implement the interface + register it — documented in a short harness authoring guide
[...]

PR #1448, opened 2026-07-04
refactor(open-core): skip private submodules in recursive init, public ENTERPRISE.md, edition in /api/version (#1443)

### LINKED ISSUE #1443: refactor(open-core): make the enterprise/OSS seam clearer — skip private submodule in recursive init, public ENTERPRISE.md, edition in /api/version
## Summary

The enterprise/OSS installation seam (#847) works, but it is only discoverable by reading code comments, and it actively bites OSS users at clone time: both submodules in `.gitmodules` are private, so the documented `git submodule update --init --recursive` fails with auth errors for anyone without org access — even though the enterprise submodule is "optional" (the `ImportError` in `src/backend/main.py` no-ops cleanly at runtime, but the clone step has already errored by then). There is also no cheap way for ops/support to tell which edition an instance is running without grepping boot logs.

## Context

Discussion of how enterprise vs OSS installs are distinguished today: presence of the optional private submodule at `src/backend/enterprise/` decides everything (`register_enterprise(app)` on import success, OSS-only on `ImportError`). Three low-cost DX gaps fell out:

1. **Clone papercut** — `--recursive` init fails for OSS cloners because the enterprise submodule is private. Setting `update = none` on it in `.gitmodules` makes recursive init skip it by default; customers opt in explicitly with `git submodule update --init src/backend/enterprise`.
2. **No public install doc** — the seam mechanics (mount command, HTTPS-PAT override for the SSH URL, rebuild, verification) live only in `main.py` comments.
3. **No edition surface** — `/api/version` doesn't say whether enterprise modules are registered; support has to grep container logs for the boot print.

## Acceptance Criteria

- [ ] `.gitmodules` sets `update = none` for the `src/backend/enterprise` submodule; fresh public clone + `git submodule update --init --recursive` completes without credentials (`.claude` submodule guidance handled/documented separately)
- [ ] Explicit opt-in path (`git submodule update --init src/backend/enterprise`) verified to still mount the submodule for entitled users
- [ ] New public `docs/ENTERPRISE.md` documents the **generic seam only**: what the submodule is, exact mount commands (SSH + HTTPS-PAT URL override), rebuild step, and verification (the `Trinity Enterprise modules registered` boot line and `enterprise_features` in `GET /api/settings/feature-flags`). No module catalog, no paid-feature names, no `enterprise_*` schema — must pass the `enterprise-docs-guard` CI grep
- [ ] `GET /api/version` returns `edition: "oss" | "enterprise"` derived from live registration state (e.g. whether any module registered with the entitlement registry)
- [ ] README / CLAUDE.md submodule setup instructions updated so the default OSS setup path no longer hits an auth failure

## Technical Notes

- Seam wiring: `src/backend/main.py` (enterprise import block, #847), `src/backend/services/entitlement_service.py` (registry; `TRINITY_OSS_ONLY=1` override), `src/backend/routers/settings.py` (`enterprise_features` flag)
- `/api/version` lives in the auth/system router area; edition should reflect runtime registration, not submodule presence on disk (a mounted-but-failed registration is effectively OSS-only)
- Standing rule for the new doc: public docs describe the entitlement/registry mechanism only (see CLAUDE.md → Enterprise Docs Are Private)

vybe:
## Summary
- **Frictionless OSS clone**: both private submodules (`.claude`, `src/backend/enterprise`) now carry `update = none` in `.gitmodules` — a fresh public clone + `git submodule update --init --recursive` (and `git clone --recurse-submodules`) completes **without credentials**; git skips them with exit 0. Verified end-to-end in clean-room clones.
- **Public seam doc**: new `docs/ENTERPRISE.md` documents the *generic* open-core mechanism only (mount, HTTPS-PAT override, rebuild, verify) — passes the `enterprise-docs-guard` grep (run locally).
- **Edition surface**: `GET /api/version` now returns `edition: "oss" | "enterprise"` **plus** `enterprise_features`, both derived from `entitlement_service.list_entitled_features()` — the same source as feature-flags, so the two surfaces can't diverge. Reflects live registration state, not submodule-on-disk. Drive-by: the hardcoded `runtimes` list was stale (missing `codex`, #1187).
- **Deploy pipeline hardened**: under `update = none` a skipped submodule **exits 0**, so `deploy-dev.yml`'s old exit-code success check would have silently degraded the dev deploy to OSS-only with a green log. It now persists the durable `update = checkout` local override before init and judges success by the populated marker file.
- **Contributor docs**: CLAUDE.md `.claude` setup rewritten to the verified config-first flow (a one-shot `--init --checkout` copies `none` into local config, so later plain updates silently skip again — the config line must come first), plus an external-contributor path via the public `abilities` dev-methodology plugin. README gains a one-line pointer.

## ⚠️ For entitled operators (existing clones)
Existing clones have **no** protective local `update` key for the enterprise submodule (unlike `.claude`, whose `update = checkout` was copied at init). After this lands, plain `git submodule update` starts *skipping* it — run once:
```
git config submodule.src/backend/enterprise.update checkout
```
(Documented in `docs/ENTERPRISE.md`; `trinity-ops-public` runbooks should pick this up too.)

## Changes
- `.gitmodules`, `.github/workflows/deploy-dev.yml`, `.github/workflows/build-without-submodule.yml`
- `src/backend/main.py` (`_build_version_payload` + `get_version` — builder stays stdlib-pure for the exec-slice test harness; handler imports `entitlement_service` function-locally so `_set_for_testing` stubs work)
- `docs/ENTERPRISE.md` (new), `README.md`, `CLAUDE.md`
- `docs/memory/requirements/infrastructure.md` (§35.2), `docs/memory/architecture.md`, `docs/memory/learnings.md`
- `tests/unit/test_926_version_endpoint.py` (+3 tests, call sites updated), `tests/registry.json`

## Test Plan
- [x] `pytest tests/unit/test_926_version_endpoint.py` — 8/8 pass (passthrough, runtimes, static function-local-import guard)
- [x] Clean-room clone + new `.gitmodules`: `--init --recursive` skips both submodules, exit 0, no credential prompt (`GIT_TERMINAL_PROMPT=0`, SSH BatchMode)
- [x] Opt-in path: after `git config submodule.src/backend/enterprise.update checkout`, plain `--init` genuinely attempts the clone
- [x] enterprise-docs-guard grep run locally over `docs/` + `CLAUDE.md` — clean
- [x] CI: `build-without-submodule.yml` now also asserts `/api/version` → `edition: "oss"` (end-to-end handler wiring)
- Note: `test_847_entitlement_seam.py::test_submodule_registers_audit_not_sso` fails locally on submodule-mounted clones — **pre-existing** staleness vs the SSO-OIDC merge (#1303), skips in CI; filed #1447.

Fixes #1443


vybe:
**/review (adversarial, second pass) — 2 findings fixed in `55fb35b6`, 3 more resolved as docs/comment corrections; no criticals remain.**

All git mechanics were re-verified empirically by an independent reviewer (10 throwaway-repo experiments — config-key syntax, init-copy semantics, pre-existing-clone asymmetry, `--remote` skip, `clone --recurse-submodules`, recurse×none interaction): every documented command works as written.

Fixed on the branch:
- **F1 (high)**: ENTERPRISE.md step 3 said "rebuild the backend" — factually wrong (the image's COPY allowlist excludes `enterprise/`; the code arrives via bind-mount only). Rewritten with the dev-restart vs prod-overlay split. Following the old text on a prod stack dead-ended at "submodule not present".
- **F2 (medium)**: deploy-dev.yml could print a green `initialized at <sha>` for a *stale* tree (marker file survives a failed update on a previously-mounted VM). Now warns on checked-out-SHA vs recorded-gitlink drift.
- **F3 (medium)**: `docker-compose.prod.enterprise.yml` comments documented update procedures (`--init --recursive`, `--remote`) that silently no-op under `update = none`; now require the one-time local override.
- **F4/F5 (low)**: docs wording corrected (ANY init path copies `none` into local config, not just `--init --checkout`); passthrough test now pins the property with a mismatched edition/features pair.

Follow-ups filed: trinity-enterprise#88 (private ENTERPRISE_LOCAL_DEV.md documents the same now-skipping commands), #1450-class comment-disclosure issue (main.py seam comment; pre-existing, out of scope here), #1447 (pre-existing stale test, unrelated).

dolho:
## Review — LGTM ✅ (minor nits, none blocking)

Read the full diff + verified against the tree. Clean, well-tested, all 18 CI checks green (incl. `backend boots without enterprise submodule` asserting `edition=oss` end-to-end). The three empirically-verified git behaviors in `learnings.md` (skip exits 0, `--init` copies the strategy local, asymmetric protection of existing clones) are exactly the traps a naive rollout hits — good that each is handled in code, not just documented.

**Verified during review:**
- `_build_version_payload` has a single caller (the handler) — both sides updated, no stale callsite. ✅
- `entitlement_service.list_entitled_features()` exists; `edition` derives from its non-emptiness → provably can't diverge from `enterprise_features`. ✅
- Function-local import matches the `routers/settings.py` pattern and is pinned by the new static guard test. ✅
- Runtimes `["claude-code","gemini-cli","codex"]` match `KNOWN_RUNTIMES` in `runtime_adapter.py` exactly. ✅
- No CI workflow depends on `.claude` auto-init, so flipping it to `update = none` is safe. ✅

**Nits (follow-up, not blocking):**

1. **`runtimes` list is still hardcoded** — this PR's drive-by fixes a staleness (`codex` missing) that recurs *by construction*; the next runtime will re-stale it. I get why it isn't derived from `KNOWN_RUNTIMES` (that lives in the agent-server tree and the builder is deliberately stdlib-pure for the exec-slice harness). Suggest a one-line comment next to the literal pointing at `runtime_adapter.KNOWN_RUNTIMES` as the source of truth, so the next person greps their way to updating both.

2. **F2 drift warning is the right instinct** — comparing the recorded gitlink SHA vs the checked-out SHA catches "populated but stale after a failed update." It only `::warning::`s (non-fatal), which is consistent with enterprise-being-optional, but it does mean a stale enterprise tree can still ship under a green deploy. That's an acceptable tradeoff given the policy; just flagging that the warning is the *only* signal — worth making sure someone actually watches deploy-dev warnings.

3. **Existing-operator migration** (the `git config submodule.src/backend/enterprise.update checkout` one-liner) is handled well and documented in three places. Confirm `trinity-ops-public` runbooks get the same note, as the PR body suggests — that's the one surface outside this repo that silently changes behavior.

Nice work on the docs — `ENTERPRISE.md` is thorough, guard-compliant, and the HTTPS-PAT path avoids ever committing a token. Approving from my side once the runtimes comment (nit 1) is added or explicitly waved off.

dolho (approved):

Issue #1360, opened 2026-06-26
docs: feature-flows.md index is 527 lines, over the 400-line guideline — condense


AndriiPasternak31:
## Problem

`docs/memory/feature-flows.md` is the **index** of vertical-slice flow docs and is meant to stay minimal and navigable. The `/sync-feature-flows` skill (Step 6) flags the index when it exceeds **400 lines** and asks for a condensation pass.

As of 2026-06-26 the index is **527 lines** — ~32% over the guideline. This is a long-standing, accumulating condition: each new feature appends a verbose multi-sentence row to the **Recent Updates** table (which never gets pruned), in addition to its category-table row. Individual sync runs correctly keep their own additions to one row, but no pass ever condenses the backlog, so the index grows unbounded.

## Why it matters

- The index is loaded for navigation/orientation; a bloated index defeats its "minimal, navigable" purpose.
- The Recent Updates rows duplicate detail that already lives in the linked flow docs (the index should point, not narrate).

## Proposed fix

1. **Trim the Recent Updates table** — cap it at the most recent N entries (e.g. last ~15–20); older rows are already discoverable via their category-table entry + the flow doc itself.
2. **Shorten Recent Updates descriptions** to one line that points at the flow doc, moving the multi-sentence detail (where not already there) into the doc.
3. Re-check line count is back under 400 after the pass.

## Acceptance criteria

- [ ] `wc -l docs/memory/feature-flows.md` < 400
- [ ] No flow-doc link removed from the category tables (navigability preserved)
- [ ] Recent Updates table retains a bounded, recent window with one-line entries

## Context

Surfaced during the `/sync-feature-flows` run for #1332. Not blocking that work — the #1332 entry itself stayed to a single row. This is the standing index-hygiene follow-up.

PR #1452, opened 2026-07-04
fix(canary): stop B-01 queue-status coherence false-firing under full-suite load (#1446)

### LINKED ISSUE #1446: bug: canary B-01 queue-status coherence violation fires under full-suite load — snapshot race or real projection gap
## Summary

Canary invariant **B-01 (queue-status coherence)** fired in two consecutive full integration-suite runs (2026-07-03 and 2026-07-04): `db.get_queued_count(agent)` returned 1 while the independently-collected queued-execution-id list for the same agent had 0 entries. Either the queued-status projection has a real coherence gap, or the canary's snapshot collection is racy under concurrent load — both are bugs, because a critical-severity invariant that false-fires under load can't be trusted as a regression tripwire.

## Context

- Failing test: `tests/test_canary_invariants.py::TestRunner::test_run_invariants_all` (tests/test_canary_invariants.py:1075), observed state `get_queued_count('a1')=1` vs snapshot count 0
- Fired in both full-suite runs while ~30 test modules were executing concurrently against the live stack; not yet reproduced in isolation
- B-01 exists precisely as a regression guard against a cache layer / status-filter drift on the queued count (see `src/backend/canary/invariants/b01_queue_status_coherence.py`), so a repeatable false-positive erodes the harness's purpose

## Acceptance Criteria

- [ ] Determine which side is wrong: a genuine `schedule_executions.status='queued'` coherence gap, or non-atomic reads inside the canary snapshot (the count query and the id-list query executing at different instants while executions transition)
- [ ] If snapshot race: make the B-01 collection read count + id-list atomically (single query or same transaction/snapshot) so concurrent transitions can't skew the comparison
- [ ] If real gap: fix the projection and add a regression test
- [ ] `test_run_invariants_all` passes reliably inside the full suite

## Technical Notes

- Canary library: `src/backend/canary/` (`snapshot.py` collects state; `invariants/b01_queue_status_coherence.py` compares)
- Related invariant context: status-as-projection (#1082) made `schedule_executions.status` CAS-guarded; B-01 compares the production accessor against an independent collection — the two reads are only meaningful if taken from the same instant
- Severity in the invariant catalog is `critical`, tier B

AndriiPasternak31:
Fixes #1446

## What & why

Canary **B-01 (queue-status coherence, Tier A / critical)** false-fired in two consecutive full integration-suite runs (2026-07-03 and 2026-07-04): `db.get_queued_count('a1')=1` vs a 0-queued snapshot — but never in isolation.

**Root cause: test harness, not production.** Several modules do a module-level, never-restored `sys.modules.setdefault("database", MagicMock())` (`test_watchdog_unit.py:61` and siblings). `setdefault` is first-writer-wins + session-persistent, and `int(MagicMock().get_queued_count("a1")) == 1`, so the canary's `_collect_queued_count_via_service` (`from database import db` resolved at *call* time) read `1` for every agent. B-01 had **no end-to-end unit coverage** — only synthetic `_snap`-built comparator tests — so nothing exercised the real `collect_snapshot()` → accessor path where the leak bites.

On the shipping SQLite backend both B-01 read paths resolve to the same file → **no production coherence bug**. (The Postgres raw-sqlite-readers vs `get_engine()` divergence is a real but deferred gap, tolerable only because the canary is default-OFF — documented in `learnings.md`.)

## The fix (test-only; production canary code untouched)

- `reload_canary` now installs a **controlled** `database` stub whose `db.get_queued_count` reads the SAME temp SQLite the snapshot side reads, via the production `get_engine()` seam. `monkeypatch.setitem` (auto-restored, last-write-wins) **overrides** any leaked foreign stub for the test's duration, and delegates to the temp DB through the real count path — so B-01 is genuinely exercised, not turned into a tautology.
- New `TestInvariantB01EndToEnd` — first e2e coverage driving `collect_snapshot()` against the fixture DB:
  - `test_coherent_with_queued_rows` — N real queued rows → both sides count N → green
  - `test_immune_to_leaked_database_magicmock` — **#1446 regression**, reproduces the exact 07-03/04 flake and proves the controlled stub overrides the leak
  - `test_fires_on_service_count_drift` — real tripwire coverage (accessor undercounts → B-01 fires once with correct observed counts)

## Verification

- Targeted + full `tests/test_canary_invariants.py` → **90 passed**
- Cross-module proof: ran the genuine leak source `test_watchdog_unit.py` in the same session *before* the canary tests → canary tests still pass (immunity holds against the real leak).

## Docs

- `docs/memory/feature-flows.md` — Recent-Updates row (oldest #1082 row trimmed per the #1360 ~20-row cap)
- `docs/memory/learnings.md` — records the `sys.modules` stub-leak-vs-untested-invariant bug class


vybe (approved):
Validated via /validate-pr: conventional commits, Fixes #1446 (P1 type-bug), test-only fix with named regression test (test_immune_to_leaked_database_magicmock) reproducing the exact flake, fixture covers all collect_snapshot() call sites, security sweep clean, no packaging concerns. Dev-merge conflict in learnings.md resolved (kept both entries).

Issue #929, opened 2026-05-24
bug/design: schedule vs agent timeout precedence is silent + SIGKILL error message is ambiguous


vybe:
## Summary

The agent execution timeout (`agent_ownership.execution_timeout_seconds`) and the per-schedule timeout (`agent_schedules.timeout_seconds`) are both 60–7200s configurable, both default to small values, and the effective ceiling is `min(agent, schedule)` enforced silently in `task_execution_service`. There is no surface — in UI, API, or error messages — that tells the operator which cap actually applied. When the smaller cap fires, the agent's Claude Code subprocess gets SIGKILLed and the error message lists three possible causes ("schedule/agent timeout exceeded, OOM kill, or operator cancel") with no indication of which one fired.

This is an ergonomic trap: an operator can configure a schedule with `timeout_seconds=7200` (the max), see "7200" in the UI, and have it silently truncated to e.g. `3600` because the agent's `execution_timeout_seconds` is set lower. Discovery happens only by hitting the wall: a long-running scheduled task gets SIGKILLed at the agent cap with the ambiguous error message, and the operator then has to manually compare the two settings and reason about precedence.

This isn't a single-line bug — asking the team to think through the design and pick an approach. Three sketches in the **Suggested Approaches** section.

## Component

Backend / Scheduling + Task Execution Service (`services/task_execution_service.py`, `db/schedules.py`, `db/agents.py`)

## Priority

P2 — there's an operator workaround (set agent cap ≥ schedule cap, or just always set both to 7200), but the silent-truncation + ambiguous-error combo is a real trap anyone scheduling long-running work will hit. Worth a design pass.

## Symptoms

1. A schedule configured with `timeout_seconds=7200` actually times out at `agent.execution_timeout_seconds` if the latter is lower
2. The UI shows the configured schedule timeout (7200), not the effective timeout
3. The DB row for the failed execution has error message:
   ```
   Execution terminated by SIGKILL after 0 tool calls / 0 turns (exit code -9).
   Likely cause: schedule/agent timeout exceeded, OOM kill, or operator cancel.
   Increase the schedule's timeout_seconds, raise agent memory, or split the
   skill into smaller steps.
   ```
   — but the operator already raised the schedule's `timeout_seconds`, and OOM is not the cause, so the message is actively misleading
4. `cleanup_service` cycles show `auto_terminated: 0` — the kill happens in the task execution path, not the watchdog path — but the operator has no way to know that without reading the source

## Reproduction

[...]

Issue #186, opened 2026-03-26
SEC: User and Agent Enumeration via differential API responses


vybe:
## User Enumeration

**Location**: `POST /api/auth/email/request`

- Unregistered email → `"If your email is registered, you'll receive a code shortly"`
- Registered email → `"Verification code sent to your email"` with `expires_in_seconds`
- Response difference confirms whether email exists in the system

## Agent Enumeration

**Location**: 30+ endpoints including `GET /api/agents/{name}`, `/logs`, `/schedules`, `/playbooks`, `/info`, `/activities`, `/capabilities`, `/executions`, `/git/*`, `/public-links`, `/tags`, `/notifications`, `/monitoring`

- Non-existent agent → `404 "Agent not found"`
- Existing but inaccessible agent → `403 "Access denied to agent"`
- Response difference confirms agent existence

## Remediation

- **Email**: Return the same generic message regardless of email existence (e.g., "Further instructions have been sent to your email address")
- **Agent**: Return `403 "Access denied"` for both non-existent and inaccessible agents (don't distinguish 404 vs 403)

## References

- [OWASP Account Enumeration Testing](https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/03-Identity_Management_Testing/04-Testing_for_Account_Enumeration_and_Guessable_User_Account)

_Source: UnderDefense Web Pentest Report, March 2026_

PR #1453, opened 2026-07-04
fix(webhooks): gate schedule/webhook creation on a live owning agent (#1445)

### LINKED ISSUE #1445: bug: valid webhook token intermittently returns 404 under concurrent load
## Summary

`POST /api/webhooks/{token}` intermittently returns **404 for a freshly-generated, valid webhook token** while the platform is under concurrent load. External integrations calling a valid webhook URL would silently miss triggers.

## Context

During the 2026-07-04 full-suite run, all four `tests/test_webhook_triggers.py::TestWebhookTrigger` tests failed with 404 where 202/413/422 was expected:

- `test_valid_webhook_returns_202` (tests/test_webhook_triggers.py:203)
- `test_webhook_with_context_body` (tests/test_webhook_triggers.py:232)
- `test_webhook_context_too_long_rejected` (tests/test_webhook_triggers.py:254)
- `test_oversized_body_returns_413` (tests/test_webhook_triggers.py:280)

The exact same flow (create schedule → generate webhook token → POST the token) was manually reproduced **immediately after the run and returned a clean 202** — i.e. not reproducible standalone. The 2026-07-03 run did not hit this. So this is a load/timing-dependent failure, not a deterministic regression — but a valid-token 404 is externally visible, so it needs investigation rather than a flaky-test shrug.

## Acceptance Criteria

- [ ] Root cause identified for how a just-generated token can 404 (read-after-write visibility across workers? token-lookup race vs schedule soft-delete filter? routing?)
- [ ] Either a product fix, or a documented proof the 404s came from test-side state (e.g. the module fixture's schedule being deleted by a concurrent test) with the tests hardened accordingly
- [ ] `TestWebhookTrigger` passes reliably inside the full suite, not just in isolation

## Technical Notes

- Webhook lookup path: `src/backend/routers/webhooks.py` → `agent_schedules.webhook_token` (partial unique index); 404 is returned both for unknown token and for a revoked/soft-deleted schedule — distinguishing which branch fired is the first diagnostic step
- Backend runs multiple uvicorn workers over shared SQLite (WAL) — worth checking read-after-write visibility for a token generated on one worker and immediately read on another
- Full pytest logs from the failing run are available if useful for correlating timestamps with backend logs

AndriiPasternak31:
Fixes #1445

## Root cause

`POST /api/webhooks/{token}` intermittently 404'd for a freshly-generated, valid token under concurrent load. The trigger is **orphan schedules**: an admin's `can_user_access_agent` returns `True` *unconditionally with no existence check*, so a schedule (and a real webhook token) could be minted on an agent that has no live `agent_ownership` row. That token then 404s **deterministically** at `get_schedule_by_webhook_token`, which since #1423 INNER-JOINs `agent_ownership` and filters `deleted_at IS NULL`. The full-suite flakiness was the module fixture creating schedules on the name-only `test_agent_name` (no ownership row) — an orphan whose token 404s.

## Fix — no-orphan creation gate

Schedule **and** webhook creation now require a **live owning agent**:

- **Router** (`create_schedule`, `generate_webhook`): access-check **first**, then `404` on `not db.is_agent_live(name)`. Ordering matters — a low-priv caller gets a uniform `403` whether or not the agent exists (no 404-vs-403 enumeration oracle); only an admin/owner ever reaches the 404. Placed before the #929 timeout check so the gate can't be probed via that path.
- **DB chokepoint** (`db/schedules.py:create_schedule` → `None`): re-enforced so the invariant holds for **every** caller (MCP, system-manifest deploy, future/internal), not just the router.
- **New predicate** `db.is_agent_live(name)`: checks `agent_ownership.deleted_at IS NULL` **without joining `users`**, matching the token-lookup predicate *exactly*. `get_agent_owner` INNER-JOINs `users`, so it would false-negative a live agent whose owner-user row is missing (FKs are off platform-wide) — a comment + a dedicated test pin why the fix must not reuse it.
- **Webhook 404 branches** now emit distinct static log lines (`webhook 404: malformed token` / `token lookup miss`) so the next occurrence is grep-attributable in Vector. The raw token is **never** interpolated (log-injection / partial-secret leak on un-validated bytes).

`get_schedule_by_webhook_token` (#1423) is unchanged — the gate keeps it from ever seeing an orphan.

## Tests

- `tests/test_webhook_triggers.py` — hardened off orphan agents: new module-scoped `webhook_agent` fixture (real `POST /api/agents`, live ownership row); new `TestWebhookCreationGate` (missing-agent 404, soft-deleted-agent webhook 404).
- `tests/unit/test_1445_schedule_creation_gate.py` — backend-agnostic (`db_harness`, SQLite + optional Postgres): `is_agent_live` truth table incl. the missing-owner-row invariant, and the `create_schedule` chokepoint (no-orphan for non-router callers + positive control).
- `tests/unit/test_929_timeout_validation.py` — access-first ordering pin (no enumeration oracle) + no-orphan 404; existing timeout tests updated to stub the new gate.

## Docs

`architecture.md` (Creation gate block), `requirements/public-access.md`, `feature-flows/webhook-triggers.md` + index row, `docs/user-docs/api-reference/webhook-triggers.md`.

## Validation

`/validate-pr` clean — no secrets/config/build-packaging gaps; all four mandated doc surfaces updated; named regression tests present. One doc-drift typo caught and fixed in-branch (#1433→#1423 in the flow index). Full integration suite (requires a running stack) to be confirmed on the reviewer's env.


vybe (approved):
Validated via /validate-pr: conventional commits, Fixes #1445 (P1 type-bug, theme-reliability), root-cause verified (orphan fixture, deterministic — not a WAL race), defense-in-depth at router + DB chokepoint, no revert of #1433/#1436, named regression tests. Dev-merge code conflict in generate_webhook resolved: kept #186's uniform-404 AuthorizedAgent gate (inline 403 stays removed) + kept the #1445 is_agent_live check; 48 affected unit tests pass locally on the resolved head.

PR #1454, opened 2026-07-04
fix(chat): fail-loud + owner-gated /task chat-session persistence (#1444)

### LINKED ISSUE #1444: bug: async chat with save_to_session completes but never creates a chat session
## Summary

Async-mode chat (dynamic-thinking / fire-and-forget submission) with `save_to_session=true` completes the execution successfully but never creates a chat session — the conversation is silently lost instead of being persisted.

## Context

Caught by two consecutive full-suite runs (2026-07-03, 2026-07-04):

- `tests/test_dynamic_thinking_status.py::TestAsyncModeSessionPersistence::test_async_mode_with_save_to_session` (tests/test_dynamic_thinking_status.py:455) — 0 sessions found before AND after the async turn
- `::test_async_mode_session_contains_messages` (tests/test_dynamic_thinking_status.py:514) — same

Environment cascade is ruled out: in the 2026-07-04 run the sibling async tests in the same module **passed** (`test_async_mode_returns_accepted`, `test_async_mode_execution_id_is_pollable`, `test_async_mode_completes_eventually`, `test_sse_stream_available_during_execution`) — so the async execution path itself worked while session persistence did not.

## Acceptance Criteria

- [ ] An async-mode chat with `save_to_session=true` creates a `chat_sessions` row and persists both user and assistant messages once the execution completes
- [ ] Both `TestAsyncModeSessionPersistence` tests pass in the integration suite
- [ ] Root cause documented (was persistence skipped on the async completion path, or does session creation race the 202 return?)

## Technical Notes

- `save_to_session` is handled in `src/backend/routers/chat.py` (model in `src/backend/models.py`)
- Suspect the async completion path finalizes the execution without running the session-persistence step the sync path runs — check where the sync path writes `chat_sessions`/`chat_messages` vs what the async terminal applier does

AndriiPasternak31:
Fixes #1444

## Problem
Async `/task` with `save_to_session=true` completes but silently never creates a chat session — the Chat-tab history write was swallowed by a `logger.warning; return None` in `_persist_chat_session`, so a dropped write was invisible. The contract was covered only by slow `requires_agent` integration tests that never gate CI.

## Fix
- **Fail-loud** (`routers/chat.py::_persist_chat_session`, shared by the sync + async branches, SUCCESS-guarded): a persistence error now logs at **ERROR with a stack trace** (message carries agent + execution_id + exc-type only — never user content, **Security F3**) and never re-raises past a completed, billed turn. The sync branch surfaces a `chat_persist_failed` marker on the response.
- **IDOR fix (Security F2)**: a caller-supplied `chat_session_id` is **owner-checked** (`session.user_id == caller`) before appending; on mismatch the write falls through to the caller's own session.
- **Param-leak hardening**: the fail-loud log uses `exc_info=True`; a `DBAPIError` from the `add_chat_message` INSERT would otherwise append a `[parameters: [...]]` tail (the user message / assistant reply) to the traceback. Set `hide_parameters=True` on both engines in `db/engine.py` so bound values stay out of any statement error, engine-wide.

## Triage
Triaged three ways (direct call, wrapper with stubbed `execute_task`, wrapper with the REAL `execute_task` + terminal-CAS row write): on stock config the in-process path is **correct**. The #1083 fire-and-forget callback path (which does not persist) is structurally **unreachable** by a manual `/task` — `ASYNC_DISPATCH_ELIGIBLE_TRIGGERS` = `{schedule, webhook}`. Callback-path persistence is deferred to the pull-mode epic (#1045/#1081).

## Tests
- New fast CI guard `tests/unit/test_1444_chat_session_persistence.py` (**6 tests, all passing**): async-wrapper persists on SUCCESS, no-persist on non-SUCCESS, fail-loud is loud/safe/non-fatal, DBAPIError params not leaked, foreign `chat_session_id` falls through (IDOR).
- The slow `test_dynamic_thinking_status.py::TestAsyncModeSessionPersistence` tests now assert `success` **before** demanding a session, disambiguating an execution failure from a persistence failure.

## Docs
`architecture.md` (chat_sessions block), `feature-flows/authenticated-chat-tab.md` + `parallel-headless-execution.md` (+ index row), `requirements/core-agent.md`.

## Follow-up (out of scope — needs its own issue)
`learnings.md` records an adjacent bug found in review: `_finalize_self_task` (`chat.py:1273`) still uses dict access `session.get("user_id")` on a now-Pydantic `get_chat_session()` result (post-#1093), so self-task `inject_result` into a chat session silently no-ops. Left untouched here to keep scope minimal.


vybe (approved):
Validated via /validate-pr: conventional commits, Fixes #1444 (P1 type-bug, theme-reliability), fix verified at both persistence call-sites via shared _persist_chat_session, security-positive (IDOR fix + hide_parameters redaction), named regression tests (test_1444_chat_session_persistence.py, 6 tests), docs across all tiers. The adjacent _finalize_self_task bug is already tracked as #1457 with PR #1462 open. Dev-merge conflicts (learnings.md, feature-flows.md) resolved keep-both.

PR #1339, opened 2026-06-24
feat(loops): configurable failure policy — fail-fast vs continue-on-error (#1167)

dolho:
## Summary
Sequential agent loops (#740) were **fail-fast**: the first failed iteration aborted the whole loop. This adds a **per-loop failure policy** so a batch-style loop can tolerate a bad iteration and keep going — bounded by a consecutive-failure circuit so a fully-broken agent still terminates. `theme-reliability`.

## Config (default preserves current behavior)
- **`on_failure`**: `abort` (default — fail-fast, backward compatible) or `continue`.
- **`max_consecutive_failures`**: default 3, range 1–100. In continue mode, the loop aborts as `failed` (`stop_reason=max_consecutive_failures`) once that many iterations fail in a row; **a success resets the streak**.

## Behavior
- **Both failure surfaces gated**: a raised exception from `execute_task` AND a non-success `TaskExecutionResult` (TIMEOUT/AGENT_ERROR/CIRCUIT_OPEN/AUTH). Each failed iteration finalizes its `agent_loop_runs` row as `failed`, then continue mode proceeds to the next run.
- **Terminal status**: reaching `max_runs` (or matching the stop-signal) with ≥1 tolerated failure → new **`completed_with_errors`** status; `failed_runs` count surfaced on the loop row + API + UI.
- **`{{previous_response}}`** carries the last *successful* response — a failed iteration never overwrites it (documented semantics).

## Surfaces (Invariant #13)
- **schema/migration**: `agent_loops` gains `on_failure`, `max_consecutive_failures`, `failed_runs` (`schema.py` + `tables.py` Core object + versioned migration).
- **api**: `POST /api/agents/{name}/loops` accepts the two fields; `LoopStatusResponse` surfaces `failed_runs`/`on_failure`/`max_consecutive_failures`.
- **mcp**: `run_agent_loop` gains the two params (`tools/loops.ts` + `client.ts`).
- **ui**: `LoopsPanel.vue` failure-policy controls + `failed_runs`/`completed_with_errors` surfacing.

## Acceptance criteria
- [x] `on_failure: abort|continue`, default `abort` (backward compatible)
- [x] Continue bounded by `max_consecutive_failures` (cutoff aborts as `failed`)
- [x] Failed iteration finalizes its run row as `failed`, then proceeds
- [x] Both failure surfaces (exception + non-success result)
- [x] Partial-success terminal status documented (`completed_with_errors` + `failed_runs`)
[...]

PR #1458, opened 2026-07-04
fix(slack): thread-scope chat session + per-speaker attribution + sender-filtered memory (#903)

### LINKED ISSUE #903: Slack: scope chat session to thread, not channel
## Summary

A Slack chat session in Trinity is currently scoped to `team_id : sender_id : channel_id` (`SlackAdapter.get_session_identifier`, `slack_adapter.py:44`). The thread (`thread_ts`) is **not** part of the key. As a result, every `@mention` and every thread reply from the same user in the same channel share one channel-wide conversation transcript. Two concurrent threads in one channel cross-contaminate each other's context, and a fresh `@mention` does not start a fresh conversation — it inherits the channel's last-10-turn transcript.

This issue scopes the session to the **thread** so that a new top-level `@mention` starts a fresh chat and each thread continues only its own context.

## Context

Surfaced while reviewing how Slack messages reach the agent. The router (`ChannelMessageRouter`) builds the agent prompt via `db.build_public_chat_context(session_id, ...)`, which replays the last 10 turns for `session_id` as a plain-text transcript — there is no persistent Claude Code session, so the channel-scoped key is the *only* thing defining "the conversation."

Concrete failure: a user with two open threads in one channel —

```
Thread A: "summarize the Q3 report"
Thread B: "what's the weather in Tokyo"
Thread A: "translate it to French"   ← prompt contains the Tokyo exchange
```

The agent sees both threads interleaved chronologically, with no thread attribution.

Slack already gives us a stable per-thread id for free: a top-level `@mention` has no `thread_ts`, so `_parse_mention` sets `thread_id = event.ts` (a unique new value); every reply under it carries that same root `thread_ts`. So thread-scoping is mostly a key change.

## Desired behavior

- A new top-level `@mention` → fresh conversation (empty context).
- A reply within a thread (mention or not) → continues that thread's own context only.
- A re-`@mention` *inside* an existing thread → continues that thread (not a new chat).
- DMs are unaffected — they have no threads and stay one continuous per-user conversation.

## Acceptance Criteria

- [ ] `SlackAdapter.get_session_identifier` scopes channel messages to the thread: `team_id : channel_id : thread_id` (drop `sender_id` for channel threads so multi-participant threads share one context).
- [ ] DMs keep the existing `team_id : sender_id : channel_id` key.
- [ ] A new top-level `@mention` resolves to a brand-new `session_id` → `build_context_prompt` returns empty history.
- [ ] Two concurrent threads in the same channel resolve to distinct `session_id`s — no transcript cross-talk.
- [ ] A thread reply and a re-`@mention` inside the same thread resolve to that thread's `session_id`.
- [ ] Unit test covering thread-vs-mention session isolation and DM continuity.
- [ ] Slack memory file / feature-flow note updated to record the session-granularity change.

## Technical Notes

Primary change is contained to `src/backend/adapters/slack_adapter.py`:

```python
def get_session_identifier(self, message: NormalizedMessage) -> str:
    team_id = message.metadata.get("team_id", "unknown")
    is_dm = message.metadata.get("is_dm", False)
    if is_dm:
        return f"{team_id}:{message.sender_id}:{message.channel_id}"
    return f"{team_id}:{message.channel_id}:{message.thread_id}"
```

- No DB schema change — `public_chat_sessions` is keyed by the opaque session identifier already.
- `max_turns=10` in `build_context_prompt` now caps history *within a thread*, which is the intended behavior.
- Out of scope: passive "observe mode" (bot absorbing non-mention thread messages into context without replying). The bot currently responds to every reply in a bot-active thread (`_parse_thread_reply`). Absorbing-without-replying is a separate, larger feature (`[NO_REPLY]` is the existing primitive).
- Group chats (`is_group`) are a separate code path that already uses fresh context — unaffected.

AndriiPasternak31:
Fixes #903

## Problem
`SlackAdapter.get_session_identifier` keyed channel sessions on `team:sender:channel`, so every user in a channel got a *separate* session and two concurrent threads collided into one. The issue's one-line suggestion (drop `sender_id`) fixes conversation **scope** but silently breaks two other concerns the single key had been conflating: speaker **attribution** (multi-participant history would replay as a flat `User:`) and per-user memory **source** (`summarize_user_memory_background` reads the whole session and writes to the *current* user's durable MEM-001 profile — Bob replying in Alice's thread would persist Alice's turns into Bob's memory, a cross-user PII leak).

## Fix
Re-key channel sessions to `team:channel:thread` (DMs keep `team:sender:channel`) and move who-said-what / whose-memory onto each message row via two nullable `public_chat_messages` columns:
- **`sender_label`** — display name; history replays attributed (`Alice:` / `Bob:`), role-label fallback when null (DMs / pre-migration rows).
- **`sender_email`** — verified speaker; the MEM-001 summarizer filters on the current user's own turns so a shared thread never feeds one user's turns into another's durable memory.

Concurrent `get_or_create_session` on a brand-new thread key is race-guarded (savepoint + `IntegrityError` re-SELECT). No migration of pre-existing channel-keyed rows (one-time "forget", expected).

### Follow-up refinements in this branch
- **Assistant-turn memory parity across ALL channel DMs.** A single-participant session must stamp the assistant reply with the recipient's email too, or its replies vanish from that user's filtered memory (a pre-#903 regression). The signal is per-adapter — Slack sets `is_dm`, groups set `is_group`, WhatsApp (DM-only) sets neither — so `_assistant_sender_email()` combines both: every DM stamps, only a Slack channel thread / group chat stays null.
- **Speaker-label sanitization** (`_sanitize_label`) — collapse newlines/control chars before a channel-controlled name lands in the transcript's structural `{speaker}:` position, so a crafted display name can't forge an `Assistant:` line.

## Migration (dual-track, #1183)
- SQLite: `public_chat_messages_sender` in `db/migrations.py`
- PostgreSQL: Alembic `0013_public_chat_messages_sender` (chains off `0012`)
- DDL in `db/schema.py`, MetaData in `db/tables.py`, model in `db_models.py`

## Tests
`tests/unit/test_903_slack_session_thread.py` (session keying, label sanitization, per-channel assistant-stamp matrix) + `tests/unit/test_903_public_chat_sender.py` (sender-filtered DB reads, single-participant assistant-in-feed parity). All 29 green; 193 adjacent router/adapter tests unaffected.

## Docs
`architecture.md`, `requirements/public-access.md` (SLACK-002), `feature-flows/slack-channel-routing.md`, `feature-flows/public-agent-links.md` (+ index row), `learnings.md`.


vybe (approved):
Validated via /validate-pr: conventional commits, Fixes #903 (P2 type-bug, theme-channels), dual-track migration verified (SQLite public_chat_messages_sender + Alembic 0013 chaining off 0012 head, schema.py/tables.py/db_models.py consistent), 29 tests incl. thread-vs-channel scoping, race guard, and sender-filtered memory (closes a cross-user memory-write leak). Dev-merge conflicts (learnings.md, feature-flows.md) resolved keep-both; database.py overlap with #1453 auto-merged, 37 affected unit tests pass locally. Note: open PR #1339 will need its Alembic revision renumbered after this lands.

PR #1462, opened 2026-07-06
fix(chat): self-task inject_result silently no-ops — dict access on Pydantic session (#1457)

### LINKED ISSUE #1451: _finalize_self_task chat-session inject silently no-ops + owner-check never runs (Pydantic .get() on get_chat_session)
## Summary

`_finalize_self_task` (the async `/task` `inject_result` path) checks chat-session ownership with **dict** access on a value that is now a **Pydantic model**, so the check raises and is swallowed — self-task result injection silently never persists, and its ownership gate never actually runs.

```python
# src/backend/routers/chat.py (in _finalize_self_task)
session = db.get_chat_session(request.chat_session_id)
if session and session.get("user_id") == user_id:   # ChatSession has no .get()
    db.add_chat_message(..., source="self_task")
```

`db.get_chat_session()` returns `Optional[ChatSession]` (a `db_models.ChatSession` Pydantic `BaseModel`, via `_row_to_chat_session`). `ChatSession` has no `.get()`, so `session.get("user_id")` raises `AttributeError`, caught by the surrounding `except Exception` (`logger.warning("[Self-Task] Failed to inject result …")`).

Net effect:
1. **Self-task `inject_result` into a chat session silently never works** — every call raises and falls into the `except`; the "result injected" WebSocket field still reports success.
2. **The ownership check never runs** — it can neither admit nor deny, because the write is never reached.

Proven: `ChatSession(...).get("user_id")` → `AttributeError: 'ChatSession' object has no attribute 'get'`.

## Context

Surfaced during the #1444 review (which hardened the *sibling* `save_to_session` persistence path and correctly uses attribute access `session.user_id`). Root cause: the `#300`/`#1093` SQLAlchemy-Core rewrite flipped `get_chat_session`'s return type from a dict/`Row` to a Pydantic model; this call-site was missed (pre-`#1093`, `.get()` was correct). Same class captured in `docs/memory/learnings.md` (model-vs-dict access after the Core rewrite, compounded by a swallowing `except`).

Kept separate from #1444: different feature (self-task `inject_result`, not `save_to_session`), untouched by that diff, and the fix revives a currently-dead behavior that needs its own test coverage.

## Acceptance Criteria

- [ ] `_finalize_self_task` uses attribute access (`session.user_id == user_id`), not `session.get("user_id")`.
- [ ] A self-task with `inject_result` and a chat session owned by the caller persists the result as an assistant message (`source="self_task"`).
- [ ] A self-task referencing a `chat_session_id` owned by a **different** user does **not** write into that session (ownership gate enforced).
- [ ] Regression test added (mirrors `tests/unit/test_1444_chat_session_persistence.py::test_foreign_chat_session_id_falls_through_to_own_session`).
- [ ] Sibling `get_chat_session` call-sites audited for the same dict-vs-model access pattern.

## Technical Notes

Fix:

```python
-    if session and session.get("user_id") == user_id:
+    if session and session.user_id == user_id:
```

Grep the other `get_chat_session` consumers while here — `src/backend/routers/public.py`, `src/backend/routers/voice.py`, and two further uses in `src/backend/routers/chat.py` — and confirm each uses attribute access on the returned `ChatSession` model. Consider a db-accessor unit test asserting the return is the model type so this can't silently regress again.

### LINKED ISSUE #1457: bug: self-task inject_result silently no-ops — dict access on Pydantic get_chat_session() result
## Summary
`_finalize_self_task` (`src/backend/routers/chat.py:1274`) validates chat-session ownership with **dict** access on a value that is now a **Pydantic model**, so the check always raises and self-task `inject_result` **silently never injects**.

## Detail
```python
# routers/chat.py:1273-1274
session = db.get_chat_session(request.chat_session_id)
if session and session.get("user_id") == user_id:   # ChatSession has no .get()
```
After the #1093 SQLAlchemy-Core rewrite, `db.get_chat_session()` returns `Optional[ChatSession]` (a Pydantic `BaseModel` via `_row_to_chat_session`), not a `sqlite3.Row`/dict. `ChatSession` has no `.get()`, so `session.get("user_id")` raises `AttributeError`, which is caught by the surrounding `except Exception` at `chat.py:1292` and logged as a warning.

**Net effect:** when `inject_result=true`, the self-task result is **never** written into the chat session, and the ownership check never actually runs. Pre-#1093 the accessor returned a dict, so `.get()` was correct then; the Core rewrite flipped the return type and this call-site was missed.

The sibling sites already use correct attribute access: `chat.py:1016` (`session.user_id != user_id`, added by #1444), `chat.py:2321`, `chat.py:2352`.

## Fix
Change line 1274 to attribute access:
```python
if session and session.user_id == user_id:
```

## Regression test
Add a fast unit guard (mirror `tests/unit/test_1444_chat_session_persistence.py`): a SUCCESS self-task with `inject_result=true` and an owned `chat_session_id` writes an `assistant` message with `source="self_task"`; a foreign `chat_session_id` does **not** inject. Assert the ownership branch is reached (not swallowed as an `AttributeError`).

## Provenance
Found during #1444 review; recorded in `docs/memory/learnings.md` (2026-07-04). Same class as #1444 — a swallowing `except` turns the regression into a silent no-op, not a crash.

dolho:
## Problem
`_finalize_self_task` (`src/backend/routers/chat.py`) validated chat-session ownership with **dict** access on the value from `db.get_chat_session()`:

```python
if session and session.get("user_id") == user_id:   # ChatSession has no .get()
```

After the #1093 SQLAlchemy-Core rewrite, `db.get_chat_session()` returns `Optional[ChatSession]` (a Pydantic `BaseModel`), not a dict. `session.get(...)` raised `AttributeError`, caught by the surrounding `except Exception` and logged as a warning. **Net effect:** with `inject_result=true`, the self-task result was **never** written into the chat session and the ownership check never ran.

## Fix
One line — attribute access, matching the sibling call-sites (`chat.py:1016`, `:2321`, `:2352`):

```python
if session and session.user_id == user_id:
```

## Regression test
`tests/unit/test_1457_selftask_inject_ownership.py` (pure unit, mirrors `test_1332_*`):
- owned session + SUCCESS + `inject_result=True` → `assistant` message with `source="self_task"`
- foreign-owned session → **no** injection
- guard: `ChatSession` exposes `.user_id` and has no `.get` (old dict access was structurally wrong)

Verified red on the buggy code, green on the fix. `3 passed`.

## Provenance
Found during #1444 review; same class as #1444 — a swallowing `except` turns the regression into a silent no-op.

Fixes #1457

---

## Also fixes #1451 (duplicate of #1457)

#1451 reports the identical defect. Its extra acceptance criterion — audit sibling `get_chat_session` consumers — done, all clean (attribute access, no dict `.get()`):
- `routers/public.py:1110` → `session.agent_name` / `session.user_id` ✅
- `routers/voice.py:54` → `chat_session.id` ✅
- `routers/chat.py:1011` (`if not session`), `:2285`, `:2317` → attribute access ✅
- (`services/agent_client.py` `session.get(...)` is a stats dict, not a `ChatSession` — out of scope)

`_finalize_self_task` was the sole offender.

Related to #1451

### DIFF
--- a/src/backend/routers/chat.py
+++ b/src/backend/routers/chat.py
@@ -1250,7 +1250,7 @@ async def _finalize_self_task(
         try:
             # Validate session exists and belongs to user
             session = db.get_chat_session(request.chat_session_id)
-            if session and session.get("user_id") == user_id:
+            if session and session.user_id == user_id:
                 # Add self-task result as a chat message
                 db.add_chat_message(
                     session_id=request.chat_session_id,
--- a/tests/unit/test_1457_selftask_inject_ownership.py
+++ b/tests/unit/test_1457_selftask_inject_ownership.py
@@ -1,0 +1,131 @@
+"""Unit tests for #1457 — self-task ``inject_result`` must actually inject.
+
+``_finalize_self_task`` (``routers/chat.py``) validated chat-session ownership with
+**dict** access (``session.get("user_id")``) on the value returned by
+``db.get_chat_session()``. After the #1093 SQLAlchemy-Core rewrite that accessor
+returns ``Optional[ChatSession]`` — a Pydantic ``BaseModel`` with no ``.get()`` — so
+the check raised ``AttributeError``, was swallowed by the surrounding
+``except Exception``, and the result was **silently never injected**.
+
+Same class as #1444: a swallowing ``except`` turns the regression into a silent
+no-op, not a crash. Guards:
+  (a) a SUCCESS self-task with ``inject_result=True`` and an **owned**
+      ``chat_session_id`` writes an ``assistant`` message with ``source="self_task"``.
+  (b) a **foreign**-owned ``chat_session_id`` does NOT inject.
+  (c) ``ChatSession`` exposes ``.user_id`` (attribute) and has no ``.get`` — proving
+      the old dict access was structurally wrong, so the ownership branch is really
+      reached, not swallowed.
+
+Pure unit tests — no backend. Mirrors test_1332_cancelled_activity_state.py.
+"""
+from __future__ import annotations
+
+import sys
+from datetime import datetime
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+_BACKEND = Path(__file__).resolve().parents[2] / "src" / "backend"
+if str(_BACKEND) not in sys.path:
+    sys.path.insert(0, str(_BACKEND))
+
+
+def _await(coro):
+    import asyncio
+
+    loop = asyncio.new_event_loop()
+    try:
+        return loop.run_until_complete(coro)
+    finally:
+        loop.close()
+
+
+def _success_result():
+    from models import TaskExecutionStatus
+
+    r = MagicMock()
+    r.status = TaskExecutionStatus.SUCCESS
+    r.response = "task done"
+    r.error = None
+    r.cost = 0.02
+    r.context_used = 120
+    r.context_max = 200000
+    return r
+
+
+def _chat_session(*, owner_id: int):
+    """A real ChatSession Pydantic model — has ``.user_id``, has no ``.get()``."""
+    from db_models import ChatSession
+
+    now = datetime(2026, 7, 6, 0, 0, 0)
+    return ChatSession(
+        id="sess-1457",
+        agent_name="test-agent",
+        user_id=owner_id,
+        user_email="owner@example.com",
+        started_at=now,
+        last_message_at=now,
+    )
+
+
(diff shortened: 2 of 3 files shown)


vybe (approved):
Validated via /validate-pr: fix verified against db layer (get_chat_session returns Pydantic ChatSession; add_chat_message accepts source kwarg on the newly-reachable path), sibling call-site audit confirmed (line 1274 was the sole .get() offender), both touched test files run locally on the PR head (10 passed), and red-verification confirmed the regression test fails on the buggy code. Security greps clean.

Issue #428, opened 2026-04-20
refactor: consolidate ExecutionQueue + SlotService + BacklogService into CapacityManager


vybe:
## Summary

Three parallel queue/capacity abstractions exist in `src/backend/services/`:

- `ExecutionQueue` — Redis LIST, **serial N=1**, used by `/chat`
- `SlotService` — Redis ZSET, **parallel N=max_parallel_tasks**, used by `/task`
- `BacklogService` — SQLite FIFO, **persistent overflow** when slots full

These are one primitive with two knobs (`max_concurrent`, `overflow_policy`) expressed as three classes with overlapping TTL logic, overlapping drain paths, and inconsistent interfaces.

## Scope

Merge into a single `CapacityManager` with:

```python
CapacityManager(agent_name, max_concurrent=N, overflow_policy='reject' | 'queue_in_memory' | 'queue_persistent')
```

- `/chat` becomes `CapacityManager(name, 1, 'queue_in_memory')`
- `/task` becomes `CapacityManager(name, max_parallel, 'queue_persistent')`
- One TTL reasoner, one drain callback, one counter source

## Depends on

- **#306** — push-based completion must land first. Much of the current TTL/drain machinery exists to compensate for blocking HTTP dispatch; the consolidation gets much smaller once #306 moves state transitions to the event consumer.

## Non-goals (keep v1 tight)

- Priority queueing (defer to v2)
- Cross-agent capacity pooling (covered by #24)
- Changing the public `/chat` vs `/task` semantics

## Migration approach (additive-first)

1. Land `CapacityManager` class alongside the existing three classes.
2. Migrate callers one at a time behind a per-agent feature flag.
3. Keep old Queue/Slot/Backlog classes until zero callers remain.
4. Delete in a separate PR after a soak period.

No multi-PR intermediate states where both paths are partially wired.

## Success criteria

- Grep for direct `SlotService` / `ExecutionQueue` / `BacklogService` instantiation returns zero hits outside `CapacityManager` and its tests.
- `/chat` and `/task` semantics unchanged observable-from-outside.
- Single drain path on capacity release.

## Context

See `docs/planning/ORCHESTRATION_RELIABILITY_2026-04.md` — Tier 2.5 Simplification.

PR #1467, opened 2026-07-06
fix(monitoring): circuit breaker opens for busy-but-healthy agents on /health timeout (#1463)

### LINKED ISSUE #1463: bug: circuit breaker opens for busy-but-healthy agents — /health probe timeouts during long CPU-bound executions count as liveness failures
## Summary

The fleet-health monitor's network probe treats any `/health` timeout as a liveness failure and feeds it into the per-agent circuit breaker (#474/#631). But an agent running a long CPU-bound execution starves the agent-server event loop enough that `/health` doesn't answer within the 10s probe timeout — so a **healthy agent doing legitimate work trips its breaker on every long run**, while the in-flight execution itself completes successfully. Every other trigger (other schedules, manual/UI, MCP) then fast-fails with `Agent circuit breaker open — agent is unhealthy` until the dormant cooldown probe recovers it ~1h later. On a production instance this cycles every 2 hours, aligned exactly with a 2-hourly long-running schedule, and floods the operator queue with `circuit_breaker_dormant` alerts (4–7/day).

## Component

Backend / Monitoring Service + Circuit Breaker (agent_client)

## Priority

P1

## Error

Observed sequence on a production instance (sanitized):

```
08:05:00 [TaskExecService] Calling agent [AGENT] /api/task (timeout=7210.0s)   <- long task in flight
08:06:40 fleet_health_check pass complete: agents=8 ... unhealthy=1
08:07:22 WARNING services.agent_client: Circuit OPENED for agent [AGENT] after 3 failures
08:11:32 WARNING services.agent_client: Circuit DORMANT for agent [AGENT] after 10 consecutive open-probe failures — switching to 3600s cooldown probing (#921)
08:20:28 Async task completed for [AGENT]: status=SUCCESS                      <- the "unhealthy" agent's task succeeds
08:27:00 [TaskExecService] CB open, fast-failing execution ... for [AGENT]     <- collateral fast-fail
```

`agent_health_checks` rows during the window: `docker` check healthy (cpu_percent ~102 — pinned on one core, well within the 2-CPU limit), `business` check healthy, `network` check `unhealthy: HTTP timeout`, aggregate `unhealthy: Network unreachable: HTTP timeout`. Once the task finishes, `/health` answers in single-digit ms.

## Location

- **File**: `src/backend/services/monitoring_service.py`
- **Function**: `check_network_health()` — the `except httpx.TimeoutException` handler (~line 331) calls `circuit.record_failure()`
- The docstring codifies the assumption: "/health is supposed to be a small, fast endpoint... A timeout here is a liveness signal — the agent is wedged (event-loop blocked, GIL contention, runaway compute)"

## Root Cause

The #474 liveness heuristic assumes `/health` unresponsiveness == wedged agent. That assumption fails for the common case of an agent-server whose event loop is starved by a *legitimate, still-running execution* (Claude Code subprocess streaming, CPU-bound work). The monitor cannot currently distinguish "wedged" from "busy", so the breaker opens exactly when the agent is doing its heaviest scheduled work — and everything else that targets the agent fast-fails.

Two aggravating factors:

1. **#1121 made this surface suddenly on upgrade**: before #1121 the monitoring loop died on every backend restart and stayed off until manually re-enabled, so many instances effectively had no health probes (and no CB churn). After upgrading past #1121 the loop auto-resumes from persisted config, and the latent interaction appears "out of nowhere" — on the observed instance the dormant-alert rate went from 0/day (monitoring silently off) to 6–7/day immediately after the upgrade.
2. The monitoring loop runs once **per uvicorn worker** (no cross-worker leader lock), so every probe failure is recorded twice — the 3-failure open threshold is effectively reached in ~1.5 probe rounds. Filed separately (see Related).

## Reproduction Steps

1. Enable fleet-health monitoring (`POST /api/monitoring/enable`, or upgrade past #1121 with the persisted setting enabled)
2. Create an agent with a schedule that runs a long CPU-heavy task (~10+ min of sustained work, e.g. large repo indexing/synthesis)
3. While the task runs, watch `agent_health_checks`: network check flips to `unhealthy: HTTP timeout` although the container is healthy and the execution is progressing
4. Within ~2–6 minutes the circuit opens (`Circuit OPENED after 3 failures`), then goes dormant after 10 open-probe failures
5. Any other schedule/manual trigger during the window fast-fails with `Agent circuit breaker open — agent is unhealthy`
6. The original execution completes with status SUCCESS; the breaker stays dormant up to 1h

## Suggested Fix

Make the liveness heuristic aware of in-flight work — the backend already knows it (slot service / active executions):

```python
# in check_network_health(), httpx.TimeoutException handler:
# A /health timeout while the agent has an active execution in flight is
# "busy", not "wedged" — observe reachable=False for the aggregate rollup
# but stay circuit-neutral.
if circuit is not None and not has_active_executions(agent_name):
    circuit.record_failure()
```

Alternative/complementary directions:
- Raise the probe timeout (or require N consecutive timeouts) when active slots > 0
- Harden the agent-server so `/health` answers under task load (serve it from a dedicated thread / prioritized handler) — that would fix the root cause rather than the heuristic

## Environment

- Trinity version: `85a7a85a` (prod compose, 2 uvicorn workers, PostgreSQL backend)
- Agent container: 2-CPU / 4 GiB limits, agent healthy throughout (no OOM, no restarts, RestartCount=0)

## Related

- #474 (health-check failure classification), #631 (health checks feed CB), #921/#924 (dormant state), #1121 (monitoring loop auto-resume — the trigger that surfaced this on upgraded instances)
- Companion bug: monitoring loop duplicated per uvicorn worker (doubles the CB failure feed) — filed separately

dolho:
## Problem (P1)
A `/health` probe timeout during a long CPU-bound execution was counted as a liveness failure and fed into the per-agent circuit breaker. So a **healthy agent doing legitimate scheduled work trips its own breaker on every long run** — the in-flight task completes `SUCCESS`, but every other trigger (other schedules, manual/UI, MCP) fast-fails with `Agent circuit breaker open` until the ~1h dormant cooldown probe recovers it. On the reported instance this cycled every 2h and flooded the operator queue with `circuit_breaker_dormant` alerts (4–7/day). #1121 (monitoring loop auto-resume on restart) surfaced it broadly on upgrade.

## Root cause
`check_network_health()`'s `httpx.TimeoutException` handler assumed `/health` unresponsiveness ⇒ wedged agent. That fails for the common case where a legitimate, still-progressing execution (Claude Code subprocess streaming, CPU-bound synthesis) starves the agent-server event loop — "busy", not "wedged". The monitor couldn't tell them apart.

## Fix
Gate `record_failure()` on in-flight work, which the backend already knows about:

```python
busy = await _agent_has_active_executions(agent_name)   # CapacityManager get_status().is_busy — pure Redis, never touches the agent
if circuit is not None and not busy:
    circuit.record_failure()
```

- Stays **circuit-neutral while busy**, but still reports `reachable=False` (error `"HTTP timeout (agent busy)"`) so the aggregate rollup is honest.
- **Fail-open**: a slot-lookup error treats the agent as idle → preserves the pre-#1463 liveness contract (a genuinely wedged *idle* agent still trips the breaker; the breaker is itself fail-open on Redis-down, so a stray idle verdict can't wrongly trip it).
- Uses the CapacityManager facade (Invariant #428). **Only the timeout leg changes** — ConnectError / ReadError / WriteError / RemoteProtocol classification is untouched.

## Tests
Extended `tests/unit/test_monitoring_health_check_classification.py` (stubs httpx + circuit + capacity_manager, no Docker/Redis):
- timeout **while busy** → circuit-neutral, no `record_failure`
- timeout **while idle** → still `record_failure` (wedge contract preserved)
- slot-lookup **raises** → fails open to `record_failure`

Verified **red** on the pre-fix gate, **green** after. Full run: `10 passed`; adjacent monitoring suites: `47 passed`.

## Scope
Out of scope (filed separately, noted in the issue): the monitoring loop runs once per uvicorn worker with no leader lock, doubling the CB failure feed. This PR fixes the misclassification; the duplication is a separate companion bug.

Docs: `feature-flows/agent-monitoring.md` (classification note + revision history), `tests/registry.json`.

Fixes #1463


vybe (approved):
Validated via /validate-pr. Verified against real code (tests stub the facade, so I checked the seam): `CapacityManager.get_status(agent_name)` exists, is async, returns `QueueStatus.is_busy` (models.py:396) — no name drift; `get_capacity_manager` factory at capacity_manager.py:668; `except CIRCUIT_FAILURE_EXCEPTIONS` (ConnectError/ConnectTimeout) precedes the TimeoutException leg so connect-failures still count; no consumer string-matches on the exact "HTTP timeout" error, so the "(agent busy)" suffix is safe. Named regression tests + feature-flow + registry updated. Noted tradeoff (documented): a genuinely wedged agent holding a slot stays circuit-neutral until slot TTL expiry — acceptable since aggregate health still reports unreachable. Upgraded body ref to `Fixes #1463`.

Issue #631, opened 2026-05-03
bug: circuit breaker flood causes SQLite DB locks and UI unresponsiveness


vybe:
## Summary

When an agent container is up but its HTTP server has crashed (only keepalive running, port 8080 unreachable), the circuit breaker probes it every 30 seconds from both backend uvicorn workers. This creates a self-reinforcing failure loop: probe failures accumulate DB writes, DB writes from two workers contend, SQLite locks cascade to WebSocket auth, and the UI becomes unresponsive for all users.

## Context

Observed in production with `cornelius-internal`. The agent container showed as `Up` but no server was listening on port 8080. Over ~3 hours, this produced:

- **400+ circuit breaker failures** — both uvicorn workers probing independently, logging "Circuit OPENED after N failures" every 30s
- **Intermittent `database is locked` errors** — concurrent DB writes from both workers under probe flood
- **WebSocket connections dropping immediately** — SQLite lock hitting `db/users.py:_get_user_by_field` during WS auth, making the UI unusable for all users

Disabling autonomy on the agent stopped new scheduled executions but did **not** stop the circuit breaker probing — that continued until the agent server was restored.

## Root Causes

1. **No backoff on circuit breaker probes** — once opened, the breaker keeps probing at fixed 30s interval regardless of how long the agent has been down
2. **Both uvicorn workers probe independently** — probe coordination is missing; two writers hitting SQLite simultaneously amplifies contention
3. **No cap on open-circuit probe writes** — each failed probe writes to DB with no ceiling, so a multi-hour outage produces hundreds of writes
4. **Circuit open state doesn't suppress probing** — the breaker should stop active probing after some maximum attempts and wait for an external signal (container restart / health check) instead

## Acceptance Criteria

- [ ] Circuit breaker probes use exponential backoff with a maximum interval (e.g. 30s → 60s → 120s → cap at 5min)
- [ ] After N consecutive open-circuit probes with no recovery, the breaker enters a `dormant` state and stops probing until the agent container is restarted or a manual health-check is triggered
- [ ] Only one worker probes at a time — coordinate via Redis (same pattern as slot service) or suppress duplicate probe writes with a short TTL lock
- [ ] A stuck-open circuit breaker does NOT produce SQLite lock contention that affects unrelated operations (WebSocket auth, UI)
- [ ] Disabling autonomy on an agent also suppresses circuit breaker probing for that agent

## Technical Notes

[...]

PR #1468, opened 2026-07-06
fix(monitoring): fleet-health loop runs once per uvicorn worker — add cross-worker leader lock (#1464)

### LINKED ISSUE #1464: bug: fleet-health monitoring loop runs once per uvicorn worker — duplicate health checks and doubled circuit-breaker failure feed
## Summary

The MonitoringService fleet-health loop is started unconditionally in the FastAPI lifespan of every uvicorn worker, with no cross-worker leader election. The prod compose runs the backend with `--workers 2`, so two independent monitoring loops run concurrently: every health check is performed and stored twice per cycle, and — because `check_network_health()` feeds probe failures into the per-agent circuit breaker (#631) — every probe failure is **counted twice**, effectively halving the breaker's failure threshold and probe budget. This directly amplifies #1463 (breaker opening for busy-but-healthy agents).

## Component

Backend / Monitoring Service (startup wiring in main.py)

## Priority

P2

## Error

Observed on a production instance — every monitoring pass is duplicated, a few ms apart:

```
08:06:40.368 fleet_health_check pass complete: agents=8 (running=8 stopped=0) healthy=7 ... unhealthy=1 duration_ms=32376
08:06:40.380 fleet_health_check pass complete: agents=8 (running=8 stopped=0) healthy=7 ... unhealthy=1 duration_ms=32296
```

And `agent_health_checks` stores two identical rows per layer per cycle (same second, e.g. two `docker` + two `network` + two `business` + two `aggregate` rows at 08:09:47, 08:10:50, 08:11:52, ...).

## Location

- **File**: `src/backend/main.py` (~lines 532–545) — `_start_monitoring_delayed()` / `asyncio.create_task(...)` in the lifespan, runs in every worker process
- **File**: `src/backend/services/monitoring_service.py` — `start_monitoring_service()` / `MonitoringService._run_loop()` have no distributed lock

## Root Cause

Lifespan code executes per worker process. The scheduler was long ago split into its own single-process service (with Redis locking) precisely to avoid multi-worker duplication — the monitoring loop (re-armed at boot by #1121) has no equivalent guard, so with N workers the fleet gets probed N times per interval.

Consequences:
1. Duplicate rows in `agent_health_checks` (storage bloat, misleading dashboards/rollups)
2. Doubled `record_failure()` feed into the circuit breaker: the 3-failure open threshold is reached in ~1.5 real probe rounds and the 10-probe dormant budget in ~5 — see #1463
3. Doubled probe load against every agent-server, worst exactly when an agent is already busy

## Reproduction Steps

1. Run the backend with `--workers 2` (prod compose default) and monitoring enabled
2. `grep "fleet_health_check pass complete" <backend logs>` — two lines per cycle, milliseconds apart
3. Query `agent_health_checks` — two rows per check layer per agent per cycle

## Suggested Fix

Give the loop a cross-process guard, mirroring the scheduler's approach:

```python
# start_monitoring_service(): acquire a Redis lock before starting the loop;
# non-holders skip. Holder refreshes TTL each cycle so a dead worker's lock expires.
lock = redis.lock("monitoring:leader", timeout=interval * 3)
if not lock.acquire(blocking=False):
    logger.info("Monitoring loop not started in this worker (another worker holds the lease)")
    return
```

Alternatively: move the loop into the scheduler service (already single-process and lock-aware), or gate on an env flag set for exactly one worker.

## Environment

- Trinity version: `85a7a85a` (docker-compose.prod.yml, `--workers 2`, PostgreSQL backend)

## Related

- #1463 — circuit breaker opens for busy-but-healthy agents (this bug doubles its failure feed)
- #1121 — monitoring loop auto-resume at boot (made the duplication permanent rather than until-next-restart)
- docker-compose.prod.yml scheduler comments — prior art: duplicate executions with multiple backend workers were fixed by single-process + locking

dolho:
## Problem (P2)
`MonitoringService` is started in the FastAPI lifespan of **every** uvicorn worker with no cross-worker guard. Prod runs `--workers 2`, so two loops probe the whole fleet each interval:
1. Duplicate `agent_health_checks` rows (two identical rows per layer per cycle → storage bloat, misleading rollups).
2. **Doubled `record_failure()` feed** into the per-agent circuit breaker (#631) — the 3-failure open threshold hits in ~1.5 real rounds and the 10-probe dormant budget in ~5. Directly amplifies #1463.
3. Doubled probe load on every agent-server.

The scheduler was long ago split into a single locked process for exactly this reason; the monitoring loop (re-armed at boot by #1121) had no equivalent guard.

## Fix
A cross-worker Redis leader lease — no new service, no schema/compose change:

```python
# each cycle, in _run_loop:
if self._try_acquire_leadership():   # SET monitoring:leader <worker_id> NX EX (3×interval),
    await self._run_check_cycle()    # refresh TTL only when the stored id is ours
```

- **Auto-failover**: every worker still runs the loop and re-checks leadership each cycle, so a dead leader's lease expires (~3 cycles) and a sibling takes over with no restart. (The issue's suggested "acquire once at startup, non-holders return" would have no failover — this re-evaluates per cycle.)
- **Own-lease-only**: refresh/release act only when the stored id equals this worker's — never steals or clobbers another worker's key.
- **Graceful handoff**: `stop()` best-effort deletes its own lease so shutdown hands off immediately instead of waiting out the TTL.
- **Fail-open**: Redis-down → act as leader (single-worker dev keeps probing; in a Redis-down prod the breaker is itself fail-open, so a doubled feed is inert).

## Tests
`tests/unit/test_1464_monitoring_leader_lock.py` (stubs `get_breaker_redis`, no real Redis):
- exactly-one leader across two workers; holder refreshes and keeps leadership
- Redis-down **and** Redis-error fail open to leader
- graceful release hands off immediately; release only deletes the caller's own lease
- TTL-expiry lets a sibling take over
- `_run_loop` runs the check cycle **iff** this worker holds the lease

`9 passed` in-file; `28 passed` across monitoring suites.

**Also verified live against real Redis** (Redis container up): two instances — one wins leadership, own-lease refresh keeps it, a non-holder's release can't steal it, and on the holder's release the sibling takes over immediately.

Fixes #1464 (companion to #1463)


vybe (approved):
Validated via /validate-pr. Verified the seam the tests stub: `get_breaker_redis()` returns a **sync** client with `decode_responses=True` (redis_breaker_util.py:63-68), so the unawaited `r.set/get` calls and the `get() == worker_id` str comparison are correct; lease TTL basis (`docker_check_interval × 3`) matches the loop's actual sleep. Resolved the dev conflict from #1467 (revision-history table — kept both rows); `monitoring_service.py`/`registry.json` auto-merged, both monitoring suites pass on the combined tree (19/19). Upgraded body ref to `Fixes #1464`. Known bounded edge (fine): a probe cycle slower than 3× interval can briefly double-probe until the next lease refresh — self-healing, strictly better than the permanent N× feed.

PR #1469, opened 2026-07-06
feat(webhooks): schedule webhook config UI + optional HMAC signature auth (trinity-enterprise#77)

dolho:
Implements **trinity-enterprise#77** (type-feature; lands in the public repo — WEBHOOK-001 is core OSS, the issue points only at public `src/backend`/`src/frontend`).

## Problem
Public schedule webhooks were backend-only: mint/rotate/revoke via REST only, no UI, and the opaque URL token was the *entire* credential — a leaked URL could trigger the schedule.

## What's added
**Signature auth (the net-new security work):**
- `services/webhook_signature.py` — HMAC-SHA256 over the raw body, constant-time verify against a per-webhook secret in `X-Trinity-Signature: sha256=…`.
- Per-schedule secret: minted `whsec_…`, returned **exactly once**, stored only as an AES-256-GCM envelope (Invariant #12). New `agent_schedules.webhook_secret_encrypted` + `webhook_auth_enabled` — **dual-track migration** (SQLite `agent_schedules_webhook_auth` + Alembic `0014`; `schema.py` + `tables.py`).
- `POST/DELETE /api/agents/{name}/schedules/{id}/webhook/secret` (enable-rotate / disable). GET reports `auth_enabled`/`has_secret`, never the secret. Rotating the URL or revoking clears the secret.
- Public trigger verifies the signature after body read+cap — **fail-closed** (401 missing/invalid; 500 on an unreadable stored secret; never a silent bypass). Off by default → existing token-in-URL webhooks unchanged.

**UI** (`SchedulesPanel.vue`): per-schedule **Webhook** panel — enable, masked/revealable URL, one-click **Copy URL**, ready-to-paste **example curl**, **Rotate URL**/**Revoke** (with warnings), **enable/rotate/disable signature**, and the signing secret shown **exactly once**. Dark-mode aware.

## Acceptance criteria
- [x] Per-schedule Webhook control, **disabled by default** (no token until enabled)
- [x] Masked/revealable token, Copy URL, copyable example `curl` (with optional context body)
- [x] Rotate + Revoke with immediate-invalidation warnings
- [x] **Signature auth** — HMAC secret scheme, off by default, 401 on unauthenticated/invalid when on; secret stored AES-256-GCM, surfaced once
- [x] **Access control decided + documented**: mint/rotate/revoke = `AuthorizedAgent`, aligning with schedule management (per the AC)
- [x] Audit-logged (unchanged); dark-mode + accessible
- [x] User-doc snippet (configure + secure a webhook, openssl signing example)
- [ ] MCP tool (explicit **stretch**) — deferred; noted for a follow-up

## Verification
- Unit: `test_77_webhook_signature.py` (12) + `test_77_webhook_secret_db.py` (8, on `db_harness` → real schema + AES-256-GCM). Updated the #1424 hardening test double for the new model fields. **24 passed.**
- SFC compiles (`@vue/compiler-sfc`).
- **Live** against real Redis + SQLite + AES-256-GCM: migration ran, `mint → encrypt → store → trigger-path decrypt → signature verify` round-trips; valid accepted, tampered/wrong-secret rejected; URL-rotate + revoke clear the secret. PASS.

## Notes
- Dual-track migration guarded by the `schema-parity` CI job; both `migrations.py` (SQLite) + Alembic `0014` land together.
- The three companion `type-bug` hardening issues referenced in the issue (dedup, soft-delete guard, pre-auth rate limit) are **already merged** on `dev` (`webhooks.py`); this builds on them.

Related to trinity-enterprise#77


vybe (approved):
Validated via /validate-pr: dual-track schema migration (SQLite + Alembic 0014 + schema.py + tables.py), AES-256-GCM secret storage (Invariant #12), fail-closed HMAC verify with constant-time compare, secret shown exactly once, architecture + user docs updated, unit + DB lifecycle tests registered. Reconciled two semantic conflicts with the just-merged #1437 (fake-schedule attrs + keyed idempotency probe); all 4 webhook unit files pass (28/28). Approving.

PR #1473, opened 2026-07-06
docs: enterprise-tracker features are entitlement-gated by default

dolho:
Clarifies the open-core routing rule in `CLAUDE.md` (SDLC → Two trackers): a feature filed in the **enterprise tracker** (`type-feature`) is a **gated enterprise module by default**, not OSS-core. The core-code-lands-public path is the explicit exception (the user's monetization call), not the default assumption — reusing OSS tables/endpoints (e.g. `agent_permissions`) does not make a feature OSS.

Prompted by a real slip (ent#84 built as a public Settings tab when it should default to gated). A matching clarification lands in the private `.claude` methodology doc (`DEVELOPMENT_WORKFLOW.md`) via a separate trinity-dev PR.

Docs-only.


vybe (approved):
Validated via /validate-pr: docs-only CLAUDE.md clarification of the open-core routing default; no security concerns. Approving.

Issue #454, opened 2026-04-22
feat(scheduler): agent-owned pre-check hook — let agents veto scheduled invocations deterministically


dolho:
## Problem

Trinity's scheduler (`agent_schedules` + `scheduler_service.py`) unconditionally fires a chat message on every cron tick. Poll-driven agents (PR reviewers, inbox triage, RSS monitors, alert routers, log scanners) burn tokens on empty wakes — a 15-minute cron on a quiet source costs ~96 fires/day × ~\$0.02–0.05 per tiny chat turn, all to have the agent respond "nothing to do."

Trinity already has the right cost-observation infra (SUB-004 token/cost tracking, cost alerts) and the right philosophical stance (GUARD-001 guardrails: *"deterministic… not relying on model compliance alone"*). What's missing is a **prevention** primitive at the scheduler layer.

## What this is NOT

- **Not** a replacement for cron schedules — cron stays as-is.
- **Not** an operator-facing config field. The operator creates a normal cron schedule like today.
- **Not** a new trigger type or schema change.

## Proposal — agent-owned pre-check endpoint

The **agent** decides (via code shipped in its template) whether a scheduled fire should actually propagate. The operator and the scheduler are unchanged; the agent gains an optional hook.

### Contract

Trinity's base-image `agent-server.py` already runs on port 8000 inside every agent container. Add one optional endpoint:

```
POST  http://agent-<name>:8000/api/pre-check
```

Agent templates implement it if they want to gate scheduled invocations. If they don't implement it, scheduler behavior is identical to today (backward compatible).

**Response shapes**:

```json
{ "fire": false, "reason": "no new PRs" }
```

```json
{ "fire": true, "message": "Review abilityai/trinity#371, abilityai/abilities#42" }
```

The optional `message` field lets the agent rewrite the chat payload with real work (e.g. the list of PRs detected by the pre-check). If absent, the scheduler uses `schedule.message` as today.

### Scheduler flow (in `services/scheduler_service.py`)

```python
def fire_schedule(schedule):
    decision = _run_pre_check(schedule.agent_name)   # new helper
    if decision.skip:
        record_execution(status='skipped', reason=decision.reason)
        return
    message = decision.message or schedule.message
    post_chat(schedule.agent_name, message)
```

`_run_pre_check` semantics:

| Response | Action |
|---|---|
| `200 {fire: true, …}` | Fire chat (use `message` if provided, else `schedule.message`) |
[...]

Issue #76, opened 2026-03-06
Configurable MCP Server URL in Admin Settings


vybe:
## Problem

The MCP server URL shown to users on the API Keys page is hardcoded in the frontend (`src/frontend/src/views/ApiKeys.vue:344-351`):

```javascript
const mcpServerUrl = computed(() => {
  const host = window.location.hostname
  if (host === 'localhost' || host === '127.0.0.1') {
    return 'http://localhost:8080/mcp'
  }
  return `http://${host}:8080/mcp`
})
```

This assumes MCP always runs on port 8080 directly. In production deployments where MCP is proxied through nginx (e.g., `https://us14.abilityai.dev/mcp`), the generated URL is wrong — users get `http://us14.abilityai.dev:8080/mcp` instead of the proxied HTTPS URL.

There is no way to override this without editing source code.

## Proposed Solution

Make the MCP server URL configurable via Admin Settings:

1. **Backend**: Add `mcp_external_url` key to `system_settings` table
   - `GET /api/settings/mcp-url` — returns configured URL (or null for default behavior)
   - `PUT /api/settings/mcp-url` — admin-only, sets the external MCP URL
   - `DELETE /api/settings/mcp-url` — resets to default (auto-detect from hostname)

2. **Frontend Settings page** (`Settings.vue`): Add "MCP Server URL" field in a new section
   - Text input with placeholder showing the auto-detected default
   - Save/Reset buttons
   - Validation: must be a valid URL ending in `/mcp`

3. **Frontend API Keys page** (`ApiKeys.vue`): Fetch configured URL on mount
   - If `mcp_external_url` is set in settings, use it
   - Otherwise, fall back to current `window.location.hostname` logic
   - The `getMcpConfig()` function and the connection info snippet both use this URL

## Files to Change

- `src/backend/routers/settings.py` — Add MCP URL endpoints
- `src/frontend/src/views/Settings.vue` — Add MCP URL config section
- `src/frontend/src/views/ApiKeys.vue` — Fetch and use configured URL

## Acceptance Criteria

- [ ] Admin can set a custom MCP server URL in Settings
- [ ] API Keys page shows the configured URL in connection snippets
- [ ] Resetting returns to auto-detect behavior
- [ ] Non-admin users cannot change the setting
- [ ] Works for both HTTP and HTTPS URLs

PR #1479, opened 2026-07-06
fix(git-sync): exempt template-committed .trinity convention files from the gitignore untrack sweep (trinity-enterprise#76)

vybe:
## Summary
- Brain-Orb-capable templates (e.g. `github:Abilityai/cornelius`) COMMIT their convention files — the `.trinity/brain-orb/{scopes,scope,search,action}` hooks and `.trinity/setup.sh`. The fleet gitignore migration (`_migrate_workspace_gitignore`, run on every operator Push) appends the fleet-wide `.trinity/` ignore and then `git rm --cached`s ignored-but-tracked files — untracking those files and pushing their deletion, so a later re-clone/recreate boots hookless and never runs startup setup.
- `_build_rm_cached_ignored_command` now pathspec-exempts `.trinity/brain-orb` and `.trinity/setup.sh` from both the emptiness probe and the removal pipe; sibling `.trinity/` runtime files are still swept.
- Compatibility check S-005 additionally accepts the `.trinity/*` star form — the only shape a template can commit re-included hooks under (git cannot re-include beneath a dir-form exclusion) — so the compat panel doesn't flag Brain-Orb templates and its auto-fix doesn't churn.

## Changes
- `src/backend/services/git_service.py` — sweep exemptions (probe + removal pipe)
- `src/backend/services/compatibility/static_checks.py` — S-005 accepts `.trinity/*`
- `tests/unit/test_github_init_gitignore.py` — real-git regression test (hooks + setup.sh survive the merge+sweep; sibling state still swept; working tree untouched)
- `tests/unit/test_compatibility_checks.py` — S-005 star-form snapshot test

## Test Plan
- [x] `pytest tests/unit/test_github_init_gitignore.py tests/unit/test_compatibility_checks.py` — 43 passed
- [x] Live e2e: created `cornelius-e2e` from `github:Abilityai/cornelius`, ran an operator Push through this backend — all 6 hook files + `setup.sh` stayed tracked, the vault capture committed, non-exempt `.trinity/` state was swept. (The pre-fix backend reproduced the bug: the first Push deleted `setup.sh` — caught live, second exemption added.)

Part of abilityai/trinity-enterprise#76 (Brain Orb full cycle from the public cornelius template — agent-side kit shipped in `Abilityai/cornelius@1ce85bd`).

### DIFF
--- a/src/backend/services/git_service.py
+++ b/src/backend/services/git_service.py
@@ -946,12 +946,19 @@ def _build_rm_cached_ignored_command(git_dir: str) -> str:
     (bash can't hold NUL bytes), then a NUL-delimited pipe to xargs so paths
     with spaces or unicode survive the round-trip. Working-tree files are
     left alone; only the index is touched.
+
+    ``.trinity/brain-orb/`` is exempt: Brain-Orb-capable templates COMMIT their
+    convention hooks there (trinity-enterprise#76), while the fleet-wide
+    ``.trinity/`` ignore appended above makes them look ignored-but-tracked —
+    without the pathspec exclusion the first push would untrack the hooks and
+    push their deletion, so a later re-clone/recreate boots hookless.
     """
+    hook_exempt = '":!.trinity/brain-orb"'
     script = (
         f"cd {shlex.quote(git_dir)} && "
-        "ignored=$(git ls-files -ci --exclude-standard) && "
+        f"ignored=$(git ls-files -ci --exclude-standard -- . {hook_exempt}) && "
         'if [ -n "$ignored" ]; then '
-        "git ls-files -ci -z --exclude-standard | "
+        f"git ls-files -ci -z --exclude-standard -- . {hook_exempt} | "
         "xargs -0 git rm --cached --quiet -r --; "
         "fi"
     )
--- a/src/backend/services/git_service.py
+++ b/src/backend/services/git_service.py
@@ -947,18 +947,20 @@ def _build_rm_cached_ignored_command(git_dir: str) -> str:
     with spaces or unicode survive the round-trip. Working-tree files are
     left alone; only the index is touched.
 
-    ``.trinity/brain-orb/`` is exempt: Brain-Orb-capable templates COMMIT their
-    convention hooks there (trinity-enterprise#76), while the fleet-wide
-    ``.trinity/`` ignore appended above makes them look ignored-but-tracked —
-    without the pathspec exclusion the first push would untrack the hooks and
-    push their deletion, so a later re-clone/recreate boots hookless.
+    ``.trinity/brain-orb/`` and ``.trinity/setup.sh`` are exempt: Brain-Orb /
+    setup-convention templates COMMIT those files (trinity-enterprise#76), while
+    the fleet-wide ``.trinity/`` ignore appended above makes them look
+    ignored-but-tracked — without the pathspec exclusions the first push would
+    untrack them and push the deletion, so a later re-clone/recreate boots
+    hookless and never runs its startup setup (verified live: the e2e Push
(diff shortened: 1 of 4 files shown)


src/backend/services/git_service.py
```diff
@@ -946,12 +946,21 @@ def _build_rm_cached_ignored_command(git_dir: str) -> str:
...
+
+    ``.trinity/brain-orb/`` and ``.trinity/setup.sh`` are exempt: Brain-Orb /
+    setup-convention templates COMMIT those files (trinity-enterprise#76), while
+    the fleet-wide ``.trinity/`` ignore appended above makes them look
+    ignored-but-tracked — without the pathspec exclusions the first push would
+    untrack them and push the deletion, so a later re-clone/recreate boots
+    hookless and never runs its startup setup (verified live: the e2e Push
+    swept setup.sh before the second exemption was added).
     """
+    exempt = '":!.trinity/brain-orb" ":!.trinity/setup.sh"'
```
dolho, line 958: Verified (real git): `.trinity/pre-check` — a documented template-committed convention hook (SCHED-COND-001, #454) — is swept by this same pipe because it isn't in the exempt set, so a template shipping it re-hits exactly the bug this PR fixes (untracked on first Push → re-clone boots without it).

Two options: add `.trinity/pre-check` here now, or scope this PR to #76 and file a follow-up for the general class.

Nit while here: this exempt list will grow one path at a time. Promoting it to a named module constant (e.g. `_TEMPLATE_COMMITTED_TRINITY_PATHS`) makes adding `pre-check` a one-liner and self-documents the growing allowlist.

dolho (commented):
**Review — approve-with-a-question. Solid, surgical fix with an excellent real-git regression test; I reproduced one sibling of the same bug that isn't covered.**

I ran the branch's actual `_build_gitignore_merge_command` + `_build_rm_cached_ignored_command` output against a real git repo mirroring the scenario (plus a `.trinity/pre-check` file):

```
BEFORE:  brain-orb/action, brain-orb/scopes, pre-check, runtime-state.yaml, setup.sh   (tracked)
AFTER merge+sweep:  brain-orb/action, brain-orb/scopes, setup.sh   ✅ survive
                    runtime-state.yaml   ✅ swept (correct)
                    .trinity/pre-check   ❌ SWEPT
```

**What's right**
- Root cause is correct: the fleet merge appends dir-form `.trinity/`, which defeats the template's `!.trinity/brain-orb/` re-include → hooks become ignored-but-tracked → the `git rm --cached` sweep untracks them. The pathspec exemption is the correct, surgical compensation, applied to **both** the probe and the removal pipe (consistent), with a positive `-- .` pathspec so the `:!` excludes have something to subtract. No injection risk (hardcoded constants, `git_dir` is `shlex.quote`d).
- S-005 change is correct and complete for the shapes in play.
- Tests are excellent — real git, asserts survival AND that siblings are still swept AND the working tree is untouched; the live e2e actually caught the `setup.sh` case.

**The one substantive point (see inline):** `.trinity/pre-check` (SCHED-COND-001 / #454 — a documented template-committed convention hook) hits the identical bug and isn't exempted. Fine to scope this PR to #76 and file a follow-up, but it shouldn't be left silent since a template that ships `pre-check` will have it untracked on the first Push.

**Nit:** `commit.gpgsign false` in the test is fine (throwaway repo), just noting it's local-repo config.

vybe:
Superseded by #1490 — the fork-to-own branch was stacked on this one, so #1490 now carries both tasks (trinity-enterprise#76 git-sync exemptions + trinity-enterprise#93 fork-to-own) as one PR against dev. Closing unmerged; commits eed834ee + 91170245 live on in feature/93-fork-to-own.

Issue #1485, opened 2026-07-06
Config Validation: 4 critical config issues found (2026-07-06)


vybe:
## Automated Config Validation Report

**Date**: 2026-07-06
**Result**: 4 critical config issues require attention — all are the known packaging-gap class (#1039/#1056): a documented/advertised config lever that never reaches the container, so setting it in `.env` silently does nothing.

### Critical Findings (P0-P1)

1. **`OPERATOR_INTAKE_ENABLED` / `DO_NOT_TRACK` privacy kill switch is inert** — `.env.example` (lines 60–70) promises *"Set to false (or DO_NOT_TRACK=1) to fully disable the outbound submission for air-gapped / privacy-strict installs"*. Both vars are read in `src/backend/config.py:33-35`, but **neither docker-compose.yml nor docker-compose.prod.yml forwards them** to the backend container — the documented privacy opt-out does not work. (The setup-consent checkbox remains the primary gate, but the env-level kill switch is advertised and broken.)

2. **`DISPATCH_BREAKER_ENABLED` (global dispatch-breaker gate, #526) is unwired and undocumented** — read in `src/backend/config.py:143`, default false. It appears in **neither compose file nor `.env.example`**. The per-agent UI toggle (`PUT /api/agents/{name}/circuit-breaker`) engages only when this global flag is also on, so the dispatch circuit breaker is un-enableable through supported configuration and the UI toggle silently does nothing.

3. **`PUBLIC_ACCESS_REQUESTS_ENABLED` env lever is inert** — documented in `.env.example` (line 58), read in `src/backend/config.py:19`, forwarded by **neither compose file**. Fails safe (default false both sides) and the `public_access_requests_enabled` system setting still works, but an operator following `.env.example` to enable CLI onboarding gets a silent no-op.

4. **`VITE_BUG_REPORTING_ENABLED` / `VITE_BUG_INTAKE_URL` cannot be set in a standard build** — `src/frontend/src/components/HelpChatWidget.vue:323-326` says operators can repoint or disable in-app bug reporting at build time, but `docker/frontend/Dockerfile.prod` declares only `ARG VITE_API_URL` and docker-compose.prod.yml passes only that build arg. The disable/repoint lever for reports sent to `intake.abilityai.dev` does not exist in the shipped build path.

### Recommended Actions

1. Forward `OPERATOR_INTAKE_ENABLED` + `DO_NOT_TRACK` in the backend `environment:` block of **both** compose files (defaults `true` / `0`).
[...]

Issue #1116, opened 2026-06-09
feat(ui): In-app bug reporting from the floating Help widget (hosted intake → GitHub issues)


vybe:
## Summary

Let users report bugs directly from within Trinity, from the **floating Help widget** (`HelpChatWidget`) that already appears on every authenticated page. Today that widget is Q&A-only (it queries the hosted `ask-trinity` endpoint); it has no way to capture a bug report. Any OSS instance — deployed anywhere, by anyone, with no GitHub credentials of its own — should be able to submit a structured, diagnostic-rich bug report that lands as a GitHub issue in `abilityai/trinity`.

## Context

Trinity is open source and self-hosted on the operator's own hardware, so a deployed instance has no credentials to file issues against `abilityai/trinity` directly. Reports are routed through a **hosted intake endpoint** (an Ability.ai-operated service that holds the GitHub token server-side and files the issue centrally) — the same hosting pattern already used by the `ask-trinity` Q&A Cloud Function the Help widget calls today. This keeps the client credential-free and works on every deployment, at the cost of running and protecting a shared intake service.

The feature lives inside the existing widget (no new floating button) so users have a single, familiar place for "I need help / something's broken."

## Acceptance Criteria

**Widget UX**
- [ ] The `HelpChatWidget` (`src/frontend/src/components/HelpChatWidget.vue`, mounted in `src/frontend/src/App.vue`) gains a "Report a bug" mode/tab alongside the existing Help Q&A — no second floating button.
- [ ] Bug report form: short title + description, both required; clear character limits; inline validation.
- [ ] Before submission, the user sees exactly what will be sent (collapsible diagnostics + screenshot preview) and must explicitly confirm — submission is opt-in and transparent.
- [ ] A clear notice that the report becomes a **public GitHub issue** in `abilityai/trinity`.
- [ ] Success state returns the created issue URL (or a tracking reference); error state is retryable.

**Diagnostics auto-capture** (shown to the user before sending, scrubbed)
- [ ] App version + git commit (from `GET /api/version`) and current route/URL.
- [ ] Browser environment: user agent, viewport, OS.
- [ ] Recent captured JS console errors/warnings (last N), **scrubbed** for tokens/PII before leaving the browser.
- [ ] Optional user-initiated screenshot of the current view, attached to the report (user can review/cancel it first).

**Hosted intake service**
[...]

Issue #187, opened 2026-03-26
SEC: JWT Tokens Not Invalidated on Logout — no revocation mechanism


vybe:
## Pentest Finding 3.3.4 — Low (CVSS 2.0)

**Location**: `config.py`, `routers/auth.py`

## Impact

JWT tokens are valid for 7 days with no server-side revocation mechanism. If a session token is exfiltrated, it remains usable for the full 7-day period regardless of password changes or logout attempts. There is no logout endpoint in the API.

## Root Cause

```python
# config.py
ACCESS_TOKEN_EXPIRE_MINUTES = 10080  # 7 days (was 30 minutes)
```

No token blacklist, `jti` tracking, or refresh token pattern exists. `grep` for `logout|blacklist|revoke|invalidate` in `routers/auth.py` returns no results.

## Remediation

- Store token `jti` claims in Redis upon logout; verify against the blacklist on each request
- Reduce `ACCESS_TOKEN_EXPIRE_MINUTES` to 60–480 minutes and implement refresh tokens
- Add `POST /api/auth/logout` endpoint

## References

- [OWASP Identification and Authentication Failures](https://owasp.org/Top10/2021/A07_2021-Identification_and_Authentication_Failures/index.html)
- [JWT for Java Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.html)

_Source: UnderDefense Web Pentest Report, March 2026_

PR #1492, opened 2026-07-06
fix(frontend): plumb VITE_BUG_REPORTING_ENABLED / VITE_BUG_INTAKE_URL as prod build args (#1489)

### LINKED ISSUE #1489: VITE_BUG_REPORTING_ENABLED / VITE_BUG_INTAKE_URL not plumbed as prod build args — in-app bug reporting can't be disabled
## Summary

In-app bug reporting (HelpChatWidget) cannot be disabled or repointed in a standard production build. `src/frontend/src/components/HelpChatWidget.vue:323-326` says operators can repoint (`VITE_BUG_INTAKE_URL`) or disable (`VITE_BUG_REPORTING_ENABLED=false`) at build time, but `docker/frontend/Dockerfile.prod` declares only `ARG VITE_API_URL`, and `docker-compose.prod.yml` passes only that build arg. The advertised levers do not exist in the shipped build path, so every standard build sends bug reports to the default hosted intake endpoint with no supported off switch.

## Context

Found by the 2026-07-06 automated `/validate-config` run (umbrella: #1485). Same packaging-gap class as #1039/#1056, on the frontend build-arg surface instead of the compose `environment:` surface (Vite vars are baked at build time, so the plumbing is Dockerfile `ARG`/`ENV` + compose `build.args`).

Reports are user-initiated (an operator clicking "report a bug"), so this is less severe than automatic telemetry — but privacy-strict / air-gapped installs are promised a disable lever that isn't wired.

## Acceptance Criteria

- [ ] `docker/frontend/Dockerfile.prod` declares `ARG VITE_BUG_REPORTING_ENABLED` and `ARG VITE_BUG_INTAKE_URL` (with the current code defaults) and exports them as `ENV` before the build step
- [ ] `docker-compose.prod.yml` frontend `build.args` passes both through from `.env` (`${VITE_BUG_REPORTING_ENABLED:-true}`, `${VITE_BUG_INTAKE_URL:-}` or the code default)
- [ ] Both vars are documented in `.env.example`
- [ ] A build with `VITE_BUG_REPORTING_ENABLED=false` hides/disables the in-app bug report action

## Technical Notes

- Consumer: `src/frontend/src/components/HelpChatWidget.vue:325-326`
- Follow the existing `VITE_API_URL` pattern in `docker/frontend/Dockerfile.prod:19-22` and `docker-compose.prod.yml` frontend `build.args`
- Dev compose is unaffected (Vite dev server reads `.env` natively)

AndriiPasternak31:
## Summary

Plumbs the two advertised in-app bug-reporting build knobs — `VITE_BUG_REPORTING_ENABLED` and `VITE_BUG_INTAKE_URL` — through the **production** frontend build path so they actually reach the shipped image. Before this, `HelpChatWidget.vue` (#1116) read both via `import.meta.env`, but `docker/frontend/Dockerfile.prod` declared only `VITE_API_URL` and `docker-compose.prod.yml` forwarded only that — so `VITE_BUG_REPORTING_ENABLED=false` in `.env` was inert and every standard prod build shipped with reporting on, pointed at the hosted intake, with no supported off switch.

Same packaging-gap class as #1039 / #1056, on the frontend **build-arg** surface instead of the compose `environment:` surface (Vite statically inlines `import.meta.env.*` at `npm run build`, so the lever must arrive as an `ARG`/`ENV` build arg, not runtime env).

Fixes #1489 · finding #4 of umbrella #1485.

## Changes

- **`docker/frontend/Dockerfile.prod`** — `ARG VITE_BUG_REPORTING_ENABLED` + `ARG VITE_BUG_INTAKE_URL` (with code-default values), exported as `ENV` **before** `RUN npm run build`.
- **`docker-compose.prod.yml`** — frontend `build.args` forwards both with `${VAR:-default}` fallbacks.
- **`.env.example`** — both documented, with a "VITE_* is world-readable, never a secret" warning and the CSP caveat.
- **`tests/unit/test_1489_vite_bug_build_args.py`** — 7 static guards over every layer, plus the load-bearing invariant that Dockerfile ARG default == compose `${VAR:-…}` == `HelpChatWidget.vue` code default, so an unset `.env` byte-reproduces today's behavior.
- **Feature-flow** `in-app-bug-reporting.md` + index row updated.

## Scope / boundaries

- The **disable** switch works end-to-end. **Repoint** (`VITE_BUG_INTAKE_URL` → self-hosted host) is still CSP-blocked because the intake host is hard-coded in `security-headers.conf` + `vite.config.js`; parameterizing the CSP from the build arg is a deferred #1485 follow-up. The AC asks for the disable lever, which is delivered.
- Dev compose is unaffected (the Vite dev server reads `.env` natively).

## Relation to umbrella #1485

This closes **finding #4** only. The other three findings (`OPERATOR_INTAKE_ENABLED`/`DO_NOT_TRACK`, `DISPATCH_BREAKER_ENABLED`, `PUBLIC_ACCESS_REQUESTS_ENABLED` — all on the backend compose `environment:` surface) are closed by the **sibling backend-trio PR**. This frontend PR is intentionally split out because the build-arg surface is a distinct fix shape.

## Testing

- `tests/unit/test_1489_vite_bug_build_args.py` — **7 passed** locally.
- Verified independently that all three defaulting layers agree on `https://intake.abilityai.dev/v1/report-bug` and `true`.

---

### Folded-in dev test-hygiene fixes (pre-existing, not part of #1489)

Rebased onto dev's tip, which is currently red from two pre-existing failures unrelated to the VITE build-arg change. Folded their fixes here to keep dev green through the rebase:

1. **sys.modules lint (from #1490)** — `test_fork_to_own.py` merged with 3 raw `del sys.modules[...]` and no baseline entry. Recorded them in `tests/lint_sys_modules_baseline.txt` (the lint's sanctioned escape hatch; zero test-behavior change). A snapshot/restore fixture was tried first but broke `test_owner_mismatch_forbidden` under full-suite ordering — reverted. Proper `monkeypatch.delitem` drain is a #1490-owned follow-up.
2. **`test_login_rate_limit_split.py` stub drift** — its fake `dependencies`/`models` module stubs drifted behind `routers/auth.py`'s imports (#187 JWT-revocation added 4 `dependencies` names + `models.User`), erroring all 8 tests at fixture setup. Synced the stubs → 8 errors → 8 passed. No open issue existed; captured in the commit.


AndriiPasternak31:
Heads-up @vybe: rebased onto latest `dev` and the `sys.modules pollution` lint went red — traced to #1490 (fork-to-own), whose `tests/unit/test_fork_to_own.py` merged with three bare `del sys.modules['services.*']` and no baseline entry, so **dev's own `backend-unit-test` tip is red** and every open PR inherits it via the merge-ref.

Rather than block this frontend PR on it, I folded in the fix as commit `43badbed`: added the linter's blessed snapshot/restore exemption pair (`_STUBBED_MODULE_NAMES` + autouse `_restore_sys_modules`, per the `test_telegram_webhook_backfill.py` precedent) — snapshots the `services.*`/`database` sys.modules subtrees around each test and restores after, so the dynamic per-test purges can't leak across the session. Behavior-preserving: the file's 40 tests still pass, lint now treats it as exempt. This also unbreaks dev's lint once merged. Shout if you'd rather own this as a separate #1490 follow-up and I'll drop the commit.

AndriiPasternak31:
Update on the folded dev-hygiene fixes (revised from my earlier comment) @vybe @dolho:

**1. sys.modules lint (from #1490).** I first tried the autouse snapshot/restore fixture on `test_fork_to_own.py` — but under full-suite ordering it re-adds modules a test deliberately deletes and broke `test_owner_mismatch_forbidden` (the `regression diff` job flagged it as a new HEAD failure; it passed in isolation, failed in the 3561-test run). Reverted that and instead **recorded the three pre-existing violations in `tests/lint_sys_modules_baseline.txt`** — the lint's explicit escape hatch for accepted pre-existing debt, zero test-behavior change. @vybe, properly draining those three `del sys.modules` to `monkeypatch.delitem` in your file is a clean #1490 follow-up if you want it.

**2. Second pre-existing dev failure unmasked.** With the lint no longer gating ahead of pytest, `test_login_rate_limit_split.py` showed 8 fixture errors on dev's tip: `routers/auth.py` grew `from dependencies import (…)` (+4 names via #187 JWT-revocation) and `from models import User`, but the test's fake-module stubs still defined the old subset, so every test errored at import. Synced the stubs (bare lambdas + a bare `User` class; none exercised by the rate-limit paths). **8 errors → 8 passed.**

Both are pre-existing on `origin/dev` (not caused by the VITE build-arg change) — folded here to keep dev green through the rebase. No issue existed for the login one; captured in the commit + here. Frontend #1489 change itself is unchanged.

vybe (approved):
Validated via /validate-pr: closing keyword ✅ (Fixes #1489), targets dev ✅, focused 8-file diff, security greps clean, feature-flow + .env.example updated, cross-layer default-agreement guard is exactly the right invariant. Ran the 15 touched unit tests locally — all pass (required checks don't run unit tests). Folded dev-hygiene fixes (sys.modules baseline for #1490 debt, auth stub sync) are justified — dev tip is red without them.

PR #1495, opened 2026-07-06
fix(canary): B-01 queue-status coherence — backend-consistent Side B + confirm-re-read (#1450)

### LINKED ISSUE #1450: canary: B-01 queue-status coherence reads are not temporally atomic nor backend-consistent (raw sqlite3 vs get_engine)
## Summary

Canary **B-01 (queue-status coherence)** collects its two compared values from two independent reads that are neither **temporally atomic** nor **backend-consistent**. On the shipping SQLite backend with the canary default-OFF these gaps are latent, so they were left out of scope for the #1446 P1 (which fixed the *test-harness* `sys.modules` leak that was the actual reported flake). This issue tracks the two production-side gaps the #1446 investigation surfaced and deferred.

Follow-up of #1446. Epic #411 (CANARY-001 harness).

## The two gaps

`src/backend/canary/snapshot.py :: collect_snapshot()` builds each `AgentSnapshot` from two separate reads, per agent:

```python
# snapshot.py ~614
execs = _collect_executions(name)            # raw sqlite3 via db.connection.get_db_connection() at DB_PATH
# snapshot.py ~627
queued_via_service = _collect_queued_count_via_service(name)   # db.get_queued_count → SQLAlchemy get_engine()
```

B-01 (`canary/invariants/b01_queue_status_coherence.py`) then fires `critical` on any inequality between `len(queued_exec_ids)` (from the first read) and `queued_count_via_service` (from the second), with **no tolerance and no confirm-re-read**.

### (a) Temporal non-atomicity
The two reads are separate queries at different instants. Under a live canary against real fleet load (e.g. the `canary-fleet-burst` generator), a concurrent enqueue or backlog-drain landing **between** the two reads produces a transient count mismatch → a spurious `critical` B-01 violation and a green→red Slack alert. The window is small but non-zero and the invariant has no confirm-read to absorb it.

### (b) Backend divergence under the configurable engine (#300/#1093)
`_collect_executions` reads raw `sqlite3` at `db.connection.DB_PATH`; `_collect_queued_count_via_service` goes through `get_engine()`, which honors `DATABASE_URL` (incl. PostgreSQL). On SQLite both resolve to the same file (why this is invisible today). **On a PostgreSQL backend they are two different databases** — the canary would compare Postgres truth to a stale/absent SQLite file, so B-01 (and any other raw-sqlite reader in the collector) is not backend-aware. This becomes fatal under the Postgres direction in `docs/planning/TARGET_ARCHITECTURE.md` and the SQLite end-of-support date (#1278, 2026-09-01).

## Why deferred, not fixed in #1446
#1446's reported symptom was a **unit-test** flake (B-01 firing only under full-suite load), root-caused to a never-restored `sys.modules.setdefault("database", MagicMock())` leak — a harness bug, fixed test-only. These two gaps are genuinely separate latent issues with no evidence of a live-canary fire yet; they're masked because the canary is default-OFF (`CANARY_ENABLED=1` only on staging/dev). See `docs/memory/learnings.md` (2026-07-04 entry) and `.plan/issue-1446-context.md` Hypothesis B.

## Acceptance criteria
1. B-01's two compared sides are sourced from the **same backend** (route `_collect_executions` — or at least B-01's queued-id list — through `get_engine()` so both honor `DATABASE_URL`). A canary-wide raw-sqlite3 → engine migration of the snapshot collector is the larger, related cleanup; scope this issue to at least B-01 consistency.
2. B-01 tolerates a transient in-flight mismatch — e.g. a single confirm-re-read of both sides before reporting, so a concurrent enqueue/drain doesn't produce a false `critical`.
3. B-01 stays a genuine coherence check, not a tautology — the two sides must remain independent code paths (production accessor vs independent id-list read), per the invariant's docstring.
4. Do **not** add a cache / second representation of the queued count (that is exactly the drift B-01 guards against, and moves away from the single-store target).
5. Regression coverage: a test that drives `collect_snapshot()` on a non-SQLite / diverged-backend configuration (or simulates the temporal race) and proves B-01 does not false-fire.

## Notes
- Keep the fix out of the CAS/`schedule_executions` transaction machinery (#1095 is research-gated NO-GO for broad changes there).
- `docs/planning/TARGET_ARCHITECTURE.md`: consult before touching the collector; the target makes `COUNT(*) WHERE status='queued'` the Lever-1 inbox-depth signal — one store, one truth.

AndriiPasternak31:
Fixes #1450

## Summary

Canary **B-01 (queue-status coherence)** collected its two compared values from two reads that were neither **temporally atomic** nor **backend-consistent**. Latent on SQLite (canary default-OFF, both reads hit the same file), but fatal under the Postgres direction. This is the production-side residue #1446 explicitly deferred (that P1 fixed only the test-harness `sys.modules` leak).

## The two gaps closed

**(a) Temporal non-atomicity** — the accessor count and the id-list read happened at different instants; a concurrent enqueue/backlog-drain landing *between* them produced a transient mismatch → spurious `critical` + green→red Slack alert.

**(b) Backend divergence (#300/#1093)** — Side B (`queued_exec_ids`) was collected via **raw sqlite3** at `DB_PATH`, while Side A (`db.get_queued_count`) honors `get_engine()`/`DATABASE_URL`. On Postgres these are two different databases, so B-01 compared Postgres truth against a stale/absent SQLite file.

## Fix

- **New Side B** `queued_ids_via_engine` (`snapshot.py::_collect_queued_ids_via_engine`) reads queued ids through the **same `get_engine()` seam** as the accessor — `SELECT id`/literal `'queued'` vs the accessor's `COUNT(*)`/`TaskExecutionStatus.QUEUED` enum, so it shares a *database* but not a *code path* (stays a genuine coherence check, not a tautology — AC #3). `queued_exec_ids` stays raw-sqlite for B-02/E-02.
- **One confirm-re-read** in `collect_snapshot()`: on a first-read mismatch, re-read both sides once. A transient race self-resolves (green); a persistent drift survives and still fires. An unconfirmable confirm degrades to a **skip**, never a false `critical` (AC #2).
- **Engine-read failure → B-01 skip** (`queued_ids_via_engine=None`), recorded in `sources_unavailable` — never falls back to the raw set, which would re-arm gap (b) on Postgres.
- No cache / second representation of the count added (AC #4 — that's the drift B-01 guards).

Collector is now *half-migrated* (B-01's queued id-set is engine-backed; running-side/known-agents/orphan reads stay raw-sqlite — invisible on SQLite, tracked for the collector-wide migration follow-up).

## Tests

`tests/test_canary_invariants.py` (all 97 pass):
- `test_b01_backend_consistency_on_diverged_backend` — new `canary_db_split` fixture models a Postgres-style diverged backend (raw file ≠ engine file) without a live PG; proves B-01 stays green when the two backends diverge (AC #5).
- `test_b01_tolerates_transient_race` — confirm-re-read absorbs a between-reads race.
- `test_b01_confirm_does_not_mask_persistent_drift` — confirm does **not** hide a real regression.
- `test_b01_skips_on_engine_read_failure` / `test_skips_when_engine_ids_none` — engine-read failure degrades to skip.
- Adds `_STUBBED_MODULE_NAMES` + autouse `_restore_sys_modules` — file is now lint-clean under `tests/lint_sys_modules.py` (the #1446 discipline).

## Docs

- `architecture.md` — B-01 row updated (both sides via `get_engine()`, confirm-re-read, engine-fail skip, half-migrated note).
- `feature-flows.md` — Recent Updates row.

## Validation

- `/validate-pr` clean: security scans clear, no new backend module (no Dockerfile packaging), no new env var. `sys.modules` linter: this file `5 → 0`. All 97 canary tests pass.
- Bug fix (no requirements.md change per Rule #1).


vybe (approved):
Validated: Fixes #1450 ✅, focused 5-file diff. Side-B migration to the get_engine() seam is correct and the non-tautology property is preserved (SELECT id + literal 'queued' vs COUNT(*) + enum). Confirm-re-read races degrade to skip, never a spurious critical — right call. Resolved the feature-flows.md index conflict against dev (kept both rows) and pushed the merge; ran tests/test_canary_invariants.py locally on the merged branch — 97 passed.

Issue #728, opened 2026-05-08
bug: agent-server.py spins at 90% CPU on OAuth token auth failure, blocking CB recovery


vybe:
## Summary

When an agent's Claude Max subscription token is expired and an execution is dispatched, `agent-server.py` enters a CPU spin loop (~87–91%, rising over time) with no claude child process present. The container becomes unresponsive to all HTTP requests, blocking both the circuit-breaker half-open recovery path and the cleanup service's re-verify step. The execution stays \`running\` in the DB until the full per-agent execution timeout fires (up to 7200 s), causing a compounding outage on that agent slot.

This is distinct from the drain deadlock fixed in #657 (normal exit path). The spin occurs specifically when claude exits immediately due to an authentication failure on an expired OAuth token.

## Component

Agent Runtime — `docker/base-image/agent_server/services/claude_code.py`

## Priority

P1 — single-agent slot blocked for up to 2 h; CB stays open during that window; no workaround short of manually restarting the container.

## Observed Behaviour

```
# agent-server.py CPU climbs and stays high; no claude child process
python3 /app/agent-server.py   87% CPU → 88% → 89.7% → 90.8% → 91% (over ~40 min)
node   moltbook-mcp             0.1% CPU  (MCP HTTP server — unrelated)
# no claude / npx process visible

# Backend logs: circuit never closes
Circuit OPENED for agent <name> after N failures  (~1 per 18 s, counting indefinitely)

# Cleanup service (every 5 min):
Skipping <execution_id> for '<agent>' — agent unreachable during re-verify (#378);
  Phase 1 stale cleanup is fallback
# Redis slot freed; DB execution stays status=running

# Execution self-resolves only when the 7200 s timeout fires
```

## Root Cause (Hypothesis)

`execute_headless_task` dispatches subprocess work via:

```python
return_code = await asyncio.wait_for(
    loop.run_in_executor(None, read_subprocess_output_with_timeout),
    timeout=timeout_seconds + 60   # e.g. 7270 s for a 7210 s timeout agent
)
```

Inside `read_subprocess_output_with_timeout`, after `process.wait()` returns (claude exits fast on auth failure), `asyncio.run(_drain_reader_threads(...))` is called. Two failure modes are suspected:

[...]

PR #1503, opened 2026-07-07
fix(agent-server): kill the process group on a budget-exceeded drain (#1502)

### LINKED ISSUE #1502: bug: budget-exceeded subprocess drain leaks a reader thread that pegs a CPU core until container restart (residual of #728)
## Context
#728 bounded the `safe_close_pipes` TextIOWrapper-lock deadlock with a 90s `_DRAIN_BUDGET_SECONDS` (`docker/base-image/agent_server/services/subprocess_lifecycle.py`, `_drain_bounded`). On budget-exceeded it returns `budget_exceeded` and **leaks the reader thread** to unwedge the task/executor. That fixed the up-to-7200s hang.

## Problem (residual of #728)
On `budget_exceeded`, the leaked reader thread (and/or an un-reaped claude grandchild still holding the pipe) then **spins at ~100% CPU with no visible `claude` child process**, and stays there **until the agent container is stopped/restarted**. The 90s cap recovers the *task/server* but not the *CPU* — a single agent slot burns a full core with no automatic recovery.

The #728 fix explicitly accepted leaking the thread ("Leaked reader threads are daemon threads and die with the container") — but "dies with the container" means the only recovery is a manual restart, and in the meantime the host CPU is pegged.

## Repro (observed on the Client Portal, but trigger-agnostic)
Dispatch a turn whose claude subprocess emits a large/blocking stream (e.g. the agent reads an ~83 KB binary file through `claude --print --output-format stream-json`). Then:
- agent-server log: `WARNING [Subprocess] Drain budget (90s) exceeded — safe_close_pipes may have deadlocked with reader thread's TextIOWrapper lock; reader threads are leaked daemon threads (pid=N). Issue #728.`
- `docker stats` → the agent container at ~100% CPU
- container `ps -eo pid,pcpu,comm` → only `agent-server.py` (no `claude`/`node` child visible)
- `POST /api/agents/{name}/executions/{id}/terminate` → **504** (agent-server can't be reached / is busy)
- only `docker stop <agent>` clears it

## Proposed fix
On `budget_exceeded`, actively tear down instead of leaking:
- SIGKILL the claude **process group** (`utils/subprocess_pgroup.py` pgid) so any grandchild holding the pipe dies and the pipe EOFs, letting the reader thread unblock/exit;
- only then leak as a last resort.

This frees the core without a container restart. Distinct from #728, which only unblocks the executor thread.

## Severity
P1 — a single agent can hold a full CPU core indefinitely with no automatic recovery; `terminate` and the cleanup/CB recovery paths can't reach the wedged agent-server.

## Related
- #728 (closed) — the drain-deadlock 90s-budget mitigation this is the residual of
- #657 — the original drain-deadlock on the normal-exit path
- #411 — R-01 canary (zombie-claude) guards a related class
- Trigger-side mitigation already shipped for the Client Portal (images sent as vision instead of read as text, so the portal no longer emits the large-binary stream that trips this) — trinity-enterprise#104.

dolho:
## Problem
#728 bounded the `safe_close_pipes` TextIOWrapper-lock deadlock with a 90s drain budget — but on budget-exceed `_drain_bounded` only unblocked the executor thread and **left the process group alive**. The leaked reader thread stays blocked reading a pipe a grandchild still holds open, **spinning at ~100% CPU with no visible `claude` child, until the container is restarted** (`terminate` 504s in the meantime).

## Fix
On budget-exceed, `_drain_bounded` now SIGKILLs the whole **process group** (via the already-imported `_terminate_process_group`, forwarding the spawn-captured `pgid`). The pipe EOFs → the daemon reader returns from `readline()` and exits → the core is freed, no container restart needed. Best-effort: the group kill never raises into the caller.

Distinct from #728 (which only unblocked the executor thread). This is the platform-level fix for the residual filed in #1502; the trigger-side mitigation for the Client Portal (images sent as vision instead of read as text) already shipped in trinity-enterprise#104.

## Testing
`tests/unit/test_drain_bounded.py` (+2): a budget-exceeded drain kills the group with the captured pgid; a failing kill is swallowed and still returns `budget_exceeded`. **10 passed** (full file, via the tests venv).

## Deploy note
Agent-server / base-image change — takes effect after a **base-image rebuild + agent recreate** (`scripts/deploy/build-base-image.sh`).

Fixes #1502.

### DIFF
--- a/docker/base-image/agent_server/services/subprocess_lifecycle.py
+++ b/docker/base-image/agent_server/services/subprocess_lifecycle.py
@@ -127,6 +127,27 @@ def _drain_bounded(
             "leaked daemon threads (pid=%s). Issue #728.",
             _DRAIN_BUDGET_SECONDS, process.pid,
         )
+        # #1502: #728 unblocked the executor thread but LEFT THE PROCESS GROUP
+        # ALIVE. The leaked reader is blocked reading a pipe a grandchild still
+        # holds open — it spins/pegs a core until the container is restarted.
+        # SIGKILL the whole group so that pipe EOFs, letting the daemon reader
+        # return from readline() and exit, freeing the CPU. Best-effort: never
+        # let cleanup raise into the caller, and only meaningful when the pgid
+        # was captured at spawn (after wait() reaped the parent, a pid lookup
+        # can't reach grandchildren — see terminate_process_group docstring).
+        try:
+            _terminate_process_group(process, 0, pgid=pgid)
+            logger.warning(
+                "[Subprocess] #1502: SIGKILLed the process group (pgid=%s, pid=%s) "
+                "after the budget-exceeded drain so the leaked reader can EOF and "
+                "free the core.",
+                pgid, process.pid,
+            )
+        except Exception as e:  # pragma: no cover — cleanup must never crash the caller
+            logger.warning(
+                "[Subprocess] #1502: group kill after budget-exceed failed "
+                "(pgid=%s, pid=%s): %s", pgid, process.pid, e,
+            )
         return "budget_exceeded"
     # done is set → _target finished, so ``outcome`` is fully settled.
     if outcome == "errored":
--- a/tests/unit/test_drain_bounded.py
+++ b/tests/unit/test_drain_bounded.py
@@ -81,6 +81,56 @@ def test_drain_bounded_returns_within_budget_when_drain_hangs(monkeypatch):
     )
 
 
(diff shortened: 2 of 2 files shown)


vybe (approved):
Validated via /validate-pr: best-effort group-kill on budget-exceed, correct pgid forwarding, failure-swallow covered by tests (test_drain_bounded +2), CI green incl. pytest head runs, security sweep clean. Note: needs base-image rebuild + agent recreate to take effect. Approving.

PR #1509, opened 2026-07-07
fix(logs): retention can delete archived originals — rw mount + dir ownership (#1478)

### LINKED ISSUE #1478: Log retention cleanup cannot delete originals in prod: backend mounts trinity-logs:/data/logs:ro
## Symptom

`LOG_RETENTION_DAYS` (#1039, wired into prod compose in #1256/v0.7.0) archives old log files but **never frees disk space** on prod deployments. Every archived file logs:

```
ERROR services.log_archive_service: Error archiving agents-2026-06-02.json: [Errno 30] Read-only file system: '/data/logs/agents-2026-06-02.json'
```

The gzip archive is written to `/data/archives` successfully, but the original in `/data/logs` cannot be unlinked, so the volume grows unbounded (~900MB/day of agent logs on a busy instance).

## Cause

`docker-compose.prod.yml` mounts the logs volume **read-only** on the backend (line ~200):

```yaml
backend:
  volumes:
    - trinity-logs:/data/logs:ro   # <-- log_archive_service cannot delete
...
vector:
  volumes:
    - trinity-logs:/data/logs      # rw (writer)
```

`log_archive_service.py` archives then calls `log_file.unlink()` on `LOG_DIR = Path("/data/logs")` — which is ro in the backend container on prod. Dev compose is presumably rw, so this only bites production.

## Observed

- eu2 (v0.7.0, `13961fac`), 2026-07-06: logs volume had grown to **32G** (files back to May 24). Manual `POST /api/logs/archive` archived 26 files (~7.5G → 164M gzipped) but all 26 deletes failed with Errno 30.
- Ops workaround applied: verified each archive (`gunzip -t`) then removed originals from the host volume path. Not sustainable fleet-wide.

## Suggested fix

Either give the backend a rw mount of the logs volume, or run the retention delete from a component that has rw access (vector sidecar / dedicated cleanup job). A startup WARNING when `LOG_DIR` is not writable would also make the failure visible instead of silent per-file errors at cleanup_hour.

## Environment

- Trinity v0.7.0 (`13961fac`), docker-compose.prod.yml, LOG_RETENTION_DAYS=30 (env), cleanup_hour=3
- Instance: eu2 (ops-managed)

dolho:
Closes #1478.

## Problem
`LOG_RETENTION_DAYS` archived old logs but **never freed disk** — every delete failed `Errno 30: Read-only file system`, volume grew unbounded (~900MB/day; eu2 hit 32G).

## Two blocking layers (both fixed)
1. **Read-only mount** — `docker-compose.prod.yml` **and** `docker-compose.yml` both mounted `trinity-logs:/data/logs:ro` on the backend (the issue assumed dev was rw — it wasn't). → changed to **rw**. Vector stays the sole writer of new lines; backend only reads (gzip) + deletes archived originals.
2. **Directory ownership** — Vector runs as root and creates `/data/logs` root:root 0755, so even rw the backend (UID 1000) can't `unlink` (deleting needs write on the *directory*). Vector can't self-fix (`cap_drop: ALL` → no CAP_CHOWN). → added a one-shot **`logs-init`** container (root alpine) that `chown 1000:1000` + `chmod 775` the dir before the backend starts (`depends_on: service_completed_successfully`). Chowns the dir only (fast, not -R); Vector's future root-owned files stay deletable via the writable dir.

Plus: `log_archive_service.start()` now logs a **WARNING** if `LOG_DIR` isn't writable — visible at startup instead of silent per-file errors at cleanup_hour.

## Verified live (dev)
- Reproduced `:ro`: `touch /data/logs/... → Read-only file system`.
- After fix: dir `1000:1000 775`; backend writes; **a root-owned file in /data/logs is deletable by UID 1000** (Errno 30 gone). `logs-init` exits 0.

## Notes
- `logs-init` runs as root (needs CAP_CHOWN) but uses a **stock alpine** image, not a Trinity-built one — Invariant #17 ("Trinity-built images non-root") doesn't apply; it's ephemeral (exits immediately).
- Alternative considered: run Vector as UID 1000 — rejected because Vector needs root for the Docker-socket / file log sources (dev override reads root-owned `/var/lib/docker`).

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -197,12 +197,19 @@ services:
       # Read-only access to Vector's logs so log_archive_service.py (reads
       # /data/logs) can run in prod. Mirrors docker-compose.yml; without it the
       # forwarded LOG_* knobs are inert (LOG_DIR not found) — the #1039 class.
-      - trinity-logs:/data/logs:ro
+      # #1478: MUST be read-WRITE — log_archive_service archives then unlink()s
+      # the originals. A ':ro' mount lets archiving succeed but every delete fail
+      # (Errno 30), so the volume grows unbounded (~900MB/day on a busy instance).
+      # Vector is still the sole writer of new log lines; the backend only reads
+      # + deletes archived originals.
+      - trinity-logs:/data/logs
     depends_on:
       redis:
         condition: service_healthy
       vector:
         condition: service_healthy  # Ensure Vector is ready before backend starts agents
+      logs-init:
+        condition: service_completed_successfully  # #1478: dir owned by UID 1000 before retention runs
     networks:
       - trinity-platform   # Redis, scheduler, vector, OTel
       - trinity-agent      # Frontend, mcp-server, agent containers, cloudflared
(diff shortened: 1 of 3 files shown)


vybe (approved):
Validated via /validate-pr: required checks green, no secrets/PII, scope focused. Approving for sequential merge to dev.

PR #1510, opened 2026-07-07
docs(target-arch): SKIP LOCKED claim semantics + agent-scoped-key auth for pull endpoints

vybe:
Docs-only refinement of the pull-coordination target design (Epic Abilityai/trinity-enterprise#1045 context — umbrella #1081). Refs #1081.

## What

Two hardening clarifications in `docs/planning/TARGET_ARCHITECTURE.md`:

1. **Atomic claim: `FOR UPDATE SKIP LOCKED` is load-bearing, not an optimization.** The claim subselect now reads `WHERE status='queued' … FOR UPDATE SKIP LOCKED`. Without it, two concurrent claimers can resolve the subselect to the same row — under READ COMMITTED the post-lock-wait re-check only re-evaluates the outer `id =` predicate — so both "win" the claim and the task double-executes.

2. **Pull endpoints authenticate with the agent's own agent-scoped MCP key, not `X-Internal-Secret`.** Same pattern as the heartbeat and the #1083 result callback; the claim is scoped server-side to the authenticated agent's queue. A fleet-wide shared secret inside every container would let one compromised agent claim a sibling's tasks (reading their payloads) and forge its results.

## Why now

Locks these two decisions into the target doc before pull-mode implementation work picks them up (`task_execution_service` / `capacity_manager` consumers are told to consult this doc per CLAUDE.md Rule 7).

Docs-only — no code change, no tests.


dolho:
**Review: ✅ Approve — merge anytime**

Docs-only (2 lines in `TARGET_ARCHITECTURE.md`), no code/tests. Both hardening points are correct and worth locking in before pull-mode implementation consumes this doc:

1. **`FOR UPDATE SKIP LOCKED` is load-bearing, not an optimization** — accurate. Under READ COMMITTED the post-lock re-check only re-evaluates the outer `id =` predicate, so two claimers resolving the subselect to the same row would both "win" → double-execute. Correct framing.
2. **Agent-scoped MCP key, not `X-Internal-Secret`** — consistent with the heartbeat (#307) and #1083 result-callback auth, and avoids the fleet-wide-shared-secret cross-agent claim/forge surface. Server-side queue scoping is the right guarantee.

No concerns.

dolho (approved):

PR #910, opened 2026-05-21
feat(enterprise): audit log dashboard + entitlement seam (#941, #847)

dolho:
## Summary
Related to **#847** (enterprise edition architecture spike) and **#941** (audit log dashboard).

This PR ships **two things**:

1. The **enterprise-edition seam** — open-core architecture for Trinity, with the private companion repo [`Abilityai/trinity-enterprise`](https://github.com/Abilityai/trinity-enterprise) mounted at `src/backend/enterprise/`. Loads conditionally so OSS-only builds (no submodule access) work unchanged.
2. The **first concrete enterprise feature** — admin audit log dashboard at `/enterprise/audit`. Frontend ships in the OSS bundle, gated by the `enterprise_features` flag exposed via `/api/settings/feature-flags`. Backend audit log endpoints stay OSS (they predate the seam via SEC-001/#20).

The original Phase 0 PoC mounted a `/api/enterprise/sso/*` mock router with seeded providers; that stub is **removed** in favour of a real, useful first enterprise demonstration. SSO returns in a later PR with a real implementation, not a stub.

### Public-repo seam (#847)

- `services/entitlement_service.py` — `EntitlementService` with `register_module(feature_id)` registry. `TRINITY_OSS_ONLY=1` env flips every check to False.
- `dependencies.py:requires_entitlement(feature_id)` — FastAPI dependency raising HTTP 403 with the missing feature name.
- `main.py` — conditional `try: from enterprise.backend import register_enterprise except ImportError`.
- `/api/settings/feature-flags` extended with `enterprise_features: list[str]`.
- `.gitmodules` entry, `docker-compose.yml` env pass-through for `TRINITY_OSS_ONLY`.
- CI workflow `build-without-submodule.yml` asserts OSS-only boot.

### Frontend enterprise scaffolding (#847)

- `views/enterprise/Index.vue` — landing page with feature catalogue cards.
- `stores/enterprise.js` — Pinia store caching `enterprise_features` from the feature-flags API.
- `NavBar.vue` — Enterprise nav link, hidden when no features entitled.
- `router/index.js` — entitlement-gated routes (`requiresEntitlement` / `requiresAnyEntitlement` metas + `beforeEach` guard).

### Audit log dashboard (#941)

- `views/enterprise/Audit.vue` — paginated table + filter form + side detail panel + hash-chain disclosure.
- `stores/auditLog.js` — domain store (entries, filters defaulting to last 24h, distinct-value lists, pagination, selected entry).
[...]

Issue #32, opened 2026-02-24
Quick Instance Deploy


vybe:
One-command agent creation from template with credentials. 'Deploy Ruby for Client X' flow.

vybe:
Incubation triage 2026-07-03: closing as superseded — instance provisioning is now epic #1215 (one-click arc: VM image #835, cross-platform bootstrap #1217, agent-guided install ent#39).

PR #1511, opened 2026-07-07
test: re-pin entitlement seam to audit+SSO now that SSO-OIDC landed (#1447)

### LINKED ISSUE #1447: test: test_submodule_registers_audit_not_sso is stale since SSO-OIDC landed (#1303) — fails locally whenever the enterprise submodule is mounted

webmixgamer:
## Summary
`test_submodule_registers_audit_not_sso` pinned the enterprise submodule's `__init__.py` to *"no `.sso` import"* — a stale assertion from the #910/#941 era, when only the removed SSO PoC stub existed. Real SSO-OIDC (#32) has since landed via #1303, so the pin **failed on every entitled clone with the submodule mounted** (CI is unaffected — it skips when the submodule is absent, which is exactly where this trains people to ignore red tests).

## Changes
`tests/unit/test_847_entitlement_seam.py` — update the pin to the current contract:
- keep the direct `register_module("audit")` assertion (UI-gating module)
- replace the two `.sso`-absence assertions with presence checks on `from .sso import register as register_sso` + `register_sso(app)` (SSO self-registers its `sso` feature_id inside `.sso.register`, so `register_module("sso")` is intentionally **not** called in `__init__.py`)
- rename `..._not_sso` → `..._and_sso` and refresh the docstring

## Test Plan
Verified locally against a mounted entitled clone (via the `trinity-backend` image, which carries all deps):
- [x] **Passes** with the enterprise submodule mounted — `1 passed`
- [x] **Skips cleanly** without it (OSS-only build) — `1 skipped`
- [x] Single test node changed; no other tests affected

CI skips this test by design (the OSS unit suite has no submodule; guarded by `build-without-submodule.yml`), so the mounted-pass evidence above is the authoritative check.

Fixes #1447

### DIFF
--- a/tests/unit/test_847_entitlement_seam.py
+++ b/tests/unit/test_847_entitlement_seam.py
@@ -236,9 +236,15 @@ def test_main_py_uses_conditional_enterprise_import():
 # -----------------------------------------------------------------------------
 
 
-def test_submodule_registers_audit_not_sso():
-    """#941: the enterprise submodule swaps the SSO PoC for `audit`
-    registration. SSO returns later with a real implementation.
+def test_submodule_registers_audit_and_sso():
+    """The enterprise submodule registers `audit` (UI-gating only, via a direct
+    `register_module("audit")`) and — since SSO-OIDC (#32) landed via #1303 — a
+    real SSO implementation, wired the same way as the other real modules
+    (`from .sso import register as register_sso` + `register_sso(app)`, which
+    self-registers the `sso` feature_id inside `.sso.register`).
+
+    This replaces the earlier `not_sso` pin from the #910/#941 era, when only
+    the removed SSO PoC stub existed and the `.sso` package was asserted absent.
 
     Static check on `src/backend/enterprise/backend/__init__.py`. The
     submodule may not be checked out in every CI job (OSS-only build),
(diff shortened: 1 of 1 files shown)


dolho (approved):
LGTM — verified against the real submodule.

**Contract confirmed** in `src/backend/enterprise/backend/__init__.py` (mounted entitled clone):
- `register_module("audit")` — direct call at L92 ✓
- `from .sso import register as register_sso` — L138 ✓
- `register_sso(app)` — L139 ✓
- and `.sso.register` self-registers `FEATURE_ID = "sso"` (sso/`__init__.py` L58), so `register_module("sso")` is correctly **absent** from `__init__.py` — the dropped assertion was right to drop.

The rename `..._not_sso → ..._and_sso` + docstring now describe the post-#1303 reality; SSO wired identically to the other real modules (user_management/siem/retention/mcp_connector/two_factor). Skip-when-absent guard is intact, so OSS-only CI stays green.

One tiny nit (non-blocking): the two new asserts are exact-substring pins, so a benign reformat of the import line (e.g. line-wrap) would false-fail — but that matches the existing static-check style in this file, so no change needed.

Good catch on the stale pin that only bit entitled clones (where CI skips). Approving.

PR #1514, opened 2026-07-07
fix(templates): curate the agent template catalog — hide fixtures, order starters first (#1513)

### LINKED ISSUE #1513: bug: agent template catalog surfaces internal test/canary fixtures + a broken stub as real templates
## Summary

`GET /api/templates` returns every directory under `config/agent-templates/` with no filtering, so **internal test/canary fixtures and a broken stub are surfaced to users as if they were real, deployable templates** — and some even rank above the genuine starters. This makes the Create Agent / Templates experience look like "a pile of irrelevant, never-tested templates" and directly undermines onboarding.

## Context

Found during an onboarding-flow audit (2026-07). Of 25 bundled template directories, only 3 are real starters (`scout` / `sage` / `scribe`). The catalog also exposes:

- **7 internal test/canary fixtures** — `test-echo`, `test-counter`, `test-delegator`, `test-codex`, `test-gemini`, `sleep-echo`, and `test-leak-hook` (whose own description says "do NOT deploy in production"). `config/agent-templates/README.md` explicitly labels these "internal test/canary fixtures — not templates to build on," yet they ship a valid `template.yaml`, so nothing filters them out of the API response.
- **2 demo agents** (`demo-researcher`, `demo-analyst`).
- **A broken `cornelius/` stub** — only `.trinity/brain-orb/` hook scripts, no `template.yaml` / `CLAUDE.md`; the catalog builder silently skips it, leaving a dead directory.

Ordering is inverted: `test-codex` / `test-gemini` carry `priority: 10` and rank **above** the real starters (`priority: 20`). Separately, the Templates tab (`src/frontend/src/views/Templates.vue`) renders its own GitHub-only grid from a separate `/api/templates` fetch, so it shows a **different** set than `CreateAgentModal.vue` — the same catalog appears two inconsistent ways.

Part of the pre-release onboarding-simplification push (First-run & self-host robustness).

## Acceptance Criteria

- [ ] `GET /api/templates` no longer returns internal test/canary fixtures or demo fixtures to normal users — filtered via an **explicit** signal (e.g. a `hidden`/`internal` flag in `template.yaml`, or a curated exclude), not fragile name matching alone.
- [ ] The broken `config/agent-templates/cornelius/` stub is removed (making it a valid default template is tracked separately).
- [ ] Ordering surfaces real starters first; no fixture outranks a genuine starter.
- [ ] The Templates tab and CreateAgentModal show the **same** curated set from a consistent source (no divergent GitHub-only grid).
- [ ] A test asserts fixtures are excluded from the default catalog response, so they can't silently reappear.

## Technical Notes

- Catalog assembly: `src/backend/services/template_service.py` (`get_all_templates` / `get_local_templates`) reads every dir with a parseable `template.yaml`; add the filter here.
- Hardcoded GitHub list: `src/backend/config.py` (`DEFAULT_GITHUB_TEMPLATE_REPOS`).
- Frontend surfaces: `src/frontend/src/views/Templates.vue` (own grid + separate fetch) and `src/frontend/src/components/CreateAgentModal.vue` (the shared modal, sole caller of `POST /api/agents`).
- Keep fixtures usable for internal tests (they can still be created by id) — the goal is to hide them from the default user-facing catalog, not delete their `template.yaml`.
- Cornelius bundling as a real default template is tracked separately; this issue only removes/neutralizes the dead stub.

vybe:
## Summary
- `GET /api/templates` no longer surfaces internal **test/canary fixtures**, **demo agents**, or the auto-deployed **trinity-system** agent as real user templates — they're excluded via an explicit `hidden: true` `template.yaml` flag (not fragile name-matching), while staying creatable-by-id so the canary/test harness is untouched.
- **Real starters now lead**: both template builders surface a coerced-int `priority`, so the router's `(priority, display_name)` sort orders `scout`/`sage`/`scribe` (priority 20) ahead of the `dd-*` suite.
- The Templates tab and Create Agent modal now show the **same curated set** (Templates.vue was GitHub-only).

## Acceptance criteria (all met)
- [x] Fixtures excluded via an explicit signal (`hidden: true`), not name matching
- [x] Broken `config/agent-templates/cornelius/` stub removed
- [x] Ordering surfaces real starters first; no fixture outranks a genuine starter
- [x] Templates tab + CreateAgentModal show the same curated source
- [x] A test asserts fixtures are excluded (+ a convention guard so they can't silently reappear)

## Changes
- `src/backend/services/template_service.py` — `_coerce_priority()` helper; surface `priority`+`hidden` in `_build_local_template`, `priority` in `_build_template`; filter `hidden` in `get_local_templates()` (LIST only — single-id + creation-by-id preserved).
- `config/agent-templates/*/template.yaml` — `hidden: true` on the 7 test/canary fixtures, 2 demo agents, and `trinity-system`.
- `config/agent-templates/scribe/template.yaml` — **quote the `usage:` value** (an unquoted `format:` colon made the file invalid YAML, so this core starter was silently dropped from the catalog).
- `src/frontend/src/views/Templates.vue` — add a Starter (local) Templates section.
- `config/agent-templates/README.md` + `docs/memory/feature-flows/template-processing.md` — reconcile docs with the hidden set and document the flag.
- `tests/unit/test_local_templates_listing.py` — exclusion, single-id survival, priority coercion, router-sort regression, real-catalog convention guard.

## Notable finds during review
- **Recovered a missing starter**: `scribe` was already absent from the catalog because its `template.yaml` failed to parse — fixed.
- **Prevented a latent 500**: surfacing a raw `priority: null`/`"high"` would `TypeError` the router sort; `_coerce_priority` guards it (with a regression test).
- `cornelius/` carried **no git-tracked files** (its only contents are globally gitignored `.trinity/` hooks) and never appeared in the catalog (no `template.yaml`) — removed the local artifact; produces no diff.

## Test plan
- [x] `pytest tests/unit/test_local_templates_listing.py` — 19 passed (11 existing + 8 new)
- [x] Adjacent template tests unaffected — `test_data_paths_template_surface`, `test_deploy_writable_templates` (5), `test_templates` (5 passed / 3 skipped)
- [x] End-to-end catalog simulation against the real `config/agent-templates/`: fixtures excluded, `scout`/`sage`/`scribe` present and ahead of `dd-*`, hidden fixtures still resolve by id
- [ ] Manual: Templates tab + Create Agent modal render the same curated set

## Follow-up (deferred, out of scope)
- Extract a shared `<TemplateCard>` component — the new local card markup mirrors the GitHub card; deferred to avoid pulling `CreateAgentModal` into a P1 bug fix.

Fixes #1513

### DIFF
--- a/config/agent-templates/README.md
+++ b/config/agent-templates/README.md
@@ -8,7 +8,7 @@ Ready-made starting points for deploying an agent to Trinity. Each subdirectory 
 
 - **From a checkout** — copy a template directory, edit its `CLAUDE.md` + `template.yaml`, then deploy it: `cd <copy>/ && trinity deploy .` (see [Deploy an agent](../../AGENTS.md#deploy-an-agent-to-trinity)).
 - **From the web UI** — *Create Agent* → pick a template by `name`.
-- **By reference** — these `name`s are the built-in templates the platform ships; they appear in `GET /api/templates` and `list_templates` (MCP).
+- **By reference** — these `name`s are the built-in templates the platform ships; they appear in `GET /api/templates` and `list_templates` (MCP). Directories marked `hidden: true` in their `template.yaml` (see [Not starting points](#not-starting-points)) are excluded from that catalog but stay deployable by id (e.g. `local:test-echo`).
 
 ## Starter templates
 
@@ -19,9 +19,6 @@ Ready-made starting points for deploying an agent to Trinity. Each subdirectory 
 | `scout` | Market research analyst — discovers trends, analyzes competitors, identifies opportunities |
 | `sage` | Strategic advisor — synthesizes research into actionable recommendations |
 | `scribe` | Content writer — reports, proposals, and client deliverables |
-| `demo-researcher` | Minimal autonomous researcher (gathers + summarizes); good first deploy |
-| `demo-analyst` | Minimal analysis agent — answers strategic questions over research findings |
-| `trinity-system` | Platform operations manager — agent health, lifecycle, fleet management (system-scoped) |
 
 `scout` → `sage` → `scribe` are designed to work as a **consulting team**: Scout researches into a shared folder, Sage strategizes over it, Scribe writes the deliverable. Deploy them together to see agent-to-agent collaboration.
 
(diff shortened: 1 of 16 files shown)


dolho:
**Review: ✅ Approve**

Clean fix for #1513. Highlights:

- **`hidden: true` over name-matching** — machine-enforced exclusion, surfaced (not filtered) in `_build_local_template` so single-id resolve + creation-by-id still work for the canary/test harness; only `get_local_templates()` (LIST) excludes. Right seam.
- **`_coerce_priority`** prevents a real latent 500: a present-but-null/string `priority:` would `TypeError` the router's `(priority, display_name)` sort. Bool-exclusion (`isinstance(True, int)`) is a nice touch. Regression-tested.
- **Recovered `scribe`** — the unquoted `usage:` colon made the YAML invalid, silently dropping a core starter. Good catch.
- **Real-catalog convention guard** fails CI if a future fixture forgets the flag.

Verified the live template dir is repo-root `config/agent-templates/` (`template_service._local_templates_dir`, line 183), so the flagged set is the one users actually see.

Interaction with #1515: cornelius is added there without `hidden` (intended — flagship), and lands in the default-priority group. Land this PR first so the coercion guard is in place before cornelius's `template.yaml` is surfaced.

dolho (approved):

PR #1515, opened 2026-07-07
feat: bundle a Brain-Orb-enabled Cornelius in the default install (abilityai/trinity-enterprise#107)

vybe:
## Summary
- Auto-seeds Trinity's flagship **Cornelius** second-brain agent on a **fresh install**, with the **Brain Orb enabled by default** — zero manual steps.
- **Path B (autoplan-reviewed): a local bundled template**, not github-native. The Brain Orb was found to be **already fully OSS** (platform flag-gated, *not* entitlement-gated), so the issue's "de-gating" dependency was **dropped**.
- github-native-at-boot was **rejected** on verified blockers: `create_agent_internal`'s dynamic-github path hard-requires a system PAT even for a *public* repo (`crud.py:302`), and a boot-time clone adds a network SPOF to startup.

## Changes
- **`config/agent-templates/cornelius/`** — bundled LOCAL template (from public `github.com/Abilityai/cornelius`): `template.yaml` (`capabilities: [brain-orb]`), `CLAUDE.md`, `.trinity/brain-orb/` hooks, a **pre-generated `resources/agent-visualization/data.json` seed** so the orb renders immediately, and a minimal seed `Brain/` vault.
- **`services/cornelius_agent_service.py`** — `ensure_seeded()`: first-run-only (durable `cornelius_seeded` flag — a **deleted** Cornelius is not resurrected), fresh-install-scoped (skipped when any non-system agent exists — established fleets aren't surprised on upgrade), **Redis SETNX lock** for the `--workers 2` race (fail-open, mirrors #1464), provisions `local:cornelius` (no PAT/network/clone), **existence-guarded** enable of `brain_orb_enabled` (never clobbers an admin OFF).
- **Two triggers** — `routers/setup.py` (fresh install: a `BackgroundTask` right after the admin account is created, since no owner exists pre-setup) + a `main.py` lifespan safety-net (upgrades: gated on `setup_completed && !cornelius_seeded`, backgrounded).
- **`create_agent_internal`** — `request` widened to `Optional[Request] = None` (boot-time caller has no HTTP request; the param was never dereferenced).
- **`db.count_non_system_agents()`** (`db/agents.py` + `database.py` facade) — the fresh-install signal. **No migration** (`system_settings` is free-form KV).

## Test Plan
- [x] New tests pass: `pytest tests/unit/test_ent107_cornelius_seed.py -v` (12 cases: gating, idempotency, fresh-install scoping, owner-deferral, docker skip, flag-seed guard, 409 convergence, SETNX lock, `local:cornelius` config, real-DB facade smoke)
- [x] Regression smoke: 191 passed across agent-creation / crud / database / setup / lifespan / quota
- [x] Setup test updated for the new background task
- [ ] Manual: fresh install → complete setup → Cornelius appears with the Brain Orb rendering the seed graph

## Notes
- **Known deviation** from the issue's decision #2: a local bundle has `origin=local` (no git upstream), so the default Cornelius won't auto-`git pull` upstream template updates; durable ownership is deferred to fork-to-own (**abilityai/trinity-enterprise#109**). Accepted at the autoplan gate.
- Bundle vendored content was secret-scanned clean; a bundle-level `.gitignore` overrides the monorepo root's `.trinity/`-prune and upstream's `data.json`-ignore so a plain `git add` stages the hooks + seed graph (verified).

Implements abilityai/trinity-enterprise#107 (private tracker — not auto-closed by this public PR; closed manually at release).


dolho:
**Review: ✅ Approve with notes**

The seeding seam is well-gated and genuinely fail-open: docker → already-seeded → fresh-install (with flag convergence) → owner-exists (defers *without* setting the flag, which is why the two-trigger design is needed) → SETNX cross-worker lock → provision → existence-guarded Brain-Orb enable → mark done only on success. `create_agent_internal(request=None)` is safe since the request is never dereferenced. Idempotency + 409-as-success backstop are correct.

Notes (non-blocking):

1. **`_release_lock()` deletes the key with no owner token.** If the 300s TTL expires mid-provision and a later boot re-acquires, the original worker's `finally` delete removes the *new* owner's lock. Low risk here (generous TTL, single-shot provision), but a check-and-delete (or Lua CAS, like the #1464 leader-lock) would close it.

2. **Cornelius sorts into the default-priority (100) group.** `template.yaml` sets no `priority:`, so after #1514's router sort it lands among the `dd-*` batch ordered by display name — it won't lead the starter list. If it's meant to be the featured flagship on the Create-Agent/Templates surface, give it an explicit low `priority:` (e.g. below scout/sage/scribe's 20).

3. **Merge order vs #1514.** Land #1514 (catalog curation + `_coerce_priority` guard) first, then this — otherwise the bundled cornelius `template.yaml` is surfaced without the coercion guard in place. No textual conflict (the old cornelius stub had no tracked files).

4. **26k-line vendored bundle** — PR states secret-scanned clean; please also confirm the pre-generated `resources/agent-visualization/data.json` seed graph carries no PII/internal content, since it ships in the public repo forever.

CI green. Tests cover the gate matrix well (12 cases + facade smoke).

dolho (approved):

Issue #589, opened 2026-04-29
SEC: Redis has no requirepass — agent containers have unauthenticated full read/write access to platform task queue (AISEC-C1)


vybe:
## Summary

Redis at `redis:6379` on the internal Docker network (`172.28.0.0/16`) requires no authentication. Any agent container can connect directly and perform arbitrary LRANGE, SET, DEL on the platform's task queue, slot tracking, and heartbeat keys — including reading and injecting tasks belonging to other users.

## Context

AISEC automated security assessment (scan 3aad5469, 2026-04-28) demonstrated the full exploit chain:

1. Inject attacker-controlled `.mcp.json` via `POST /api/agents/{name}/credentials/inject`
2. Drive the agent's bash to `nc -w3 redis 6379` and capture output
3. Redis responds to `PING` with `+PONG` — `requirepass` is not set (version 7.4.8, uptime 19 days)
4. `LRANGE` on a foreign queue reveals task records of **other users**:

```
agent:queue:ad_copy_generator_seller   LIST  (other user, source_user_id=1)
agent:queue:competitor-ad-swipefile-seller  LIST  (other user)
```

An attacker can **inject task records into another user's queue**; the receiving Claude Code agent executes them as that user.

**Root cause gap vs. prior fix:** UnderDefense issue #172 (Docker Image Allowlist) was fixed on 2026-04-23 and given an "A — Excellent" grade. That fix closed one vector but did not address the actual root cause: **Redis itself has no password**. Any legitimately deployed agent container still has direct unauthenticated access.

CVSS 9.9 (AV:N/AC:L/PR:L/UI:N/S:C/C:H/I:H/A:H), CWE-306.

## Acceptance Criteria

- [ ] Redis `requirepass` set via `${REDIS_PASSWORD}` env var in `docker-compose.yml`; `.env.example` documents the variable
- [ ] Redis ACL configured so platform services (`backend`, `mcp-server`) have scoped command access; agent containers have **no** Redis ACL entry
- [ ] Agent containers cannot reach `redis:6379` directly — network segment separation (platform-internal network vs agent network)
- [ ] Integration test asserts an agent-network container cannot authenticate to Redis without the platform password
- [ ] All existing Redis clients (`slot_service.py`, `backlog_service.py`, `event_bus.py`, `cleanup_service.py`, `credential_encryption.py`) use `REDIS_URL=redis://:${REDIS_PASSWORD}@redis:6379`

## Technical Notes

[...]

PR #1520, opened 2026-07-08
test: fix dev test-suite drift + harness bugs (no product changes)

vybe:
## Summary

A full-suite run on `dev` (Docker, local) surfaced **38 failures + 2 errors**. Every one was root-caused to a **stale test or a test-harness issue — zero product regressions**. This PR is **test-files-only (0 lines of production code)**.

Verified after fixes: **74 unit + 162 integration = 236 passed** across the touched files.

## Fixes (what drifted, why)

| Cluster | Fix | Root cause |
|---|---|---|
| `test_platform_prompt_unit` (5) | stub lambdas accept `**_kwargs` | production compose path now forwards `runtime=` (#1187) |
| `test_cb_probe_execution_close` (7) | add `get_setting_value` to `_CanaryTempDB` | canary temp-DB stub leaks onto module-level `from database import db`; under full-suite order it serviced `get_platform_default_model` and `AttributeError`'d |
| `test_fork_to_own` (2 errors) | defensive `list_all_agents_fast` backfill | leaked bare `services.docker_service` stub (same `sys.modules` cross-file leak class) |
| `test_files_guardrail_bypass`, `test_mcp_validator_endpoint` (2) | assert `"Disallowed credential file path"` | intentional message wording change (#1305) |
| `test_event_subscriptions` (1) | expect **403** not 400 | intentional #186 enumeration-safety collapse (commit `34ddf71a`) |
| `test_activities` (1) | derive valid triggers from canonical `_TRIGGER_BUCKETS` | hard-coded allow-set drifted behind `webhook` / channel triggers; now tied to source of truth |
| `test_whatsapp_integration` (4) | read pending-login Redis via backend's authed client | bare `redis-cli GET` fails `NOAUTH` (Redis ACL mandatory since #589) but exits 0, defeating the `returncode` guard. **Product code verified correct.** |

Clusters 1–3 pass in isolation and only failed under full-suite ordering (`sys.modules` leakage). A durable follow-up would harden the `tests/unit/conftest.py` sys.modules baseline restore — out of scope here.

## ⚠️ Not in this diff (environment, not code)

Running the full suite locally still shows a handful of failures/skips that are **not** fixed by this PR because they are environmental — a clean CI run is unaffected:

- **9 unit failures** (`bcrypt` 72-byte error ×3, `ModuleNotFoundError: PIL` ×6) — a **drifted local `tests/.venv`** (bcrypt 5.0.0, no Pillow). `requirements-test.txt` already pins `bcrypt>=4.2.0,<5` and `Pillow>=11.1.0`. **Resync your local venv:** `cd tests && .venv/bin/pip install -r requirements-test.txt`.
- **~60 infra-broken skips** ("Agent server not ready", "testfix container not found", 429s) — the shared test agent was never ready (`TEST_AGENT_NAME` unset, no `testfix` fixture). Needs a persistent test agent.
- **4 quota failures** (`test_dynamic_thinking_status` async ×3, one 429) — the real Claude subscription hit its usage cap during the 50-min run.
- **3 environmental** (`workspace_available`/`auto_switch` defaults, `agent_rename` name collision) — pass in a clean CI env / fresh DB; code defaults are correct.

## Test plan

- [x] `pytest` on all touched files (leak-order preserved for canary→cb_probe): 236 passed
- [x] Confirmed diff touches **no** `src/backend/` production code

### DIFF
--- a/tests/test_activities.py
+++ b/tests/test_activities.py
@@ -441,11 +441,12 @@ class TestTimelineForDashboard:
         assert_status(response, 200)
         activities = response.json().get("activities", [])
 
-        # Valid trigger types for Dashboard color coding.
-        # #1260: include "loop" (#740/#1150) and "voip" (#1056) — real trigger
-        # values added after this test was written.
-        valid_triggers = {"schedule", "agent", "manual", "user", "mcp", "public",
-                          "fan_out", "loop", "voip", None}
+        # Valid trigger types for Dashboard color coding. Derive from the canonical
+        # source of truth (db/schedules._TRIGGER_BUCKETS — the Dashboard bucket map)
+        # rather than a hand-maintained literal set, so this test doesn't drift every
+        # time a trigger/channel value is added (webhook, telegram/slack/whatsapp, etc.).
+        from db.schedules import _TRIGGER_BUCKETS
+        valid_triggers = set(_TRIGGER_BUCKETS) | {None}
 
         for activity in activities:
             triggered_by = activity.get("triggered_by")
--- a/tests/test_canary_invariants.py
+++ b/tests/test_canary_invariants.py
@@ -593,6 +593,14 @@ class _CanaryTempDB:
             ).first()
         return int(row[0]) if row else 0
 
+    def get_setting_value(self, key: str, default=None):
+        # No settings table in the B-01 temp DB; mirror the real facade's
+        # empty-store return so callers that leak onto this stub via the
+        # module-level `from database import db` reference (settings_service.
+        # get_setting <- task_execution_service.get_platform_default_model)
+        # get a clean fallback instead of an AttributeError.
+        return default
+
 
 def _reload_canary_with_temp_db(fake_redis, monkeypatch):
     """Evict + reimport the canary modules and install the controlled `database`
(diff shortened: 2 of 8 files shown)


dolho:
**Review: ✅ Approve — safe to merge**

Test-files-only, zero production lines (verified: diff touches only `tests/`). Each of the 40 failures is credibly root-caused to test drift or a harness bug, not a regression:

- `test_platform_prompt_unit` — stubs now accept `**_kwargs` to match the `runtime=` forwarding added in #1187. Correct.
- `test_event_subscriptions` 400→403 — matches the #186 enumeration-safety collapse (source-not-found folded into the not-permitted 403). Correct.
- `*_credential file path` message asserts — tracks the intentional #1305 wording.
- `test_activities` deriving valid triggers from `_TRIGGER_BUCKETS` — good move, ties the assertion to the source of truth instead of a literal that keeps drifting.
- `test_whatsapp_integration` reading pending-login through the backend's authed Redis client — real harness bug (bare `redis-cli` returns `NOAUTH` with exit 0 since #589, silently defeating the returncode guard). Product code untouched.

CI green across all 3 seeds × base/head. One note (already called out in the PR): clusters 1–3 are band-aided per-file against `sys.modules` cross-file leakage; the durable fix is a `conftest.py` baseline restore. Fine to defer, but worth a follow-up issue so this doesn't recur file-by-file.

dolho (approved):

PR #1524, opened 2026-07-08
refactor: make context counters model-specific (1M-context models) + re-sync model catalog (#1521)

### LINKED ISSUE #1521: refactor: make context counters model-specific (1M-context models) + re-sync available-model catalog
## Summary

Context-usage counters across the platform assume a flat **200,000-token** context window. Models now ship larger windows — Claude Opus 4.8/4.7 with a **1M-token** context (`[1m]` suffix), Gemini at 1M, Codex at 272K — so the usage bars and percentages are computed against the wrong denominator. On a 1M-context turn the counter reads ~5× too high (e.g. showing "90% used" when the turn is actually at ~18%).

The real per-model window *is* already reported by the runtime — `stream_parser.py` reads `contextWindow` from the result stream's `modelUsage` — but it gets overridden by a hardcoded `200000` fallback in ~15 places whenever it doesn't propagate. This refactor consolidates those scattered magic constants into **one model → context-window source of truth** and makes the fallback model-aware instead of a flat 200K. As part of the same pass, re-sync the **available-model catalog** (the picker preset list and the agent-server alias note have drifted — e.g. the Claude 5 family and correct default windows are missing/stale).

## Context

- The `200000` literal is duplicated across backend, the agent-server mirror, the scheduler, and the frontend. There is no single place that answers "what is model X's context window?", so each site guesses 200K.
- `docker/base-image/agent_server/state.py:100` (`_get_default_context_window`) already branches Gemini → 1M vs Claude → 200K, but ignores the `[1m]` 1M-context Claude variants and every other site re-hardcodes 200K independently.
- The `[1m]` alias suffix is understood at dispatch (`docker/base-image/agent_server/routers/chat.py` validates `opus[1m]`/`sonnet[1m]`/`haiku[1m]`) but the context-window accounting never learns that those turns have a 1M window.
- Model catalogs have drifted: the picker (`ModelSelector.vue` `PRESET_MODELS`, last synced 2026-06-06 / #1080) and the alias note in `agent_server/routers/chat.py` ("Sonnet 4.5 / Opus 4.5" aliases) are stale and maintained separately.

## Known fallback sites (grounding, not exhaustive)

Backend / scheduler:
- `src/backend/services/agent_client.py` (`context_window`/`or 200000`, ~3 sites)
- `src/backend/services/task_execution_service.py` (`metadata.get("context_window") or 200000`, ~3 sites)
- `src/backend/routers/chat.py` (`or 200000`, ~2 sites incl. the salvage path)
- `src/backend/services/agent_service/stats.py` (`contextMax` default 200000)
- `src/backend/db_models.py` (`total_context_max: int = 200000`, ~2 models)
- `src/scheduler/models.py` (`context_max: int = 200000`)

Agent-server (mirror — Invariant #5 parity):
- `docker/base-image/agent_server/config.py` (`DEFAULT_CONTEXT_WINDOW = 200000`, `GEMINI_CONTEXT_WINDOW = 1000000`)
- `docker/base-image/agent_server/state.py` (`_get_default_context_window`)
- `docker/base-image/agent_server/models.py` (`context_window: int = 200000`)

Frontend:
- `src/frontend/src/composables/useSessionActivity.js` (defaults 200000, ~3 sites)
- `src/frontend/src/components/AgentTile.vue` (`ctxStats?.contextMax || 200000`)
- `src/frontend/src/components/SessionPanel.vue` (`total_context_max || 200000`)
- percentage renderers in `SchedulesPanel.vue`, `TasksPanel.vue`, `ExecutionDetail.vue` (consume `context_max`, so they inherit whatever denominator upstream picks)

## Acceptance Criteria

- [ ] A single model → context-window catalog (one canonical map, consumed by backend, the agent-server mirror per Invariant #5, and the frontend) replaces the scattered `200000` literals; the `[1m]` 1M-context Claude variants, Gemini (1M), and Codex (272K) resolve to their real windows.
- [ ] The runtime-reported `contextWindow` (from `modelUsage` in `stream_parser.py`) remains the primary value; the fallback is derived from the catalog by model id, never a flat constant.
- [ ] A turn on a 1M-context model displays utilization against a 1M denominator end-to-end: session tab (`SessionPanel.vue`), agent tile (`AgentTile.vue`), execution/schedule/task rows (`SchedulesPanel.vue`, `TasksPanel.vue`, `ExecutionDetail.vue`).
- [ ] Unknown / free-text model ids fall back to a sane documented default (not a silently-wrong 200K) — e.g. the model's family default, else a conservative floor with a logged warning.
- [ ] The available-model catalog is re-synced against the canonical model overview: the `ModelSelector.vue` preset list and the agent-server alias note are current (add the Claude 5 family, drop retired presets, fix stale "Sonnet 4.5 / Opus 4.5" alias labels) and, ideally, derive from the same shared catalog so they can't drift again.
- [ ] No schema change required (the `context_max` / `total_context_max` columns stay; only the *values* written to them become model-correct). Confirm defaults in `db_models.py` / `src/scheduler/models.py` don't reintroduce a flat 200K when metadata is present.

## Technical Notes

- Invariant #5 (agent server mirrors the backend) applies — the catalog must live somewhere both sides can consume, or be vendored with a parity check like `credential_paths.py`.
- Prefer resolving the window from the model id at the point the metadata is finalized (`task_execution_service` / `agent_client`) so persisted rows carry the correct `context_max`; the frontend fallback should only ever be a last-resort display guard.
- Keep the catalog data-only and easy to bump (a comment pointing at the canonical model overview, as `ModelSelector.vue` already does) so the recurring "update the model list" chore is a one-file edit.
- Out of scope: changing how tokens are *counted* (the cumulative-vs-per-call accounting in `stream_parser.py` is correct and documented) — this is only about the *denominator* (max window) and the model list.

vybe:
## Summary
- The `% context used` gauge assumed a flat **200K** window, reading up to ~5× too high on 1M-context models (Opus 4.8/4.7, Sonnet 5, Fable 5 = 1M; Gemini = 1M; Codex = 272K). This consolidates ~15 scattered `200000` literals + the 5 disagreeing agent-server resolvers into **one model→context-window catalog**.
- The runtime-reported `modelUsage.contextWindow` stays **primary** (only it knows the effective plan/auth/beta window); the catalog is the **fallback** — bare Claude → **200K safe floor**, `[1m]` → 1M, Gemini → 1M, Codex → 272K, unknown → 200K + logged warning.
- **Why the bare-Claude floor is 200K, not 1M:** the effective window is auth/plan-dependent (a Pro-without-credits Opus turn is genuinely 200K). On the fallback path we have no tier info, so failing "too full" is safe; a fixed 1M fallback would hide an imminent 200K compaction wall. Real 1M is delivered by the primary value.

## Changes
- **New** `src/backend/services/model_context.py` — pure-stdlib catalog (`resolve_context_window` + `DEFAULT_CONTEXT_WINDOW`), **vendored byte-identically** into `docker/base-image/agent_server/model_context.py` (Invariant #5, parity-tested).
- **Critical fix**: `ClaudeCodeRuntime.get_context_window` was **dead code** — the real Claude fallback was the static `ExecutionMetadata.context_window = 200000` default. Now the catalog **seeds** `metadata.context_window` at construction in the Claude + headless paths (the stream value overrides it), guarded by a regression test.
- Unified 5 resolvers (`state.py`, `claude_code`, `gemini_runtime`, `codex_runtime`, ABC-now-non-abstract); fixed `state.py` init-ordering so a `[1m]` agent reports 1M pre-first-turn.
- Backend downstream: 13 `or 200000` → `or DEFAULT_CONTEXT_WINDOW`.
- Re-synced `ModelSelector.vue` presets (added Fable 5, Sonnet 5) + agent-server alias note.
- Docs: `architecture.md` (Context-window catalog note), `model-selection.md`, `learnings.md`.
- **No schema change** — columns/DDL unchanged; only written *values* become model-correct.

## Test Plan
- [x] New: `pytest tests/unit/test_model_context_catalog.py tests/unit/test_model_context_parity.py` — 35 pass (resolution table incl. `[1m]`/unknown/None, byte-parity, Claude-metadata seed regression, AgentState init-ordering)
- [x] Full unit suite: 3611 passed (4 pre-existing unrelated flakes in `fork_to_own`/`capacity_manager` — pass in isolation, don't touch this code)
- [x] No secrets in diff; scope clean; `/review` → 0 critical

Fixes #1521


dolho:
**Review (direct).** Consolidates scattered `200000` literals + disagreeing per-runtime resolvers into one `model_context.resolve_context_window` catalog for the % context denominator.

Strengths: total function (bounded output, never raises); the **safe-floor invariant** is the right call — bare Claude ids resolve to the guaranteed 200K floor (over-report usage = safe) rather than the 1M ceiling (which would hide a 200K compaction wall), with `[1m]` as the explicit 1M opt-in and the runtime-reported value as the authority. Byte-identical vendored copy (Invariant #5) guarded by `test_model_context_parity.py`.

Notes (non-blocking):
- **Documented tradeoff**: a bare `sonnet-5`/`fable-5` on the *fallback* path resolves to 200K even though they're natively 1M — safe (fuller bar), and the docstring calls it out (decision A). Just flagging it's intentional, not a bug.
- `_FAMILY_PREFIX_WINDOWS` is `startswith`-ordered most-specific-first; fine today, but a future `claude-…` id that should differ from the 200K floor would need the runtime value (which it has). OK.

Merge conflict in `tests/registry.json` (both sides appended test entries) resolved — kept all three entries, JSON validated. CI re-running post-rebase. ✅ once green.

dolho (approved):

Issue #1552, opened 2026-07-09
refactor(roster): derive client message_count/last_active from public_chat at read time instead of a parallel counter


webmixgamer:
## Summary

Follow-up to #1533. That issue was fixed **tactically** by reviving the per-channel counter (`chat_links.message_count`) via a new `record_inbound()` upsert. This issue captures the reframe that the strategy review recommended instead, so the stopgap is an explicit decision rather than the accidental permanent design.

## Context

The same router that writes `chat_links` already persists **every** channel message to `public_chat`:

- `message_router.py` → `db.get_or_create_public_chat_session(agent_name, session_identifier, channel)`
- `message_router.py` → `db.add_public_chat_message(...)` for the user turn and the assistant turn

So a real, live, per-session message record already exists — **and it is backfilled**, whereas the `#1533` counter starts at deploy.

The proposal: drop the counter and derive the roster's two columns at read time in `list_clients_for_agent`:

- `message_count` = `COUNT(*) FROM public_chat_messages WHERE role='user'` (note: `public_chat_sessions.message_count` counts user **+** assistant turns, so it is the wrong source)
- `last_active`   = `MAX(timestamp)`

Then `chat_links.message_count` can be dropped outright, along with the write path added in #1533.

## Why this wasn't done in #1533

- It rewrites the roster spine (both `list_clients_for_agent` implementations + a cross-table aggregation + a column drop), well beyond that issue's `complexity-low` bug scope.
- The join is **not** a clean equijoin. Telegram's `get_session_identifier` is `{bot_id}:{sender_id}:{chat_id}` — keyed on `bot_id`, not the `binding_id` that `chat_links` uses. WhatsApp's is `{binding_id}:{sender_id}`, which is clean.
- **It does not fix Slack "for free."** Slack's `get_session_identifier` is thread-scoped for channel messages and deliberately drops `sender_id` (#903), so per-user attribution is impossible for channel threads from that table. It would work for Slack DMs only.

## The regret this avoids

Three artifacts now track "how much has this channel user talked": `chat_links.message_count` (user-only, from-deploy, Telegram/WhatsApp), `public_chat_sessions.message_count` (user + assistant, full history, all channels), and raw `public_chat_messages`. Identity stays split from activity. Whoever picks up per-client controls or client analytics next will hit the same split.

## Acceptance criteria

- [ ] Decide: derive-at-read vs keep the counter (this issue may legitimately close as "won't do")
[...]

PR #1554, opened 2026-07-09
fix(channels): count inbound client messages on the Sharing-tab roster (#1533)

### LINKED ISSUE #1533: bug: Sharing-tab client roster shows 0 messages for every client — message_count is never incremented
## Summary

The client roster on the agent Sharing tab displays **0 messages for every external client**, including clearly active users (correct `last_active`, real conversations). The `message_count` column on the channel chat-link rows is initialized to `0` at creation and never incremented afterward — the increment plumbing exists but has no callers.

## Context

Observed on the Sharing-tab roster (aggregated by `src/backend/services/client_roster_service.py` across Telegram + WhatsApp). Clients sort correctly by `last_active` (it is touched on inbound traffic via `get_or_create_chat_link`), yet all show zero messages — so the roster reads as "active user, 0 messages", which is self-contradictory.

**Root cause:** dead increment path.
- `db/telegram_channels.py::increment_message_count` and `db/whatsapp_channels.py::increment_message_count` exist, and `database.py` exposes them as `increment_telegram_message_count` / `increment_whatsapp_message_count` — but a repo-wide search finds **zero call sites** outside `database.py`.
- Chat links are created with `message_count=0` (`db/telegram_channels.py`, `db/whatsapp_channels.py`) and nothing ever updates the column.
- The roster (`list_clients_for_agent` in both DB modules → `client_roster_service.py`) faithfully reports the stored value: always 0.

## Acceptance Criteria

- [ ] Inbound message handling increments `message_count` on the matching chat link for **both** Telegram and WhatsApp (the two channels the roster aggregates today)
- [ ] Counting semantics defined and documented in the code: count inbound client messages (one per user message; assistant replies not counted)
- [ ] Sharing-tab roster shows non-zero, growing counts for active clients after the fix
- [ ] Unit tests cover the increment on the inbound path for both channels (and that a failed increment never blocks message processing — counter is best-effort)
- [ ] Note in the fix: historical counts are not backfilled (links start counting from deploy); roster UI copy should not imply lifetime totals

## Technical Notes

- Wire the increment where the chat link is resolved on the inbound path (adjacent to the existing `get_or_create_chat_link` / `last_active` touch) in `adapters/telegram_adapter.py` and `adapters/whatsapp_adapter.py` — or fold the increment into the same touch UPDATE to avoid a second write per message.
- Facade methods already exist: `database.py::increment_telegram_message_count` / `increment_whatsapp_message_count`.
- Roster read path needs no change: `client_roster_service.py` already surfaces `message_count`.

webmixgamer:
## Summary

The agent Sharing tab's client roster showed **`0` messages for every external client**. Investigating it turned up a wider fault than the issue describes — one dead call site causing three symptoms:

1. **`message_count` was always `0`** — the reported bug.
2. **`last_active` was also stale.** It was the `/login` timestamp, not last activity.
3. **Clients who never ran `/login` never appeared in the roster at all.**

**Root cause:** `get_or_create_chat_link` *and* `increment_message_count` (Telegram + WhatsApp) had **zero callers** outside the `database.py` facade. The shared inbound path (`message_router._handle_message_inner`) writes `public_chat_sessions` — a *different* table — so `*_chat_links` rows were only ever created by `set_*_verified_email`. The read path faithfully echoed a column nobody wrote, and `client_roster_service`'s own docstring asserted the data existed.

> ⚠️ **Correction to the issue body.** Issue #1533's Context says *"`last_active` … is touched on inbound traffic via `get_or_create_chat_link`"*, and its Technical Notes say to wire the increment "adjacent to the existing `get_or_create_chat_link` / `last_active` touch". That call site **does not exist**. Implementing the issue literally would have shipped an increment that never fires for unverified users and left `last_active` wrong. The scope widening here was reviewed and approved before implementation.

## Changes

- **`adapters/base.py`** — new default-no-op `record_inbound_activity(message, agent_name)` hook on the `ChannelAdapter` ABC, alongside the existing `enrich_message` / `on_response_sent` defaults. Slack and VoIP inherit the no-op (Invariant #9).
- **`adapters/message_router.py`** — calls it once per delivered DM at **step 5c**, *after* `_enforce_access_policy` (so an unauthenticated stranger who messages the bot cannot create unbounded chat-link rows) and **skipped for groups** (a chat link is DM-keyed; counting group traffic would list members who never DM'd the agent). Wrapped in `try/except` — a counter write must never block message processing.
- **`adapters/telegram_adapter.py`, `adapters/whatsapp_adapter.py`** — override the hook.
- **`db/telegram_channels.py`, `db/whatsapp_channels.py`** — new `record_inbound()`: one atomic `INSERT … ON CONFLICT DO UPDATE` replacing `get_or_create` + `increment`. This removes a cross-worker SELECT-then-INSERT race (`--workers 2`; the old `get_or_create_chat_link` had no `IntegrityError` catch), halves the writes per message, and refreshes a stale display name via `COALESCE(excluded, existing)`. Because `increment_message_count` was `last_active`'s only live writer, one call fixes the count and the timestamp together.
- **Dead-code removal** — the four now-unused methods, their facade delegations, and the `_row_to_chat_link` helper only they called. Verified zero callers across `src/`, `tests/`, and the private `trinity-enterprise` submodule.
- **`SharingPanel.vue`** — the `Messages` header now states counts are not backfilled.
- **Docs** — `architecture.md`, `telegram-integration.md` (which documented the deleted methods and claimed `message_count` was "Incremented per message"), `whatsapp-integration.md`, plus a `learnings.md` entry.

No schema change, so no migration on either track.

## Test Plan

- [x] **Dual-backend DB tests** through the real `db_backend` harness — roster read-back `0→1→2`, `last_active` advances, username backfills, `/login` interplay, per-client isolation. **Passes on SQLite *and* real PostgreSQL** (`on_conflict_do_update` portability is the main technical risk; SQLite alone would not have proven it).
- [x] **Real router pipeline** (`_handle_message_inner`) — DM counts; group, access-denied do not; a raising hook still delivers the reply.
- [x] **Adapter override bodies executed for real**, and **mutation-verified**: typo the metadata key to `sender_username` and the suite goes red. Without this test, that typo ships green.
- [x] **Facade delegation** guarded — a wholesale `database = MagicMock()` stub cannot see a missing/renamed facade method (`learnings.md`, 2026-07-06).
- [x] Full `tests/unit`: **3648 passed**. Channel/router suites vs a live backend: **97 passed**. The 5 failures reproduce identically on a pristine `origin/dev` worktree (macOS-Docker `pgroup` ×2, a `Europe/Kiev` tz case, 2 live-API anon-session tests). **Zero regressions.**
- [x] **Verified against a live instance.** Real Telegram `Update` payloads POSTed to the real webhook endpoint with a valid secret token, through the real transport → `parse_message` → router → hook → adapter → production SQLite:
  - `require_email=true`, unverified sender → `Access denied` at the router, **no row created** (proves the hook sits after the gate)
  - DM #1 → `('alice', 1, …032Z)`; DM #2 → `('alice', 2, …822Z)` — one row, no duplicate, `last_active` advanced
  - Group message (valid `@mention`) → reaches the router (`START channel=-100123456`, `DONE`) and is **not** counted
  - `GET /api/agents/{name}/clients` → `[{"identity":"@alice","message_count":2,…}]`

## Notes

- Historical counts are **not** backfilled — links start counting at deploy. The roster header says so.
- Access-denied and pending clients stay at `0`; that is deliberate ("delivered turns, not gate-rejected attempts") and documented in `record_inbound()`.
- Follow-up **#1552** records the read-time-derivation reframe (deriving the roster from `public_chat_messages`, which would be backfilled and cover Slack DMs) that this tactical fix deliberately defers.

Fixes #1533


vybe (approved):
Validated via /validate-pr: Fixes #1533. `record_inbound()` atomic upsert removes a cross-worker SELECT-then-INSERT race; regression test `test_1533_client_message_count.py` present. All required checks green (incl. schema-parity, pg-migrations). Security scan clean. Approving.

PR #1556, opened 2026-07-09
fix(avatar): prevent thinking-token truncation dropping the color-scheme block

vybe:
## Problem

Agent avatars were coming out wildly inconsistent and off-palette — not following the color scheme.

**Root cause:** Avatar consistency lives entirely in a fixed technical block that the refinement model (`GEMINI_TEXT_MODEL`) is asked to append to every prompt — background `#1a1f2e`/`#111827`, indigo rim `#6366f1`, head-and-shoulders framing, 85mm lens, 5600K key light. The refinement call in `_call_gemini_text` capped `maxOutputTokens` at **512**, but since #1130 the default model is `gemini-3.5-flash`, a **thinking** model whose reasoning phase draws from the output budget. Reasoning consumed **~491 of 512 tokens**, truncating the refined prompt (`finishReason=MAX_TOKENS`) to a ~90-char fragment that dropped the entire technical block. The image model then received only a bare subject description → inconsistent, off-scheme avatars.

This was a silent regression from #1130: the previous default `gemini-2.0-flash` was non-thinking, so all 512 tokens went to visible output and the block survived.

## Fix (`src/backend/services/image_generation_service.py`)

- **Disable thinking** (`thinkingConfig.thinkingBudget=0`) — prompt refinement is a deterministic rewrite that needs no chain-of-thought.
- **Raise `maxOutputTokens` 512 → 4096** — headroom so the block survives even when thinking stays on.
- **Retry once without `thinkingConfig` on HTTP 400** — a thinking-mandatory model (e.g. `gemini-2.5-pro`, in #1130's model list) rejects `budget=0` with a 400. Since `refine_prompt` silently falls back to the raw prompt on any error (the exact trap that hid this bug), an unhandled 400 would re-break avatars on that model.

## Verification

- **End-to-end against the live model:** all 6 technical-block tokens now survive refinement (`#1a1f2e`, `#111827`, `#6366f1`, `head-and-shoulders`, `85mm`, `5600K`) — was **0/6** before.
- **Unit tests:** 3 new regression tests + full suite green (`43 passed`).

```
tests/unit/test_image_generation_service.py::TestCallGeminiText::test_disables_thinking_and_has_output_headroom PASSED
tests/unit/test_image_generation_service.py::TestCallGeminiText::test_retries_without_thinking_config_on_400 PASSED
tests/unit/test_image_generation_service.py::TestCallGeminiText::test_persistent_400_still_raises PASSED
```

## Scope / notes

- Code-only change (no schema) — no dual migration needed.
- Existing already-generated avatars are cached image files; they'll benefit only on regeneration (`POST /api/agents/avatars/generate-defaults` or per-agent Regenerate).
- **Follow-up (not in this PR):** consistency still relies on an LLM copying a verbatim block, which is inherently brittle (the `social` use case was also observed leaking markdown). A more durable design would append the fixed technical block in code after refinement. Happy to do that separately.

### DIFF
--- a/src/backend/services/image_generation_service.py
+++ b/src/backend/services/image_generation_service.py
@@ -252,28 +252,57 @@ class ImageGenerationService:
         """
         url = f"{GEMINI_API_BASE}/{GEMINI_TEXT_MODEL}:generateContent"
 
-        payload = {
-            "system_instruction": {
-                "parts": [{"text": system_prompt}],
-            },
-            "contents": [
-                {
-                    "role": "user",
-                    "parts": [{"text": user_message}],
-                }
-            ],
-            "generationConfig": {
+        def _build_payload(disable_thinking: bool) -> dict:
+            gen_config = {
                 "temperature": 0.7,
-                "maxOutputTokens": 512,
-            },
-        }
+                # #1130 follow-up: GEMINI_TEXT_MODEL defaults to a *thinking* model
+                # (gemini-3.5-flash). Its reasoning phase draws from the output
+                # budget, so a tight cap (was 512) let thinking consume ~491 tokens
+                # and truncate the refined prompt (finishReason=MAX_TOKENS) to a
+                # ~90-char fragment — dropping the AVATAR fixed technical block
+                # (background #1a1f2e/#111827, indigo rim #6366f1, head-and-shoulders
+                # framing, 85mm lens, 5600K key). That block is the sole source of
+                # avatar color-scheme + consistency, so avatars came out wildly
+                # different and off-palette. Generous headroom keeps the full block
+                # intact even when thinking stays on.
+                "maxOutputTokens": 4096,
+            }
+            if disable_thinking:
+                # Prompt refinement is a deterministic rewrite that needs no
+                # chain-of-thought. Disabling thinking makes it fast/cheap and
+                # immune to the budget squeeze above. Not every model allows a zero
+                # budget (e.g. gemini-2.5-pro is thinking-mandatory → HTTP 400), so
+                # the caller retries without this field on a 400.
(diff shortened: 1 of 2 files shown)


dolho (commented):
### Review — LGTM, one non-blocking suggestion

Root-cause writeup is excellent and the fix is correct: disabling thinking (`thinkingBudget=0`) + raising the cap to 4096 addresses both the squeeze and the headroom, and the 400-retry gracefully covers thinking-mandatory models. Nice touch that the same retry also makes it robust for a *non*-thinking model that rejects an unknown `thinkingConfig` field. Test coverage is solid — happy path, 400-retry, and persistent-400-still-raises all guarded.

**Non-blocking suggestion — the silent-truncation class isn't fully closed.** `_call_gemini_text` still returns `candidates[0].content.parts[0].text` without inspecting `finishReason`. The 4096 headroom makes truncation *unlikely*, but on the retry path (thinking **on**, for a thinking-mandatory model) reasoning still draws from the output budget — if it ever eats 4096, you get the exact original failure again: a truncated fragment that drops the AVATAR technical block, returned silently and passed to the image model.

Consider making a future recurrence *loud* rather than silent — e.g. after parsing:

```python
cand = data["candidates"][0]
if cand.get("finishReason") == "MAX_TOKENS":
    raise RuntimeError("Gemini text refinement truncated (MAX_TOKENS)")
```

`refine_prompt` already falls back to the raw prompt on a raise, so this degrades no worse than today but turns a silent off-palette avatar into a caught, logged failure. (Alternatively, assert the refined text still contains a sentinel from the fixed block.) Not a merge blocker — the headroom + thinking-off almost certainly suffices for the default model.

dolho (approved):

Issue #1562, opened 2026-07-10
Prettify Agent Permissions Matrix: bigger labels + no grid jump on grant/revoke


dolho:
## Problem

The **Agent Permissions Matrix** (`AgentPermissionsMatrix.vue`) has two UX issues:

1. **Hard to read.** The whole grid is `text-xs` with `w-8 h-8` (32px) cells; agent row/column labels are tiny and easy to miss.
2. **Grid jumps on every update.** The success/error toast is rendered inline between the toolbar and the grid (`v-if="message"`). On each grant/revoke it appears (pushing the grid down), then auto-dismisses after 3.5s (grid snaps back up) — a distracting layout jump every time you toggle a cell.

## Scope

Presentation-only polish of `AgentPermissionsMatrix.vue`. **No** change to the grant/revoke logic or the store/backend endpoints.

## Acceptance criteria

- [ ] Grid and agent labels are comfortably legible (larger base font + cells).
- [ ] Toggling a cell (grant/revoke) does **not** shift the grid vertically — the toast occupies a reserved slot.
- [ ] Column labels stay readable at their rotated orientation.
- [ ] No regression to grant/revoke, bulk row/col actions, filter, or the selected-pair rail.

vybe:
Fixed by #1563 (merged to `dev`). Promoting to `status-in-dev` — the PR referenced the issue only via its branch name, so auto-promotion didn't fire.

PR #1563, opened 2026-07-10
fix(ui): prettify Agent Permissions Matrix — bigger labels, no grid jump

dolho:
## What

Presentation-only polish of the **Agent Permissions Matrix** (`AgentPermissionsMatrix.vue`).

**Readability**
- Base font `text-xs` → `text-sm`; cells `w-8 h-8` → `w-11 h-11`; granted checkmark `text-base`.
- Row/column agent labels get `font-medium`; corner axis hints `10px` → `text-xs`; column-label headroom `8rem` → `10rem`.

**No grid jump on update**
- The success/error toast was rendered inline between the toolbar and the grid (`v-if="message"`), so it pushed the grid down on every grant/revoke and snapped it back on the 3.5s auto-dismiss. It now lives in a reserved `min-h-[2.5rem]` slot → the grid no longer reflows.

## Scope

One file, markup/CSS only. **No** change to grant/revoke logic, the store methods, or the backend/enterprise endpoints.

## Verification

Grant/revoke exercised end-to-end against the live stack via the exact endpoints the store calls:

| Step | Result |
|------|--------|
| `GET /api/enterprise/permissions-matrix` | 200, agents + edges |
| `POST /api/agents/asdfsdf/permissions/cornelius` | 200 → re-read shows edge `(asdfsdf, cornelius)` |
| `DELETE /api/agents/asdfsdf/permissions/cornelius` | 200 → re-read shows `[]` |

Direction stays one-way (reverse edge never added). Component compiles clean (vite HMR, no error).

_Caveat: the anti-jump fix is verified by reasoning + reserved-height layout, not a recorded browser session; the `ui` label runs the Playwright e2e suite here._

Related to #1562

### DIFF
--- a/src/frontend/src/components/AgentPermissionsMatrix.vue
+++ b/src/frontend/src/components/AgentPermissionsMatrix.vue
@@ -60,13 +60,15 @@
         </span>
       </div>
 
-      <!-- Toast -->
-      <div v-if="message" :class="[
-        'mb-3 text-sm rounded-md px-3 py-2',
-        message.type === 'error'
-          ? 'bg-status-danger-50 dark:bg-status-danger-900/30 text-status-danger-700 dark:text-status-danger-300'
-          : 'bg-status-success-50 dark:bg-status-success-900/30 text-status-success-700 dark:text-status-success-300'
-      ]">{{ message.text }}</div>
+      <!-- Toast: reserved fixed-height slot so appear/disappear never shifts the grid -->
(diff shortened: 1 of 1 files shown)


vybe (approved):
/validate-pr ✅ — UI-only polish of AgentPermissionsMatrix (+17/-15, one Vue file). Required checks green (cancelled e2e was a superseded run). No secrets.

Issue #1561, opened 2026-07-09
bug(sync-health): `list_git_enabled_agents()` does not exclude soft-deleted agents — the 60s poller hammers dead agents and poisons their circuit breakers


webmixgamer:
## Summary

`SyncHealthService` polls every git-sync-enabled agent once a minute via `db.list_git_enabled_agents()` (`src/backend/services/sync_health_service.py:91`). That accessor (`src/backend/db/schedules.py:3103`) selects from `agent_git_config` filtered **only** on `sync_enabled == 1` — it does **not** join `agent_ownership` and does **not** exclude `deleted_at IS NOT NULL`.

Soft delete preserves `agent_git_config` rows by design, so the poller keeps calling agents whose container was removed — indefinitely. Each attempt raises `httpx.ConnectError`, which is exactly the exception class that increments the transport circuit breaker.

## Context

The sibling accessor was already hardened for this. `list_all_enabled_schedules()` (`src/backend/db/schedules.py:437`) carries a docstring stating it filters `agent_ownership.deleted_at` *"otherwise the scheduler fires"* for soft-deleted agents. The git-sync poller was simply missed by that pass.

Observed on a dev instance after an agent was soft-deleted at `11:42:52` (container removed):

```
Error getting git status for <name>: All connection attempts failed     # once a minute, forever
11:45:14  Circuit OPENED for agent <name> after 3 failures
12:18:19  Circuit DORMANT for agent <name> after 10 consecutive open-probe failures — switching to 3600s cooldown
12:18:19  [CB] circuit_breaker_dormant operator-queue entry emitted for <name>
```

Three consequences, all real:

1. **The breaker for a deleted agent is driven to `DORMANT`**, which then poisons any future agent created under the same name — see the companion issue #1560. The arithmetic corroborates the poller as the driver: with `CIRCUIT_FAILURE_THRESHOLD = 3` and cooldowns of 30 + 60 + 120 + 240 + 300×6 s ≈ 36 min against a 60s poll interval, `11:42 + ~36 min = 12:18`, matching the `DORMANT` log to the minute.
2. **A bogus operator-queue alert** (`circuit_breaker_dormant`) is raised naming an agent that no longer exists, and it sits unresolved in the Operations queue.
3. **Unbounded error churn** — one failed HTTP call per deleted agent per minute, forever, plus the operator-queue sync loop retrying delivery of that alert every 5s once any container reappears under the name.

[...]

Issue #1560, opened 2026-07-09
bug(lifecycle): stale transport circuit-breaker Redis key survives agent delete/recreate — a fresh healthy agent fast-fails as "unhealthy"


webmixgamer:
## Summary

The transport circuit breaker lives in Redis under `agent:circuit:{agent_name}` — keyed by **name**, not by container identity. Nothing clears that key when an agent is deleted, recreated, or renamed. An agent created with a previously-used name therefore **inherits the dead agent's breaker state**, and every execution fast-fails with

```
Agent circuit breaker open — agent is unhealthy
```

(`src/backend/services/task_execution_service.py:1016`) — while the new agent is perfectly healthy and is never contacted. The message is actively misleading: Trinity reports the agent unhealthy without having made a single call to it.

Worst case the fresh agent is unusable for **up to one hour**. Once the old breaker reaches `DORMANT` (after `CIRCUIT_DORMANT_AFTER_OPEN_PROBES = 10` failed probes, `src/backend/services/agent_client.py:76`), its cooldown is `CIRCUIT_DORMANT_COOLDOWN_SECONDS = 3600.0` (`:83`), and it only closes when that elapses and a half-open probe succeeds.

## Context

The asymmetry is the tell. The delete path already cleans up the *sibling* per-agent Redis state:

- `src/backend/routers/agents.py:510` → `heartbeat_service.clear_heartbeat(agent_name)`
- `src/backend/routers/agent_rename.py:147` → same, on rename

The circuit-breaker key is simply forgotten. `reset_circuit()` already exists (`src/backend/services/agent_client.py:541`) but is only reachable from the two admin reset endpoints and the autonomy toggle — never from the agent lifecycle.

Reproduced end-to-end from backend logs on a dev instance:

| Time (UTC) | Event |
|---|---|
| `11:42:52` | agent soft-deleted; container removed |
| `11:45:14` | `Circuit OPENED for agent <name> after 3 failures` |
| `12:18:19` | `Circuit DORMANT … after 10 consecutive open-probe failures — switching to 3600s cooldown` |
| `17:15:51` | agent **recreated** under the same name — fresh, healthy container |
| `17:16:03` | first user message → `[TaskExecService] CB open, fast-failing execution …` → `FAILED: Agent circuit breaker open — agent is unhealthy` |
| `17:18:27` | dormant cooldown elapses; half-open probe hits the healthy agent; circuit **closes** |
| `17:28:31` | next user message → **success** |

[...]

PR #1565, opened 2026-07-10
fix(sync-health): exclude soft-deleted agents from list_git_enabled_agents (#1561)

dolho:
## Problem

`db.list_git_enabled_agents()` selected from `agent_git_config` on `sync_enabled == 1` alone — no `agent_ownership` join, no `deleted_at` filter. Soft delete preserves the `agent_git_config` row, so the 60s `SyncHealthService` poller kept HTTP-polling removed containers **forever**. Each `httpx.ConnectError` is exactly the class that increments the transport circuit breaker → after ~3 min it OPENs, after ~36 min it goes **DORMANT** and emits a bogus `circuit_breaker_dormant` operator-queue alert for an agent that no longer exists. `GET /api/fleet/sync-audit` (same accessor) also listed dead agents.

## Fix

- **`list_git_enabled_agents()`**: join `agent_ownership`, filter `deleted_at IS NULL` — mirrors the #834 hardening on `list_all_enabled_schedules()`. One change fixes both consumers (poller + audit endpoint).
- **`git_service.get_git_status`**: bare `print()` → `logger.warning` (the issue's "minor, same code path" — those failures were invisible to structured logging).

## Audit sweep (acceptance criterion)

`list_git_enabled_agents` was the **sole** background-loop accessor missing the `deleted_at` filter:

| Loop | Enumeration | Soft-deleted excluded? |
|------|-------------|------------------------|
| sync_health (60s) | `list_git_enabled_agents` (DB) | **was the bug — fixed here** |
| monitoring (30s) | `list_all_agents_fast()` (Docker) | ✓ removed container not listed |
| operator_queue (5s) | `list_all_agents_fast()` + running (Docker) | ✓ |
| scheduler | `list_all_enabled_schedules` (DB) | ✓ already hardened (#834) |
| capacity (60s) | `list_agents_with_queued` (DB) | ✓ DB-only backlog drain, no agent HTTP |
| heartbeat (5s) | Redis `seen`, cleared on delete | ✓ |

Every other per-agent HTTP poller enumerates from **Docker**, so removed containers are inherently absent; the remaining DB-driven loops either already filter or make no agent HTTP call. No other offender found.

## Tests

`tests/unit/test_sync_health_service.py::TestSoftDeletedExcluded`:
- accessor excludes a soft-deleted agent (live agent still listed);
- one `_poll_cycle()` polls only the live agent — **zero HTTP to the soft-deleted one**, no `sync_state` row, no operator-queue entry.

Full `test_sync_health_service.py` (9) + `test_fleet_sync_audit.py` / `test_73_sync_health_bulk.py` (13) green locally.

## Scope note

This removes the **source** of breaker poisoning. The **inheritance** side — a stale `agent:circuit:{name}` key surviving delete/recreate — is the companion defect **#1560** and intentionally out of scope here.

Related to #1561

### DIFF
--- a/src/backend/db/schedules.py
+++ b/src/backend/db/schedules.py
@@ -3101,10 +3101,30 @@ class ScheduleOperations:
             return result.rowcount > 0
 
     def list_git_enabled_agents(self) -> List[AgentGitConfig]:
-        """List all agents with git sync enabled."""
+        """List all agents with git sync enabled.
+
+        Joins `agent_ownership` and excludes soft-deleted agents (#1561).
+        `agent_git_config` rows survive soft delete by design, so without this
+        filter the 60s `SyncHealthService` poller (and `GET /api/fleet/sync-audit`)
+        keeps hitting removed containers forever — each `httpx.ConnectError`
+        poisons the transport circuit breaker and eventually drives it to
+        DORMANT + a bogus `circuit_breaker_dormant` alert for an agent that no
+        longer exists. Mirrors the #834 hardening on `list_all_enabled_schedules()`.
+        """
         stmt = (
             select(agent_git_config)
-            .where(agent_git_config.c.sync_enabled == 1)
+            .select_from(
+                agent_git_config.join(
+                    agent_ownership,
+                    agent_ownership.c.agent_name == agent_git_config.c.agent_name,
+                )
+            )
+            .where(
+                and_(
+                    agent_git_config.c.sync_enabled == 1,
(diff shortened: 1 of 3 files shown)


vybe (approved):
/validate-pr ✅ — adds `agent_ownership.deleted_at IS NULL` to `list_git_enabled_agents()` so the sync-health poller stops chasing soft-deleted agents; covered by a soft-delete exclusion test. Also upgrades a bare print() to structured logging. Required checks green.

Issue #1559, opened 2026-07-09
bug(recovery): soft-delete recovery dead-ends — `recover` tells you to `start`, but `start` 404s when the container is gone


webmixgamer:
## Summary

Admin soft-delete recovery (#834 Phase 1c) can restore the `agent_ownership` row but **cannot bring the agent back**. `recover_agent` explicitly defers container creation to the operator, and the endpoint it defers to refuses to run without an existing container — which never exists after a delete.

## Evidence

`src/backend/routers/admin_recovery.py::recover_agent`:

```python
"""Clear `deleted_at` on the agent_ownership row. Audit-logged.
...
The container is NOT started automatically — operator does that explicitly.
"""
```

`src/backend/services/agent_service/lifecycle.py::start_agent_internal`:

```python
container = get_agent_container(agent_name)
if not container:
    raise HTTPException(status_code=404, detail="Agent not found")
```

`DELETE /api/agents/{name}` removes the container (the workspace volume survives). So after any soft delete, `POST /api/agents/{name}/start` always 404s, and there is no other endpoint that creates a container for an existing agent (`recreate_container_with_updated_config` requires an `old_container` to copy config from).

`create_agent_internal` cannot fill the gap either — it 409s on a reserved name, and `is_agent_name_reserved` deliberately matches soft-deleted rows (`crud.py:184-189`).

Net effect: a recovered agent is visible in the API and UI, holds all its child rows, and can never be started.

## Reproduction

1. `DELETE /api/agents/<name>` (soft delete; container removed, `agent-<name>-workspace` volume retained)
2. `POST /api/admin/soft-deleted/agents/<name>/recover` → `200`, `deleted_at` cleared
3. `POST /api/agents/<name>/start` → **404 `Agent not found`**

## Manual workaround (what recovery should be doing)

Deleting the `agent_ownership` row and re-creating the agent with the same name and template restores it completely: `crud.py:805` calls `volume_get` before `volume_create`, so the existing `agent-<name>-workspace` is reattached untouched, and `startup.sh` skips the clone because `.git` already exists on the volume (`"Repository already exists on persistent volume - skipping clone"`). Child rows key off the `agent_name` string, so Telegram bindings, chat sessions, and executions reattach on their own. This proves the data model supports a clean rebuild — only the API surface is missing.

## Acceptance criteria

[...]

PR #1567, opened 2026-07-10
fix(recovery): rebuild missing container on start for recovered agents (#1559)

dolho:
## Problem

Admin soft-delete recovery (#834 Phase 1c) could restore the `agent_ownership` row but **never bring the agent back online**. `recover_agent` deferred container creation to the operator; the only endpoint it deferred to — `POST /api/agents/{name}/start` → `start_agent_internal` — 404s when the container is gone (`get_agent_container` → None). And `recreate_container_with_updated_config` needs an `old_container` to copy config from, which never exists after a delete. Net: a recovered agent was visible, held all its child rows, and could never start.

## Fix

**`start` auto-recreates a missing container for a live agent** (the AC's option 2):

- `start_agent_internal`: when there's no container **but a live `agent_ownership` row exists** (recovered agent), rebuild it instead of 404. A genuinely nonexistent agent (no ownership row) still 404s. (`db.get_agent_owner` already filters soft-deleted rows, so a *not-yet-recovered* agent correctly still 404s — you must recover first.)
- **`recreate_missing_container(agent_name)`**: reconstructs the container spec from persisted state — `agent_ownership` (resources, capabilities, subscription, PAT, guardrails) + `agent_git_config` (repo/PAT) + the surviving **volume's `template.yaml`** for `type`/`runtime` (labels are gone with the container; the volume survives). Reuses the existing `agent-<name>-workspace` volume — **never recreated, no data loss**; `startup.sh` sees `.git` and skips the clone.
- Extracted the shared **folder-provision + secured `containers_run` tail** into `_provision_folders_and_run_agent_container()`, used by *both* recreate paths — so the full security posture (cap-drop ALL, AppArmor, noexec tmpfs, derived `TRINITY_AGENT_AUTH_TOKEN`, resource limits) is byte-identical and can't drift. **Not a hand-rolled `docker run`** (AC).
- `recover_agent`: message now points to `start` (which rebuilds); `needs_container_recreate` is honored by that path.

## AC coverage

- [x] Recovery restores a **startable** agent (start recreates the missing container from persisted config)
- [x] Existing workspace volume reused, never recreated (base_volumes pins `agent-<name>-workspace`; volume_get, no volume_create)
- [x] Full security posture via the shared supported path (not hand-rolled)
- [x] Subscription / resource limits / capabilities / guardrails / PAT survive (sourced from `agent_ownership` + `agent_git_config`)
- [x] `needs_container_recreate` honored by `start_agent_internal`
- [~] Regression: unit-tested the rebuild wiring + config reconstruction with a mocked docker/db layer (below). The full "reaches running + /stats 200" assertion needs a real container — I'll verify that against the local stack and report (per review).

## Tests (`tests/unit/test_start_agent_skip_inject.py::TestRecoverRecreate1559`)

- `start` 404s when there's no container **and** no ownership row;
- `start` **rebuilds** (calls `recreate_missing_container`) when the container is missing but the agent is live, then actually starts;
- `recreate_missing_container` reconstructs `type`/`runtime` from the volume's `template.yaml` and pins the existing workspace volume as the base mount.

All green locally (`test_start_agent_skip_inject.py`, `test_admin_recovery.py`).

## Note on MCP key

The old agent-scoped MCP key's plaintext is unrecoverable (only the hash is stored), so the rebuild **mints a fresh** agent-scoped key — same wiring as create; enables collab + heartbeat.

Related to #1559

### DIFF
--- a/src/backend/routers/admin_recovery.py
+++ b/src/backend/routers/admin_recovery.py
@@ -144,13 +144,14 @@ async def recover_agent(
             f"Agent {agent_name} recovered: all relational state "
             f"(chat history, schedules, sharing, permissions, "
             f"credentials config) is restored and the agent is visible "
-            f"again. The Docker container was removed at soft-delete and "
-            f"is NOT recreated by recovery — bringing the agent back "
-            f"online (container recreate from the preserved workspace "
-            f"volume) is tracked as #834 Phase 2. Until then the agent "
-            f"shows status=stopped with needs_start=true."
+            f"again. The Docker container was removed at soft-delete; "
+            f"call POST /api/agents/{agent_name}/start to bring it back "
+            f"online — start now rebuilds the missing container from the "
+            f"preserved workspace volume and persisted config (#1559)."
         ),
         "agent_name": agent_name,
+        # #1559: honored by start_agent_internal, which rebuilds the container
+        # from persisted state when it finds a live agent with no container.
         "needs_container_recreate": True,
     }
 
--- a/src/backend/services/agent_service/lifecycle.py
+++ b/src/backend/services/agent_service/lifecycle.py
@@ -17,6 +17,7 @@ from database import db
 from services.docker_service import (
     docker_client,
     get_agent_container,
+    get_next_available_port,
 )
 from services.docker_utils import (
     container_stop, container_remove, container_start, container_reload,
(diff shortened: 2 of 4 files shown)


vybe:
**Triage note:** the `regression diff` job flags **5 new backend unit-test failures** introduced by this branch (not present on BASE), starting with:

```
[E] test_agent_readiness_probe::test_returns_true_after_initial_connect_errors
```

`backend-unit-test` / `regression diff` aren't required checks, so they won't lock the merge button — but these look like genuine regressions from the container-rebuild-on-start change. Worth fixing the readiness-probe fixture/logic before landing.

<sub>Automated triage from `/open-pull-requests`.</sub>

vybe (approved):
Validated via /validate-pr: recovery recreate funnels through the shared _provision_folders_and_run_agent_container tail (security posture can't drift), genuinely-nonexistent agents still 404, spec reconstructed from persisted DB state + volume template.yaml (RO, network-less throwaway container). Regression tests name #1559 incl. the negative path. Security scan clean; CI green incl. full pytest matrix. ✅

PR #1571, opened 2026-07-10
fix(reliability): decouple agent autonomy from the circuit breaker (#1557)

### LINKED ISSUE #1557: bug: Disabling agent autonomy forces circuit breaker DORMANT and blocks all inbound user responses

webmixgamer:
## Summary

- Disabling an agent's **autonomy** forced its **transport** circuit breaker `dormant` (`force_circuit_dormant(reason="autonomy_disabled")`). The `execute_task` gate consults that breaker for **every** trigger, so a healthy paused agent fast-failed **all inbound chat** (manual UI, Telegram, Slack, public link, webhooks) with `"Agent circuit breaker open — agent is unhealthy"` — the agent was never contacted (failures=0, `/health`=200).
- Autonomy governs **proactive work only**. It now acts solely via `set_schedule_enabled(...)` and never touches the breaker. Inbound chat to a paused agent works normally.
- The misleading fast-fail message is split to name the breaker that actually fired: **transport → unreachable**, **dispatch → auth-dead** (was a blanket "unhealthy").

## Why this direction (fix the mechanism, not gate-by-trigger)

`force_circuit_dormant` was the #631 AC#5 hook ("disabling autonomy suppresses breaker probing"). It conflated *"administratively paused"* with *"transport unhealthy"*. Removing the conflation — rather than exempting "inbound" trigger types at the gate — is correct because:

- **#631's flood protection does not depend on it.** A genuinely-down agent is still throttled by the breaker's own failure-driven backoff/dormant path (fed by the pollers' real probes), plus the #1464 leader lock (one worker probes) and #1121 (monitoring default-off). The hook only changed *when* dormant is reached, and it wrote an inconsistent dormant that self-healed a healthy agent to `closed` in ~300s anyway — so its only real effect was a ~300s DoS on a healthy paused agent.
- **Proactive suppression already lives at the right layer** — the schedules are disabled, independent of the breaker.
- The gate-by-trigger alternative would need an "inbound" allow-list over `_TRIGGER_BUCKETS`, which is a **display** map (missing keys silently bucket to "Other") and unsafe as an authorization gate.

A reviewed alternative (short-circuit health *monitoring* on autonomy-off to show a "paused" state) was **rejected**: it re-commits the same conflation on the monitoring axis and would mask a real crash on a paused-but-live agent — which, post-fix, still takes inbound traffic and so *should* be monitored.

## Interaction with #1560 (PR #1568, merged)

#1560 made `start_agent_internal` clear the breakers on any recreate, which introduced a documented interaction: restarting an autonomy-disabled agent silently wiped its `dormant` marker so it answered again while the autonomy flag stayed `false`. Removing the autonomy↔breaker coupling here **dissolves** that interaction — there is no autonomy-driven marker to wipe.

## Changes

- `services/agent_service/autonomy.py` — remove the `force_circuit_dormant`/`reset_circuit` block (both branches). Dropping the enable-path `reset_circuit` also removes a latent footgun (it would wipe a legitimately-open breaker on a wedged agent) and is safe because #1560 already clears breakers on every real (re)start.
- `services/task_execution_service.py` — new pure `_circuit_breaker_error(transport_open, dispatch_open)` helper; used at the Step 3b fast-fail. Every branch keeps the substring `circuit breaker open`.
- `services/agent_client.py` — `force_circuit_dormant` kept as a documented test-only/operator primitive (no production caller).
- Docs: architecture.md (Circuit Breakers), feature-flows (autonomy-mode, dispatch-circuit-breaker, task-execution-service), learnings.md.

## Test Plan

- [x] New unit `tests/unit/test_1557_autonomy_breaker_decoupled.py` — structural regression guards (autonomy.py no longer references the breaker; still disables schedules) + the 3 message branches + the pinned-substring guard. **Mutation-proven**: reverting `autonomy.py` to HEAD makes exactly the 2 structural guards fail.
- [x] New opt-in integration `tests/integration/test_1557_autonomy_inbound.py` (`TRINITY_AUTONOMY_TEST_AGENT`) — real Redis + API: disabling autonomy on a running healthy agent leaves the circuit not-dormant and `POST /task` returns 200. Passed live.
- [x] Regression sweep — 99 breaker/monitoring tests + 58 execution-gate tests pass; `sys.modules` lint gate green. (Pre-existing `fakeredis`-collection skips are environmental, unrelated.)
- [x] Verified live: autonomy off → circuit `closed` → `/task` returns 200; forced-open transport → new honest message.

Fixes #1557


vybe (approved):
Validated via /validate-pr: Fixes #1557 (P1 bug) with named regression tests (unit/test_1557_autonomy_breaker_decoupled.py structural guards + opt-in integration leg), architecture.md + 3 feature-flow docs updated, learnings entry added, security greps clean, no new env vars. Conflict-resolution merges (learnings.md, tests/registry.json — both-appended entries, kept both) re-ran the full suite green. Approving.

PR #1586, opened 2026-07-13
fix(tests): conftest no longer deletes real agents named test-* (#1558)

### LINKED ISSUE #1558: bug(tests): conftest api_client fixture deletes every live agent whose name starts with `test-`
## Summary

`tests/conftest.py`'s session-scoped `api_client` fixture enumerates **every agent on the target Trinity instance** and deletes any whose name begins with `test-`. Because nearly every API test depends on `api_client`, running *any* of `tests/*.py` against a live instance destroys real agents — silently, with no confirmation and no opt-in.

## Evidence

`tests/conftest.py` (in the `api_client` fixture, `scope="session"`):

```python
response = client.get("/api/agents")
if response.status_code == 200:
    agents = response.json()
    leftover = [a["name"] for a in agents if a["name"].startswith("test-")]
    for name in leftover:
        cleanup_test_agent(client, name)
```

The target instance is whatever `TRINITY_API_URL` points at, authenticated with `TRINITY_TEST_PASSWORD`. There is no guard on the URL being localhost, no `--dangerously-allow-destructive` flag, and no confirmation prompt.

## Observed impact

While running the channel/router suites against a **local dev instance** to validate a fix, this fixture soft-deleted a developer's real agent named `test-agent-2` (created weeks earlier, with a bound Telegram bot, chat history, and a GitHub-sync binding). Nothing in the invocation suggested it would mutate existing agents. The audit log records `stop` + `delete` originating from the test container's IP on the platform network.

Agents that happened not to match the prefix (`testfix`, `testfix-2`) survived purely by naming luck.

The blast radius is bounded only by the naming convention: **any agent named `test-*` on a staging or production instance is destroyed** the moment someone points the suite at it. `type-bug` rather than a nice-to-have because it is silent, unprompted data loss.

## Why the current design exists

The comment says: *"Also cleans up leftover test agents from prior runs to avoid quota exhaustion."* That is a legitimate need — the suite creates ephemeral agents (`test-stopped-<hex>`, `test-shared-session-<hex>`) and a crashed run leaves them behind.

## Acceptance criteria

- [ ] The fixture never deletes an agent it did not create in the current session, OR it only deletes agents matching the suite's own generated-name pattern (a dedicated prefix such as `pytest-ephemeral-<hex>`, not the broad `test-`)
- [ ] Ephemeral agents created by the suite are registered in a session-scoped set and torn down explicitly by name
- [ ] A destructive sweep, if retained at all, is opt-in behind an explicit env flag AND refuses to run when `TRINITY_API_URL` is not localhost
- [ ] `tests/README` (or the suite docstring) states plainly that the suite mutates the target instance
- [ ] Regression test: the fixture leaves a pre-existing `test-`-prefixed agent untouched

## Notes

Two follow-on hardenings worth considering: rename the suite's generated agents away from the `test-` prefix so no user agent can ever collide, and have `cleanup_test_agent` refuse any name it cannot prove the session created.

dolho:
## Problem (P1, silent data loss)

The session-scoped `api_client` fixture in `tests/conftest.py` enumerated **every agent** on the target instance and deleted any whose name started with `test-`:

```python
leftover = [a["name"] for a in agents if a["name"].startswith("test-")]
for name in leftover:
    cleanup_test_agent(client, name)
```

Because nearly every API test depends on `api_client`, running *any* `tests/*.py` against a live instance destroyed real agents — no confirmation, no opt-in, no URL guard. It already soft-deleted a developer's real `test-agent-2` (bound Telegram bot, chat history, GitHub sync). Any `test-*` agent on staging/prod was one invocation from deletion.

## Fix — fail-closed on every axis

- **Dedicated prefix**: the suite now names every agent it creates `pytest-ephemeral-*` (a namespace no human uses), never the broad `test-`. Renamed all conftest agent generators: `test_agent_name`, `module_agent_name`, `stopped_agent`, `shared_agent`.
- **Session registry**: created names are registered (`register_created_agent`) and reclaimed by name at session end — belt-and-suspenders over each fixture's own teardown.
- **Startup leftover-sweep is now opt-in**: gated behind `TRINITY_TEST_CLEANUP_SWEEP`, **refuses any non-localhost target** (`is_local_target`), and only removes provably suite-owned names. The decision is a pure, fail-closed helper `select_sweepable_agents(names, base_url, *, enabled)` → `[]` unless enabled **and** localhost, then only `pytest-ephemeral-` (or session-registered) names. **Default: no sweep at all.**
- **`cleanup_test_agent(require_suite_owned=True)`** refuses — with no stop/delete call — a name the session can't prove it created. Used by every sweep path.
- **Docs**: the conftest docstring and `tests/README.md` now state plainly that the suite mutates the target instance and must point at localhost.

## Acceptance criteria

- [x] Fixture never deletes an agent it didn't create; the sweep only matches the dedicated `pytest-ephemeral-` prefix (or session-registered names), never broad `test-`.
- [x] Suite-created agents registered in a session set, torn down explicitly by name.
- [x] Destructive sweep is opt-in behind an env flag **and** refuses non-localhost `TRINITY_API_URL`.
- [x] `tests/README` + suite docstring state the suite mutates the target instance.
- [x] Regression test: a pre-existing `test-`-prefixed agent is left untouched.
- [x] (Follow-ons from the issue Notes) generated agents renamed off `test-`; `cleanup_test_agent` refuses names it can't prove the session created.

## Tests

`tests/unit/test_1558_conftest_agent_sweep_guard.py` — **17 cases, pure (no network)**: the exact `test-agent-2`-survives scenario, opt-in/localhost gating, prefix-only selection, `is_local_target` truth table, and the `cleanup_test_agent` refusal path (asserts no API call for an unowned name).

```
17 passed  (run in a throwaway container off trinity-backend:latest)
```

Note: `tests/utils` is a symlink to `tests/testing_utils/` — the changed file is `tests/testing_utils/cleanup.py`.

Related to #1558

Fixes #1558

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,5 +1,14 @@
 # Trinity Tests
 
+> ⚠️ **This suite mutates the target instance.** The API fixtures **create and
+> delete real agents** on whatever `TRINITY_API_URL` points at. **Point it at a
+> local dev instance — never staging/production.** Every agent the suite creates
+> is named with the `pytest-ephemeral-` prefix and torn down by name; the suite
+> **never deletes an agent it did not create**. A leftover-sweep for crashed
+> runs is **OFF by default** — enable with `TRINITY_TEST_CLEANUP_SWEEP=1`, and
+> even then it only removes `pytest-ephemeral-` agents on a localhost target
+> (#1558).
+
 ## Quick start
 
 ```bash
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,14 @@
 """
 Shared pytest fixtures for Trinity API tests.
 
+⚠️  THIS SUITE MUTATES THE TARGET INSTANCE. The API fixtures CREATE and DELETE
+real agents on whatever ``TRINITY_API_URL`` points at. Point it at a local dev
+instance — never staging/production. Every agent the suite creates is named
+with the ``pytest-ephemeral-`` prefix and is torn down by name; the suite never
+deletes an agent it did not create. The optional leftover-sweep (for crashed
+runs) is OFF by default and, even when enabled, only removes
+``pytest-ephemeral-`` agents on a localhost target (#1558).
+
 OPTIMIZATION NOTES (2025-12-09):
 - Module-scoped agent fixture: Creates ONE agent per test FILE (not per test)
 - Session-scoped shared agent: Single agent for tests that can share
@@ -12,6 +20,9 @@ Configuration:
 - TRINITY_TEST_PASSWORD: Test user password (default: password)
 - TRINITY_MCP_API_KEY: MCP API key for authenticated tests
 - TEST_AGENT_NAME: Pre-existing agent for agent-server tests
+- TRINITY_TEST_CLEANUP_SWEEP: opt-in (1/true) — sweep leftover
+  ``pytest-ephemeral-`` agents from prior crashed runs at session start.
+  Refuses unless TRINITY_API_URL is localhost. Default OFF (#1558).
 """
 
 # Skip test files that require backend context (can't be run from test suite)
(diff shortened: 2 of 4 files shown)


vybe (approved):
Validated via /validate-pr: fail-closed sweep policy with strong regression tests; full pytest matrix green. Approving for merge.

PR #1589, opened 2026-07-13
fix(operator-queue): quarantine persistently-failing creates to stop the sync hot-loop (#1525)

### LINKED ISSUE #1525: bug: operator-queue create_item KeyError on missing created_at causes an infinite sync-retry loop
## Summary

`OperatorQueueOps.create_item()` reads several fields from the agent-supplied request dict via **bracket access** — `item["id"]`, `item["title"]`, `item["question"]`, and `item["created_at"]`. If an agent writes a request to its `.trinity/operator-queue.json` that omits `created_at` (or any of those keys), `create_item` raises `KeyError: 'created_at'`.

Because the backend's per-agent sync only calls `create_item` for requests that are `pending` **and** `not db.operator_queue_item_exists(req_id)`, and the failing create never persists the row, `operator_queue_item_exists` stays `False` forever. So **every subsequent sync cycle re-attempts the same create and re-fails**, producing a persistent error loop (observed roughly every 2–5s) and the request never surfaces to the operator. The loop only stops when the agent flips the request out of `pending` (e.g. to `acknowledged`).

## Component

Backend / Operator Queue (`db.operator_queue` + `services.operator_queue_service`)

## Priority

P2 — a single malformed agent request produces a permanent, high-frequency ERROR loop and silently drops a legitimate operator-queue alert/approval. Self-limiting (stops when the request leaves `pending`) and requires malformed input, so not P1; but it is noisy and data-losing while active.

## Error

```
ERROR services.operator_queue_service: Failed to create queue item req-<id>: 'created_at'
```
(underlying exception is `KeyError: 'created_at'`, repeating on every sync cycle)

## Location

- **File**: `src/backend/db/operator_queue.py`
- **Function**: `OperatorQueueOps.create_item()` (def at line 77)
- **Offending line**: line 102 — `created_at=item["created_at"],`
- Same fragility on lines 87 / 97 / 98 — `item["id"]`, `item["title"]`, `item["question"]`
- **Retry driver**: `src/backend/services/operator_queue_service.py`, the sync loop's `if req_status == "pending" and not db.operator_queue_item_exists(req_id): ... db.create_operator_queue_item(...)` branch (the `except` logs but does not quarantine the bad request, so it is retried indefinitely).

## Root Cause

`create_item` was rewritten to the SQLAlchemy `make_insert(operator_queue).values(...)` pattern as part of the configurable-backend refactor (#300 / #1093). That rewrite uses hard bracket access for `created_at` (and `id`/`title`/`question`) with no default or validation, so a request missing `created_at` becomes an unhandled `KeyError`. Combined with the "create only if not already in DB" guard in the sync loop, a permanently-failing create turns into an unbounded retry loop. This is pure-Python and DB-backend-agnostic (fails identically on SQLite and PostgreSQL).

## Reproduction Steps

1. Have an agent write a request to `.trinity/operator-queue.json` with `status: "pending"`, valid `id`/`title`/`question`, but **no `created_at`** field.
2. Let the backend operator-queue sync run.
3. `create_item` raises `KeyError: 'created_at'`; the item is never persisted.
4. Because `operator_queue_item_exists(id)` stays `False`, every subsequent sync re-attempts the create and re-logs the error — a persistent ~2–5s ERROR loop; the request never appears in the operator queue.

## Suggested Fix

Default `created_at` to "now" when absent, and treat the required text fields defensively so a single malformed request cannot hot-loop:

```python
from datetime import datetime, timezone

def create_item(self, agent_name: str, item: Dict) -> str:
    item_id = item.get("id")
    if not item_id or "title" not in item or "question" not in item:
        # Malformed request: skip rather than raise (caller retries forever otherwise)
        raise ValueError(f"operator-queue request missing required fields: {item.get('id')!r}")

    created_at = item.get("created_at") or datetime.now(timezone.utc).isoformat()
    ...
    stmt = make_insert(operator_queue).values(
        id=item_id,
        ...
        title=item.get("title", ""),
        question=item.get("question", ""),
        created_at=created_at,
        ...
    ).on_conflict_do_nothing(index_elements=["id"])
```

Additionally, consider having the sync loop **quarantine** a request whose create raises (e.g. record a failure marker / skip after N attempts) so a persistently-malformed entry cannot generate an unbounded error stream.

## Environment

- Trinity `dev` @ `58354873` (present since the #300 backend refactor, `#1093`).
- Affects both SQLite and PostgreSQL backends (KeyError raised before any DB call).

## Related

- `create_item` introduced/rewritten in #300 configurable database backend (#1093).
- Retry driver: `services/operator_queue_service.py` sync loop (pending-create branch).

dolho:
## Problem

A `pending` operator-queue request whose DB create keeps failing was re-attempted on **every ~5s sync cycle forever**: the failing create never persists the row → `operator_queue_item_exists(id)` stays `False` → the sync loop re-creates + re-`ERROR`-logs it indefinitely, and the request never reaches the Operating Room.

## Finding (part of this was already fixed)

The specific `created_at` **KeyError** from the issue was **already made defensive on `dev`** by #1426 — `create_item` now uses `.get(...) or utc_now_iso()` for the previously hard-indexed fields plus `on_conflict_do_nothing`. So the field-defaulting case no longer loops.

But the **retry-loop class remains**: the sync loop still re-attempts *any* failing create (a DB error, a constraint, a future field) every cycle with no bound. This PR closes that — the "quarantine" the issue's *Suggested Fix* explicitly calls for.

## Fix

- **`services/operator_queue_service._sync_agent`**: track per-request consecutive create-failure counts. After `MAX_CREATE_ATTEMPTS` (3), the request is **quarantined** (skipped) — one WARN at the threshold instead of an unbounded ~5s ERROR stream. A create that later succeeds clears the counter; an in-memory safety valve caps the map so it can never grow without bound. Retried again after a process restart or once the request changes.
- **`db/operator_queue.create_item`**: the last hard-indexed field (`id`) now uses `.get` and raises a clear **`ValueError`** (which the caller quarantines) instead of an opaque `KeyError`; the WS-broadcast `item["id"]` hardened too.

## Tests

`tests/unit/test_1525_operator_queue_quarantine.py` — **4, pure/mocked (no backend/DB)**:
- a persistently-failing create is attempted **at most the cap**, not once per cycle (the bug);
- a create that succeeds clears the counter;
- a healthy request leaks no quarantine state;
- `create_item` raises `ValueError` (not `KeyError`) on a missing `id`.

```
4 passed
```
Run in a throwaway container off `trinity-backend:latest`. Backend-backend-agnostic (fails/loops identically on SQLite + PostgreSQL); no schema change.

Related to #1525

Fixes #1525

### DIFF
--- a/src/backend/db/operator_queue.py
+++ b/src/backend/db/operator_queue.py
@@ -84,7 +84,13 @@ class OperatorQueueOperations:
         Returns:
             The item ID
         """
-        item_id = item["id"]
+        # #1525: `id` was the last hard-indexed field. The sync loop guards on a
+        # truthy id before calling, but keep the DB boundary self-defensive so an
+        # id-less item can never KeyError-hot-loop here (raise a clear ValueError
+        # the caller quarantines, rather than an opaque KeyError).
+        item_id = item.get("id")
+        if not item_id:
+            raise ValueError("operator-queue item is missing a required 'id'")
         options_json = json.dumps(item.get("options")) if item.get("options") else None
         context_json = json.dumps(item.get("context")) if item.get("context") else None
 
--- a/src/backend/services/operator_queue_service.py
+++ b/src/backend/services/operator_queue_service.py
@@ -29,6 +29,14 @@ _websocket_manager = None
 
 QUEUE_FILE_PATH = ".trinity/operator-queue.json"
 DEFAULT_POLL_INTERVAL = 5  # seconds
+# #1525: after this many consecutive failed create attempts for the same request
+# id, quarantine it — stop re-attempting so one persistently-failing create can't
+# produce an unbounded ~5s ERROR loop. It's retried again after a process restart
+# (or if it leaves `pending` / gets created some other way).
+MAX_CREATE_ATTEMPTS = 3
+# Safety valve so the in-memory quarantine map can never grow without bound if an
+# agent streams unique failing ids; cleared wholesale past this size.
+_MAX_QUARANTINE_ENTRIES = 5000
 
 
 def set_websocket_manager(manager):
@@ -44,6 +52,10 @@ class OperatorQueueSyncService:
         self.poll_interval = poll_interval
         self._task: Optional[asyncio.Task] = None
         self._running = False
+        # #1525: req_id → consecutive create-failure count. Bounds the retry loop
+        # for a request whose DB create keeps raising (malformed input, a DB
+        # error, …) so it can't hot-loop forever. Entry is dropped on success.
+        self._create_failures: dict[str, int] = {}
 
     def start(self):
         """Start the background polling loop."""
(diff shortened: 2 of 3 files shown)


vybe (approved):
Validated via /validate-pr: docs, security, packaging, and test-adequacy checks pass. Approving for merge.

PR #1594, opened 2026-07-13
chore(dev-skills): bump .claude — /cso v1.1 Trinity-shape refresh

vybe:
## Summary

Pointer bump of the private `.claude` submodule to trinity-dev `2bdd362` — the `/cso` security-audit skill refreshed against Trinity's current shape (`architecture.md` + `TARGET_ARCHITECTURE.md`).

**Stale checks fixed** (previously flagged intentional design or under-checked reality):
- Setup check inverted: the single-use bootstrap token was deliberately removed (ent#49) — now verifies post-setup lockout instead of flagging the token's absence
- Redis: checks the actual ACL model (`backend`/`scheduler` with `-@dangerous`, mandatory passwords) instead of bare `requirepass`; names the two-network invariant (#589) and the hard-coded agent-creation network sites
- Stack facts: dual-track SQLite+PostgreSQL/Alembic (#1183), standalone scheduler, TS MCP server
- Webhook FP rule: no HMAC on the token-authenticated `/api/webhooks/{token}` is by-design (signing opt-in, ent#77), not a finding

**New-surface checks added**:
- Agent-scoped-key self-boundaries on heartbeat (#307), result callback (#1083), reports (#918) — sibling-key rejection
- Backend→agent `X-Trinity-Agent-Token` completeness via the #1159 static guard
- ent#77 fail-closed signature verification; Twilio/Slack/Telegram transport checks
- Vendored-copy parity (Invariant #5) — credential-path deny-list drift = injection bypass
- A01: enumeration uniformity (#186) + `AuthorizedAgent` vs `OwnedAgentByName` mutation tiers
- LLM/secrets: name-only MCP tool descriptions (#846), Brain Orb voice-token config-lock, webhook `context` framing, `backlog_metadata` as a G-04-class secret surface, enterprise-docs-guard coverage (ent#45)

Also adds the `metadata.version`/changelog block + what's-new banner (skill was previously unversioned).

## Test plan

- [x] Skill structure verified intact after edits (all 15 phases, checklist, report format unchanged)
- [x] Submodule commit pushed to trinity-dev `main` (`2bdd362`)
- [ ] Next `/cso` run confirms no false-fires on the setup-token / Redis / webhook checks

Docs-only skill change in the private submodule — no product code touched.


dolho:
## Review — LGTM

Submodule pointer bump only (`.claude` `2228cfc7` → `2bdd362b`). Confirmed locally that `2bdd362` is reachable and is the CSO v1.1 refactor commit (`refactor(cso): Trinity-shape refresh — v1.1`), so the pointer is not dangling. No product code touched; the skill doesn’t run in CI. CI green.

Standard submodule-bump caveat: the actual skill diff isn’t reviewable from the public repo (private `trinity-dev`), so the substance rides on the trinity-dev side. The stale-check corrections called out in the description (setup post-lockout instead of the removed bootstrap token ent#49, the real Redis ACL model + two-network invariant #589, dual-track SQLite+PG #1183, the by-design tokenless webhook ent#77) all match current `architecture.md`, so the refresh is pointed the right way.

Only open item is the unchecked box — a `/cso` run confirming no false-fires on the setup-token / Redis / webhook checks. Fine to land the pointer bump ahead of that.

dolho (approved):

Issue #1596, opened 2026-07-13
bug: git-sync agent workspaces grow unboundedly — auto-committed data churn bloats .git (44GB observed on one agent)


vybe:
## Summary

Agents using git sync (auto-commit of workspace state) accumulate unbounded `.git` history when their skills write large, frequently-churned data files (datasets, JSON snapshots, indexes). Git retains every historical version of every committed file, and `git gc --auto` only stacks incremental packs without consolidating. On one production instance this has grown a single agent's `.git` to **44GB** and the fleet's combined workspace git history to **~110GB+**, driving the host disk to 90% — and the same bloat is pushed to the GitHub remote on every sync.

## Component

Agent Runtime / git-sync (workspace auto-commit) — plus agent base image git hygiene

## Priority

P1 — threatens instance availability (disk exhaustion); no built-in mitigation exists; manual surgery on workspace repos is the only current recourse.

## Observed State (one agent, sanitized)

```
$ git count-objects -vH        # inside the agent workspace
count: 49450                   # loose objects
size: 9.04 GiB
in-pack: 4264089               # 4.26M objects — from only 2,146 commits
packs: 2267                    # gc --auto keeps stacking packs, never consolidates
size-pack: 33.06 GiB
size-garbage: 1.06 GiB
# .git total ≈ 44GB; workspace checkout itself is a small fraction of that
```

Other agents on the same instance show the same curve at earlier stages (23GB, 16GB, 13GB, 5GB…). Growth compounds on every auto-commit cycle and is mirrored to the GitHub remote.

## Root Cause

1. Workspace auto-commit snapshots **everything** in the workspace, including large data files that change on every scheduled run — each change stores a full new blob in history.
2. No default `.gitignore` conventions (or size guardrails) exclude data/cache directories from sync.
3. No maintenance strategy: thousands of small packs accumulate (`git gc --auto` never consolidates under this write pattern), loose objects pile up, and reachable history is never compacted, so growth is strictly monotonic.

## Reproduction Steps

1. Enable git sync / auto-commit for an agent whose skill writes a multi-MB data file (e.g. a JSON snapshot) on each run
2. Schedule the skill to run frequently (e.g. every 15–60 min)
3. Let it run for several weeks
4. Observe `.git/objects` growing by the file size × run count, `git count-objects -vH` showing thousands of packs, and the same growth on the remote repo

## Suggested Fix

Investigation needed — likely a combination of:

[...]

Issue #1595, opened 2026-07-13
bug: git auto-gc can never complete inside agent containers — unbounded, silent workspace .git bloat (273GB observed; 97%-garbage repos)


obasilakis:
## Summary

Git's **auto-gc detaches into the background by default** (`gc.autoDetach=true`). A detached background process inside an agent container reparents to **PID 1** — which is exactly the signature the #817 orphan sweeper exists to SIGKILL. So `git gc` is killed within ≤30s, **every time, on every git-enabled agent, forever**. It has never once completed.

Consequences compound into a self-reinforcing ratchet:

- Packs never consolidate → **git cannot delta-compress across separate packs** → every version of every file is stored standalone.
- Unreachable objects are never pruned.
- Each killed gc **abandons its `tmp_pack_*` write buffers** — so the failed maintenance is *itself* a garbage generator.
- Bigger repo → slower gc → even more reliably killed. It never recovers.

This is **silent**: no error surfaces, agents keep working, and the only artifact is a stale `.git/gc.pid`. It ran undetected for ~2 months on our fleet and only surfaced as a disk-capacity incident.

This is a **distinct defect from #1505** (which is about the allowlist *seam*): the point here is that the platform offers **no viable git-maintenance path at all** for git-enabled agents, and **no signal** when their repos rot.

## Component

Agent Runtime (base image) — `agent_server` orphan sweeper (`utils/orphan_allowlist.py`) × git-enabled agent workspaces

## Priority

P1 — unbounded disk growth on every git-enabled agent, no operator-visible signal, and it has already caused a multi-day silent outage (below).

## Evidence (production, eu2 — 10 git-enabled agents)

Probed on the host. Worst case, `oracle-4-politics`:

| Signal | Value |
|--------|-------|
| Workspace / `.git` | 54G / **44G** |
| **Working tree (checked-out files)** | **20K** |
| Total git objects | **4,313,374** |
| Reachable from refs **+ reflog** | **131,780** → **96.9% of the repo is unreachable garbage** |
| Pack files | **2,267** (33.06 GiB) — a healthy repo has 1–5 |
| Orphaned `tmp_pack_*`/`tmp_obj_*` | 36 files, **1.06 GiB** (git reports these as `garbage:`) |
| Largest 8 blobs (219MB, 197MB, 190MB…) | **all unreachable** — dangling |
| `.git/gc.log` | **absent** — gc never *errored*, it was killed hard |
[...]

Issue #1505, opened 2026-07-07
orphan sweep: evaluate protecting the live agent-server subtree (descendants_of(sweep_pid)) vs per-site transient registration


vybe:
## Context

#1501 fixed the orphan sweep SIGKILLing in-flight brain-orb hooks via per-site transient-pid registration (`ProcessRegistry.add_transient_pid` around each spawn). During review, an independent strategy pass proposed a strictly-more-general seam that was deliberately **not** adopted in the P1 fix because it changes the sweep's safety philosophy and deserves its own discussion + canary soak.

## The proposal

Add `allowlist.update(descendants_of(sweep_pid))` to `resolve_allowlist` (`agent_server/utils/orphan_allowlist.py`) — protect the whole **live** agent-server subtree.

**Why it's plausibly safe:** a genuinely leaked orphan always reparents to PID 1 (nothing in the container is a subreaper), never *into* the agent-server subtree — so reparented leaks stay killable. The docstring's existing objection ("protect every child of PID 1 shields reparented orphans") does not apply to the agent-server's subtree.

**What it buys:** covers every current and future agent-server spawn site with zero per-call-site contract — brain-orb hooks, the `routers/git.py` / auto-sync `subprocess.run` git children (see below), Gemini's post-turn sweeps, and anything added later. No new state, no TTL, no "remember to register" latent bug class.

**What it costs:** flips the sweep from deny-by-default to allow-by-default for the agent-server subtree. A hung/leaked **live-chained** child (e.g. a wedged git push with no timeout) is shielded forever instead of reaped in ≤30s. #817 was built deny-by-default from hard production lessons; #1501's registration approach preserves that guarantee (protection is scoped + deadline-bounded).

## Known residual exposure (documented in #1501)

`agent_server/routers/git.py` runs git via in-process `subprocess.run` — those children are bare agent-server children, sweep-exposed exactly like the pre-fix hooks. They're short-lived (~1–2s), so the per-op kill probability is roughly `runtime / 30s` (a few percent per auto-sync cycle) — rare, and would surface as sporadic unexplained sync failures. If the seam is rejected, these sites should get the transient-registration treatment (or a shared wrapper) instead.

## Acceptance criteria

- [ ] Decision recorded: adopt the seam, extend per-site registration to git.py, or accept the exposure (with rationale in `orphan_allowlist.py`'s docstring either way)
[...]

Issue #1169, opened 2026-06-12
Agent runtime data volumes: declared data paths with snapshot/restore and portable export


vybe:
## Summary

Agents accumulate runtime data — SQLite databases, large incremental files, working datasets — that cannot live in the git-synced template repo (it bloats fast and doesn't belong in version control), yet must survive container recreate, template re-pulls, and subscription auto-switch, and must move with the agent when it migrates to another Trinity instance.

Proposed model: **agent = code (git) + credentials (`.credentials.enc`) + data (declared volume)**. Runtime data becomes a third declared artifact class, parallel to the existing credential pattern.

## Context

Today the only options are committing data to the template repo (bloat, churn, merge conflicts) or leaving it in the container filesystem (lost on recreate — exactly the events that already destroy runtime state today: template re-pull, subscription auto-switch, image upgrade). Neither is acceptable for database-backed or data-heavy agents.

The platform already has the adjacent primitives:
- `.trinity/persistent-state.yaml` (S4, #383) — the "paths that survive reset" allowlist
- Per-agent named volumes (`agent-{name}-public`, FILES-001; shared-folder volumes)
- Backend-side Docker SDK file extraction (`get_archive`) that never mounts the agent workspace (`src/backend/services/agent_shared_files_service.py`)
- The `~/.trinity/pre-check` hook pattern for agent-side lifecycle hooks
- Credential export/import as the portability surface

## Design Sketch

1. **Declare** — `data_paths:` section in `template.yaml` (or an extension of the S4 persistent-state allowlist) listing runtime-data paths, e.g. `~/data/`, `~/db/agent.sqlite`. Gitignored by convention.
2. **Mount** — Trinity mounts declared paths from a named per-agent volume (`agent-{name}-data`), same pattern as `agent-{name}-public`. Survives recreate/re-pull/auto-switch.
3. **Durability** — a backend snapshot job tars the volume on a schedule into the existing `trinity-data` bind mount (e.g. `/data/agent-data-snapshots/`) with retention. For SQLite, snapshot via a quiesce hook (`~/.trinity/pre-snapshot`, shebang-selected like pre-check) or `sqlite3 .backup` to avoid mid-write corruption.
4. **Portability** — "move agent to another Trinity" = template URL + `.credentials.enc` + data snapshot tar. Extends the existing export/import surface.

## Acceptance Criteria

- [ ] `template.yaml` (or persistent-state allowlist) supports declaring runtime data paths
[...]

PR #1597, opened 2026-07-13
fix(git-sync): bound agent .git bloat — maintenance repack + ignore conventions + size observability (#1596, PR-1)

dolho:
## Problem (P1 — disk exhaustion)

Git-sync auto-commit (`agent_server/routers/git.py::_run_auto_sync_once`) does `git add -A` → commit → push every 15-min heartbeat. Under skills that write churny data files, `.git` grew unboundedly: **one agent's `.git` reached 44 GB / 2,267 packs / 4.26M objects**; the fleet hit ~110 GB and the host 90%.

## ⚠️ Root cause update (see #1595) — scope narrowed; destructive prong dropped

Further investigation (credit: parallel work on **#1595 / #1505**) found the **true** root cause is **not** reachable history retention — it's that **git maintenance can never complete inside agent containers**: git's detached `gc --auto` reparents to PID 1, the #817 orphan sweeper **SIGKILLs** it (a bare SIGKILL, so it leaves stale `.lock`/`gc.pid` litter), and every subsequent gc then aborts on the stale lock — permanently disabling maintenance for that repo. **~97% of the bloat is unreachable garbage, not committed history** (a plain non-destructive `git gc --prune=now` took one agent 44 GB → 1.6 GB, `fsck` clean, HEAD unchanged).

Consequence for this PR:
- The **destructive history-compaction** prong originally proposed here was **dropped** — it solves the wrong problem (unreachable garbage needs no history rewrite) and is unsafe (a long rebase/gc on a big repo is itself SIGKILLed mid-operation by the same sweeper).
- The **load-bearing fix lives in #1595** (sweeper SIGTERM-then-SIGKILL + allowlist git maintenance + stale-lock cleanup), tracked separately to avoid conflicting edits to the sweeper.

### 1. Maintenance repack (agent side)
After a successful push, `_maybe_run_git_maintenance` runs `git repack -adl` + `git gc --prune=now` once the pack count crosses `GIT_MAINTENANCE_PACK_THRESHOLD` (default 20). Self-throttling, best-effort, non-destructive. **Note:** its reliability on large repos depends on the #1595 sweeper fix (an in-container git process is currently sweep-exposed) — it fully self-heals small/medium repos now and all repos once #1595 lands.

### 2. Ignore conventions
`git_service._GITIGNORE_PATTERNS` now also excludes `node_modules/`, `.venv/`, `venv/`, `__pycache__/`, `*.pyc`/`*.pyo`, `.pytest_cache/`, `.mypy_cache/`, `.ruff_cache/`, `.ipynb_checkpoints/`, `*.sqlite`/`*.sqlite3`/`*.db`. Merged into each agent on sync (also **untracks** already-committed matches). Datasets/indexes belong in `data_paths` (#1169).

### 3. Observability
- Agent measures `.git` size (`du -sb`) into `sync-state.json.git_dir_bytes`, returned via `GET /api/git/status`.
- New `agent_sync_state.git_dir_bytes` column (dual migration: SQLite + Alembic `0019`); `SyncHealthService` persists it; surfaced on `GET /api/agents/sync-health`, `GET /api/fleet/sync-audit`, and the per-agent sync-state endpoint.

## Acceptance criteria

- [x] Maintenance: consolidating repack bounds pack count + reclaims loose/garbage (non-destructive).
- [x] Ignore conventions: default `.gitignore` excludes data/cache/venv/index/deps.
- [x] Observability: `.git` size in sync health + fleet audit.
- [ ] **Moved to #1595 (real root cause):** sweeper SIGTERM-then-SIGKILL + git-maintenance allowlist + stale-lock cleanup — the change that lets maintenance actually run.
- [ ] Deferred: warn-on-large-file auto-commit guardrail.

## Tests

`tests/unit/test_1596_git_sync_observability.py` — `git_dir_bytes` round-trips + a partial (failed) update preserves the last size; the default `.gitignore` includes the bulk-data set **and** keeps the pre-existing credential/`content/` safety rules. Schema + Alembic parity green.

## Deploy note

Agent-side maintenance + `.git`-size measurement land on a **base-image rebuild** + agent recreate. The DB/backend surfacing works immediately. Full de-bloat of the existing 44 GB agents comes from **#1595** (fix the sweeper so `gc --prune=now` can run to completion); this PR bounds future growth and gives operators the size curve.

Related to #1596 · root cause #1595 · #1505

### DIFF
--- a/docker/base-image/agent_server/routers/git.py
+++ b/docker/base-image/agent_server/routers/git.py
@@ -95,6 +95,7 @@ _SYNC_STATE_DEFAULT: Dict = {
     "last_sync_at": None,
     "last_error_summary": None,
     "consecutive_failures": 0,
+    "git_dir_bytes": None,  # #1596: last-measured .git on-disk size
 }
 
 
@@ -124,11 +125,14 @@ def _write_sync_state_file(
     last_sync_status: str,
     last_sync_at: Optional[str] = None,
     last_error_summary: Optional[str] = None,
+    git_dir_bytes: Optional[int] = None,
 ) -> Dict:
     """Persist one sync outcome.
 
     consecutive_failures is bumped on `failed`, reset on `success`, untouched
     on `never`. last_error_summary is cleared on success, kept on never.
+    git_dir_bytes (#1596) is updated when measured; a None here preserves the
+    last known value.
     """
     prior = _read_sync_state_file(home_dir)
 
@@ -140,6 +144,9 @@ def _write_sync_state_file(
         prior["last_error_summary"] = None
     # else 'never' — leave counters alone
 
+    if git_dir_bytes is not None:
+        prior["git_dir_bytes"] = git_dir_bytes
+
     prior["last_sync_status"] = last_sync_status
     prior["last_sync_at"] = last_sync_at or datetime.now(timezone.utc).isoformat()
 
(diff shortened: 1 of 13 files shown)


dolho:
Heads-up: I force-pushed this branch back to the **non-destructive PR-1 scope** and dropped the destructive history-compaction commits I'd added.

Reason: the parallel investigation in **#1595 / #1505** identified the true root cause — git maintenance can never complete inside agent containers (detached `gc --auto` reparents to PID 1 → #817 orphan sweeper SIGKILLs it → stale `.lock`/`gc.pid` litter → every later gc aborts). Confirmed in code: `resolve_allowlist`/`_hard_protected_pids` protect only agent-server + its parent chain, so an in-container git process is sweep-exposed. The bloat is ~97% **unreachable garbage**, not committed history — a plain non-destructive `gc --prune=now` reclaims it (44 GB → 1.6 GB, HEAD unchanged).

So rewriting history was the wrong fix (and unsafe — a long rebase would be SIGKILLed mid-operation by the same sweeper). The load-bearing change (sweeper SIGTERM-then-SIGKILL + git-maintenance allowlist + stale-lock cleanup) belongs in **#1595**. This PR stays scoped to the two prongs that stand on their own — ignore conventions + observability — plus the non-destructive repack (whose large-repo reliability depends on #1595).

vybe (approved):
Validated via /validate-pr: non-destructive self-throttling repack (pack-count gated, in-process child so the #817 sweeper doesn't apply), git_dir_bytes observability wired agent→sync-state→dashboard/audit, dual-track migration correct (SQLite agent_sync_state_git_dir_bytes + Alembic 0019 revising 0018), schema.py/tables.py/architecture.md synced, named regression tests. Security scan clean; CI green. Heads-up applied at merge time: #1550 must re-chain its 0019/0020 Alembic revisions onto this one. ✅

Issue #69, opened 2026-03-06
Filter agents by owner on Dashboard and Agents pages


vybe:
## Summary

Add the ability to filter agents by owner on both the Dashboard and Agents pages, so users can quickly find agents belonging to a specific owner.

## Requirements

- Add an owner filter dropdown/selector to the **Dashboard** page
- Add an owner filter dropdown/selector to the **Agents** page
- Populate the filter with the list of distinct agent owners
- Filtering should update the displayed agent list in real-time (client-side)
- "All owners" option to clear the filter
- Filter state should persist during the session (not reset on navigation)

## Motivation

As the number of agents grows, users need a way to quickly narrow down the list to agents owned by a specific person or team. This improves discoverability and reduces visual clutter.

## Acceptance Criteria

- [ ] Dashboard page has an owner filter
- [ ] Agents page has an owner filter
- [ ] Selecting an owner shows only their agents
- [ ] "All" option resets the filter
- [ ] Filter works with existing search/sort functionality

PR #1602, opened 2026-07-14
fix: ephemeral budget gate no longer counts the execution's own pre-created row (#1601)

### LINKED ISSUE #1601: bug: ephemeral budget gate counts the execution's own pre-created row — max_executions=1 ghosts can never run
## Symptom

Live smoke of ghost activation (trinity-enterprise#148, entitled build): a fresh ephemeral agent created with `max_executions: 1` refuses its **first** `/task` with **410 `ephemeral_budget_exhausted`**. Both smoke attempts produced `failed` rows with `ephemeral_exhausted: ghost agent budget spent (budget_exhausted)` — the ghost never executed anything:

```
('CD-MAFwB6...', 'failed', 'manual', 'ephemeral_exhausted: ghost agent budget spent (budget_exhausted)', ...)
```

## Root cause

The `/task` router (`routers/chat.py::execute_parallel_task`) pre-creates the execution row with `status=RUNNING` (`db.create_task_execution`) **before** `execute_task` → `CapacityManager.acquire`. The #69 admission gate then counts `terminal + active` (`count_ephemeral_budget_usage`, active = queued/running/pending_retry) — which includes the execution's **own** pre-created row. For `max_executions=1`: `0 terminal + 1 active(self) ≥ 1` → deny. Every pre-created-row path (task, scheduler) loses the ghost's **final** budgeted execution; `max_executions=1` ghosts are unusable.

The active-count itself is correct in intent (bounds concurrent check-then-act overshoot) — the bug is only that the row being admitted counts against itself.

## Fix

`acquire` already knows the `execution_id` being admitted. Thread it into the gate:

- `count_ephemeral_budget_usage(agent_name, exclude_execution_id=None)` — exclude that row id from the count when given.
- Gate in `capacity_manager.acquire` passes `exclude_execution_id=execution_id`.
- The terminal discard hook keeps the un-excluded count (the just-finished row must count).

Paths that acquire before creating the row are unaffected (the id isn't in the table yet — exclusion is a no-op).

## Acceptance criteria

- [ ] A `max_executions=1` ghost admits and runs its first execution, then hard-discards at the terminal hook
- [ ] A ghost at `terminal ≥ max_executions` still denies admission (410)
- [ ] A concurrent second admission (another active row, budget 1) still denies
- [ ] Unit coverage in `tests/unit/test_69_ephemeral_agents.py`

Found by the live smoke required by trinity-enterprise#148 (activation of #69 / PR #1580).

vybe:
## Summary

- The `/task` and scheduler paths pre-create the execution row with `status=RUNNING` **before** `execute_task` → `CapacityManager.acquire`, so the #69 ephemeral admission gate's `terminal + active` count included the row being admitted: a `max_executions=1` ghost denied its first (only) run with 410 `ephemeral_budget_exhausted`, and every N-budget ghost lost its final run. Found by the live activation smoke required by trinity-enterprise#148.
- Fix: `acquire` already knows the `execution_id` being admitted — thread it through as `count_ephemeral_budget_usage(exclude_execution_id=...)` and exclude that row from the count. Paths that acquire before creating the row are unaffected (the id isn't in the table yet — exclusion is a no-op). The terminal discard hook keeps the un-excluded count (the just-finished row must count).
- Concurrency overshoot bound preserved: two racing admissions for the last budget slot each exclude only their own row, so they still see each other and both deny.

## Changes

- `src/backend/db/agent_settings/ephemeral.py` — optional `exclude_execution_id` on `count_ephemeral_budget_usage`
- `src/backend/services/capacity_manager.py` — gate passes the admitted `execution_id`
- `src/backend/database.py` — facade delegation carries the kwarg
- `tests/unit/test_69_ephemeral_agents.py` — new: DB-layer exclusion (own row / unknown id no-op) + gate-level #1601 scenario (own row admits, racing sibling still denies, id threaded through); existing monkeypatched doubles updated to the new signature

## Test plan

- [x] `tests/unit/test_69_ephemeral_agents.py` — **42 passed** (40 existing + 2 new)
- [x] Adjacent units: `test_capacity_manager.py`, `test_backlog.py` — 101 passed combined
- [x] Live verification on the local stack: pre-fix, a fresh `max_executions=1` ghost 410'd its first `/task`; post-fix the same flow admits, runs (`SMOKE-OK`), and hard-discards via the terminal hook within seconds (container removed, ownership purged, audit `ephemeral_discard` row)

Fixes #1601

### DIFF
--- a/docs/memory/feature-flows/ephemeral-agents.md
+++ b/docs/memory/feature-flows/ephemeral-agents.md
@@ -88,8 +88,12 @@ recorded accepted-risk follow-up.
   `terminal+running+queued ≥ max_executions` raises
   `EphemeralBudgetExhausted` — nothing admitted or enqueued. Covers every
   admission surface through the one facade (`/chat` + `/task` + scheduler +
-  a2a + loop iterations). Fail-open on DB error. Routers map it to **410
-  Gone** (`ephemeral_budget_exhausted`); `execute_task` maps it to a FAILED
+  a2a + loop iterations). The count **excludes the row being admitted**
+  (#1601: `/task` + scheduler pre-create a RUNNING row before `acquire` —
+  counting it denied a `max_executions=1` ghost its first run; racing
+  sibling admissions still see each other, so the overshoot bound holds).
+  Fail-open on DB error. Routers map it to **410 Gone**
+  (`ephemeral_budget_exhausted`); `execute_task` maps it to a FAILED
   row with `TaskExecutionErrorCode.EPHEMERAL_EXHAUSTED`.
 - **Terminal hook**: `_maybe_discard_exhausted_ephemeral` — spawned via
   `_spawn_bg` after a CAS-won terminal in `apply_result` (both branches,
@@ -168,11 +172,12 @@ database.py.
 
 ## Testing
 
-`tests/unit/test_69_ephemeral_agents.py` (40 tests, db_harness — real engine,
+`tests/unit/test_69_ephemeral_agents.py` (42 tests, db_harness — real engine,
 never a wholesale-mocked `database`): column live-select; mixin accessors
 round-trips incl. purge-refuses-durable + cascade read-back; facade
 delegations; acquire-gate deny matrix (expired/exhausted/active-counted,
-fail-open) proven to fire BEFORE slot work; key-fence allow/deny matrix;
+fail-open) proven to fire BEFORE slot work, incl. own-row exclusion at both
+the DB and gate layers (#1601); key-fence allow/deny matrix;
 Part 2 guard matrix (name+key-id); budget hook fail-open + trigger; discard
 full-path + idempotent re-run + half-discarded resume + lock contention +
 audit-row read-back; atomic quota INCR-with-cap + Redis-down fallback.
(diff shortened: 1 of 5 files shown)


dolho:
## Review — LGTM

Traced the logic and the race analysis holds. The root cause (the `/task` + scheduler pre-created `RUNNING` row counting against its own budget in the `terminal+active` gate) is correctly fixed by excluding the admitted `execution_id`, and the overshoot bound is preserved because each racing admission excludes **only its own** row → siblings still see each other → both deny for the last slot. Acquire-before-create paths are a clean no-op (id not yet in the table). Tests cover DB-layer exclusion, the gate-level #1601 scenario, and the racing-sibling deny. CI green.

Non-blocking notes:

1. **Implicit invariant worth pinning down:** the fix assumes `acquire`’s `execution_id` is *the same id* as the pre-created row. If any pre-create path ever calls `acquire` with a mismatched (or `None`) id, the exclusion silently no-ops and the off-by-one returns — with no test to catch it. Holds today; a one-line comment at the pre-create sites (or an assert) would keep it from regressing.
2. `if exclude_execution_id:` treats `""` as “no exclusion”. Fine given ids are non-empty urlsafe tokens — just flagging the truthiness vs `is not None`.
3. The max_executions=1 double-first-run race resolves to **both deny** (safe, no overshoot) rather than one winning. Reads intentional and correct; noting it since the test asserts the sibling denies but not that at least one ultimately runs.

Nice catch surfacing it via the trinity-enterprise#148 activation smoke.

dolho (approved):

Issue #118, opened 2026-03-14
refactor: Split db/schedules.py — extract execution queries (1092 lines)


vybe:
## Problem

`src/backend/db/schedules.py` (1092 lines) mixes schedule CRUD with execution tracking and statistics queries. The `ScheduleOperations` class has 3 functions at CC >= 14:

| Function | CC |
|----------|----|
| `update_schedule` | 18 |
| `_row_to_schedule_execution` | 14 |
| `get_all_agents_execution_stats_dual` | 14 |

## Proposed Solution

Split following the agent mixins pattern (#102):

1. **`db/schedule_crud.py`** — Schedule CRUD (create, read, update, delete, list) ~300 lines
2. **`db/schedule_executions.py`** — Execution record management (create, update status, get by ID) ~400 lines
3. **`db/schedule_stats.py`** — Statistics and aggregation queries ~200 lines
4. **`db/schedules.py`** — Facade class inheriting all mixins ~100 lines

## Acceptance Criteria

- [ ] Each mixin < 400 lines
- [ ] `ScheduleOperations` API unchanged (facade inheritance)
- [ ] All schedule/execution tests pass

## Source

Refactor audit 2026-03-13

oleksandr-korin:
Closing as superseded — file-split refactor tickets created en masse on 2026-03-14 from a one-shot complexity scan. The `/code-health` weekly automation now scores hotspots dynamically and creates better-targeted issues from current churn × complexity, so static "split N-line file" tickets are no longer the right unit of work. The underlying complexity signal will resurface organically if/when it matters. Reference: backlog lean review on 2026-05-28.

PR #1603, opened 2026-07-14
docs(open-core): harden the enterprise-disclosure guard + trim ephemeral packaging framing

vybe:
Public half of the docs-boundary work tracked in **abilityai/trinity-enterprise#151** (follow-up to trinity-enterprise#45). Docs + CI only — no code, no behavior change.

## The rule these PRs land

> **The doc follows the code.** A document lives in the repository that holds the code it describes.

A feature filed in the private tracker whose code shipped in OSS (Brain Orb, Cornelius, the Grid dashboard, the MCP connector) keeps its **public** flow — the tracker never decides the doc's home, and hiding docs for public code protects nothing while leaving contributors without a map. What must never go public is the reconstructable *design*: private-module internals, private `enterprise_*` schema, and the gating rationale.

## 1. The guard was watching the wrong things

`enterprise-docs-guard.yml` matched four keywords (`SIEM|SCIM|2FA|SSO`-ish). It therefore missed **every** private table added since it was written — portal sessions/messages, retention config, SSO providers, user invites.

- **Private schema is now matched generically**: any `enterprise_*` identifier, with an explicit allowlist for the five OSS-legitimate names (`enterprise_features` — the public feature-flag field; `enterprise_schema_migrations` — the two-track runner; `enterprise_connectors*` — an OSS table since #118; `enterprise_docs`; `enterprise_registered`). Generic on purpose: an explicit list cannot see a table that ships after it.
- **Deliberately NOT guarded**: `SSO` (public GitHub-PAT docs legitimately say "SSO authorization required"), `retention`, `audit`, `user management` — each names an OSS subsystem, so guarding them would false-fire on core docs. Documented in-file with the reason.
- **A guard must name what it forbids, and this file is public.** Every token in the pattern is already public elsewhere (`client_portal`, `permissions_matrix`, `SIEM`, `SCIM` ship in the OSS Vue bundle; `two_factor` was already in the old pattern), so the guard adds no new disclosure. `portal_auth` was dropped for exactly this reason. There is an in-file note so the next editor doesn't reintroduce the problem.
- Uses `grep -P` for the allowlist lookahead, and notes that `docs/` is recursive — so `docs/user-docs/**`, which auto-publishes to a public search index via `sync-docs-to-vertex.yml`, is in scope.

## 2. Ephemeral ghost agents — the one real public leak

Its private module is **43 lines of `register_module()`**; every mechanic is OSS code. Yet the flow, `architecture.md`, and the requirements entry all *led* with the entitlement/quota/packaging framing.

Trimmed the packaging narrative; kept the gate as what it actually is — a registry read in `crud.py`, observable in public source either way. The mechanism (admission gate, hard discard, GC, ghost key fence) stays fully documented, because OSS contributors maintain that code.

## 3. Flow index

Neutral pointer that flows for private-submodule code live in that repo. No catalog.

## Verification

```
✅ guard passes on the tree
✅ guard catches a planted `enterprise_portal_sessions` line
✅ workflow YAML parses
```

## Companion PRs

- **trinity-enterprise#152** — private per-module flows (7 modules that had none) + the private user-doc tree
- **trinity-dev** — `/sync-feature-flows`, `/feature-flow-analysis`, `/update-docs`, `/generate-user-docs` route by edition and run this grep before reporting done; `/review` + `/validate-pr` block a PR that trips it; `DEVELOPMENT_WORKFLOW.md` gains a **Docs Routing** section (it routed issues and code, but never docs — which is how this drifted)


dolho:
## Review — LGTM (verified empirically)

Ran the new pattern locally against the tree and some adversarial inputs:

- ✅ **Passes clean** on `docs/ CLAUDE.md` + seam files — no false-fires.
- ✅ **Catches** planted `enterprise_portal_sessions`, `enterprise_user_2fa`, `enterprise_sso_providers`.
- ✅ Allowlisted names (`enterprise_features`, `enterprise_schema_migrations`, `enterprise_connectors_table`, `enterprise_docs`, `enterprise_registered`) correctly ignored.
- ✅ **Prefix-evasion is closed** — the `\b` anchors on the allowlist mean `enterprise_features_secret` and `enterprise_connector_secrets` are still **caught**. That subtlety is the thing an explicit-list guard would miss, so the generic `enterprise_(?!allowlist)` shape is the right call.

The reasoning that every named token is already public (OSS Vue bundle / prior pattern) is sound — the guard adds no disclosure, and dropping `portal_auth` for that reason is the correct instinct.

Non-blocking:
1. `grep -P` needs GNU/PCRE grep — fine on `ubuntu-latest`, just noting the runner dependency.
2. `-i` also lowercases the lookahead, so `Enterprise_Features` stays exempt — verified, good.
3. Future watch-out: excludes are hard-coded (`archive`/`releases`/`security-reports`). A *new* point-in-time doc tree added later wouldn’t be excluded and could false-fire on a legitimately-archived enterprise mention — worth a comment pointing at the git-history-scrub follow-up so the next person knows the boundary.

Doc-follows-the-code framing in the flow index reads well.

dolho (approved):

Issue #1501, opened 2026-07-07
bug: orphan sweep SIGKILLs in-flight brain-orb hook runs — brain-orb/refresh returns 502


vybe:
## Summary

The agent-side brain-orb router runs its `~/.trinity/brain-orb/action` hook as a bare child of the agent-server (`asyncio.create_subprocess_exec`). The #817 orphan sweeper's allowlist does not cover the agent-server's own hook subprocesses, so any hook run still alive when a 30s periodic sweep fires is SIGKILLed mid-run. `POST /api/brain-orb/refresh` (hook timeout 180s — the "fold inbox into KB graph" job) is effectively guaranteed to die on any non-trivial KB, surfacing as a 502 on `POST /api/agents/{name}/brain-orb/refresh` in the UI. Observed in production on a KB agent immediately after deploying `994576a9` (dev).

Notably, the #102 change (merged in #1494) fixed exactly this failure class for post-voice processing by dispatching it as a registered backend execution — its own code comment says *"a hook-forked detached `claude -p` is unregistered and reaped by the orphan sweep"* — but `refresh` was left on the bare-hook path.

## Component

Agent Runtime (base image) — `agent_server` brain-orb router + orphan sweeper allowlist

## Priority

P1 — brain-orb refresh is broken for any agent whose refresh takes longer than the gap to the next sweep (0–30s). Retry does not help; no full workaround (see below).

## Error

Agent container log (timestamps UTC, agent name redacted):

```
09:07:12 WARNING [OrphanSweep] SIGKILL cgroup orphan: pid=1037 pgid=1 cmd=python3 /home/developer/.trinity/brain-orb/action
09:07:12 WARNING [OrphanSweep] SIGKILL cgroup orphan: pid=1038 pgid=1 cmd=<python subprocess spawned by the hook>
09:07:12 WARNING [OrphanSweep] SIGKILL cgroup orphan: pid=1039 pgid=1 cmd=git -C /home/developer log --since=... (spawned by the hook)
09:07:12 INFO  [OrphanSweeper] periodic sweep killed 3 orphan(s) (preserved 0 in-flight execution pid(s))
09:07:12 WARNING agent_server.routers.brain_orb: brain-orb hook action exit -9:
09:07:12 INFO: "POST /api/brain-orb/refresh HTTP/1.1" 502 Bad Gateway
```

Backend then proxies the 502 to the frontend (`POST /api/agents/{name}/brain-orb/refresh` → 502 in the browser console).

## Location

[...]

Issue #389, opened 2026-04-18
Sync health observability (S1)


vybe:
## Summary

Add sync health tracking, dashboard indicators, and auto-sync to catch silent desync before it accumulates.

## Context

**Problem P1**: An agent can run for weeks with failing syncs while appearing "healthy" in the dashboard. No sync-health signal exposed at fleet level, and no autonomous sync cadence.

**Problem P6**: The ahead/behind count compares against \`origin/main\`, not the working branch. External writes to the working branch are invisible.

### Database: New agent_sync_state table
- [ ] \`last_sync_at\`, \`last_sync_status\` (success/failed)
- [ ] \`consecutive_failures\` counter
- [ ] \`last_error_summary\`
- [ ] \`last_remote_sha_main\`, \`last_remote_sha_working\`

### Dashboard indicators
- [ ] Each agent row shows sync-health dot:
  - Green: <24h since sync, no failures
  - Yellow: 24h–7d since sync
  - Red: >7d or last attempt failed or \`behind_working > 0\`

### Operator queue alerts
- [ ] If \`consecutive_failures >= 3\`, emit \`sync_failing\` entry to operator queue
- [ ] Surfaced in Operating Room

### Fix P6: Dual ahead/behind
- [ ] \`/api/git/status\` returns both tuples:
  - \`ahead_main\` / \`behind_main\` (template improvements)
  - \`ahead_working\` / \`behind_working\` (working branch divergence)
- [ ] Health indicator keys off \`behind_working\`

### Auto-sync (S1a)
- [ ] Per-agent flag \`git_auto_sync_enabled\` (default true for GitHub-template agents)
- [ ] Agent server runs lightweight sync heartbeat every 15 min
- [ ] Guarded by \`GIT_SYNC_ENABLED\` env var

### Config flag
- [ ] \`freeze_schedules_if_sync_failing\` (default false, opt-in)
- [ ] If enabled, pauses schedules when sync health is red

## Technical Notes

**Files to modify:**
- \`src/backend/db/schema.py\`, \`db/migrations.py\` — new table
- \`src/backend/services/monitoring_service.py\` — git sync checks
- \`docker/base-image/agent_server/routers/git.py\` — dual ahead/behind, auto-sync loop
- \`src/frontend/src/views/Dashboard.vue\` — sync health dots

Part of Epic #381

PR #1618, opened 2026-07-14
fix(git-sync): survivable git maintenance — auto-gc off, registered guarded repack, bloat alerts (#1595)

### LINKED ISSUE #1595: bug: git auto-gc can never complete inside agent containers — unbounded, silent workspace .git bloat (273GB observed; 97%-garbage repos)

vybe:
## Summary
- **Root cause fixed at the source**: git auto-gc always detaches to PID 1 and the orphan sweep SIGKILLs it — it has never once completed on any git-enabled agent (273GB of workspaces, one 44GB/97%-garbage `.git` observed). The base image now disables it (`gc.auto=0` + `autoDetach` + `maintenance.auto/autoDetach`), making the platform's auto-sync maintenance pass the single owner of repo upkeep.
- **The maintenance pass is now actually survivable**: every auto-sync subprocess is orphan-sweep-registered via the new `agent_server/utils/registered_run.py` seam (#1501 pattern; process-**group** timeout kill), the cycle runs off the event loop (`asyncio.to_thread` — no more `/health`/heartbeat starvation during a 30-min repack), and a non-blocking repo lock replaces the accidental serialization (mutating git endpoints 409 `agent_busy`).
- **Guarded so it can't hurt the host**: pack-OR-loose trigger (with auto-gc off, garbage accumulates loose), 1-hour prune grace (concurrent Claude git ops are never corrupted — replaces `--prune=now`), memory-bounded repack, free-disk preflight, exponential failure backoff, env-tunable budget, per-cycle + startup stale-lock reaping (the 12-day silent `index.lock` freeze class).
- **The failure class is observable now**: `pack_count`/`loose_objects`/`maintenance_failures` flow into `agent_sync_state` (agent-supplied ints coerced at the boundary; dual-track schema + Alembic `0023`) and edge-triggered `git_bloat` operator alerts fire at a 10 GiB ceiling or 3 consecutive maintenance failures.
- **Security fix found in review**: PAT-bearing remote URLs are now redacted from persisted git stderr at both sinks (sync-state error summary + push-collision operator-queue entry) — this exposure predates the branch (since #389). CSO diff report included.

## Changes
- `docker/base-image/`: Dockerfile (system gitconfig), startup.sh (lock reap), `agent_server/utils/registered_run.py` (new), `agent_server/routers/git.py` (maintenance + lock + hygiene + redaction), `agent_server/auto_sync.py` (to_thread)
- `src/backend/`: `services/git_service.py` (config mirror), `services/sync_health_service.py` (coercion + alerts), `db/{schema,tables,migrations,sync_state}.py` + `migrations/versions/0023_agent_sync_state_gc_signals.py`
- Docs: architecture.md, requirements/github.md §11.9 (maintenance-ownership contract + named residuals), feature-flows/git-sync-health.md, learnings.md, CSO diff report
- Tests: `tests/unit/test_1595_git_maintenance.py` (37), `tests/unit/test_1595_sync_health_signals.py` (8)

## Test Plan
- [x] New tests pass: `pytest tests/unit/test_1595_git_maintenance.py tests/unit/test_1595_sync_health_signals.py -v`
- [x] Adjacent suites unaffected: 137 passed across orphan-sweep/git/process-registry/sync-health/schema-parity suites; both module orders verified
- [x] Alembic revision chains off `0022`; SQLite + PG tracks both present (schema-parity + pg-migrations CI)
- [ ] Post-merge manual: rebuild base image, recreate one git-enabled agent, confirm first auto-sync cycle reports `pack_count`/`loose_objects` in `GET /api/agents/sync-health`

## Rollout & ops notes
- Ships in the base image → fleet needs `build-base-image.sh` + agent recreate.
- eu2 recovery of pre-existing bloat is ops-side (Abilityai/trinity-ops-agent#127) — use `GIT_MAINTENANCE_TIMEOUT_SECONDS` for the first pass.
- **Ops advisory**: instances with failed pushes before this fix may hold PATs in retained `operator_queue` collision rows (≤90d) — one-time scrub + opportunistic rotation recommended on shared fleets.
- Residuals (documented in flow doc): sub-repos unmaintained; history growth → squash/geometric follow-up; `/api/git/status` loop blocking → evidence filed on #1505.

Fixes #1595


dolho (commented):
## Review — #1595 survivable git maintenance

Read the full diff end-to-end (agent-server maintenance path, backend health signals, schema/migration parity, docs, CSO report). This is a strong, well-reasoned change: root cause (auto-gc detaches → orphan-swept → never completes) is fixed at the source, the maintenance pass is now off-loop + sweep-registered + repo-locked, and the whole failure class is finally observable + alertable. Approving in spirit; notes below are minor.

### ✅ Correctness spot-checks that hold up
- **Failure counter actually increments** — `_maybe_run_git_maintenance` returns `"failed"` on the exception path (not `None`), so `_write_sync_state_file` bumps `maintenance_failures` and arms the exponential backoff. The `"backoff"`/`"skipped_low_disk"` skip-statuses correctly record without touching the counter, so the streak survives across cycles and the 3-failure alert fires once per episode.
- **Edge-triggered alerts fire once** — `prior_maintenance_failures < THRESHOLD <= new` and the `git_dir_bytes` ceiling both compare prior-DB-row → new, mirroring `sync_failing`. A `0` reset re-arms (`_merged` treats `0` as a value, only `None` falls back).
- **Boundary coercion** — `_coerce_nonneg_int` rejects `bool` (int subclass) and out-of-range, matching the `/health clone_status` posture for agent-writable JSON. Good.
- **No lock nesting / deadlock** — auto-sync (`_run_auto_sync_once`) and the `@_with_repo_lock` operator endpoints both use *non-blocking* `acquire()`; a `threading.Lock` shared across the event-loop thread and the `to_thread` worker is fine here because neither ever blocks on it. Contention → skip (cycle) or 409 `agent_busy` (endpoint).
- **Timeout contract** — `run_registered` raises `TimeoutExpired` (not `CalledProcessError`) on the process-group kill; every caller's inner `except Exception` catches it and writes a `failed` state, so a swept repack never silently reads as success.
- **Schema/migration parity** — SQLite `agent_sync_state_gc_signals` + Alembic `0023` (chains off `0022`, now on dev) + `schema.py`/`tables.py` all carry the three columns. `ADD COLUMN IF NOT EXISTS` keeps fresh-PG baseline a no-op.

### 🔸 Minor / for your call (non-blocking)
1. **PAT redaction scope.** `_redact_url_userinfo` covers the two *persisted* sinks (sync-state error summary + push-collision operator-queue row). The `sync_to_github` / `pull_from_github` endpoint responses still return raw `stderr` to the (authenticated) operator, and any `logger.*(stderr)` debug lines aren't redacted. Lower risk (auth'd caller / masked-by-Vector), and the CSO report calls it accepted — just confirming that's the intended boundary rather than an oversight.
2. **`index.lock` 1h reap vs concurrent Claude git.** Claude executions run git *outside* `_REPO_LOCK`, so the reaper's age gate (>1h) is what protects an in-flight `git add`. Reasonable (no single git op holds it that long) and well-documented — flagging only because it's the one place the repo lock doesn't cover.
3. **Whole-file `sys.modules` lint exemption (my CI fix).** The `_STUBBED_MODULE_NAMES` + `_restore_sys_modules` pattern exempts the *entire* file from the pollution linter (per the sanctioned precedent), so future `sys.modules` mutations in these two test files won't be caught. Acceptable tradeoff given the import-time base-image loader, but noting it.

### CI
The `lint (sys.modules pollution check)` job was failing — the two new #1595 test suites do import-time `sys.modules` eviction that the linter bans. Pushed a fix (`b42e6ae`) adopting the sanctioned `_STUBBED_MODULE_NAMES` + autouse `_restore_sys_modules` snapshot/restore pattern. Lint is now green; both suites still pass locally (45/45). Rest of CI re-running.

dolho (approved):

PR #1625, opened 2026-07-15
fix(telemetry): cold container-stats cache reports no_data, not running_count 0 (#1617)

### LINKED ISSUE #1617: bug: /api/telemetry/containers cold cache reports running_count 0 — indistinguishable from "no containers running", misleads diagnostics
## Summary

`GET /api/telemetry/containers` uses a stale-while-revalidate, **per-uvicorn-worker in-process cache**. A worker whose cache is cold returns `{"running_count": 0, ..., "stale": true, "cache_age_seconds": null}` — the same shape as a healthy answer of "zero containers running". Because the cache is per worker, consecutive requests can flip between `running_count: 0` and the true count depending on which worker serves them, and a rarely-queried instance can serve data that is hours stale (observed: `cache_age_seconds: 5726`). In a real incident this was read as "the backend lost its Docker connection" and drove a wrong root-cause report from a client's diagnostic tooling.

## Component

Backend / Telemetry Router

## Priority

P3 (diagnostic quality; no functional breakage)

## Error

```
GET /api/telemetry/containers  (worker A, cold cache)
{"running_count": 0, "containers": [], "cached": false, "stale": true, "cache_age_seconds": null}

GET /api/telemetry/containers  (worker B, seconds later, 3 agent containers running)
{"running_count": 3, "containers": [...], "cached": true, "stale": true, "cache_age_seconds": 43.0}
```

## Location

- **File**: `src/backend/routers/telemetry.py`
- **Line**: ~269–292 (cold-cache path returns `_empty_payload(0)`; per-process cache)

## Root Cause

The cold-cache branch returns `_empty_payload(0)`, i.e. a literal `running_count: 0`, plus `stale: true`. Consumers cannot distinguish "no data yet" from "zero containers". Compounding it: the cache lives in each worker process, so freshness varies per request, and a worker that never receives a request keeps a cold cache indefinitely (stale-while-revalidate only refreshes on demand).

## Reproduction Steps

1. Run the backend with multiple uvicorn workers and several agent containers running
2. Restart the backend (all caches cold)
3. Immediately `GET /api/telemetry/containers` a few times in a row
4. Observe some responses with `running_count: 0` / `cache_age_seconds: null` and others with the true count, depending on the worker hit

## Suggested Fix

- Cold cache: return `running_count: null` (and/or an explicit `"no_data": true`) instead of `0`, so consumers can't mistake "not yet measured" for "nothing running"
- Consider moving the cache to Redis (shared across workers) so responses are consistent regardless of serving worker
- Document that `stale: true` refers to cache freshness, not Docker connectivity

## Environment

- Trinity version: `a88f5fc8`
- Docker prod compose, multi-worker uvicorn

## Related

- `src/backend/routers/telemetry.py`
- Surfaced during the same investigation as #1615

dolho:
## Summary

`GET /api/telemetry/containers` serves a per-uvicorn-worker, stale-while-revalidate **in-process** cache. The cold-cache branch returned `_empty_payload(0)` — a literal `running_count: 0` with `stale: true` — which is **byte-identical to a healthy "zero containers running"** answer. Because the cache is per worker, consecutive requests could flip between `0` and the true count depending on which worker served them, and a rarely-queried worker kept a cold cache indefinitely (observed `cache_age_seconds: 5726`). In a real incident this read as "the backend lost its Docker connection" and drove a wrong root-cause report (#1617).

## Fix

Make the cold state **self-describing** so it can't be mistaken for a real zero:

- **Cold cache** → `running_count: null` + `no_data: true`.
- **Genuinely-empty fleet** (refresh ran, nothing running) → `running_count: 0` + `no_data: false` (unchanged).
- Fresh + stale branches also carry `no_data: false` for a consistent contract.
- Docstring now states `stale` = **cache freshness only**, not Docker connectivity (unreachable Docker is the existing HTTP 503).

Directly implements suggested fix 1 (`running_count: null` / explicit `no_data`) and #3 (document `stale` semantics) from the issue.

**Deferred (issue's optional "consider"):** moving the cache to Redis for cross-worker consistency. The misleading-diagnostics defect is fully closed by making cold state self-describing; a shared-cache rework is a larger, separate change and out of scope for a P3 diagnostic-quality fix.

## Consumer safety

No in-repo consumer reads `/api/telemetry/containers` — the frontend uses only `/api/telemetry/host`; the `running_count` references in `OverviewPanel`/`network.js`/`executions.js` come from `/api/executions/stats` (a different endpoint/model). The endpoint's only known consumer is external diagnostic tooling (per the issue), so widening the contract with `no_data` + a nullable `running_count` is backward-additive.

## Changes

- `src/backend/routers/telemetry.py`: cold branch → `running_count: null` + `no_data: true`; `no_data: false` on fresh/stale; docstring clarifies `stale` semantics.
- `tests/unit/test_1096_telemetry_cache.py`: update cold-start expectation to the new shape; new `test_cold_no_data_is_distinct_from_real_zero_running` proving cold≠real-zero.
- `tests/test_telemetry.py`: live-API tests poll past the cold window (skip/tolerate `no_data`).

## Test Plan

- [x] `pytest tests/unit/test_1096_telemetry_cache.py -q` → 12 passed
- [x] Router + test modules parse clean
- [ ] Live-API `tests/test_telemetry.py::TestContainerTelemetry` (test-runner suite, needs a running stack)

Fixes #1617

### DIFF
--- a/src/backend/routers/telemetry.py
+++ b/src/backend/routers/telemetry.py
@@ -265,8 +265,19 @@ async def get_container_stats(current_user: User = Depends(get_current_user)):
     Non-blocking (#1096): served from a short-TTL cache that is refreshed by a
     background task. The request never waits on the Docker daemon, so a fleet
     of agents can no longer pin a uvicorn worker for ~6s and stall the UI.
-    Adds three fields on top of the original contract: `cached` (bool),
-    `stale` (bool), and `cache_age_seconds` (float|None).
+    Adds four fields on top of the original contract: `cached` (bool),
+    `stale` (bool), `cache_age_seconds` (float|None), and `no_data` (bool).
+
+    Field semantics (#1617):
+    - `stale` is about CACHE FRESHNESS ONLY — the cached payload is older than
+      the TTL and a refresh was scheduled. It does NOT mean Docker is
+      unreachable (that surfaces as HTTP 503 below).
+    - `no_data` is True only on a cold cache (this worker has never completed a
+      refresh); in that case `running_count` is `None`, NOT `0`. A cold cache
+      per uvicorn worker previously returned `running_count: 0`, which is
+      indistinguishable from a healthy "zero containers running" and misled
+      diagnostics. `no_data: True` / `running_count: None` lets consumers tell
+      "not yet measured" from a real zero.
     """
     if not docker_client:
         raise HTTPException(status_code=503, detail="Docker not available")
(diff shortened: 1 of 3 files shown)


vybe (approved):
Validated via /validate-pr.

Contract change is safe: I grepped the whole repo — nothing consumes `/api/telemetry/containers` outside the tests. The frontend `running_count` hits (stores/executions.js, network.js, OverviewPanel.vue) all read `/api/executions/stats`, a different endpoint. So making `running_count` nullable breaks no consumer.

- `no_data: true` / `running_count: null` correctly distinguishes 'not yet measured' from a real zero; `test_cold_no_data_is_distinct_from_real_zero_running` proves both sides of that.
- Named regression test present for the reported issue.

Security scan clean.

Two notes, neither blocking:
1. Body edited — it read 'suggested fixes #1', which the issue-status-on-merge workflow regex (`/\b(?:fix(?:e[sd])?|close[sd]?|resolve[sd]?)\s+#(\d+)/gi`) matched as a closing ref to **#1** (a long-merged PR), and would have stamped status-in-dev on it while leaving #1617 stranded in status-in-progress. Now reads 'suggested fix 1' + `Fixes #1617`.
2. Follow-up: `docs/memory/feature-flows/host-telemetry.md:146` documents this endpoint's response schema and still shows `running_count` as always-int with no `no_data` field. Not merge-blocking (bug-fix tier), but the flow doc is now stale.

Issue #737, opened 2026-05-08
feat: serve A2A Agent Cards from template.yaml for external discoverability


vybe:
## Summary

Generate and serve an A2A-compliant Agent Card (\`/.well-known/agent-card.json\`) for each Trinity agent by reading \`template.yaml\` skills and \`CLAUDE.md\`. External A2A orchestrators (AWS Bedrock, Azure Copilot, Google ADK) can discover and call Trinity agents without knowing Trinity's internal API. No agent-server changes required — the backend generates and proxies the card.

## Context

An Agent Card is the A2A protocol's discovery primitive: a JSON document at \`/.well-known/agent-card.json\` that describes an agent's capabilities, skills, and auth requirements. Trinity's \`template.yaml\` already contains the raw material (name, description, skills, credential requirements). Serving a card from that data makes Trinity agents first-class citizens in the A2A ecosystem with zero agent-code changes. See \`docs/research/a2a-protocol.md\` § 2.2 and § 13.2 Option A/B.

## Acceptance Criteria

- [ ] New backend endpoint \`GET /api/agents/{name}/a2a/agent-card\` returns a valid A2A v1.0 Agent Card JSON
- [ ] Card fields populated from: \`template.yaml\` (name, description, version, skills array) + agent ownership (capabilities)
- [ ] Backend proxies \`GET /.well-known/agent-card.json\` for requests with \`X-Trinity-Agent: {name}\` header (or via subdomain / path convention TBD)
- [ ] \`skills[]\` in card populated from \`template.yaml\` skills list; each skill includes \`id\`, \`name\`, \`description\`, \`tags\`, \`examples\` where available
- [ ] \`securitySchemes\` declares \`bearerAuth\` (Trinity MCP API key); public card omits sensitive details
- [ ] Extended card (requires auth) adds internal endpoint URLs and full skill schemas
- [ ] \`capabilities.streaming: true\` set for all agents (agent-server supports SSE)
- [ ] Card cached in Redis with 5-min TTL, invalidated on agent update/restart
- [ ] New MCP tool \`get_agent_card(agent_name)\` returns the card JSON for introspection
- [ ] Integration test: fetch card, validate against A2A v1.0 schema

## Technical Notes

- A2A Agent Card schema: \`docs/research/a2a-protocol.md\` § 2.2
- \`template.yaml\` skills array → A2A \`skills[]\`: \`id\`, \`name\`, \`description\`, \`tags\`, \`examples\`
[...]

Issue #160, opened 2026-03-24
bug: Headless task returns empty response for skills with context: fork


vybe:
## Summary

Scheduled and manual headless tasks that invoke skills with `context: fork` in their frontmatter consistently fail with "Task returned empty response". The skill's forked context does its work but the parent Claude Code process exits with code 0 and no text in `response_parts`, triggering the empty response check. This has caused 8 consecutive daily failures for a specific agent since March 17.

## Component

Agent Runtime / `agent_server/services/claude_code.py` (headless task execution)

## Priority

P2 — Scheduled skill invocations silently fail daily; workaround is to invoke via MCP chat instead of headless task.

## Error

```
[TaskExecService] Failed to execute task on agent: Task returned empty response
```

Agent container returns HTTP 500:
```json
{"detail": "Task returned empty response"}
```

Execution records show: `claude_session_id: null`, `cost: null`, `context_used: null` — indicating the stream-json output from Claude Code contained no `result` message with text content.

## Location

- **File**: `docker/base-image/agent_server/services/claude_code.py`
- **Line**: ~996-999
- **Function**: `execute_headless_task()`

## Root Cause

When a headless task message triggers a skill with `context: fork` (e.g., `/synthesize-learnings`), Claude Code forks the conversation context to a sub-process. The forked process may do all the work (file reads, writes, tool calls), but the parent process that Trinity is watching via `--output-format stream-json` exits with code 0 and an empty `result` field. The current code treats any empty `response_parts` as a hard failure:

```python
response_text = "\n".join(response_parts) if response_parts else ""
response_text = sanitize_text(response_text)

if not response_text:
    raise HTTPException(
        status_code=500,
        detail="Task returned empty response"
    )
```

This check doesn't account for `context: fork` skills where the meaningful work happens in the forked context, not the parent stream.

**Timeline**: Last successful execution was March 16. First failure March 17. Commits `c6af307` (default model to sonnet for subscription compatibility) and `e47a9f8` (create execution records for all /api/chat calls) landed between the last success and first failure, possibly changing how model selection or execution tracking interacts with forked contexts.

## Reproduction Steps

1. Create an agent with a skill that has `context: fork` in its frontmatter
2. Schedule or manually trigger that skill via headless task execution (UI chat or scheduler)
3. Observe the execution fails with "Task returned empty response"
4. Note: invoking the same skill via MCP `chat_with_agent` works fine (different code path)

## Evidence

[...]

Issue #157, opened 2026-03-23
spike: Agent Task Backlog — async task queue for inter-agent pipelines


vybe:
## Overview

Today, agent-to-agent communication in Trinity is **synchronous** — the caller blocks until the target agent responds. The `chat_with_agent` MCP tool has an `async: true` flag that returns an `execution_id`, but there is **no way to retrieve the result later**. This means agents cannot submit work to each other and continue independently.

This spike proposes an **Agent Task Backlog** — a persistent, ordered queue of tasks per agent that any actor (human, agent, schedule) can submit to, with results retrievable by execution ID.

---

## Problem

1. **Blocking calls limit throughput** — Agent A calls Agent B and waits 2+ minutes for a response. A can't do anything else.
2. **No shared work queue** — Humans can trigger tasks from the UI, but there's no concept of a persistent backlog that agents also submit to.
3. **No result retrieval** — The `async: true` mode returns an execution_id, but no MCP tool exists to fetch the result. The async path is a dead end.
4. **No pipeline pattern** — Without submit + retrieve, agents can't form asynchronous pipelines (A submits to B, B submits to C, A checks results later).

---

## Current State

| Component | What Exists | Gap |
|-----------|-------------|-----|
| **Execution Queue** (`services/execution_queue.py`) | Redis-backed, max 3 queued, 600s TTL | Ephemeral, not a deliberate backlog |
| **`chat_with_agent` MCP tool** | `async: true` returns execution_id | No `get_execution_result` tool |
| **`/api/agents/{name}/task`** | Stateless parallel execution | No queue semantics, no persistence |
| **`schedule_executions` table** | Stores execution results with IDs, status, response, cost | Only for schedule-triggered executions |
| **Process Engine** | Multi-step YAML workflows | Heavyweight for ad-hoc task submission |
| **Tasks UI panel** | Manual task triggering | No backlog view, no queue status |

---

### New Concept: Agent Task Backlog

A persistent, per-agent queue where tasks are submitted and processed FIFO (with optional priority). Every actor uses the same queue.

### Backend Changes

1. **`POST /api/agents/{name}/backlog`** — Submit a task, returns `execution_id` immediately
   - Accepts: `message`, `priority` (optional), `metadata` (optional)
   - Persists to database (not just Redis)
   - Source tracking: who submitted (human user, agent, schedule)

2. **Backlog Worker** — Background loop per agent
   - Picks next pending task from backlog (FIFO or priority-ordered)
[...]

PR #1628, opened 2026-07-15
feat(a2a): A2A interoperability — inbound server (well-known + JSON-RPC/SSE) + control MCP tools (ent#157/#160)

dolho:
## Summary

The complete public surface for **A2A (agent-to-agent) protocol interoperability** — one feature, one PR. Two halves:

1. **A2A inbound server (ent#157)** — make an *exposed* Trinity agent genuinely reachable + taskable over the open A2A protocol by an external orchestrator (Google ADK, LangChain, AWS Bedrock AgentCore, another Trinity).
2. **A2A control MCP tools (ent#160)** — drive the exposure/config plane (toggle exposure, read the card, manage the inbound allow-list, register outbound endpoints) over MCP.

The **entitlement-gated enterprise backend** ships in a separate repo (the private submodule): **trinity-enterprise#161**. It's the *writer/management* half; this PR is the *reader/server* + *MCP* half. (The submodule is a distinct git repo, so it can't share a PR with the public code.)

## A2A inbound server (ent#157)

**OSS `a2a_exposed` primitive** (edition-agnostic, default OFF — safe by default):
- `agent_ownership.a2a_exposed` column (schema.py + tables.py + two-track migration: `db/migrations.py` + Alembic `0024`) + `A2AExposureMixin` accessors + surfaced on `GET /api/agents` (mirrors `mcp_exposed`). OSS owns the column + read/enforcement; the **write** is entitlement-gated by the enterprise setter (core-primitive + enterprise-knob, like `users.suspended_at`).

**Public server** (`routers/a2a.py` → new `a2a_server_router` in `main.py`):
- `GET /a2a/{name}/.well-known/agent-card.json` — unauthenticated discovery; **uniform 404** for non-exposed/non-existent (no enumeration oracle). Safe by default: the flag can't be set without the entitled setter → OSS-only builds always 404.
- `POST /a2a/{name}` — **JSON-RPC 2.0**: `message/send` (sync → `execute_task(triggered_by="a2a")`), `message/stream` (SSE), `tasks/get`, `tasks/cancel`. JSON-RPC error codes (`-32700/-32600/-32601/-32602`) + A2A task-not-found (`-32001`). **Bearer = a Trinity MCP key**, per-call `get_current_user` (fail-closed 401). Owner/shared + per-agent inbound **allow-list** via the new `services/a2a_gate` seam (OSS no-op; enterprise provider; fail-open). **Trigger-boundary idempotency** on `messageId` (Invariant #18). Every task **audit-logged** (`source="a2a"`). `taskId == execution_id`.
- **Card honesty**: `protocolVersion` `"0.3.0"`, `url` → the JSON-RPC endpoint, `preferredTransport=JSONRPC`, Bearer scheme.
- Phased (seams ready): `tasks/resubscribe` (`-32004`), incremental token streaming.

## A2A control MCP tools (ent#160)

`src/mcp-server/src/tools/a2a.ts` — 7 tools: `get_agent_a2a_config`, `set_agent_a2a_exposure`, `get_agent_a2a_card` (proxies the OSS #737 served card), `set_a2a_inbound_allowlist`, `register_a2a_endpoint`, `list_a2a_endpoints`, `remove_a2a_endpoint`. Honest gating (unentitled-403 / OSS-404 → structured `not_entitled`/`not_found`, never silent success); mutations owner/admin + human-only; outbound credentials write-only (never echoed). `a2a_exposed` flows natively on `list_agents`/`get_agent` via the OSS column (no extra fetch).

## Test Plan

- [x] Backend: `tests/unit/test_157_a2a_inbound_server.py` (16) + `test_a2a_card_service.py` updated for the honest card. Schema/Alembic parity guards green; pg-migrations, backend-boots-without-enterprise, non-root all green.
- [x] MCP: `src/mcp-server/src/tools/a2a.test.ts` (10); full suite **100 pass**, `tsc` clean.
- [x] regression-diff green (updated the two #737 card assertions).

## Companion PR

**trinity-enterprise#161** — the entitlement-gated `enterprise.backend.a2a` module: the exposure **setter** (write-throughs to this OSS column — without it the flag stays OFF and every public route 404s, the intended safe default), curated-skills + inbound allow-list + outbound endpoint registry (AES-256-GCM creds), and the allow-list provider it plugs into `a2a_gate`.

Related to abilityai/trinity-enterprise#157, abilityai/trinity-enterprise#158, abilityai/trinity-enterprise#160, abilityai/trinity-enterprise#180

(Fully qualified `owner/repo#N` so the cross-repo links actually resolve — `trinity-enterprise#157` without the owner does not. Deliberately **not** a closing keyword: the convention here is relate-don't-close, and the private issues get `status-in-dev` set manually at merge since the status automation is same-repo only.)

---
**Update — A2A config UI (ent#158) added.** An owner-only **A2A tab** on Agent Detail (`components/A2aPanel.vue`), gated on the `a2a` entitlement (never a blank tab in OSS): exposure toggle (→ the OSS `a2a_exposed` write-through), one-click-copy Agent Card URL, advertised skills (read-only), inbound allow-list add/remove, outbound endpoint registry (credentials write-only, never echoed). No new backend endpoints — proxies the surface above. This PR now covers the full public feature: **inbound server (#157) + control MCP tools (#160) + config UI (#158)**.


dolho:
## 📖 A2A user guide + screenshots

Full guide added at [`docs/user-docs/integrations/a2a-protocol.md`](docs/user-docs/integrations/a2a-protocol.md). Reproduced here for review, with the live UI screenshots.

Trinity speaks the open **[A2A protocol](https://a2a-protocol.org)** — an exposed agent is reachable by any A2A orchestrator (Google ADK, LangChain, Bedrock, another Trinity), and can call external A2A agents in turn.

> **Availability:** exposing over A2A requires the A2A capability to be entitled. Otherwise the tab is hidden and public routes 404 — off + invisible by default.

---

### Part 1 — Expose an agent (the new **A2A** tab)

A2A config panel

1. **Flip "Expose over A2A" on** (OFF by default; while off the public routes 404).
2. **Copy the Agent Card URL** — the public discovery URL, one click.
3. **Review advertised skills** (from `template.yaml`).
4. **Inbound allow-list** — restrict which caller identities may task the agent (empty = any owner/shared caller).
5. **Outbound endpoints** — external A2A endpoints the agent may call; credentials stored encrypted, never echoed.

Agent Card URL section

Same controls are available as MCP tools (`get_agent_a2a_config`, `set_agent_a2a_exposure`, `set_a2a_inbound_allowlist`, `register_a2a_endpoint`, …). Exposure + credential ops are owner/admin and **human-only**.

---

### Part 2 — Consume an exposed agent (external orchestrator)

**1. Discover** — fetch the card:
```bash
curl -s http://localhost:8001/a2a/new_cool_agent/.well-known/agent-card.json | jq
# → { "protocolVersion":"0.3.0", "url":".../a2a/new_cool_agent",
#     "preferredTransport":"JSONRPC", "securitySchemes":{"bearerAuth":{...}} }
```

**2. Authenticate** — a Trinity MCP API key (Settings → MCP Keys) as `Authorization: Bearer trinity_mcp_…`. Unauth → `401`.

**3. Task** — JSON-RPC 2.0:
```bash
curl -s -X POST http://localhost:8001/a2a/new_cool_agent \
  -H "Authorization: Bearer trinity_mcp_YOUR_KEY" \
  -d '{"jsonrpc":"2.0","id":1,"method":"message/send",
       "params":{"message":{"messageId":"req-1",
         "parts":[{"kind":"text","text":"Summarize today's sales."}]}}}'
# → Task { status:{state:"completed"}, artifacts:[{parts:[{text:"…"}]}] }
```
- `message/stream` for SSE (`working` → final `completed`); `tasks/get` / `tasks/cancel` take `{"id":"<taskId>"}`.
- Re-sending the same `messageId` returns the original task **without re-executing** (idempotent).

---

### Notes
- **Safe by default** — exposure OFF per agent; non-exposed/non-existent → uniform `404` (no enumeration).
- **Front door** — external clients hit your public URL; Trinity proxies `/a2a/` to the backend (nginx prod / dev proxy). Set `PUBLIC_CHAT_URL` so the card's `url` is externally reachable.
- Every inbound task is audit-logged (`source=a2a`).

**JSON-RPC error codes:** `-32700` parse · `-32600` bad envelope · `-32601` method not found · `-32602` invalid params · `-32001` task not found.

vybe (changes requested):
Validated via `/validate-pr`. Impressive scope and the fundamentals are solid — dual-track migrations are textbook (all four surfaces present, `0024` chains to the verified sole head `0023`, matches the `0010_mcp_exposed` house pattern byte-for-byte), Invariants #4/#14 clean, SQL is all bound-param Core, and the JSON-RPC path is fail-closed 401 with uniform 404s. The negative-path test coverage is genuinely good.

Three findings I'd want resolved before this lands. All verified against the code, not inferred.

---

### 1. CRITICAL — cross-caller response disclosure via `messageId` idempotency

`routers/a2a.py`:
```python
message_id = message.get("messageId")
decision = idempotency_service.begin(f"a2a:{agent_name}", message_id)
if decision.replay and not decision.in_flight and decision.snapshot:
    return _rpc_result(rpc_id, decision.snapshot)
```

The scope carries **no caller identity**, the key is a **peer-controlled protocol field**, and the stored snapshot contains the agent's full response text (`artifacts[0].parts[0].text`).

**Failure:** users A and B both have shared access to agent `bot`. A sends `messageId: "req-1"`. B sends `messageId: "req-1"` → **B receives A's agent output**, and B's own task silently never executes.

This isn't hypothetical: A2A `messageId` is mandatory, auto-generated by SDKs, and the spec only requires *per-client* uniqueness. This PR's own guide uses `"messageId": "req-1"` as the worked example.

I know `make_agent_scope(name)` + caller-supplied key is the house shape (`chat.py:205`, `fan_out.py:63`) — the difference is `Idempotency-Key` there is opt-in and deliberately chosen, while `messageId` is automatic and collision-prone. Suggest folding the caller principal into the scope.

### 2. CRITICAL — unauthenticated well-known route has no rate limit

`routers/a2a.py` — `a2a_well_known_card` → `_serve_card`

Per request against an exposed agent: DB read → **live uncached Docker API call** (`docker_service.get_agent_container`) → **HTTP call into the agent container** (`_fetch_template_data`, 5s timeout). No throttle at either layer — `services/rate_limiter` isn't imported in the diff, and `nginx.conf` has no `limit_req`/`limit_conn`.

Every other unauthenticated route in the repo carries a per-IP limit: `public.py:50` (`PUBLIC_LINK_LOOKUP_RATE_LIMIT = 60`, annotated "pentest 3.3.2"), `files.py:44` (`_DOWNLOAD_RATE_LIMIT = 60`), `webhooks.py:99` (added by #1424 for exactly this endpoint shape).

The card URL is **published by design**, so the agent name is public. An unauthenticated flood saturates the Docker API (called synchronously inside `async def`, so it stalls the event loop) and hammers the agent container → fleet-wide degradation from one URL.

Worth noting this PR is what *makes* the route public — the docs it replaces said "Trinity does **not** serve that route yet — it requires a separate access-policy decision per agent." That decision seems to have been made implicitly; the per-IP limit is the piece that didn't come with it.

### 3. CRITICAL — public seam file names a private enterprise table (standing rule; CI green by accident)

`services/a2a_gate.py:13`:
> `* Enterprise build → the provider consults ``enterprise_a2a_inbound_allowlist``:`

`enterprise_a2a_inbound_allowlist` matches the guard's own pattern and is not allowlisted. **The guard won't catch it** — `.github/workflows/enterprise-docs-guard.yml` greps `docs/** CLAUDE.md` plus `SEAM_FILES='src/backend/main.py src/backend/services/entitlement_service.py'`. `a2a_gate.py` is a *new* open-core seam the stale list doesn't cover.

This is exactly the #1461 class the guard's own header calls out: "a comment there is unguarded prose that can name the paid catalog just as a doc can." Fix is one comment edit (describe the mechanism, not the table) — and please add `a2a_gate.py` to `SEAM_FILES` so the next one is caught.

The `docs/` side is clean.

---

### Warnings

4. **`tasks/cancel` reports success unconditionally.** `terminate_execution_on_agent(...)` returns a `bool` that's discarded; the handler always returns `"canceled"`. No `status=CANCELLED` write, no slot release, no CAS precondition (contrast #1082 status-as-projection, and `chat.py:2408`'s dedicated `db.cancel_queued_execution` branch). Cancelling a **queued** execution → registry 404s → helper returns True → client told "canceled" → the task drains minutes later, runs, bills, and performs side effects the caller believed cancelled.

5. **SSE disconnect wedges the `messageId` for 24h.** `_gen()` wraps `_run_a2a_task` in `except Exception`, but a client disconnect raises `asyncio.CancelledError` — a `BaseException` since 3.8, so uncaught. Neither `complete()` nor `fail()` runs → row stays `in_flight` → every retry with that `messageId` gets `-32603 "already in progress"` for 24h.

6. **Allow-list identity contract mismatch.** `caller_identity = getattr(current_user, "email", None) or current_user.username` — always email/username. But `A2aPanel.vue` placeholders `"did:example, https://caller/…, or key id"` and the guide says "a caller URL, client id, or key id". An operator entering a DID makes the list non-empty → every real caller (an email) is denied, or they think they've restricted access when they haven't.

7. **The access gate has zero coverage.** `test_157_a2a_inbound_server.py` defines `state["access"] = {"bot"}` but no test ever varies it — the exposed-but-inaccessible → uniform 404 branch is never exercised.

8. **Allow-list fails open; docs present it as a hard boundary.** `check_inbound_allowed` returns `True` on provider exception (deliberate, documented in code) — but the guide says "only listed identities are accepted; everyone else gets 403" with no mention that a policy-backend outage admits every owner/shared caller.

### Suggestions

- `await request.json()` is uncapped (contrast #1083's callback and #1424's webhook, which caps *before* parsing). Mitigated by nginx `client_max_body_size 25m`, but not if `:8000` is reachable directly.
- `message/stream` replay returns JSON, not SSE — a streaming SDK breaks on replay.
- `_a2a_state_for` maps everything non-`success` → `failed`, so a CANCELLED execution reports `failed`, not `canceled`.
- `architecture.md` not updated: line 784 still says "A2A v1.0" (now `0.3.0`); the `agent_ownership` block lists `mcp_exposed` but not `a2a_exposed`; the two new public routes, `services/a2a_gate.py`, and `triggered_by="a2a"` are undocumented. Only `docs/user-docs/` was touched.

### Merge ordering

Companion enterprise PR abilityai/trinity-enterprise#161 imports `from services import a2a_gate` and calls `get_a2a_exposed`/`set_a2a_exposed` — none of which exist in OSS until **this** PR lands. I verified the API contracts match exactly on both sides, so it's purely sequencing: **#1628 → dev first, then #161**, and any submodule-pointer bump must land on a branch that already contains this PR.

dolho:
Re-validated via `/validate-pr` after the fix commit (`3cff879c`).

> **Disclosure:** I authored `3cff879c`, the commit this validates. A self-review is structurally weaker than an independent one — the three criticals below are confirmed by live behaviour and by tests I checked fail against the unfixed code, but the *judgement* that they're adequately fixed is not independent. Weight it accordingly.

## The three criticals are resolved — verified live, not just by tests

The backend bind-mounts `src/backend` into `/app`, so this branch is what the running stack serves. All checks below are against the live instance on PostgreSQL (Alembic at `0024`, applied cleanly from `0023`).

**1. Cross-caller disclosure via `messageId`** — the scope now carries the caller principal (`_a2a_idem_scope`), preferring `mcp_key_id` over `username` because agent-scoped keys all resolve to one owner user; username alone would have collapsed two calling agents into a shared namespace and left a narrower version of the bug. Reverting to `f"a2a:{agent_name}"` makes the new end-to-end test fail with *"caller B's task was skipped — B replayed A's cached result"*, so it reproduces the reported failure rather than just asserting the fix.

**2. Unauthenticated well-known route had no rate limit** — added per-IP (60/min), placed ahead of the DB/Docker/agent work, trusted-proxy aware via `_get_client_ip` (nginx sets `X-Real-IP`). Live flood of 75 requests: **57×200 then 18×429 with `Retry-After: 50`**. A test asserts the limiter runs *before* any Docker/DB call — limiting after the expensive part wouldn't have fixed the flood.

**3. Seam file named a private table** — comment now describes the mechanism, and `a2a_gate.py` is added to `SEAM_FILES`. I verified the guard both passes now *and* catches the old wording when re-injected — it was previously blind to this file.

Warnings 4–8 addressed: `tasks/cancel` routes queued rows through the backlog CAS and returns `TaskNotCancelable` (-32002) rather than reporting success for a task that will later run and bill; SSE catches `CancelledError` (a `BaseException`, so the old `except Exception` never saw it) and releases the in-flight row; the allow-list UI/guide now say "email" and state the gate fails open; the access gate has coverage. Suggestions folded in: `cancelled` → `canceled`, body capped before parse (live: `-32600 "Request body too large"`), stream replay returns SSE.

Other live checks: non-exposed and nonexistent agents both return a uniform 404; JSON-RPC without auth is fail-closed 401; the card serves `protocolVersion 0.3.0`. Tests: **44 pass** (was 24). Adjacent suites green (77 passed incl. `test_186_enumeration_uniformity`, `test_models_centralized`).

## Validation findings

| Category | Status | Notes |
|---|---|---|
| PR size | ✅ | 34 files (under the 50 threshold) |
| Issue link | ✅ | `Related to trinity-enterprise#157, #160` — both resolve, P2/type-feature/complexity labelled. "Related to" (not "Closes") is correct per the cross-tracker manual-close convention |
| Commit messages | ✅ | conventional prefixes, explain *why* |
| Architecture | ✅ | now records the routes, `a2a_exposed`, `0.3.0`, and the seam service |
| Migrations (Inv #3) | ✅ | all four surfaces; `0024` chains to sole head `0023`; verified applying cleanly on live PG |
| MCP surface (Inv #13) | ✅ | `tools/a2a.ts` + `.test.ts` + client/server wiring |
| Security sweep | ✅ | no keys, tokens, real emails, public IPs, `.env`, or credential files |
| Infrastructure | ✅ | `nginx.conf` `/a2a/` proxy justified (front-door routing, SSE); `client_max_body_size 25m` applies |
| Build packaging | ✅ | Dockerfile uses a `*.py` glob — no gap (the skill's explicit-name check is stale and false-positives on `database.py`; confirmed present in the image) |
| Config packaging | ➖ | no new `os.getenv()` vars |
| **Requirements** | ❌ | see below |
| **Feature flow** | ⚠️ | see below |

### ❌ Requirements not updated for a new capability

`requirements/mcp.md` §32 covers only **#737 A2A Discoverability** — §32.1 is the *authenticated* card endpoint, explicitly scoped "Phase 1". The inbound server is a distinct new capability (two public routes, a new DB column, exposure flag, allow-list gate, MCP control tools) with **no requirement entry**. CLAUDE.md Rule 1 wants the area file updated *before* implementing; Rule 4's tier for "new capability" is requirements + feature-flow. This is the one item I'd genuinely block on — everything else is either fixed or a judgement call.

### ⚠️ Stale docs the PR makes wrong

`docs/user-docs/faq/collaboration.md:55` currently states:

> "There is no public unauthenticated `/.well-known` route yet"

This PR ships exactly that route, so the FAQ becomes **actively false on merge** — and `docs/user-docs/**` auto-publishes to a public search index via `sync-docs-to-vertex.yml`. Same paragraph also describes the card as authenticated-only.

`A2A v1.0` is stale in four more places (the card is `0.3.0`): `requirements/mcp.md:54,63`, `user-docs/integrations/mcp-server.md:123`, `user-docs/README.md:92`, `faq/collaboration.md:55`. The `architecture.md:784` instance is fixed; these are the same class.

### ⚠️ No feature flow

A vertical slice across router + seam + MCP tools + Vue panel + private module is exactly the shape `feature-flows/` exists for — compare `mcp-connector.md` for the analogous #846/ent#46 feature. Not a merge blocker on its own, but it's the tier the methodology asks for.

### Suggestion

`proxy_read_timeout 86400` on `location /a2a/` is inherited from the SSE pattern and is right for `message/stream`, but it also applies to the unauthenticated well-known card, where a 24h read timeout is more rope than that route needs.

## Recommendation

**REQUEST CHANGES** — narrowly, on documentation:

- [ ] Add a requirements entry for the A2A inbound server (`requirements/mcp.md`, alongside §32)
- [ ] Fix `faq/collaboration.md:55` — it will be false on merge and it auto-publishes publicly
- [ ] Update the four stale `A2A v1.0` → `0.3.0` references
- [ ] Consider a `feature-flows/a2a-inbound-server.md`

The code findings from the previous review are resolved and hold up under live exercise. Merge ordering from that review still stands: **#1628 → `dev` first, then trinity-enterprise#161**, which imports `a2a_gate` and `get_a2a_exposed`/`set_a2a_exposed` — none of which exist in OSS until this lands.

dolho:
Requested changes applied in `aada0476`. (Same disclosure as above — I wrote the commit I'm reporting on.)

- [x] **Requirements entry** — `requirements/mcp.md` §32.2 (inbound server, ent#157) + §32.3 (control over MCP, ent#160). Written as FR-N bullets in the §45 house style, recording the *why* behind each decision — the caller-scoped dedup, the limit-before-work ordering, honest cancel, and the fail-open gate — since those are what a future reader would otherwise reverse-engineer from the diff.
- [x] **`faq/collaboration.md`** — now describes exposure, both public routes, and that tasking still needs a key.
- [x] **Stale `A2A v1.0`** — corrected in **eight** places, not the four I first counted. Four were in `a2a_card_service.py`, the file that emits `"0.3.0"`.
- [x] **Feature flow** — `feature-flows/a2a-inbound-server.md`, indexed.

**One finding I missed in the review above:** `faq/advanced-features.md:51` carried a *second* false claim — "the public `/.well-known/agent-card.json` route is not served yet". Same class as the `collaboration.md` one, same auto-publishing tree; my grep pattern just didn't reach it. Fixed here.

Also refreshed the router + card-service docstrings, which still described a single-endpoint Phase 1 surface and called public serving a follow-up needing "a separate access-policy decision" — that decision is the per-agent exposure flag, and it shipped in this PR.

Verification: 44 tests still green; `enterprise-docs-guard` passes against the new prose (it greps `docs/**`, and §32.2 describes the seam mechanism without naming the private table); live card still serves `0.3.0`. I temporarily set `a2a_exposed=1` on a scratch agent to re-check the live route and **reverted it to 0**.

The blocking documentation item is cleared from my side. The code findings were resolved in `3cff879c`. Merge ordering unchanged: **#1628 → `dev`, then trinity-enterprise#161**.

Worth an independent pass before merge — two of the three criticals here were fixed and then validated by the same author, and the two false-doc findings show that a fresh reader looking at a different corner keeps finding things.

dolho:
Re-validated via `/validate-pr` at `a5c49eb6`.

> **Disclosure, third time:** I wrote every commit under review here (`3cff879c`, `aada0476`, `a5c49eb6`). This pass is only reliable as a check on *what I introduced* — it is not an independent judgement that the fixes are right. Treat the ✅ column below as "the mechanical checks pass", not as review.

## Status

| Category | Status | Notes |
|---|---|---|
| Base branch | ✅ | `dev` |
| PR size | ✅ | 42 files (was 34; under the 50 threshold) |
| Issue link | ✅ | `Related to trinity-enterprise#157, #160` — both resolve, P2/type-feature |
| Commit messages | ✅ | conventional, explain *why* |
| **Requirements** | ✅ | **was ❌** — `requirements/mcp.md` §32.2 (ent#157) + §32.3 (ent#160); §32.1's stale Behavior/Phase-2 text corrected |
| **Feature flow** | ✅ | **was ⚠️** — `feature-flows/a2a-inbound-server.md`, indexed; all 4 sibling links resolve |
| Architecture | ✅ | routes, `a2a_exposed`, `0.3.0`, seam service |
| Migrations (Inv #3) | ✅ | `check_alembic_parity.py` run with real SHAs: **PASS — both tracks updated**, `0024` paired with the SQLite DDL |
| Invariant guards | ✅ | `test_models_centralized` (#14) 4✅ · `test_186_enumeration_uniformity` (#8) 29✅ · `test_1560_agent_redis_key_parity` 4✅ · `test_alembic_revision_id_length` 27✅ |
| MCP surface (Inv #13) | ✅ | `tools/a2a.ts` + `.test.ts` |
| Enterprise guard | ✅ | passes over the enlarged docs tree — §32.2 describes the seam mechanism without naming the private table |
| Build/config packaging | ➖ | no new top-level module (glob COPY), no new `os.getenv()` |
| Feature tests | ✅ | 44 pass |

## Found in this pass (both mine, both fixed in `a5c49eb6`)

1. **Non-canonical status label.** §32.1 said `✅ Shipped` — a label the legend doesn't define. The four canonical values are ⏳ Not Started / 🚧 In Progress / ✅ Implemented / ❌ Removed; `✅ Implemented` is used 179× across `requirements/`, `✅ Shipped` exactly once, introduced by me. (#737 *is* closed, so ✅ itself was correct.)
2. **Index row stale.** `requirements.md` still routed `mcp.md` as "A2A discoverability", which no longer points a reader at the inbound server now in that area file.

Nothing else new. Everything in the previous comment's checklist is cleared.

## Recommendation

**APPROVE — with the standing caveat that this needs a reviewer who isn't me.**

The mechanical position is clean: the three criticals from vybe's review are fixed and exercised live, the documentation tier the methodology asks for is present, every invariant guard passes, and the dual-track migration is verified with real SHAs. What's missing is adversarial judgement. Across three passes I've now found two false doc claims and a bad status label *in my own work* — which is evidence the self-review loop still catches things, and equally evidence that it keeps finding them because one author's blind spots don't move.

Specifically worth an independent look:
- `_a2a_idem_scope`'s principal choice (`mcp_key_id or username or "anonymous"`) — the `"anonymous"` fallback is unreachable through the authenticated route today, but it's a shared-namespace bucket if a future caller ever reaches it with neither field.
- The allow-list fail-open decision. It's deliberate, documented, and consistent with Trinity's availability bias — but it's the kind of call worth a second opinion, not a self-affirmation.

Merge ordering unchanged: **#1628 → `dev`, then trinity-enterprise#161**.

dolho:
### Two additions on this branch

**1. Merge conflict resolved** (`cdc561ff`). The PR was `CONFLICTING`; sole conflict was `tests/registry.json`, where both sides appended entries at the same array position. Resolved as a union and verified programmatically that the merged set is exactly |HEAD ∪ dev| = 82 — nothing dropped, no duplicates. (The nightly bot's skip comment should clear on its next run.)

Also worth noting for whoever re-reviews: @vybe's `CHANGES_REQUESTED` from 07-15 was addressed by `3cff879c` the following day (messageId dedup scoped per `(agent, caller principal)`, public card rate-limited, seam de-genericised). The review state is stale rather than outstanding.

**2. Configurable exposed skills** (trinity-enterprise#180) — `a55b439a` + `ef6b8b80`, requested by the maintainer.

An exposed agent's card advertises every `template.yaml capabilities[]` tag, and this PR's well-known route is unauthenticated by design — so that full list is world-readable. This lets an operator choose what the outside is told.

It extends **this PR's own seam** rather than adding a module: `a2a_gate` gains a second provider (`exposed_skills(agent) -> Optional[List[str]]`), and both card surfaces go through one helper (`_card_with_exposed_skills`) so they can't disagree and a future third surface inherits the filter. OSS registers no provider → identity function → cards byte-identical. The config/storage/UI live in the enterprise a2a module (where exposure and the inbound allow-list already are).

**It is a disclosure control, and every layer says so.** `message/send` dispatches free-form text via `execute_task`, so hiding a skill hides it — it does not make it unreachable. That sentence is in requirements §32.4, the seam docstring, the filter, the endpoint and the UI copy, because the risk here isn't a bug, it's an operator reading it as an access control. A real boundary would be `allowed_tools`/guardrails on `triggered_by="a2a"` — separate work, own threat model.

Requirements §32.4 landed before the code; public docs describe the seam mechanism only, per the standing enterprise-docs rule. 17 new OSS tests (110 green across the A2A suites), 6 enterprise tests (module suite 24).

Worth a look in review: writing the tests surfaced that a malformed provider return would have failed **closed** and invisibly — a `str` iterates into single characters, matches no id, and silently empties the card, which reads as "this agent has no capabilities" rather than "the provider is broken". Malformed returns now take the same fail-open path as a raised error.

dolho (commented):
## Review — A2A inbound server (ent#157/#160)

Read the full public surface (`routers/a2a.py`, `services/a2a_gate.py`, `a2a_card_service.py`, `db/agent_settings/a2a_exposure.py`, the MCP tool). The security fundamentals are strong and the reasoning is written down where it matters. Three findings below, one worth acting on before merge; none are blockers on their own.

### What holds up
- **Uniform 404 on the public surface** (`_serve_card`/`_authorize_inbound`) — non-exposed / inaccessible / non-existent all answer the same 404, so `/a2a/{name}` is not an enumeration oracle (Invariant #8). Exposure is set only by the entitled enterprise setter, and `set_a2a_exposed` carries the `deleted_at IS NULL` guard, so an OSS/unentitled build is exposed-off by construction and a soft-deleted agent can't be flipped exposed.
- **Body cap before parse** (`_MAX_RPC_BODY_BYTES` via `request.body()`), **per-IP card rate limit before any DB/Docker/agent work**, trusted-proxy-aware client IP — the flood surface is closed at the cheap edge.
- **`messageId` dedup scoped per (agent, caller principal)** — the docstring on `_a2a_idem_scope` correctly identifies that `messageId` is peer-controlled and only unique *per client*; scoping on the agent alone would replay caller A's stored response (with its full text) to caller B. Keying on `mcp_key_id` is the right principal.
- **Honest cancel semantics** — terminal → `TASK_NOT_CANCELABLE`, `queued` → CAS `cancel_queued_execution`, running → `terminate_execution_on_agent`; a lost race returns "could not cancel" rather than lying "canceled" while the task drains and bills later.
- **`asyncio.CancelledError` handled explicitly in `_stream_task`** — it's a `BaseException`, so the `except Exception` wouldn't run `fail(decision)` and the row would sit `in_flight` for the full 24h TTL, wedging every retry of that messageId. Catching it is exactly right.

### Findings

**1. (low–med, security/disclosure) The public well-known card can leak `trinity.template` as the card `name`.**
`_fetch_template_data`'s `label_fallback` sets `display_name = labels.get("trinity.template", agent_name)`, and `generate_a2a_card` maps `display_name → card["name"]` (a2a_card_service.py:100,118). That fallback is taken whenever the agent is **stopped** or the agent-server template fetch fails / `has_template=False` — so a stopped exposed agent publishes its internal template/repo identifier (e.g. `github:Org/private-repo`) as its public, unauthenticated card name, while a running one shows the real `template.yaml` name. This is the exact token #846's CSO pass excluded from advertised surfaces as a cross-tenant disclosure + prompt-injection vector — and here it's on a fully unauthenticated route. Recommend the public fallback use `agent_name` (already public via the URL) rather than the template label.

**2. (minor, consider) The JSON-RPC task POST has no request-rate cap; the card GET does.**
`a2a_well_known_card` is rate-limited, but `POST /a2a/{name}` — the endpoint that dispatches a billed `execute_task` per `message/send` — relies solely on auth + `CapacityManager`/backlog for flood protection. That's defensible (authenticated + capacity-bounded, and distinct messageIds still queue rather than run), but a per-principal rate cap on `message/send` would match the card route's own reasoning ("the expensive part is what a flood amplifies") and bound backlog spam. Not a blocker; worth a follow-up if not this PR.

**3. (clarity) The "inbound allow-list" gates Trinity identities, not external peer DIDs.**
`_authorize_inbound` requires `can_user_access_agent` (owner/shared/admin) *first*, then passes `current_user.email or username` to `check_inbound_allowed`. So every inbound caller is already an owner/shared Trinity user, and the allow-list is a second layer keyed on Trinity email/username — not an external agent DID as the "A2A inbound allow-list" name might imply. Worth a one-line doc note that entries are Trinity identities, so an operator doesn't populate it with peer DIDs expecting them to match.

### Merge state (context, resolved)
I brought this branch up to current `dev` and resolved the three conflicts (migrations list, the `db/schedules.py`→package split #1481, `tests/registry.json` union). Caught the #1676-class Alembic hazard — this branch's `0025_a2a_exposed` and dev's own `0025_display_label` both chained off `0024` (two heads → PG boot fails); re-chained the A2A revision to **`0029_agent_ownership_a2a_exposed`** onto dev's head `0028_agent_reminders`. Verified a single Alembic head and a clean Postgres boot at `0029` with both `a2a_exposed` + `display_label` columns present. PR is now `MERGEABLE`.

Net: ship-worthy. I'd fix finding 1 before merge (small, and it's the only one touching an unauthenticated disclosure); 2 and 3 are fine as follow-ups.

AndriiPasternak31 (dismissed):
Validated via `/validate-pr`. Requesting changes on one hard blocker that CI can't currently see. First, a correction to the PR's own status, because it's unfair to the work sitting here:

**The existing `CHANGES_REQUESTED` is stale.** Both criticals from the 15 Jul review appear to be addressed in the commits through 20 Jul — `messageId` dedup is now scoped per `(agent, caller principal)` on `mcp_key_id` rather than on the agent alone, and the per-IP card rate limit runs *before* any DB / Docker / agent-container work. That review just hasn't been dismissed, so the PR reads as more blocked than it is.

### Blocking — Alembic branch point

`src/backend/migrations/versions/0029_agent_ownership_a2a_exposed.py` declares:

```python
revision = "0029_agent_ownership_a2a_exposed"
down_revision = "0028_agent_reminders"
```

`dev` now carries `0029_product_events.py` with the **same** parent:

```python
revision = "0029_product_events"
down_revision = "0028_agent_reminders"
```

Two revisions off `0028_agent_reminders` is a branch point, so `alembic upgrade head` raises *"Multiple head revisions are present"* — which is a **PostgreSQL boot failure**, not a migration warning (`init_database()`'s non-SQLite branch calls `upgrade_to_head()`). Needs renumbering to chain off the current head, `0030_slack_channel_allow_proactive`. Note this was correct when written — the review at the time verified `0024` chaining off the then-sole head `0023` — it's pure drift from sitting open. Worth re-checking the head at rebase rather than picking the next free number, since #1752 is also holding an unmerged revision.

The `pg-migrations` CI job runs a real `alembic upgrade head` and would have caught this, but it can't: the PR is conflicting, so there's no `refs/pull/N/merge` for CI to build and **no checks are reporting at all**.

### Also needed

**Conflicts** in `src/backend/db/migrations.py`, `src/mcp-server/src/server.ts`, `tests/registry.json`, `docs/user-docs/README.md`.

**The issue reference doesn't resolve.** `ent#157/#160` in the title isn't a form GitHub parses. It needs the fully-qualified closing keyword (`Fixes abilityai/trinity-enterprise#157`) to be a real link — and even then the status automation is same-repo only, so someone has to set `status-in-dev` on the private issue manually at merge. Flagging so it doesn't strand in `status-in-progress`.

### What holds up

Documentation is the most complete of any PR currently open — `architecture.md`, `requirements.md` + the area file, a new feature-flow doc, *and* the flow index row. Dual-track migration has all four surfaces. Entitlement gating is wired. Three test files including negative paths. At 46 files it's just under the split threshold.

Renumber the revision, rebase, and I'd expect this to go green.

dolho:
## /review Report

**Branch**: `feature/157-a2a-inbound-server` → `dev` (merge-base `39f29c6`, post-rebase)
**Files Changed**: 46 (+3404/−115)
**Scope**: CLEAN (bundled — see below)
**Plan Completion**: 6 done / 0 partial / 0 not done / 2 changed / 0 unverifiable

Diffed against the **merge-base**, so the 124 commits `dev` gained during the rebase aren't misattributed here.

---

### Scope

**Intent** — ent#157 (inbound A2A server) + ent#158 (config UI) + ent#160 (MCP control tools) + ent#180 (exposed-skill curation).
**Delivered** — exactly those four, no unrelated changes.

Four issues in one PR is bundling, but they're one feature epic and the title says so. Two out-of-area edits, both checked and both legitimately required: `db/schedules/analytics.py` (new trigger bucket) and `enterprise-docs-guard.yml` (adds the new seam file `a2a_gate.py` to `SEAM_FILES` — correctly *widening* the guard).

---

### Critical Findings

**[C1] Enum completeness: `a2a` missing from `_VALID_TRIGGERS` — the Executions filter silently returns everything** (Confidence: 10/10) — **FIXED in `6680ad1e`**

`src/backend/routers/executions.py:28`

Evidence — the constant's own comment states the failure mode:
```python
# NOTE: this is a filter ALLOW-LIST, not a DB enum — `triggered_by` is a plain
# TEXT column. An unknown value here degrades to "no filter" (see below), which
# silently returns EVERY execution instead of none, so a trigger that reaches
# this endpoint must be listed or its filter lies. `room` is ent#169.
_VALID_TRIGGERS = {"schedule", "manual", ..., "reminder", "room"}   # no "a2a"
```
and the degradation at line 61:
```python
triggered_by=triggered_by if triggered_by in _VALID_TRIGGERS else None,
```

Issue: the diff adds `"a2a"` to `_TRIGGER_BUCKETS` (`db/schedules/analytics.py:45`) but not to `_VALID_TRIGGERS`. There are **three** constants enumerating triggers and this reached one. `?triggered_by=a2a` on the fleet Executions list therefore returned every execution on the install rather than the A2A ones — the filter lies instead of erroring, which is the exact outcome the comment warns about.

Why it matters: A2A is a surface where an *external* party drives execution and spend. "Show me only what the outside world triggered" is the first question an operator asks, and it silently answered with everything.

Fix applied: added `"a2a"` to the set, with the `room`-style provenance note. 408 tests pass.

---

### Informational Findings

**[I1] `_AUTONOMOUS_TRIGGERS` also omits `a2a` — deliberate or missed?** (Confidence: 6/10)
`src/backend/services/task_execution_service.py:259`

The third trigger constant governs the unresolved-slash-command Operating Room alert:
> *"Triggers with no human watching the reply — an unresolved-command run here is invisible without an explicit signal... An interactive-ish trigger (a user's /task, mcp, or public turn) still classifies as FAILED but the caller already sees the 'Unknown command' reply."*

Genuinely arguable both ways, which is why it's informational and I did **not** auto-fix it. A2A `message/send` does return the reply to its caller, so it is "interactive-ish" like `mcp`/`public` — omission is defensible. But that caller is an *external orchestrator*, not a human who will read "Unknown command: /foo" and act. Suggestion: state the choice in a comment either way, so the next reader doesn't have to re-derive it (and doesn't "fix" it by accident).

**[I2] AC names `effect_guard`; the implementation uses trigger-boundary idempotency, and the deviation isn't recorded** (Confidence: 8/10)
`src/backend/routers/a2a.py:429`

ent#157 AC: *"reuse `effect_guard` (#1084) so at-least-once re-delivery can't double-execute a side-effecting task."* The diff instead uses `idempotency_service.begin/complete/fail` keyed on `messageId` (Invariant #18). `grep effect_guard` over `routers/a2a.py` + `services/a2a_gate.py`: no match.

I think the implementation is **right** and the AC is imprecise — these are different layers per `architecture.md`: trigger-boundary dedup stops a re-POST creating a second *execution* (which is what A2A inbound is), while `effect_guard` is a per-*sink* guard on the agent's outbound tool calls, already wired at those sinks. A2A inbound is a trigger boundary, not a sink.

The problem is only that nothing says so. The PR body and `docs/memory/feature-flows/a2a-inbound-server.md` both discuss the idempotency choice at length but never mention `effect_guard`, so the AC reads as unmet. Suggestion: one sentence in the feature-flow doc — "the AC's `effect_guard` reference is the sink layer; the trigger boundary is Invariant #18, and here is why."

**[I3] gitleaks allow-list widened to the whole of `docs/user-docs/`** (Confidence: 7/10)
`.gitleaks.toml:57`

```toml
'''(^|/)docs/user-docs/''',        # generated user-facing docs carry the same curl `Authorization: Bearer`
```

This disables secret scanning across a tree that (per the neighbouring comment) **auto-publishes to a public search index** via `sync-docs-to-vertex.yml`. The motivating problem is real and well-stated — a commit-range scan flags the placeholder in the commit that introduced it, so editing the head can't clear it. But the remedy is broader than the cause: a real credential pasted into any future user-doc is now unscanned, on a public repo, in the tree with the widest blast radius.

Suggestion: narrow to the cause — a `stopwords`/regex allowlist entry for the `trinity_mcp_…` placeholder shape, or a path scoped to `a2a-protocol.md`, rather than the whole directory. Not blocking; flagging because the *scope* of the exemption outlives the commit that needed it.

**[I4] No rate limit on `POST /a2a/{name}`** (Confidence: 6/10)
`src/backend/routers/a2a.py:383`

The unauthenticated card route is per-IP limited (`A2A_CARD_RATE_LIMIT`, correctly *before* any DB/Docker work). The JSON-RPC route has none. Mostly fine: it's authenticated and `message/send` admits through `CapacityManager` (slots + `max_backlog_depth`). But `tasks/get` is a cheap unbounded DB read behind the same auth, and A2A callers are external systems with retry loops. Suggestion: reuse `rate_limiter.check` per `(agent, caller identity)` on the non-executing methods.

**[I5] `tasks/resubscribe` has an implemented error path and zero tests** (Confidence: 7/10)
`src/backend/routers/a2a.py:526`

Returning `-32004 unsupported` is a good phased choice — explicit beats a generic method-not-found. But `grep -ci resubscribe tests/unit/test_157_a2a_inbound_server.py` → `0`, while every other method has coverage. It's the one method whose contract could silently change when someone implements it.

---

### Clean Categories (verified, with citations — not assumed)

- **Auth boundaries** — `_authorize_inbound` (`a2a.py:357`) is **self-uniform** per Invariant #8: non-exposed → 404 and inaccessible → 404, with 403 reachable *only after* access is proven, so the public surface is not an enumeration oracle. `tasks/get`/`tasks/cancel` both gate on `_exec_field(row, "agent_name") != agent_name` (lines 474, 493) — no cross-agent execution read.
- **Connector-key containment** — the inline access check bypasses the `_enforce_connector_scope` helper, so I checked whether the fence is escapable. It isn't: `dependencies.py:383-399` enforces it centrally in `get_current_user` against a 2-route allowlist, so a connector key gets 403 before reaching `/a2a/`. The ghost fence (`dependencies.py:409`) applies at the same point. The `test_1310_auth_wiring.py` allowlist entry is therefore justified, and its stated reason matches the code.
- **Migrations (Invariant #3)** — all four tracks consistent: `schema.py` DDL, `tables.py` MetaData, SQLite `_migrate_agent_ownership_a2a_exposed`, Alembic `0031`. Additive nullable column `DEFAULT 0` — no data-loss path. `ScriptDirectory.get_heads()` resolves to exactly one head (32-revision chain), verified inside the backend image after the rebase re-chained `0029→0031` onto `0030_slack_channel_allow_proactive`.
- **Soft-delete** — all three accessors in `db/agent_settings/a2a_exposure.py` (lines 30, 45, 59) guard `deleted_at IS NULL`, so a soft-deleted agent can't be read or flipped exposed.
- **Frontend** — no `v-html` in `A2aPanel.vue`; every one of the 9 async paths has a `catch` with a user-facing message. Tab gated on `sessionsStore.a2aAvailable && can_share && !isSystem` (`AgentDetail.vue:830`), so OSS/unentitled builds never render a dead tab.
- **Public-repo / enterprise-docs rule** — the docs diff contains no private-schema or paid-catalog token; the only match is a line *describing* the rule. `a2a_gate.py` was correctly added to the guard's `SEAM_FILES`. gitleaks + enterprise-docs-guard both pass.
- **Error handling** — the SSE generator's `except asyncio.CancelledError` (`a2a.py:561`) is a subtle and correct catch: a client disconnect raises a `BaseException` that `except Exception` misses, which would strand the idempotency row `in_flight` for the full 24h TTL. Caught, documented, and released.
- **Magic numbers** — body cap, rate limits and every JSON-RPC error code are named constants (`a2a.py:211-231`).

### Notes (not findings against this PR)

- `stores/agents.js` uses raw `axios` rather than the shared `api` client (Invariant #7). The new code follows the file: **85** raw `axios.` calls, **0** `api.` calls, all pre-existing. Consistent, so not charged here — but the whole store is a standing invariant violation worth its own issue.
- `tests/registry.json` carries a duplicate entry for `test_181_agent_display_label.py`. I verified it exists on `origin/dev` too, so it predates this branch and I left it alone.

### Summary
- **Critical: 1** — found and fixed in `6680ad1e`
- **Informational: 5** — I2 and I3 are the two worth a decision before merge
- **Scope: clean**

The security-sensitive parts of this PR — the public unauthenticated surface, the uniform-404 discipline, the fail-open seam, the SSE cancellation path — are noticeably careful. The one real defect was in the boring part: a new enum value wired into one of the three constants that enumerate it.

dolho:
Addendum on **[I2]** (`effect_guard` vs trigger-boundary idempotency) — I checked `docs/memory/learnings.md` while capturing learnings and this exact confusion is **already in the ledger**, from Brain Orb Phase 4a (trinity-enterprise#61):

> *"The issue said 'wire the #1084 effect guard,' but `idempotency_service.effect_guard` builds its scope as `effect:{execution_id}` and `resolve_and_validate_execution` fail-opens to no-dedup when there's no execution... The correct primitive is the trigger-boundary `Idempotency-Key` path."*

So this is the **second** issue whose AC names `effect_guard` where the trigger boundary is the right layer. That independently confirms the implementation choice here was correct — I2 downgrades to purely a documentation gap. Still worth the one sentence in the feature-flow doc, precisely because the AC text will keep saying `effect_guard` and the next reader will keep checking.

Also appended a ledger entry for the C1 class (`ae05b775`). The durable part isn't "remember three constants" — it's the asymmetry: the trigger constant that's easiest to remember (analytics) is the one that degrades *safely* to `Other`, while the one in a different file degrades to **no filter** and returns every row with no error and nothing to complain about. Reviewing an enum addition means finding the call sites whose fallback is permissive first, not the ones that are easiest to grep.

Concrete follow-up worth considering: a parity test asserting every trigger the code can emit appears in `_VALID_TRIGGERS`, in the shape of `test_1560_agent_redis_key_parity.py` — a comment at the site warning about the trap has now failed to prevent it twice (`room` was added for the same reason).

dolho:
@AndriiPasternak31 — all three items handled. CI is now reporting and fully green (it couldn't run at all before, since the conflicts meant no `refs/pull/N/merge` to build — which is exactly why the Alembic branch point was invisible).

**Alembic branch point** — fixed in `f2b53dab`. Renumbered to `0031_agent_ownership_a2a_exposed` chaining off `0030_slack_channel_allow_proactive`, plus the stale cross-reference in the mirrored SQLite migration's docstring.

Took the advice to re-check the head at rebase rather than pick the next free number, and verified the *tree* rather than reading `down_revision` lines — run inside the backend image, since the revision graph is what CI walks:

```
heads: ['0031_agent_ownership_a2a_exposed']
depth: 32 | top 4: ['0031_agent_ownership_a2a_exposed', '0030_slack_channel_allow_proactive',
                    '0029_product_events', '0028_agent_reminders']
SINGLE HEAD OK
```

`pg-migrations` now passes on a real `alembic upgrade head`.

**Conflicts** — resolved, branch rebased onto `dev` (was 124 behind, now 0). All four were append-collisions where both sides added an entry; kept both in each case:
- `db/migrations.py` — SQLite migration registry, a2a appended after dev's two
- `mcp-server/src/server.ts` — tool groups, `createRoomTools` + `createA2ATools` both retained (TypeScript compiles, 100 MCP tests pass)
- `tests/registry.json` — spliced and validated by a real `json.loads` + a delta check against `dev` (exactly the 2 a2a entries added, 0 lost) rather than eyeballing the braces
- `docs/user-docs/README.md` — took dev's newer "100+ MCP tools" figure and this PR's own A2A line

**Issue references** — corrected. You're right that `ent#157` doesn't parse; worth adding that `trinity-enterprise#157` (which the body actually had) doesn't resolve either — a cross-repo reference needs the owner, `abilityai/trinity-enterprise#157`. Now fully qualified, and ent#158/#180 added since the PR grew to cover those too. Kept as **"Related to"** rather than `Fixes`, per the relate-don't-close convention here — so your point about setting `status-in-dev` manually at merge stands either way. Thanks for flagging it; that's the kind of thing that stalls silently.

**Also, unprompted:** I ran `/review` over the rebased diff and it found one real defect — `triggered_by="a2a"` was wired into `_TRIGGER_BUCKETS` but not `_VALID_TRIGGERS`, so `?triggered_by=a2a` on the Executions list silently returned *every* execution instead of the A2A ones. Fixed in `6680ad1e`, with the full report [here](https://github.com/Abilityai/trinity/pull/1628#issuecomment-5131104482).

Re-requesting review. The 15 Jul `CHANGES_REQUESTED` from @vybe is still stale as you noted.

dolho:
@vybe all three CRITICALs and the four warnings are resolved, and I've just re-merged `dev` (only conflict was the append-only `learnings.md` ledger — both entries kept; migrations auto-merged, single Alembic head `0031`). Now MERGEABLE.

**CRITICAL 1 — messageId cross-caller disclosure.** `_dedup_scope` now keys on `(agent, caller principal)` → `f"a2a:{agent_name}:{principal}"`, so each caller gets a private `messageId` namespace and a collision can only replay the caller's own snapshot (`a2a.py:289`).

**CRITICAL 2 — unauthenticated card route unthrottled.** `A2A_CARD_RATE_LIMIT = 60/min` per IP via `rate_limiter.enforce` on `_serve_card` (`a2a.py:213,345`), matching the `public.py`/`files.py`/`webhooks.py` shape.

**CRITICAL 3 — seam file named a private table.** The `enterprise_a2a_inbound_allowlist` reference is gone from `a2a_gate.py` (describes the mechanism now), and `a2a_gate.py` is added to `SEAM_FILES` in `enterprise-docs-guard.yml` so the next one is caught.

**WARN 4 — tasks/cancel reported success unconditionally.** Now checks terminal state first, routes a `queued` row through `db.cancel_queued_execution` (backlog CAS, not a registry call that 404s and lies), reads the bool, and returns `_A2A_TASK_NOT_CANCELABLE` when the cancel didn't land (`a2a.py:488`).

**WARN 5 — SSE disconnect wedged the messageId 24h.** `except asyncio.CancelledError:` now calls `idempotency_service.fail(decision)` before re-raising, so a client disconnect releases the in-flight row instead of leaving it stuck (`a2a.py:561`).

**WARN 6 — allow-list identity contract mismatch.** The panel now states the identity is **email (or username)** with a `caller@example.com` placeholder, matching `caller_identity = email or username` — the `did:example` copy that would have made every real caller fail is gone.

**WARN 7 — access gate had zero coverage.** `test_157_a2a_inbound_server.py` now has an "Access gate (exposed ≠ accessible)" section that varies `state["access"]` and exercises the exposed-but-inaccessible → uniform 404 branch.

CI is green (no failing checks). Ready for re-review.

dolho:
## Rebased onto `dev` — conflicts resolved, now MERGEABLE

Branch was 57 commits behind and `CONFLICTING`. Replayed the 21 commits onto `dev` (`8f335101` → `c4ba9cf6`); the dev-merge commit is gone, so the history is linear.

### Conflicts (3, all append-collisions — both sides kept)

| File | Collision |
|---|---|
| `db/migrations.py` | MIGRATIONS list: dev's `channel_report_back_columns` + `telegram_progress_indicator`, then a2a last (correct application order) |
| `tests/registry.json` | both sides appended entries; spliced and re-parsed to prove the JSON is valid (106 entries) |
| `docs/memory/learnings.md` | append-only ledger; dev's entries then this branch's |

### Two things the conflict resolution could not catch

**1. Two Alembic heads.** The a2a revision chained onto `0030`, but dev has since landed `0031_channel_report_back` and `0032_telegram_progress_indicator` — leaving a duplicate `0031` number and a branched graph, which `alembic upgrade head` refuses (PG boot break). Renamed `0031_agent_ownership_a2a_exposed` → `0033_` and re-pointed `down_revision` at `0032_telegram_progress_indicator`.

Verified two ways, not just by reading:
```
ScriptDirectory.get_heads() → ['0033_agent_ownership_a2a_exposed']   # exactly one, 34-revision chain

# and the real thing, against PostgreSQL 16 on an empty database:
alembic upgrade head
  … 0032_telegram_progress_indicator -> 0033_agent_ownership_a2a_exposed
information_schema → agent_ownership.a2a_exposed | integer | default 0
alembic_version    → 0033_agent_ownership_a2a_exposed
```

The SQLite track needed no fix — its entry appends after dev's two, which is the right order.

**2. The enterprise submodule pointer was silently dropped.** The bump to `740063e3` (the A2A module — ent#160 registration wrapper, #180 curated skills) rode on the dev-merge commit that a flat rebase discards, so the rebased branch quietly reverted to dev's `e898cef8` and an enterprise build would have lost the module this PR's OSS seam exists for. Confirmed `e898cef8` is an **ancestor** of `740063e3` (restoring is a fast-forward, not a regression) and restored it.

Both fixes are in `c4ba9cf6`.

### Verification

- No conflict markers anywhere in the tree
- Touched-file set identical to pre-rebase; each of the 12 overlapping files diffs as **a2a-only additions** on top of dev — no dev content lost, nothing duplicated
- `test_schema_parity` + `test_alembic_parity_guard` + `test_alembic_revision_id_length` → 64 passed
- `test_157_a2a_inbound_server` + `test_a2a_exposure_mixin` → 55 passed
- Frontend `npm run build` + `check:tokens` clean
- Full local unit suite running; CI's base/head regression diff is the authoritative gate

### One pre-existing gap, unchanged by the rebase

`a2a` is in `visibleTabs` but missing from `DEEP_LINK_TABS` in `AgentDetail.vue`, so `?tab=a2a` doesn't deep-link. Byte-identical to the pre-rebase branch — a branch bug, not rebase damage. Left alone since this was a rebase; say the word and it's a one-line commit.

Pre-rebase tip `8f335101` is kept locally as `backup/1628-pre-rebase`.

dolho:
Rebased onto current `dev` and re-chained the Alembic revision. Status against both reviews:

## @AndriiPasternak31's blocker — fixed

The drift had gone one step further than your review saw: the branch was on `0033_agent_ownership_a2a_exposed`, and `dev` now carries `0033_agent_evaluations` — same parent `0032_telegram_progress_indicator`, so both a **filename** and a **graph** collision.

Now `0035_agent_ownership_a2a_exposed` chained to `0033_agent_evaluations`. **0034 is skipped deliberately**: PR #1901 holds `0034_skill_sources` unmerged off the same head. Whichever of the two lands second still has to re-chain — that is inherent to two open PRs both adding a revision — but skipping the number keeps the filenames from colliding too, so that rebase is a one-line `down_revision` edit rather than a rename plus an edit. The reasoning is now in the revision docstring so the next rebase doesn't re-derive it, per your point about re-checking the head rather than picking the next free number.

Verified structurally rather than by eye: the version graph has exactly **one head** and no dangling `down_revision` (35 revisions walked). 61 alembic-guard tests green.

**Conflicts** resolved in all four files you listed plus `docs/memory/learnings.md`. Two are append-only registries (`tests/registry.json`, `learnings.md`) where both sides had appended — unioned, ours last, no reordering. The registry union was re-serialized from parsed JSON rather than line-spliced, because splicing loses the separating comma (I did exactly that on the first attempt and caught it on the re-parse).

**Issue reference**: already fully-qualified in the PR body — `abilityai/trinity-enterprise#157/#158/#160/#180`. Deliberately `Related to` rather than a closing keyword (relate-don't-close is the convention here), with a note in the body that the private issues need `status-in-dev` set manually at merge since the automation is same-repo only. Flagging so it isn't read as an oversight.

## @vybe's review — all three criticals addressed in `a5351922`

1. **messageId disclosure** — dedup scope now carries the caller principal (`mcp_key_id`), so two callers cannot collide on `req-1`.
2. **Unauthenticated card rate limit** — per-IP limit runs **before** any DB / Docker / agent-container work, matching `public.py`/`files.py`/`webhooks.py`.
3. **Enterprise table named in a public seam** — comment describes the mechanism, no table name; `a2a_gate.py` added to `SEAM_FILES` so the guard covers it.

The warnings are addressed too, not just the blockers:

- **W4 (cancel lies)** — `tasks/cancel` now reports what actually happened: terminal states short-circuit, a *queued* row goes through `db.cancel_queued_execution` (CAS, the `chat.py` branch's shape), and a lost race returns an error instead of `"canceled"`.
- **W5 (SSE disconnect)** — explicit `except asyncio.CancelledError` with the reason in a comment (it's a `BaseException` since 3.8, so `except Exception` never saw it); the key no longer wedges for 24h.
- **W6 (identity contract)** — the allow-list placeholder is `caller@example.com`, matching what `caller_identity` actually resolves to.
- **W7 (access gate coverage)** — `209a38ab` adds the exposed-but-inaccessible branch alongside SSE, idempotency, task-state and soft-delete cases.
- Non-`success` no longer flattens to `failed`: `cancelled` maps to `canceled`, which the spec treats as distinct.

Local verification after the rebase: **440 passed** across `a2a / 157 / 180 / migration / schema / registry / trigger`. CI can finally report — while the PR was conflicting there was no `refs/pull/1628/merge` to build, which is why no checks were showing at all.

Ready for re-review. Both existing `CHANGES_REQUESTED` are now stale; please dismiss or re-review when you get a chance.

vybe (approved):
Approving. All three of my criticals are resolved, all five warnings addressed, and Andrii's Alembic blocker is fixed — each verified against the current head rather than taken from the PR body.

**My criticals:**

1. **Cross-caller `messageId` disclosure** — scope is now `a2a:{agent}:{principal}` via `_a2a_idem_scope`, not `a2a:{agent}`. Two callers sharing an agent can no longer collide on a peer-controlled protocol field.
2. **Unauthenticated card route** — `A2A_CARD_RATE_LIMIT = 60`/60s per IP through the shared `services/rate_limiter`, enforced *before* the DB read, the Docker call and the container fetch, with trusted-proxy-aware client resolution. Matches the `public.py` / `files.py` / `webhooks.py` (#1424) precedent.
3. **Seam file naming the private table** — the `enterprise_a2a_inbound_allowlist` reference is gone, and `a2a_gate.py` is added to `SEAM_FILES` in `enterprise-docs-guard.yml`, so the next one gets caught. That second half is the part that matters.

**Warnings:** `tasks/cancel` now splits queued (`db.cancel_queued_execution`, backlog CAS) from running (`-32002` when `terminate_execution_on_agent` returns False) instead of reporting success unconditionally; `_a2a_state_for` maps `cancelled → canceled`; the SSE generator catches `asyncio.CancelledError` explicitly with the `BaseException`-since-3.8 reasoning written down; the access gate now has coverage that actually varies `state["access"]`. Docs went well past what I asked — `architecture.md`, `requirements.md` + the area file, a new feature-flow doc, and the flow-index row.

**Andrii's blocker:** the revision is renumbered `0035_agent_ownership_a2a_exposed` chaining off `0034_skill_sources`. I recomputed the graph across all 35 revisions on dev rather than trusting the number — `0034_skill_sources` is the sole head, so this is a single head after merge. `pg-migrations` is green, which it could not report before because the PR was conflicting.

---

**I pushed two drift fixes** (`dbbf7fd6`) rather than bounce this again — both caused by dev moving under the branch, neither touching the feature:

- **`regression diff` was failing deterministically across all three seeds** on `test_2048_pull_pilot_reach`. #2057 landed 08-07, after this branch's last commit, asserting `_AUTONOMOUS_TRIGGERS - PULL_REACHABLE_TRIGGERS == set(_STRANDED)` — and this PR adds `a2a` to `_AUTONOMOUS_TRIGGERS`. That is the guard working exactly as intended: a new trigger must surface as an *unreviewed addition*. I classified it as **stranded** and wrote the reason at the site — `message/send` consumes `result.response` synchronously to build the JSON-RPC artifact, and a pull-claimed row produces no synchronous response, so pull dispatch structurally cannot serve it. Same reason `fan_out` and `loop` are stranded. Shout if you'd have called it differently; it's a reviewable decision, not a mechanical one.
- **Reverted the `src/backend/enterprise` bump to `740063e3d`.** That commit exists only on trinity-enterprise#161's unmerged branch — 6 behind that branch's own tip and diverged from enterprise `main` (main +5/-7). Merging it would point dev at unmerged code missing five commits of enterprise main, with green CI: the stale-pin mode where entitled features 403 everywhere and nothing looks wrong. The pin stays at dev's `e898cef83`; the pointer moves in its own PR once ent#161 is on enterprise main.

Verified locally at `dbbf7fd6`: 170 passed across the four a2a suites, the pull-pilot suite, the Invariant #8 auth-wiring guard and the Alembic revision-id lint. All 23 checks green.

**Sequencing from here** (unchanged from my ent#161 review): this merges to `dev` first, because `enterprise/backend/a2a/__init__.py` imports `services.a2a_gate` and calls `get_a2a_exposed`/`set_a2a_exposed` — none of which exist in OSS until this lands. Then ent#161 → enterprise `main`. Then the pointer bump, on a branch that already contains this PR. I'll set `status-in-dev` on ent#157/#158/#160/#180 manually at merge, since the automation is same-repo only.

Issue #1402, opened 2026-07-01
feat: MAX_REDELIVERY cap + async operator-queue human-gate lever (lease-reaper backstop)


vybe:
## Summary

Build the re-delivery cap that parks a poison task as `FAILED(MAX_REDELIVERY)` into the operator queue with an alert. This same lever is the landing spot for **irreversible-and-un-confineable effects** that the platform cannot auto-recover. It is **unbuilt today** — the existing `retry_count` column is #678's reader-race auto-retry, not a lease cap.

## Context

Under the pull / work-stealing model (#1081), lease-expiry re-delivery re-runs failed turns. Two cases must escalate to a human instead of looping forever:
1. a **poison task** that keeps failing on every re-delivery, and
2. an **irreversible effect the platform cannot make safe** (no confined Trinity tool fronting the rail, no provider-native idempotency key — e.g. an action through the agent's own key / `gh` / `curl`).

Both land in the operator queue (OPS-001). This lever is doubly load-bearing and gates defaulting pull mode on for any side-effect-bearing agent.

## Design constraint: NO synchronous user gates — comms are asynchronous via the operator queue

The human gate is the **asynchronous operator queue (OPS-001)** — **not** a blocking in-turn prompt. This is a hard design constraint, not an implementation detail:

- A turn that needs human input **MUST NOT** hold a worker open waiting for a response. Blocking a turn on a human would pin a worker and violate async-first (Principle #3) — the exact failure mode the pull model exists to remove.
- Instead, the turn **parks** a question/approval into the operator queue and **ends**. The operator responds later; resolution re-enters **asynchronously** (a new execution, or the operator-queue respond → re-trigger path).
- **Agents must be designed to treat all user/operator communication as asynchronous — fire-and-park, never block-and-wait.** An agent should never assume a synchronous human answer is available mid-turn.
- This contract must be surfaced to agents (agent-facing docs + the platform system prompt) so agent authors build for it explicitly.

## Acceptance Criteria

- [ ] Per-execution re-delivery counter + configurable cap; over-cap → status `FAILED(MAX_REDELIVERY)`.
- [ ] Over-cap task **parked into the operator queue with an alert** (never silently dropped).
- [ ] Irreversible-and-un-confineable effects route to the **same** async human gate.
- [ ] The gate is **asynchronous** (operator queue) — never a synchronous in-turn block; a turn never holds a worker waiting on a human.
[...]

Issue #1632, opened 2026-07-15
bug: operator-queue create path has no rate limit or size caps — flooding/operator-fatigue surface (blocking pull default-ON)


vybe:
## Summary

The operator-queue create path has no per-agent rate limit and no size caps: `create_item` accepts unbounded `title`/`question`/`context`, the sync service ingests the agent's queue file every 5s (up to the generic 100 MB file cap), and agents choose `priority: critical` freely (drives the red-pulse NavBar badge). Contrast `agent_reports`: 256 KB payload cap + 30/60s `services/rate_limiter.py` limiter.

## Context

Found in the #1402 review (finding S2). #1402 makes the operator queue the approval channel for irreversible actions (payments, destructive ops), so a compromised/prompt-injected agent flooding plausible "approve this" items becomes a meaningful attack: operator fatigue → reflexive approval. Markdown rendering is DOMPurify-sanitized (XSS covered) — the exposure is volume and social engineering, not script injection. **This should land before pull-mode default-ON** multiplies autonomous re-delivered turns.

## Acceptance Criteria

- [ ] Per-agent creation rate limit at the sync ingestion boundary (reuse `services/rate_limiter.py` — sliding window, fail-open), with over-limit items skipped + WARN logged + a single aggregated alert (not one per skipped item).
- [ ] Length caps on `title` / `question` / `context` at ingestion (named limits, truncate-with-marker or reject — pick one and document it).
- [ ] Platform-created items (lease-reaper poison-park, DB-first) are exempt from the per-agent agent-file limits.
- [ ] Existing legitimate flows unaffected (pipeline escalations, #1402 approvals): limits generous enough for real usage — document chosen numbers.
- [ ] Tests: over-limit flood is capped; caps enforced; reaper items unaffected.

## Technical Notes

- `src/backend/services/operator_queue_service.py` (sync ingestion), `src/backend/db/operator_queue.py` (create), `services/rate_limiter.py` (shared primitive — do not hand-roll Redis counters, see architecture.md Container Security → Rate limiting #1023).
- Related: #1525 (quarantine of failing creates), #1402 (approval-channel contract), #1081 (pull default-ON gate list).

AndriiPasternak31:
/claim — picking this up as part of the reliability/security open-ticket push.

AndriiPasternak31:
@vybe — scope-sync ahead of the #1632 draft PR (ingestion depth/rate/size caps at the `_sync_agent` seam):

[...]

Issue #1631, opened 2026-07-15
bug: operator_queue.id is a global primary key — cross-agent id collision silently swallows requests


vybe:
## Summary

`operator_queue.id` is a `TEXT PRIMARY KEY` shared fleet-wide; the create path uses `on_conflict_do_nothing(index_elements=["id"])` (`src/backend/db/operator_queue.py`) and the 5s sync loop's dedup is id-only (`operator_queue_item_exists(req_id)`, no agent scoping). Two agents choosing the same request id (e.g. the old date-serial example `req-20260307-001`) means the second agent's request is **silently never created** — never shown to an operator, never responded to, and never flipped terminal in the agent's file (write-back queries are agent-scoped), so the agent waits forever with zero log signal.

## Context

Found in the #1402 review. The #1402 contract mitigates by mandating execution-id-derived request ids (`approval-{execution_id}-{slug}`) in the platform prompt and agent guide, but the schema-level fix is scoping uniqueness to the agent. A related lower-severity wrinkle: the lease reaper's platform-created `poison-{execution_id}` ids share the same global namespace, so an adversarial agent can pre-create a matching id to replace the platform's alert text with its own (`_create_park_item` still returns True on conflict, so the row still parks — visibility is degraded, not lost).

## Acceptance Criteria

- [ ] Queue-item uniqueness is scoped to `(agent_name, id)` (dual-track migration: SQLite `db/migrations.py` + Alembic revision + `db/schema.py` + `db/tables.py`), OR — if the PK migration is judged too disruptive — a cross-agent conflict on create is detected and logged loudly (WARN + audit) instead of silently swallowed.
- [ ] Sync-loop dedup (`operator_queue_item_exists`) checks the same scoped key.
- [ ] Idempotent re-insert of the SAME agent's same id keeps today's on-conflict no-op behavior (the #1402 re-park dedup relies on it; the reaper's alert-first idempotency too).
- [ ] Regression test: two agents writing the same id both get visible queue items (or the loud-conflict path fires).

## Technical Notes

- `src/backend/db/operator_queue.py` (create/dedup), `src/backend/services/operator_queue_service.py` (sync), `services/lease_reaper_service.py` (`poison-{eid}` items).
- Migration reminder: schema changes touch FOUR files (schema.py, tables.py, migrations.py, Alembic revision) — see `docs/memory/learnings.md` 2026-06-23.

Issue #1629, opened 2026-07-15
bug: pull-claimed turns bypass platform system prompt composition — platform contract never delivered on the pull path


vybe:
## Summary

Pull-claimed turns run **without the platform system prompt**. The agent-side pull worker (`docker/base-image/agent_server/services/pull_worker.py`) executes a claimed task via `execute_headless(system_prompt=overrides.get(\"system_prompt\"))`, where `task_overrides` is reconstructed verbatim from `backlog_metadata` (`src/backend/services/pull_coordination_service.py::_build_claim_response`). Nothing on this path calls `platform_prompt_service.compose_system_prompt` — that composition happens only on the push/drain dispatch path (`task_execution_service`).

Net effect: a pull-claimed turn — **including every lease-reaper re-delivery** — receives none of the platform instructions: the operator-queue protocol, file-sharing rules, user-memory rules, and the #1402 async human-gate contract (fire-and-park). The #1402 contract is delivered via the platform prompt, so without this fix it never reaches the exact turns it governs.

## Context

Found during #1402 plan review (independent strategy voice, verified against code). The pull path is dark today (`PULL_MODE_PILOT_AGENTS` empty), so nothing user-facing is broken yet — but this is **blocking for pull default-on** and for #1402's contract to be real on re-delivered turns. Pre-existing gap from the Phase 0–3 pilot merge (#1550), worth fixing regardless of #1402.

## Acceptance Criteria

- [ ] A pull-claimed turn receives the same composed, runtime-aware platform system prompt a push-dispatched turn does (composed with the caller's `system_prompt` override, runtime resolved from the `trinity.agent-runtime` label).
- [ ] Composition is fail-open: on any composition error the turn still runs with the caller prompt only (parity with today) plus a WARN log — a prompt failure never blocks dispatch.
- [ ] Re-delivered turns (`redelivery_count > 0`) get a deterministic re-delivery framing banner prepended to the claimed message **at read time, never persisted** — stating attempt N of cap and the irreversible-effect park guidance (from the #1402 review: the model otherwise cannot observe that a turn is a re-delivery; `prior_trace` is a reserved field until #1401 wiring lands).
- [ ] Unit tests: claim response carries the composed prompt; banner present iff `redelivery_count > 0`; the stored `schedule_executions.message` is never mutated.

## Technical Notes

[...]

PR #1635, opened 2026-07-15
feat: async operator human-gate contract — fire-and-park prompt + derived request ids (#1402)

vybe:
## Summary

- Ships the remaining Abilityai/trinity#1402 deliverable: the **authored async human-gate contract** agents receive. The MAX_REDELIVERY cap + poison-park *mechanism* already shipped in Phase 3 (#1550) — verified in-code during planning; this PR is the contract surface + docs + tests.
- Platform prompt "Operator Communication" rewritten around **fire-and-park, never block-and-wait**: park the request → end the turn → process `responded` items in a later turn; **ask-before-irreversible-actions** guidance; **execution-id-derived request ids** (`approval-{execution_id}-{slug}`) closing the global-PK collision trap and making re-parks under re-delivery idempotent; `expires_at` + resume-instructions guidance for agents with no future turn.
- `_mode_guidance` task mode gains an explicit park carve-out — "execute to completion, do not ask questions" no longer contradicts the gate on the exact autonomous path it governs.
- Synced copies: `config/trinity-meta-prompt/prompt.md` (dead reference copy, injection removed in Abilityai/trinity#136) + new agent-guide async-comms section (incl. explicit honor-system framing — this is a compliance contract and audit trail, **not** a security boundary).
- Docs: `requirements/scheduling.md` §10.13 (raw-attempts / no-reset-on-progress cap semantics + rationale, cap=0 design note, the **still-blocking list for pull default-on**), `TARGET_ARCHITECTURE.md` stale claims fixed (cap was described as "all unbuilt today"), `operating-room.md` (async contract, poison-park item type, Abilityai/trinity#136 prompt-integration correction), `learnings.md` (prompt-delivery-path pitfall).

## Changes

- `src/backend/services/platform_prompt_service.py` — text-only: `PLATFORM_INSTRUCTIONS` Operator Communication rewrite + `_mode_guidance` carve-out
- `config/trinity-meta-prompt/prompt.md`, `docs/TRINITY_COMPATIBLE_AGENT_GUIDE.md` — synced contract copies
- `docs/memory/requirements/scheduling.md`, `docs/planning/TARGET_ARCHITECTURE.md`, `docs/memory/feature-flows/operating-room.md`, `docs/memory/feature-flows.md`, `docs/memory/learnings.md`
- `tests/unit/test_1402_prompt_contract.py` (+ registry) — sentinel phrases in every runtime build, old block-inviting wording gone, Operator Communication section byte-identical across runtimes (MCP-name-free), task-mode carve-out, composed end-to-end prompt, synced-copy parity

## Residual (filed, referenced from the docs)

- **#1629** — pull-claimed turns bypass platform-prompt composition (P1, blocks pull default-on) + the deterministic re-delivery banner
- **#1630** — operator-queue respond → re-trigger dispatch (no-next-turn re-entry)
- **#1631** — `operator_queue.id` global-PK cross-agent collision (schema scoping)
- **#1632** — operator-queue create rate/size caps (flooding / operator fatigue; blocking pull default-ON)

## Test Plan

- [x] `pytest tests/unit/test_1402_prompt_contract.py` — 17 passed
- [x] `pytest tests/unit/test_platform_prompt_runtime.py tests/test_platform_prompt_unit.py tests/unit/test_public_channel_prompt.py` — all passing (no assertions on the changed wording elsewhere; repo-wide grep confirmed the only other old-wording copy is a historical 2026-03 spec, left per point-in-time convention)
- [x] No code-path changes: diff touches one string constant + one pure string-returning function + docs + tests

Fixes Abilityai/trinity#1402


dolho:
Validated via `/validate-pr`. Genuinely nice piece of work — the diff is exactly what it claims (I checked: zero `def`/`return`/`if`/assignment lines change; it's one string constant, one pure string-returning function, docs, and tests), the residual issues Abilityai/trinity#1629–#1632 all exist and are open, and the "`config/trinity-meta-prompt/prompt.md` is a dead reference copy" claim holds up — it's referenced only by the new test and an *archived* deploy script, is in no Dockerfile and no runtime path. 17/17 tests pass locally. The `_mode_guidance` carve-out closes a real contradiction, and the test that composes an end-to-end task-mode prompt is the right way to prove it.

One finding, verified against the code rather than inferred.

---

### ⚠️ The contract promises an `expired` flip that no code performs

The new prompt tells every agent:

> **Set `expires_at`** on requests that gate an action. If it passes without a response the platform marks the item `expired` — treat that as "not approved; do not proceed."

`db/operator_queue.py:430` implements `mark_expired()` correctly — `UPDATE status=expired WHERE status=pending AND expires_at < now`. **Nothing calls it.** Every reference in the repo:

```
src/backend/database.py:2265        return self._operator_queue_ops.mark_expired()   ← facade delegator
src/backend/db/operator_queue.py:430    def mark_expired(self) -> int:               ← the definition
docs/memory/feature-flows/operating-room.md:389,453                                  ← docs
```

No background service, no router, no APScheduler job, no sweep in `cleanup_service`, nothing in `src/scheduler/`. The 5s `operator_queue_service` sync loop doesn't do it either. So a `pending` item past `expires_at` stays `pending` forever.

**Why it matters beyond a doc nit:**

- The failure direction is fail-safe — the agent never sees `responded`, so it never proceeds. Good. But the agent is told to *rely on* an event that will never fire, and the guidance sits on the irreversible-action path, which is the one place the contract most needs to be true.
- `pending` rows are **never retention-deleted** by design (`db/operator_queue.py:373`, Abilityai/trinity#1142 — only terminal rows are pruned, and `expired` is one of the terminal states this would produce). Every gating request an operator ignores becomes a permanent row. That compounds Abilityai/trinity#1632 (no create rate/size caps), which this PR itself flags as blocking pull default-ON.
- `operating-room.md` already documents the *propagation* half — "still-`pending` entries in the agent's file are flipped in place to their terminal DB status (`cancelled` or `expired`)" — so the write-back path for `expired` exists and is dead code for want of a producer.
- This PR propagates the same promise into two more places: `scheduling.md` §10.13 and `operating-room.md` ("an `expired` flip means 'not approved — do not proceed'").

I searched open and closed issues — nothing tracks this, so it isn't a known-and-deferred gap.

**Options**, in the order I'd rank them:
1. Wire `db.mark_expired()` into the `cleanup_service` sweep — it's an existing, tested-shaped method and a 5-min cycle is well inside the tolerance for a human-scale gate. Makes the contract true.
2. Soften the prompt to "an operator can let a gating request lapse; treat a stale request as not approved" and file the expiry sweep as a follow-up alongside Abilityai/trinity-enterprise#329/#1632.

Either is fine; shipping the sentence as-is is what I'd avoid, because the honor-system framing this PR is careful to state elsewhere ("compliance contract and audit trail, not a security boundary") makes it *more* important that the platform-side promises are literally accurate.

---

### Suggestions (non-blocking)

- **`execution_id` is conditional.** `_render_context` emits `- **Execution ID**: …` only `if ctx.execution_id`, but the contract says "Derive the `id` from your current execution ID (see the Execution Context block)" with no fallback for when it's absent. Interactive/chat contexts hit that path today, and Abilityai/trinity#1629 (pull-claimed turns bypass prompt composition) is a second way to land there. A clause like "if no Execution ID is present, use a random suffix" would close it.
- **Mixed placeholder syntax.** The JSON example uses `approval-<execution_id>-deploy` while the prose right below uses `approval-{execution_id}-{short-slug}`. The `<…>` form matches the memory block at line 151, so it's defensible — but two syntaxes three lines apart invites a literal `{short-slug}` in someone's queue file.
- The PR notes "the only other old-wording copy is a historical 2026-03 spec, left per point-in-time convention" — FWIW a repo-wide grep for `by reading the file and looking for items` now returns only the test file, so that copy appears already gone or worded differently. No action; just noting the claim is now stale in the safe direction.

---

### Validation summary

| Category | Status | Notes |
|---|---|---|
| Base branch | ✅ | `dev` |
| PR size | ✅ | 10 files (+350/−37) |
| Issue link | ✅ | `Fixes Abilityai/trinity#1402` — resolves, P2/type-feature/theme-reliability. Same-repo `Fixes` on a dev PR closes at the dev→main cut, per the documented SDLC |
| Commit messages | ✅ | conventional; the single commit explains *why* |
| Requirements | ✅ | `scheduling.md` §10.13 — cap semantics, rationale, cap=0 note, still-blocking list |
| Architecture | ➖ | no endpoint/schema/component change; `architecture.md` documents prompt *plumbing* (runtime-awareness), not prompt text |
| Feature flow | ✅ | `operating-room.md` updated + index row |
| Feature-flow format | ✅ | matches house style |
| Build/config packaging | ➖ | no new top-level module, no new `os.getenv()` |
| Code quality | ✅ | text-only, verified; scope matches the stated purpose |
| Tests | ✅ | 17/17; sentinels across all 4 runtime builds, negative assertion on the old wording, cross-runtime section parity, task-mode carve-out, composed end-to-end, synced-copy parity |

### Recommendation

**REQUEST CHANGES** — for one line:

- [ ] Either wire `db.mark_expired()` into a sweep, or reword the `expires_at` guidance (and its two doc echoes) so the contract doesn't promise platform behaviour that doesn't exist.

Everything else is ready. If you take option 2, I'd file the sweep as a follow-up next to Abilityai/trinity#1632, since the never-expiring `pending` rows and the missing create caps are the same unbounded-growth story.

obasilakis (changes requested):
Checked the contract against the code rather than just the prose, and it holds up:

- `expires_at → expired` is real — `mark_expired` runs in the 5s sync loop (`operator_queue_service.py:106`), so the prompt isn't promising behaviour the platform lacks.
- `execution_id` is populated and rendered on the autonomous path (`task_execution_service.py:1170` → `platform_prompt_service.py:525`), so the derived-id instruction resolves.
- The collision rationale is well-founded: `operator_queue.id` is `Column("id", Text, primary_key=True)` and `create_item` does `.on_conflict_do_nothing(index_elements=["id"])` — a colliding id genuinely is swallowed silently.
- `Fixes #1402` is legitimate: the #1550 mechanism (`MAX_REDELIVERY`, `redelivery_count`, alert-first poison-park) is on dev in `lease_reaper_service.py`, so this contract surface completes the ACs.

Tests: 17 passed, and still pass rebased onto current dev. One necessary change before merge.

## The #1629 residual claims are stale — that fix merged to dev a minute before this PR opened

Three docs assert present-tense code state that is false on the merge target:

1. `docs/memory/requirements/scheduling.md` §10.13, in the **"Still blocking pull default-on"** list:
   > pull-path platform-prompt delivery + re-delivery banner (#1629 — pull-claimed turns currently receive NO platform instructions at all)
2. `docs/planning/TARGET_ARCHITECTURE.md`:
   > the pull-path *delivery* of that prompt (pull-claimed turns receive no platform instructions today) is **#1629**, still blocking default-on for effect-bearing agents
3. `docs/memory/feature-flows/operating-room.md`:
   > **Known gap (#1629)**: pull-claimed turns bypass platform-prompt composition entirely and currently receive none of these instructions

`f6567b56` ("fix(pull): compose the platform system prompt on the pull-claim path + re-delivery banner", #1633) merged to dev at **14:00:26Z**; this PR opened at **14:01:39Z**. On dev, `pull_coordination_service.py:87` calls `compose_system_prompt(...)` for pull-claimed turns, and #1633 shipped the re-delivery banner alongside it. Pure timing — the branch predates the fix by about a minute.

Citing #1629 as a tracked residual is still correct: it's `status-in-dev`, open only until the release cut. What needs changing is the present-tense code claims and its place in the still-blocking list. That list is the gate for defaulting pull mode on, so a resolved blocker sitting in it means the next reader either redoes #1629's work or mis-sequences the epic — which is exactly the kind of thing these docs exist to prevent.

Suggested wording: `#1629 — fix merged to dev (#1633), pending release`, and drop it from the still-blocking list. The remaining blockers there are unaffected (#548/#333 trace fidelity, #1401 `prior_trace` injection, fail-closed `execution_id`, #1632 queue caps).

Worth noting the irony: this PR's own `learnings.md` entry is precisely about the pull path not receiving the platform prompt. That lesson stands on its own — it's what caused #1629 to be filed and fixed. Only the "still open / never delivered" framing needs updating.

## Rebase needed

The branch conflicts with dev on `tests/registry.json` only, and it's additive on both sides (dev's #1600/#1649/#1615 entries vs this PR's #1402 entry). I rebased locally to check: resolution is keep-both (80 dev entries + 1 = 81, no duplicate files), and the prompt + pull suites pass afterwards — 68 passed, 1 skipped.

Everything else looks good. The sentinel-lock approach, the cross-runtime byte-identity check, and the meta-prompt/agent-guide parity tests are a good fit for a contract that lives in prose.

obasilakis (approved):
Resolving my own changes-request: I rebased the branch onto current dev and pushed the fix myself (e2d182ac).

**Rebase** — only `tests/registry.json` conflicted, additive on both sides. Resolution is keep-both: dev's 80 entries + this PR's #1402 entry = 81, no duplicate files. The feature commit is unchanged at +350/−37.

**The fix** (docs only) — corrected the three stale #1629 claims to say shipped-pending-release instead of unbuilt, and moved it out of the "Still blocking pull default-on" list in both `scheduling.md` and `TARGET_ARCHITECTURE.md`. #1629 stays referenced (it's `status-in-dev`, open until the release cut). `learnings.md` I deliberately left untouched — it records *why* #1629 was filed, which is accurate history and still the durable lesson.

**Verified after the rebase**: `test_1402_prompt_contract.py` + `test_platform_prompt_runtime.py` + `test_1081_pull_endpoints.py` → 68 passed, 1 skipped.

Since I both requested and made this change, please sanity-check my wording on the blocker lists before merging — I'd rather you disagree with the phrasing now than inherit it. Approving on the substance: the contract is well-founded (I verified `expires_at → expired` fires, `execution_id` resolves, and the global-PK collision is real), and the sentinel-lock + cross-runtime byte-identity tests are a good fit for a contract that lives in prose.

Issue #1644, opened 2026-07-15
bug: retention prunes have no blast-radius guard — a mistyped window deletes most of a table within 5 minutes


vybe:
## Summary

Split out from #1638 (deliberately deferred there to keep a P0 data-loss fix small).

#1638 fixed the specific way a *default change* destroyed data. It did not add any guard on the prune itself. Nothing in `cleanup_service` bounds destruction relative to what exists: `_sweep_retention_772` deletes up to `RETENTION_CHUNK_SIZE_PER_CYCLE = 5000` rows per cycle, per sweep, unconditionally, and re-sweeps every 300s.

So the remaining hole is the operator path: an admin who types `5` instead of `50` into `PUT /api/settings/ops/config` (no validation, no clamp) has most of their execution history gone within 5 minutes, silently. #1638's inversion makes the *default* safe; it does nothing about a written row.

## Why this is the durable fix

#1638's fix is specific to one mechanism. A blast-radius guard catches:
- a mistyped retention window (the operator path above)
- any *future* default change that slips the #1638 discipline
- the `agent_soft_delete` path, where expiry destroys agent data volumes (#1581) — unrecoverable

It's the only proposed mitigation whose value doesn't expire.

## Suggested approach

Before any prune, compare the candidate count against the table count. If a single cycle would delete more than N rows **or** more than X% of the table:

1. **Refuse** — do not delete.
2. Log at ERROR with the resolved window, its source (`db-row`/`code-default`), and the candidate/total counts.
3. Raise an `operator_queue` item (OPS-001 already exists and is exactly the async human-gate primitive this needs — there is no synchronous operator at boot).
4. Proceed only on explicit acknowledgment.

Note the interaction with chunking: `RETENTION_CHUNK_SIZE_PER_CYCLE` currently splits one catastrophic delete into a long sequence of individually-unremarkable log lines, which is part of why #1638 was invisible. The guard should evaluate against the *total* candidate set, not the per-cycle chunk.

## Acceptance criteria

- [ ] A prune that would exceed the threshold refuses and logs at ERROR
- [ ] An `operator_queue` item is raised naming the agent/table, the window, and the row counts
- [ ] Acknowledgment allows the prune to proceed
- [ ] Threshold is configurable, with a sensible default, surfaced in Settings
- [ ] Guard evaluates the total candidate set, not the per-cycle chunk
- [ ] Test: an over-threshold prune deletes nothing and queues an item
[...]

Issue #1581, opened 2026-07-12
bug: Docker volumes are never deleted — volume_remove has zero callers; agent purge leaks workspace/public/shared volumes forever


vybe:
## Problem

`services/docker_utils.py::volume_remove` has **zero callers** repo-wide. No lifecycle path — soft-delete (`DELETE /api/agents/{name}`), `cascade_delete`, or the 180-day retention hard-purge (`purge_agent_ownership` via `cleanup_service._sweep_soft_deleted_agents`) — removes any Docker volume. `agent-{name}-workspace`, `agent-{name}-public`, and shared-folder volumes survive even a full hard-purge and leak indefinitely: an unbounded disk leak on every Trinity install, proportional to agent churn.

Found during the trinity-enterprise#69 design review (ephemeral ghost agents sidestep it by being volume-less; durable agents remain affected).

## Fix sketch

- At retention hard-purge time (the moment the agent is unrecoverable), remove the agent's named volumes: workspace + public + shared (labels `trinity.platform=agent-workspace|agent-public|agent-shared` + `trinity.agent-name`).
- Double-guard the destructive call: label AND `agent-{name}-*` naming must both match; refuse otherwise.
- Tolerate in-use/NotFound errors (retry next sweep); consider a one-shot reconciliation sweep for volumes whose agent row no longer exists (bounded per cycle).
- Do NOT remove at soft-delete — recovery within the retention window must keep the workspace.

## Acceptance criteria

- [ ] Retention purge removes the purged agent's workspace/public/shared volumes
- [ ] Orphan volumes (no ownership row, past grace) reclaimed by a bounded sweep
- [ ] Destructive call double-guarded (label + name); durable-agent volumes unreachable by a bug
- [ ] Unit tests incl. in-use/NotFound tolerance

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1065, opened 2026-06-04
feat(retention): OSS 5-day community floor + prod-compose LOG_* fix (#1039 — OSS floor)

dolho:
## Summary

**Stage 1 of Abilityai/trinity-enterprise#31 — the OSS floor** (the issue's "must land first" chunk). The enterprise `retention` module (private submodule) + Settings → Retention UI follow as separate PRs on the Abilityai/trinity#847 seam.

## Prod bug fixed (silent no-op)
`docker-compose.prod.yml` omitted `LOG_RETENTION_DAYS` / `LOG_ARCHIVE_ENABLED` / `LOG_CLEANUP_HOUR` from `backend.environment:`. Prod launches standalone (`-f docker-compose.prod.yml`, no base-compose merge), so an operator-set `LOG_RETENTION_DAYS` never reached the container and retention always fell back to the code default. **Added the three lines.**

## 5-day community floor
Lowered the operator-tunable defaults to the 5-day community floor:

| Setting | Was | Now |
|---|---|---|
| `LOG_RETENTION_DAYS` | 90 | 5 |
| `execution_log_retention_days` | 30 | 5 |
| `execution_row_retention_days` | 90 | 5 |
| `health_check_retention_days` | 7 | 5 |
| `agent_soft_delete_retention_days` | 180 | 5 |
| `schedule_soft_delete_retention_days` | 30 | 5 |
| `audit_log_retention_days` | 365 | **365 (exempt — integrity floor)** |

New `COMMUNITY_RETENTION_FLOOR_DAYS` / `RETENTION_OPS_KEYS` constants — the enterprise module reuses these to clamp unentitled writes.

## Read surface
`GET /api/settings/retention` (admin) reports the effective windows in use + the active edition (`community` vs `enterprise` via the `retention` entitlement) + documented precedence: **enterprise (license) DB setting → env → 5-day community default**.

## Design note
OSS does **not** hard-clamp env/OPS values — they remain an unsupported self-host escape hatch (per the issue: *"not a cryptographic lock on a constant"*). The clamp-to-floor lives in the enterprise `retention` module's managed setter (Stage 2).

## ⚠️ Release-notes call-out
This sharply shortens **soft-delete recovery** on community installs (agent 180→5, schedule 30→5). Deliberate per the issue; an enterprise license restores longer windows.

## Verification
- `tests/unit/test_retention_floor.py` — **7/7** (floor defaults, audit exemption, read-surface edition + windows, audit 365-floor, enterprise-set OPS window).
[...]

PR #1645, opened 2026-07-15
fix: apply retention floor by seeding, not by lowering prune-time defaults (#1638)

### LINKED ISSUE #1638: bug: retention floor (#1065) silently deletes pre-existing execution history on upgrade — currently dev-only, ships to everyone at next release
## Summary

#1065 (`43411375`, currently **`dev`-only — not in any release**; see *Release exposure* below) lowers five community retention defaults to **5 days**. Because `cleanup_service` resolves each window **DB-row-first with the code default as fallback**, and runs a prune as a **startup sweep**, any pre-existing install that has no `system_settings` rows for those keys — which is the default state, since nothing writes them — hard-deletes its execution history within ~20 seconds of its first boot on the affected code. Silently, with `/health` green.

**This is not a dispute of the community floor itself.** That is a deliberate open-core decision and is documented in-code. The defect is narrower: a **default change is applied retroactively and destructively to data that already exists**, with no migration entry, no warning, and no window to react.

Observed on a managed production instance: **5352 of 5593 `schedule_executions` rows (95.7%) — roughly 3 months of history — deleted 20 seconds after restart.** It was recoverable only because an unrelated pre-update backup happened to predate the sweep by 16 minutes. A site without that backup loses the data permanently and has no way to know it happened.

## Release exposure — no released user has been affected (corrected)

**This report originally said "v0.8.0". That was wrong, and the correction matters.** `43411375` landed **2026-07-14**, six days *after* the `v0.8.0` tag (`b3b7078b`, 2026-07-08):

```
compare v0.8.0...43411375  ->  status: ahead, ahead_by: 646, behind_by: 0
compared against all 13 tags -> "ahead" in every case  (contained in ZERO releases)
main...43411375            ->  diverged (not on main either)
```

So the floor exists **only on `dev` and feature branches**. **No released user has ever been exposed**, and the affected instance was tracking `dev`, not the release. Severity is unchanged — it hits every user the moment the next release is cut — but the cohort is currently dev-trackers only, which means **this is fixable before it ever reaches anyone**. That is the good-news framing, and the reason to prefer the backfill fix below over a post-hoc recovery path.

### Why it was misreported as v0.8.0 (a real DX trap worth its own look)

`VERSION` on `dev` reads **`0.8.0`** — byte-identical to the `v0.8.0` tag — because it isn't bumped until the next release cuts. So `GET /api/version` returns `0.8.0` on an instance 646 commits past the tag, and is **indistinguishable from the actual release**. Anyone triaging by reported version (operators, support, this report) will mis-attribute dev-line behavior to a shipped version. Consider a dev/pre-release marker (`0.8.0+dev.<sha>`, or surfacing `git_branch` alongside `version` in the Build Info dialog) so the two are separable.

## Component

Backend — `cleanup_service` + `settings_service`

## Priority

**P0** — irreversible data loss on a routine upgrade, with no operator signal.

## Error

There is no error, and that is precisely the problem. The upgrade succeeds, health checks pass, and the only trace is an INFO line:

```
[Cleanup] Deleted 5108 schedule_executions rows older than 5 days (#772)
[Cleanup] Deleted  244 schedule_executions rows older than 5 days (#772)
[Cleanup] Startup sweep: {'execution_logs_pruned': 1223, 'execution_rows_pruned': 5108, ..., 'total': 6331}
```

(Two lines because each worker sweeps independently.)

## Location

- `src/backend/services/settings_service.py:72` — `OPS_SETTINGS_DEFAULTS`, the five lowered windows
- `src/backend/services/settings_service.py:118` — `OPS_SETTINGS_DESCRIPTIONS`, which still advertises the **old** defaults (see factor 4)
- `src/backend/services/cleanup_service.py:~140` — `_read_retention_settings()`, `db.get_setting_value(key, OPS_SETTINGS_DEFAULTS.get(key, "0"))`
- `src/backend/services/cleanup_service.py` — `_cleanup_loop()` runs `run_cleanup()` as a startup sweep

## Root Cause

Four factors compound. Individually each is defensible; together they make silent data loss the default outcome of an upgrade.

**1. A lowered default is applied retroactively to existing data.** The five windows:

| Key | Pre-#1065 | #1065 |
|-----|-----------|-------|
| `execution_log_retention_days` | 30 | 5 |
| `execution_row_retention_days` | **90** | **5** |
| `health_check_retention_days` | 7 | 5 |
| `agent_soft_delete_retention_days` | **180** | **5** |
| `schedule_soft_delete_retention_days` | 30 | 5 |

An install that never touched retention has no rows, so it silently inherits the new floor and its existing history is destroyed. Changing a default for *new* installs is reasonable; rewriting the history of *existing* ones is a different act.

**2. The prune runs as a startup sweep.** `_cleanup_loop()` calls `run_cleanup()` immediately, so deletion happens ~20s after boot rather than at some later off-peak hour. Even a well-informed operator has no window to intervene between "backend is up on the new version" and "history is gone."

**3. No `docs/migrations/` entry.** This is the most consequential omission. `docs/migrations/` is the established convention for operator-facing changes, and tooling reads it to surface required actions before a restart. The most destructive change in v0.8.0 ships without one, so no upgrade tooling can surface it. The #1039 release note is not a substitute — it isn't in the path an upgrade actually consults.

**4. The module contradicts itself.** `OPS_SETTINGS_DEFAULTS` was updated to `"5"`, but `OPS_SETTINGS_DESCRIPTIONS` still says:

```python
"execution_row_retention_days": "Days to retain finished schedule_execution rows; rows older than this are deleted (default: 90, 0 = disabled, #772)",
"agent_soft_delete_retention_days": "Days to retain soft-deleted agents before hard-purge (default: 180, 0 = disabled, #834)",
"execution_log_retention_days": "Days to retain the JSONL transcript on schedule_executions (default: 30, 0 = disabled, #772)",
```

Anywhere these descriptions surface, an operator is told the default is 90 while the code deletes at 5.

**5. Mass deletion logs at INFO.** Destroying 95% of a table is reported at the same level as routine no-op cleanup, so it does not stand out in logs or trip any alerting.

Note that an enterprise `retention` entitlement does **not** mitigate this: `enterprise/backend/retention/service.py` mirrors the same floor (`_OPS_DEFAULTS = {k: "5"}`) and only adds a managed setter that clamps to `max(value, 5)`. An entitled instance upgrading is exposed identically.

## Reproduction Steps

1. Run any instance on code predating `43411375` for more than 5 days so `schedule_executions` accumulates history. Do not set any retention settings — this is the default state; nothing writes those rows.
2. Confirm none exist: `GET /api/settings` returns no `*_retention_days` keys.
3. Update to `dev` at or past `43411375` (or any future release containing it) and restart the backend.
4. Within ~20 seconds, every execution older than 5 days is hard-deleted. Health stays green; no error is raised.
5. `SELECT MIN(started_at) FROM schedule_executions` now returns a timestamp exactly 5 days old, and re-sweeps every 300s keep it there.

## Suggested Fix

**Preferred — backfill on upgrade, so the floor applies to new installs only.** When a retention key has no row and the install predates the floor, write the previous default rather than silently applying the new one. This fully preserves the open-core intent (fresh installs get the 5-day community floor) without retroactively deleting existing users' data:

```python
# One-off migration / startup reconciliation
_PRE_1065_DEFAULTS = {
    "execution_log_retention_days": "30",
    "execution_row_retention_days": "90",
    "health_check_retention_days": "7",
    "agent_soft_delete_retention_days": "180",
    "schedule_soft_delete_retention_days": "30",
}
# Only for installs that existed before the floor landed — a fresh DB gets the new defaults.
for key, prior in _PRE_1065_DEFAULTS.items():
    if db.get_setting_value(key, None) is None:
        db.set_setting(key, prior)
```

**If the retroactive floor is intentional and must stand, then at minimum:**

1. **Ship a `docs/migrations/` entry.** This alone would have prevented the incident — it is the file upgrade tooling reads.
2. **Do not prune destructively on the first boot after an upgrade.** Skip the startup sweep when the resolved window came from a code default rather than an explicit row, and log loudly instead.
3. **Log destructive prunes at WARNING**, including the row count and resolved window, so mass deletion is visible.
4. **Fix `OPS_SETTINGS_DESCRIPTIONS`** to match `OPS_SETTINGS_DEFAULTS` (factor 4) — this is an unambiguous inconsistency regardless of the outcome here.

A dry-run/report mode (`retention_dry_run=true` logging what *would* be deleted) would also let operators see the blast radius before committing.

## Environment

- Affected commit: `43411375` (`feat(retention): OSS 5-day community floor + prod-compose LOG_* fix (#1039 — OSS floor) (#1065)`), authored 2026-07-14
- Branch: **`dev`** — **not** contained in `v0.8.0` (646 commits behind it) nor any other tag; **not on `main`**
- Reported version on the affected instance: `0.8.0` — misleading, see *Release exposure* (the dev `VERSION` file still reads `0.8.0`)
- Backend: PostgreSQL 16 (also applies to SQLite — the resolution path is backend-agnostic)
- OS: Ubuntu 22.04 / Docker

## Related

- `src/backend/services/settings_service.py` (`OPS_SETTINGS_DEFAULTS`, `OPS_SETTINGS_DESCRIPTIONS`)
- `src/backend/services/cleanup_service.py` (`_read_retention_settings`, `_cleanup_loop`)
- #1039 (OSS retention floor), #1065 (this change), #772 (execution/health retention), #834 (soft-delete retention)

vybe:
## Summary

- **The 5-day community retention floor (#1039/#1065) was applied by lowering `OPS_SETTINGS_DEFAULTS`** — but those are read at *prune* time as the fallback for any install with **no** `system_settings` row, which is the default state since nothing writes them. The dict wasn't "the default for new installs"; it was live, retroactive policy for every install that never opted in, evaluated against a hard `DELETE` on a **startup sweep**. A real instance lost **95.7% of `schedule_executions`** (~3 months) with a green `/health` and one INFO line.
- **Never shipped in a release.** `43411375` is not an ancestor of `v0.8.0` (the tag predates it by six days) and is contained in **no tag** — only `dev`-tracking instances were ever exposed. This fixes it before it reaches a single release user. (The issue's `Environment: v0.8.0` is inaccurate.)
- **Inverts the mechanism rather than backfilling.** Code defaults return to their wide historical values and become a **safety floor, not a policy knob**; the community floor now reaches **fresh installs only**, by seeding rows into an empty DB.

## Why invert instead of the backfill migration the issue proposed

The decisive argument is **failure direction**:

| | default = 5 (backfill) | default = 90 (this PR) |
|---|---|---|
| Misclassify existing as fresh | **Data destroyed** | Harmless |
| Miss the migration entirely | **Data destroyed** | Safe |
| `/ops/reset` clicked | **Data destroyed** | Safe |
| Future default edit | **Data destroyed** | Safe |

With the default at 5 every hole fails destructive; at 90 every hole fails safe. Inverting is also **less code**: no fresh-vs-existing discriminator, no dual-track SQLite+Alembic pair, no idempotency crash-loop risk.

**No open-core cost.** The floor was never enforced: any admin can widen a window via `PUT /api/settings/ops/config` (no clamp), and the enterprise clamp is a **minimum** (`max(value, 5)`). Enterprise sells the managed panel, not the capability — so keeping existing data concedes nothing. New installs still get 5-day windows.

## Changes

| File | What |
|---|---|
| `config.py` | `COMMUNITY_FRESH_INSTALL_SEED` + floor constant (leaf module — `settings_service` imports `db` from `database`, so defining them there is circular) |
| `database.py` | `_seed_fresh_install_retention{,_engine}` — seeds a fresh DB only; fail-safe + idempotent |
| `settings_service.py` | Defaults back to 30/90/7/180/30; re-exports the constants |
| `cleanup_service.py` | Boot-time effective-window + **source** log before the first sweep; WARNING on cap-hitting prunes; dead fallback literals removed |
| `routers/settings.py` | `/ops/reset` skips retention keys; `/retention` reports real precedence + per-key `source` |
| `docs/migrations/RETENTION_DEFAULTS_2026-07.md` | Runbook, leading with restore-from-backup |

**Seeding contract:** runs inside `init_database()` after `init_schema` and **before** `_ensure_admin_user` — an empty `users` table is the fresh-install signal. Never raises (`init_database()` runs at *import*; a raise would crash-loop boot permanently) and is idempotent (`INSERT OR IGNORE` / `on_conflict_do_nothing` — both migration locks fail open, so two workers can race it).

**`agent_soft_delete_retention_days` is now exempt from the floor in every edition.** Its expiry chains `purge_agent_ownership` → `remove_agent_volumes` (#1581), destroying agent workspaces and `data_paths` runtime data (#1169). That's a *recovery* window, not a log window — 180 → 5 was never a deliberate decision.

## Also fixed (each one independently re-armed the bug)

- **`POST /api/settings/ops/reset` deleted every OPS row it knew about** — "reset to defaults" was one admin click from re-arming mass deletion.
- **`/api/settings/retention` advertised `"enterprise → env → community-default"`** — no env layer exists for these five keys (grep `EXECUTION_ROW_RETENTION_DAYS`: zero hits). The one documented mitigation an operator could reach for did nothing.
- **`OPS_SETTINGS_DESCRIPTIONS` told the admin UI "default: 90" while the code deleted at 5.** Now pinned by a test.
- **`test_retention_floor.py` asserted the destructive default *as the requirement*** (`assert OPS_SETTINGS_DEFAULTS[key] == "5"`). That's why #1065 shipped green — rewritten to assert the inverse.

## Test Plan

- [x] `pytest tests/unit/test_1638_retention_upgrade_safety.py tests/unit/test_retention_floor.py` — 22 passed
- [x] Full affected surface (retention/cleanup/settings), random order — **76 passed**
- [x] **Mutation-tested**: re-introducing the bug (`execution_row_retention_days` → `"5"`) fails **4 tests**
- [x] **Mutation-tested**: removing the `/ops/reset` skip fails its guard test
- [x] Byte-compile + secret scan clean

Covered: fresh → seeded; existing → never seeded (the regression); soft-delete exempt; idempotent under re-run; never clobbers an operator value; never raises on a broken DB; reset can't strand a window; descriptions match defaults; `init_database` pinned to import-time (moving it into the lifespan would put the seed into an ordering contest with `cleanup_service` it currently cannot lose).

## Deferred (filed, not dropped)

- **#1644** — blast-radius guard: refuse a prune that would delete >N% of a table, escalate via operator-queue. The **durable** fix (catches operator fat-fingers and any future default change); deferred to keep a P0 data-safety fix small.
- **abilityai/trinity-enterprise#168** — enterprise `_OPS_DEFAULTS` still mirrors the floor; reports 5d while OSS prunes at 90d (reporting divergence only).

Fixes #1638


AndriiPasternak31:
LGTM

AndriiPasternak31 (approved):

Issue #175, opened 2026-03-26
SEC: SSH Private Key and Password Transmitted Plaintext in API Response


vybe:
## Pentest Finding 3.1.4 — High (CVSS 8.5)

**CVSS Vector**: [CVSS:4.0/AV:L/AC:L/AT:N/PR:L/UI:N/VC:H/VI:H/VA:H/SC:N/SI:N/SA:N](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:L/AC:L/AT:N/PR:L/UI:N/VC:H/VI:H/VA:H/SC:N/SI:N/SA:N)

**Location**: `POST /api/agents/{name}/ssh-access` — `services/ssh_service.py` lines 56–83, `routers/agents.py` line 1368

## Impact

When SSH access is enabled with `auth_method=key`, the server generates an SSH key pair entirely server-side and returns the **private key in plaintext** in the HTTP response body. The private key is:
- Transmitted over HTTP (unencrypted by default)
- Potentially logged by reverse proxies
- Stored in browser history
- Accessible in any intermediate system between client and server

Exploitation consequences:
- **Docker network MitM** — compromise of any container on the same Docker network yields the private key
- **Host-level attacker** — `nsenter` or `tcpdump` on the host captures all internal Docker HTTP traffic
- **Key impersonation** — server retains every private key ever issued without client knowledge
- **Container takeover** — SSH key grants developer shell access to agent containers with credentials, `.env` files, `.mcp.json`
- **Credential pivot** — once inside a container, injected credentials are directly accessible at `/home/developer/.env`

## Root Cause

`services/ssh_service.py` lines 56–83 generate the keypair server-side with `NoEncryption()` and return the private key bytes directly. `routers/agents.py` line 1368 includes `private_key` verbatim in the JSON response.

## Steps to Reproduce

```bash
POST /api/agents/shared/ssh-access
Authorization: Bearer {USER_TOKEN}
# Response includes full unencrypted private key in "private_key" field
```

## Remediation

- Remove server-side keypair generation entirely — change the API to accept a client-supplied `public_key` parameter
- Server only calls `inject_ssh_key(public_key)` to append to `authorized_keys`; private key never leaves the client
- If server-side generation must be kept for UX reasons, mandate TLS-only deployment and suppress logging of the response body
- Enforce owner-only (not just `can_user_access_agent`) access on the endpoint

## References

[...]

PR #1647, opened 2026-07-16
fix(ssh): remove the broken password auth method from ssh-access (#1615)

### LINKED ISSUE #1615: bug: password ssh-access is broken end-to-end — import crypt crashes on Python 3.13 and agent sshd disallows password auth anyway
## Summary

`POST /api/agents/{agent_name}/ssh-access` with `auth_method: "password"` always returns HTTP 500. The handler imports the stdlib `crypt` module, which was removed in Python 3.13 (PEP 594) — the backend image runs Python 3.13, so the import raises before any work happens. Independently, even if the hash were set, the agent base image's sshd runs with `PasswordAuthentication no`, so password login can never succeed. The password option is dead code that presents as a working choice in the API, MCP tool (`get_agent_ssh_access`), and UI.

## Component

Backend / SSH Service + Agent Runtime (base image sshd)

## Priority

P2 (key-based auth works as the workaround; password option fully broken)

## Error

```
File "/app/routers/agent_ssh.py", line 97, in create_ssh_access
    if not await ssh_service.set_container_password(agent_name, password):
File "/app/services/ssh_service.py", line 125, in set_container_password
    import crypt
ModuleNotFoundError: No module named 'crypt'
```

## Location

- **File**: `src/backend/services/ssh_service.py`
- **Line**: ~125 (function-level `import crypt` in `set_container_password`)
- **Also**: agent base image sshd config — containers run sshd (started via `docker/base-image/startup.sh`) with `PasswordAuthentication no` / `KbdInteractiveAuthentication no` / `UsePAM yes`

## Root Cause

1. `crypt` was deprecated in Python 3.11 and **removed in 3.13**; the backend image is on 3.13 (`/usr/local/lib/python3.13/`). The function-level import crashes every password-method request with a 500.
2. Even with a hash set via `usermod -p`, the in-container sshd rejects password authentication by configuration, so the feature cannot work as shipped.

## Reproduction Steps

1. Deploy Trinity with the Python 3.13 backend image
2. Start any agent
3. `POST /api/agents/{agent}/ssh-access` with body `{"auth_method": "password", "ttl_hours": 1}`
4. Observe HTTP 500 with the `ModuleNotFoundError` above in backend logs
5. (Even after patching the import: `ssh developer@host -p <agent_port>` with the password → `Permission denied` because sshd disallows password auth)

## Suggested Fix

Preferred: **remove the `password` auth method** from the API, MCP tool, and UI — the BYOK key flow is strictly more secure (the tool description already calls key auth "more secure"), and keeping a never-working option misleads users into long debugging sessions.

If password auth is to be kept instead:
- Drop `import crypt`; the function already contains a `chpasswd` fallback — make that the primary path (no host-side hashing needed), or hash inside the container with `openssl passwd -6`.
- Ship an sshd drop-in (`/etc/ssh/sshd_config.d/`) in the agent base image enabling `PasswordAuthentication yes` for the `developer` user only.

## Environment

- Trinity version: `a88f5fc8`
- Backend image: Python 3.13
- OS: Ubuntu (GCP VM), Docker prod compose

## Related

- `src/backend/routers/agent_ssh.py` (`create_ssh_access`)
- `src/backend/services/ssh_service.py` (`set_container_password`)
- `docker/base-image/startup.sh` (sshd startup)

dolho:
## Summary

Takes the issue's **preferred fix**: remove the `password` auth method rather than repair it. Password SSH was broken end-to-end and presented as a working choice — a user picking it got a 500 and a long debugging session for a path that could never have worked.

Two independent reasons it was unfixable-as-shipped:
1. `set_container_password` did a function-level `import crypt`. `crypt` was removed from the stdlib in **Python 3.13** (PEP 594) and the backend image is on 3.13 → `ModuleNotFoundError` before any work happened.
2. Even with a hash set, the agent base image's sshd runs `PasswordAuthentication no` (`docker/base-image/Dockerfile:92`) → password login can never succeed.

Key auth (BYOK) is strictly more secure and already the documented-better path — the MCP tool description called it "more secure" while offering the broken alternative alongside it.

## Changes

**Surface removal**
- `routers/agent_ssh.py` — `auth_method` other than `"key"` → **400** naming key auth as the alternative. Placed before any Docker work.
- `services/ssh_service.py` — `generate_password()` and `set_container_password()` deleted (with them, the `crypt` import).
- `models.py` — `SshAccessRequest.auth_method` documented as key-only.
- MCP (`tools/agents.ts`, `client.ts`, `types.ts`) — `auth_method` gone from the tool surface; `SshConnectionInfo.password` removed; `auth_method: "key"`.
- **No UI change** — there is no SSH frontend surface (the issue lists one; a repo-wide grep finds none). The endpoint is admin/API-only.

**`clear_container_password` is deliberately kept** — cleanup-only. A backend on Python <3.13 *could* set a password successfully, so a pre-upgrade Redis row (`auth_type="password"`) may still be in flight. Those rows carry a TTL of ≤24h, so the branch is dead within a day of deploy; it exists so the last legacy credentials get locked rather than left set. Nothing writes `auth_type="password"` any more. This is now stated in the code and the flow doc — the surviving references otherwise read as an incomplete removal.

**Tests** — `tests/unit/test_1615_ssh_password_removed.py` (12). The existing suite guards the *service* layer; nothing pinned the router, which is the surface that produced the 500. The handler is a plain `async def`, so these call it directly rather than fighting FastAPI's DI.

The headline test asserts the **reason**, not just the status: without the guard a password request falls through to the key path and 400s on the missing `public_key` — right code, wrong reason, and the guard could be deleted with the test still green. Reverting the guard fails **8 of 12**.

**Docs** — `feature-flows/ssh-access.md` documented password as supported and embedded code for three deleted functions; rewritten around the one remaining flow (also fixes a stale `#175` claim that the response still returns `private_key`). `async-docker-operations.md` cited `set_container_password`. `architecture.md`'s API row now names key-based/BYOK.

## Verification

Unit: **29 pass** (12 new + 13 existing ssh + models-centralized). `tsc --noEmit` clean.

Exercised against a live agent container, not just mocks:

```
# the exact repro from the issue — was 500 ModuleNotFoundError: crypt
POST /api/agents/portal-client/ssh-access {"auth_method":"password","ttl_hours":1}
→ HTTP 400 {"detail":"Password SSH auth is no longer supported. Use key-based auth:
   generate a key locally (ssh-keygen -t ed25519) and pass its public_key."}

# key auth still works, end to end
POST ... {"auth_method":"key","public_key":"ssh-ed25519 AAAA..."}  → HTTP 200
  auth_method: key | user/port: developer/2228
  private_key in response: False | password in response: False

ssh -i <privkey> -p 2228 developer@localhost 'whoami@hostname'
→ SSH KEY AUTH OK: developer@ad52a380de0d
```

Agent sshd confirmed `PasswordAuthentication no` / `PubkeyAuthentication yes`. Test key revoked from `authorized_keys` afterwards.

## Test Plan

- [x] `pytest tests/unit/test_1615_ssh_password_removed.py tests/unit/test_ssh_service.py` — 25 passed
- [x] New tests verified to fail against the unfixed code (8/12 on guard removal)
- [x] `npx tsc --noEmit` in `src/mcp-server` — clean
- [x] Live: password → 400, key → 200, real SSH login succeeds

Related to #1615

Fixes #1615


vybe (approved):
Validated via /validate-pr: takes the issue's preferred fix (remove, don't repair) — password auth was doubly broken (crypt removed in Python 3.13; agent sshd runs PasswordAuthentication no). Three surfaces updated in sync (backend router/model/service, MCP client/tool/types, docs incl. architecture.md endpoint table). Named regression tests (unit test_1615_* + live API/e2e suite) cover the 400-not-500 refusal, key-auth default, and no-private-key/password-leak response shape. The kept clear_container_password cleanup path for pre-upgrade Redis rows is correctly reasoned (TTL-bounded, dead within a day). Resolved the registry.json append conflict (all entries kept, JSON validated). Security battery clean. Approving.

PR #1637, opened 2026-07-15
refactor(frontend): remove legacy /enterprise/client-portal, point card to /portal

dolho:
## What

The enterprise dashboard **Client Portal** card linked to the superseded operator-auth portal (`/enterprise/client-portal` → `views/enterprise/ClientPortal.vue`). This:

1. **Fixes the link** — the card now points to the current public client portal `/portal`.
2. **Removes the old portal** — deletes the `/enterprise/client-portal` route and its two components (`ClientPortal.vue`, `PortalChat.vue`).

## Unchanged (the active portal)

`views/Portal.vue`, `components/portal/*`, `stores/clientPortal.js`, and the backend `/api/enterprise/client-portal/*` endpoints — all still used by the current `/portal` surface. No backend change.

## Verification

- `vite build` passes (only a pre-existing chunk-size warning).
- Live dev-server preview: served `router/index.js` no longer contains the old route; `Index.vue` card resolves to `/portal`; the two deleted `.vue` modules 404 (SPA fallback); `Portal.vue` still serves.
- No dangling references to the removed route/components anywhere in `src/frontend`.

## Follow-ups (not in this PR)

- Operator-feedback UX when a portal sign-in email has no access (silent no-leak gate leaves operators with no signal).
- `DATABASE_URL` empty → silent SQLite fallback (an instance can believe it's on Postgres while running on SQLite).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Fixes #1652

PR #1648, opened 2026-07-16
fix(ui): restore the top nav on the enterprise pages (#1636)

### LINKED ISSUE #1636: bug: Enterprise pages lose the top nav bar
## Summary

Navigating to the Enterprise page drops the main top navigation. The user lands on a page with no way back to the rest of the platform except the browser back button. The nav bar should stay pinned at the top like it does on every other authenticated page.

## Context

Trinity has no global chrome — `App.vue` renders only `<router-view>`, so every authenticated view is responsible for importing and rendering `components/NavBar.vue` itself. `Dashboard.vue`, `Settings.vue`, `Operations.vue`, and `Templates.vue` all do.

None of the views under `src/frontend/src/views/enterprise/` reference `NavBar` at all, so the chrome silently disappears on those routes:

| Route | View | NavBar |
|-------|------|--------|
| `/enterprise` | `views/enterprise/Index.vue` | ❌ missing |
| `/enterprise/audit` | `views/enterprise/Audit.vue` | ❌ missing |
| `/enterprise/client-portal` | `views/enterprise/ClientPortal.vue` | ❌ — likely intentional (standalone client surface, see below) |

`/enterprise/client-portal` is a client-facing surface (its route already sets `hideHelpWidget: true` with a comment that the operator help widget "isn't relevant to a portal client"), so it is probably meant to be standalone like `/portal` and `/m`. Confirm that intent while fixing the other two rather than blanket-adding NavBar everywhere.

The per-view NavBar convention is the underlying reason a new view can silently ship without chrome. Fixing that convention (hoisting NavBar into `App.vue` behind a `meta.standalone` opt-out) is a bigger refactor — worth considering, but out of scope here unless it's the cheaper fix.

## Acceptance Criteria

- [ ] `/enterprise` renders the main NavBar at the top, matching Dashboard/Settings/Operations
- [ ] `/enterprise/audit` renders the main NavBar at the top
- [ ] The active-route highlight in NavBar behaves correctly on both routes
- [ ] Page content spacing/layout is not broken by the added chrome (no double top padding, no overlap)
- [ ] Standalone surfaces (`/portal`, `/m`) still render without NavBar — no regression
- [ ] The intent for `/enterprise/client-portal` is confirmed and documented in the route comment: either it gains NavBar, or the comment states explicitly that it is a standalone client-facing surface
- [ ] Dark mode renders correctly on both fixed pages

## Technical Notes

- `src/frontend/src/App.vue` — global shell, currently `<router-view>` + `HelpChatWidget` only
- `src/frontend/src/components/NavBar.vue` — the shared top nav
- `src/frontend/src/views/enterprise/{Index,Audit,ClientPortal}.vue` — the affected views
- `src/frontend/src/router/index.js` — enterprise route defs, entitlement guard, and the existing `hideHelpWidget` meta precedent for standalone surfaces
- Follow the pattern in `Dashboard.vue`/`Operations.vue` for import + placement
- If the `App.vue` hoist is chosen instead, a `meta.standalone: true` flag on `/portal`, `/portal/c/:sessionId`, `/m`, `/login` (and possibly `/enterprise/client-portal`) mirrors the `hideHelpWidget` precedent already in the router

dolho:
## Summary

`/enterprise` and `/enterprise/audit` rendered no top nav — the user landed with no way back to the platform except the browser's back button.

Root cause is as the issue describes: Trinity has no global chrome. `App.vue` renders only `<router-view>`, so every authenticated view mounts `NavBar` itself. The enterprise views never did, so the chrome silently vanished on those routes.

> **Review with `git diff -w`.** The real change is **21 added lines, 0 deletions**. Wrapping each template re-indents its body, which inflates `Audit.vue` to ~1266 lines of pure whitespace churn.

## Changes

- `views/enterprise/Index.vue`, `views/enterprise/Audit.vue` — wrap the existing centred container in the house layout (`min-h-screen` shell → `NavBar` → content), matching `Settings.vue` / `Templates.vue`. The original `p-6 max-w-*` containers are untouched, so spacing is unchanged and there's no double padding.
- `router/index.js` — document the `/enterprise/client-portal` intent (see below).

**Active highlight needed no change.** NavBar's Enterprise link already keys off `$route.path.startsWith('/enterprise')`, so it highlights correctly on both routes once NavBar renders.

**`/enterprise/client-portal` deliberately keeps NO NavBar**, and the route comment now says so rather than leaving it ambiguous: it's a client-facing surface, not an operator page — platform nav would offer a client links they can't use. It sits alongside `/portal` and `/m`. (Ambiguity was the actual bug here: the missing comment is why a reader couldn't tell "standalone by design" from "same defect as its siblings".)

## Scope

The issue floats hoisting NavBar into `App.vue` behind a `meta.standalone` opt-out — *"unless it's the cheaper fix"*. It isn't: **7 views** render NavBar today, so hoisting means touching ~12 files, adding `meta.standalone` to every standalone route, and risking double-nav on every authenticated page. Left as the follow-up the issue describes.

## Verification

Production `vite build` succeeds, and the built output confirms the wiring rather than just the intent:

- `Index-*.js` (contains `enterprise-landing`) → references `NavBar` ✅
- `Audit-*.js` (contains `audit-dashboard`) → references `NavBar` ✅
- `Portal-*.js`, `ClientPortal-*.js`, `MobileAdmin-*.js` → **no** `NavBar` ✅ (standalone surfaces unregressed, AC #5)

Both wrappers use the same `min-h-screen bg-gray-100 dark:bg-gray-900` shell as the rest of the platform, so dark mode inherits the existing behaviour (AC #7).

**Not visually confirmed in a browser** — worth stating plainly. This repo's `docker-compose.override.yml` deliberately repoints the frontend dev server at the `trinity-portal-cleanup` worktree ("so Vite serves the old-client-portal cleanup at localhost:8001"), so localhost:8001 doesn't serve this branch, and I didn't want to disturb that setup. Verification here is build + bundle-level. A reviewer with the override reverted (`docker compose up -d frontend`) can eyeball `/enterprise` and `/enterprise/audit` in a few seconds.

## Interaction with #1637

PR #1637 removes `/enterprise/client-portal` entirely. If it lands first, the route comment added here disappears with the route — no conflict in substance, and the `Index.vue`/`Audit.vue` fixes are independent of it.

## Test Plan

- [x] `vite build` succeeds (production config)
- [x] Enterprise chunks reference NavBar; Portal/ClientPortal/MobileAdmin chunks do not
- [x] Existing `p-6 max-w-*` containers unchanged — no double padding
- [ ] Reviewer: eyeball `/enterprise` + `/enterprise/audit` light/dark with the frontend mount pointed at this branch

Closes #1636

### DIFF
--- a/src/frontend/src/router/index.js
+++ b/src/frontend/src/router/index.js
@@ -177,6 +177,10 @@ const routes = [
     path: '/enterprise/client-portal',
     name: 'EnterpriseClientPortal',
     component: () => import('../views/enterprise/ClientPortal.vue'),
+    // STANDALONE, no NavBar — deliberate (#1636). This is a client-facing
+    // surface, not an operator page: the platform nav would offer a client
+    // links they can't use. The sibling /enterprise pages DO render NavBar;
+    // this one is the exception, like /portal and /m.
     // hideHelpWidget: the platform "Trinity Help" widget (bottom-right) overlaps
     // the client chat composer's Send button and isn't relevant to a portal
     // client — keep it off this standalone surface.
--- a/src/frontend/src/views/enterprise/Audit.vue
+++ b/src/frontend/src/views/enterprise/Audit.vue
@@ -16,6 +16,7 @@
  */
 import { computed, onMounted, ref, watch } from 'vue'
 import { useAuditLogStore } from '../../stores/auditLog'
+import NavBar from '@/components/NavBar.vue'
 
 const store = useAuditLogStore()
 // Direct reactive access — Pinia state is already reactive, and the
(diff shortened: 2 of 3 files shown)


dolho:
image

vybe (approved):
Validated via /validate-pr: mechanical NavBar mount on the two operator-facing enterprise pages (Audit, Index) — the diff is a wrap+re-indent, verified no template content changed; the client-portal route's standalone status is documented as deliberate. Security battery clean. Approving.

Issue #1649, opened 2026-07-16
bug: proactive group messages (send_group_message) never persisted to channel session history


dolho:
## Summary

Split from #1600, which fixed proactive **DM** persistence (`send_message`). The group/channel equivalent — MCP `send_group_message` (#349/#350) — has the same gap: a proactive group message is delivered and never written to session history, so the agent has no record of its own broadcast.

Verified: neither backend endpoint persists anything.

- `routers/telegram.py::send_telegram_group_message` (`POST /api/agents/{name}/telegram/groups/{chat_id}/messages`) — no `add_public_chat_message` call
- `routers/slack.py::send_agent_slack_channel_message` (`POST /api/agents/{name}/slack/channels/{channel_id}/messages`) — same

## Why this wasn't folded into #1600

The DM fix works because a DM's session key is fully derivable from the recipient: each adapter's `get_session_identifier()` is driven with a synthetic message and yields exactly the key an inbound reply resolves to. **Groups don't have that property**, and the answer differs per channel:

- **Telegram**: `get_session_identifier` returns `f"{bot_id}:{sender_id}:{chat_id}"` — with **no group branch**, so a group session is scoped per *(sender, chat)*. A proactive broadcast has no sender, so there is no single existing session it belongs to. Options: a synthetic agent-sender key, a per-chat shared session, or writing to every participant's session. All change group-context semantics.
- **Slack**: channel sessions are **thread-scoped** (`team:channel:thread`, #903). A proactive channel post either starts a new thread (its own `ts` becomes the key, knowable only after the send returns) or posts top-level. Reusing the "last" thread would be a guess.

That's a design decision about group-context semantics, not a mechanical port — hence a separate issue.

## Acceptance Criteria

- [ ] Decide the group-session model per channel (Telegram per-chat vs per-sender; Slack new-thread vs top-level) and document the decision
- [ ] A successful proactive Telegram group send appends an assistant turn to whatever session the following inbound group message resolves to — **assert identifier equality**, since persisting into a *different* session looks like a fix but leaves the bug intact (this was the trap in #1600)
- [ ] Same for Slack channel sends
- [ ] Persist only on confirmed delivery — a failed send writes no phantom turn
- [ ] Attribution follows #903: `sender_label` = agent name; `sender_email` = **None** for a shared/multi-participant thread (per `_assistant_sender_email`), so a broadcast is never folded into one participant's MEM-001 memory. This differs from the DM case in #1600, where the recipient's email IS stamped.

[...]

Issue #951, opened 2026-05-26
Notify requester on approval of channel access request


vybe:
## Summary

When a pending access request is approved via `POST /api/agents/{name}/access-requests/{id}/decide`, the approving owner sees the action complete, but the **requesting user is never notified on the channel they used**. They have no signal that they can now message the agent — they have to guess and retry. Confirmed broken on Telegram; status on Slack and WhatsApp unknown but likely the same gap.

## Context

The `#311` Unified Channel Access Control flow upserts a pending `access_requests` row when a verified-email user without sharing tries to chat an agent on Telegram / Slack / WhatsApp, and the user is told "🔒 Your access request is pending approval." (`src/backend/adapters/message_router.py:426`).

On approval the endpoint (`src/backend/routers/sharing.py:229-258`):
1. Marks the request approved
2. Inserts into `agent_sharing`
3. Optionally whitelists the email
4. Broadcasts an `agent_shared` WebSocket event for the dashboard

It does **not** send a proactive message back to the requester on the originating channel. Result: the user who asked for access is left waiting indefinitely with no callback, even though they're now allowed.

There is already a proactive messaging primitive (`services/proactive_message_service.py`, `routers/messages.py`, #321) that can target a verified email — so the building block exists. The decision endpoint just isn't using it.

## Acceptance Criteria

- [ ] On approval, the requester receives a message on the channel they originated the request from (Telegram / Slack / WhatsApp) telling them they can now chat the agent
- [ ] Rejection does **not** notify the requester (deliberate — owners may not want to confirm the agent exists)
- [ ] If the user has chat links on multiple channels, prefer the channel recorded on the `access_requests.channel` column (the one they actually asked from)
- [ ] If the channel binding/transport is unavailable at decision time, the approval still succeeds (notification is best-effort, not blocking)
- [ ] Audit log records whether the post-approval notification was delivered, skipped, or failed
- [ ] Manual test path documented for each of Telegram, Slack, WhatsApp

## Technical Notes

- Decision endpoint: `src/backend/routers/sharing.py:208-274`
- Channel column on `access_requests` already records `'telegram' | 'slack' | 'whatsapp' | 'web'` (see `architecture.md` schema)
- Proactive send building block: `services/proactive_message_service.py` resolves verified-email → channel binding and dispatches
- Channel adapters expose `send_message` / equivalent via `adapters/base.py:ChannelAdapter`
[...]

Issue #321, opened 2026-04-13
feat: proactive agent messaging to channel users (Telegram / Slack / web)


vybe:
## Problem

Trinity agents can only **respond** to channel messages. They cannot initiate a conversation — there is no way for a scheduled task, a process step, or an agent-to-agent signal to push a message to a specific user on Telegram / Slack / web.

Concretely:

- `telegram_adapter.send_response()` only runs inside the inbound pipeline
- `send_notification` MCP tool broadcasts to the Trinity WebSocket, not to external channels
- The `announce` skill can post to a preconfigured channel via env var, but it is a CLI side-channel, not an agent runtime capability

This blocks real use cases: reminders, monitoring alerts, scheduled reports, follow-ups after long-running work.

## Proposed approach

One new MCP tool — `send_message` — that resolves a recipient by verified email (the identity key from #311) and dispatches through the correct channel adapter.

```
send_message(
  recipient_email: str,           # verified email; MUST be in agent_sharing OR owner
  text: str,
  channel: \"auto\" | \"telegram\" | \"slack\" | \"web\" = \"auto\",
  reply_to_thread: bool = False,  # continue last thread if one exists
)
```

**Resolution:**
- `telegram` → look up `telegram_chat_links` where `verified_email = recipient_email` AND `binding_id` belongs to the calling agent
- `slack` → look up existing DM channel in `slack_channel_agents` / `slack_active_threads`
- `web` → WebSocket push + persisted message on `public_chat_sessions` (user sees it on next load)
- `auto` → prefer the channel the user most recently used

## Reuse map

| Existing primitive | Role |
|---|---|
| `agent_sharing` (#311) | Authorization — recipient must be in the agent's allow-list |
| `telegram_chat_links.verified_email` (#311) | Telegram recipient resolution |
| `telegram_adapter._send_message()` | Outbound transport — already battle-tested |
| `slack_service.send_message()` | Slack transport — already used for auth prompts |
| MCP auth context (`agent_name`, `scope='agent'`) | Caller identity — agents can only message their own users |

## What's actually new

- MCP tool `send_message` in `src/mcp-server/src/tools/` (new module or extend `notifications.ts`)
- Backend endpoint `POST /api/agents/{name}/messages` — does the authorization + channel dispatch
- Rate limiting (per agent, per recipient) to prevent spam — reuse the adapter-layer limiter
- Optional: `last_active_channel` column on `agent_sharing` or a lightweight `user_channel_activity` table for `auto` routing

[...]

PR #1650, opened 2026-07-16
fix(channels): persist proactive messages to channel session history (#1600)

### LINKED ISSUE #1600: bug: proactive messages (send_message) never persisted to channel session history — agent unaware of its own outbound messages
## Summary

When an agent sends a proactive message to a user via the MCP `send_message` tool (#321) — e.g. over Telegram — the delivered message is never persisted to the channel session history (`public_chat_messages`). The next inbound message from that user builds its conversation context from `db.build_public_chat_context(session_id, ...)`, which contains only turns persisted by the channel message router — so the agent has no record of the proactive message it just sent. In the ensuing exchange the agent is unaware of its own outreach: it can repeat itself, contradict the earlier message, or fail to understand the user's reply ("yes, sounds good" — to what?).

Reported against Telegram, but the same gap applies to all proactive delivery channels: none of the `_deliver_*` helpers in `src/backend/services/proactive_message_service.py` write history.

## Context

- **Inbound flow** (`src/backend/adapters/message_router.py`): step 11 persists BOTH the user turn and the assistant reply via `db.add_public_chat_message(...)` with per-speaker attribution (#903), and the context prompt for each turn is built from that history via `db.build_public_chat_context(session_id, ...)`.
- **Proactive flow** (`src/backend/services/proactive_message_service.py`, #321): `_deliver_telegram` / `_deliver_slack` / `_deliver_whatsapp` resolve the recipient and call the channel API directly — no `add_public_chat_message` call exists anywhere in the service (`_deliver_web` is unimplemented v1).
- The persistence seam already exists for Telegram and WhatsApp: both delivery helpers look up the chat link, and `telegram_chat_links.session_id` / `whatsapp_chat_links.session_id` identify the recipient's channel session. The Slack path resolves the recipient by email across workspaces and has no chat-link/session mapping today.

## Acceptance Criteria

- [ ] A successful proactive Telegram delivery appends an assistant-role message to the recipient's channel session history (`sender_label` = agent name, mirroring router step 11 attribution semantics, #903)
- [ ] The next inbound turn's `build_public_chat_context` includes the proactive message, so the agent responds aware of what it sent (regression test: proactive send → inbound reply → context contains the proactive message)
- [ ] Same behavior for WhatsApp (chat link already carries `session_id`)
- [ ] Slack: either persisted the same way (resolving/creating a session for the DM) or explicitly documented as a known limitation in the `send_message` tool description + user docs — no silent inconsistency
- [ ] Persist only on confirmed successful delivery — a failed send must not write a phantom assistant turn into history
- [ ] Recipient with no session yet (NULL `session_id` on the chat link) degrades gracefully: message still delivered, no crash; decide and document whether a session is created
- [ ] MEM-001 interplay: the persisted assistant turn's `sender_email` follows the `_assistant_sender_email` semantics (#903) so sender-filtered summarization doesn't fold proactive messages into the wrong user's memory

## Technical Notes

- Likely fix point: after each successful `_deliver_*` in `proactive_message_service.py`, call `db.add_public_chat_message(session_id, "assistant", text, sender_email=..., sender_label=agent_name)` using the chat link's `session_id`.
- Check whether `send_group_message` (MCP `channels.ts`, #349/#350) has the same gap for proactive group/channel sends — if so, fix in the same pass or file a follow-up.
- Out of scope: approval notifications (#951), TTS voice replies (epic #24).

dolho:
## Summary

A proactive `send_message` (#321) was delivered and written **nowhere**. The next inbound turn builds context from `build_public_chat_context`, which only ever contained turns the message router persisted — so the agent had no record of its own outreach. It would repeat itself, contradict the message, or fail to parse the reply (*"yes, sounds good"* — to what?).

Now fixed for **all three DM channels** (Telegram, WhatsApp, **and Slack**).

## ⚠️ The issue's suggested fix cannot work — worth reading before reviewing

The issue proposes keying off the chat link's `session_id`, and states the seam *"already exists for Telegram and WhatsApp"*. **It doesn't.**

`telegram_chat_links.session_id` and `whatsapp_chat_links.session_id` are **never written by any code path**. Across the whole repo they appear in exactly three places, all `SELECT`s:

```
db/telegram_channels.py:273   telegram_chat_links.c.session_id,     ← select
db/telegram_channels.py:411   telegram_chat_links.c.session_id,     ← select
db/whatsapp_channels.py:370   whatsapp_chat_links.c.session_id,     ← select
```

`record_inbound` (the only writer of those tables) inserts `telegram_user_id` + `message_count` and never a session. They're vestigial columns.

**Implemented as suggested, the fix would have persisted nothing — silently — and looked done.** AC-6 ("recipient with no session yet") isn't an edge case under that design; it's every recipient.

## Approach

Resolve the session by **identifier** via `get_or_create_public_chat_session`, and make the identifier provably equal to the inbound one.

That equality is the whole fix. Persisting into *a* session is worthless — it has to be the session the recipient's next message resolves to, or the write succeeds and the bug survives invisibly. So each `_deliver_*` sets `DeliveryResult.session_identifier` by driving **its own adapter's** `get_session_identifier()` with a synthetic DM, never by re-deriving the format in the service:

| Channel | Key | Derived from |
|---|---|---|
| Telegram | `{bot_id}:{user}:{chat}` | binding `bot_id` + chat link's `telegram_user_id` (a DM's chat_id IS the user id — the existing send already relies on this) |
| WhatsApp | `{binding_id}:{phone}` | binding id + chat link's `wa_user_phone` |
| Slack | `{team}:{user}:{dm_channel}` | workspace `team_id` + looked-up user + opened DM channel |

Why not just format the string in the service? The formats have subtle fallbacks that would drift — Telegram's `bot_id` defaults to `""` in the webhook but `"unknown"` in the adapter. A mismatch mints a *second* session and reintroduces the bug in a place nothing checks.

**Slack is fixed, not documented as a limitation** (AC-4 allowed either). It has no chat-link table, but `_deliver_slack` already resolves team, user, and DM channel — everything its DM key needs.

## AC coverage

- **AC-1/2/3** — Telegram/WhatsApp persist; the next turn's context includes the message (identifier equality is what guarantees it)
- **AC-4** — Slack persisted, not deferred
- **AC-5** — persist only on confirmed delivery; a failed send writes no phantom turn
- **AC-6** — moot under this design (the column was never the key); the session is created when absent, and that's documented
- **AC-7** — `sender_email` = recipient. A proactive send targets ONE verified email — a DM, i.e. single-participant — so `_assistant_sender_email`'s rule (#903) resolves to the recipient, keeping sender-filtered MEM-001 in the right user's memory

**Scope:** persistence hangs off `_send_message_inner`, not the shared `_deliver_*` helpers, so `send_access_grant_notification` (out of scope, #951) is excluded **structurally** rather than by a flag. Fail-soft throughout: the message is already delivered when persistence runs, so a DB error is logged loudly and never surfaces as a delivery failure.

## `send_group_message` → #1649

Confirmed the same gap (0 persistence calls in both `send_telegram_group_message` and `send_agent_slack_channel_message`), but it's a different problem, not a port: Telegram group sessions are per-*(sender, chat)* and a broadcast has no sender; Slack channels are thread-scoped on a `ts` only knowable after the send. That's a semantics decision — filed as **#1649** with the analysis.

## Test Plan

- [x] `pytest tests/unit/test_1600_proactive_history.py` — 12 passed
- [x] `pytest tests/test_proactive_audit_unit.py` — existing suite green (16 total)
- [x] Reverting the `_persist_outbound` call fails `test_successful_delivery_persists` (the end-to-end regression). **Honest note:** only that 1 of 12 flips — the parity tests drive the adapters, which this PR doesn't change; they exist to catch *future* drift, not to prove today's fix.
- [ ] Not exercised against live Telegram/Slack/WhatsApp — needs real channel credentials

Related to #1600

Fixes #1600

### DIFF
--- a/docs/memory/feature-flows/proactive-messaging.md
+++ b/docs/memory/feature-flows/proactive-messaging.md
@@ -15,6 +15,7 @@ Enables agents to send proactive messages to specific users by verified email ac
 - **Mandatory audit logging**: All proactive sends logged via platform_audit_service
 - **Multi-channel delivery**: Auto-selection tries telegram → slack → web. WhatsApp is an explicit-only channel (`channel="whatsapp"`) and not part of `auto` — Twilio's 24-hour session window makes it unreliable for inactive recipients.
 - **MCP tool access**: Agents use `send_message` MCP tool for outreach
+- **History-aware (#1600)**: a delivered message is appended to the recipient's channel session, so the agent's next turn knows what it sent
 
 ## Architecture
 
@@ -160,6 +161,56 @@ actor_id=agent_name
 target_type="user"
 target_id=recipient_email
 ```
+
+### 5. Session-History Persistence (#1600)
+
+A delivered message is appended to the recipient's channel session, so the next
+inbound turn's `build_public_chat_context` includes it. Without this the agent had
+no record of its own outreach — it would repeat itself, contradict the message, or
+fail to parse the reply ("yes, sounds good" — to what?).
+
+```python
+# _send_message_inner, on confirmed delivery only (never on failure — no phantom turn)
+self._persist_outbound(agent_name, recipient_email, text, result)
+  -> db.get_or_create_public_chat_session(agent_name, result.session_identifier, channel)
+  -> db.add_public_chat_message(session_id, "assistant", text,
+                                sender_email=recipient_email,   # DM => single participant (#903)
+                                sender_label=agent_name)
+```
+
+**The session key must equal the inbound one.** Persisting into *a* session is
+useless; it has to be the one the recipient's next message resolves to, or the
+write succeeds and the bug survives invisibly. So each `_deliver_*` sets
+`DeliveryResult.session_identifier` by driving **its own adapter's**
+`get_session_identifier()` with a synthetic DM — never by re-deriving the format
+in the service, which would drift from the adapter:
+
+| Channel | Key | Derived from |
+|---------|-----|--------------|
+| Telegram | `{bot_id}:{user}:{chat}` | binding `bot_id` + chat link's `telegram_user_id` (a DM's chat_id IS the user id) |
+| WhatsApp | `{binding_id}:{phone}` | binding id + chat link's `wa_user_phone` |
+| Slack | `{team}:{user}:{dm_channel}` | workspace `team_id` + looked-up user + opened DM channel (`is_dm=True` selects the DM branch, not the thread branch) |
+
+**The session is created if absent.** `telegram_chat_links.session_id` and
+`whatsapp_chat_links.session_id` look like the natural key but are **never written
+by any code path** — vestigial columns, read in three SELECTs and nowhere else. So
+"reuse the link's session" isn't available; `get_or_create_public_chat_session` with
+the adapter's identifier lands in the same session regardless.
+
+**Attribution (#903)**: role `assistant`, `sender_label` = agent name,
+`sender_email` = the recipient. A proactive send targets ONE verified email — a DM,
+i.e. a single-participant session — so `_assistant_sender_email`'s rule resolves to
+the recipient, keeping sender-filtered MEM-001 summarization in the right user's
+memory.
+
+**Fail-soft**: the message is already delivered when this runs, so a persistence
+error is logged (loudly) and never surfaces as a delivery failure.
+
+**Scope**: `send_access_grant_notification` shares the `_deliver_*` helpers but does
+NOT persist — its history semantics are out of scope (#951). The call sits on the
+`send_message` path, which draws that line structurally rather than with a flag.
+Proactive **group** sends (`send_group_message`, #349/#350) have the same gap and a
+different session model — tracked in #1649.
 
 ## File Locations
 
--- a/src/backend/services/proactive_message_service.py
+++ b/src/backend/services/proactive_message_service.py
@@ -62,6 +62,13 @@ class DeliveryResult:
     channel: str
     message_id: Optional[str] = None
     error: Optional[str] = None
+    # #1600: the recipient's channel-session key, set by the `_deliver_*` helper
+    # on success so the caller can persist the outbound turn into the SAME
+    # session an inbound reply will use. Each helper derives it by calling its
+    # own adapter's `get_session_identifier()` — never by re-deriving the format
+    # here, which would drift from the adapter and silently mint a second
+    # session. None ⇒ the channel can't resolve a session (nothing to persist).
+    session_identifier: Optional[str] = None
 
 
 class ProactiveMessageService:
@@ -275,6 +282,14 @@ class ProactiveMessageService:
                 )
                 if result.success:
                     self._increment_rate_limit(agent_name, recipient_email)
+                    # #1600: record the outbound turn in the recipient's channel
+                    # session so the agent's next turn knows what it sent.
+                    # Deliberately here and not in `_deliver_*`: those helpers are
+                    # shared with `send_access_grant_notification`, whose
+                    # persistence is out of scope (#951). Keeping the call on the
+                    # send_message path draws that line structurally rather than
+                    # with a flag.
+                    self._persist_outbound(agent_name, recipient_email, text, result)
                     await self._audit_send(
                         agent_name, recipient_email, ch, True,
                         message_preview=message_preview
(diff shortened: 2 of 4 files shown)


dolho:
## Test-coverage correction (`c51f08e9`)

Asked directly whether the tests really cover this, I checked instead of asserting — and they didn't.

I sabotaged the service two ways, each of which silently derives a **different** session key and therefore leaves the bug fully intact:

- Telegram metadata key typo: `bot_id` → `botid_TYPO` (yields `unknown:123:123`)
- Slack probe dropping `is_dm` (takes the thread branch, not the DM branch)

**All 12 tests passed against both.** The suite would have shipped green over a non-fix.

The hole: `TestSessionIdentifierParity` compares two probes constructed *in the test file*. It proves the adapters are self-consistent — never in doubt — and never touches the probe the **service** builds. It was testing the wrong side of the seam.

`c51f08e9` adds `TestDeliveryPathProducesTheInboundKey`, which drives the real `_deliver_telegram` / `_deliver_slack` with mocked channel APIs and asserts the returned `session_identifier` equals the key an inbound DM resolves to. Both sabotages now fail their test and nothing else. Also asserts a failed send carries no session key (nothing to persist against). 15 tests.

### Still not covered — stating it rather than implying it

**AC-2's end-to-end round-trip is not tested**: *"proactive send → inbound reply → context contains the proactive message"*. Every `db` call in this suite is mocked, so nothing proves `add_public_chat_message` and `build_public_chat_context` actually round-trip through a real store. What's verified is that the right call is made with the right arguments against the right session key — not that the agent genuinely sees the message on its next turn.

Closing that needs a real DB (the sqlite harness) or a live channel. Worth doing before this is trusted in prod; happy to add it if a reviewer wants it in this PR rather than a follow-up.

dolho:
Validated via `/validate-pr` at `a8b4131b`.

> **Disclosure:** I wrote every commit here. The mechanical checks below are facts; the judgement that the fix is *right* is not independent. The two real findings this pass produced both came from CI and from adversarially attacking my own tests — not from reading them approvingly.

## Findings this pass

**1. CI caught a real regression — my tests poisoned another suite.** The base-vs-head diff flagged two EXISTING tests failing under HEAD:

```
test_1609_proactive_rate_limits::test_limit_is_sourced_from_settings_not_the_constant
test_1609_proactive_rate_limits::test_zero_limit_is_unlimited_and_skips_redis
```

My fixture stubbed `sys.modules` and forced a re-import so the stubs bound at import time. Re-importing creates a **new module object** and rebinds the `proactive_message_service` attribute on the `services` **package**; monkeypatch restores `sys.modules` but not that attribute. #1609 reaches the module via `import services.proactive_message_service as pms` (package attribute → my re-imported object) while its `svc` fixture goes through `sys.modules` (→ the original). It patched one object and exercised another, so its patch silently did nothing and the real limit answered.

Fixed in `a8b4131b` by dropping `sys.modules` manipulation entirely and patching the bound names — which is the pattern #1609 itself documents in a comment ("patch the name as bound in the CONSUMER module ... robust to sys.modules identity quirks"). Strictly less machinery, nothing to leak. Both files now pass in both orders and under random seeds. **Local runs never showed it** — it needs 1600-then-1609 in one session; the seed matrix is what found it.

**2. The tests didn't cover the fix** (`c51f08e9`, detailed in the comment above). Sabotaging the service two ways — telegram `bot_id`→`botid_TYPO`, slack probe dropping `is_dm`, each silently sending the turn to a **different** session and leaving the bug intact — passed all 12 tests. The parity tests compared two probes written in the test file and never touched the service's probe. Now covered by `TestDeliveryPathProducesTheInboundKey`; both sabotages fail it.

## Validation summary

| Category | Status | Notes |
|---|---|---|
| Base branch | ✅ | `dev` |
| PR size | ✅ | 4 files (+566/−1) |
| Issue link | ✅ | `Related to #1600` — resolves, P2/type-bug/theme-channels |
| Commit messages | ✅ | conventional; each explains *why*, incl. the two self-corrections |
| Requirements | ➖ | bug fix (tier = commit message only). Note: #321 proactive messaging has **no** requirements section at all — a pre-existing gap, not introduced here |
| Architecture | ➖ | no endpoint/schema/component change |
| Feature flow | ✅ | `proactive-messaging.md` — new §5 documents the key derivation, the vestigial-column finding, #903 attribution, and the #951/#1649 scope lines |
| Build/config packaging | ➖ | no new top-level module, no new `os.getenv()` |
| Code quality | ✅ | scoped to the stated purpose; matches router step 11's persistence shape |
| Tests | ✅ | 15 + existing suites green; lint OK |
| CI | ⚠️ | re-running after `a8b4131b` — the regression diff must go green before merge |

## Known gaps (stated, not hidden)

- **AC-2 is not tested end-to-end.** *"proactive send → inbound reply → context contains the proactive message"* — every `db` call is mocked, so nothing proves `add_public_chat_message` + `build_public_chat_context` round-trip through a real store. Needs the sqlite harness or a live channel.
- **Not exercised against live Telegram/Slack/WhatsApp.** A live Slack test was set up on a dev instance but is blocked on a workspace bot token.
- `test_no_session_identifier_skips_persistence_without_crashing` covers a **defensive** branch that can't currently fire: `_deliver_web` always raises, so no channel returns success without a key today.

## Recommendation

**APPROVE once CI is green** — pending the regression diff clearing on `a8b4131b`.

The fix itself is sound and the reasoning is documented where the next reader will hit it. But I'd weight an independent review heavily here: across three passes this PR produced a test suite that passed over a sabotaged fix and a fixture that broke a neighbouring suite. Both were found by adversarial checks and CI, not by me reading my own work. The AC-2 round-trip is the gap I'd most want closed before this is trusted in production.

vybe (approved):
Validated via /validate-pr: session-key parity is asserted through the adapters' own get_session_identifier (both probe-parity and service-probe tests — the latter catch a sabotaged probe), persist-only-on-confirmed-delivery, fail-soft persistence, #903 attribution verified (sender_email=recipient for single-participant DM). Confirmed the db facade call shape matches the inbound router's (agent_name as link_id is the router's own convention, message_router.py:455). Security battery clean. Approving.

PR #1654, opened 2026-07-16
fix: Cornelius template + seed vault declare only what the bundle ships (#1646)

### LINKED ISSUE #1646: bug: Cornelius template advertises semantic_search and skills the bundled template doesn't ship (search falls back to substring scan)
## Summary

`config/agent-templates/cornelius/template.yaml` declares capabilities including `semantic_search`, `spreading_activation`, and `usage_based_learning`, and declares `sub_agents:` and `skills:` sections — but the bundled template does not include `resources/local-brain-search/` (nor a `.claude/` skills directory), so the semantic path can never run and several declared skills/sub-agents don't exist in the bundle.

## Context

The `.trinity/brain-orb/search` hook implements a deliberate backend ladder (per its docstring): try the local-brain-search daemon (semantic FAISS search via `run_search.sh`), then fall back to a stdlib keyword scan over `Brain/**/*.md`. With `resources/local-brain-search/run_search.sh` absent from the bundle, `try_semantic()` fails immediately and every fresh install serves substring matching (`{"backend": "keyword"}`) while the template metadata claims semantic search.

Since Cornelius is auto-seeded on fresh installs (`cornelius_agent_service`), this mismatch is part of the first impression of the platform's knowledge story — the template should claim exactly what it ships.

## Acceptance Criteria

- [ ] `template.yaml` `capabilities:` reflects what the bundled template actually does at every bootstrap state — either bundle the semantic stack, or trim the claims to the keyword floor (and re-add on upgrade)
- [ ] Every skill / sub-agent declared in `template.yaml` exists in the bundle, or the declaration is removed
- [ ] The keyword → semantic upgrade ladder in the search hook keeps working unchanged (`{"backend": ...}` stays honest)

## Technical Notes

Trimming the claims is a metadata-level fix and can land immediately; actually bundling a local semantic stack is a larger feature tracked separately. This issue is satisfied by either, as long as claims and behavior match.

obasilakis:
## Summary

- The bundled Cornelius is a **vendored snapshot** of the public `Abilityai/cornelius` repo. The snapshot took `template.yaml` + `CLAUDE.md` + `.gitignore` but **not** `.claude/` or `resources/local-brain-search/` — so it advertised ~24 skills, 9 sub-agents and a FAISS semantic tier the bundle cannot serve.
- Cornelius is **auto-seeded on every fresh install** (ent#107), so this was the platform's first impression of the knowledge story.
- The bundle materializes via a verbatim `cp -r` of `/template` with **no git origin**, so the missing files never arrive later. The claims were permanently false, not merely stale.

Nothing caught it: `template.yaml` has **no CI validation**, and compatibility checks X-003/X-004 only run against a *live* agent. Systemic guard filed as **#1655**.

## Why trim rather than bundle

The issue allows either. ent#173 already owns "make it real" and says so explicitly:

> The short-term truthfulness fix (public tracker) can trim the claims; **this issue ships the real floor.**

Its epic lists this issue as launch hygiene **independent of the ADR**. Both are `status-incubating` (H8 readout targets 2026-09-01), so waiting ships a lying template until September.

Shipping the skills was evaluated and rejected on evidence: upstream's skills are entangled with the author's personal environment (LBS, Smart Connections, a chart MCP, Apollo) — none configured, and **upstream's own `.mcp.json.template` doesn't configure them either**. The clean subset was 35/49 skills but only **1/10 sub-agents**.

## Changes

| File | What |
|---|---|
| `template.yaml` | `capabilities`: drop the 3 semantic tokens (**`brain-orb` kept — load-bearing**). `sub_agents`/`skills`: removed (0 shipped). `use_cases`: drop the semantic lookups + wording naming deleted machinery. `mcp_servers`: drop `playwright` + `trinity`. `metrics`: drop `index_status`/`memory_learning`/`advice_requests` |
| `CLAUDE.md` | Removed instructions to invoke `/advise`, `/recall`, `/refresh-index` and to shell out to `resources/local-brain-search/run_*.sh`; removed 2 dangling `@`-imports; folder structure now matches the bundle (775 → ~350 lines) |
| `Brain/` (6 files) | Seed vault no longer advertises unshipped commands; removed the fabricated "Vault Statistics" block (~1,067 notes / 13 MOCs — **16 `.md` and 2 MOCs ship**) |
| `tests/unit/test_1646_cornelius_template_claims.py` | New — 11 tests, the missing guard |

**`CLAUDE.md` was the bigger defect.** `template.yaml` misleads a human reading the Info tab; `CLAUDE.md` misled the *agent*, telling it to execute scripts that don't exist.

Two fields the AC didn't name but which mattered more than `capabilities`:
- **`use_cases`** — renders above the fold and is **clickable**, so a false entry prefills a task the agent cannot serve.
- **`mcp_servers`** — `playwright`/`trinity` were breaking **X-004** (which compares against `.mcp.json.template`); it now passes.

## Test Plan

- [x] `pytest tests/unit/test_1646_cornelius_template_claims.py -v` → **11 passed**
- [x] **Verified the tests catch the bug** — stashed the fix, re-ran: **8 of 10 fail**. The 2 staying green are the state-independent guards (`brain-orb` preserved, ladder intact). Reverting only `Brain/` makes the new vault test name all 6 files and 9 commands.
- [x] No regressions: `test_ent107_cornelius_seed`, `test_compatibility_checks`, `test_local_templates_listing`, `test_data_paths_template_surface` → **82 passed** total
- [x] `template.yaml` parses; functional keys `crud.py` reads (`type`, `resources`, `credentials`, `shared_folders`) intact

Tests assert **every declaration resolves to a file** rather than pinning today's list, so a future re-vendor with the same drift fails here.

## Review

An independent review caught that the first pass stopped at `template.yaml` + `CLAUDE.md` while the seed vault still made every removed claim — on a *more* visible surface (`Brain/README.md`: "Start here ... try `/advise`"). Fixed, and the suite extended to scan `Brain/**/*.md`. It also closed a false-negative hole in one test and **corrected a finding of mine**: `git.ignore_paths` entries for `.claude/commands/trinity/` and `local-brain-search/data/` are *not* dead (the former is created at runtime by the trinity plugin, the latter returns with ent#173) — removing them would have been the regression.

`/cso --diff`: **0 findings** at the 8/10 gate. All safety sections of the system prompt survive the splice (ETHICAL BOUNDARIES, Preserve User Agency, Respect Cognitive Privacy, …); no `deny`/`guard` text removed; no new endpoint/exec/upload surface; no secrets.

## Scoped out (deliberate)

- **Dead seed vault** → **#1656**. The 2 shipped MOCs link to 11 MOCs that don't exist (22 of 23 wikilinks dead), and the orb's 514-node graph doesn't match the 16 on-disk notes. This PR removes the false *prose* but does not invent a note corpus — that's a product decision, and it interacts with the seed-export work.
- **`metrics`** remaining entries are inert (`MetricsPanel.vue` has no imports — no renderer).
- `requirements/core-agent.md` describes this template as `capabilities: [brain-orb]`, already inaccurate before this change.
- **Existing agents unaffected** — the `.trinity-initialized` guard means the copy only runs on first boot.

Fixes #1646


vybe (approved):
Validated via /validate-pr: security greps clean, 11 regression tests + 71 adjacent template-consumer tests pass locally, scope-disciplined bug fix with named regression coverage. Approving.

Issue #1080, opened 2026-06-05
Update and actualize the available LLM model list across the app (add Opus 4.8, retire EOL models)


vybe:
## Summary

The app's selectable LLM model lists have drifted from Anthropic's current Claude lineup. The canonical user-facing list (`src/frontend/src/components/ModelSelector.vue`, `PRESET_MODELS`) was last synced **2026-05-13** and tops out at **Claude Opus 4.7** — but **Opus 4.8 (`claude-opus-4-8`) is now the flagship** and isn't selectable anywhere in the product. Separately, two models still offered in the list — `claude-opus-4-20250514` and `claude-sonnet-4-20250514` — are scheduled to **retire on 2026-06-15**, after which any agent / schedule / session pinned to them returns a 404. Refresh every surface that defines or defaults a model so the offered set matches the live lineup and no user can pin a soon-to-be-retired ID.

## Context

Model IDs are hard-coded across **~10 locations spanning 5 surfaces** with no single source of truth, so the list rots silently as Anthropic ships new models and retires old ones. This is also time-sensitive: the 2026-06-15 retirement is ~10 days out, and `claude-3-haiku-20240307` already passed its 2026-04-19 retirement.

**Current Anthropic lineup** (authoritative reference — [Models Overview](https://platform.claude.com/docs/en/about-claude/models/overview)):

| Tier | Models |
|------|--------|
| Current | `claude-opus-4-8` (flagship, 1M ctx), `claude-opus-4-7`, `claude-opus-4-6`, `claude-sonnet-4-6`, `claude-haiku-4-5` |
| Legacy (still active) | `claude-opus-4-5`, `claude-opus-4-1`, `claude-sonnet-4-5` |
| Deprecated — **retires 2026-06-15** | `claude-opus-4-20250514`, `claude-sonnet-4-20250514` |
| Retired (404 today) | `claude-3-7-sonnet-*`, `claude-3-5-haiku-*`, `claude-3-opus-*`, … |

**Surfaces that define or default a model (all need review):**

1. Frontend canonical picker — `src/frontend/src/components/ModelSelector.vue` (`PRESET_MODELS`)
2. Frontend admin platform-default dropdown — `src/frontend/src/views/Settings.vue`
3. Frontend task default — `src/frontend/src/components/TasksPanel.vue` (localStorage fallback)
4. Backend platform default — `src/backend/services/settings_service.py` (`PLATFORM_DEFAULT_MODEL_VALUE`)
[...]

PR #1662, opened 2026-07-16
fix: add Claude Fable 5 / Sonnet 5 to the platform default and public-channel model lists (#1660)

### LINKED ISSUE #1660: bug: Fable 5 / Sonnet 5 missing from platform default-model dropdown and PUBLIC_CHANNEL_MODELS whitelist (drifted from ModelSelector presets)
## Summary

The per-agent model picker (`src/frontend/src/components/ModelSelector.vue`) gained the Claude 5 family presets — `claude-fable-5` and `claude-sonnet-5` — but the two hand-synced model lists never followed:

- **Platform default-model dropdown** (`src/frontend/src/views/Settings.vue`, Settings → General) still offers only Sonnet 4.6 / Opus 4.8 / 4.7 / 4.6 — an admin cannot pick Fable 5 or Sonnet 5 as the platform default from the UI.
- **`PUBLIC_CHANNEL_MODELS` whitelist** (`src/backend/services/settings_service.py`) still holds the pre-Claude-5 set, so the per-agent public-channel override (#894, `PUT /api/agents/{name}/public-channel-model`) rejects the new models with a 422.

The code comment on `PUBLIC_CHANNEL_MODELS` says it "mirrors the current presets in ModelSelector.vue (kept in sync by hand)" — that sync has drifted. Present on both `main` and `dev`.

## Context

Found while checking whether Fable 5 is selectable as a default model. Owners can already pick it per agent/schedule via ModelSelector, and `services/model_context.py` already resolves its context window — only the default-model and public-channel surfaces lag. Note the platform default is saved via the generic `PUT /api/settings/{key}` path with **no server-side model whitelist**, so the default-model restriction is frontend-only while the public-channel override is server-validated — the surfaces disagree about which models exist.

## Acceptance Criteria

- [ ] `PUBLIC_CHANNEL_MODELS` in `src/backend/services/settings_service.py` includes `claude-fable-5` and `claude-sonnet-5`, restoring the documented mirror of the ModelSelector presets
- [ ] The platform default-model dropdown in `src/frontend/src/views/Settings.vue` offers Claude Fable 5 and Claude Sonnet 5 (recommended default stays Sonnet 4.6 — this widens the selectable list, no change to `PLATFORM_DEFAULT_MODEL_VALUE`)
- [ ] Selecting Fable 5 as a per-agent public-channel override passes validation (no 422) and resolves at chat time
- [ ] Graceful-degradation posture preserved: a saved model later removed from the whitelist still falls back to the platform default (`db.get_public_channel_model` #1080 behavior unchanged)
- [ ] Drift guard: a unit test (or at minimum a cross-referencing comment on all three lists) ties `PUBLIC_CHANNEL_MODELS`, the Settings.vue dropdown, and the ModelSelector presets together so the next preset addition can't silently drift

## Technical Notes

- ModelSelector presets already carry the new models (`src/frontend/src/components/ModelSelector.vue`).
- Three spots to touch: `PUBLIC_CHANNEL_MODELS` frozenset in `src/backend/services/settings_service.py`, the platform default-model `<option>` list in `src/frontend/src/views/Settings.vue`, and optionally a parity test for the drift guard.
- No schema change, no migration; behavior-safe for existing saved values.

obasilakis:
## Summary

`ModelSelector.vue` gained the Claude 5 presets (`claude-fable-5`, `claude-sonnet-5`); the two hand-synced lists never followed. So an admin couldn't pick either as the platform default, and the per-agent public-channel override (#894) rejected them with a 422.

- `settings_service.PUBLIC_CHANNEL_MODELS` — add both models
- `Settings.vue` platform default-model dropdown — offer both (recommended default unchanged: Sonnet 4.6, `PLATFORM_DEFAULT_MODEL_VALUE` untouched)
- New drift guard pinning the current lineup across all three lists

## Changes

| File | Change |
|------|--------|
| `src/backend/services/settings_service.py` | +2 entries in `PUBLIC_CHANNEL_MODELS` |
| `src/frontend/src/views/Settings.vue` | +2 `<option>`s in the default-model `<select>` |
| `tests/unit/test_1660_model_list_drift.py` | new — 9 tests |

## On the drift guard

The three lists are **subsets by policy, not identity**, so the guard pins the current generation rather than diffing whole lists:

- `ModelSelector` `PRESET_MODELS` (9) — display list; the field is free-text, so it enforces nothing
- `PUBLIC_CHANNEL_MODELS` (7) — current-gen, server-validated. Haiku is in **on purpose** (#894 wanted a cheap model for public traffic)
- `Settings.vue` dropdown (6) — Opus/Sonnet tiers. Haiku is out **on purpose** (#1080 AC: "includes the current Opus/Sonnet options")

It also asserts both subset directions (neither list may offer a model the picker dropped) and that `PLATFORM_DEFAULT_MODEL_VALUE` is itself an option — a fallback with no matching `<option>` renders a blank `<select>`.

## Test Plan

- [x] `pytest tests/unit/test_1660_model_list_drift.py -v` → 9 passed
- [x] Guard is not vacuous: 4 failed / 5 passed on the pre-fix tree
- [x] `pytest tests/unit/test_894_public_channel_model.py tests/unit/test_platform_default_model_regression.py` → 25 passed
- [ ] Manual: Settings → General offers Fable 5 / Sonnet 5; setting Fable 5 as a per-agent public-channel override saves without a 422

## Known gaps (not in scope, worth filing)

- **`platform_default_model` has no server-side validation.** It's written via the generic `PUT /api/settings/{key}`, so the dropdown is the only restriction — an admin can curl any string. The public-channel override *is* server-validated, so the two surfaces disagree. Reconciling it is a breaking change (it'd 422 the existing `[1m]` extended-context opt-in and legacy defaults, and break `tests/test_platform_default_model.py:217`), so it belongs in its own issue.
- **A saved default outside the dropdown renders a blank `<select>`** (Vue drops an unmatched `v-model` value). Pre-existing; reachable today via a legacy, Haiku, or `[1m]` default.
- **The lists stay hand-synced.** #1080 already noted a shared catalog as "worth considering, out of scope"; this guard is the workaround, not the fix.

Fixes #1660

### DIFF
--- a/src/backend/services/settings_service.py
+++ b/src/backend/services/settings_service.py
@@ -33,6 +33,8 @@ _PLATFORM_MODEL_CACHE_TTL = 60.0
 # this set after it was saved is treated as unset (→ platform default) by
 # `db.get_public_channel_model`, matching the #1080 graceful-degradation posture.
 PUBLIC_CHANNEL_MODELS = frozenset({
+    "claude-fable-5",
+    "claude-sonnet-5",
     "claude-opus-4-8",
     "claude-opus-4-7",
     "claude-opus-4-6",
--- a/src/frontend/src/views/Settings.vue
+++ b/src/frontend/src/views/Settings.vue
@@ -254,7 +254,9 @@
                       class="block flex-1 px-3 py-2 border border-gray-300 dark:border-gray-600 rounded-md shadow-sm focus:outline-none focus:ring-action-primary-500 focus:border-action-primary-500 dark:bg-gray-700 dark:text-white text-sm"
                     >
                       <option value="claude-sonnet-4-6">Claude Sonnet 4.6 — Fast + smart (recommended)</option>
-                      <option value="claude-opus-4-8">Claude Opus 4.8 — Most capable</option>
+                      <option value="claude-fable-5">Claude Fable 5 — Most capable, longest tasks (latest)</option>
+                      <option value="claude-sonnet-5">Claude Sonnet 5 — Fast + smart, 1M context (latest)</option>
+                      <option value="claude-opus-4-8">Claude Opus 4.8 — Most capable Opus</option>
                       <option value="claude-opus-4-7">Claude Opus 4.7</option>
                       <option value="claude-opus-4-6">Claude Opus 4.6</option>
                     </select>
--- a/tests/unit/test_1660_model_list_drift.py
+++ b/tests/unit/test_1660_model_list_drift.py
@@ -1,0 +1,133 @@
+"""Drift guard for the hand-synced model lists (#1660).
+
+`ModelSelector.vue` is the de-facto canonical picker (#1080). Two other lists are
+kept in sync with it **by hand** — there is no shared Python/Vue model registry:
+
+  1. `services/settings_service.PUBLIC_CHANNEL_MODELS` — server-validated whitelist
+     for the per-agent public-channel override (#894); current-gen only (no legacy).
+  2. The platform default-model `<option>` list in `views/Settings.vue` — the admin
+     fleet-wide default; Opus/Sonnet tiers only (#1080 AC), so Haiku is deliberately
+     absent and is NOT asserted here.
+
+The #1660 bug: ModelSelector gained the Claude 5 presets and neither list followed,
+so an admin couldn't pick them as the platform default and the public-channel PUT
+422'd them. These tests assert the current lineup reaches all three lists.
+
+Deliberately narrow: this pins the *current-generation* models by name rather than
+diffing whole lists, because the three lists are legitimately different sizes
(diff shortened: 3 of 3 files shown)


vybe (approved):
Validated via /validate-pr. Focused P1 bug fix (Fixes #1660, closing keyword present so the issue auto-promotes to status-in-dev on merge): +2 whitelist entries, +2 dropdown options, and a well-shaped drift guard that pins the current generation across all three hand-synced lists (including both subset directions and the PLATFORM_DEFAULT_MODEL_VALUE blank-select guard). Ran the 9 new tests locally against the PR head: all pass. Full CI green including the pytest head matrix and e2e.

PR #1663, opened 2026-07-16
docs(learnings): mirrored constants drift, and the mirror's tests pin the drift

obasilakis:
## Summary

Appends one entry to the `docs/memory/learnings.md` ledger. Docs only — no code.

This is the **#1638 class recurring**, one repo over. #1638's own entry records the trap that *"the test asserted the destructive default as the requirement"*; the same trap just reappeared in a constant mirrored across a repo boundary, where **three** tests asserted the diverged value as the requirement. That's exactly what the ledger exists to stop a third time, and it's what `/autoplan` reads before planning.

## The durable rule

A mirrored constant is a copy with no owner. *"Mirrors X"* in a comment is a hope, not a mechanism — and the comment is what makes the drift invisible, because it reads as if something enforces it. If you can't import the source of truth, a **parity test** that reads the real thing and fails on drift is mandatory.

Three refinements, each learned by getting it wrong first:

1. **Read the source, don't import it**, when the import is the very thing you were avoiding — a parity test that imports the heavy module inherits its side effects, passes in isolation, and fails in-suite.
2. **Assert the regression's signature, not a proxy** — "no value may equal the floor" false-fires the moment upstream legitimately adopts that number, and then *contradicts* the parity test.
3. **Grep for sibling constants** derived the same way before calling it fixed.

Plus a verification corollary: proving a guard fires by reintroducing a bug with a same-byte-length edit (`"7"` → `"5"`) won't invalidate `__pycache__` on restore — you'll chase a phantom failure.

## Disclosure check

Written generically and deliberately: it says **"a module in a separate repo"** and names no module, no internals, no panel, no entitlement/gating. Everything concrete in it is **OSS-side and already public** — `OPS_SETTINGS_DEFAULTS`, `cleanup_service`, `init_database()`, the 30/90/7/180/30 values, #1638 itself.

Worth flagging: `enterprise-docs-guard` would **not** have caught a leak here — its pattern (`SIEM|SCIM|client_portal|permissions_matrix|two_factor|enterprise_[a-z_]+`) matches none of this vocabulary. Green CI on this PR is not evidence of compliance; the compliance is in the wording, and it's worth a human read on that basis.

## Merge note

⚠️ **Expect a conflict with #1654.** That PR also appends an entry to `learnings.md` at the same EOF position (right after the 2026-07-15 #1638 entry). Whichever lands second needs a trivial rebase — both entries are additive and distinct, keep both.

## Test plan

- [x] Docs only — no code, no tests affected
- [x] Guard-pattern scan over the diff: zero tokens
- [x] Verified it names no module / internals / gating
- [x] Only `learnings.md` staged (worktree's submodule pointer deliberately excluded)

Context: `abilityai/trinity-enterprise#168` / PR `abilityai/trinity-enterprise#179`


vybe (approved):
Validated via /validate-pr: docs-only ledger entry (+4 lines), all checks green, disclosure check holds (names only OSS-side artifacts already public in architecture.md; the retention module's existence is already publicly documented). Merge-conflict warning with #1654 is moot — base already contains that entry and GitHub reports MERGEABLE.

Issue #237, opened 2026-04-01
bug: Slack file download fails — get_bot_token called with wrong argument


pavshulin:
## Summary

`SlackAdapter.download_file()` passes `message.metadata.get("team_id")` (a string) to `get_bot_token()`, but `get_bot_token()` expects the `team_id` string directly. However, looking at the actual signature, `get_bot_token(self, team_id: str)` takes a string — so `message.metadata.get("team_id")` is correct.

The actual bug: on the dev server, `get_bot_token` is receiving `message` (the full NormalizedMessage object) instead of `message.metadata.get("team_id")`.

## Fix

```python
# Wrong (passes full message object):
bot_token = self.get_bot_token(message)

# Correct (passes team_id string):
bot_token = self.get_bot_token(message.metadata.get("team_id"))
```

## Impact

All Slack file downloads fail silently — bot token lookup returns None, files are skipped.

## Introduced in

PR #235 (Slack inbound file sharing)

PR #1670, opened 2026-07-17
fix(sanitizer): linear KEY=value redaction — the #1661 CPU spin is the regex, not the pipe

### LINKED ISSUE #1661: bug: agent-server leaks a busy-spinning reader thread (100% CPU) on stdout-pipe-race / drain deadlock — #728 fix incomplete
## Summary

`agent-server.py` again pegs a full core at ~100% CPU indefinitely, via a **different trigger than #728**. A subprocess (headless `claude`) **stdout-pipe race / reader-thread drain deadlock** leaves a **leaked daemon reader thread busy-spinning** until the container is recreated. The #728 fix's own instrumentation is what surfaces it (`... reader threads are leaked daemon threads (pid=...). Issue #728.`), so the fix is present but does not cover this path.

Closed sibling: #728 (`agent-server.py spins at 90% CPU on OAuth token auth failure`). Same 100% CPU-spin symptom, different cause (reader-thread drain deadlock, not OAuth). Filing new rather than reopening since the trigger differs.

## Environment

- Trinity (platform): `d169e255` (PR #1645)
- Agent base image: `trinity-agent-base:latest`, **built 2026-07-15 13:30 UTC** (recent — contains the #728 fix code)
- Instance: construct; observed on `agent-cornelius-m` and `agent-agent-rick`
- Affected container recreated 2026-07-15 13:33 UTC (RestartCount=0); spin **recurred within ~27h** of that clean rebuild.

## Symptom

- Container CPU pegged ~100% (one core), **memory flat** (spin, not a leak).
- Occurs with **no active task execution** (no `claude`/`node` subprocess running).
- Recurs every few hours; self-clears when the stuck thread eventually drains, then reappears on the next task that hits the race. Persisted across a fresh image rebuild.
- On heavy-pipeline agents it has tripped the dispatch **circuit-breaker** (agent-unhealthy) during coincident load.

## Root-cause chain (from agent-server logs)

1. Headless task runs `claude --print --output-format stream-json`; a tool emits a large result (e.g. 44 KB, persisted-output), and an **stdio MCP child interleaves on the agent-server pipe**:
   ```
   WARNING agent_server.services.stream_parser: Failed to parse line as JSON: {"type":"user",...
   WARNING agent_server.services.headless_executor: [Recovery] Result event lost (stdout pipe race) ... recovering as soft success
   WARNING agent_server.services.headless_executor: ... Likely cause: stdio MCP child interleaved with claude on the agent-server pipe.
   ```
2. On subprocess exit, pipe close deadlocks against the reader thread's `TextIOWrapper` lock; drain budget blows and reader threads are leaked:
   ```
   WARNING agent_server.services.subprocess_pgroup: [Subprocess] Reader thread(s) still busy after process exit (pid=16375, stuck_count=1) — killing process group, then waiting 30s for natural drain
   WARNING agent_server.services.subprocess_lifecycle: [Subprocess] Drain budget (90s) exceeded — safe_close_pipes may have deadlocked with reader thread's TextIOWrapper lock; reader threads are leaked daemon threads (pid=16782). Issue #728.
   ```
3. The leaked daemon reader thread **busy-spins at 99.9% CPU** until the process is restarted.

## Evidence it is the leaked reader thread (not task work)

- `top -bH` inside the container: thread **TID 16784** in state `R`, 99.9% CPU, CPU-time climbing 2:52 → 5:10 across samples; main + other threads idle.
- `/proc/16784/wchan = 0`, `/proc/16784/syscall` empty → **pure userspace spin**, not blocked on I/O.
- TID `16784` is sequential to the leaked `pid=16782` named in the `Issue #728` warning → it is that subprocess's leaked reader thread.
- Agent-server (PID 35) thread count elevated to 7 while spinning; **`docker restart` drops it to baseline and CPU to ~3%** (confirmed mitigation, but temporary).

## Reproduction signature

Heavy scheduled-pipeline agent (`/pipeline-tick`, `/heartbeat-procedure`) whose tools produce **large tool outputs** while an **stdio MCP server** is attached → periodic `stdout pipe race` → `Drain budget exceeded` → one leaked reader thread pinned at 100% CPU.

## Suggested area

`agent_server/services/subprocess_lifecycle.py` (`safe_close_pipes` / drain budget) and `subprocess_pgroup.py` reader-thread teardown: when the drain budget is exceeded, the reader threads are abandoned as daemons and continue to spin. They need to be **forcibly terminated / their fds hard-closed** so the abandoned thread exits its read loop instead of busy-spinning. The stdio-MCP-interleave on the agent-server pipe (the upstream trigger) is a second, related fault worth isolating (MCP child should not share claude's stdout pipe).

## Ops handling

Ops tracks this as `trinity-ops-agent#133` (+ `#208` for agent-rick, `#134` for the CB trip). Config-only mitigation applied: `docker restart agent-cornelius-m` (clears the spin; recurs). No source changes made (ops boundary). This report is the dev-team handoff for the durable fix.

dolho:
Fixes #1661. **The spinning thread was never blocked on a pipe — it was computing.** That is why the #728 and #1502 pipe fixes are present in the affected image and do not cover this path.

Related to #1661

## Root cause

The sanitizers' patterns describe **variable names** (`.*TOKEN.*` = "a name containing TOKEN"), but stage 3 composed each one into a **line-scanning** regex:

```
(.*TOKEN.*)=(["\']?)([^\s"\']+)\2
```

Two unbounded `.*` around a literal, re-scanned from every offset of the line. Measured on this branch:

| line size | `sanitize_text` (before) |
|---|---|
| 1 KB | 45 ms |
| 2 KB | 274 ms |
| 4 KB | 1 501 ms |
| 8 KB | **8 572 ms** |

~5.7× per 2× input (≈ n^2.5) → **~10 CPU-minutes for the 44 KB tool-result line** named in the report. `.*TOKEN.*` alone accounts for 6.8s of the 8 KB figure.

The headless reader calls `sanitize_dict(raw_msg)` on **every** parsed message (`headless_executor.py:610`), so one 44 KB tool result puts the reader thread into a ~10-minute C-level regex burn. That single fact explains every observation in the report:

| Reported symptom | Explanation |
|---|---|
| 100% of one core, `R`, `wchan=0`, no syscall, memory flat | a regex in C, not I/O |
| "Reader thread(s) still **busy** after process exit" | busy, literally — computing |
| kill group → 30s wait → 90s budget → "leaked" (#728/#1502) | those make a **blocked** reader EOF; this one was never blocked |
| "Result event lost (stdout pipe race) → soft success" | the result line wasn't lost — the reader hadn't reached it yet |
| spin persists with **no active task**, then self-clears | the regex finishes |
| recurs on the next large-output task; `docker restart` clears it | next big line, same burn |

## Fix

One linear pass: find `key=value` pairs with a delimiter-bounded key, then test the (short) key against the name patterns.

| line size | before | after |
|---|---|---|
| 8 KB | 8 572 ms | **0.4 ms** |
| 44 KB | ~10 min | **1.2 ms** |
| 500 KB | — | 18 ms |
| 200 KB single unbroken token | — | 11.7 ms |

The reader's actual per-message call, `sanitize_dict` on a 44 KB `tool_result`: **1.6 ms**.

## It also fixes a security bug

The greedy `.*` spans to the **last** `=` on the line, so on a multi-pair line the old code redacted the **wrong** pair and left the real credential in place:

```
GH_TOKEN=supersecret PLAIN=harmless
  old -> GH_TOKEN=supersecret PLAIN=***REDACTED***      # leaks the token
  new -> GH_TOKEN=***REDACTED*** PLAIN=harmless

env: ANTHROPIC_API_KEY=sk-ant-realkey123456 LOG_LEVEL=debug
  old -> ... ANTHROPIC_API_KEY=sk-ant-realkey123456 LOG_LEVEL=***REDACTED***
  new -> ... ANTHROPIC_API_KEY=***REDACTED*** LOG_LEVEL=debug
```

Stages 1–2 (known env values; value-shaped patterns like `sk-ant-[…]{20,}`) mask this for many real keys, which is why it went unnoticed — the exposure is a credential whose *value* matches nothing known, on a line with 2+ pairs.

**Suffix matching is preserved deliberately.** The old regex could start matching mid-token, so `ANTHROPIC_.*` redacts `MY_ANTHROPIC_KEY=x`. A `fullmatch` rewrite would quietly redact *less* — a leak, not a speedup — so the key test is anchored with `\Z` to reproduce the old reach exactly.

## Scope

- **Both copies fixed.** The backend layer has the identical flaw and sanitizes execution logs before DB write, so it burned backend CPU on the same input.
- **Pattern lists untouched.** They differ (agent-side is a superset: `DB_.*`, `REDIS_.*`, …), so `MY_DB_PASS=x` is redacted agent-side but not backend-side. That asymmetry predates this issue and is now pinned by a test — widening a list changes what gets redacted and belongs in its own change.
- **Not addressed** (the upstream trigger, which the report itself suggests isolating): the stdio MCP child interleaving on claude's stdout pipe. This fix removes the CPU burn and the drain/leak/soft-success cascade; the interleave remains its own fault.

## Verification

31 new tests (`tests/unit/test_1661_sanitizer_linear.py`) covering both copies: 44 KB budget, a shape assertion that 8× input isn't ~300× time (the old curve), the giant-unbroken-token case, multi-pair redaction, and the single-pair behaviour contract captured from the *pre-fix* implementation. 510 passing across the sanitizer/executor suites; the 117 existing sanitizer tests pass unchanged.

Along the way I also reproduced the `safe_close_pipes` ↔ `readline()` deadlock the log line mentions (`close()` never returns while a reader is mid-`readline`) — real, but it leaves the reader **blocked at 0% CPU**, so it is not the spin. Noting it as a separate latent issue rather than folding it in.

### DIFF
--- a/docker/base-image/agent_server/utils/credential_sanitizer.py
+++ b/docker/base-image/agent_server/utils/credential_sanitizer.py
@@ -71,6 +71,51 @@ SECRET_VALUE_PATTERNS = [
 
 # Compiled patterns for performance
 _sensitive_var_re = [re.compile(p, re.IGNORECASE) for p in SENSITIVE_VAR_PATTERNS]
+
+# --- #1661: linear KEY=value redaction -------------------------------------
+# The patterns above describe VARIABLE NAMES (`.*TOKEN.*` means "a name
+# containing TOKEN"). Stage 3 below used to compose each one into a
+# LINE-scanning regex — `(.*TOKEN.*)=(["\']?)([^\s"\']+)\2` — i.e. two
+# unbounded `.*` around a literal, re-scanned from every offset of the line.
+# Cost was superlinear in line length: ~6.8s for `.*TOKEN.*` alone on an 8 KB
+# line, ~10 CPU-minutes on the 44 KB tool-result lines that triggered #1661.
+# That is what pegged a core: the agent-server reader thread was not blocked on
+# a pipe, it was *computing* — which is why the #728/#1502 pipe fixes never
+# helped, and why it "self-cleared" (the regex eventually finished).
+#
+# One linear pass instead: find `key=value` pairs, then test the (short) key.
+# The key is bounded by delimiters, and the lookbehind stops the engine from
+# retrying every offset inside a long unbroken token (a 44 KB base64 blob would
+# otherwise reintroduce quadratic cost).
+_KV_LINE_RE = re.compile(r'(?<![^\s"\'=])([^\s"\'=]+)=(["\']?)([^\s"\']+)\2')
+
+# A name pattern must match a SUFFIX of the key, not the whole key: the old
+# composed regex could start matching mid-token, so `DB_.*` redacted
+# `MY_DB_PASS=x`. Anchoring with \Z reproduces exactly that (a `fullmatch`
+# rewrite would silently redact LESS — a leak, not a speedup).
+_SENSITIVE_KEY_SUFFIX_RE = [
+    re.compile(r'(?:' + p + r')\Z', re.IGNORECASE) for p in SENSITIVE_VAR_PATTERNS
+]
+
+
+def _is_sensitive_kv_key(key: str) -> bool:
+    """True if `key` (a short `KEY=` name) names a credential."""
+    return any(r.search(key) for r in _SENSITIVE_KEY_SUFFIX_RE)
+
+
+def _redact_kv_match(match: "re.Match") -> str:
+    """Redact the value of a sensitive `key=value` pair, keep everything else.
+
+    Mirrors the old replacement byte-for-byte: the value (and its quotes, which
+    the old `\1=` replacement also dropped) becomes the placeholder; text
+    around the pair is untouched.
+    """
+    key = match.group(1)
+    if _is_sensitive_kv_key(key):
+        return f"{key}={REDACTION_PLACEHOLDER}"
+    return match.group(0)
+
+
 _secret_value_re = [re.compile(p) for p in SECRET_VALUE_PATTERNS]
 
 # Cache for known credential values (loaded from environment)
@@ -163,13 +208,10 @@ def sanitize_text(text: str) -> str:
 
     # 3. Redact key=value pairs where key is sensitive
     # Handle: KEY=value, KEY="value", KEY='value'
-    for var_pattern in _sensitive_var_re:
-        # Match: SENSITIVE_VAR=somevalue or SENSITIVE_VAR="somevalue"
-        kv_pattern = re.compile(
-            r'(' + var_pattern.pattern + r')=(["\']?)([^\s"\']+)\2',
-            re.IGNORECASE
-        )
-        result = kv_pattern.sub(r'\1=' + REDACTION_PLACEHOLDER, result)
+    # #1661: ONE linear pass (see _KV_LINE_RE) — this used to compile a
+    # line-scanning regex per name pattern, which cost CPU-minutes on a large
+    # line and pegged a core.
+    result = _KV_LINE_RE.sub(_redact_kv_match, result)
 
     return result
 
--- a/docker/base-image/agent_server/utils/credential_sanitizer.py
+++ b/docker/base-image/agent_server/utils/credential_sanitizer.py
@@ -87,6 +87,13 @@ _sensitive_var_re = [re.compile(p, re.IGNORECASE) for p in SENSITIVE_VAR_PATTERN
 # The key is bounded by delimiters, and the lookbehind stops the engine from
 # retrying every offset inside a long unbroken token (a 44 KB base64 blob would
 # otherwise reintroduce quadratic cost).
+# The lookbehind is load-bearing, not decoration: it is what keeps this LINEAR.
+# Without it the engine retries the key at every offset inside a long non-matching
+# run, which is exactly the quadratic shape this issue is about. CodeQL reports
+# `py/polynomial-redos` here because its model ignores lookbehinds; the adversarial
+# cases it names (`!`*n, `!=`+`!=!`*n) are pinned as tests in
+# tests/unit/test_1661_sanitizer_linear.py and run in ~1ms at 64 KB. Do not
+# "simplify" the lookbehind away.
 _KV_LINE_RE = re.compile(r'(?<![^\s"\'=])([^\s"\'=]+)=(["\']?)([^\s"\']+)\2')
 
 # A name pattern must match a SUFFIX of the key, not the whole key: the old
--- a/src/backend/utils/credential_sanitizer.py
+++ b/src/backend/utils/credential_sanitizer.py
@@ -55,6 +55,48 @@ SENSITIVE_KEY_PATTERNS = [
 _secret_value_re = [re.compile(p) for p in SECRET_VALUE_PATTERNS]
 _sensitive_key_re = [re.compile(p, re.IGNORECASE) for p in SENSITIVE_KEY_PATTERNS]
 
+# --- #1661: linear KEY=value redaction -------------------------------------
+# The patterns above describe KEY NAMES (`.*TOKEN.*` means "a name containing
+# TOKEN"). The stage below used to compose each one into a LINE-scanning regex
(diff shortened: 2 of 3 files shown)


dolho:
### CodeQL `py/polynomial-redos` (#237) — investigated, not waved through

CodeQL flagged the new `_KV_LINE_RE` as a **high-severity ReDoS**. Given this PR exists *because* of a regex complexity pathology, that got a real investigation rather than a dismissal.

**It is a false positive in effect, and here is why:** the lookbehind `(?<![^\s"'=])` means the key can only start after a delimiter, so the engine cannot retry `([^\s"'=]+)=` at every offset inside a long non-matching run — which is the quadratic shape CodeQL assumes. Its model does not account for lookbehinds.

**Measured**, both strings CodeQL names plus 9 more adversarial shapes:

| input | 2 KB | 8 KB | 32 KB |
|---|---|---|---|
| `!`*n (CodeQL) | 0.0 ms | 0.1 ms | 0.6 ms |
| `!=` + `!=!`*n (CodeQL) | 0.0 ms | 0.0 ms | 0.1 ms |
| unterminated quote `K="`+`x`*n | 0.1 ms | 0.3 ms | 1.0 ms |
| trailing `=`, missing values, all-`=`, quoted pairs, huge value | — | — | ≤ 7.4 ms |

Every shape: **4× input → ~4× time**. Worst case 7.4 ms at 32 KB (vs 8 572 ms for an 8 KB line before this PR).

**What I did instead of just dismissing it:**

1. The adversarial cases — including CodeQL's two strings — are now **tests** (`test_adversarial_inputs_stay_linear`, 64 KB budget). If someone later "simplifies away" the lookbehind, they reintroduce the exact pathology this issue is about and **CI fails** instead of a reviewer having to remember this thread.
2. The code says the lookbehind is load-bearing and why.
3. Only then dismissed alert #237 as a false positive, citing the tests.

Happy to restructure the regex instead if a reviewer prefers a shape the analyser can prove — the trade is a less faithful match extent (the quote handling needs a backreference or an alternation) for a green analyser, and I judged the executable proof the better deal.

obasilakis (approved):
Verified the diagnosis independently rather than taking the numbers on trust — it holds.

Reproduced the old cost curve from scratch: 8 KB = 8533 ms (PR claims 8572 ms), growth ~6.8x per 2x input, so the 44 KB tool-result line is genuinely CPU-minutes. New path at 44 KB = 1.2 ms (PR claims 1.2 ms). Confirmed `sanitize_dict(raw_msg)` runs inside the reader loop (`headless_executor.py:610`, same shape in `claude_code.py:358`), which is what puts the burn on the reader thread and explains every observation in the report — state R, wchan=0, no syscall, memory flat, self-clears, recurs on the next big output. The re-diagnosis is correct and #1661's own suggested area was a red herring.

Also confirmed the lookbehind is load-bearing rather than folklore: removing it takes 64 KB from 1.8 ms to 23404 ms (cleanly quadratic). The 'do not simplify away' comment is earned and should stay.

Two non-blocking notes:

1. The comment 'Anchoring with \Z reproduces exactly that' overclaims. `\Z` preserves reach within a token, but the old `.*` also spanned spaces and quotes across tokens. A differential run over a realistic corpus shows 9/19 lines redact less. Most of that is accidental over-redaction worth losing (the old net only ever hit the LAST pair on the line, so it was never a real property). But `"GH_TOKEN"=v` and `GH_TOKEN ="v"` are genuinely-sensitive key names the old code reliably redacted and the new one never will. Rare formats, not worth blocking — but the comment asserts an exactness that doesn't hold, and the next maintainer will trust it. Worth softening to state the actual boundary (delimiter-bounded key).

2. The two copies now share ~50 lines of scan logic with no parity guard. `test_the_two_copies_carry_different_pattern_lists` pins the list asymmetry nicely, but nothing pins that `_KV_LINE_RE` / `_is_sensitive_kv_key` / `_redact_kv_match` stay in sync — and this bug existed identically in both copies precisely because they drift silently.

On the declined residuals: agreed on both calls. Post-fix the interleave degrades to a parse warning and the `safe_close_pipes` deadlock parks a thread at 0% CPU — neither reproduces the P1 symptom. Leaving them in the PR body rather than inflating the backlog; the interleave will re-surface on its own with better evidence than we could write today.

158 tests pass locally (31 new + 127 existing unchanged).

PR #1674, opened 2026-07-17
fix(retention): refuse a prune that would delete most of a table (#1644)

### LINKED ISSUE #1644: bug: retention prunes have no blast-radius guard — a mistyped window deletes most of a table within 5 minutes

obasilakis:
## Summary

**#1638 made the retention *defaults* safe. It added no guard on the prune itself.** This adds the backstop that doesn't care *how* a bad window arrived.

Before each of the **7** window-driven destructive prunes, count the candidate set (bounded) and if it exceeds **1000**, **refuse** — log ERROR, raise an ops alarm, delete nothing. An admin unblocks it with one call.

The threat isn't hypothetical. `PUT /api/settings/ops/config` is `Dict[str, str]` → `if key in OPS_SETTINGS_DEFAULTS` → bare upsert. **No type, range, clamp, or audit.** Garbage fails *safe* (`"abc"` → 0 → sweep disabled); **a small valid integer is the catastrophic input**, and no range check catches `5` typed for `50`.

## Two corrections to the issue's own premises

- **There is no per-cycle cap.** #1644 says the sweep "deletes up to `RETENTION_CHUNK_SIZE_PER_CYCLE = 5000` rows per cycle". It doesn't — 6 of 7 accessors are `while True:` drain loops; `chunk_size` bounds each *transaction*. One call deletes everything past the cutoff. **#1638 deleting 5352 rows in one startup sweep is the proof** — a real 5000 cap couldn't produce that number. Comment, docstring, and architecture.md corrected.
- **The enterprise clamp protects nothing.** `_clamp` is `max(value, 5)` — a *monetization floor*, not a safety check. An entitled admin typing `5` is accepted and *is* the #1638 scenario. It's also a parallel endpoint, so every edition routes around it.

| decision | why |
|---|---|
| **Stateless detection**, not a last-seen-window watermark | The watermark row is deletable via the same endpoints that cause the bug → silently disarms the guard. `learnings.md` #1638 Lesson 2. The ack *is* state, but deleting an ack fails **safe**. |
| **Absolute counts, no percentage** | No denominator exists for `execution_log` (its predicate *includes* `execution_log IS NOT NULL`); a total costs a full scan and divides by zero on an empty table; and a percentage **inverts** on the agent sweep — 3 purged agents ≈ 0% of any table but 3 destroyed volume sets. Steady state is a trickle; any anomaly is large. Table size is irrelevant. |
| **Threshold is a constant, not a setting** | Started as an operator knob. Nobody can reason about the value, and the panel needed a caption explaining *bigger is worse* — a control that must explain which way is safe is the wrong control. Worse: a mutable constant read at action time gating a destructive op **is #1638 one level up**. Deleting the knob deleted its clamp, endpoint, blocklist entry, and a whole fail-closed branch. |
| **Ack is the gate; the queue item is only an alarm** | `create_item` is a blind `INSERT … ON CONFLICT DO NOTHING` — a load-bearing item would **wedge shut permanently** once it hit any terminal state (Clear All → `cancelled`), and `prune_terminal_items` would delete the approval at 90 days: *the sweep deleting its own authorization*. Decorative ⇒ both dissolve. |
| **Counts share each prune's predicate by construction** | A mirrored copy drifts. Matters most for `operator_queue`, whose predicate derives a second cutoff internally (`max(responded, terminal)`). |
| **Agents floor at 0** | Every purge destroys Docker volumes (#1581). Unrecoverable ⇒ always acked. |

**Fail-closed everywhere.** Count throws, ack lookup throws → refuse. A guard that fails open is worse than no guard: it manufactures confidence.

## Also fixed (found while investigating)

- 🐛 **`RETENTION_OPS_KEYS` held 5 of 7 windows** — so `/ops/reset` **deleted** `agent_reports_retention_days` and `operator_queue_retention_days` while returning `"retention windows unchanged"`, and both were invisible to `GET /api/settings/retention` and boot logging.
- 🐛 **Retention inputs had `border-gray-300` without `border`** — no width applied, so browsers rendered default number-input chrome. Now matches the app-wide convention (96 other inputs use it).
- 🐛 **A 422 dumped FastAPI's raw Pydantic array into the page.** Added `utils/apiError.js` + pre-flight validation. *~198 other callsites share this latent bug — left alone.*

## ⚠️ Security gap filed separately

**trinity-ops-agent#232** (private — public `SECURITY.md` forbids disclosing here). `require_admin` never calls `reject_agent_principal`, and an agent key resolves to its owner **carrying the owner's role**. On a default install (Cornelius/trinity-system are admin-owned) an agent's own MCP key can `PUT /api/settings/execution_row_retention_days {"value":"1"}` through the unblocklisted catch-all. **This PR does not close it** — but the ack endpoint applies `reject_agent_principal` explicitly rather than trusting `require_admin`, so the gate isn't reachable that way (verified: 403).

## Test Plan

- [x] `pytest tests/unit/test_1644_retention_guard.py` — **26 pass** (dual-backend SQLite + PG)
  - predicate parity: count == what each prune actually touches
  - fail-closed: count throws → refuse; ack lookup throws → refuse; failed alarm never flips a refusal
  - ack bound to window, scoped to setting, single-use
  - threshold direction pinned (`<= PREVIOUS`) + asserted **not** settings-backed
  - sentinel parity with canary L-03 + uncreatable as an agent name
  - the #1638 scenario refused end-to-end; `evaluate()` deletes nothing
- [x] **Full unit suite: 4133 passed.** The one failure (`test_1474_read_boundary_z`) is **pre-existing** — verified identical on stashed `dev`.
- [x] Verified live: clamp rejects `999999999`; catch-all 422s the guard key; ack 409s on window mismatch; **agent principal with `role="admin"` gets 403**.
- [x] Frontend builds; `npm run check:tokens` passes.

## Expected behaviour worth knowing

A legitimate first-enable of retention on a mature install **will** trip the guard once. That's intended — the guard can't distinguish a large legitimate backlog from a mistyped window, so it asks once. That's what keeps it a small predicate instead of a "was this narrowed?" detector.

Fixes #1644


vybe (approved):
Validated via /validate-pr. P1 fix with a clean Fixes #1644 link, architecture.md + requirements/lifecycle-observability.md updated, and a named regression suite (`test_1644_retention_guard.py`). Verified the load-bearing properties in the diff: fail-closed on every guard error path, ack endpoint is admin + `reject_agent_principal` (agent keys can't approve mass deletion), ack is window-bound (409 on mismatch) + single-use (consumed only after a successful prune), ACK_KEY_PREFIX blocklisted from the generic settings catch-all, and the alarm carries counts/identifiers only (G-04 lesson). The `_retention-guard` sentinel can't collide with a real agent (sanitize strips leading `_`) and is excluded from canary L-03. Ran the 4 touched unit test files locally on the PR head merged with current dev (incl. #1688): **87 passed**; #1688's operator-queue guard tests also re-pass on the merged state (6 passed). Non-blocking suggestion: consider adding `retention-guard-` to `_RESERVED_ID_PREFIXES` so an agent-authored lookalike alarm id is folded out at the sync seam (same mimicry-only class as the #1688 LOW).

Issue #516, opened 2026-04-26
bug: SIGKILL/timeout terminations of claude subprocess misclassified as authentication failure


vybe:
## Summary

When the agent-server's claude subprocess is killed by an external signal (schedule timeout SIGKILL, OOM-kill, parent SIGTERM, etc.), agent-server falls through into the auth-failure heuristics and surfaces the error as `Authentication failure: Subscription token may be expired or revoked. Generate a new one with 'claude setup-token'.` In the operator UI this looks identical to a real expired-token incident, sending operators down the wrong diagnostic path while the actual cause (skill exceeded its schedule timeout) goes unnoticed and keeps repeating every cron tick.

This is the same shape as #361 (max-turns misclassified as auth) — same file, same heuristic block — but a different exit path. #361 only special-cased the max-turns exit; the signal-kill exit still falls through.

## Component

Agent Runtime (`agent_server.services.claude_code`)

## Priority

P2 — feature impaired with workaround (raise the schedule timeout); diagnostic UX is broken so the workaround is hard to discover.

## Error

Symptom in agent-server log on every signal-kill:

```
[Subprocess] Reader thread(s) stuck after process exit (pid=NNNNN, stuck_count=1) — killing process group and closing pipes to unwind
[Subprocess] 1 reader thread(s) leaked for pid=NNNNN after close+killpg; continuing anyway
[Headless Task] Error reading stdout: I/O operation on closed file.
[Headless Task] Auth failure (fallback detection): Subscription token may be expired or revoked. Generate a new one with 'claude setup-token'.
POST /api/task HTTP/1.1 503 Service Unavailable
```

The first three lines are the SIGKILL fingerprint (process gone, but stdout pipe is still held → reader thread stuck → forced unwind). The fourth line is the misclassification.

What surfaces in the platform `schedule_executions.error` column:

```
Authentication failure: Subscription token may be expired or revoked. Generate a new one with 'claude setup-token'.. Check subscription token or API key configuration.
```

…even though the token is fine, sister agents on the same subscription are running successfully at the same moment, and the actual cause is the schedule timeout SIGKILL.

## Location

- **File**: `docker/base-image/agent_server/services/claude_code.py`
- **Lines**: ~1255–1280 (the `if return_code != 0:` block, after the max-turns special-case from #361)
- Two heuristic paths both wrongly fire for signal-kills:
  - `_is_auth_failure_message(error_preview) or _is_auth_failure_message(verbose_transcript)` — line ~1263
[...]

PR #1675, opened 2026-07-17
fix: surface Claude Code execution_error instead of a success placeholder (#1673)

### LINKED ISSUE #1673: bug: agent-server masks error_during_execution as a successful `context: fork` placeholder — failed session resumes recorded as success, Sessions-tab fallback never fires
## Summary

The agent-server records Claude Code `error_during_execution` results as **successful** executions. When a headless task fails with an error result whose `result` text is empty (e.g. `--resume` against a Claude session whose JSONL no longer exists: "No conversation found with session ID"), the finalize path falls through to the #160 `context: fork` placeholder branch and returns HTTP 200 with "(Task completed with no direct output — skill may use `context: fork`.)". The execution is stored as `status=success`, `error=null`.

Downstream this defeats the Sessions-tab self-healing designed in EXEC-023 Phase 2.2: the router's resume-not-found fallback only fires on `result.status != "success"`, so the stale cached Claude UUID is never cleared — worse, step 6 **re-caches the same stale UUID** (the result event echoes the resume target as `session_id`). The session is permanently broken: every turn "succeeds" in ~2s at $0 with the placeholder text, `consecutive_resume_failures` stays 0, and nothing surfaces to the user or operator.

Observed in production on a dev-tracking instance (2026-07-17): a Sessions-tab session created 2026-05-14 whose cached Claude JSONL no longer exists on disk (consistent with Claude Code's default 30-day `cleanupPeriodDays` transcript cleanup — guaranteed to happen to any session idle >30 days). Two consecutive user turns each failed the resume and were both recorded as green successes.

## Component

Agent Runtime (agent-server: `headless_executor`, `stream_parser`) — consequence in Backend (`routers/sessions.py`)

## Priority

P1

## Error

What Claude Code actually emitted (captured in the stored `execution_log`, while the execution row says `success`):

```json
{
  "type": "result",
  "subtype": "error_during_execution",
  "is_error": true,
  "num_turns": 0,
  "total_cost_usd": 0,
  "errors": ["No conversation found with session ID: <claude-session-uuid>"]
}
```

Agent-server log sequence for the failing turn:

```
[Task] Resuming session <claude-session-uuid>: ...
[Headless Task] Resuming session: <claude-session-uuid>
[Headless Task] Starting task <execution-id>: claude --print --output-format stream-json --verbose...
WARNING stream_parser: Claude Code result is_error=true: type=execution_error, message=
WARNING [Headless Task] Task <execution-id> produced a result but did not exit — finalizing early, terminating teardown
INFO [Headless Task] Task ... exited cleanly with no assistant text in the parent stream (cost=$0, duration=0ms) — likely a `context: fork` skill. Returning placeholder response.
POST /api/task HTTP/1.1" 200 OK
```

Note `message=` is empty in the warning — the parser reads `msg["result"]`, but for `error_during_execution` the text lives in `msg["errors"]`.

## Location

Three cooperating defects:

1. **`docker/base-image/agent_server/services/stream_parser.py` (~L281–296)** — the generic `execution_error` branch sets `metadata.error_message = result_text` (from `msg.get("result", "")`), which is empty for `error_during_execution`; the actual message is in `msg.get("errors", [])`. (The `max_turns` branch two lines up already reads `errors` — this branch doesn't.)
2. **`docker/base-image/agent_server/services/headless_executor.py` (~L907–935, finalize)** — the error dispatch handles `error_type` values `rate_limit`, `max_turns`, and `authentication_failed`, but **`execution_error` is never checked**. With `return_code == 0` (finalize-early path) and `cost_usd == 0` (present, so `is not None`), control falls into the #160 empty-response placeholder branch (~L1055–1090) and returns 200 success.
3. **`src/backend/routers/sessions.py` (~L645–680, ~L703)** — by design the resume-not-found fallback (`_is_resume_not_found` markers: "no conversation found" / "session not found") clears the cached UUID and retries cold — but only when `result.status != "success"`. Because of (2) it never fires, and step 6 re-caches the stale UUID from `result.session_id`, locking the session into the broken state.

## Root Cause

Claude Code error results with empty `result` text are indistinguishable from the legitimate `context: fork` "clean exit, no parent-stream text" case in the finalize path, because `metadata.error_type == "execution_error"` is never consulted before the placeholder branch. The placeholder branch was added in #160 specifically for *clean* exits; an `is_error=true` result should never reach it.

## Reproduction Steps

1. In the Sessions tab, create a session with an agent and exchange at least one turn (a Claude UUID gets cached).
2. Remove the corresponding Claude Code JSONL transcript inside the agent container (equivalently: let the session idle past Claude Code's `cleanupPeriodDays`, default 30 days, or recreate the container without that transcript).
3. Send another message in the same session.
4. Observe: HTTP 200, assistant reply is the `context: fork` placeholder text, execution recorded `status=success` / `error=null` / `num_turns=0` / cost $0; the `session_resume_fallback` log line never appears; `consecutive_resume_failures` stays 0; every subsequent turn repeats the failure.

## Suggested Fix

Surface `execution_error` in finalize before the empty-response handling, and make the parser fall back to `errors[]` for the message:

```python
# stream_parser.py — generic execution_error branch
errors = msg.get("errors", [])
metadata.error_type = "execution_error"
metadata.error_message = result_text or (errors[0] if errors else "Execution error")
```

```python
# headless_executor.py — finalize, alongside the rate_limit/max_turns/auth checks
if ctx.metadata.error_type == "execution_error":
    err = sanitize_text(ctx.metadata.error_message or "Execution error")
    logger.error(f"[Headless Task] Claude Code execution error: {err[:300]}")
    raise HTTPException(status_code=502, detail=f"Execution error: {err[:300]}")
```

With that in place, the existing sessions-router fallback works as designed: the 502 detail contains "No conversation found", `_is_resume_not_found` matches, the stale UUID is cleared, and the turn retries cold — the session self-heals with no user-visible failure. Agent containers need a base-image rebuild to pick up the agent-server fix.

## Environment

- Trinity version: dev @ `13a29857` (v0.8.0 + dev)
- Agent base image: built from the same tree
- OS: Ubuntu 22.04, Docker prod compose

## Related

- #160 — origin of the `context: fork` placeholder branch (this bug is that branch swallowing genuine error results)
- EXEC-023 Sessions tab, Phase 2.2 (resume-failure fallback that this defeats)

obasilakis:
## Summary

- A Claude Code `result` with `is_error=true` and **empty** `result` text was recorded as a **successful** execution — it fell through to the #160 `context: fork` placeholder and returned HTTP 200 (`status=success`, `error=null`).
- The canonical trigger is `--resume` against a session whose JSONL no longer exists. Claude Code's default 30-day transcript cleanup makes this **guaranteed** for any session idle >30 days.
- Downstream this defeated the Sessions-tab self-healing: the resume fallback only fires on `result.status != "success"`, so the stale UUID was never cleared — and step 6 **re-cached** it from the echoed `session_id`. Sessions wedged permanently: every turn "succeeded" in ~2s at $0 with placeholder text, `consecutive_resume_failures` stayed 0, nothing surfaced.

Observed in production 2026-07-17 on a session created 2026-05-14 — two consecutive user turns both recorded green.

## Root cause — two cooperating defects

| # | File | Defect |
|---|------|--------|
| 1 | `stream_parser.py` | Generic `execution_error` branch read `msg["result"]` (empty for this subtype) and never fell back to `msg["errors"]` → `error_message = ""`. The `max_turns` branch two lines up already read `errors[]`. |
| 2 | `headless_executor.py` | Finalize dispatched on `error_type` for `rate_limit` / `max_turns` / `authentication_failed` but **never `execution_error`**. With `return_code == 0` and `cost_usd` present, control reached the #160 placeholder branch. |

`routers/sessions.py` needed **no change** — its fallback was correct all along, just never reachable.

## Changes

- `docker/base-image/agent_server/services/stream_parser.py` — fall back to `errors[0]` when result text is empty; log the resolved message.
- `docker/base-image/agent_server/services/headless_executor.py` — surface `execution_error` as **502** before the empty-response handling.
- `tests/unit/test_1673_execution_error_not_success.py` — 11 new tests.
- `tests/registry.json` — registered.

## Design notes

- **Why 502, not 503** — keeps it clear of SUB-003's auth auto-switch, which substring-matches on auth wording.
- **No collision with #678** — `_is_reader_race_signature` requires a **dict** body with `recovery_attempted`; this detail is a plain string, so the 502 auto-retry can't fire. Pinned by a test.
- **Placement** — after the `return_code != 0` block, so it can't preempt #516 signal classification (a SIGKILLed timeout must stay a signal exit). Non-zero exits were already failing correctly; only the clean-exit path was broken.
- **#160 preserved** — the placeholder still fires for genuine clean exits (`error_type is None`). Narrowed, not removed; pinned by a test and the existing #160 suite.

## Test Plan

- [x] 11 new tests pass: `pytest tests/unit/test_1673_execution_error_not_success.py -v`
- [x] **Tests proven to catch the bug**: reverting only the two source fixes turns 9 of 11 red. The 2 that stay green are the regression guards for existing behavior (correct).
- [x] Neighbours unaffected: #160 fork suite, pipe-drop, empty-result-classification, finalize-snapshot, #970 timeout — 50 passed.
- [x] Broader sweep (`-k "stream or parser or session or headless or task_execution or 678 or 160 or 520"`) — 289 passed, 1 skipped.

### Verified live on a real agent (base image rebuilt, `cornelius`, Claude Max subscription)

**A/B against the real `POST /api/task` + `headless_executor`, identical request both times:**

| | pre-fix | post-fix |
|---|---|---|
| HTTP | **200 OK** | **502** |
| body | `"(Task completed with no direct output — skill may use \`context: fork\`.)"` | `"Execution error: No conversation found with session ID: …"` |

Pre-fix, the response body carried the placeholder while the `execution_log` beside it read `"subtype":"error_during_execution","is_error":true` — the failure sold as success, exactly as reported.

**Reproducing the bug required care, and this is worth knowing for anyone re-testing it.** The naive repro (delete a session's JSONL, send a turn) does **not** reproduce it: local claude 2.1.212 emits the `error_during_execution` event *and exits 1 immediately*, so it takes the pre-existing `return_code != 0` path, which already surfaced the error correctly and self-healed. Verified: pre-fix code also recovers on that path.

The real bug needs the **teardown race** the production trace shows — claude emits the result but lingers past `_WAIT_POLL_S = 2.0` (a stdio MCP child holding the pipe), so finalize-early forces `return_code = 0` and the placeholder branch becomes reachable. Reproduced deterministically with a shim emitting the verbatim claude event then holding stdout.

That also explains why this is intermittent in production rather than constant: it fires only when teardown outlives the 2s poll.

- [x] Verified live (see above). Agent containers still need a base-image rebuild + recreate to pick this up on any deployment.
- [x] The `session_resume_fallback` → clear stale UUID → cold retry → self-heal chain confirmed end-to-end on the sibling path (`stale_uuid=…` logged, `cached_claude_session_id` rotated, reply returned).

Fixes #1673


vybe (approved):
Validated via /validate-pr. Both cooperating defects traced (stream_parser errors[] fallback + finalize execution_error→502 before the #160 placeholder branch); placement after return_code!=0 preserves #516 signal classification; string detail can't trip the #678 dict-body auto-retry (pinned). Named regression suite proven red-on-revert (9/11), #160 placeholder preserved for genuine clean exits, live A/B verified. All checks green.

Issue #1671, opened 2026-07-17
bug: agent rename is not gated on volume-base ownership — the #1664 one-row-per-base invariant is enforced at create only


obasilakis:
Found during the /review of PR #1666 (#1664/#1665/#1667). That PR establishes the invariant that **one ownership row claims one volume base** and enforces it at agent creation. Rename — the only other producer of `volume_base_name` — is not gated, so it can still mint the collision.

## Summary

PR #1666 added `agent_ownership.volume_base_name` and gated creation on it:

```python
# services/agent_service/crud.py:298
if not config.ephemeral and db.is_volume_base_reserved(config.name):
    raise HTTPException(status_code=409, detail="Agent name unavailable: its data volumes still belong to another agent (it was renamed).")
```

`db.is_volume_base_reserved` has exactly two enforcement call-sites — create (above) and the purge/orphan sweep (`cleanup_service.py:855,965`). The rename endpoint checks neither the volume base nor anything that stands in for it:

```python
# routers/agent_rename.py:95 — Docker container only
existing = get_agent_container(sanitized_name)

# db/agent_settings/metadata.py:137 — agent_name column only
taken = conn.execute(select(1).where(agent_ownership.c.agent_name == new_name)).first()
```

Neither sees `volume_base_name`, so renaming **into** a base another row already claims is permitted.

## Component

Backend / Agent Rename (`routers/agent_rename.py`, `db/agent_settings/metadata.py::rename_agent`) × the #1664 volume-identity model

## Priority

P1 — cross-tenant data exposure of the same class as #1667, but requires a two-rename sequence and is not a regression (pre-#1666 this path was equally ungated, with worse consequences).

## Reproduction Steps

An ordinary ops swap is enough:

1. Create agent `prod-agent`; enable file sharing (creates `agent-prod-agent-public`). Volumes live under base `prod-agent`.
2. Rename `prod-agent` → `prod-agent-old`. Row: `agent_name='prod-agent-old'`, `volume_base_name='prod-agent'`. Volumes keep the old base — by design (#1664).
3. Create a *new* agent named `prod-agent` → **409, correctly refused** by the #1664 create gate.
4. Instead, create agent `staging-agent`, then rename `staging-agent` → `prod-agent`:
   - `get_agent_container('prod-agent')` → None (the container is `agent-prod-agent-old` now) → passes.
   - `select(1).where(agent_name == 'prod-agent')` → no row (it's `prod-agent-old`) → passes.
[...]

Issue #1669, opened 2026-07-17
bug: rename + recreate silently empties an agent's public/shared volumes (they name off the live agent name)


dolho:
Found during the #1665 audit (PR #1668) — the fourth site of the rename/volume-identity class, and deliberately left out of that PR because it is not a mechanical swap.

## Summary

`get_public_volume_name` / `get_shared_volume_name` (`db/agent_settings/file_sharing.py`, `db/shared_folders.py`) name off the agent's **live** name:

```python
return f"agent-{agent_name}-public"    # and -shared
```

Rename keeps the agent's existing volumes (Docker renames neither a volume nor its immutable label), and `recreate_container_with_updated_config` deliberately **drops** the old public/shared mounts and re-adds them from current config:

```python
if dest == db.get_public_mount_path(): continue          # skipped...
...
public_volume_name = db.get_public_volume_name(agent_name)   # ...re-added by CURRENT name
try:    await volume_get(public_volume_name)
except docker.errors.NotFound:  await volume_create(name=public_volume_name, ...)   # fresh + EMPTY
```

So for a renamed agent: **rename + any recreate (guardrails / resources / shared-folder change, operator rebuild) silently empties its publish dir and its shared-out folder.** The data is orphaned, not deleted — #1664's union ownership predicate keeps the reclaim sweep off the old base — but it is unreferenced, and consumers mounting `agent-{source}-shared` (`shared-in/{agent}`) see an empty folder from that point on.

Confirmed against a real DB:

```
agent renamed p1665-old -> p1665-new; workspace pin = 'p1665-old'
  public volume the recreate will mount : agent-p1665-new-public
  shared volume the recreate will mount : agent-p1665-new-shared
  ...but its published/shared data is in: agent-p1665-old-public / agent-p1665-old-shared
```

Severity is below the workspace case (#1665): the FILES-001 download path serves bytes copied to `/data/agent-files`, so already-shared links keep working — what breaks is the agent's staging dir contents and cross-agent shared folders.

## Why this is not a one-liner

The obvious fix — resolve through `volume_base_name` like #1665 did for the workspace — is **wrong on its own for agents already renamed in the field**: their post-rename recreate has already created `agent-{new}-public` and they have been writing to it, possibly for weeks. Switching the lookup to the pinned base would mount `agent-{old}-public` and orphan the *newer* data instead — the same bug pointed the other way.

So this needs a decision:

[...]

Issue #1667, opened 2026-07-17
bug: agent creation silently reuses a pre-existing Docker volume of the same name (get-then-create, no adopt/refuse decision)


dolho:
Found during the /review of PR #1666 (#1664). That PR closes the rename-driven half; this is the remainder.

## Summary

`crud.py` provisions the agent home volume with **get-then-create** semantics:

```python
agent_volume_name = f"agent-{config.name}-workspace"
try:
    await volume_get(agent_volume_name)      # exists -> silently REUSED
except docker.errors.NotFound:
    await volume_create(name=agent_volume_name, labels={...})
volumes[agent_volume_name] = {"bind": "/home/developer", "mode": "rw"}
```

An existing volume is adopted **silently** — there is no decision point, no log line, and the labels of the pre-existing volume are never checked against the agent being created. Whatever data is in it becomes the new agent's `/home/developer`, including any `.env` / `.credentials.enc` left there.

PR #1666 gates the **rename** case (`db.is_volume_base_reserved(name)` → 409: a live or soft-deleted row still claims that base). That gate cannot cover a volume with **no owning row**:

- the retention purge's `remove_agent_volumes` hit an in-use 409 and skipped, and the orphan sweep has not reclaimed it yet (it now requires 3 consecutive unattached cycles + a 1h creation grace, so this window is ~15+ min wide, by design);
- the backend crashed between `volume_create` and the `agent_ownership` INSERT (the 1h creation-grace exists precisely because creation writes the volume BEFORE the row);
- a volume restored from a backup, or created by hand.

In all of these `is_volume_base_reserved` correctly answers "free" (no row claims it) and the new agent adopts a stranger's disk.

## Impact

Data from a previous agent of the same name resurfaces inside a new agent — potentially a different owner's. Not a fresh workspace, and never announced. Lower severity than the rename case #1666 fixes (that one reused a **live** agent's volume, and both containers wrote it concurrently); here the predecessor is gone, so it is disclosure of residue rather than concurrent corruption.

## Fix — needs a decision, not a one-liner

The create path should make adoption explicit. Options, roughly in order of preference:

1. **Refuse + require an explicit opt-in.** If `agent-{name}-workspace` exists at create time, 409 with a distinct code (`volume_exists`), and let the caller either pick another name or pass an explicit `adopt_existing_volume: true`. Safest; the only cost is a new failure mode for a name whose volume is mid-reclaim.
[...]

PR #1676, opened 2026-07-17
feat(agents): editable display label with an immutable slug (ent#181)

dolho:
Implements trinity-enterprise#181. **OSS-core** (maintainer decision — an agent's display name is basic UX, not a paid capability).

## The problem

An agent has one name: its slug. So "rename" is the heavyweight identity change — `PUT /rename` stops the container, rewrites `agent_name` across ~20 tables, clears every per-agent Redis keyspace, renames avatar files, and **still leaves the agent's volumes under the old base**, because Docker can rename neither a volume nor its immutable `trinity.agent-name` label. That single operation is the root of #1664 (P0) / #1665 / #1667 / #1669 / #1671.

But "call it Marketing Bot" is the common case, and it should touch none of that.

## The fix: render a label, never resolve it

A per-agent `display_label` that is presentation only. The slug stays the identity every route, container, volume, MCP key and A2A card keys on.

**Verified end-to-end on the live stack** — this is the property the whole feature turns on:

```
PUT /api/agents/a2a-demo/label {"label":"Marketing Bot"}
  -> {"label":"Marketing Bot","display_name":"Marketing Bot"}
GET /api/agents          -> a2a-demo carries display_label="Marketing Bot"
container agent-a2a-demo  -> still exists (nothing restarted)
GET /api/agents/a2a-demo -> 200 (the slug still routes)
PUT ... {"label":"   "}  -> label:null, display_name:"a2a-demo"  (blank reverts to the slug)
```

## Shape

- **`agent_ownership.display_label TEXT`**, nullable. NULL = "render the slug", so no backfill and every existing agent is unchanged until someone sets one. Dual-track migration per Invariant #3 (SQLite + Alembic `0025`, schema.py, tables.py).
- **`DisplayLabelMixin`** (Invariant #2), with a **batched** reader for the fleet list so the hottest endpoint stays one query (mirrors `get_tags_for_agents`), and the four `DatabaseManager` facade pass-throughs — the #1666 lesson that mocked tests can't see a facade gap.
- **`GET`/`PUT /api/agents/{name}/label`**, owner-only. The read never coerces `label` to the slug, so the UI can tell "no label" from "label equals slug" and show an empty field. WS `agent_label_changed` broadcast.
- **Frontend — one resolution helper** (`utils/agentName.js`), not per-site `label || name` (which would show one agent under two names — §1.3.1 FR-3). The header pencil edits the label and keeps the slug visible as secondary text (FR-4); list/tile surfaces render the label with the slug in the tooltip. The store goes through the shared axios client (Invariant #7).
- **The slug rename is demoted, not removed** (FR-5): a separate "Rename the id instead…" affordance with copy that states what it actually does. Owners who need it keep it; it stops being the default gesture.

Requirements §1.3.1 landed before the code. 28 tests.

## Notes for review

- **Two branch-collision hazards this shares with the merged #1666 and the in-flight A2A #1628, worth watching at merge time:**
  1. **Alembic head.** This chains `0025` onto dev's current head `0024_volume_base_name`. The A2A branch (#1628) *also* adds a `0025`. Whichever merges second must re-chain to `0026` — same fix I just did on #1628 when #1666 landed under it. (SQLite's track is a list and unions cleanly; only Postgres branches, which is why `pg-migrations` is the job that catches it.)
  2. **Local DB stamp.** Switching between these branches re-stamps the shared local Postgres, so a checkout can leave the backend crash-looping on "Can't locate revision" until you re-stamp to the common ancestor and restart. Not a code issue — a dev-ergonomics one — but I hit it twice today, so flagging it.
- `npm run build` fails locally on an unrelated pre-existing issue (`mermaid` declared in package.json, not installed, `AgentWorkspace.vue`) — reproduces with this branch stashed. My four touched SFCs compile cleanly via `@vue/compiler-sfc`.


vybe (approved):
Validated via /validate-pr. Requirements §1.3.1 landed before code; dual-track migration complete (SQLite agent_ownership_display_label + Alembic 0025 chained on 0024, schema.py + tables.py); label is rendered-never-resolved with one frontend resolution helper; owner-gated PUT with soft-delete guard; batched fleet read; slug rename demoted with honest copy. Real-DB harness tests pin the invariants (slug never moves, NULL=slug, no name reservation).

Minor follow-ups (non-blocking): (1) tests/registry.json carries a duplicate entry for test_181_agent_display_label.py — dedupe next touch; (2) GET/PUT /api/agents/{name}/label isn't in architecture.md's endpoint table yet; (3) cross-tracker ref — set status-in-dev on trinity-enterprise#181 manually after merge. All checks green.

PR #1679, opened 2026-07-17
fix(ssh): enforce ephemeral-key TTL on the filesystem + injection-safe inject/remove (#1616)

### LINKED ISSUE #1616: bug: injected SSH public keys are silently wiped whenever an agent container is recreated
## Summary

SSH keys injected via `POST /api/agents/{agent_name}/ssh-access` are appended to `/home/developer/.ssh/authorized_keys` on the container's **overlay filesystem**. Any operation that recreates the agent container (rebuild, recreate-on-start, lifecycle recovery) silently deletes every injected key while its Redis TTL metadata says it is still valid. The user experience is: SSH worked yesterday, today the same key gets `Permission denied (publickey)` with no error anywhere — this cost a client and two ops teams a multi-hour misdiagnosis (the client concluded the backend had lost its Docker socket).

## Component

Backend / SSH Service + Agent Runtime (container lifecycle)

## Priority

P2 (workaround exists — re-mint the key after any recreate — but the failure is silent and very misleading)

## Error

```
ssh -p <agent_port> developer@<host>
developer@<host>: Permission denied (publickey).
```
No backend error; key-injection metadata in Redis (`ssh_access:<agent>:<key_comment>`, TTL still live) says access is valid.

## Location

- **File**: `src/backend/services/ssh_service.py`
- **Line**: ~73–89 (append to `/home/developer/.ssh/authorized_keys` via `container_exec_run`)

## Root Cause

`authorized_keys` lives only on the container overlay layer. `docker restart` preserves it, but any **recreate** resets the filesystem to the image state, so all injected keys vanish. The key metadata (Redis, TTL-based) is not consulted on container start, so nothing re-injects still-valid keys and nothing invalidates the metadata either — state silently diverges.

Verified on a live instance: an agent recreated at T had an empty `.ssh/` afterwards; keys injected pre-recreate were gone while their TTL metadata remained; a freshly minted key on the new container worked immediately (platform injection path itself is healthy).

## Reproduction Steps

1. Generate a keypair locally and `POST /api/agents/{agent}/ssh-access` with `auth_method: "key"` and the public key, e.g. TTL 12h
2. Verify SSH login works
3. Recreate the agent container (e.g. the platform's rebuild/recreate path, or stop + remove + start via the API)
4. SSH with the same key → `Permission denied (publickey)`, though the TTL has hours remaining

## Suggested Fix

On agent container start (backend's lifecycle/ready path), re-inject all unexpired public keys for that agent from the stored metadata. This requires storing the public key material alongside the existing Redis metadata (only public keys — the server never holds private keys, so persisting them is safe). Alternative: relocate `.ssh` onto the agent's persistent volume (e.g. symlink into the workspace volume), though re-injection from metadata keeps TTL enforcement in one place.

Either way, expired-key cleanup should keep working against the chosen store.

## Environment

- Trinity version: `a88f5fc8`
- Docker prod compose, Ubuntu (GCP VM)

## Related

- `src/backend/services/ssh_service.py` (inject / remove key paths)
- Companion issue: #1615 (password ssh-access broken end-to-end)

AndriiPasternak31:
Fixes #1616

## Problem

Two independent reliability defects in ephemeral SSH access:

1. **Expired keys were never removed from the filesystem.** An ephemeral key's TTL was enforced *only* on its Redis metadata. `SshService.cleanup_expired_credentials()` — the code that removes the key's line from the `authorized_keys` file `sshd` actually reads — existed but had **zero callers** (`SSH_ACCESS_CLEANUP_INTERVAL` was defined and never read). On a preserved (never-recreated) volume, an **expired** key lingered in the file and still granted login past its stated expiry.
2. **Inject/remove were injection-prone and non-idempotent.** `inject_ssh_key` interpolated the public key into an `sh -c` string that docker-py `shlex.split`s *and* the container re-parses (double-unquoting) — a quote raised `ValueError`, a `$(...)`/backtick opened command substitution — and a repeat inject duplicated the line. `remove_ssh_key` used a substring `sed -i '/comment/d'` that could over-delete an unrelated key.

## Fix

- **Wire the expired-key sweep** into the 5-min `cleanup_service` loop (`_sweep_expired_ssh_credentials`, report field `ssh_credentials_expired`; fail-open, self-contained).
- **`expires_at`-driven, grace-backed sweep.** Deciding expiry from a `ttl in [0,60]s` window against the 300s cadence caught only ~20% of keys (the auto-expiring Redis row vanished between ticks). Now the sweep compares `now >= expires_at`, and `store_credential_metadata` keeps the Redis row alive `SSH_ACCESS_CLEANUP_GRACE_SECONDS` (600s > cadence) past true expiry so every expired key is observed by ≥1 cycle. The Redis row is deleted **only after** the file line is gone, so a transient failure retries. Legacy rows with no `expires_at` fall back to the old window.
- **`SCAN`, not `KEYS`.** A live probe caught that the sweep scanned with `redis_client.keys()`, which the backend's `-@dangerous` Redis ACL user is **forbidden** to run (`NoPermissionError`) — the fail-open handler would have swallowed it to 0-cleaned every cycle, leaving the fix inert. Switched all three `ssh_service` scans (`cleanup_expired_credentials`, `list_active_keys`, `cleanup_agent_credentials`) to `scan_iter`. A mocked test can't see this, so a durable static guard asserts `redis_client.keys(` never reappears in `ssh_service.py`.
- **Injection-safe inject/remove.** The key/comment flow via the exec `environment` (never string-interpolated) in a single atomic `sh -c`. `inject_ssh_key` gains `skip_if_present=True` (append-if-absent via `grep -qxF`). `remove_ssh_key` matches the comment as an exact awk `$NF` last-field via `ENVIRON` (no regex/metachar escaping, no substring over-delete) and swaps a temp file in only on success (never truncates `authorized_keys`).

## Scope note

#1616's reported **recreate-wipe** symptom is a *separate, unidentified* mechanism — a normal never-renamed agent is volume-safe on recreate (`/home/developer` is a named volume forwarded by the config-recreate path), and the #1664/#1665/#1667 volume-identity family (fixed by #1666) postdates the reporter's build. This PR fixes the **enforceable half** (expired-key file-side revocation + safe inject/remove); no start-time re-injection hook was added (it is near-inert and would mask real data loss).

## Tests

`tests/unit/test_1616_ssh.py` (new, 29 tests) + updated `test_ssh_service.py`:
- **Real-shell** (not mocked) inject/remove: captures the exact `(cmd, environment)` the service generates and runs it through a real `sh` against a temp HOME — append-if-absent idempotency, 600/700 perms, injection-proofness (`$(...)`, backtick, quotes, `;`, `|`, `&&`, backslash, `$HOME`), exact-last-field removal that won't over-delete a superset comment.
- **Sweep**: cross-cycle-miss (expired but outside the old 60s window → still removed), legacy-row TTL fallback, delete-only-after-removal retry-safety, metadata-TTL-carries-grace, fail-open on Redis error, `cleanup_service` wiring + a static guard the sweep is actually invoked (dead-code class).

Local: 58 passed (SSH + cleanup suites). Docs: `architecture.md` (Background Services), `feature-flows/ssh-access.md`, `feature-flows/cleanup-service.md`, feature-flows index.

_Draft — I'll flip to ready for review._


vybe (approved):
Validated via /validate-pr. Wires the zero-caller cleanup_expired_credentials into the 5-min sweep (the only thing that actually revokes a key-auth credential on a preserved volume); expires_at-driven with a Redis-TTL grace sized to the cadence; SCAN not KEYS (real ACL constraint caught live, pinned by a static guard); inject/remove pass caller data via exec environment only (no shell interpolation), awk exact last-field match + temp-file swap so removal can neither over-delete nor truncate. Real-shell tests. New SSH_ACCESS_CLEANUP_GRACE_SECONDS follows the (pre-existing) unwired SSH_ACCESS_* family precedent with a safe default — flagging the whole family for /validate-config as a follow-up, not blocking. Dev merge (docs index + learnings) resolved keeping both sides; 89 tests green on the merged tree. All checks green.

PR #1681, opened 2026-07-17
Scrub terminal backlog_metadata PII in the retention sweep (#1449)

### LINKED ISSUE #1449: reliability: callback-path chat-session persistence (idempotent) + backlog_metadata PII scrub — deferred from #1444
Deferred follow-up from #1444 (async `/task` + `save_to_session` persistence hardening).

## Background

#1444 hardened the **in-process** chat-session persistence path (`routers/chat.py::_persist_chat_session`, driven by the async wrapper `_run_async_task_with_persistence`): fail-loud on write errors, owner-gated `chat_session_id`, SUCCESS-guarded, with a fast unit regression guard.

Triage for #1444 established that the **#1083 fire-and-forget callback path does NOT persist chat sessions**, but it is **structurally unreachable by a manual `/task` today**: `ASYNC_DISPATCH_ELIGIBLE_TRIGGERS` is a hardcoded `frozenset({"schedule","webhook"})` and no `save_to_session` turn carries an eligible trigger. So callback-path persistence is **currently dead code** — correctly descoped from #1444 (see the /autoplan Decision Audit there).

Under the **pull / work-stealing** direction (Epic #1045, umbrella #1081), *every* terminal will arrive via a result-report path and `apply_result` becomes the single finalizer. At that point the callback path becomes live for `save_to_session` turns, and the missing persistence step would silently lose Chat-tab history again.

## Scope (do this when pull-mode makes the path reachable — not before)

1. **Callback-path chat-session persistence, built idempotently.**
   - Persist `save_to_session` / `user_message` / `create_new_session` / `user_id` / `user_email` as **durable execution metadata** (the queued path already does this via `backlog_metadata` — precedent) so BOTH finalizers (inline sync + result callback) can run persistence.
   - Extract a `chat_session_persistence` **service** (Invariant #1 — service layer, not a router→router import) so `apply_result` / the callback endpoint can call it without importing `routers/chat.py`.
   - **Key on `execution_id`** and make it **replay-safe**: a re-delivered/crash-recovered terminal must **upsert**, not double-write messages or drop the session. `chat_messages` currently has **no `execution_id` column** — add one via a **dual-track migration** (Invariant #9: SQLite `db/migrations.py` + an Alembic revision + `db/schema.py`/`tables.py`), then dedup the user+assistant insert on `(session_id, execution_id, role)` (or equivalent). A terminal-CAS-only guard is NOT sufficient — it reintroduces the #1444 silent-loss bug in the crash window.
   - Target-arch note: this is a reversible/idempotent internal DB effect — no `effect_guard` key needed, but it MUST be replay-safe because the callback can legally be re-delivered (#1085 re-delivery governor).

2. **Queued-path `backlog_metadata` PII scrub (pre-existing, security).**
   - `services/backlog_service.py` captures the full `ParallelTaskRequest` (incl. `message` / `user_message`) into `schedule_executions.backlog_metadata` for drain replay. On a **terminal** row this is stale PII sitting in the DB indefinitely.
   - Extend the #772 retention sweep (`cleanup_service`) to **NULL `backlog_metadata` on terminal rows** (it's only needed while `status='queued'`).

## Acceptance criteria

- A `save_to_session` turn finalized via the result-callback path creates a `chat_sessions` row + user & assistant `chat_messages` (integration-testable once pull-mode is on).
- A **re-delivered / duplicate** terminal for the same `execution_id` does NOT create duplicate messages or a second session (idempotent upsert).
- `backlog_metadata` is NULL on terminal `schedule_executions` rows after a retention sweep.
- Dual-track migration for the new `chat_messages.execution_id` column (SQLite + Alembic), schema-parity CI green.

## Not in scope

- The in-process path (done in #1444).
- Turning pull-mode default-ON for side-effect agents (gated separately on #1084 / Epic #1045/#1081).

Refs: #1444, #1083, #1082, #1085, Epic #1045, umbrella #1081.

AndriiPasternak31:
Fixes #1449

## What

Scrubs stale drain-replay **PII** from `schedule_executions.backlog_metadata` on
authoritative-terminal rows. `backlog_service.enqueue` `json.dumps`es the full
request — `user_message` / `user_email` / `system_prompt` — into that column so a
**queued** task can be reconstructed at drain. Nothing reads it once a row leaves
`status='queued'` (the drain claims only queued rows; the #1083/#1081 result
callbacks read the POST payload, not the row's metadata; canary E-04/G-04 are
queued-scoped), so on a terminal row it is stale PII sitting in the DB indefinitely
— bounded only by the 90-day `execution_row_retention_days` DELETE. This NULLs it as
soon as the row reaches a terminal state.

## How

- **`db.scrub_terminal_backlog_metadata(chunk_size)`** (`db/schedules.py`) — chunked
  `SELECT id … LIMIT N` → `UPDATE … SET backlog_metadata=NULL WHERE id IN (…)`, each
  chunk its own transaction (short write lock), mirroring `prune_execution_logs`.
  Facade delegate on `DatabaseManager`.
- **`cleanup_service._sweep_retention_772`** — appended as an **unconditional**
  sub-sweep (not age-gated, no ops-config key). Count feeds `CleanupReport` + the
  WAL-checkpoint sum; the blob itself is **never** logged (count-only — it carries PII).

### Two load-bearing decisions
- **FAILED is EXCLUDED** (`_AUTHORITATIVE_TERMINALS = success/cancelled/skipped`). A
  FAILED row is resurrectable to SUCCESS via a late token-gated CAS (`park_expired_lease`
  keeps its `claim_token`), so its drain-replay intent must survive. FAILED PII stays
  bounded by the 90-day row DELETE. (Regression-guarded by a dedicated test.)
- **Not an operator knob** — the scrub is a **security invariant**, so it has no
  retention window and runs every cycle even when all #772 windows are `0`. A fixed
  default sidesteps the #1638 floor-by-seed trap.

## Migrations / schema

**None.** This scrubs *values* in an existing column — no DDL, no new table, no
column. So there is **no dual-track (SQLite + Alembic) migration to add**, and the
`schema-parity` CI job is trivially green (no schema file touched).

## Tests

- `tests/unit/test_1449_backlog_metadata_scrub.py` (real DB via `db_harness`, not a
  mock): authoritative-terminal NULL-out, **FAILED-exclusion resurrection guard**,
  queued/running/pending_retry untouched, multi-pass chunking, idempotency,
  `chunk_size<=0` disabled, canary queued-scope smoke.
- `tests/unit/test_cleanup_inner_sweeps.py`: report field + WAL sum + "runs even when
  retention disabled" guards.
- `tests/unit/test_schedule_status_observability.py`: registers the new writer in the
  #1082 status-writer guard.

Local run: **35 passed / 1 skipped** (skip = Postgres dual-backend, no PG reachable);
canary + retention-prune regression sweep **142 passed**; `sys_modules` lint green.

## Docs

`requirements/infrastructure.md` §12.11 (RETENTION-002), `architecture.md` Cleanup
Service sweep line, `feature-flows.md` Recent-Updates row, and
`feature-flows/cleanup-service.md` all updated.

## Scope note (reviewer)

#1449 bundles two deliverables. This PR ships **only** the `backlog_metadata` PII
scrub (a live security win with zero pull dependency). The sibling — **callback/pull-path
chat-session persistence** — is **deferred to the #1081 single-applier work** (dead
code today: the callback path is unreachable by a manual `/task`, pull-mode is OFF),
with the verified design captured in the plan. The original bundled "reconciliation
sweep" design was **killed** in review — it would double-write live in-process-drained
chat rows. If you'd rather keep #1449 open for the deferred half, drop the `Fixes` to
`Refs`.


vybe (approved):
Validated via /validate-pr. Clean security-invariant sweep: chunked SELECT-ids-then-UPDATE mirroring prune_execution_logs (short write locks), scoped to authoritative terminals with FAILED deliberately excluded (resurrectable via late token-gated CAS — regression-guarded), unconditional (no ops key — dodges the #1638 floor-by-seed trap), count-only logging. No schema change so no dual-track migration needed (verified). Real-DB harness tests + cleanup-cycle characterization + #1082 status-writer registration; requirements §12.11 + architecture + feature-flows all updated. Dev merge (index rows) resolved; 104 tests green on the merged tree incl. the three-way cleanup_service merge with #1616/#1578. All checks green.

PR #1685, opened 2026-07-17
fix(chat): Continue-as-Chat resume — close the IDOR and the KeepAlive drop (#1672)

### LINKED ISSUE #1672: bug: 'Continue as Chat' (EXEC-023) broken or unreliable across use cases — test and validate the full matrix
## Summary

The **Continue as Chat** feature (EXEC-023) — the button on the Execution Detail page that reopens a completed execution as an interactive chat via `?tab=chat&resumeSessionId=<claude_session_id>` — is reported not working, or at least not working in all use cases. The full path (button → route → forced-legacy chat mode → `resume_session_id` payload → agent-server `--resume`) needs systematic testing, and the broken cases fixed.

## Context

Reported by the operator: continuing an execution in chat fails in at least some scenarios. The flow crosses several recent changes that make silent breakage plausible:

- The Unified Chat tab (#1112) made **Session mode the default**; Continue-as-Chat must force the legacy `ChatPanel` via a transient `routeForcedMode` without rewriting the saved preference — a routing seam that can regress.
- Fire-and-forget dispatch (#1083) writes the sentinel `claude_session_id='dispatched_async'` on async executions — passing that as a resume id would produce a broken resume.
- Some executions legitimately have **no** `claude_session_id` (NULL), and headless-task JSONL persistence is only auto-enabled for timeouts > 600s; the session-cleanup reaper also deletes JSONLs outside the keep set — so a resume target may simply no longer exist on the agent.
- Non-Claude runtimes (e.g. Codex) lack `--resume` continuity entirely (#1187).

## Acceptance Criteria

- [ ] A documented test matrix covers Continue-as-Chat from each execution source: manual chat/task, schedule, webhook, loop iteration, MCP-triggered, and async-dispatched (#1083) executions
- [ ] Executions with no usable session (NULL `claude_session_id`, `dispatched_async` sentinel, reaped/expired JSONL) either hide the button or fail gracefully with a clear message — never a silent no-op or broken chat
- [ ] Clicking Continue as Chat lands on the Chat tab in legacy mode with the resume banner shown, and the next message actually resumes the execution's Claude session (verified end-to-end, with evidence)
- [ ] The forced-legacy mode does not overwrite the user's saved chat-mode preference (`localStorage['trinity.chatMode']`)
- [ ] Button is hidden or clearly unsupported for runtimes without resume support (Codex)
- [ ] Broken cases found by the matrix are fixed or split into follow-up issues, each traced back to this one

## Technical Notes

Key files in the path:
- `src/frontend/src/views/ExecutionDetail.vue` — Continue as Chat button (EXEC-023), passes `resumeSessionId: execution.claude_session_id`
- `src/frontend/src/views/AgentDetail.vue` — `resumeSessionId` from route query; `routeForcedMode` forces legacy chat mode
- `src/frontend/src/components/ChatPanel.vue` — resume banner, sends `resume_session_id` in the task payload
- `src/backend/routers/chat.py`, `src/backend/services/task_execution_service.py` — backend pass-through
- `docker/base-image/agent_server/routers/chat.py` + runtime adapters — actual `--resume` invocation
- `docker/base-image/agent_server/services/jsonl_recovery.py` / `src/backend/services/session_cleanup_service.py` — JSONL persistence + reaping rules that determine whether a resume target still exists

obasilakis:
## Summary

Fixes **"Continue as Chat" (EXEC-023)**, which was broken/unreliable across cases **and** leaked private conversations across users on shared agents.

- **Reliability root cause:** `AgentDetail` is `<KeepAlive>`-cached, so the code forcing the legacy `ChatPanel` for a resume landing ran only in `onMounted` — which does **not** fire on the common nav path (open agent → click execution → *Continue as Chat* re-activates the cached instance via `onActivated`). `SessionPanel` rendered instead of the resuming `ChatPanel` and the resume id was silently dropped; it "worked" only after a hard page reload — the exact "works sometimes" symptom.
- **Security root cause (IDOR):** `resume_session_id` → `claude --resume <id>` was passed **unvalidated and unauthorized** through `POST /api/agents/{name}/task`. Execution rows are agent-scoped and expose `claude_session_id`, but that id is a per-user secret (the Session tab 404s on it). Any operator on a **shared** agent could read a peer's session id from the executions list and resume the peer's private conversation.

## Changes

**Backend**
- `routers/chat.py` — resume gate on `/task`: reject the #1083 `dispatched`/`dispatched_async` sentinels (400) + authorize ownership via `db.resume_session_belongs_to_user` (404, enumeration-safe; admin bypass). Keyed on the resume id's presence, **not** the spoofable `X-Source-Agent` header. No id-shape check — ownership is the guard, which also preserves **Codex** resume (its `thread_id` isn't a UUID).
- `db/schedules.py` + `database.py` — new `resume_session_belongs_to_user` primitive + facade delegation.
- `routers/paid.py` — sentinel reject only (no Trinity `current_user` to bind ownership to; see follow-up).

**Frontend**
- `views/AgentDetail.vue` — extracted `applyDeepLinkRouting()` called from **both** `onMounted` and `onActivated`; clears `routeForcedMode` on non-resume landings so legacy mode can't stick for the SPA session.
- `views/ExecutionDetail.vue` — `canContinueAsChat` hides/rejects the sentinel button case.
- `components/ChatPanel.vue` — auto-select guard `!isResumeMode` → `!resumeSessionIdLocal`, so **dismissing the banner** no longer lets auto-select null the resume id and drop `--resume`.

**Docs** — `feature-flows/continue-execution-as-chat.md` (Security Considerations rewritten + revision entry), `learnings.md` (IDOR + KeepAlive bug classes).

## Test Plan

- [x] `pytest tests/unit/test_1672_continue_as_chat_resume_auth.py` — **15 passed**. Real accessor SQL (owned / foreign-user / wrong-agent / unknown / sentinel-agnostic) + endpoint gate (sentinel→400, foreign→404, owned/admin pass, **regular-user X-Source-Agent spoof still 404**, agent-key foreign→404).
- [x] Regression: `test_chat_admission`, `test_chat_dispatched_marker`, `test_946_task_idempotency_on_deny`, `test_1457_selftask_inject_ownership`, `test_async_task_persistence` — **35 passed**.
- [ ] **e2e `continue-as-chat.spec.js`** — reproduces the KeepAlive path + asserts the saved chat-mode pref is not overwritten. `@interactive`, runs in CI on `ui`-labeled PRs. **Authored without a live stack and NOT executed locally** — validate in CI.
- [ ] Manual matrix (needs a stack): resume from schedule / webhook / loop / MCP / async-dispatch executions.

## Notable review finding

The first version of the backend gate (`if resume_session_id and not x_source_agent`) was **still exploitable** — an adversarial-review subagent found a regular user (`agent_name=None`, so SELF-EXEC-001 never checks them) could set `X-Source-Agent` and skip the gate, leaving the IDOR fully open. The gate now ignores the header; a regression test pins the spoof case.

## Follow-ups (not in this PR)

- No index on `schedule_executions.claude_session_id` → the auth lookup full-scans (rare action, LOW). Adding it is a dual SQLite+Alembic migration.
- Paid `/api/paid/{name}/chat` resume has only the sentinel guard — full payer→session authorization needs a payer-identity binding that doesn't exist yet.

Fixes #1672


vybe (approved):
Validated via /validate-pr. IDOR closed correctly: ownership gate keys on the authenticated principal (never the spoofable X-Source-Agent header — the adversarial-review bypass is pinned by test), 404 mirrors the Session tab's enumeration-safe response, sentinel reject covers the #1083 rows, and ownership-not-shape preserves Codex resume. KeepAlive fix extracts one idempotent deep-link handler called from both onMounted and onActivated. Real-SQL accessor tests + endpoint gate tests (15) incl. the spoof attack; e2e spec noted as CI-validated only. All checks green.

PR #1688, opened 2026-07-18
fix(operator-queue): normalize reserved-prefix guard against lookalikes (#1631 follow-up)

obasilakis:
## Summary

Fast-follow to #1684 (merged). The `/cso --diff` audit of the #1631 change found one **LOW**: the reserved-prefix hijack guard (`operator_queue_service.py`) was exact-match (`req_id.startswith(_RESERVED_ID_PREFIXES)`), so a prompt-injected agent could author an item with a **lookalike** id (`Poison-x`, uppercase/leading-space) that evaded the filter and appeared as a **platform-styled** operator alert with attacker text.

- **Not a hijack.** An independent refutation proved collision/suppression is impossible — the only id that can collide with the platform's real `poison-{execution_id}` slot *necessarily* starts with `poison-` and is rejected. Every evasion produces a *different*, non-colliding row. So this is operator-**phishing polish**, not a new capability (an agent can already author arbitrary queue items).
- **Fix:** normalize (`strip().lower()`) before the prefix check. The platform mints these prefixes lowercase and unpadded, so a normalized match can only be impersonation.

This hardening was committed during the audit but **stranded** when #1684 squash-merged mid-`/cso`; landing it on its own branch.

## Changes

- `src/backend/services/operator_queue_service.py` — `strip().lower().startswith(...)` in the reserved-prefix guard.
- `tests/unit/test_1631_operator_queue_agent_scoped_ids.py` — new `test_reserved_prefix_guard_folds_case_and_whitespace` (rejects ` Poison-x`, `SYNC-FAILING-x`, `\tgit-bloat-x`).
- `docs/security-reports/cso-2026-07-18.md` — the audit report.

## Test Plan

- [x] `tests/unit/test_1631_operator_queue_agent_scoped_ids.py` → `6 passed` (5 original + fold-case).
- [x] No behavior change to the legitimate path — agent-authored non-reserved ids are unaffected.

Refs #1631. Hardening follow-up to #1684.

### DIFF
--- a/docs/security-reports/cso-2026-07-18.md
+++ b/docs/security-reports/cso-2026-07-18.md
@@ -1,0 +1,63 @@
+# CSO Audit — 2026-07-18
+
+- **Mode**: daily (`--diff`), 8/10 confidence gate
+- **Scope**: branch `feature/1631-operator-queue-id-collision` vs `origin/dev` (issue #1631)
+- **Phases run**: 0, 1, 2, 4, 7, 9 (scoped to diff), 12, 13, 14
+
+## Summary
+
+The #1631 change scopes operator-queue request-id uniqueness per agent (surrogate
+uuid PK + agent-authored `request_id` column with `UNIQUE(agent_name, request_id)`).
+Audit found **no CRITICAL/HIGH/MEDIUM findings**. One LOW (guard hardening) was
+surfaced and fixed in-branch. The change is a net security **improvement** — it
+closes a cross-agent write (agent B acknowledging its id flipped agent A's row).
+
+## Findings
+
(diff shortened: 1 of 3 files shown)


vybe (approved):
Validated via /validate-pr: focused hardening follow-up to #1684 — normalize-before-check on the reserved-prefix guard, with the named lookalike regression test (`test_reserved_prefix_guard_folds_case_and_whitespace`). Ran `tests/unit/test_1631_operator_queue_agent_scoped_ids.py` locally on the PR head merged with dev: 6 passed. No security/packaging concerns (diff strictly tightens an existing guard; no new env vars, modules, or endpoints). Note: body uses bare `Refs #1631`, so bumping the issue to status-in-dev manually.

PR #1690, opened 2026-07-18
fix(canary): read the configured SQL backend so the invariant harness is not blind on PostgreSQL (#1540)

### LINKED ISSUE #1540: Canary invariant harness is blind on PostgreSQL — reads stale SQLite /data/trinity.db
> Greenlit 2026-07-13 (incubation promotion sweep) — `status-ready`, P2.

## Summary
The canary invariant harness (CANARY-001) opens a **SQLite** connection unconditionally, so on a **PostgreSQL**-backed deployment it inspects a stale/empty `/data/trinity.db` instead of the live PG database. The SQL-tier invariant checks therefore run against zero live rows and return **vacuously green** — the harness is effectively blind to the live database at that tier.

## Evidence
- `src/backend/db/connection.py::get_db_connection` hardcodes `sqlite3` and ignores a PostgreSQL `DATABASE_URL`.
- On a PG instance, `POST /api/canary/run-cycle` returned **0 violations**, but its SQL collector was reading a frozen, read-only `/data/trinity.db` with **no running rows**.
- SQL-sourced checks affected (blind on PG): **S-01, E-05, E-01, E-02, B-01** (slot–row bijection, terminal closure, phantom reversal, dispatched-has-session, queue coherence).
- Redis- and Docker-sourced checks (ZSET occupancy, zombie `claude` `ps`) are **unaffected** — those read live sources.
- The check *logic* is fine — only the SQL collector's data source is wrong: invoking the check functions directly against a hand-built snapshot with a live pull lease, **S-01 correctly excludes the leased row** and **E-05 fires** (both as designed).

## Impact
The canary is the orchestration-invariant **safety net**. On PostgreSQL it watches nothing at the SQL tier. This matters specifically for the pull / work-stealing migration (#1081): a **default-ON** decision leans on the canary to catch slot/lease/status regressions, and on the now-standard PG backend those checks would silently pass.

## Discovered
During the #1081 / #946 pull-coordination live pilot (2026-07-08). See `docs/testing/PULL_MIGRATION_RECONCILED_TESTING.md` §8, gap **G3**.

## Proposed direction (not greenlit — for triage)
Make the canary SQL collector resolve its connection from the app's `DATABASE_URL` (PG-aware), the same way the rest of the backend does, instead of hardcoding `sqlite3`. Keep SQLite working during the dual-track window (#1183).

## Acceptance criteria (draft)
- On a PostgreSQL instance, `run-cycle`'s SQL-tier checks read the **live** PG `schedule_executions` / slot state, not `/data/trinity.db`.
- A synthetic slot–row mismatch on PG **fires S-01**; a clean state does not.
- SQLite deployments unchanged.
- Canary works on both engines (dual-track, #1183).

AndriiPasternak31:
Fixes #1540

## Summary

The CANARY-001 invariant harness opened a **SQLite** connection unconditionally (`db.connection.get_db_connection` hardcodes `sqlite3` and ignores a non-SQLite `DATABASE_URL`), so on a **PostgreSQL**-backed deployment its SQL-tier collectors inspected a stale/empty `/data/trinity.db` instead of the live PG database. `_collect_known_agents` returned zero rows → `snap.agents` empty → the whole per-agent loop (S-01/E-05/B-02/E-04/G-04/B-01) plus the fleet-wide reads (E-02/E-03/G-03/E-06/L-03) went **vacuously green**. The harness was effectively blind to the live database at that tier.

This finishes the #1450 collector-wide migration:

- **6 SQL-tier collectors → `get_engine()`/`DATABASE_URL` seam** — `_collect_known_agents`, `_collect_executions`, `_collect_terminal_executions`, `_collect_terminal_rows`, `_collect_enabled_schedules`, `_collect_orphan_refs` now read through the same engine seam B-01 already used, so the whole harness is backend-consistent on PostgreSQL.
- **`PRAGMA table_info` → `sqlalchemy.inspect()`** — the column-existence probes (older-image skips) are now portable across SQLite and PG instead of SQLite-only DDL introspection.
- **Composite-PK `agent_tags` handling** — the orphan-scan primary key is taken from the **static** `db/tables.py` `Table` definition, not runtime reflection; `agent_tags` has a composite PK, so it maps to a synthetic `row_id` rather than assuming a single-column PK.

## Verification

- **Canary suite:** `tests/test_canary_invariants.py` → **139 passed, 1 skipped** (the 1 skip is the real-Postgres dialect gate, `TRINITY_TEST_PG_URL`). Re-run at ship stage, still green.
- **Real-PG proof:** the dialect path was exercised against a real `postgres:16-alpine` at both implement and review stages (via the `TRINITY_TEST_PG_URL`-gated test).
- **verify-local `--skip-agent`:** PASS — build + import-smoke, boot + health, and the integration suite all green.

## Notes

- **No schema change** — this is a read-path repoint only, so the dual-track migration rule (Invariant #3) does not apply.
- The `sqlite.*` `sources_unavailable` labels are kept **verbatim** as an internal skip-contract prefix for L-03/E-02, not a backend claim — they are a stable contract string, not a description of the DB in use.

## Follow-ups (for Andrii to file)

- A runtime **"silent re-blinding" operator alarm** — a canary cycle whose SQL-tier collectors all read zero rows on a fleet that is provably non-empty should surface loudly rather than passing green (defense against a future regression of exactly this bug class).
- The `.claude` submodule **test-runner catalog** entry for the new canary coverage is prepared locally but left **uncommitted** — the core team lands `.claude` changes via the private `trinity-dev` submodule, not through this public-repo PR.


vybe (approved):
Validated via /validate-pr: engine-seam repoint un-blinds the canary on PG, named #1540 regression coverage incl. real-PG dialect gate, composite-PK handling for agent_tags. TRINITY_TEST_PG_URL is test-only (no compose wiring needed). All checks green. Approving.

Issue #1158, opened 2026-06-11
security: rotate and remove embedded third-party API credential from trinity-cli


vybe:
## Summary

The published \`trinity-cli\` package (PyPI + Homebrew) embeds an obfuscated third-party API credential used for the access-request notification feature. The credential is recoverable from any installed copy and from package archives, so it must be treated as burned.

This is the follow-up that #758's Technical Notes called for ("should be moved to an env var in a follow-up") but which was never filed.

> **Note:** exact location and mechanism intentionally omitted from this public issue until the credential is rotated. Maintainers: see the internal audit report dated 2026-06-11.

## Context

Found during the 2026-06-11 repository audit (finding N1, Critical). Hardcoding any credential in a shipped artifact violates the repo's own public-repo secrets policy; history rewrite is futile since package archives exist — rotation is the real control.

## Acceptance Criteria

- [ ] Credential revoked/rotated in the provider console (do this FIRST, before any code change lands)
- [ ] Embedded credential and its decoder removed from the CLI source
- [ ] Access-request notification either proxied via a vendor-hosted endpoint (CLI holds no secret) or the feature is removed from the CLI
- [ ] Secret-scanner (e.g. gitleaks) job added to CI with a rule covering this credential format and obfuscated-blob heuristics
- [ ] Patch release published through the CLI publish flow

## Technical Notes

Old CLI versions already fail gracefully when the send fails, so revoking the key does not break installed clients.

vybe:
Resolved. The embedded credential has been **revoked** in the provider console (re-probe returns \`API key is invalid\`), and the secret + its decoder + the outbound send have been removed from the CLI in #1162 (deletes \`_notify.py\`; the no-URL \`trinity init\` branch now just points users to email access@ability.ai directly — the CLI carries no credential).

Closing as the live-key risk is fully mitigated. Two hardening follow-ups are tracked separately rather than blocking this:
- gitleaks CI rule covering \`re_…\`-shaped secrets and obfuscated blobs (will land with the broader secret-scanner job, see #851 / #850 deltas)
- patch CLI release to publish the cleaned package

PR #878, opened 2026-05-17
fix(security): non-root USER directives for production images (#874)

AndriiPasternak31:
## Summary

Closes #874 — CSO MEDIUM defense-in-depth gap flagged persistent since the 2026-04-05 audit (~6 weeks).

Trinity-built production images now run as non-root:

| Service | User | UID |
|---------|------|-----|
| backend | `trinity` | 1000 |
| scheduler | `trinity` | 1000 |
| mcp-server | `node` (built-in) | 1000 |
| frontend (prod) | `nginx` (`nginxinc/nginx-unprivileged`) | 101 |

Backend joins the host `docker` group via `group_add: ${DOCKER_GID:-999}` so UID 1000 keeps `/var/run/docker.sock` access on Linux. macOS Docker Desktop ignores `group_add` (UID translation handles it).

The dev Vite frontend image (`docker/frontend/Dockerfile`) is intentionally exempt — it has no production attack surface.

## Why it matters

Before this PR, any RCE in the FastAPI backend landed as **root** inside the backend container, and `/var/run/docker.sock` (mounted `:ro`) became a fleet-wide reconnaissance primitive — enumerate every agent container, owner labels, network bindings; read in-container env vars (`ANTHROPIC_API_KEY`, `REDIS_URL` with backend creds, `SECRET_KEY`). The socket is read-only so it's not a direct container escape, but "no USER" turned a single-point RCE into the worst kind of secondary blast radius.

## What's new in this PR vs the original ticket

In addition to the four Dockerfile changes from the ticket, the branch fixes the operational gaps that came out of a `/review` pass on the implementation:

1. **CI false positive** — the original `verify-non-root` step asserted `GET /api/agents` returns `[]`, but `list_all_agents_fast` catches every Docker exception and returns `[]` so the check passed even when `group_add` was broken. Now `docker exec trinity-backend python -c "docker.from_env().ping()"` runs the real round-trip.
2. **Fresh prod install** — backend bind-mounts `${TRINITY_DATA_PATH:-./trinity-data}` to `/data`. If the host dir didn't pre-exist, Docker created it root-owned and UID 1000 couldn't create `trinity.db`. `start.sh` now pre-creates it with UID 1000.
3. **Non-Debian Linux hosts** — `.env.example` shipped `DOCKER_GID=999` (Debian) which silently fails on RHEL/Fedora (~991) or Arch (990). `.env.example` now ships blank and `start.sh` auto-detects via `getent group docker`.
[...]

PR #1692, opened 2026-07-18
ci(security): gitleaks commit-time secret scanning on every PR (#1164)

### LINKED ISSUE #1164: ci: add gitleaks secret-scanning rule so embedded API keys (incl. re_-prefixed) cannot reland
## Summary

Add a gitleaks (or equivalent) secret-scanning step to CI so that an embedded API credential — like the one removed in PR #1162 / #1158 — is caught at PR time instead of by a later audit. The leaked key was a `re_`-prefixed provider key that default rulesets may not cover, and it was XOR-split across two base85 halves, which no scanner catches — so the rule set should cover both the raw prefix pattern and high-entropy string pairs in suspicious contexts.

## Context

- Audit finding N1 (#1158): an obfuscated third-party API credential shipped in the published CLI. Removal merged in PR #1162; prevention was explicitly deferred to a follow-up.
- This is a public repository — any committed secret is world-readable immediately and permanently (git history).

## Acceptance Criteria

- [ ] CI workflow runs gitleaks (or equivalent) on every PR against the diff/branch
- [ ] Custom rule catches `re_`-prefixed API keys in addition to the default ruleset (sk-, ghp_, xox*, AKIA, etc.)
- [ ] A test commit containing a fake `re_`-shaped key fails the check (verified once, then reverted)
- [ ] False-positive handling documented (allowlist file or inline annotation) so legitimate placeholder examples in docs don't block PRs

## Technical Notes

- Candidate home: a new `.github/workflows/secret-scan.yml` or a job appended to the existing container-security workflow (`.github/workflows/container-security.yml`) — standalone is cleaner since secret scanning should not be path-filtered.
- Entropy-based detection (gitleaks' default entropy rules) is the only realistic guard against deliberately obfuscated/split secrets; expect limited coverage there and treat the prefix rule as the primary win.

AndriiPasternak31:
Fixes #1164

## Summary

Adds commit-time secret scanning on every PR — the deferred prevention half of #1158 / PR #1162, closing the gap where an embedded `re_`-prefixed Resend key shipped in the published CLI and was only caught by a later audit. Public repo → any committed secret is world-readable and permanent, so the guard runs on **every** PR (no path filter, no label gate — #878 lesson).

- **`.github/workflows/secret-scan.yml`** — runs the **MIT gitleaks CLI** binary, version- and sha256-pinned (v8.30.1), **not** `gitleaks/gitleaks-action` (which requires a paid `GITLEAKS_LICENSE` for org-owned repos and fails on run 1). `permissions: contents: read`.
- **Commit-range scope** via `git merge-base` (`--log-opts`) — scans only the commits the PR adds past the fork point, never the base branch's own advance or the settled ancestor history.
- **Custom `trinity-resend-api-key` rule** (`.gitleaks.toml`) for the `re_` prefix the default v8.30.1 ruleset misses — `[extend] useDefault = true` keeps the built-in `sk-`/`ghp_`/`xox*`/`AKIA` rules + stopwords; the repo-prefixed rule id can never override a future default rule.
- **`--redact=100`** — findings are masked in the world-readable CI log so a match never re-leaks the secret.
- **`--exit-code 2` tri-branch, fail-closed** — clean (0) / finding (2 → fail) / any other non-zero = scanner/config/git error (→ fail closed, never misdiagnosed as a finding).

## Enforcement gap (honest)

**Detection shipped; enforcement is a follow-up.** The workflow runs on every PR, but the check is **NON-BLOCKING until a repo admin adds `secret-scan` as a required status check on `dev`/`main` branch protection** — a repo-settings toggle a code PR cannot perform. Its unconditional trigger is precisely what makes requiring it safe (never left "Expected — waiting forever"). Until that toggle, a red scan does **not** block merge: this is "detection shipped; enforcement = follow-up", not "cannot reland / solved".

## Invariant #3 (dual-track migrations)

N/A — no schema change (CI YAML + config + docs only).

## Verification

**verify-local intentionally skipped** — this is a CI-YAML/config change with no runtime surface, outside verify-local's image-build / boot / migration coverage. Local-proof evidence instead:

- Pinned-binary **sha256 verified** against the release checksum.
- Clean commit-range scan → **exit 0**.
- Planted a `re_`-shaped key → scan **exit 2, redacted** in output; removed via `git reset` (never committed).
- The `.gitleaks.toml` fixture-commit is **not** flagged (self-allowlist holds).
- **actionlint** and **shellcheck** green on the workflow.

## Ratified deviations (from review)

1. **No #1158 fingerprint in `.gitleaksignore`** — the settled leak's two base85 halves are **verified undetectable** by gitleaks under both `useDefault` and this config (neither half is a `re_` key nor keyword-adjacent; generic entropy on base85 is unreliable). Fabricating a fingerprint gitleaks would never emit is false coverage. `.gitleaksignore` stays a non-load-bearing baseline for future known-accepted findings; the range-scoped CI gate does not depend on it.
2. **`docs/memory/**` allowlist** — the live engineering docs carry API-usage examples (curl `Authorization: Bearer` commands, truncated JWTs, `KEY=`/`condition=` samples) the default `curl-auth-header`/`generic-api-key` rules flag; **8 such FPs verified this session**. Narrow scope: only `docs/memory/**` (plus `tests/**` and historical `docs/archive|releases|security-reports/**`) — **not** a blanket `docs/**` — is excluded. A broader `docs/**` exclusion is deferred to reviewer judgement; other doc trees stay scanned and rely on the inline `gitleaks:allow` escape hatch.

## Honest limits

Split / obfuscated secrets evade regex + entropy (the #1158 base85 halves are the proof) — **credential rotation is the real defense** against an original leak; this gate is defense-in-depth against a **verbatim re-land**. `tests/**` and the historical doc trees are documented **non-scanned zones** (gitleaks pre-skips allowlisted paths before per-finding regex). Distinct from GUARD-002, which is **runtime** output scanning — this is **commit-time source** scanning; complementary layers.

## Follow-ups (for Andrii)

- **File the enforcement-toggle issue** when this merges: a repo admin must add `secret-scan` as a required status check on `dev`/`main` branch protection to make the gate blocking.
- Optional / deferred: tighten the `re_` rule regex (digit/uppercase requirement) — RE2 has no lookahead, so the entropy floor is currently the only FP lever.


vybe (approved):
Validated via /validate-pr: pinned+sha256-verified MIT gitleaks binary (not the org-licensed action), least-privilege workflow, commit-range scope, redacted output, fail-closed tri-branch. Grep hits are gitleaks' own documented example patterns only. Docs complete. Honest enforcement-gap note acknowledged — follow-up is the required-check toggle. Approving.

PR #1686, opened 2026-07-17
feat(github): per-user GitHub credentials (abilityai/trinity-enterprise#162)

obasilakis:
## Summary
- Lets each user store a personal GitHub PAT so a **non-admin** can create an agent from a repo the shared admin token can't see. Agent creation resolves **per-agent → owner's per-user (live) → global**, so each agent carries its creator's scope instead of the fleet-wide admin token.
- **OSS-core** (confirmed): edition-agnostic onboarding plumbing, lands in the public repo; filed on the enterprise tracker per routing only. Same per-user encrypted-credential primitive already shipped free by `subscription_credentials`.
- Shape A — the per-user PAT is a **create-time** credential that seeds the existing #347 per-agent PAT; no reactive recreate, no new propagation path.

## Design decisions (surfaced at plan review)
- **Persist only the user tier, never the global fallback.** Literal AC #4 ("persist the resolving credential") contradicts AC #9: `github_pat_propagation_service` skips any agent with a per-agent PAT, so persisting the *global* PAT would silently drop every new agent out of admin-PAT rotation on single-admin installs. We persist as the per-agent PAT only when `tier ∈ {per_user, fork}`; a global-fallback agent keeps `github_pat_encrypted` NULL.
- **Two separate ladders — the safety pin.** The create resolver (`resolve_github_pat`, 3-tier) is deliberately distinct from the recreate/restart ladder (`get_github_pat_for_agent`, 2-tier per-agent→global). Making the recreate ladder re-derive the live per-user tier would force-recreate a running agent (killing in-flight work) the instant its owner adds a PAT in Settings (the #1560/#1557 class). The recreate ladder is untouched except the settings_service relocation.
- **`users` column, not a new table.** A per-user PAT is 1:1 with a user (like `suspended_at`) — cheaper and more strandable if #177 (GitHub App) lands.
- Public-repo-no-PAT (**AC #8**) is **out of scope → sibling #123** (owns the `startup.sh` gate + base-image rebuild). Commented on the issue.

## Changes
- **DB**: `users.github_pat_encrypted` (AES-256-GCM) across schema.py/tables.py/migrations.py + Alembic `0025` (head `0024`); 4 accessors + facade delegations; column excluded from the general user dict.
- **Resolver**: `resolve_github_pat(agent_name, owner_id)` in `settings_service`; `get_github_pat_for_agent` relocated there (kills service→router import, Invariant #1) and re-exported from `routers/git.py`.
[...]

Issue #1484, opened 2026-07-06
refactor(agents): decompose create_agent_internal (830 lines, CC 120) with characterization tests first


vybe:
## Summary

`create_agent_internal` (`src/backend/services/agent_service/crud.py:118`) is the single worst function in the backend: **830 lines, cyclomatic complexity 120** (thresholds: 200 / 30). It has grown +116 lines / +9 CC since the 2026-06-02 audit (714 lines, CC 111) — every agent-creation feature lands inside it.

## Context

Flagged as the top critical finding by `/refactor-audit` (report: `docs/reports/refactor-audit-2026-07-06.md`). Previously one acceptance criterion inside the grab-bag issue #1028; split out because this function needs its own test-first PR series. Everything flows through it: template processing, runtime selection (#1187), volume/network provisioning, credential + MCP config injection, `data_paths` materialization (#1169), agent-scoped MCP key generation, ownership registration, avatar, container start.

## Acceptance Criteria

- [ ] Characterization tests written FIRST covering the major creation paths (GitHub template / local template / system agent) and the error/rollback paths
- [ ] Function decomposed into named stages (indicatively: validate → provision → configure → register → start) behind the existing signature — callers unchanged
- [ ] Partial-failure cleanup behavior preserved exactly (container/volume/ownership rollback on mid-creation errors)
- [ ] No stage function >100 lines; CC <20 per function
- [ ] Network attachment still hard-codes `trinity-agent-network` (network-topology rule: agents never join the platform network)
- [ ] No behavior change: unit suite green; agent create/start integration tests pass

## Technical Notes

- Do NOT refactor sibling functions in the same PR — `deploy_local_agent_logic` (306 lines, CC 40) and `recreate_container_with_updated_config` (265 lines, CC 46) stay under #1028.
- Stage boundaries should mirror the rollback points that already exist in the function's except/cleanup blocks — extract along those seams, don't invent new ones.

AndriiPasternak31:
/claim — picking this up as part of the July reliability/refactor batch.

PR #1694, opened 2026-07-18
feat(reminders): agent self-reminders — one-shot deferred self-trigger (#1296)

### LINKED ISSUE #1296: feat: agent self-reminders — agent-callable one-shot deferred self-trigger ("remind me to do X later")
## Summary

Give agents a first-class **self-reminder** primitive: while running, an agent can schedule a **one-shot, future re-invocation of itself** with a message it chooses — "remind me to check this PR in 2 hours," "follow up with the user tomorrow at 9am," "re-run the digest after the export finishes." When the timer fires, Trinity dispatches a normal execution of that same agent carrying the reminder message, through the standard capacity/slot path. The agent can list and cancel its pending reminders.

This is the missing time-deferred sibling of `run_agent_loop` (#740): a loop runs iterations *back-to-back now*; a reminder fires *once, later*.

## Context

Long-running autonomous agents constantly need to defer work to a future wall-clock time and then reliably pick it back up. Today there is **no agent-callable way to do that** — every existing primitive misses on one axis:

- **`agent_schedules` (cron):** recurring and operator-oriented. Validation is cron-only (`croniter` / APScheduler `CronTrigger`) — there is no one-shot "fire once at T" mode. An agent that wants a single future nudge would have to create a cron schedule and then self-delete it after one fire, which is a leaky hack and pollutes the owner's Schedules tab.
- **`run_agent_loop` (#740):** sequential bounded iterations executed immediately; no deferral to a future time.
- **`emit_event` / `subscribe_to_event` (EVT-001):** reactive (fires on another agent's event), not time-based.
- **`send_notification`:** passive — surfaces a message to the platform/operator; it does **not** re-invoke the agent later.
- **Pre-check hook (#454):** gates an *already-scheduled* run; it doesn't create a future trigger.
- **Operator queue (OPS-001):** human escalation, not agent self-scheduling.

So the gap is a durable, **one-shot, agent-initiated deferred self-trigger**. This fits Trinity's agent-owned-pipeline philosophy (requirements §34, #919): the agent owns *when and why* to follow up; Trinity provides only the thin, standardized deferred-trigger surface — it does not become a workflow engine.

## Acceptance Criteria

- [ ] Via an **agent-scoped MCP key**, an agent can schedule a one-shot self-reminder with **either an absolute time or a relative delay** plus a message payload.
- [ ] When the timer fires, the backend dispatches a normal execution of **that same agent** carrying the reminder message, via the standard `capacity_manager` admit/slot path (sharing the agent's `max_parallel_tasks` budget, like loops #740). The execution's `triggered_by` reflects the reminder origin.
- [ ] Reminders are **durable** — they survive a backend / standalone-scheduler restart and still fire, or are deterministically reconciled when past-due (mirror the cleanup-service startup hooks used for loops #740).
- [ ] An agent can **list its pending reminders** and **cancel** one before it fires.
- [ ] **Self-only authorization:** an agent-scoped key can set/list/cancel reminders only for itself; cross-agent attempts are rejected (mirror the agent-scoped gating used by heartbeat and operator-queue #1101).
- [ ] The create boundary accepts an **`Idempotency-Key`** and de-dupes per Invariant #18 (new trigger types must accept idempotency keys).
- [ ] **Abuse bounds** enforced: max pending reminders per agent, min/max delay window, max message length.
- [ ] MCP tools added and documented; three-surface sync (backend router + MCP tool) per Invariant #13.

## Technical Notes

- **Storage — prefer reuse over a new table.** Extending `agent_schedules` with a nullable one-shot `fire_at` (cron NULL ⇒ one-shot) likely beats a new `agent_reminders` table: it inherits soft-delete (#834), `schedule_executions` history, the idempotency wiring, and restart recovery for free. Evaluate vs. a dedicated lightweight table during design. Either way the firing path must use APScheduler `DateTrigger` (one-shot), not `CronTrigger`.
- **Durability:** must not be an in-memory `asyncio` timer — the standalone scheduler process has to pick it up after a restart, same as cron schedules.
- **New trigger boundary ⇒ Invariant #18:** route the create endpoint through `idempotency_service` (`begin`/`complete`/`fail`).
- **Execution path:** dispatch through `task_execution_service.execute_task(...)` with a new `triggered_by` value (e.g. `"reminder"`) so it shows up correctly in Executions and the Overview analytics `_TRIGGER_BUCKETS` (#1107) — add the bucket so it never silently lands in `Other`.
- **Auth/gating:** agent-scoped MCP key resolves to self; reject reminders targeting any other agent.
- **MCP surface:** new tools (e.g. `set_reminder`, `list_reminders`, `cancel_reminder`) — fold into `src/mcp-server/src/tools/schedules.ts` or a new `reminders.ts`.
- **Relevant files:** `src/backend/routers/schedules.py`, `src/backend/services/scheduler_service.py`, `src/backend/services/task_execution_service.py`, `src/backend/db/schema.py` + `db/migrations.py` (if extending `agent_schedules`), `src/mcp-server/src/tools/`.
- **Likely sizing:** medium — new trigger type + one-shot firing + MCP tool + idempotency + restart recovery + abuse bounds. (Final complexity to be set at grooming.)

## Related

- Sibling primitive: #740 (`run_agent_loop`) — immediate sequential execution; this is its deferred one-shot counterpart.
- Adjacent triggers: #291 (webhook triggers), #454 (agent-owned pre-check hook), EVT-001 (event pub/sub), OPS-001 (operator queue).
- Philosophy: requirements §34 / #919 (agent-defined pipelines — Trinity provides the thin surface, the agent owns the logic).
- Constraint: Invariant #18 (idempotency on trigger boundaries).

AndriiPasternak31:
Fixes #1296

## Summary

Agent **self-reminders** — a durable one-shot deferred self-trigger, the time-deferred sibling of loops. An agent schedules a message to itself at an absolute `fire_at` (ISO) **XOR** a relative `delay_seconds`; when the reminder comes due the platform runs a normal execution of that same agent.

- **Durable `agent_reminders` table** — pending reminders survive restarts (no in-memory-only timers).
- **Standalone-scheduler `DateTrigger` fire path** — the single-instance standalone scheduler arms one APScheduler `DateTrigger` per pending row (`_reconcile_reminders` at boot + a 60s sync-loop own-try + full-reload). Single-instance by design, so there is **no `--workers 2` double-fire** (unlike the multi-worker backend).
- **Fires through the existing internal execute-task** — on fire the scheduler creates the `schedule_executions` row itself (real id, `__manual__` schedule) and dispatches `triggered_by="reminder"` through the **idempotency-keyed** internal execute-task on the **standard `capacity_manager` admit** — no bespoke dispatch path.
- **At-least-once delivery** — a `firing` state + a `fire_attempts` cap + stale-`firing` reclaim guarantee a due reminder is not silently dropped.
- **3 self-scoped MCP tools** — `set_reminder` / `list_reminders` / `cancel_reminder`, self-only (connector keys rejected, off the ghost fence); create is idempotent over raw input (terminal rows excluded from replay).
- **`triggered_by="reminder"` wired into all three trigger constants** — first-class "Reminders" analytics bucket, no silent `Other`.
- **Full #1644 retention-guard wiring** — the `agent_reminders` retention sweep goes through the blast-radius guard like every other window-driven prune.

Bounds: pending 25 / durable daily 100 / min-60s / max-30d / timeout clamped to the agent cap.

## 5-track migration (Invariant #3)

The new `agent_reminders` table lands on all five surfaces, parity trio green:

1. `src/backend/db/schema.py` — DDL
2. `src/backend/db/migrations.py` — SQLite versioned migration
3. `src/backend/migrations/versions/0028_agent_reminders.py` — Alembic revision (`down_revision = 0027_users_github_pat`, the current dev head from #1686)
4. `src/backend/db/tables.py` — MetaData
5. `AGENT_REFS` — rename/cascade coverage

## Verification

- **verify-local FULL PASS** — real prod images + agent base image, isolated sibling stack, incl. **fresh-DB migration** (both tracks) and **agent-exercise** end-to-end.
- **`tests/scheduler_tests` 224/224**, **`tests/unit/test_1296_reminders.py` 33/33**.
- The **9 broad-sweep reds are pre-existing on the dev tip**, not from this branch: 8× `test_fork_to_own` (broken by #1686's per-user GitHub-credential change — fixed by the #1484 PR in this batch) + 1× `test_1474`.

## Design notes

- **Timeout = outcome-unknown → assume-dispatched.** A `TimeoutException` on the internal execute-task is treated as "we don't know the outcome, assume it dispatched" — we do **not** force a FAILED row. The consequence is a documented **at-least-once double-execute window**, which is bounded (by the `fire_attempts` cap) and reaped (stale-`firing` reclaim). A clean failure, by contrast, writes a status-guarded FAILED + retry.
- **`__manual__` rows no-op the cron paths.** The reminder's `schedule_executions` row carries a `__manual__` schedule id, so the cron retry/validation paths correctly no-op on it (it is not a recurring schedule).

## SHIP-stage note

One config-packaging fix added at ship (final commit): the six `REMINDER_*` / `MAX_*REMINDER*` backend `os.getenv()` tunables are wired into `.env.example`, `docker-compose.yml`, and `docker-compose.prod.yml` `backend.environment:` — matching how `REPORT_RATE_LIMIT` (#918) is wired. The feature ships functional on code defaults regardless; this makes the operator levers actually reachable on deploy (the #1056/#1039 packaging class).

## Follow-ups (for Andrii to file)

- **PG-backed CAS contention test variant** — the fire-claim CAS is exercised on SQLite; add a Postgres-backed contention variant once PG is the primary.
- **Frontend `ExecutionsPanel` trigger-filter "reminder" option** — UI surface, out of scope for this PR.
- **Pull-mode note** — when Epic #1045/#1081 (pull / work-stealing) lands, the fire step swaps from direct execute-task to **enqueue**; the reminder becomes a producer of pull-mode work rather than a direct dispatcher.


vybe (approved):
Validated via /validate-pr: 5-track migration complete (schema/migrations.py/Alembic 0028 on 0027 head/tables.py/AGENT_REFS), all 6 REMINDER_* knobs wired into both compose files + .env.example, at-least-once fire semantics with bounded double-execute window documented, triggered_by=reminder in all trigger constants. Rebased onto the #1693 schedules package (bucket lines re-applied to analytics.py, test import retargeted — same lockstep pattern #1693 used). Local runs green: reminders+analytics+CAS-observability 65/65, scheduler suite 17/17. Approving.

Issue #384, opened 2026-04-18
Reset-preserve-state operation (S3)


vybe:
## Summary

New endpoint to adopt latest upstream (main) while preserving agent-specific state files.

## Context

**Problem P2/P3**: When parallel history exists, both UI options ("Pull First" / "Force Push") make things worse. The only valid recovery is: adopt main as baseline, overlay workspace state, push to working branch. Currently requires SSH access.

### New endpoint
- [ ] `POST /api/agents/{name}/git/reset-to-main-preserve-state`
- [ ] Backend proxies to agent server

### Guardrails
- [ ] ALWAYS snapshot persistent-state files to `.trinity/backup/<timestamp>/` before reset
- [ ] Create tar archive downloadable via Files panel
- [ ] Refuse if agent has no git config
- [ ] Refuse if agent is currently running a task (check activity service)

### Agent-server implementation
- [ ] New endpoint proxied from backend
- [ ] Uses S4's `_reset_preserving_persistent_state()` subroutine
- [ ] Commits "Adopt main baseline, preserve state"
- [ ] Pushes with `--force-with-lease`

### New agent-server primitives
- [ ] `POST /api/agent-server/snapshot?paths=...` — returns tarball of matched files
- [ ] `POST /api/agent-server/restore` — restore from tarball with path allowlist

### Response
```json
{
  "success": true,
  "files_preserved": ["workspace/decisions.csv", ".trinity/config.yaml", ...],
  "commit_sha": "abc123",
  "backup_path": ".trinity/backup/2026-04-18T..."
}
```

## Technical Notes

Depends on #383 (S4 persistent state allowlist).

**Files to modify:**
- `src/backend/routers/git.py` — new endpoint
- `src/backend/services/git_service.py` — proxy logic
- `docker/base-image/agent_server/routers/git.py` — implementation
- `docker/base-image/agent_server/routers/files.py` — snapshot/restore primitives

Part of Epic #381

PR #1697, opened 2026-07-20
feat(skills): full-directory skill packages + frontmatter contract (abilityai/trinity-enterprise#183)

vybe:
## Summary

Phase 1 of the skills-platform epic (abilityai/trinity-enterprise#182): a skill becomes a **versioned directory package** instead of a lone `SKILL.md`.

- **Full-directory injection.** Previously `inject_skills()` wrote one file per skill — `scripts/`, templates, and resources were silently dropped, so any multi-file skill broke after injection with no warning. Injection now ships the whole directory through the **existing** agent-server restore primitive (#384/#1169): no new endpoint, no agent-server change, Invariant #5 untouched.
- **Frontmatter contract.** `automation`, `user_invocable`, `allowed-tools`, and `requires: {packages, binaries, env}` are parsed platform-side and surfaced via REST + MCP, alongside package metadata (`multi_file`, `file_count`, `size_bytes`, `version`).
- **Honest results.** Per-skill `{status, files_written, error?, warnings[]}` with named codes — unmet declared deps or a partially-shipped package produce warnings, never silent success.

Gating decided at design time per the open-core rule: **OSS-core** (the library mechanism is already OSS; gating the fix would fork the service and leave OSS with the known-broken silent-drop path). Epic monetization lands in Phase 2 (placement / skill-runner, trinity-enterprise#139).

| Decision | Why |
|---|---|
| Tar source = `git archive HEAD` | Atomic against a concurrent admin re-sync (`git reset --hard`); git member modes carry exec-bit truth |
| Version = git **tree SHA** | Deterministic across clones by construction — a content hash of tar bytes embeds checkout mtimes and would never match twice |
| Meta member appended **LAST** | Restore extracts in member order; meta-first would pin a new-version marker over old/partial files and defeat the repair loop |
| **Manifest-based** prune (prev manifest − new, dir-confined, capped) | Platform deletes only what it previously wrote — runtime artifacts (`__pycache__`, downloaded models) and agent-authored files are untouchable |
| REGTYPE-only members | An out-of-archive symlink member raises `KeyError` inside `restore_from_tar` mid-extraction; following one at build time would exfiltrate backend-readable files |
| Gitignore + untrack injected dirs | The 15-min auto-sync deliberately commits `.claude/`, so every package would otherwise land in every agent's GitHub repo — the #1595/#1596 bloat class |

## Changes

**Backend**
- `services/skill_packaging.py` (new) — pure primitives: hardened contract parse, archive member vetting, tar assembly, prune diff
- `services/skill_service.py` — injection rewrite (packaging, restore transport, prune, dep probe, injection lock, CLAUDE.md rebuild); contract-aware discovery with commit-keyed cache
- `routers/skills.py` — contract pass-through, 409 on lock contention, name guard on bulk assign
- `services/agent_service/lifecycle.py` — start path passes `force=False` (skip version-unchanged), surfaces warning counts
- `db_models.py` — `SkillInfo` extended (all additive with defaults)

**MCP** — `tools/skills.ts`: contract fields on list/get; `sync_agent_skills` surfaces per-skill warnings even on success

**Docs** — `requirements/skills.md` (§21.4 rewritten + new §21.6), `feature-flows/skill-injection.md` (rewritten), `skills-library-sync.md`, `skills-on-agent-start.md`, `architecture.md`, plus two `learnings.md` entries

**No DB schema change. No new endpoints. No frontend change.**

## Test Plan

- [x] `pytest tests/unit/test_ent183_skill_packages.py -v` — **52 passed**
- [x] Neighbors green: `test_skill_service_user_agent`, `test_start_agent_skip_inject`, `test_1560_agent_redis_key_parity`, `test_1410_unresolved_slash_command`, `test_agent_readiness_probe`, `test_inject_assigned_credentials`, `test_subscription_auto_switch_no_cred_import` (73 passed total)
- [x] `npx tsc --noEmit` clean for the changed MCP module
- [ ] Manual: sync a multi-file library skill, assign, inject, verify `scripts/` present + executable, `.trinity-skill.json` written, `.gitignore` updated

High-value coverage: round-trips against the **real** `restore_from_tar`, a **real-git** end-to-end (repo → archive → filter → restore, tree-SHA determinism across two clones), and the injected python scripts compiled + executed with hostile inputs.

## Review history

Planned via `/autoplan` (premise gate + independent strategy and engineering voices) and reviewed via `/review` with an adversarial pass. All findings applied before this PR: prune confined to the skill dir against an untrusted agent-side manifest; injection-lock TTL raised above the worst-case injection with token-checked release; repair-delete no longer fires on transport failure (data-loss window); name guard applied before any in-container exec; fallback skills no longer queue chmod paths; list cache keyed on the clone's live commit (`--workers 2` staleness).

Fixes abilityai/trinity-enterprise#183


dolho:
## Review — skill packages + frontmatter contract

**Solid. No blockers.** Reviewed all 5 code files against the agent-server `restore_from_tar` contract and the `_matches_any` glob matcher.

### Verified correct
- **Restore contract match** — `restore_from_tar` returns member `name` (arcname); the `sent - restored` accounting compares like-for-like. Glob `.claude/skills/<name>/**` matches nested files + the meta member (`_matches_any` `/**` prefix branch).
- **Name-guard coverage complete** — bulk assign → 422, single assign → 404 via `get_skill`→`_skill_dir`, inject loop per-name, `_skill_dir` realpath containment as the second guard. A traversal-shaped name never reaches path math or an exec argument.
- **Lock semantics** — `SkillInjectionBusy` is raised before `lock_token` is bound, so the `finally` can never delete another holder's lock. Random-token check-and-delete release; TTL exceeds worst-case; fail-open on Redis down.
- **Prune confinement** — prefix-locked to the skill dir against the untrusted agent-side manifest, capped, meta excluded; the stored manifest excludes the meta member by construction.
- **Warning-list aliasing** — the `warnings` list threaded through `_restore_skill` / prune / `_apply_dep_warnings` is one object, so appends after `results[name] = outcome` are still reflected. Consistent.
- **`r["status"]` always present** on every result path → no `KeyError` in the final counts.
- **Meta-appended-last + repair-only-on-HTTP-error** — a transport failure does NOT delete the dir (data-loss window closed, matches the PR claim).
- **Exec injection surface** — base64-pipe charset is shell-safe; embedded names are JSON-encoded and regex-gated.
- All three `inject_skills` callers updated; the return-shape change is additive.

### Minor (non-blocking)
- `list_skills` returns the cached list object directly (shared mutable). Callers are read-only, but a `.copy()` is cheap insurance.
- The total-byte cap skips unchanged skills (they aren't re-sent) — correct; a one-line comment noting the intent would help.
- `_run_agent_python` swallows stderr (`2>/dev/null`) — advisory probes, fine, but makes field debugging harder.

### Note
The final Test Plan box (manual multi-file round-trip: `scripts/` present + executable, `.trinity-skill.json` written, `.gitignore` updated) is still unchecked — worth doing before merge since it's the exact silent-drop path this PR fixes. Automated coverage (52 pass + real-`restore_from_tar` round-trip) covers most of it.

Targets `dev`. Approve pending the manual verify.

dolho (approved):

PR #1687, opened 2026-07-17
feat(agents): agent display label across operator, monitoring, executions & graph (#1643)

dolho:
## What

Follow-up **5/5** of the #964 display-name track — spread the agent display label to the surfaces the ent#181 foundation (#1676) didn't reach: operator queue, monitoring, executions, the collaboration graph, tab titles, and prose/toasts.

Builds directly on #1676 (merged to `dev`): `agent_ownership.display_label`, `GET`/`PUT /label`, the `utils/agentName.js` resolver, and the `agent_label_changed` WS handler already ship. **No backend change here.**

## The decision (AC #1)

Dense operational surfaces (executions, operator queue, monitoring) read only the **slug** from their own API payloads. Rather than add a mutable `display_name` to each of those high-volume endpoints — staleness risk + N sources of truth — the label is resolved **on the frontend off the loaded agents**, from one place, staying live via the existing `agent_label_changed` handler:

- `agentsStore.displayNameForSlug(slug)` → label or slug (prose)
- `agentsStore.agentRefForSlug(slug)` → agent object or bare slug (feeds `agentNameTooltip`)

An unloaded slug falls back to itself, so a cold surface never regresses. Recorded as requirements **§1.3.1 FR-6**.

## Render rule by class (#964)

| Class | Rule | Surfaces |
|-------|------|----------|
| Dense operational tables | **slug primary**, label as hover tooltip | ExecutionsPanel · operator QueueCard/QueueList/QueueItemDetail/ResolvedCard/NotificationsPanel · MonitoringPanel (alert + health) · MobileAdmin ops/notif · RoleMatrix (100px, slug-only + tooltip) |
| Prose / toasts | **label alone** | Agents.vue toasts · Settings.vue confirm + subscription option · PortalConversation header/switcher/placeholder · PortalFilesPanel headings/toast |
| Collaboration graph | render label, **`data.label` stays the slug** (action key) | network.js node data (agent + system) + live handler · AgentNode |
| Reviewed case-by-case | label + slug link / tooltip | ExecutionDetail breadcrumb + source-agent · PermissionsPanel · ReplayTimeline · FoldersPanel · AgentWorkspace + AgentBrainOrb headers |

Notes:
- **`AgentAvatar` always receives the slug** (incl. every touched surface).
- **Comma-joined lists** (GitHub-PAT propagation failure list) keep the slug — long labels make them unreadable (per the issue's own note).
[...]

Issue #1643, opened 2026-07-15
feat: Agent display name across remaining surfaces — operator, monitoring, executions, graph (#964 follow-up 5/5)


vybe:
## Summary

Apply the display name to the remaining surfaces: operator queue, monitoring, executions, the collaboration graph, tab titles, and prose/toast strings. Lowest value per site and safely deferrable — until this lands these surfaces keep showing the slug, which is correct behavior, just not the polished version.

Follow-up 5 of 5 from the design decision on #964.

## Context

Per the #964 render rule, these split across three surface classes:

- **Dense operational tables** (machine-facing) — the slug stays primary; the display name is supplementary at most. Executions rows, operator queue rows, monitoring alerts, and `RoleMatrix` (a 100px column header — extremely tight).
- **Prose / toasts** (identity already established) — display name alone reads better than a concat.
- **Collaboration graph** — needs a store change before it can show anything.

Most of these components read `agent_name` out of **API payloads**, not the agents store, so the backend payloads need the field or the frontend needs a slug→display resolver. That decision is the main design work in this issue.

## Acceptance Criteria

- [ ] Decision recorded: add `display_name` to operator/monitoring/executions payloads, **or** build a frontend slug→display resolver from the agents store (with the tradeoff stated)
- [ ] Dense operational tables keep the **slug primary** per the #964 render rule:
  - [ ] `src/frontend/src/components/ExecutionsPanel.vue` table rows
  - [ ] `src/frontend/src/components/operator/` — `QueueCard`, `QueueList`, `QueueItemDetail`, `ResolvedCard`, `NotificationsPanel`
  - [ ] `src/frontend/src/components/MonitoringPanel.vue` alert + health rows
  - [ ] `src/frontend/src/components/process/RoleMatrix.vue` (100px column — slug only)
  - [ ] `src/frontend/src/views/MobileAdmin.vue` ops rows
- [ ] Collaboration graph shows the display name **without** breaking actions:
  - [ ] Sibling field added in `src/frontend/src/stores/network.js` (both the agent and system-agent node data)
  - [ ] `data.label` remains the **slug** — it is the API key for `router.push` / `toggleAutonomy` / `toggleAgentRunning` in `AgentNode.vue`
- [ ] Prose / toast strings use the display name alone — `Agents.vue` notifications, `Settings.vue` confirms, `PortalChat.vue` empty state, `PortalFilesPanel.vue` headings
[...]

Issue #964, opened 2026-05-28
spike: Support a friendly display name on agents separate from the slug-based name


vybe:
## Summary

Today an agent's name is its identifier — slug-cased, lowercase, no spaces (e.g. `tom-marketing-operations-agent`). Users want to call it `TOM — Marketing operations` in the UI. The current create / rename flows reject non-slug input.

This is a **feature spike**, not a fix. Decide whether to add a separate `display_name` field, or to relax slug rules (and pay the schema/Docker cost).

## Context

Reported during a UI/UX review session. The slug is load-bearing — per Invariant #1 / schema in `docs/memory/architecture.md`, `agent_name` is the PK across ~20 tables and is used as a Docker container label (`trinity.agent-name`), volume name (`agent-{name}-public`), and route segment (`/api/agents/{name}/...`). Changing the slug rules globally is high blast-radius.

Cheapest cleanest path is almost certainly: keep the slug as today, add an optional `display_name TEXT` column to `agent_ownership`, render it where humans look (avatar caption, agent list, header), keep the slug everywhere machines look (URLs, MCP, Docker labels). Doable but worth a design pass before code.

## Open Questions

- Add `display_name` (UI-only) and keep the slug, or actually relax slug rules?
- If `display_name`: who can set it (owner only? collaborators?), what's the max length, unicode handling, profanity / abuse policy
- How does it interact with `rename_agent` flow (#834 cascade rename)?
- Does the MCP tool surface need it (`get_agent_info`, `list_agents`)? External orchestrators?
- Where exactly does it render — agent list, agent header, avatar caption, anywhere else?
- Search / disambiguation: today users type the slug to find an agent — does display name need to be searchable?

## Acceptance Criteria

- [ ] Decision recorded (issue comment) on display_name approach vs slug relaxation
- [ ] If display_name approach: short design doc (issue comment is fine) covering schema, API, MCP surface, render sites
- [ ] Implementation broken into follow-up issues — this spike does not ship code
- [ ] Update `docs/memory/architecture.md` schema section if the decision lands

## Technical Notes

- Constraints from Invariant #1 (`docs/memory/architecture.md`): `agent_name` is PK across many tables, used as Docker label, volume name, route segment
- Existing rename flow: `routers/agent_rename.py`, `services/agent_service/` rename, manual cascade across 17+ tables
- Sites that probably need the display name if we add one: Agent list view, Agent detail header, Agent Avatar dialog, agent cards in collaboration graph, system agent picker
[...]

PR #1699, opened 2026-07-20
feat(agents): agent display name in pickers, search, and sort (#1642)

### LINKED ISSUE #1642: feat: Agent display name in pickers, search, and sort (#964 follow-up 4/5)
## Summary

Make agents findable by their display name: the `Display (slug)` concat treatment in pickers and filters, search that matches both fields, and a stated sort key. Requires a store-shape change first — four components currently iterate over arrays of bare strings.

Follow-up 4 of 5 from the design decision on #964.

## Context

These belong together because they share one store change and one purpose: **finding the right agent**. That's exactly what the "slug must stay visible" constraint from #964 protects. Today `src/frontend/src/views/Agents.vue` filters on `agent.name` only — once display names exist, a user types "TOM" and gets zero results.

Per the #964 render rule, pickers and filters are the surface class where disambiguation matters most, so they carry the slug inline via concat rather than dropping it.

## Acceptance Criteria

- [ ] Store shape changed so agent lists carry both names — `agentNames` / `availableAgents` are currently arrays of bare strings and cannot carry a second field. Affects:
  - [ ] `src/frontend/src/components/ExecutionsPanel.vue`
  - [ ] `src/frontend/src/components/ReportsPanelFleet.vue`
  - [ ] `src/frontend/src/components/operator/QueueList.vue`
  - [ ] `src/frontend/src/components/operator/NotificationsPanel.vue`
- [ ] All six `<option>` dropdowns render `Display name (slug)` when a display name is set, and the bare slug when not — also covers `src/frontend/src/views/FileManager.vue` and `src/frontend/src/views/Settings.vue`
- [ ] `<option>` values remain the **slug** — filtering/selection still keys on slug, never the display name
- [ ] Agent-list search in `Agents.vue` matches **both** slug and display name (case-insensitive)
- [ ] Sort key for "Name (A-Z)" decided and documented — display name when set, else slug (or an explicit alternative recorded on the issue)
- [ ] All `getActivityState(agent.name)` / `getSuccessBarPercent(agent.name)` / `getAgentTags(agent.name)` style lookups still key on the slug
- [ ] Long display names truncate gracefully in narrow selects without hiding the slug entirely

## Technical Notes

`Settings.vue` already string-concatenates a suffix onto an `<option>` label (`{{ agent.name }}{{ ... ? ' (on X)' : '' }}`) — that's the precedent for the concat pattern here.

`<option>` elements render text only and cannot hold a second line, which is why this surface class uses concat rather than the two-line idiom used on the roomy surfaces.

The three duplicated agent-list layouts in `Agents.vue` must change together — they're near-identical blocks and it's easy to update one and miss the others.

Depends on: the #964 schema/API/MCP issue. Best sequenced after the roomy-surfaces issue so the store change lands once the field is proven.

dolho:
## What

Follow-up **4/5** of the #964 display-name track — make agents findable by their display name. Builds on the `display_label` foundation (#1676) and the frontend resolvers from #1643 (#1687), both already in `dev`.

## Changes (per the ACs)

- **Pickers carry the slug inline.** New `agentOptionLabel(agent)` → `Display name (slug)` when a distinct label is set, else the bare slug. Wired into all **six** `<option>` dropdowns — `ExecutionsPanel`, `ReportsPanelFleet`, operator `QueueList` + `NotificationsPanel`, `FileManager`, `Settings` (subscription assignment). Every `<option>` **value stays the slug**, so filter/select never keys on the label.
- **Search matches both fields.** `Agents.vue` name filter now matches slug **or** display name (case-insensitive) — fixes the exact bug in the issue: typing "TOM" against a `tom-marketing-ops` slug returned nothing.
- **Sort key — decided + documented (AC).** "Name (A-Z / Z-A)" sorts by the **display name when set, else the slug** (`agentDisplayName`, in the store's `_getSortedAgents`). Rationale: sorting by the slug while the row renders the label orders the list by an invisible key. (Recorded on #1642 and in requirements §1.3.1 FR-7.)

## Design decision — no store-shape change

The issue's AC suggested reshaping `agentNames` / `availableAgents` (bare slug-string arrays) into `{name, display_label}` objects. I took the lower-blast-radius path instead: resolve the label off the loaded agents via the FR-6 resolvers (`agentRefForSlug`), which stay live through the `agent_label_changed` WS handler. This keeps every existing consumer of those getters working, mirrors how #1643 solved "payloads only carry the slug", and degrades gracefully (an unloaded slug renders as itself). Same visible outcome — `Display (slug)` in the dropdowns — without touching four components' data contracts.

Every per-agent lookup (`getActivityState` / tags / stats / router actions) still keys on `agent.name`; only the option label, the search predicate, and the sort comparator changed.

## Verification

All 7 touched SFCs compile via `@vue/compiler-sfc`; the 2 JS files pass `node --check`; `agentOptionLabel` / the sort comparator / the dual-field search unit-verified (labelled → `TOM (tom-mktg)`, unlabelled → slug, sort orders by visible name, search matches both).

Related to #1642

Fixes #1642

### DIFF
--- a/docs/memory/requirements/core-agent.md
+++ b/docs/memory/requirements/core-agent.md
@@ -76,6 +76,23 @@
   store isn't fetched yet); the next navigation self-heals. Comma-joined agent
   lists (e.g. the GitHub-PAT propagation failure list) keep the slug — long
   labels make them unreadable.
+- **FR-7 — Findable by display name: pickers, search, sort (#1642)**: the
+  picker surface class carries the slug **inline** — `<option>`s render
+  `Display name (slug)` via `agentOptionLabel` (else the bare slug), and the
+  `<option>` **value stays the slug** so filtering/selection never keys on the
+  label. Six dropdowns: `ExecutionsPanel`, `ReportsPanelFleet`, operator
+  `QueueList` + `NotificationsPanel`, `FileManager`, `Settings` (subscription
+  assignment). `Agents.vue` name search matches **both** the slug and the
+  display name (case-insensitive) — otherwise typing "TOM" against a
+  `tom-marketing-ops` slug returns nothing. **Sort-key decision (AC):** the
+  "Name (A-Z / Z-A)" sort orders by the **display name when set, else the slug**
+  (`agentDisplayName`, in the store's `_getSortedAgents`) — sorting by the slug
+  while the row renders the label would order the list by an invisible key. Every
+  per-agent lookup (`getActivityState`/tags/stats/router actions) still keys on
+  `agent.name`; only the option label, the search predicate, and the sort
+  comparator changed. No store-shape change — the label is resolved off the
+  loaded agents (FR-6 resolvers), so `agentNames`/`availableAgents` stay
+  slug-string arrays.
 
 ### 1.4 Agent Deletion
 - **Status**: ✅ Implemented
--- a/src/frontend/src/components/ExecutionsPanel.vue
+++ b/src/frontend/src/components/ExecutionsPanel.vue
@@ -70,7 +70,7 @@
               : 'border border-gray-200 dark:border-gray-600 hover:border-gray-300 dark:hover:border-gray-500'"
           >
             <option value="">All agents</option>
-            <option v-for="name in agentNames" :key="name" :value="name">{{ name }}</option>
+            <option v-for="name in agentNames" :key="name" :value="name">{{ agentOptionLabel(agentsStore.agentRefForSlug(name)) }}</option>
           </select>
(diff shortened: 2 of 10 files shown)


vybe (approved):
Validated via /validate-pr: verified every touched component already binds `agentsStore` and imports the resolver (incl. `agentDisplayName` in stores/agents.js and Agents.vue), `<option>` values stay the slug, and Operations.vue fetches agents on mount so the operator-panel labels resolve. Deviation from AC#1 (resolve via agentRefForSlug instead of changing the store shape) is documented in requirements/core-agent.md FR-7. All checks green.

Issue #1641, opened 2026-07-15
feat: Render agent display name on the nine roomy UI surfaces (#964 follow-up 3/5)


vybe:
## Summary

Render the display name with the slug beneath it on the nine UI surfaces that have room for a two-line treatment, and retire the template-derived fallback in favour of the #964 precedence chain. This is where users first see the feature.

Follow-up 3 of 5 from the design decision on #964.

## Context

Of ~67 agent-name render sites, only nine are roomy enough for "display name primary, slug secondary". Three of them (`OverviewPanel.vue`, `InfoPanel.vue`, and the public chat header via `src/backend/routers/public.py`) already render a template-derived friendly name — this issue switches them onto the durable precedence chain so the old, inconsistent path retires rather than lingering alongside the new one.

The core constraint from #964: **when a display name is set, the slug stays visible** so it's always clear which agent you're looking at.

## Acceptance Criteria

- [ ] Two-line treatment (display name primary, slug secondary) on all nine roomy surfaces:
  - [ ] `src/frontend/src/components/AgentHeader.vue` (badge row directly below is the natural slug slot)
  - [ ] `src/frontend/src/components/OverviewPanel.vue`
  - [ ] `src/frontend/src/components/AgentTile.vue` (reuse the existing `.t-repo` secondary line)
  - [ ] `src/frontend/src/components/portal/PortalBriefing.vue`
  - [ ] `src/frontend/src/components/portal/PortalSidebar.vue`
  - [ ] `src/frontend/src/components/portal/PortalConversation.vue`
  - [ ] `src/frontend/src/views/enterprise/PortalChat.vue`
  - [ ] `src/frontend/src/views/enterprise/ClientPortal.vue`
  - [ ] `src/frontend/src/views/MobileAdmin.vue`
- [ ] When no display name is set, the slug is promoted to primary and there is **no dangling empty secondary line**
- [ ] `OverviewPanel.vue` / `InfoPanel.vue` / public chat header read the precedence chain (DB → template → slug) rather than template-only
- [ ] `src/backend/routers/public.py` reads the column instead of falling back through container labels; the friendly name now renders when the agent is **stopped**
- [ ] `src/backend/routers/voice.py` stops synthesising a name via `agent_name.replace("-", " ").title()`
- [ ] `AgentAvatar` still receives the **slug** — verified that setting a display name does not change any agent's avatar colour or initials
- [ ] Dark mode + truncation verified on every surface
- [ ] Frontend e2e snapshots updated

[...]

PR #1700, opened 2026-07-20
fix(ui): Dashboard Timeline tile leads with the display name, slug beneath (#1641)

dolho:
## What

The Dashboard **Timeline** view's compact agent tiles were naming agents by the **slug** (e.g. `a2a-demo`), with the display name only in the hover tooltip. Spotted live: an agent labelled "Zeta Ops" still read `a2a-demo` on its tile.

That was the dense-table treatment #1643 applied when it bucketed `ReplayTimeline` with the operational rows. But this tile is **roomy** — `lg` avatar, three rows, autonomy toggle, capacity meter, 72px tall — so by the #964 render rule it's a roomy surface: **lead with the display name, keep the slug visible as secondary text.**

## Change

`ReplayTimeline.vue` name cell → a tight two-line stack:
- **display name** on the name line (`agentDisplayName`)
- the **slug** as a small muted line directly beneath — rendered **only when a distinct label is set** (`hasDistinctLabel`), so an unlabelled agent is byte-identical to before (slug on one line, no second row).

The hover tooltip still carries `label · slug`. `row.name` (the slug) still drives the avatar, `navigateToAgent`, and `toggleAutonomy` — display-only change.

## Verified on the live instance

With `a2a-demo` labelled "Zeta Ops" and `slack-tester` labelled "TOM Marketing":

```
a2a-demo      → primary "Zeta Ops"        slug line "a2a-demo"
slack-tester  → primary "TOM Marketing"   slug line "slack-tester"
cornelius     → primary "cornelius"       (no second line — unlabelled)
```

SFC compiles via `@vue/compiler-sfc`; behavior confirmed against the real fleet from `/api/agents`.

## Scope

Partially addresses **#1641** (roomy surfaces, #964 follow-up 3/5) and corrects the `ReplayTimeline` classification from #1643. Kept off the #1642 PR (pickers/search/sort) since it's a different surface class.

Related to #1641

### DIFF
--- a/src/frontend/src/components/ReplayTimeline.vue
+++ b/src/frontend/src/components/ReplayTimeline.vue
@@ -163,12 +163,12 @@
                     <span
                       class="text-sm font-semibold text-gray-900 dark:text-white truncate cursor-pointer hover:text-blue-600 dark:hover:text-blue-400"
                       @click="navigateToAgent(row.name)"
-                      :title="agentNameTooltip(agentsStore.agentRefForSlug(row.name))"
+                      :title="agentNameTooltip({ name: row.name, display_label: row.displayLabel })"
                     >
-                      {{ agentDisplayName(agentsStore.agentRefForSlug(row.name)) }}
+                      {{ agentDisplayName({ name: row.name, display_label: row.displayLabel }) }}
                     </span>
                     <span
-                      v-if="hasDistinctLabel(agentsStore.agentRefForSlug(row.name))"
+                      v-if="hasDistinctLabel({ name: row.name, display_label: row.displayLabel })"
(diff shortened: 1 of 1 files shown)


dolho:
image

vybe (approved):
Validated via /validate-pr: resolvers imported, and the label is sourced off the networkStore agent/nodeData the tile is already built from (`nodeData` is in scope in the same computed) rather than the agents store the Dashboard never populates — that sourcing is the load-bearing detail here. Slug stays visible as the secondary line and avatar/navigate/toggle still key on `row.name`. All checks green.

Note: intentionally left without a `Fixes` keyword — #1641 covers nine roomy surfaces and this fixes the Timeline tile only, so the issue should stay open rather than auto-promote.

Issue #1704, opened 2026-07-20
bug: agent plugin install-state is lost on container recreate (~/.claude.json is gitignored)


dolho:
## Summary

Claude Code records which marketplace plugins an agent has installed in `~/.claude.json`. That file is **gitignored** (it's in `git_service._GITIGNORE_PATTERNS`, and correctly so — it also holds session/runtime state and secrets). It is also **not** part of the durable home volume in any recreate-surviving way. So when an agent container is **recreated** (config-drift recreate, image/template change, sub-switch), the agent **silently loses its installed plugins** — both the `~/.claude/plugins/` cache and the `~/.claude.json` manifest that records them are gone.

## Context / how this surfaced

Split out from #1702 (the open question in its acceptance criteria). #1702 adds `.claude/plugins/` to the fleet-wide gitignore so the plugin **cache** stops getting auto-committed into agent repos — a re-installable cache does not belong in git. But that surfaced the deeper question: if the cache is (correctly) ignored, **how does an agent's plugin selection survive a recreate?**

Today it doesn't. A user runs `/plugin install trinity@abilityai` (or `/trinity:onboard`), the agent works, then a routine recreate wipes the plugins with no signal.

This matters because Trinity **actively steers agents onto plugins** — `CLAUDE.md` recommends `/plugin marketplace add abilityai/abilities`, and the `trinity` plugin is meant to be installed by agents.

## What NOT to do

Do **not** un-ignore `~/.claude/plugins/` or `~/.claude.json` to "fix" persistence:
- the cache is a re-installable copy (bloat, #1702);
- `~/.claude.json` carries session state and secrets and must stay out of git.

## Possible directions (design needed)

- A small **declared plugin manifest** the agent owns (e.g. `~/.trinity/plugins.yaml` — installed marketplace + plugin@version list), committed like other agent config, with a startup hook that re-runs `/plugin install` for anything listed but not present. Mirrors the "agent owns its declared state" pattern (`data_paths` #1169, persistent-state S4).
- OR persist `~/.claude/plugins/` + the plugin section of `~/.claude.json` on the durable volume across recreate (heavier; re-introduces the bloat/secret concerns).

## Acceptance criteria

- [ ] Decide the mechanism (declared manifest + re-install hook vs. volume persistence), recorded on this issue
- [ ] An agent's installed plugins survive a container recreate
- [ ] No plugin cache or `~/.claude.json` gets committed to the agent repo (keep #1702's gitignore intact)

## Technical notes

[...]

PR #1705, opened 2026-07-20
fix(git-sync): ignore .claude/plugins/ so plugin caches aren't auto-committed (#1702)

### LINKED ISSUE #1702: bug: .claude/plugins/ is missing from the fleet-wide gitignore — installed plugin caches get auto-committed into agent repos
## Summary

`_GITIGNORE_PATTERNS` in `src/backend/services/git_service.py` ignores six `.claude/*` runtime directories under an explicit policy comment — *"Claude Code runtime — commit commands/skills/agents, exclude runtime data"* — covering `.claude/projects/`, `.claude/statsig/`, `.claude/todos/`, `.claude/debug/`, `.claude/sessions/`, and `.claude/shell-snapshots/`. **`.claude/plugins/` is absent from that list.**

Claude Code installs marketplace plugins into `~/.claude/plugins/cache/<plugin>@<version>/` (a full copy of each plugin: skills, agents, hooks). Because a Trinity agent's `HOME` **is** its git repo root (`ENV HOME=/home/developer`, `WORKDIR=/home/developer`, repo at the same path), that cache lands inside the agent's git working tree and is auto-committed by the 15-minute sync loop — the same repo-bloat class as #1596, just via a different writer.

## Context

Found while reviewing the platform skill-injection work (abilityai/trinity-enterprise#183), which had to add per-skill `.gitignore` lines and untracking for exactly this reason: anything the platform or the agent writes under `~/.claude/` is inside the repo by construction.

This is a path users are actively steered onto — `CLAUDE.md` recommends `/plugin marketplace add abilityai/abilities` and the `trinity` plugin is meant to be installed *by agents* (`/trinity:onboard`). Every plugin install is a fresh directory tree committed to the agent's GitHub repo, and every plugin update commits another copy.

## Acceptance Criteria

- [ ] `.claude/plugins/` added to `_GITIGNORE_PATTERNS` alongside the existing Claude Code runtime entries
- [ ] Existing agents pick it up through the idempotent gitignore merge on next sync, and already-committed matches are untracked via the `git rm --cached` path added in #1596 (stops future churn; does not shrink existing history)
- [ ] Verified against a live agent with a plugin installed: the plugin cache no longer appears in `git status`, and the plugin still loads afterwards
- [ ] **Open question resolved before closing**: does plugin *install state* need a committed manifest to survive a container recreate? Claude Code records installed plugins in `~/.claude.json`, which is **already** gitignored — so a recreated agent may silently lose its plugins today. If that is the case, track it as its own issue rather than un-ignoring the cache (a re-installable cache does not belong in git either way)

## Technical Notes

- `src/backend/services/git_service.py` — `_GITIGNORE_PATTERNS`, the "Claude Code runtime" block; merge helper `_build_gitignore_append_command`; untrack pattern established by #1596
- Sibling issues in the same bloat class: #1595 (auto-gc can never complete in-container), #1596 (auto-committed data churn)
- Root cause behind the whole category is the agent layout (repo root == `$HOME`) — tracked separately

dolho:
## What

`_GITIGNORE_PATTERNS` in `git_service.py` ignored six `.claude/*` runtime dirs but **not `.claude/plugins/`**. Claude Code installs marketplace plugins into `~/.claude/plugins/cache/<plugin>@<version>/` (a full copy — skills, agents, hooks). Since an agent's `HOME` **is** its git repo root, that cache lands in the working tree and the 15-min sync loop auto-commits it — and every plugin update commits another copy. Same repo-bloat class as #1596, different writer, on a path users are steered onto (`/plugin marketplace add abilityai/abilities`, `/trinity:onboard`).

## Change

- Added `.claude/plugins/` to the "Claude Code runtime" block of `_GITIGNORE_PATTERNS`.
- Existing agents pick it up through the idempotent gitignore merge on the next Push, and any already-committed cache is untracked via the `git rm --cached` step `_migrate_workspace_gitignore` already runs (#1596) — stops future churn, doesn't shrink history.
- Kept the gitignore block in `docs/TRINITY_COMPATIBLE_AGENT_GUIDE.md` in sync — the `test_doc_and_constant_in_sync` guard fails CI if any `_GITIGNORE_PATTERNS` entry is missing from the doc.

## Open question (AC) — resolved

> does plugin *install state* need a committed manifest to survive a container recreate?

**Yes, and it's lost today.** `~/.claude.json` (the plugin manifest) is *also* gitignored, so a recreated agent silently loses its plugins. The right resolution per the AC: **keep the cache ignored** (it's re-installable and doesn't belong in git either way) and **track the persistence gap separately** → filed as **#1704**. This PR does not un-ignore anything.

## Verification

- `.claude/plugins/` is present in both the `_GITIGNORE_PATTERNS` tuple **and** the doc's `gitignore` block → `test_doc_and_constant_in_sync` passes (46 patterns, none missing).
- `git check-ignore` in a real repo confirms `.claude/plugins/cache/abilities@1.0/skills/x.md` is **ignored**.
- The compat static check `S-006` uses a hardcoded subset that doesn't include `plugins`, so its `_GOOD_GITIGNORE` fixture is unaffected (no test churn); the auto-fix still adds the pattern since `fixes.py` reuses `_GITIGNORE_PATTERNS`.
- `git_service.py` parses.

_(No live agent currently has a plugin installed, so the "plugin still loads" half of the AC's live check wasn't reproducible here — but the ignore semantics are proven and the merge/untrack mechanism is the same one #1596 runs in production for the sibling `.claude/*` patterns.)_

Related to #1702 · follow-up #1704

Fixes #1702

### DIFF
--- a/docs/TRINITY_COMPATIBLE_AGENT_GUIDE.md
+++ b/docs/TRINITY_COMPATIBLE_AGENT_GUIDE.md
@@ -190,6 +190,7 @@ __pycache__/
 .claude/debug/
 .claude/sessions/
 .claude/shell-snapshots/
+.claude/plugins/
 # Keep: .claude/commands/, .claude/skills/, .claude/agents/, settings.local.json
 
 # Temporary files
--- a/src/backend/services/git_service.py
+++ b/src/backend/services/git_service.py
@@ -924,6 +924,12 @@ _GITIGNORE_PATTERNS: Tuple[str, ...] = (
     ".claude/debug/",
     ".claude/sessions/",
     ".claude/shell-snapshots/",
+    # Marketplace plugin caches (#1702): Claude Code copies each installed
+    # plugin (skills/agents/hooks) into ~/.claude/plugins/cache/<plugin>@<ver>/.
+    # Since HOME == the agent's repo root, that lands in the working tree and
+    # the 15-min sync loop commits it (and every plugin update commits another
+    # copy) — repo bloat, same class as #1596. Re-installable, so never in git.
+    ".claude/plugins/",
     # Temporary files
     "*.log",
     "*.tmp",


vybe (approved):
Validated via /validate-pr: one-line addition to the fleet-wide gitignore + the matching doc block, which the existing test_github_init_gitignore.py parity test enforces as a pair. All checks green.

Issue #1709, opened 2026-07-20
bug: retention blast-radius guard has no in-product approval path — agent purge and #1581 volume reclaim can never run


vybe:
## Summary

`#1644` shipped the retention blast-radius guard with `FLOOR_AGENTS = 0` — *"every purge destroys Docker volumes. Always ack."* — so **every** agent hard-purge, even a single one, exceeds the threshold and is refused until an admin acknowledgement exists.

The only way to record that acknowledgement is `POST /api/settings/retention/acknowledge`. That endpoint has **no frontend caller and no MCP tool**: `src/frontend/src/views/Settings.vue` wires only the read-only `GET /api/settings/retention`.

Net effect: no operator can approve, the guard always refuses, the `#834` agent purge sweep never completes — and `#1581`'s Docker-volume reclaim, which shipped in the *same* release, can never run. Volumes still leak forever, now for a different reason.

## Context

Found by `/release-plan` during the v0.8.5 pre-freeze coherence review (2026-07-20).

The guard's design is correct and should not change: destroying agent data volumes is irreversible, so a human gate is right, and `FLOOR_AGENTS = 0` is a deliberate "always ask". This issue is only about the **missing approval surface** — the guard is fail-safe (nothing is wrongly deleted), but it makes a shipped fix inert and the release cannot honestly claim `#1581` without it.

## Design decision to settle first

`GET /api/settings/retention` currently reports the effective windows plus `guard.max_rows` — it does **not** report that a prune is pending approval, so the UI has nothing to render a prompt from. Two options:

1. **Extend the GET** to report pending refusals (setting key, window in force, candidate count, last-refused timestamp).
2. **Read the existing alarm** — the guard already raises an operator-queue item on the reserved `_retention-guard` sentinel with counts and identifiers.

Option 1 keeps the Settings panel self-contained and avoids coupling a settings surface to the operator queue; option 2 adds no endpoint surface. Pick one before building.

## Acceptance Criteria

- [ ] An admin can approve a pending retention prune from Settings → Retention without leaving the UI
- [ ] The confirmation names **what** will be deleted: the setting key, the window in force, and the candidate count — approving blind is not acceptable for an irreversible volume deletion
- [ ] The action is **admin-only and human-only** — it must reject an agent-scoped principal (`reject_agent_principal`), since an agent key resolves to its owner carrying the owner's role and the default install is admin-owned
[...]

Issue #1310, opened 2026-06-22
refactor(auth): consolidate duplicated auth/dependency patterns behind factories (INV-8, split from #654)


AndriiPasternak31:
Split from #654 (architecture-invariant drift). #654 now tracks **INV-14** (model centralization, done in #1308); this issue carves out **INV-8** because it is **security-sensitive** and needs new abstractions + a security review, not a mechanical move.

## Invariant
**#8 — Auth Pattern**: every authenticated endpoint uses `Depends(get_current_user)`; agent-scoped endpoints use `AuthorizedAgent` / `OwnedAgentByName`; role-gated endpoints use `require_role(...)` / `require_admin`. Duplicated/inline auth wiring should collapse behind these shared dependencies (threshold: ≤5 bespoke sites).

## Ground truth (re-measured 2026-06-22)
The 2026-05-05 automated report undercounted ("51 sites / 19 files"). Actual:

- **64 sites across 22 files** (threshold ≤5).
- Breakdown: **~27 mechanical** (swap to the existing dependency), **~22 need a *new* auth factory** (no current shared helper fits), **~15 inline** checks to fold in.
- `slack.py` carries **8** sites — **overlaps the channel-adapter area (Pavlo)**; coordinate before touching.

## Why this is its own issue (security-relevant)
- ~22 sites need **new** auth-factory design — getting the boundary wrong is a privilege/authz risk.
- Requires **behavioral verification + a security pass** (CSO-style): confirm each consolidated site enforces the *same* owner/admin/role/agent-scope check it did before. INV-8 is the security-relevant invariant of the original drift report; do **not** fold it into a mechanical PR.

## Suggested approach
1. Design the missing auth factories first (cover the ~22 gap cases).
2. Migrate the ~27 mechanical sites + ~15 inline checks.
3. Coordinate `slack.py` with the channel owner.
4. CSO/security review of the consolidated boundaries.

Refs: parent #654, sibling #1309 (INV-7), completed INV-14 PR #1308.

AndriiPasternak31:
/claim — picking this up as part of the reliability/security open-ticket push.

PR #1715, opened 2026-07-20
fix(retention): in-product approval path for the blast-radius guard (#1709)

dolho:
## What

The `#1644` blast-radius guard ships with `FLOOR_AGENTS = 0` — every agent hard-purge is refused until an admin acknowledges via `POST /api/settings/retention/acknowledge`. That path was **doubly dead**, so `#834`'s purge sweep and `#1581`'s volume reclaim (same release) could never run:

1. **No UI caller.** `Settings.vue` only read `GET /retention` — no admin could ever approve.
2. **The endpoint itself 500'd.** It called an *unimported* `require_admin` (a FastAPI `Depends` factory, not an imperative helper) — a `NameError` the `#1310` auth-wiring refactor introduced. **Found by testing live**, not by reading. Fixed to `assert_admin(current_user)` (the imported helper the rest of the file uses).

## Design decision (the issue asked to settle GET vs operator-queue)

**Extend the GET, computed by re-running the guard live.** It reports `pending_acknowledgements` — `{key, label, window_days, candidate_count, floor}` — for the two low-floor irreversible sweeps (agents floor 0, schedules floor 100). Freshest possible, no stale/"approved" state, **no coupling** a settings surface to the operator queue, and naturally single-use (an acked or pruned sweep returns `allowed=True` and drops off). Each check is bounded to `floor+1` rows.

## Frontend (Settings → Retention)

A prominent amber block lists each pending prune, **names exactly what will be deleted** (label + count + window + key), and offers **"Approve deletion"** → `POST /acknowledge {key, window_days}`. Window-bound (409 surfaced as a readable inline error), single-use (item clears after approval + refetch), **honest empty state** when nothing's pending. Admin+human gate stays server-side.

## AC coverage

| AC | Status |
|----|--------|
| Approve a pending prune from Settings without leaving the UI | ✅ |
| Confirmation names key + window + candidate count | ✅ |
| Admin-only **and** human-only (`reject_agent_principal`) | ✅ (server-side; and fixed the 500 that blocked it) |
| Window sent matches force; mismatch → readable **409** | ✅ |
| Single-use; UI re-arms, no stale "approved" | ✅ (pending recomputed live) |
| Honest empty state with a next action | ✅ |
| With #1581, an approved purge removes the agent's volumes | ✅ (purge path now reachable — see live verification) |
| Approval audit-logged with acting admin | ✅ (already server-side; now actually reachable) |
| Test fails if the ack endpoint loses its UI caller | ✅ `test_1709_retention_ack_ui.py` |

## Verified end-to-end on a live stack

Inserted a soft-deleted `agent_ownership` row past the 180-day window → `GET` surfaced it (`candidate_count: 1`, label names the volume loss) → wrong window `999` → **409** → correct window `180` → **200**, pending **cleared** → residue removed. (No real agent touched.)

## Tests
- `test_1709_retention_ack_ui.py` — static guard: fails if `Settings.vue` drops the acknowledge call or the pending render.
- `test_retention_floor.py` — pending surfaces when over threshold, drops off when acked, skips disabled windows, empty when clear; the ack endpoint is **callable** (NameError regression) and **409s** on window mismatch.

_(Validated live end-to-end; unit tests parse and mirror that behavior — CI runs them, host has no pytest.)_

Related to #1709 · #1644 · #1581 · #834

### DIFF
--- a/src/backend/routers/settings.py
+++ b/src/backend/routers/settings.py
@@ -217,12 +217,12 @@ async def acknowledge_retention_prune(
     to approve a mass deletion of their own audit trail, and it lives in a table
     one of the guarded sweeps prunes.
 
-    Human-only. `require_admin` alone is NOT sufficient today: an agent-scoped MCP
+    Human-only. Admin-role alone is NOT sufficient today: an agent-scoped MCP
     key resolves to its owner *carrying the owner's role*, so on an install whose
     agents are admin-owned (the default — see cornelius_agent_service.CORNELIUS_OWNER)
-    an agent key passes `require_admin`. `reject_agent_principal` is therefore
-    applied explicitly here rather than relied upon from `require_admin`. See
-    abilityai/trinity-ops-agent#232 for the underlying fix.
+    an agent key passes the admin check. `reject_agent_principal` is therefore
+    applied explicitly here. See abilityai/trinity-ops-agent#232 for the
+    underlying fix.
 
     The ack is bound to `window_days`: approving a prune at 30 days does not
     approve one at 1 day. It is single-use — `cleanup_service` consumes it once the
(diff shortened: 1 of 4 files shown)


dolho:
image

vybe (approved):
Validated: fixes a real live bug — settings.py:236 calls require_admin() which is NOT imported (only assert_admin is), a NameError that 500s the retention-ack endpoint on dev today, leaving #834's purge + #1581's volume reclaim unreachable. Fix (assert_admin) + reachable UI approve path + regression tests (test_retention_floor asserts the endpoint is callable & 409s on window mismatch). All referenced symbols (FLOOR_AGENTS/FLOOR_SCHEDULES/evaluate, db.count_soft_deleted_*_past_retention) verified on dev. No schema change. Required checks green; pytest head-seed matrix green.

Issue #1712, opened 2026-07-20
bug: per-agent circuit-breaker toggle still reports success while the global gate is off (#1487 residual)


vybe:
## Summary

`#1487` had two halves. The packaging half **is fixed** — `DISPATCH_BREAKER_ENABLED` is now forwarded by `docker-compose.yml`, `docker-compose.prod.yml`, and `.env.example`, so the global gate is enableable on a standard deployment.

The honesty half is not. `config.py:143` still defaults to `false`, and `PUT /api/agents/{name}/circuit-breaker` accepts an owner's toggle, persists `agent_ownership.circuit_breaker_enabled`, and returns success — while the breaker never engages, because the two-tier gate requires the global flag too. The owner gets a control that reports success and does nothing, with no indication why.

`.env.example` now states this outright: *"Default false — with this off, the owner-facing UI toggle silently no-ops."* That is a known-false affordance shipped deliberately.

## Context

Found by `/release-plan` during the v0.8.5 pre-freeze coherence review (2026-07-20). Filed separately because `#1487` will close with this release and would otherwise take the residual with it.

This is a Product Quality Bar item — **honest status** and **graceful fallback when a flag or dependency is absent**. The two-tier gate itself is correct design (fleet-wide kill switch over a per-agent opt-in); the gap is that only one tier is visible to the person operating the other.

## Acceptance Criteria

- [ ] When `DISPATCH_BREAKER_ENABLED` is false, the per-agent circuit-breaker toggle communicates that the feature is globally disabled — it does not silently accept and report success
- [ ] The UI names the **reason** and the **remedy** (the global flag, and that an admin must set it) rather than only disabling the control
- [ ] `GET /api/agents/{name}/circuit-breaker` already returns `config.global_enabled` — the frontend consumes it rather than inferring state
- [ ] An owner who has toggled it on while the global flag is off sees honest state on reload, not a persisted-but-inert "on"
- [ ] Admins retain a discoverable path to the global setting
- [ ] No change to the gating semantics — both flags are still required for the breaker to engage

## Open question

Whether the global gate should stay `false` by default. It is already two-tier (per-agent `circuit_breaker_enabled` defaults to 0), so flipping the global default to `true` would change nothing for agents that have not opted in, and would remove the confusing state entirely. Worth deciding here rather than carrying a permanently-false default plus UI to explain it.

## Technical Notes

- Gate: `src/backend/config.py:143`
[...]

Issue #1491, opened 2026-07-06
PUT /api/agents/{name}/circuit-breaker gives no signal when the global DISPATCH_BREAKER_ENABLED gate is off


AndriiPasternak31:
### Context
Follow-up split out of #1487 (compose-forwarding of `DISPATCH_BREAKER_ENABLED`, landed in the #1485 config-hygiene PR).

The dispatch circuit breaker is two-tier gated: it engages only when **both** the global `DISPATCH_BREAKER_ENABLED` env flag **and** the per-agent `agent_ownership.circuit_breaker_enabled` toggle are on (see architecture.md → *Circuit Breakers*, and `services/dispatch_breaker.py`).

### Problem
An owner can `PUT /api/agents/{name}/circuit-breaker {"enabled": true}` and get a `200` with `circuit_breaker_enabled: true`, while the breaker still does nothing because the global flag is off. The response already surfaces `config.global_enabled` on `GET`, but the `PUT` gives the owner no indication their toggle is inert — the classic "the switch is on but nothing happens" trap. #1487 made the global flag *reachable*; it didn't close the UX gap.

### Proposed fix
On the `PUT` handler (`routers/agents.py`, ~L400–411), when `enabled=true` is set while `DISPATCH_BREAKER_ENABLED` is false, include a non-fatal advisory in the response body (e.g. `"warning": "Per-agent toggle saved, but the breaker stays inactive until the platform-wide DISPATCH_BREAKER_ENABLED flag is enabled."`) and/or surface it in the owner UI. Keep the write succeeding — the stored preference is still valid for when the global flag flips on.

### Acceptance
- `PUT .../circuit-breaker {"enabled": true}` with the global flag off returns `200` + an advisory field flagging the inert state.
- No behavior change when the global flag is on.
- Owner-facing surface (UI toggle) reflects the advisory.

### Non-goals
No change to breaker logic or the two-tier gate itself — advisory only.

Issue #1487, opened 2026-07-06
DISPATCH_BREAKER_ENABLED global gate unwired — dispatch circuit breaker (#526) un-enableable, UI toggle silently no-ops


vybe:
## Summary

The global gate for the per-agent dispatch circuit breaker (RELIABILITY-007, #526) cannot be enabled through supported configuration. `DISPATCH_BREAKER_ENABLED` is read at `src/backend/config.py:143` (default `false`) but appears in **neither** compose file **nor** `.env.example`. Since the breaker engages only when the per-agent `circuit_breaker_enabled` flag AND this global flag are both on, the owner-facing toggle (`PUT /api/agents/{name}/circuit-breaker`) silently does nothing on every standard deployment.

## Context

Found by the 2026-07-06 automated `/validate-config` run (umbrella: #1485). Same packaging-gap class as #1039/#1056 — but worse than a tuning knob: an entire shipped reliability feature is un-enableable, and the UI presents a toggle that has no effect.

```python
# src/backend/config.py:143
DISPATCH_BREAKER_ENABLED = os.getenv("DISPATCH_BREAKER_ENABLED", "false").lower() == "true"
```

Gating documented in `docs/memory/architecture.md` (Circuit Breakers, #526): "per-agent `circuit_breaker_enabled` (default OFF) AND global `DISPATCH_BREAKER_ENABLED` must both be on."

## Acceptance Criteria

- [ ] `DISPATCH_BREAKER_ENABLED` is forwarded in the backend `environment:` block of **both** `docker-compose.yml` and `docker-compose.prod.yml` (default `false`)
- [ ] A documented entry is added to `.env.example` explaining the two-tier gating (global flag + per-agent toggle)
- [ ] With `DISPATCH_BREAKER_ENABLED=true` in `.env` and the per-agent toggle on, `GET /api/agents/{name}/circuit-breaker` reports `config.global_enabled: true`

## Technical Notes

- Read site: `src/backend/config.py:143`; consumed by `services/dispatch_breaker.py` and `services/capacity_manager.py`
- Consider whether the per-agent PUT should surface a warning when the global flag is off (currently the toggle succeeds silently) — optional follow-up, not required for this fix

PR #1717, opened 2026-07-20
fix(reliability): honest per-agent circuit-breaker toggle in Settings (#1712)

dolho:
## What

The per-agent dispatch breaker is two-tier — per-agent `circuit_breaker_enabled` **and** the platform-wide `DISPATCH_BREAKER_ENABLED` must both be on. The endpoint already returns `config.global_enabled` (GET) and a `warning` (PUT) when a toggle is inert (#1491/#1499). The residual #1712 gap was the **UI**: there was no owner-facing toggle at all (a growth-path TODO in `SettingsPanel`), so the "control that reports success and does nothing" had no honest surface.

## Change — new `ReliabilityPanel.vue` (agent Settings tab)

- **Consumes `config.global_enabled`** from the endpoint (doesn't infer) → honest state on every load, including a persisted-but-inert "on".
- **Global flag OFF → names reason + remedy:** an amber notice explains the platform-wide breaker is off and that **an admin must set `DISPATCH_BREAKER_ENABLED=true`** in the environment. The toggle stays operable (owner may pre-set the preference) with a **"saved · inactive"** badge — it does *not* silently accept-and-succeed.
- **On save while global-off**, surfaces the endpoint's `warning` as the outcome (not a plain "Saved"). When both flags are on → **"active"** badge.

No change to gating semantics — `config.py` untouched, both flags still required.

## Open question (global default) — decided

Keep `DISPATCH_BREAKER_ENABLED=false` for this release. Flipping it to `true` would start engaging the breaker for any **already-opted-in** agent mid-freeze — a runtime behaviour change needing soak. The honest UI removes the confusion without that risk. Revisit post-freeze. (Recorded on #1712.)

## AC coverage

| AC | Status |
|----|--------|
| Global-off → toggle communicates it's globally disabled (no silent success) | ✅ |
| UI names the reason + remedy (not only disabled) | ✅ |
| Frontend consumes `config.global_enabled` | ✅ |
| Honest state on reload (no persisted-but-inert "on") | ✅ |
| Admins keep a discoverable path to the global setting | ✅ (names the env flag + who sets it) |
| No change to gating semantics | ✅ |

## Verified live

On this install `DISPATCH_BREAKER_ENABLED` is off → `GET .../circuit-breaker` returns `config.global_enabled:false`; `PUT {enabled:true}` returns the inert `warning`, which the panel surfaces. Static guards (`test_1712_*`) fail if the panel drops the `global_enabled` read, the warning surface, the reason/remedy copy, or its mount. Both SFCs compile.

Related to #1712 · #1487 · #526

### DIFF
--- a/.github/workflows/backend-unit-test.yml
+++ b/.github/workflows/backend-unit-test.yml
@@ -56,7 +56,12 @@ jobs:
     # Per-PR diff job — no base/head exists on a push, so skip there.
     if: github.event_name != 'push'
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    # The suite runs ~7-8 min on a normal runner, but GitHub's shared runners
+    # vary ~2-2.5x; a slow one was hitting the old 15-min cap at ~78% done and
+    # getting killed (a spurious red — the tests were all passing). 25 gives real
+    # headroom for the slow tail without masking a genuine hang (pytest-timeout
+    # still bounds any single test).
+    timeout-minutes: 25
     strategy:
       fail-fast: false
       matrix:
--- a/src/frontend/src/components/ReliabilityPanel.vue
+++ b/src/frontend/src/components/ReliabilityPanel.vue
@@ -1,0 +1,130 @@
+<template>
+  <div class="p-6">
+    <h3 class="text-lg font-medium text-gray-900 dark:text-white mb-2">Reliability — dispatch circuit breaker</h3>
+    <p class="text-sm text-gray-500 dark:text-gray-400 mb-4">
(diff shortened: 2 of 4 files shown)


vybe (approved):
Validated: frontend-only (ReliabilityPanel.vue + SettingsPanel.vue) — consumes config.global_enabled for an honest per-agent breaker toggle, no gating-semantics change (config.py untouched). CI timeout bump 15m→25m is justified (spurious runner-variance kills). Static guard test_1712 present. Required checks green; pytest head-seed matrix green.

PR #1708, opened 2026-07-20
feat(install): agent-driven one-shot install — unattended start.sh + agent runbook (ent#39)

dolho:
Implements the local install slice of **Abilityai/trinity-enterprise#39** (one-line install + agent-guided onboarding). OSS-core — per the issue's own note, "the OSS installer must work standalone," so this lands in the public repo. First PR of the epic; the two follow-ups are noted at the bottom.

## What

Make the install path **agent-runnable and one-shot** with no interactive hard-stops in the happy path, verified serving before it claims success, ending on a friendly next-steps card — plus a canonical **agent-facing runbook** so any capable agent can drive and narrate the whole thing.

### `scripts/deploy/start.sh` (the engineering core)
- **Unattended mode** (`TRINITY_UNATTENDED=1` / `--unattended`): a blank `ADMIN_PASSWORD` is generated (strong, 24-char) and surfaced in the summary instead of the interactive `exit 1`. Interactive default keeps fail-fast (now pointing at `--unattended`), so a human is never handed a password they didn't note.
- **Pre-flight**, one consolidated fail-fast: Docker daemon + Compose v2 (hard); ports 80/8000/8080/6379 reported as a warning (a re-run legitimately holds them). Portable probe via `bash /dev/tcp`.
- **Serving verification**: replaced the fixed `sleep 5` with a bounded poll of the backend `/health` (up only after migrations + lifespan). Timeout → warning, never a crash.
- **Next-steps card**: Web UI URL, login (+ the generated admin password when unattended), the `abilities` marketplace + `trinity:onboard` commands, a UI create-agent pointer, and a model-API-key warning when none is set.

### Docs
- **`docs/AGENT_INSTALL_GUIDE.md`** — deterministic runbook written *for the installing agent*: pre-flight → unattended install → independent serving check → friendly landing, with a "how to communicate" section (confirm prerequisites, surface secrets once, never fabricate a model key, end on a card). At a **stable raw-GitHub URL**, declared the single source of truth kept in sync with the scripts.
- **README** promotes the agent-driven one-shot path as the recommended entry point + documents `--unattended`.

## AC coverage

| AC | This PR |
|----|---------|
| Copy-pasteable "ask your agent to install" prompt | ✅ (README + guide) |
| Non-interactive install, no hard-stops; secrets surfaced | ✅ |
| Pre-flight (Docker/Compose/ports) with remediation | ✅ |
| Verify actually serving before success | ✅ |
| Friendly next-steps card | ✅ (terminal) |
[...]

PR #1719, opened 2026-07-21
ci: raise backend-unit-test pytest timeout 15m → 25m (stop spurious runner-variance cancels)

dolho:
## What

Raises the `backend-unit-test.yml` pytest matrix job's `timeout-minutes` **15 → 25**.

## Why

The suite runs **~7-8 min** on a normal GitHub runner, but the shared runners vary **~2-2.5×**. A slow one sits at ~78% of the suite (all tests passing — pure dots in the log) when it hits the 15-min cap and gets **killed**. That surfaces as a red `pytest (…, seed …) fail` **plus** a `regression diff` fail — even though the diff itself reports **"✅ No new failures."**

Seen **three times this cycle** on PRs that can't possibly break pytest:
- **#1700** — Timeline tile (frontend-only)
- **#1712** — circuit-breaker toggle (frontend + one static file-reading test)
- **#1708** — installer/docs (zero Python)

Each needed a manual empty-commit/branch patch. This fixes it fleet-wide so no PR flakes on runner variance.

## Safety

25 min covers the observed ~19-min worst case with margin, and does **not** mask a genuine hang — `pytest-timeout` still bounds any single test. One-line, workflow-only change.

### DIFF
--- a/.github/workflows/backend-unit-test.yml
+++ b/.github/workflows/backend-unit-test.yml
@@ -56,7 +56,12 @@ jobs:
     # Per-PR diff job — no base/head exists on a push, so skip there.
     if: github.event_name != 'push'
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    # The suite runs ~7-8 min on a normal runner, but GitHub's shared runners
+    # vary ~2-2.5x; a slow one was hitting the old 15-min cap at ~78% done and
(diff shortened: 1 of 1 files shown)


vybe (approved):
Validated via /validate-pr: required checks green, security scan clean, scope focused.

Issue #1713, opened 2026-07-20
bug: src/scheduler/utils.py declares byte-parity vendoring with no test enforcing it (and the claim is already false)


vybe:
## Summary

`src/scheduler/utils.py` (added by `#1474`) declares a byte-parity contract that nothing enforces:

> *"CANONICAL SOURCE: src/backend/utils/helpers.py … `utc_now_iso` / `to_utc_iso` are vendored here byte-for-byte from the backend — edit the backend copy and regenerate this mirror. Same 'regenerate-from-backend' discipline as `failure_classifier.py`."*

Two problems. First, **no parity test exists** — the repo's three other declared mirrors each have one (`tests/unit/test_credential_paths_parity.py`, `tests/unit/test_model_context_parity.py`, and the `failure_classifier` parity check in `tests/unit/test_904_sigkill_no_false_auth.py`). Second, the claim is **already false as written**: `to_utc_iso` is functionally identical but textually different (the backend uses an `if`/`else` with comments; the mirror uses an early return), so "regenerate and diff" cannot mechanically verify it today.

## Context

Found by `/release-plan` during the v0.8.5 pre-freeze coherence review (2026-07-20).

The standalone scheduler cannot import from the backend package, so vendoring is the right call — but an unenforced vendoring claim is worse than none, because it invites the assumption that a guard exists. A future edit to `utc_now_iso`/`to_utc_iso` on the backend side silently desyncs the scheduler's timestamp format, which is precisely the `#1474` bug class: naive timestamp strings that JS `new Date(...)` parses as local time, so every non-UTC browser renders schedule timestamps shifted.

Trinity has already been bitten by mirrored-constant drift — see `docs/memory/learnings.md` on mirrored constants where the mirror's own tests pinned the drift.

## Acceptance Criteria

- [ ] A test fails when the backend and scheduler implementations diverge behaviorally
- [ ] The test asserts **agreement on output**, not source-text equality: both `utc_now_iso()` produce the same format (Z-suffixed, `T` separator), and both `to_utc_iso` agree across naive, aware-UTC, and aware-non-UTC inputs
- [ ] The docstring is corrected to match what is actually enforced — "behavioral mirror" if the test is behavioral, "byte-for-byte" only if a literal source comparison is enforced
- [ ] The same treatment is applied or explicitly ruled out for `src/scheduler/failure_classifier.py`, whose parity check lives in an unrelated test file and is easy to miss
[...]

Issue #1474, opened 2026-07-06
bug: schedule-triggered executions store naive started_at (no timezone suffix) — UI shows wrong relative time


vybe:
## Summary

Schedule-triggered executions store `started_at` as a **naive** ISO timestamp (no `Z` suffix), while backend-created executions store Z-suffixed UTC. JavaScript parses the naive form as *local* time, so in any non-UTC browser the agent Tasks list shows schedule-triggered rows shifted older by the viewer's UTC offset — a dispatcher and its self-task workers that started 42 seconds apart appear hours apart.

## Context

Operator report from a production instance (deployed at `3d2538c0`), 2026-07-06:

| Execution | Actual start (UTC) | UI shows (UTC+1 browser) |
|---|---|---|
| Dispatcher (`triggered_by: schedule`) | 11:00:00 | **"2h ago"** ❌ |
| 3 workers (`triggered_by: self_task`) | 11:00:23–11:00:42 | "33–34m ago" ✅ |

All four rows started within 42 seconds of each other; the schedule-triggered one appears ~1.6h older. This misled an operator into believing the dispatcher and its self-tasks ran hours apart.

Serialization evidence (same row is mixed — created by the scheduler, finalized by the backend):

```
# scheduler-created execution — naive, no timezone suffix:
"started_at":   "2026-07-06T11:00:00.207634"
# same row, completed by the backend — Z-suffixed:
"completed_at": "2026-07-06T11:00:42.219882Z"
```

`agent_schedules.last_run_at` is also naive (`"2026-07-06T11:00:00.616516"`).

## Root Cause

Two layers:

**Write side — the standalone scheduler serializes naive timestamps.** Cron-fired executions are created by the scheduler container, whose raw-SQL DB layer uses `datetime.utcnow().isoformat()` (no `Z`) instead of the backend's ISO-Z convention:

- `src/scheduler/database.py` `create_execution()` (~L389) — the reported `started_at`
- also `create_skipped_execution` (~L450), scheduler-side `complete_execution` (~L514–525), `update_schedule_run_times` (~L351/354 — why `agent_schedules.last_run_at`/`next_run_at` are naive too), and the process-schedule variants (~L798–1019)

[...]

PR #1720, opened 2026-07-21
test(scheduler): behavioral parity guard for vendored timestamp helpers (#1713)

dolho:
## What

`src/scheduler/utils.py` (#1474) declared **byte-for-byte** vendoring of `utc_now_iso` / `to_utc_iso` from `src/backend/utils/helpers.py` with **no test** — and the claim was **already false**: `to_utc_iso` is functionally identical but textually different (backend `if/else` + comments; the mirror an early return), so "regenerate and diff" can't verify it. An *unenforced* vendoring claim is worse than none — it invites the assumption a guard exists, and a future backend edit silently re-introduces the #1474 bug class (naive strings a JS `new Date(...)` renders shifted by the viewer's tz).

## Change

- **`tests/unit/test_1713_scheduler_utils_parity.py`** — asserts agreement on **output**, not source text (prior art: `test_model_context_parity`):
  - both `utc_now_iso()` emit the `Z`-suffixed, `T`-separated ISO shape;
  - both `to_utc_iso` agree across **naive, aware-UTC, and aware-non-UTC** inputs (all the same instant → same `Z` string);
  - loads each copy in isolation via `spec_from_file_location` (the scheduler can't import the backend package — the whole reason vendoring exists);
  - pins `parse_scheduler_ts` to **naive UTC** on purpose — the deliberate read-side asymmetry #1474 needs for `aware − naive` duration math — so a parity refactor can't accidentally couple it to the backend's *aware* parse.
- **Docstring corrected**: "behavioral mirror", not "byte-for-byte", and it now names the enforcing test (discoverable from the mirrored file).

## `failure_classifier.py` — explicitly ruled out

Its two copies **are** genuinely byte-identical, its byte-diff test (`test_904_sigkill_no_false_auth.py::TestBackendSchedulerParity`) is valid and already enforced + referenced from that file. Only `utils.py` needed a behavioral guard — the byte approach is *correct* for `failure_classifier` (identical) and *impossible* for `utils.py` (textually divergent). The docstring now contrasts the two so the difference isn't a trap.

## AC coverage

| AC | Status |
|----|--------|
| Test fails when the two implementations diverge behaviorally | ✅ |
| Asserts agreement on **output**, not source-text; covers naive/aware-UTC/aware-non-UTC | ✅ |
| Docstring matches what's enforced ("behavioral mirror") | ✅ |
| Same treatment applied or ruled out for `failure_classifier.py` | ✅ ruled out (genuinely byte-identical, already guarded) |
| Test discoverable from the mirrored file | ✅ (named in the docstring) |
| Doesn't pin `parse_scheduler_ts` to aware | ✅ (asserts it stays **naive** UTC) |

Verified: all parity assertions pass against the current files.

Related to #1713 · #1474

### DIFF
--- a/src/scheduler/utils.py
+++ b/src/scheduler/utils.py
@@ -4,9 +4,14 @@ Timezone-aware timestamp helpers for the standalone scheduler (#1474).
 CANONICAL SOURCE: src/backend/utils/helpers.py. The scheduler is a *separate*
 package (own pyproject.toml / Dockerfile that copies only src/scheduler) and
 cannot import src/backend at runtime, so ``utc_now_iso`` / ``to_utc_iso`` are
-vendored here byte-for-byte from the backend — edit the backend copy and
-regenerate this mirror. Same "regenerate-from-backend" discipline as
-``failure_classifier.py``.
+vendored here. This is a **behavioral** mirror, not a byte copy: it must AGREE
+ON OUTPUT with the backend (same Z-suffixed ISO format), and ``to_utc_iso`` is
+functionally identical but written with an early return rather than the
+backend's if/else — so "regenerate and diff" cannot mechanically verify it.
+Edit the backend copy and keep the *outputs* in sync; the contract is enforced
+by ``tests/unit/test_1713_scheduler_utils_parity.py`` (#1713). (Contrast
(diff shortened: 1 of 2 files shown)


vybe (approved):
Validated via /validate-pr: required checks green, security scan clean, scope focused.

Issue #1714, opened 2026-07-20
bug: bulk watchdog sweeps emit no task-completion event — swept executions never wake their subscriber (#1578 residual)


vybe:
## Summary

`#1578` made the backend emit `agent.task.completed` / `agent.task.failed` at every **CAS-won** execution terminal, so a subscribed orchestrator is woken with a report-back task instead of polling.

Bulk watchdog sweeps are the documented exception. They terminate executions with a `COUNT`-style bulk update and carry no per-row context, so they emit **no** completion event. An execution reaped by a bulk sweep therefore never wakes its subscriber — the caller waits until it times out on its own.

This was named as a residual in the `#1578` work and tracked nowhere.

## Context

Found by `/release-plan` during the v0.8.5 pre-freeze coherence review (2026-07-20).

Low impact today: the individually-handled terminal writers all emit correctly (`apply_result` success and failure branches, `_write_terminal_and_gate` for timeout/budget/crash, the `#1083` lease-reaper, and the pull sink). Bulk sweeps are the rare path. The impact rises with `#1081` pull coordination, where re-delivery and lease expiry make bulk reclamation more common, and where waking the caller is the whole point of the async-first design.

## Acceptance Criteria

- [ ] A bulk-swept execution emits the same `agent.task.failed` event an individually-reaped one does, or the gap is closed by routing bulk sweeps through a per-row terminal path
- [ ] The emit stays **matching-subscription gated** — no `agent_events` row and no dispatch when nothing subscribes, so a large sweep with no subscribers costs nothing
- [ ] A sweep reclaiming many rows does not produce a thundering herd — the `#1085` governor primitives apply, or the emit is batched/paced
- [ ] The three-layer loop safety from `#1578` still holds: reserved `agent.task.*` namespace, no reserved self-subscription, and the `X-Event-Trigger` recursion break
- [ ] Emission remains fail-open — an event failure never blocks or reverts the sweep's terminal write
- [ ] If the residual is instead accepted, it is documented in `docs/memory/architecture.md` under Task Completion Events as a deliberate, bounded exception rather than an unstated one

## Technical Notes

- Emit helpers: `src/backend/services/event_dispatch_service.py` → `emit_task_terminal_event` / `spawn_task_terminal_event`
- Covered terminal writers today: `task_execution_service.apply_result`, `_write_terminal_and_gate`, `cleanup_service._process_stale_slot_reclaims`, `pull_coordination_service.apply_task_result`
[...]

Issue #1578, opened 2026-07-11
feat: system-emitted agent.task.completed/failed events at execution terminal (async caller report-back)


vybe:
## Summary

When a caller (e.g. an orchestrator agent) dispatches a long-running task via `chat_with_agent(parallel=true, async=true)` — or receives a `queued_timeout` receipt from the 25s sync abort (#914) — the only way to learn the outcome is polling `get_execution_result`. Nothing fires when an execution reaches terminal: `apply_result` writes the row, releases the slot, and broadcasts UI-only WS telemetry. `docs/planning/TARGET_ARCHITECTURE.md` (§Async-First Communication) already specifies the missing half: *"Caller subscribes to `agent.task.completed` / `agent.task.failed` events via the event bus, or polls `get_execution_result`."* This issue builds those system-emitted completion events.

## Context

- Concrete failure mode: an orchestrator agent dispatches a long task to a worker agent; the orchestrator's turn ends; when the worker finishes, nothing wakes the orchestrator, so nobody reports back to the human.
- EVT-001 event pub/sub today carries only **agent-emitted** events (`emit_event`) — a completion callback depends on the worker's LLM remembering to emit at the end of its turn, which is not reliable and not enforceable.
- The #946 pull pilot reroutes sequential agent→agent calls to async receipts and explicitly notes "the backend emits no completion event — polling is the contract." This issue upgrades that contract in the direction the target architecture already commits to, making fire-and-park orchestration (#1402 authoring contract) practical.

## Proposal

Fire a synthetic completion event at the single terminal chokepoint — `task_execution_service.apply_result`, on the **CAS-won branch only** (replayed/late terminals never double-fire):

- Source agent = the executing agent; event type `agent.task.completed` or `agent.task.failed` by terminal status.
- Payload: `{execution_id, status, triggered_by, summary_or_error, duration_ms, cost}` — supports the existing `{{payload.field}}` template interpolation.
- Delivery reuses the existing EVT-001 subscription dispatch (`_trigger_subscription` → async task enqueued to the subscriber) — the same "reply lands in the caller's queue" shape the pull model wants.
- Emit only when a matching subscription exists (`find_matching_event_subscriptions` first) — zero subscriptions ⇒ zero event rows, no fleet-wide spam.

## Acceptance Criteria

[...]

PR #1722, opened 2026-07-21
fix(events): bulk watchdog sweeps emit task-completion events (#1714)

dolho:
## #1714 — bulk watchdog sweeps emit no task-completion event (the #1578 residual)

`#1578` made every **individually-reaped** terminal writer emit `agent.task.completed`/`.failed` so a subscribed orchestrator is woken with a report-back task. **Bulk watchdog sweeps were the documented exception:** `mark_stale_executions_failed` / `mark_no_session_executions_failed` fail many rows with a COUNT-style bulk write and no per-row context → a bulk-swept execution emitted nothing and **never woke its subscriber** — the caller waited out its own timeout. Impact rises with `#1081` pull coordination (bulk reclamation becomes common, and waking the caller is the whole point of async-first).

## Fix — route the bulk sweeps through the same per-row emit

- **db** (`db/schedules/cleanup.py`): both bulk-fail fns take an optional `collect_failed` list and append the **CAS-won** `(execution_id, agent_name)` rows (gated on `result.rowcount` — only rows this sweep actually won, never one a real terminal beat it to). **The int-count return is unchanged**, so every existing caller/test is untouched — no signature break.
- **service** (`cleanup_service.py`): the two sweeps collect those rows; `_emit_bulk_terminal_events` emits `agent.task.failed` per row via the shared `spawn_task_terminal_event` → `emit_task_terminal_event` helper.

## AC coverage

| AC | How |
|----|-----|
| Bulk-swept row emits the same `agent.task.failed` | ✅ per CAS-won row through the shared helper |
| Matching-sub gated; **free when nobody listens** | ✅ cheap `db.has_task_terminal_subscribers()` short-circuits the whole loop; the helper still does the precise per-agent match |
| No thundering herd | ✅ emits **paced in batches** (50, yield between) — burst bounded, **no row dropped** (a hard cap would reintroduce the miss for the overflow) |
| #1578 loop safety holds | ✅ reserved `agent.task.*` namespace, no reserved self-sub, `X-Event-Trigger` recursion break — all inherited from the shared helper (suppresses a `triggered_by="event"` origin) |
| Fail-open | ✅ best-effort; the FAILED terminal is already committed — an emit failure never blocks/reverts it |
| Residual documented as covered | ✅ `architecture.md` + `task-completion-events.md` updated from "documented residual" → "covered" |

## Tests
`tests/unit/test_1714_bulk_sweep_completion_events.py` (13): DB collect (CAS-won only, count contract preserved, fresh row not collected), the subscriber gate (enabled/disabled), the service emit (empty→noop, no-subs→noop, per-row FAILED args, missing-id skip, fail-open, large-batch pacing). Existing cleanup + #1578 suites: **77 passing**, no regression.

**Live-verified** on the Postgres stack end-to-end: a bulk no-session sweep of a stale execution for a subscribed source agent emitted a real `agent_events` `agent.task.failed` row (`subscriptions_triggered=1`).

Related to #1714 · #1578 · raises-impact #1081


vybe (approved):
Validated via /validate-pr: required checks green, security scan clean, named regression test for #1714 present.

PR #1727, opened 2026-07-21
Fix dashboard header overlap at narrow desktop widths

erichroepke:
## Summary

- allow the dashboard header stats and controls to wrap cleanly
- make host telemetry compact below 1100px by hiding sparklines and shrinking the disk bar
- keep the fix scoped to dashboard/header layout

## Root cause

The compact dashboard header forced the status/telemetry row and controls into one horizontal line. At narrow desktop widths, fixed-width telemetry elements competed with the filter/view controls and caused crowding/overlap.

## Validation

- `npm run check:tokens`
- `npm run build`

Note: Docker was unavailable in this session, so the final production build was run locally after installing frontend dependencies from the lockfile.

### DIFF
--- a/src/frontend/src/components/HostTelemetry.vue
+++ b/src/frontend/src/components/HostTelemetry.vue
@@ -149,14 +149,17 @@ onUnmounted(() => {
 .host-telemetry {
   display: inline-flex;
   align-items: center;
+  flex-wrap: wrap;
   gap: 12px;
   font-size: 12px;
+  min-width: 0;
 }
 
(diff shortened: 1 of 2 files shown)


erichroepke:
Superseded by #1756, rebuilt from current `dev` and expanded to cover the complete narrow-viewport Dashboard regression tracked in #1754. Retargeting this older `main`-based draft would expose an unrelated 27-file diff, so I am closing it in favor of the clean replacement.

PR #1734, opened 2026-07-22
feat(portal): agent avatar on chat-history rows (ent#186)

dolho:
Related to Abilityai/trinity-enterprise#186 (frontend half; the backend title-generation half is Abilityai/trinity-enterprise#194).

## What

Client-portal history rows and cross-chat search results identified their agent with a 2px color dot, while the Agents section directly above rendered full `PortalAvatar` components. A client with several agents had to hover for the tooltip to learn whose conversation a thread was.

## How

Both lists now render `PortalAvatar` at `size=22` — the same component (avatar-URL-with-initials-fallback) already used in four other portal places.

- The avatar URL is resolved **from the roster already loaded at sign-in** — an `avatarByAgent` computed map keyed by `agent_name`, built from the existing `roster` prop. **No additional request per row**, and no change to the sessions payload.
- An agent with no generated avatar falls through to `PortalAvatar`'s existing initials + deterministic tint — no new empty state.
- Search rows move the snippet into the title column so history and search share one layout (previously an `ml-4`-indented sibling that assumed the dot's width).
- `agentColor` is no longer imported directly by the sidebar (`PortalAvatar` still uses it internally for the tint).

## Acceptance criteria

- [x] History rows show the conversation's agent avatar instead of the color dot
- [x] Search-result rows use the same treatment
- [x] Initials + color fallback works when the agent has no avatar
- [x] Row height / truncation hold at the sidebar's narrow width, on mobile, in dark mode — `min-w-0` + `truncate` retained on both lines; `PortalAvatar` is `shrink-0`; no new colors (the avatar carries its own tint, dark-mode-agnostic)
- [x] No additional network request per row

## Verification

`@vue/compiler-sfc` parse + `compileScript` + `compileTemplate` on the changed SFC → compiles clean. (A full `npm run build` fails on this machine for an unrelated reason — `mermaid` is in `package.json` but missing from the local `node_modules`, breaking `AgentWorkspace.vue`; untouched by this PR.)

### DIFF
--- a/src/frontend/src/components/portal/PortalAvatar.vue
+++ b/src/frontend/src/components/portal/PortalAvatar.vue
@@ -1,11 +1,11 @@
 <template>
   <span
     class="inline-flex shrink-0 items-center justify-center rounded-full overflow-hidden text-white font-semibold select-none"
-    :style="{ width: size + 'px', height: size + 'px', background: imgOk ? 'transparent' : tint, fontSize: Math.round(size * 0.4) + 'px' }"
+    :style="{ width: size + 'px', height: size + 'px', background: showImage ? 'transparent' : tint, fontSize: Math.round(size * 0.4) + 'px' }"
     :title="name"
   >
     <img
-      v-if="avatarUrl && imgOk"
+      v-if="showImage"
       :src="avatarUrl"
       :alt="name"
       class="w-full h-full object-cover"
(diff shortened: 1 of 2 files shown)


vybe (approved):
Validated via /validate-pr: small focused UI change, reuses PortalAvatar + existing roster (no new requests), security clean. Cross-tracker ent#186 stays open (titles half pending) — noting for manual status bump.

PR #1735, opened 2026-07-22
feat(skills): skill-runner execution surface + close the raw library read (ent#139)

dolho:
Related to Abilityai/trinity-enterprise#139 (OSS half; the gated module is Abilityai/trinity-enterprise#195).

## 1. Security fix — close the unrestricted library read

`GET /api/skills/library/{name}` returns a skill's **full content** — SKILL.md plus any bundled `scripts/`, i.e. executable material — behind a bare `get_current_user`. An agent-scoped key resolves to its owner, so **any agent could pull arbitrary executable content out of the library and run it locally**: self-acquisition with no chokepoint.

Now calls `reject_agent_principal`. This is edition-agnostic and stands on its own, which is why it lives in OSS rather than behind the entitlement. **Listing stays open** (name + description) so agents can still discover what exists — only the content fetch is closed.

⚠️ **Behavior change:** an agent-scoped MCP key calling `get_skill` now gets 403. Agents get skills two supported ways instead — assigned/injected by an operator, or executed on the runner under a per-skill allow-list. Neither needs raw content over REST.

## 2. MCP tools

`run_skill(skill_name, input)` and `list_runnable_skills` — structurally `run_playbook` (`tools/connector.ts`) pointed at the library. Both auto-register from the existing tool group; no `server.ts` change.

The tool-side allow-list check is **not authorization** — it exists only to turn a predictable refusal into a message listing what IS available. The backend re-checks the ACL at dispatch, and a failed pre-check falls through to the backend rather than failing open into "assume permitted". On an OSS build (no module mounted) both tools return an honest `enabled: false` / "not available on this platform" instead of throwing a stack trace at the agent.

## 3. Grid tile — first agent-class variant

`AgentTile.vue` keyed off `agent.type === 'skill-runner'` (the `trinity.agent-type` label), so:

- no new per-agent field on the agents payload;
- it **degrades to the standard tile automatically** — with the module unmounted no agent carries that type.

Adds a `SKILL RUNNER` identity chip (same shape as `.sys-badge`), a teal avatar ring, and an exposed-skill chip (or a warn chip when the runner exists but execution is off).

The ring uses a CSS var (`--gv-ring-runner`, defined per-theme in `FleetGrid.vue`) rather than a Tailwind class — **there is no teal token in `tailwind.config.js`, and an invented class name renders no border at all, silently**. Light and dark values both defined.

Status rides the Grid's **existing** visibility-aware `refreshBatchData` — no new polling loop — and the endpoint is admin-only, so a non-admin viewer keeps `null` and gets the standard tile.

Deliberately scoped as an in-tile variant rather than a new component, so it does not pre-empt epic Abilityai/trinity-enterprise#94's tile-variant architecture.

## Verification

- `npx tsc --noEmit` on the MCP server → clean.
- `@vue/compiler-sfc` parse + `compileScript` + `compileTemplate` on `AgentTile.vue` and `FleetGrid.vue` → clean. (A full `npm run build` fails on this machine for an unrelated pre-existing reason: `mermaid` is in `package.json` but missing from the local `node_modules`, breaking `AgentWorkspace.vue`.)
- `pytest tests/unit -k "skill or models_centralized or auth_wiring or enumeration"` → **138 passed, 3 skipped**.
- Live: `GET /api/skills/library` still 200 for an admin JWT.

## Not done

The Grid tile has **not been seen rendered** — no skill-runner agent exists on this stack (the module refuses to provision without a configured skills library), so the variant is verified by compilation and code inspection only, not visually.

### DIFF
--- a/src/backend/routers/skills.py
+++ b/src/backend/routers/skills.py
@@ -15,7 +15,13 @@ from typing import List
 from fastapi import APIRouter, Depends, HTTPException
 
 from models import User
-from dependencies import get_current_user, require_admin, get_authorized_agent_by_name, get_owned_agent_by_name
+from dependencies import (
+    get_current_user,
+    require_admin,
+    reject_agent_principal,
+    get_authorized_agent_by_name,
+    get_owned_agent_by_name,
+)
 from database import db
 from db_models import AgentSkill, SkillInfo, AgentSkillsUpdate
 from services.skill_service import skill_service, SkillInjectionBusy
@@ -72,7 +78,20 @@ async def get_skill(
 ):
     """
     Get details for a specific skill including full content.
+
+    **Human/operator surface only** (ent#139). This returns the skill's full
+    content — SKILL.md plus any bundled `scripts/` — so it is executable
+    material, not metadata. An agent-scoped key resolves to its owner user, so
+    before this gate any agent could pull arbitrary executable content out of
+    the library and run it locally: self-acquisition, which is exactly the
+    supply-chain risk the skill runner exists to avoid.
+
+    Agents get skills two supported ways instead: assigned/injected by an
+    operator, or executed on the dedicated runner under a per-skill allow-list.
+    Neither needs raw content over REST. Listing (name + description) stays open
+    so an agent can still *discover* what exists.
     """
+    reject_agent_principal(current_user)
     skill = skill_service.get_skill(skill_name)
     if not skill:
         raise HTTPException(status_code=404, detail=f"Skill '{skill_name}' not found")
--- a/src/frontend/src/components/AgentTile.vue
+++ b/src/frontend/src/components/AgentTile.vue
@@ -1,10 +1,10 @@
 <template>
-  <div class="gtile" :class="{ system: isSystemAgent }">
+  <div class="gtile" :class="{ system: isSystemAgent, runner: isSkillRunner }">
     <!-- Avatar half-out on the left edge -->
     <div class="gtile-avatar">
       <div
         class="rounded-full border-2 shadow-md overflow-hidden"
-        :class="isSystemAgent ? 'border-accent-purple-400 dark:border-accent-purple-500' : 'border-action-primary-400 dark:border-action-primary-500'"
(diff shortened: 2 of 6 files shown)


dolho:
Live-tested against a real instance alongside Abilityai/trinity-enterprise#195.

**Verified live:**
- `GET /api/skills/library/{name}` — agent-scoped key **403**, admin **200**. The gating works and does not disturb operators.
- Listing (`GET /api/skills/library`) still **200** for an agent key, so discovery is preserved as intended.
- **MCP end-to-end** over real JSON-RPC with an agent's own `TRINITY_MCP_API_KEY`: after rebuilding the MCP image, tool count 97 → **102**, both `run_skill` and `list_runnable_skills` advertised. `list_runnable_skills` returned exactly the caller's permitted set (1 of 2 library skills); `run_skill("word-count")` returned a real result with cost and `execution_id`; `run_skill("secret-skill")` returned the structured refusal with `available: ["word-count"]`.
- Grid tile **data** confirmed: the runner reports `type: "skill-runner"` on `/api/agents`, and the status endpoint yields the expected badge/chip/tooltip values.

**Fixed after live check (b56b319d):** the chip rendered "1 skills exposed" — pluralized.

**Not verified:**
- **The tile was never seen rendered.** Data + compile only — no browser tooling available in this session. Worth a visual pass before merge, particularly the teal ring/badge in dark mode.
- `get_skill` remains *advertised* as an MCP tool to agent-scoped keys even though it now always 403s for them. Harmless but slightly untidy; hiding it from agent keys would need a `canAccess` change I did not want to make unasked.

**Note for reviewers:** the MCP server is a baked image (no bind mount), so `npm run build` on the host is not enough — `docker compose build mcp-server` is required to pick these tools up.

vybe (approved):
Validated: OSS half of ent#139 (standalone security fix closing the agent-key library-content read + graceful-fallback MCP tools + degrading grid tile). Enterprise half merged with review fixes (single-slot pin). Earlier CI failure was the base-side hung-job flake; rerun green.

PR #1721, opened 2026-07-21
feat(telemetry): local product-event capture — activation funnel, Tier-1 (ent#184)

dolho:
## ent#184 — Tier-1 local product events (OSS capture + gated view)

**Gating shape (confirmed):** OSS-core **capture** + entitlement-gated **funnel view**. This is the public-repo half; the enterprise funnel backend ships in `trinity-enterprise` (`feature/184-telemetry-funnel`).

### Capture — OSS-core, default-ON, **zero network egress**
- **`product_events`** table (`installation_id`, `event_type`, `event_context`, `created_at`) — `db/schema.py` + `db/tables.py` MetaData + **dual-track migration** (SQLite `product_events_table` + Alembic `0029_product_events`).
- **`POST /api/product-events`** — writes ONE local row and returns; never phones home. `event_type` checked against a **fixed allow-list** (unknown → 422) so the table can't be spammed; `context` byte-capped.
- **`OnboardingWizard.vue`** emits its step transitions (`setup_started` → `setup_step_create` → `setup_step_credential` → `setup_completed`/`setup_dismissed`) via `stores/productTelemetry.js` — fire-and-forget, never blocks/breaks the wizard.
- **First-value events** (`first_agent_created`, `first_chat`, …) are **derived on read** by the enterprise view from `audit_log`/`agent_activities`/`agent_schedules` — never re-emitted, so they survive restart by construction.
- `db/product_events.py` — `ProductEventOperations` (record + count-by-type + chronological list for Tier-2 backfill + optional prune), wired into the facade.

### Funnel view — ships in the OSS bundle, hidden unless entitled
- `components/settings/ActivationFunnelPanel.vue` — step funnel + drop-off % + first-value tiles + **honest empty state**; fetches the gated enterprise endpoint.
- `Settings.vue` gains an admin-only **Activation** tab, `requires: 'telemetry'` → invisible in OSS-only builds (standard feature-flag gating).

### Docs
- Requirements **§45** (`lifecycle-observability.md`) — generic seam only (no enterprise module design in public docs).
- User-facing `docs/PRODUCT_EVENTS.md` — what's recorded, that it never leaves the box, how to view.

### AC coverage
- [x] Wizard step transitions + first-value events recorded/derived locally, default-ON, survive restart
- [x] Operator-facing funnel with step counts / drop-off + honest empty state *(view lands with the enterprise PR)*
- [x] **Zero network egress** — the emit endpoint writes one local row; documented local-only
[...]

PR #1739, opened 2026-07-22
fix(seam): scope the enterprise ImportError arm to the top-level import (#1653)

### LINKED ISSUE #1653: bug: main.py's enterprise `except ImportError` reports a real registration bug as "submodule not present — this is normal"
## Summary

`main.py`'s enterprise-registration `try` block catches `ImportError` **first**, and `ModuleNotFoundError` subclasses `ImportError`. So a **real bug** inside `register_enterprise(app)` — any enterprise module lazily importing an OSS seam that doesn't exist — is reported as *"this is normal"*, and the diagnostic handler never runs.

Surfaced while reviewing trinity-enterprise#161 (credit: @vybe, who spotted it and offered to file).

## Current code (`src/backend/main.py:998`)

```python
try:
    from enterprise.backend import register_enterprise
    register_enterprise(app)
    print("Trinity Enterprise modules registered", flush=True)
except ImportError:
    print(
        "Trinity Enterprise submodule not present — OSS-only build "
        "(this is normal; enterprise modules are an optional private submodule)",
        flush=True,
    )
except Exception as e:
    # A BUG in enterprise registration ... must NOT take down the core platform
    ...
```

The `except ImportError` arm is meant for **one** thing: the top-level `from enterprise.backend import register_enterprise` failing on an OSS-only clone. But it wraps `register_enterprise(app)` too, so it also swallows any `ModuleNotFoundError` raised *deep inside* module registration.

## Why it matters

Enterprise modules lazily import OSS seams inside `register()` — established pattern (`two_factor/__init__.py:41` does `from services import mfa_gate`; the A2A module does `from services import a2a_gate`). If that seam is missing — e.g. the OSS PR shipping it hasn't merged, or a submodule pointer is bumped onto a branch without it — the failure mode on a **mounted enterprise install** is:

- prints `"Trinity Enterprise submodule not present — OSS-only build (this is normal)"` — **actively false**, and the most misleading message available;
- the `except Exception` traceback that would name the real cause **never prints**;
- boot does not crash; earlier-registered modules stay live, so the install looks healthy;
- the failing module's router may still be mounted while `register_module(...)` never ran → **every one of its routes 403s forever**, with no diagnostic anywhere.

An operator sees "this is normal" and a feature that 403s. There is nothing in the logs to connect the two.

## Suggested fix

Narrow the `ImportError` arm to only the case it's for — the top-level import:

```python
try:
    from enterprise.backend import register_enterprise  # type: ignore[import-not-found]
except ImportError:
    print("Trinity Enterprise submodule not present — OSS-only build ...", flush=True)
else:
    try:
        register_enterprise(app)
        print("Trinity Enterprise modules registered", flush=True)
    except Exception as e:
        # real bug -> loud, with traceback; still doesn't take down core
        ...
```

A `ModuleNotFoundError` from inside `register()` then hits the diagnostic arm and prints a traceback naming the missing seam.

## Acceptance Criteria

- [ ] A `ModuleNotFoundError` raised **inside** `register_enterprise(app)` is reported as a bug (traceback, module named) — not as "submodule not present"
- [ ] A genuinely absent submodule (OSS-only clone) still prints the calm "this is normal" line and boots clean — no regression to the OSS path
- [ ] Core boot still survives a broken enterprise module (the existing degrade-don't-crash contract holds)
- [ ] Regression test for both branches: absent submodule → normal message; register() raising ModuleNotFoundError → diagnostic

## Notes

- `.github/workflows/*` greps stdout for `"Trinity Enterprise modules registered"`, and the `backend boots without enterprise submodule` job asserts the OSS-only path — worth checking neither string assertion breaks.
- Low priority in the sense that nothing is broken today; high leverage in that it converts a silent, misleading failure into an obvious one for **every** future enterprise module.

dolho:
Fixes #1653.

Related to Abilityai/trinity-enterprise#196 (OSS half; the enterprise half — per-module registration isolation — is Abilityai/trinity-enterprise#197).

## The bug

The enterprise load block caught bare `ImportError` and printed:

```
Trinity Enterprise submodule not present — OSS-only build
(this is normal; enterprise modules are an optional private submodule)
```

Correct for an OSS clone. But `ImportError` also covers a **mounted** submodule whose module failed to import — e.g. importing a name from an OSS table that hasn't landed yet.

Observed live: the submodule was present and correctly mounted, a module failed, and the operator was told **"this is normal"** while a paid feature was silently absent. Diagnosing it required hand-running `register_enterprise()` in the container to see the real traceback.

## The fix

Narrow the benign branch to its actual signature — a `ModuleNotFoundError` whose missing module **is** the `enterprise` package:

| Situation | Exception | Branch |
|---|---|---|
| OSS clone, no submodule | `ModuleNotFoundError(name="enterprise.backend")` | benign ✓ |
| Module imports a missing **name** | plain `ImportError` (not a `ModuleNotFoundError`) | **loud** ✓ |
| Module needs an uninstalled package | `ModuleNotFoundError(name="pyotp")` | **loud** ✓ |
| Anything else (schema, migration, mount) | `Exception` | **loud** ✓ (unchanged) |

The loud handler was already correct and is only factored into `_report_enterprise_failure()` so both paths share it. No behavior change for a genuine OSS build.

## Test-pin update

`tests/unit/test_847_entitlement_seam.py` statically pinned the literal `from .sso import register as register_sso` + `register_sso(app)` in the **enterprise** `__init__.py`. The paired enterprise PR registers each module through an isolating helper, so that exact pair no longer exists.

The pin moves to the **intent** — "the submodule registers `sso`" — and still accepts the pre-ent#196 form, so this cross-repo static check doesn't break on a behavior-preserving refactor in the other repo.

## Verification

New `tests/unit/test_ent196_enterprise_load_reporting.py` (4 passed, 1 skipped — the skip is the "no submodule" case, which can't be exercised in a checkout that has one). Includes a static guard that fails if a bare `except ImportError` ever comes back in front of the benign message.

Full OSS unit suite: **4667 passed, 16 skipped, 5 failed**. All 5 failures reproduce on a clean `origin/dev` worktree and are unrelated:
- 3× `test_admin_email_login` — `AttributeError: module 'bcrypt' has no attribute '__about__'` (local bcrypt/passlib version mismatch)
- `test_1474_read_boundary_z::test_schedules_summary_last_run_at_normalized`
- `test_1472_schedule_validation::test_accepts_valid[0 4 * * *-Europe/Kiev]`

Enterprise-docs guard run locally against `docs/`, `CLAUDE.md`, `main.py`, `entitlement_service.py` — clean (the new comment describes the generic seam and names no module).

Verified live: with the enterprise submodule mounted and one module genuinely broken, the log now reads `Trinity Enterprise registration completed with 1 FAILED module(s): …` instead of the reassuring line.

### DIFF
--- a/src/backend/main.py
+++ b/src/backend/main.py
@@ -1014,22 +1014,19 @@ def _report_enterprise_failure(exc: BaseException) -> None:
 
 try:
     from enterprise.backend import register_enterprise  # type: ignore[import-not-found]
-    register_enterprise(app)
-    # `print(..., flush=True)`: this import block runs at module init,
-    # which is BEFORE `lifespan` calls `setup_logging()`. The default
-    # Python logger drops INFO-level records, so `logger.info` here
-    # would be silently swallowed. Print to stdout instead — docker
-    # logs captures it for ops + the CI workflow greps for it.
-    print("Trinity Enterprise modules registered", flush=True)
 except ModuleNotFoundError as e:
-    # "Not present" is ONLY true when the enterprise package itself is missing.
-    # A bare `except ImportError` here also swallowed a *failed* import inside a
-    # mounted submodule — e.g. a module importing a name that doesn't exist yet
-    # — and reported it as the reassuring line below. An operator then saw
-    # "this is normal" while a feature they paid for was silently absent, and
-    # diagnosing it needed a hand-run `register_enterprise()` to see the real
-    # traceback. Narrow the benign case to its actual signature; everything
-    # else falls through to the loud handler.
+    # SCOPE: this arm covers ONLY the top-level import. It used to wrap
+    # `register_enterprise(app)` too, so a ModuleNotFoundError raised deep inside
+    # a module's `register()` — enterprise modules lazily import OSS seams there
+    # — was reported as the calm line below. On a MOUNTED install that message is
+    # actively false, the diagnostic traceback never printed, and the failing
+    # module's routes could stay mounted-but-unentitled, 403-ing forever with
+    # nothing in the logs connecting the two (#1653, spotted by @vybe).
+    #
+    # Even here "not present" is only true when the missing module IS the
+    # enterprise package. If `enterprise.backend` imports some third-party
+    # package that isn't installed, that is a real failure wearing the same
+    # exception type — so check which module went missing.
     if (e.name or "").split(".")[0] == "enterprise":
         print(
             "Trinity Enterprise submodule not present — OSS-only build "
@@ -1038,8 +1035,23 @@ except ModuleNotFoundError as e:
         )
     else:
         _report_enterprise_failure(e)
-except Exception as e:
+except ImportError as e:
+    # A non-ModuleNotFoundError ImportError from the top-level import (e.g. the
+    # package exists but imports a name that doesn't) is a real bug, not an
+    # absent submodule.
     _report_enterprise_failure(e)
+else:
+    try:
+        register_enterprise(app)
+        # `print(..., flush=True)`: this block runs at module init, BEFORE
+        # `lifespan` calls `setup_logging()`. The default Python logger drops
+        # INFO records, so `logger.info` here would be silently swallowed. Print
+        # to stdout — docker logs captures it and CI greps for this exact string.
+        print("Trinity Enterprise modules registered", flush=True)
+    except Exception as e:
+        # ANY failure inside registration is a bug and gets the diagnostic —
+        # including ModuleNotFoundError, which is why this is a separate try.
+        _report_enterprise_failure(e)
 
 
 # WebSocket endpoint
--- a/src/backend/main.py
+++ b/src/backend/main.py
@@ -997,6 +997,21 @@ app.include_router(ws_tickets_router)  # WebSocket auth tickets (#550)
 # repo is restructured into `backend/` and `frontend/` subdirs so the
 # same repo can be dual-mounted (`src/backend/enterprise/` for Python,
 # `src/frontend/src/enterprise/` for Vite).
+def _report_enterprise_failure(exc: BaseException) -> None:
+    """Surface a real enterprise-registration failure loudly.
+
+    A BUG in registration (schema init, migration, router mount, a bad import)
+    must NOT take down the core platform — degrade and report instead of
+    crashing boot. Modules that registered before the failure stay active.
+    """
+    import traceback
+    print(
+        f"Trinity Enterprise registration FAILED — continuing OSS-only: {exc!r}",
+        flush=True,
+    )
+    traceback.print_exc()
+
+
 try:
     from enterprise.backend import register_enterprise  # type: ignore[import-not-found]
     register_enterprise(app)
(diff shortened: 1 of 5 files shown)


dolho:
## Extensive testing + one addition

### The OSS-only path, verified on a genuinely submodule-less build

The backend image ships **without** `enterprise/` (it's bind-mounted in dev), so I built a real OSS-only tree — `src/backend` copied minus `enterprise/` — and booted it against the same env:

```
$ docker run --rm --env-file .env -v <oss-tree>:/app <backend-image> \
    sh -c 'python -c "import main"'
Trinity Enterprise submodule not present — OSS-only build (this is normal; …)
IMPORT OK
```

So the narrowed branch still reports a real OSS clone correctly — the change doesn't regress the case the message exists for.

The loud path was observed live in the opposite direction earlier: with the submodule mounted and Alembic unable to resolve a revision, boot printed `Trinity Enterprise registration FAILED — continuing OSS-only: CommandError(...)` with a traceback, instead of the reassuring line.

### Added: `EntitlementService.unregister_module` (6bc7f19f)

Fault injection against the real `register_enterprise` (details in Abilityai/trinity-enterprise#197) turned up a third bug: a module that claims its feature_id and *then* fails leaves a **ghost entitlement** — advertised by `GET /api/settings/feature-flags`, UI rendered by the OSS bundle, every call 404s because the router never mounted.

The registering side needs a way to withdraw its own claim, so this adds the edition-agnostic primitive here and the enterprise side uses it (snapshot the registry around each `register()`, withdraw whatever leaked on failure). Idempotent; withdrawing an unregistered id is a no-op. Nothing in OSS calls it today — it exists for the registering side, same shape as `register_module`.

3 new tests in `test_847_entitlement_seam.py` cover withdraw / idempotency / no-collateral-damage.

### Suites

- `test_847_entitlement_seam.py` + `test_ent196_enterprise_load_reporting.py` → **23 passed, 1 skipped** (the skip is the no-submodule case, unreachable in a checkout that has one — covered end-to-end by the container test above instead).
- Enterprise-docs guard re-run against `docs/`, `CLAUDE.md`, `main.py`, `entitlement_service.py` — clean; the new docstring describes the generic seam and names no module.
- Full OSS unit suite result is unchanged from the PR body: 4667 passed, 5 pre-existing failures that reproduce on a clean `origin/dev` worktree.

dolho:
## This PR now also fixes #1653 — and #1653's approach found a hole in mine

#1653 (credit @vybe) describes exactly the bug this PR was opened for; I hit it independently while testing trinity-enterprise#196. Reading it, its **suggested fix is better than my original one in a case I had wrong**, so I've adopted it.

### The hole in the name-check-only approach

My first pass kept one `try` and narrowed the arm to `ModuleNotFoundError` + "is the missing module the `enterprise` package?". That mishandles the case the issue is actually about:

| scenario | exception | name-check verdict | correct |
|---|---|---|---|
| OSS clone, no submodule | `ModuleNotFoundError(name='enterprise')` | benign | benign ✓ |
| module imports a missing NAME | plain `ImportError` | loud | loud ✓ |
| missing third-party dep | `ModuleNotFoundError(name='pyotp')` | loud | loud ✓ |
| **`register()` imports a missing SIBLING** | `ModuleNotFoundError(name='enterprise.backend.gone')` | **benign** ← wrong | loud |

A module doing `from .missing_sibling import x` inside `register()` produces a name whose first segment *is* `enterprise` — indistinguishable from an absent submodule by name alone. And lazy imports inside `register()` are exactly the established pattern the issue cites.

### Both parts are needed

- **#1653's structural fix**: the `ImportError` arms guard only the top-level import; `register_enterprise(app)` moves to the `else` arm under `except Exception`. This catches the sibling case — and anything else registration can raise.
- **The name check, kept for the top-level import**, where structure cannot help: if `enterprise/backend/__init__.py` needs a third-party package that isn't installed, that's a real failure wearing the same exception type as an absent submodule. #1653's snippet alone would report it as "this is normal".

### Verified live — all four branches

```
absent submodule (real OSS tree, enterprise/ removed)  -> calm line + IMPORT OK
missing third-party dep in the top-level import        -> diagnostic
from .gone_module import x inside register()           -> diagnostic
   ModuleNotFoundError("No module named 'enterprise.backend.gone_module'")   ← the trap case
the live product_events failure on this instance       -> diagnostic, real cause named
```

That last one is the bug in the wild: this instance currently has trinity-enterprise#184's telemetry module importing an OSS table whose OSS half (#1721) hasn't merged. Before: *"submodule not present — this is normal"*. After: `registration FAILED — ImportError("cannot import name 'product_events' …")`.

### Acceptance criteria (#1653)

- [x] `ModuleNotFoundError` inside `register_enterprise(app)` reported as a bug with a traceback naming the module
- [x] Genuinely absent submodule still prints the calm line and boots clean — verified on a real OSS tree, not a mock
- [x] Core boot still survives a broken enterprise module (degrade-don't-crash holds; the live instance is proof)
- [x] Regression tests for both branches

### Tests

`tests/unit/test_1653_enterprise_load_reporting.py` — 7 passed (renamed from the ent#196 file). The structural test parses `main.py` with **`ast`**, not string search: the surrounding comments mention `register_enterprise(app)` in prose and a substring check matched those (it did, and failed the first run).

Also pins the CI string contract — `build-without-submodule.yml` greps for `"Trinity Enterprise submodule not present"`, so a reworded message would break CI silently.

Related tests: 87 passed, 3 skipped.

vybe (approved):
Validated via /validate-pr: OSS half of the ent#196 seam pair (unregister_module + honest ImportError narrowing), named regression tests, security clean. Enterprise half (ent#197) merged.

Issue #15, opened 2026-02-16
Security: Agent Credential Leakage in Execution Logs [FIXED]


vybe:
## Security Bug: Environment Variable Exposure via Agent Tools

**Severity**: High  
**Status**: ✅ FIXED in commit `8c05bb3`  
**Reported**: 2026-02-13  
**Fixed**: 2026-02-16  

---

## Description

Sensitive environment variables (including `TRINITY_MCP_API_KEY`, `OPENAI_API_KEY`, and custom credentials from `.env`) were being exposed in the `execution_log` stored in the database.

### Attack Vectors

| Vector | Description |
|--------|-------------|
| **Command Leak** | Agent executes `env` or `printenv` via Bash, dumping all secrets to stdout |
| **Direct Retrieval** | User asks "What is my API key?", agent reads environment and outputs the secret |

### Impact

In both cases, secrets were:
1. ❌ Returned to the LLM (leaked to model provider)
2. ❌ Persisted in the `execution_log` or `response` columns in the database
3. ❌ Displayed in the UI history

---

## Evidence

**Verified on Production (2026-02-15):**
- **Vector 1**: Running "Check environment variables" resulted in a full dump of `.env` including `TRINITY_MCP_API_KEY`
- **Vector 2**: Execution showed the agent successfully retrieving and displaying API tokens upon request

Example leaked pattern (BEFORE FIX):
```
TRINITY_MCP_API_KEY=trinity_mcp_...
TRINITY_MCP_URL=http://mcp-server:8080/mcp
OPENAI_API_KEY=sk-...
```

---

## Fix Implementation

Multi-layer credential sanitization was implemented:

### Agent-Side (Primary Layer)
| File | Change |
|------|--------|
| `docker/base-image/agent_server/utils/credential_sanitizer.py` | NEW - Sanitization utility |
| `docker/base-image/agent_server/services/claude_code.py` | Sanitizes subprocess output and response text |
| `docker/base-image/agent_server/routers/credentials.py` | Refreshes sanitizer cache after credential injection |

### Backend-Side (Defense-in-Depth)
| File | Change |
|------|--------|
| `src/backend/utils/credential_sanitizer.py` | NEW - Pattern-based sanitization |
| `src/backend/routers/chat.py` | Sanitizes execution_log, tool_calls, response before DB persistence |

### Patterns Detected and Redacted
- OpenAI keys (`sk-*`, `sk-proj-*`)
- Anthropic keys (`sk-ant-*`)
[...]

PR #1740, opened 2026-07-22
feat(mcp): room tools for shared multi-agent sessions (ent#169)

dolho:
OSS half of Abilityai/trinity-enterprise#169 (enterprise module: Abilityai/trinity-enterprise#210).

Five MCP tools — `create_room`, `list_rooms`, `read_room`, `post_to_room`, `close_room` — proxying the enterprise rooms API. Tool code ships in the public server even though the feature is gated (Invariant #13).

## Thin by design

All authorization is server-side and nothing here decides access:

- **Membership is the grant** — a non-member gets a uniform 404 from the backend.
- **The acting identity comes from the API key**, so an agent-scoped key always posts as its own agent and cannot post as another participant. There is no `sender` parameter to spoof.
- The tools' only local logic is turning an absent/forbidden route into a structured result instead of a transport error.

## Graceful fallback, verified

On an OSS build (or unentitled instance) the routes are absent, so the tools return:

```json
{ "error": "shared_sessions_not_enabled",
  "message": "Shared sessions are not enabled on this Trinity instance." }
```

Verified live in both directions: with the module unregistered → the fallback above; with it registered → real data.

## Live verification

Real MCP JSON-RPC with an agent's own `TRINITY_MCP_API_KEY` (tool count 100 → **105**):

- `list_rooms` → the agent's rooms with status, `stop_reason`, message and participant counts
- `read_room` with `since` → transcript tail + participants
- `post_to_room` → landed as `#15 agent/portal-demo`, i.e. recorded as the **agent**, resolved from auth context

## Known gap

`TrinityClient.request()` takes no headers argument, so `post_to_room` cannot send an `Idempotency-Key`. The endpoint accepts one — Invariant #18 is satisfied at the boundary — but wiring it through the MCP client needs a client-signature change I didn't want to make unasked. Left as a follow-up.

**Note for reviewers:** the MCP server is a baked image, so `docker compose build mcp-server` is needed to pick these up; `npm run build` on the host is not enough.

### DIFF
--- a/src/backend/routers/executions.py
+++ b/src/backend/routers/executions.py
@@ -21,7 +21,11 @@ from services.agent_service.helpers import accessible_agent_names, narrow_to_age
 router = APIRouter(prefix="/api/executions", tags=["executions"])
 
 _VALID_STATUSES = {"running", "queued", "success", "failed", "error", "cancelled", "skipped"}
-_VALID_TRIGGERS = {"schedule", "manual", "agent", "mcp", "chat", "session", "public", "webhook", "fan_out", "loop", "reminder"}
+# NOTE: this is a filter ALLOW-LIST, not a DB enum — `triggered_by` is a plain
+# TEXT column. An unknown value here degrades to "no filter" (see below), which
+# silently returns EVERY execution instead of none, so a trigger that reaches
+# this endpoint must be listed or its filter lies. `room` is ent#169.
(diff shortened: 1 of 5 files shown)


vybe (approved):
Validated: thin server-authorized proxy tools with structured not-enabled fallback + a real allow-list filter fix with named test. Enterprise rooms backend (ent#210) merged with the enumeration-oracle fix.

PR #1742, opened 2026-07-22
fix(executions): make tool_calls a summary, not a copy of the transcript (#1741)

### LINKED ISSUE #1741: bug: schedule_executions.tool_calls stores a verbatim copy of execution_log — breaks the tool-call metric and survives log retention
## Summary

`task_execution_service.apply_result` writes the **entire raw execution log** into `schedule_executions.tool_calls`:

```python
# services/task_execution_service.py:1786-1790
if isinstance(exec_log, list) and len(exec_log) > 0:
    execution_log_json = json.dumps(exec_log)
    execution_log_json = sanitize_execution_log(execution_log_json)
    tool_calls_json = execution_log_json        # <- the same blob, verbatim
```

Verified on a live execution — the two columns are byte-identical:

```
tool_calls identical to execution_log: {'identical': True, 'log_len': 23438, 'tc_len': 23438}
```

That causes three distinct problems.

## 1. The tool-call metric is always 0 (silently)

`db/schedules/analytics.py::get_agent_schedules_summary` parses that column expecting **chat-shaped** entries:

```python
for entry in parsed:
    if isinstance(entry, dict) and (entry.get("name") or entry.get("tool")):
        tool_total_by_sched[row["schedule_id"]] += 1
```

But the stored value is a Claude Code **stream-json event list** — `{"type": "system", "subtype": "init", ...}`, `{"type": "assistant", "message": {...}}`, etc. Tool calls live one level deeper, as `tool_use` blocks inside `message.content`.

Measured on a real scheduled run that used three tools:

```
event types: {'system': 13, 'assistant': 11, 'rate_limit_event': 1, 'user': 4, 'result': 1}
entries the rollup would count:  0 of 30
ACTUAL tool_use calls nested in assistant messages: 4
tools used: {'ToolSearch': 2, 'WebFetch': 1, 'mcp__trinity__report': 1}
```

So the Schedules-tab inline stats and the Overview "Schedules performance" section (#1115) report **`tool_call_total: 0`** for every scheduled execution, forever. It reads as the *fact* "this schedule uses no tools" rather than "not measured" — the failure mode is a confident wrong number, not a blank.

## 2. It survives `execution_log` retention (the one that matters)

`prune_execution_logs` (#772) nulls **only** `execution_log`:

```python
update(schedule_executions)
  .where(schedule_executions.c.id.in_(ids))
  .values(execution_log=None)
```

`tool_calls` is never mentioned anywhere in `db/schedules/retention.py` (0 occurrences). Since it holds a byte-identical copy of the log, after the `execution_log_retention_days` sweep (default **30**) an operator believes the transcript is gone while a verbatim copy persists until the row itself is deleted at `execution_row_retention_days` (default **90**).

**A retention control is silently bypassed for ~60 days** by a column whose name gives no hint it contains the transcript. Same class as #1449 (`backlog_metadata` scrub), which was treated as a security invariant — and that scrub does not cover `tool_calls` either.

## 3. Every execution stores its log twice

23 KB duplicated for a single trivial run. Multiplied across the fleet and the 90-day row window, this is pure waste in the largest table.

## How it was found

Dogfooding: built a scheduled agent that checks a news page daily. It demonstrably used WebFetch and the `report` MCP tool; the Schedules performance panel reported 0 tool calls. Pulling that thread led to the shared blob.

## Acceptance criteria

- [ ] `tool_calls` holds a tool-call *summary* (or the metric is computed from the real `tool_use` blocks) so `tool_call_total` reflects reality
- [ ] Whatever remains in `tool_calls` is covered by the same retention sweep as `execution_log` — a transcript copy must not outlive the transcript
- [ ] The log is not stored twice per execution
- [ ] Existing rows are handled (backfill/scrub), not just new ones
- [ ] Regression test: a scheduled execution that uses N tools reports N, and a log-retention sweep leaves no transcript copy behind

## Notes

- If the intent was "keep the whole log because parsing is hard", then the column is misnamed and the analytics parser is the thing that's wrong — either way the current pair is inconsistent.
- Whichever direction is chosen, the retention coupling has to be fixed regardless of the metric.

dolho:
Related to #1741.

## The bug

`apply_result` assigned `tool_calls_json = execution_log_json` — the entire Claude Code stream-json transcript, verbatim, into a column named `tool_calls`. Verified byte-identical on a live run (23,438 chars each).

| | before | after |
|---|---|---|
| `tool_call_total` for a run that made 4 tool calls | **0** | **4** |
| execution-detail "Tool Calls" panel | **empty** | 4 rows |
| copy survives `execution_log` retention | **yes, ~60 days** | no |
| transcript stored twice | yes | no |

The metric didn't fail loudly — it reported the *fact* "this schedule uses no tools". Tool uses are nested inside `assistant` events' `message.content`, and the rollup scanned only top level: **0 of 30** entries matched.

## Why derive it backend-side

The chat path was already correct: `/api/chat` returns `execution_log_simplified` (`{type, tool, input, success, duration_ms}`) — which is exactly the shape both the analytics parser (`entry.get("name") or entry.get("tool")`) and the UI (`filter(c => c.type === 'tool_use')`) expect.

`/api/task` returns **only** `raw_messages`. So `apply_result` genuinely had nothing else to store — this wasn't a wrong-variable typo.

Teaching `/api/task` to send the simplified log would need a base-image rebuild and leave a mixed-fleet window where old agents still write transcripts. Deriving it in `services/tool_call_summary.py` works on **every agent image immediately**.

**Honest trade-off:** `success` and `duration_ms` come from the agent's own tracking and cannot be reconstructed from the transcript. Both render behind `v-if` in the UI, so they degrade to "not shown" rather than wrong. Having `/api/task` also send `execution_log_simplified` would restore them — clean follow-up, not required for correctness.

## Retention

`prune_execution_logs` now nulls `tool_calls` alongside `execution_log`.

The predicate also had to widen: gating on `execution_log IS NOT NULL` alone would **permanently strand** the copies on rows a previous sweep already nulled — precisely the rows where the copy most obviously outlived what it duplicates. Predicate and #1644 blast-radius guard stay in sync by construction (they share `_execution_log_prune_predicate`).

## Existing rows

An idempotent, bounded pass in the cleanup sweep rewrites legacy blobs in place. Non-destructive — the transcript remains in `execution_log`, so a bad conversion loses nothing recoverable. Classified in the #1082 update-site inventory as a non-status writer (that guard caught the new site, as designed).

## Verified end to end on a live instance

```
[Cleanup] Converted legacy tool_calls on 34 executions to the summary shape (#1741)

Daily watermelon watch   executions=1  tool_call_total=4   (was 0)
Cron fire probe          executions=1  tool_call_total=0   (used no tools — a TRUE zero)
```

The second row is the control: a schedule explicitly told not to use tools still reports 0, so the fix didn't just turn every zero into a number.

Recovered from the real bugged row: `['ToolSearch', 'WebFetch', 'ToolSearch', 'mcp__trinity__report']`.

## Tests

`tests/unit/test_1741_tool_call_summary.py` — 15 passed. Extraction from the real nested shape (including a test asserting the *old* scan returns 0 on the same input), pass-through of already-simplified entries, malformed input never raising, bounds on count and input size, idempotent conversion, and two guards pinning the retention coupling.

Suite: 262 passed across retention/cleanup/schedule. Three failures, two pre-existing on clean `origin/dev` (`test_1474_read_boundary_z`, `test_1472_schedule_validation` — local tz/bcrypt env), and the third was the #1082 inventory tripwire firing on my new update site, now classified.

## How it was found

Dogfooding — built a scheduled agent that monitors a news page daily. It demonstrably used WebFetch and the `report` MCP tool; the Schedules performance panel said 0 tool calls. Pulling that thread found the shared blob and the retention hole behind it.

Fixes #1741


vybe:
**/validate-pr: REQUEST CHANGES — real regression, not the CI flake.**

The regression-diff failure survived a rerun and is deterministic across all three seeds:

- `test_1083_apply_result.TestSuccessGolden::test_success_won_writes_full_row` — fails under HEAD in seeds 12345/67890/99999, passes under BASE.

The tool_calls-summary change appears to break the #1083 `apply_result` golden contract (that test pins the full row a CAS-won SUCCESS terminal writes, presumably including `tool_calls` carrying the old verbatim shape). Either the golden needs updating to the new summary shape, or `apply_result`'s writer path was missed by the summarization and now diverges from the sync path — please determine which before re-requesting review.

Holding this out of the 0.8.5 payload until green. (The earlier 25-min hung-job failures on this and sibling PRs *were* flakes — reruns cleared #1734/#1735 — but this one is real.)

vybe (approved):
Validated via /validate-pr: closing keyword ✅ (Fixes #1741, P2/type-bug), security greps clean, no schema change, CI full pytest matrix ×3 seeds + regression diff green, touched unit files re-run locally (48 passed). Design notes: retention coupling shares the prune predicate with the #1644 guard by construction ✅. One non-blocking observation: the backfill rescans from the lowest id each cycle and converted rows stay in the candidate set, so on a busy install conversion can stall beyond the first max_rows id-ordered rows — self-limiting, since the widened prune predicate nulls legacy blobs once they age past the log window. Operators on busy installs may see a one-time #1644 retention-guard ack prompt after merge (the widened predicate legitimately enlarges the candidate set).

PR #1744, opened 2026-07-22
fix(executions): reconcile fleet totals with the per-agent view (#1743)

### LINKED ISSUE #1743: Fleet execution totals include soft-deleted agents' spend that no per-agent surface attributes
## Summary

`GET /api/executions/stats` (the Operations → Executions stat cards) counts executions belonging to **soft-deleted** agents. `GET /api/agents/execution-stats` (the Dashboard Grid tiles) does not. Neither is wrong on its own, but the two cannot be reconciled and nothing on either surface explains the difference.

Measured on a dev instance, 24h window:

| surface | executions | cost |
|---|---|---|
| DB truth | 16 | $0.8933 |
| `GET /api/executions/stats?hours=24` | 16 | $0.8933 |
| sum of `GET /api/agents/execution-stats` tiles | **12** | **$0.6291** |
| **gap** | **4** | **$0.2642** |

$0.26 — **30% of the window's spend** — is visible in the fleet cards and attributable to nothing on the dashboard.

The gap is one soft-deleted agent:

```
ownership row: {'agent_name': 'trinity-skill-runner', 'deleted_at': '2026-07-22T08:39:13.947615Z'}
orphaned executions: 4  cost: 0.2642
```

## Why both behaviours are individually defensible

- Keeping the execution rows is deliberate — cost is billing truth, and the architecture says as much for ephemeral ghosts ("exec/cost stats inclusive (billing truth)"). Soft-deleted agents are also recoverable, so discarding their history would be wrong.
- Not rendering a tile for a deleted agent is obviously right — there is nothing to act on.

The problem is only that **no surface reconciles them**. An operator auditing spend finds a number they cannot attribute, with no hint that deleted agents are the cause.

## It grows

Every deleted agent's spend accumulates in the fleet totals with no attribution until `execution_row_retention_days` (default 90) removes the rows. On a fleet with churn — ephemeral/ghost agents especially, which are created and discarded by design — the unattributable fraction is unbounded within that window.

## Suggested fix

Cheapest honest option: have `GET /api/executions/stats` return the deleted-agent portion as its own field (e.g. `deleted_agent_count` / `deleted_agent_cost`), and let the Executions panel show it as a "deleted agents" line. The totals then reconcile with the tiles by subtraction, and the spend stays attributable.

Alternatives — excluding deleted agents from fleet stats (loses billing truth) or rendering ghost tiles (clutters the dashboard with un-actionable agents) — both seem worse.

## Acceptance criteria

- [ ] Fleet execution totals and the sum of per-agent stats reconcile, or the difference is explicitly surfaced
- [ ] Spend belonging to deleted agents stays visible and attributable (not silently dropped)
- [ ] Regression test asserting the two surfaces reconcile in the presence of a soft-deleted agent with executions

## Found by

A cross-surface consistency audit — checking that every surface reporting a number agrees with the DB. Same method that found #1741; all other checked metrics (counts, success_rate, cost, duration avg/p95, context_avg) agreed exactly across `/api/executions/stats`, agent analytics, schedules analytics-summary and the tile endpoint.

## Noted but NOT filed (no live defect)

`success_rate` is **0–1** from `/api/agents/{name}/analytics` and per-schedule analytics, but **0–100** from `/api/agents/execution-stats`. Every current consumer handles its own source correctly (`ScheduleAnalyticsCard`/`OverviewPanel` multiply by 100; `AgentTile` does not), so nothing mis-renders today — but the same field name carrying two scales is a trap for the next consumer.

dolho:
Fixes #1743.

## The problem

`GET /api/executions/stats` (Operations stat cards) counts every accessible execution row. `GET /api/agents/execution-stats` (Dashboard tiles) renders only **live** agents. Execution rows outlive their agent on purpose — cost is billing truth, and a soft-deleted agent is recoverable — so the two surfaces differed by the deleted agents' share, with nothing explaining the gap.

Measured on a dev instance, 24h window:

| surface | executions | cost |
|---|---|---|
| fleet stats | 16 | $0.8933 |
| sum of tiles | 12 | $0.6291 |
| **gap** | **4** | **$0.2642** (30% of the window's spend) |

## The fix

Neither endpoint was wrong on its own, so **neither changes its totals**. The difference is *reported* instead:

```json
{ "total": 16, "total_cost": 0.8933,
  "deleted_agent_count": 4, "deleted_agent_cost": 0.2642 }
```

`total − deleted_agent_count` now reconciles with the tiles by subtraction, and the spend stays attributable rather than being silently dropped from either side.

Computed in the **same single pass** via a LEFT JOIN matching only live ownership rows, so a soft-deleted agent (`deleted_at` set) and a hard-purged one (no row at all) collapse together — both are equally invisible to the tiles, so both belong in the slice.

The Executions panel shows one line, and only when the count is non-zero:

> Includes **4** executions ($0.26) from deleted agents, which have no tile on the dashboard.

## Things worth a reviewer's eye

- **The join made columns ambiguous.** `agent_name`, `status`, `started_at` and `cost` all exist on both sides; PostgreSQL rejects the ambiguous form outright (SQLite would have tolerated some of it). Every column in that query is now qualified.
- **The empty-accessible-set early return** short-circuits before the query, so it needed the new keys too — otherwise a caller with no accessible agents loses the fields.
- **Fields are additive with defaults**, so a client predating them still validates.
- The `?agent=` / non-admin filtered path is covered: the split is computed over exactly the filtered set, not the whole table.

## Verification

Live, after the change:

```
fleet stats        : total=16  cost=0.8933
deleted-agent slice: count=4   cost=0.2642
fleet minus slice  : total=12  cost=0.6291
sum of tiles       : total=12  cost=0.6291
RECONCILES: True
```

`tests/unit/test_1743_deleted_agent_spend.py` — 8 passed: the slice itself, the reconciliation invariant, soft-deleted **and** purged both counted, zero-when-clean (so the UI line stays hidden), the agent-filtered path, the empty-set early return, the windowed path, and the model defaults.

Related suites: 212 passed, 3 skipped.

## Found by

A cross-surface consistency audit — checking every reported metric against the DB across all surfaces. Everything else agreed exactly (counts, cost, success_rate, duration, context); this was the one disagreement.

## Not addressed here

`success_rate` is 0–1 from the analytics endpoints and 0–100 from `execution-stats`. Every current consumer handles its own source correctly, so there's no live defect — noted in #1743's body rather than fixed here, since changing either scale would silently break whichever consumer wasn't updated in the same breath.

### DIFF
--- a/src/backend/db/schedules/stats.py
+++ b/src/backend/db/schedules/stats.py
@@ -385,13 +385,15 @@ class ScheduleStatsMixin:
                 "total": 0, "success_count": 0, "failed_count": 0,
                 "total_cost": 0.0, "success_rate": 0.0,
                 "running_count": 0, "queued_count": 0, "hours": hours,
+                "deleted_agent_count": 0, "deleted_agent_cost": 0.0,
             }
 
         bind: Dict = {}
         agent_where = ""
         if agent_names is not None:
             keys = [f"an{i}" for i in range(len(agent_names))]
-            agent_where = "WHERE agent_name IN (%s)" % ",".join(f":{k}" for k in keys)
+            # Qualified: the #1743 LEFT JOIN puts agent_name in both tables.
+            agent_where = "WHERE se.agent_name IN (%s)" % ",".join(f":{k}" for k in keys)
             bind.update(dict(zip(keys, agent_names)))
 
         # Single-pass query: windowed totals via conditional aggregation +
@@ -401,7 +403,7 @@ class ScheduleStatsMixin:
         # integer raises; SQLite tolerated it. #300.) The named :cutoff bind is
         # reused across all four CASE expressions.
         if hours:
-            time_cond = "started_at > :cutoff"
+            time_cond = "se.started_at > :cutoff"
             bind["cutoff"] = iso_cutoff(hours)
         else:
             time_cond = "1=1"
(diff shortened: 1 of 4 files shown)


vybe (approved):

PR #1746, opened 2026-07-22
security: revoke an agent's MCP keys when the agent is deleted (#1745)

### LINKED ISSUE #1745: security: deleting an agent does not revoke its API key — the credential keeps admin-inherited fleet access and can mint new keys
## Summary

Deleting an agent leaves its agent-scoped MCP API key `is_active = 1` and fully functional. The credential keeps working against both the REST API and the MCP server, inherits its owner's role, and can **mint a fresh key of its own** — so revoking the orphaned key afterwards does not contain it.

Reproduced end to end on a dev instance.

## Reproduction

```bash
POST /api/agents                      # create agent "keytest"
POST /api/agents/keytest/start
K=$(docker exec agent-keytest printenv TRINITY_MCP_API_KEY)

DELETE /api/agents/keytest            # {"message":"Agent keytest deleted"}
```

DB state after deletion:

```
ownership deleted_at: 2026-07-22T12:06:46Z     <- agent is gone
mcp key is_active   : 1                        <- credential is NOT
```

The same `K` still works:

| endpoint | result |
|---|---|
| `GET /api/users/me` | **200** |
| `GET /api/agents` | **200** — enumerates all 11 agents on the instance |
| `GET /api/agents/{other}` · `/info` · `/schedules` · `/executions` | **200** for agents unrelated to the deleted one |
| `GET /api/activities/timeline` | **200** |
| `GET /api/audit-log` | **200** — admin-only surface |
| `GET /api/executions/stats`, `/api/reports` | **200** |
| MCP server `initialize` | **200 — accepted**, so it can still call tools |

## The part that makes it P1

The dead credential can **mint a new one**:

```bash
POST /api/mcp/keys  -d '{"name":"probe-1743-delete-me"}'
→ 200 {"id":"DsMYlyWIzCRu6kfBlscdKw","key_prefix":"trinity_mcp_CKrBeAq_", ...}
```

So the natural containment action — "delete the agent", or later "revoke that stale key" — does not contain anything. Anything that held the agent's key (the container, a leaked env dump, a prompt-injected agent that read its own `TRINITY_MCP_API_KEY`) can mint an independent, user-scoped key that outlives every trace of the agent.

*(The probe key was deleted immediately; verified gone.)*

## Why the reach is so wide

Two issues compound:

1. **This bug** — agent deletion does not deactivate `mcp_api_keys` rows for that agent.
2. **Known, accepted risk** — an agent-scoped key resolves to its owner and inherits the owner's role, so on a default admin-owned install it reaches admin surfaces. #69 Part 2 records this as *"Fleet-wide narrowing of remaining agent-key breadth = accepted-risk follow-up"*, and the same class was just fixed narrowly for the skill runner (trinity-enterprise#195, where an agent granted itself skills and disabled a feature).

(2) alone is a documented trade-off. (1) turns it into a credential that survives the thing it was scoped to.

## Not hypothetical on this instance

Five active agent-scoped keys are already bound to agents that no longer exist:

```
agent-pytest-ephemeral-ssh-access-877586-key -> pytest-ephemeral-ssh-access-877586
agent-pytest-ephemeral-ssh-access-d10d17-key -> pytest-ephemeral-ssh-access-d10d17
agent-pytest-ephemeral-ssh-access-554de6-key -> pytest-ephemeral-ssh-access-554de6
agent-tel-live-test-key                      -> tel-live-test
agent-trinity-skill-runner-key               -> trinity-skill-runner
```

Three are leftover ephemeral test agents, which is its own signal: #69's hard-discard path is supposed to purge a ghost's rows via `purge_ephemeral_agent_ownership` + `clear_agent_runtime_state`, and the key survived that too.

The canary's **L-03** invariant already flags this class ("no live row in any cross-cutting table … agent-scoped MCP keys") — it fired on this instance and nobody was watching, because the canary is disabled by default.

## Acceptance criteria

- [ ] Deleting an agent deactivates (or deletes) its agent-scoped `mcp_api_keys` rows, in the same transaction as the ownership write
- [ ] The ephemeral hard-discard path does the same
- [ ] An existing orphaned key is deactivated by a backfill/sweep — five already exist here
- [ ] A revoked/orphaned key is rejected by **both** the REST API and the MCP server
- [ ] Regression test: create agent → capture key → delete agent → key is rejected, and cannot mint a new key

## Note on scope

Soft-delete is recoverable, so "deactivate on delete / reactivate on recover" may be preferable to a hard delete of the key row — worth a decision rather than assuming. Either way the key must not authenticate while the agent is deleted.

dolho:
Fixes #1745.

## The bug

Soft-deleting an agent left its agent-scoped MCP key `is_active = 1`. `mcp_api_keys` is *already* registered CASCADE in `AGENT_REFS` with the rationale **"an orphaned key must not survive its agent"** — but `cascade_delete()` only runs at the **hard purge**, after the whole soft-delete window (default 180 days). The intent was recorded; the window was wide open.

## The fix

Deactivate (not delete) **in the same transaction as the ownership write**, and reactivate on recover.

Deactivation rather than deletion because soft-delete is *recoverable* — the credential state has to be reversible too, or a recovered agent silently cannot talk to the platform. `validate_mcp_api_key` already requires `is_active = 1`, so this closes the REST and MCP paths together; a test pins that dependency, because the fix is cosmetic without it. Covers `scope='agent'` **and** `scope='connector'` — both are per-agent credentials and both are CASCADE entries.

A cleanup sweep handles keys orphaned before the fix, and any that slip through a path bypassing `delete_agent_ownership`. Not age-gated — an orphaned credential is a security invariant, not a retention window (the #1449 reasoning).

## Verified by replaying the original attack

```
BEFORE delete:  /api/agents          200

DELETE /api/agents/keytest2

AFTER delete:   /api/users/me        401
                /api/agents          401
                /api/audit-log       401
                POST /api/mcp/keys   401   ← the persistence path, closed
                MCP initialize       401
```

Recover/re-delete round trip flips the key back on, then off:

```
recovered+started -> key 200
DELETE            -> key 401,  deleted_at set,  is_active 0
```

Backfill on the real instance:

```
[Cleanup] Deactivated 12 MCP key(s) belonging to agents that are no longer live (#1745)
per-agent keys for non-live agents: 15 total, 0 still active
```

## Things a reviewer should weigh

- **Deactivate vs delete.** I chose reversible. The trade-off is that the *same* key value comes back on recovery — if the credential was already leaked, recovery restores a compromised key. Deleting and re-minting on recover would be stronger but breaks a running container's env. Worth a call.
- **A manually-revoked key is reactivated by recovery.** The delete path cannot distinguish "operator revoked this" from "delete revoked this" without a `revoked_by` column. Current behaviour is recorded in a test named for it rather than left to be discovered.
- **Only `delete_agent_ownership` is covered.** Any future path that removes an agent another way is caught by the sweep within 5 minutes, not instantly.

## Unrelated bug found while testing (not fixed here)

A **recovered agent cannot be deleted again until it is started**. Recovery deliberately does not recreate the container, and `DELETE /api/agents/{name}` looks up the container first, so it returns `404 "Agent not found"` for an agent whose ownership row is very much alive. It briefly looked like my fix had failed — it hadn't; the delete never ran. Pre-existing and out of scope, but worth its own issue.

## Tests

`tests/unit/test_1745_agent_key_revocation.py` — 11 passed: delete revokes, connector scope too, other agents' and user-scoped keys untouched, recover restores, the delete→recover→delete round trip, backfill over soft-deleted **and** never-existed agents, idempotency, and the `is_active` gate the whole fix depends on.

Related suites: 584 passed, 3 skipped.

### DIFF
--- a/src/backend/database.py
+++ b/src/backend/database.py
@@ -1024,6 +1024,14 @@ class DatabaseManager:
     def get_agent_mcp_api_key(self, agent_name: str):
         return self._mcp_key_ops.get_agent_mcp_api_key(agent_name)
 
+    def set_agent_keys_active(self, agent_name: str, active: bool) -> int:
+        """#1745: activate/deactivate an agent's per-agent MCP keys."""
+        return self._mcp_key_ops.set_agent_keys_active(agent_name, active)
+
+    def deactivate_orphaned_agent_keys(self) -> int:
+        """#1745: deactivate per-agent keys whose agent is no longer live."""
+        return self._mcp_key_ops.deactivate_orphaned_agent_keys()
+
     def delete_agent_mcp_api_key(self, agent_name: str):
         return self._mcp_key_ops.delete_agent_mcp_api_key(agent_name)
 
--- a/src/backend/db/agents.py
+++ b/src/backend/db/agents.py
@@ -375,6 +375,26 @@ class AgentOperations(
         with get_engine().connect() as conn:
             return int(conn.execute(stmt).scalar() or 0)
 
+    @staticmethod
+    def _deactivate_agent_keys_in_txn(conn, agent_name: str, *, active: bool) -> int:
+        """Flip per-agent key activity on an EXISTING connection (#1745).
+
+        Runs inside the caller's transaction so the credential state can never
+        diverge from the ownership state — a key left active after a committed
+        delete is the whole bug.
+        """
+        from .tables import mcp_api_keys
+
+        return conn.execute(
+            update(mcp_api_keys)
+            .where(
+                mcp_api_keys.c.agent_name == agent_name,
+                mcp_api_keys.c.scope.in_(("agent", "connector")),
+                mcp_api_keys.c.is_active == (0 if active else 1),
+            )
+            .values(is_active=1 if active else 0)
+        ).rowcount or 0
+
     def delete_agent_ownership(self, agent_name: str) -> bool:
         """Soft-delete the agent ownership row (Issue #834 Phase 1a).
 
@@ -400,6 +420,16 @@ class AgentOperations(
                 .values(deleted_at=utc_now_iso())
             )
             if result.rowcount > 0:
+                # #1745: revoke the agent's credentials in the SAME transaction.
+                # `mcp_api_keys` is CASCADE in AGENT_REFS precisely because "an
+                # orphaned key must not survive its agent", but that cascade only
+                # runs at the hard purge — i.e. after the whole soft-delete
+                # window (default 180 days), during which the key kept
+                # authenticating, enumerating the fleet, and able to mint a fresh
+                # key of its own. Deactivate (not delete) so recovery can restore
+                # them; `validate_mcp_api_key` requires is_active=1, so this
+                # closes both the REST and MCP paths.
+                self._deactivate_agent_keys_in_txn(conn, agent_name, active=False)
                 return True
             # rowcount==0 — either already soft-deleted or doesn't exist
             row = conn.execute(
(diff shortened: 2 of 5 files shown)


vybe (approved):

PR #1748, opened 2026-07-22
fix(agents): delete an agent whose container is gone (#1747)

### LINKED ISSUE #1747: bug: an agent whose container is gone becomes invisible, undeletable, and blocks its own name
## Summary

`DELETE /api/agents/{name}` looks up the Docker container first and treats its absence as "agent not found":

```python
# src/backend/routers/agents.py:511
container = get_agent_container(agent_name)
if not container:
    raise HTTPException(status_code=404, detail="Agent not found")
```

But an agent's identity lives in `agent_ownership`, not Docker. So an agent with a **live ownership row and no container** is simultaneously invisible, undeletable, and holding its own name hostage.

The code immediately above already knows this hazard — the ephemeral branch is deliberately placed *before* the container lookup:

> `#69: this branch runs BEFORE the container lookup: a half-discarded ghost (crash residue: live row, no container) must be force-discardable, never 404.`

Ordinary agents get no such treatment.

## Reproduction

```bash
POST /api/agents            # create "orphan-probe"
docker rm -f agent-orphan-probe          # container lost (prune, crash, manual cleanup)
```

State — the ownership row is untouched and live (`deleted_at: None`), yet:

| action | result |
|---|---|
| `GET /api/agents` | **not listed** (Docker-as-truth, Invariant #11) |
| `DELETE /api/agents/orphan-probe` | **404 "Agent not found"** |
| `POST /api/agents` with the same name | **"Agent already exists"** |
| its MCP key | still `is_active = 1` |
| `POST /api/agents/orphan-probe/start` | **200** — the only escape |

So from the operator's side: *the agent vanished from the list, I can't delete it, and I can't recreate it under that name.* The only way out is to call `/start` on an agent you cannot see — which is not discoverable, and fails on its own terms if the image/subscription/template is broken (exactly the situations that tend to produce a missing container).

## How agents reach this state

1. **Recovering a soft-deleted agent.** #834 Phase 1c does this *by design*: recovery is metadata-only, `needs_container_recreate=true`, "operator runs `POST /api/agents/{name}/start`". Between recover and start, the agent is in this state and cannot be re-deleted.
2. `docker system prune` / manual `docker rm` on the host.
3. A crash between the ownership-row write and container creation.
4. Any Docker-side loss (daemon reset, disk cleanup, host migration).

## Suggested fix

Mirror what the ephemeral branch already does — decide from the **ownership row**, not from Docker. The container removal is best-effort cleanup, not an existence check:

```python
container = get_agent_container(agent_name)
if not container and not db.agent_ownership_exists(agent_name):
    raise HTTPException(status_code=404, detail="Agent not found")
# ... container stop/remove stays best-effort (it is already wrapped in try/except)
```

The soft-delete itself, cascade, name reservation and Redis cleanup all key off the row and work fine without a container.

The 404 detail is also misleading — the agent *is* found; only its container is missing. If a 404 is kept for the genuinely-unknown case, the missing-container case deserves a distinct message.

## Acceptance criteria

- [ ] An agent with a live ownership row and no container can be deleted
- [ ] A genuinely unknown agent name still returns 404
- [ ] A recovered-but-not-yet-started agent is deletable (the #834 Phase 1c flow)
- [ ] Regression test: create → remove the container out-of-band → delete succeeds and the name is reusable

## Found by

Testing #1745. After a recover/delete round trip the key appeared to survive deletion — it looked like the #1745 fix had failed. It hadn't: the delete returned 404 and never ran. Worth noting because the symptom impersonates a different bug.

dolho:
Fixes #1747.

## The bug

`DELETE /api/agents/{name}` decided existence from Docker:

```python
container = get_agent_container(agent_name)
if not container:
    raise HTTPException(status_code=404, detail="Agent not found")
```

Identity lives in `agent_ownership`, so an agent with a live row and no container was **invisible** (the listing is Docker-as-truth, Invariant #11), **undeletable**, and **still holding its own name** — escapable only by calling `POST /start` on an agent you cannot see.

## The fix

404 now requires **both** no container and no live row; container teardown becomes conditional.

```python
container = get_agent_container(agent_name)
if not container and not db.is_agent_live(agent_name):
    raise HTTPException(status_code=404, detail="Agent not found")
```

Everything downstream — soft-delete, queue cancel, Redis keyspace clearing, name reservation — keys off the agent **name** and never needed a container. Teardown is guarded because `container_stop(None)` would otherwise raise into the generic `except` and log a misleading error.

This is the same reasoning the **ephemeral branch directly above already applies**: it sits before the container lookup precisely so a half-discarded ghost ("live row, no container") is force-discardable, *"never 404"*. Ordinary agents just never got it. A test pins that ordering so this change didn't disturb #69.

## Verified live

```
create agent -> docker rm -f agent-stuck-probe   (live row, no container)
DELETE  -> HTTP 200,  deleted_at set             (was 404 forever)

unknown agent         -> 404   (still)
already soft-deleted  -> 404   (still)
```

## Two corrections to the issue as I filed it

1. **"the name is reusable" was wrong.** After this delete the name stays reserved — that is the normal #834 soft-delete policy for *every* delete (180 days, `is_agent_name_reserved` sees soft-deleted rows), not something this fix should change. The AC is corrected in the commit message; recreating under the same name still returns 409, correctly.
2. **Scope.** The same three-line guard appears in four endpoints in this module — `delete`, `stop`, `logs`, `activities`. My first patch attempt used a blanket string replace and would have changed all four; the assertion caught it. Only `delete` traps the agent, so only `delete` is changed. (`activities` 404ing on a missing container is arguably odd for a DB-backed read, but that is a separate question.)

## Tests

`tests/unit/test_1747_delete_without_container.py` — 6 passed. The static checks parse the module with **`ast`** and assert against the delete endpoint's source specifically, because that same guard exists in four functions and a file-wide string search would happily pass on the wrong one. Plus behavioural coverage of `is_agent_live` across all three states (live / already-deleted / unknown) and the containerless soft-delete.

Related suites: 858 passed, 3 skipped.

### DIFF
--- a/src/backend/routers/agents.py
+++ b/src/backend/routers/agents.py
@@ -508,8 +508,21 @@ async def delete_agent_endpoint(agent_name: str, request: Request, current_user:
             )
         }
 
+    # #1747: an agent's identity lives in `agent_ownership`, not in Docker.
+    # This used to 404 whenever the container was missing, which left an agent
+    # with a live row simultaneously invisible (the listing is Docker-as-truth),
+    # undeletable, and still holding its own name — escapable only by calling
+    # `/start` on an agent you cannot see. Agents reach that state routinely:
+    # the #834 Phase 1c recovery flow leaves them there BY DESIGN (metadata-only,
+    # "operator runs POST /start"), as does any docker prune / daemon reset or a
+    # crash between the row write and container creation.
+    #
+    # Decide existence from the ROW and treat the container as best-effort
+    # cleanup — exactly what the ephemeral branch above already does for the same
+    # "live row, no container" residue. An already-soft-deleted agent still 404s:
+    # it is deleted, and `is_agent_live` is False for it.
     container = get_agent_container(agent_name)
-    if not container:
+    if not container and not db.is_agent_live(agent_name):
         raise HTTPException(status_code=404, detail="Agent not found")
 
     # Issue #834 Phase 1a: agent delete is now a SOFT-delete. We stop +
@@ -522,11 +535,20 @@ async def delete_agent_endpoint(agent_name: str, request: Request, current_user:
     # (default 180). At that point the #816 cascade_delete primitive
     # tears down all the child rows and on-disk artifacts in one shot.
 
-    try:
-        await container_stop(container)
-        await container_remove(container)
-    except Exception as e:
-        logger.warning(f"Error stopping/removing container: {e}")
+    if container:
+        try:
+            await container_stop(container)
+            await container_remove(container)
+        except Exception as e:
+            logger.warning(f"Error stopping/removing container: {e}")
+    else:
+        # Nothing to tear down. Everything below — soft-delete, queue cancel,
+        # Redis keyspace clearing, name reservation — keys off the agent NAME,
+        # so the delete completes normally without a container (#1747).
+        logger.info(
+            f"Deleting {agent_name} with no container present (#1747) — "
+            "proceeding with the ownership-row delete"
+        )
 
     # BACKLOG-001: in-flight queued tasks can't be recovered (the
     # container is gone), so cancel them now rather than waiting for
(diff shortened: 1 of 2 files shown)


vybe (approved):

Issue #1106, opened 2026-06-09
feat(loops): web UI for sequential agent loops (#740 Phase 2)


vybe:
## Summary

Sequential agent loops (#740, Phase 1) shipped as a **headless, API-and-MCP-only** capability — there is no web UI. Today a user can only start, observe, or stop a loop via `POST /api/agents/{name}/loops`, the `run_agent_loop` / `get_loop_status` / `stop_loop` MCP tools, or raw REST. The only thing visible in the existing UI is the side effect: each iteration appears as an ordinary execution row in the fleet executions dashboard tagged `triggered_by="loop"`. This issue adds the Phase 2 UI surface so loops are usable from the web app.

## Context

Phase 1 explicitly deferred the UI ("Timeline integration: iterations appear as normal `schedule_executions` rows tagged with `loop_id` — no dedicated dashboard surface in Phase 1"). Confirmed against `dev`: no `Loops` view, no `stores/loops.js`, and no frontend reference to `agent_loops` / `run_agent_loop` / `loop_id`. Without a UI, the feature is invisible to operators who don't use MCP, and there's no way to see "loop X, run 3 of 10," watch progress live, or stop a runaway loop from the browser.

The backend, REST endpoints, and MCP tools already exist on `dev` (PR #902) — this is a frontend-only slice over existing plumbing.

## Acceptance Criteria

- [ ] **Start** — a "Run loop" affordance on the Agent Detail page with a form for the documented inputs: `message` template (with `{{run}}` / `{{previous_response}}` helper text), `max_runs` (1–100), `stop_signal`, `delay_seconds`, `timeout_per_run`, `model`, `allowed_tools`. Posts to `POST /api/agents/{name}/loops`.
- [ ] **List** — the agent's loops are listed with status (`queued` / `running` / `completed` / `stopped` / `failed` / `interrupted`), `runs_completed / max_runs`, and `stop_reason`, via `GET /api/agents/{name}/loops`.
- [ ] **Detail / live progress** — a loop detail view shows per-run summaries (run number, status, cost, duration, response preview) and the last full response, via `GET /api/loops/{id}`; updates live (WebSocket `loop_run_completed` / `loop_completed` events already emitted by `loop_service.py`).
- [ ] **Stop** — a Stop control that calls `POST /api/loops/{id}/stop` and reflects the cooperative `stopping` → `stopped` transition.
[...]

PR #1750, opened 2026-07-23
feat(ui): Sessions view — rooms rail, transcript, participants, human participation (ent#170)

dolho:
The operator UI for shared multi-agent sessions. Builds on the ent#169 backend (Abilityai/trinity-enterprise#210 + Abilityai/trinity#1740).

**This was blocked until now** — when I picked up #170 the rooms backend didn't exist. It does now and runs on the dev instance, so the UI is built against a live, verified contract rather than invented from the epic.

## What it is

A top-level `/sessions` view — a human and several agents working one topic in a persistent transcript. Three panes:

- **Rooms rail** — active/closed list (live dot, participant + message counts) and a New-session dialog: pick agents from the accessible roster, set message/cost/TTL budgets, optional scribe.
- **Transcript** — sender row per message with an Agent/You badge + tinted-initials avatar (`PortalAvatar`), portal bubble styling, `@mentions` highlighted, system event lines, per-agent metadata (execution link), and a typing-dots "working" indicator.
- **Participants rail** — presence dots, budget meters that warn near exhaustion, expiry countdown.
- **Composer** — `@mention` autocomplete over participants, the "mention wakes an agent" hint, optimistic append rolled back on failure, `Idempotency-Key` on send.

## Gating (portal precedent)

Gated Vue in the OSS bundle: the route + NavBar entry appear only when `shared_sessions` is in `enterprise_features`, and the router guard catches a direct URL. **Ships dark in OSS/unentitled builds** and lights up only once the ent#169 backend is entitled — no coupling to the submodule being present. Invariants followed: single `api.js` client (#7), domain store (#6), WS-through-the-bus (#10), DOMPurify markdown (H-005).

## Live updates

Follows the loops pattern (#1106): the store reacts to fleet-wide `room_message` / `room_participant_state` / `room_closed` events for the room on screen and refetches over REST (thin WS payloads, the #918 pattern), plus a 12s backstop poll while any agent is working.

## Verification — and its honest limit

- **All 6 SFCs + the 3 edited files compile** (`@vue/compiler-sfc`), store is valid JS, HMR applied with no errors.
- **Contract verified against the running ent#169 backend**: every field the components read — room (`status`/`stop_reason`/`cost`/`max_messages`/`expires_at`/…), each message (`seq`/`sender_kind`/`sender_identity`/`kind`/`mentions`/`execution_id`), each participant (`kind`/`identity`/`role`/`left_at`) — is present in the live `/api/rooms` and `/api/rooms/{id}` responses. There is a real room on the instance (a 3-way agent conversation) for the panes to render.
- **I could not visually confirm the render** — no browser tooling in this session. The above is contract- and compile-level verification, not a screenshot. The e2e-smoke AC (create → mention → response renders with identity + working lifecycle) needs a human or Playwright pass. Flagging plainly rather than claiming it works on screen.

## Known gaps (deliberate, stated)

- **Per-message cost/duration not shown.** The room-detail response returns an `execution_id` per message and a room total, but not per-message cost — so the metadata line links the execution without its cost. A small ent#169 backend addition (join `schedule_executions`) would fill it.
- **No artifact/report card** for messages referencing a shared file or report — the backend doesn't surface that linkage on a message yet.

Both are backend-surface gaps, not UI shortcuts; noted for the follow-up rather than faked in the UI.

### DIFF
--- a/src/frontend/src/components/NavBar.vue
+++ b/src/frontend/src/components/NavBar.vue
@@ -71,6 +71,17 @@
                  is hidden entirely. The store's `featureFlagsLoaded`
                  guard means the link doesn't flicker on first paint
                  (loadFeatureFlags fires in onMounted below). -->
+            <!-- Shared sessions (ent#170) — visible only when the rooms
+                 backend is entitled (shared_sessions). Hidden entirely in
+                 OSS/unentitled builds, like every enterprise surface. -->
+            <router-link
+              v-if="enterpriseStore.isEntitled('shared_sessions')"
+              to="/sessions"
+              class="border-transparent text-gray-500 dark:text-gray-400 hover:border-gray-300 dark:hover:border-gray-600 hover:text-gray-700 dark:hover:text-gray-200 inline-flex items-center px-1 pt-1 border-b-2 text-sm font-medium"
+              :class="{ 'border-blue-500 dark:border-blue-400 text-gray-900 dark:text-white': $route.path.startsWith('/sessions') }"
+            >
+              Sessions
+            </router-link>
             <router-link
               v-if="enterpriseStore.hasAnyEnterprise"
               to="/enterprise"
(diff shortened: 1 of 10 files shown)


vybe (approved):
Validated via subagent review: entitlement gating verified at route+NavBar+guard (ships dark on OSS), sole v-html through DOMPurify, frontend invariants honored. Known non-blocking: mention-highlighter post-sanitize string surgery (follow-up), ttl_hours '' serialization nit, no visual pass yet — needs a Sessions smoke on an entitled instance before release notes claim it.

PR #1751, opened 2026-07-23
security: validate archive-supplied .mcp.json on deploy-local (ent#213)

dolho:
Related to Abilityai/trinity-enterprise#213 (private security tracker). OSS code, so the fix lands here.

## The asymmetry

`deploy-local` copied an archive `.mcp.json` into the agent workspace **without** `validate_mcp_config()`; the post-deploy inject path (`routers/credentials.py`) validates it. Claude Code auto-loads `~/.mcp.json` via `--mcp-config` on the next execution, so an unvalidated archive config was an ingress the inject-path guard (#598, AISEC-C2 Layer 2) never covered — command substitution, shell metacharacters, a `${VAR}` reference to a reserved var (`LD_PRELOAD`/`PATH`/…), oversize, unknown fields.

Surfaced by an external security report. **Severity is defense-in-depth / consistency, not critical RCE** (the issue is explicit on this): the demonstrated PoC runs an allowlisted runtime within the deploying creator's own trust boundary and would pass the validator anyway — blocking code execution via approved runtimes is the deferred sandboxed-MCP Layer-3 work. This PR closes the validation **asymmetry** between the two ingresses; it does not, alone, stop an allowlisted-runtime server.

## Fix

Same validator, same 400 shape as the inject path, run **before** `_prepopulate_workspace_from_template` writes the file into the volume — a validate-after-copy would leave the malicious config on disk.

Extracted the guard into a small helper `_validate_archive_mcp_config` so it's unit-testable without the Docker-bound deploy flow, and a wiring test pins that it runs before pre-pop.

## Parity check (the issue asked for it)

- **`.credentials.enc` archive** — already validates (`credential_encryption.py:66-73`). ✓
- **git_service `.mcp.json` references** — gitignore/allowlist patterns and a comment, not a workspace ingress. ✓
- **deploy-local** — was the only gap. Fixed.

## One thing worth calling out

While writing the tests I assumed setting `env: {"LD_PRELOAD": "/tmp/x.so"}` would be rejected. It isn't — the validator's reserved-env gate blocks a `${LD_PRELOAD}` **reference** in a value, not setting the key literally. That's not a hole in this fix (my change just calls the same validator both paths use — parity is the whole point), but it's a reminder that the threat model is `${VAR}`-reference-scoped, and setting a literal reserved env key is currently allowed. If that should also be blocked, it belongs in `mcp_validator.py` and would then apply to both ingresses automatically. Flagging rather than silently widening scope.

## Tests

`tests/unit/test_213_deploy_mcp_validation.py` — 13 passed. Reject path (malformed JSON, command substitution, shell metacharacters, blocked runtime, reserved env-ref, unknown field, `trinity` redefined as stdio, oversize) each 400 with a named reason; accept path (valid stdio, empty `mcpServers`, canonical `trinity` entry); and the wiring/ordering guard.

Related suites: 357 passed, 3 skipped. Backend imports `deploy.py` cleanly (the new import + helper).

### DIFF
--- a/src/backend/services/agent_service/deploy.py
+++ b/src/backend/services/agent_service/deploy.py
@@ -33,6 +33,7 @@ from services.docker_service import get_agent_container
 from services.docker_utils import container_stop
 from utils.helpers import sanitize_agent_name
 from services.settings_service import get_agent_quota_for_role
+from services.mcp_validator import validate_mcp_config, McpValidationError
 from .helpers import get_agents_by_prefix, get_next_version_name, get_latest_version
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


vybe (approved):
Validated: closes the deploy-local .mcp.json validation asymmetry with the same validator + 400 shape as the inject path; guard runs before workspace pre-populate (pinned by test). CodeQL py/path-injection dismissed as false positive — dest_path is sanitize_agent_name'd (deploy.py:412) + normalized-containment-checked (:478-499), filename is a constant. Cross-tracker ent#213 needs a manual close at release.

PR #1757, opened 2026-07-23
feat(git): PAT-free clone of public github: templates (trinity-enterprise#123)

vybe:
## Summary
- Creating an agent from a **public** `github:owner/repo` template now succeeds with **no GitHub PAT** anywhere — anonymous, read-only, **source-mode** clone. Gating prerequisite for the fresh-install fleet-provisioning epic (abilityai/trinity-enterprise#122 → first-run manifest seed ent#124, UI manifest install ent#126).
- Visibility is validated over the **git transport** (credential-less `git ls-remote` probe — same transport as the real clone, immune to the anonymous REST 60/hr cap): definitive miss → combined named 400 *"not found or private — add a GitHub token"*; transient → **fail-closed 502** (monitoring is default-off, a fail-open would ship a silently empty agent).
- Every push surface fails honestly for a tokenless agent: backend `sync`/`reset` return a named `no_write_credentials` 409 (baked env **or** per-agent PAT row — covers the #1264 live-injection window), the container's push remote is blackholed with a self-describing URL, and the MCP `git_sync` hint stops suggesting `chat_with_agent` (chat can't fix missing credentials).
- Closes a second silent-empty-agent route while at it: the #1559 rebuild-recovery seam (`_apply_persisted_auth_env`) gated github env on `pat AND repo` — now repo-only, PAT-conditional, with source-mode/branch re-derivation.

## Changes
- `src/backend/services/agent_service/crud.py` — tokenless gate (`""`→`None` normalization, source-mode-only 400 incl. explicit `None`), owner/repo charset guard, tokenless ls-remote validation branch, repo-only env assembly (token vars + `GIT_SYNC_AUTO` require a PAT)
- `src/backend/services/agent_service/lifecycle.py` — rebuild-seam gate flip
- `src/backend/services/git_service.py` — `probe_anonymous_repo_access`, `_agent_has_write_credentials` + guards; `src/backend/routers/git.py` — reset 409 mapping
- `docker/base-image/startup.sh` — repo-only clone gate, conditional `CLONE_URL`, `GIT_TERMINAL_PROMPT=0`, push blackhole, workspace-`.env` PAT fallback, private-repo failure cause
- `src/mcp-server/src/tools/git.ts` — `no_write_credentials` hint suppression
- Docs: `requirements/github.md` §11.11, `architecture.md`, `github-sync.md` + `mcp-git-tools.md` flows, `learnings.md`

## ⚠️ Deploy note
**Base-image rebuild is a hard ordering requirement.** An un-rebuilt image still has the old `GITHUB_PAT`-gated clone block — tokenless creation on it yields an empty agent with no failure marker. `./scripts/deploy/build-base-image.sh` before (or with) backend rollout.

## Test Plan
- [x] New: `pytest tests/unit/test_ent123_tokenless_clone.py -v` — 50 tests (gate, charset guard, probe outcomes, env assembly, rebuild seam, stderr classification, write-credential guard, startup.sh static)
- [x] Pinned regressions green (130): `test_1484_create_agent_characterization` · `test_1439_clone_status_health` · `test_1574_gh_token_wiring` · `test_ent162_per_user_github_pat` · `test_1264_per_agent_pat_propagation` · `test_fork_to_own`
- [ ] `/verify-local` (base-image build + real agent boot) recommended before release — the anonymous in-image clone is the one path unit tests can't prove

Refs abilityai/trinity-enterprise#123 (cross-repo — closed manually at release)


dolho:
## /review report — `feature/ent123-pat-free-clone` → `dev`

**Files**: 13 (+1276/−63) · **Merge-base**: `7508fc57` · **Scope**: CLEAN · **Tests**: 43

### Critical findings: **0**

I chased one suspected command-injection path hard and it does **not** hold — recording the trace so a future reviewer doesn't have to redo it.

**Investigated and REJECTED — slash-less injection bypass**
The new charset guard is conditional:
```python
if "/" in template_str and not _GITHUB_REPO_PATH_RE.match(template_str):
```
so a slash-less value skips it entirely (`foo;whoami` and `foo$(id)` both pass), and `startup.sh:165` runs the clone through `eval` with unquoted interpolation. That looked like RCE in the agent container.

**It isn't reachable**: `crud.py:423` rejects a slash-less `repo_path` with a 400 before it can become `GITHUB_REPO`. So both paths are closed — slash present → charset guard; slash absent → 400. Worth knowing the two guards are *jointly* load-bearing: weakening either (e.g. relaxing the 400 to a warning) reopens the `eval`.

### Verified clean (evidence, not assumption)
- **Injection surface** — `probe_anonymous_repo_access` uses `create_subprocess_exec` (no shell) with a fixed `base` prefix, so the repo arg can never start with `-` or reach a shell
- **Branch input** — validated by the pre-existing GIT-002 `isalnum()` check; an invalid branch is dropped to `None`, so no metacharacter reaches `CLONE_BRANCH`
- **PAT exposure in logs** — clone failures echo through `sed "s|oauth2:[^@]*@|oauth2:***@|g"`, so the token is scrubbed before it reaches container logs (which Vector captures). This is the failure path most likely to leak, and it's covered.
- **Fail-closed on transient** — a probe that can't reach GitHub returns 502 rather than admitting the create. Correct direction: monitoring is default-off, so a fail-open would ship a silently empty agent.

Nice catch in the PR itself: recognising that removing the PAT-ful REST validation also removed an *incidental* injection barrier, and replacing it with an explicit one. That's the kind of second-order consequence that usually ships unnoticed.

### 🟡 Informational

**I1 — stderr substring matching is locale/version fragile** (Confidence 7/10)
`_ANON_PROBE_DEFINITIVE_PATTERNS` matches English git stderr (`"repository not found"`, `"authentication failed"`, …). A localised or reworded git build would fall through to `"transient"` → **502**. That's the *safe* direction (fail-closed, no security impact), but it turns a clean "private repo — add a token" 400 into a confusing infrastructure error. Consider also treating a non-zero exit with *empty* stderr as definitive, or pinning `LC_ALL=C` on the subprocess so the messages are stable.

**I2 — the probe inherits the full backend environment** (Confidence 6/10)
`env={**os.environ, "GIT_TERMINAL_PROMPT": "0"}` hands every backend secret to the git subprocess. Nothing exploits it today, but the whole point of this probe is that it is *credential-less* — passing a minimal env (`PATH`, `HOME`, `GIT_TERMINAL_PROMPT`, `LC_ALL`) would make that property structural rather than incidental, and would rule out an inherited `GIT_*` / proxy var quietly changing probe behaviour.

### Summary
**Critical: 0** · Informational: 2 · Scope: clean · 43 tests

The security-sensitive parts (charset guard, branch validation, PAT redaction, fail-closed probe) are all genuinely handled — I verified each rather than assuming.

dolho (approved):

PR #1758, opened 2026-07-23
feat(slack): per-agent bot routing seam — OSS half of ent#222

dolho:
The **OSS companion** to the enterprise per-agent Slack bots feature (trinity-enterprise#222, enterprise PR trinity-enterprise#232). Open-core split: OSS owns the seam, the private module owns the policy — same shape as the connector (#118), rooms (#1740), skill-runner (#1735).

## What this adds
- **`adapters/per_agent_bot.py`** — an **inert-by-default** hook registry (`set_resolver` / `set_token_provider`). With no resolver registered (OSS-only builds, or before the enterprise module loads) **every path is a no-op and channel routing is byte-for-byte unchanged**. Both hooks **fail open** — an error falls back to normal routing, so a bug here can never take Slack offline.
- **`message_router._resolve_agent_and_token`** — consult the per-agent resolver first; a match wins over the channel binding and replies through the per-agent bot token. Falls through to channel routing when there's no binding / no token.
- **`slack_adapter.parse_message`** — stamp the **receiving** bot identity (`authorizations[0].user_id` + `api_app_id` + `team_id`) into `message.metadata`, so the resolver can key on which bot got the event. Additive; inert for every non-Slack channel.

## Tests
5 unit tests: inert-by-default, resolver reads recipient bot, non-Slack ignored, both hooks fail-open, token provider. **Green.**

## Honest scope
- The **inert-by-default** guarantee means this is safe to merge ahead of the enterprise side — it changes nothing until an enterprise resolver registers.
- **Needs live verification:** the end-to-end path (a real Slack event routing to a per-agent bot) can't be proven by mocking — the `authorizations` recipient-bot extraction is the live point. Verified in isolation; end-to-end verification needs a live per-agent Slack app from the design partner.
- **Still to come in this OSS PR (or a follow-up):** the enterprise-gated Vue channel panel (UI) to configure a per-agent bot. The backend seam is the keystone; the UI is additive.

Pairs with: trinity-enterprise#232 (module + transport manager), which registers the resolver + token provider into this seam.

### DIFF
--- a/src/backend/adapters/message_router.py
+++ b/src/backend/adapters/message_router.py
@@ -784,6 +784,20 @@ class ChannelMessageRouter:
         (the caller short-circuits). Both return silently — there is no token to
         reply with on the no-token path.
         """
+        # ent#222: if the event was received by an agent's DEDICATED Slack bot,
+        # that agent wins over the channel binding and its own bot token is used
+        # for the reply. Inert for non-Slack / OSS-only (no resolver registered).
+        from adapters import per_agent_bot
+        pab_agent = per_agent_bot.resolve_from_message(message)
+        if pab_agent:
+            pab_token = per_agent_bot.get_token(pab_agent)
+            if pab_token:
+                logger.debug(f"[ROUTER:{channel}] per-agent bot → agent {pab_agent}")
+                return pab_agent, pab_token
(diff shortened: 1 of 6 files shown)


dolho:

Issue #1609, opened 2026-07-14
Make proactive channel-message rate limits configurable (Slack/Telegram/DM guardrails)


vybe:
## Summary

The proactive channel-messaging rate limits (10 messages/hour/channel, 100 messages/hour/agent for Slack and Telegram; 10 messages/hour/recipient for proactive DMs) are hardcoded module constants. They were introduced as anti-spam guardrails (#349, #350, #321), but they cannot be tuned, which blocks legitimate high-volume autonomous use cases — e.g. an agent that posts a Slack message for each inbound support request. Make them admin-configurable with the current values as defaults.

## Context

A community report hit these caps building an autonomous Slack-posting agent and asked whether they are configurable (they are not) and whether the enterprise edition lifts them (it does not — they are OSS guardrails, not entitlement gates).

The hardcoded constants:

- `src/backend/routers/slack.py` — `_SLACK_PROACTIVE_PER_CHANNEL = 10`, `_SLACK_PROACTIVE_PER_AGENT = 100` (gates `POST /api/agents/{name}/slack/channels/{channel_id}/messages`, backing MCP `send_group_message`)
- `src/backend/routers/telegram.py` — `_PROACTIVE_RATE_LIMIT_PER_GROUP = 10`, `_PROACTIVE_RATE_LIMIT_PER_AGENT = 100` (same caps, bespoke in-memory bucket)
- `src/backend/services/proactive_message_service.py` — `RATE_LIMIT_MAX_PER_HOUR = 10` per agent-recipient pair (backing MCP `send_message`)

The codebase already has the pattern to follow: the inbound channel limiter reads `channel_rate_limit_max` / `channel_rate_limit_window` from `settings_service` with hardcoded fallbacks (`src/backend/adapters/message_router.py`), and `max_parallel_tasks_ceiling` (#506) established the admin-tunable-guardrail shape (runtime-resolved from `system_settings`, range-validated, no per-process cache so `--workers 2` stays consistent, no migration needed).

Note: replies to inbound Slack/Telegram messages (DM/@mention/thread via the channel adapter) are NOT gated by these caps — only agent-initiated proactive sends are. That distinction should land in the docs as part of this issue.

## Acceptance Criteria

- [ ] The three proactive cap families (Slack per-channel/per-agent, Telegram per-group/per-agent, proactive-DM per-recipient) are admin-configurable, read at request time via `settings_service` (no restart required, `--workers 2`-consistent)
[...]

PR #1761, opened 2026-07-24
feat(slack): per-channel proactive-messaging consent (ent#223)

dolho:
Backend for **ent#223** — the partner fast-follow from the 2026-07-23 call. Built **OSS-core** (confirmed with the owner): the flag lives on `slack_channel_agents`, an OSS table, and **Invariant #3** forbids an enterprise migration from ALTERing an OSS table — so consent here is an edition-agnostic safety primitive (the `users.suspended_at` #995 pattern).

## Problem
Proactive consent was per-**recipient** (`agent_sharing.allow_proactive`, keyed by verified email). In an open Slack workspace nobody authenticates, so there's no recipient record to opt in — and channel posts had **no consent gate at all**, only the owner check + #1609 rate caps.

## What changed
- `slack_channel_agents.allow_proactive` — dual-track migration (SQLite `db/migrations.py` + Alembic `0029`), plus `schema.py` DDL and `tables.py` MetaData.
- **Default posture, deliberately split so nothing silently flips:**
  - **NEW** bindings → `0` (deny). Binding an agent to a channel isn't itself consent to unprompted posts.
  - **EXISTING** bindings → backfilled to `1` (allow), because channel posts worked with no gate before this. The AC is explicit that existing behavior must not silently flip.
  - The backfill runs only when *this* caller added the column, so an operator's later explicit **OFF survives re-runs** (tested).
- **Enforcement** in `send_agent_slack_channel_message`, placed between the **404** (not bound) and the **429** (rate capped) so all three refusal reasons stay distinguishable, returning a named actionable `proactive_not_allowed`. Rate caps still apply on top.
- **Toggle endpoint** `PUT /api/agents/{name}/slack/channels/{channel_id}/proactive` (owner-gated) so it isn't API-only.

## Tests
8 unit tests — default-deny for new bindings, flag on both read paths, toggle + scoping, unknown-binding miss, and the migration's no-silent-flip backfill, post-migration default, and idempotency. Green on SQLite (harness also runs PG when `TEST_POSTGRES_URL` is set).

## ⚠️ Alembic head collision to resolve at merge
This is `0029_...` off `0028_agent_reminders`. My open PR #1752 also adds a `0029_agent_evaluations` off the same parent — **two heads would break boot**. Whichever merges second needs a one-line `down_revision` rebase.

## Not in this PR (remaining ent#223 AC)
- The **channel-binding UI toggle** (`SlackChannelPanel.vue`) — the endpoint is here, the switch isn't yet.
- **MCP `send_group_message`** relaying the named reason to the agent.
- **Docs** (Slack proactive = per-channel consent; DM stays per-recipient).

Unblocks ent#224 (completion report-back is gated on this consent).

### DIFF
--- a/docs/memory/feature-flows/proactive-messaging.md
+++ b/docs/memory/feature-flows/proactive-messaging.md
@@ -11,6 +11,18 @@ Enables agents to send proactive messages to specific users by verified email ac
 ## Key Features
 
 - **Explicit opt-in consent**: Users must enable `allow_proactive` flag on their sharing record
+- **Two consent units, by channel shape (ent#223)**: a **DM** to a person is consented
+  **per recipient** (`agent_sharing.allow_proactive`, keyed by verified email). A post to a
+  **Slack channel** is consented **per channel binding**
+  (`slack_channel_agents.allow_proactive`) — in an open Slack workspace users never
+  authenticate, so there is no verified-email recipient record to opt in, and the channel
+  is the only meaningful consent unit. Default posture: a **new** binding denies (binding
+  an agent is not itself consent); **existing** bindings were backfilled to allow, because
+  channel posts had no consent gate before ent#223 and must not silently break. Toggled in
+  the Slack channel panel, or via
+  `PUT /api/agents/{name}/slack/channels/{channel_id}/proactive`. Refusal is a named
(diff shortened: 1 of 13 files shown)


dolho:

PR #1762, opened 2026-07-24
feat(slack): report a finished task back to its originating channel/thread (ent#224)

dolho:
Closes the partner scenario from the 2026-07-23 call: *user asks in Slack → agent delegates a ~10-minute job → job succeeds → **nobody tells the user**.*

> **Stacked on #1761** (ent#223 consent) — it needs the per-channel consent gate. Base is the #1761 branch; retarget to `dev` once that merges.

## The issue called this "a join". It wasn't — two things were missing
1. **Channel context didn't survive delegation.** The A→B path never forwarded `source_channel*`, so B's row had **no destination**. `ParallelTaskRequest` gains `parent_execution_id`; `chat_execution_service` copies the parent's channel/chat/thread down (fail-open). MCP `chat_with_agent` gains an optional `execution_id`, mirroring the #1084 effect tools.
2. **Nothing reported at the terminal.** New `channel_completion_report` posts the outcome to the originating channel/thread, hooked at **both** CAS-won chokepoints beside `spawn_task_terminal_event` (#1578) — **success and failure**, since a silent failure is the bug.

## ⚠️ The load-bearing safety rule
A direct channel turn is **synchronous and already replies inline**, and those rows carry `triggered_by` == the channel name. Reporting them would **duplicate every normal Slack reply in a customer workspace**. So we report **only when the context was INHERITED** (`triggered_by` not in `{slack, telegram, whatsapp}`). That single condition is what keeps this from spamming the partner.

Also: ent#223 consent-gated · at-most-once via `effect_guard` (#1084) keyed on the **resolved destination**, never the generated body · Slack-only in v1 · fail-soft (a reporting failure never disturbs an execution that already completed and was billed).

## Tests — 12, green
Double-post guard across all three channel triggers · delegated reporting into the right thread · consent + unbound gates · no-context / non-Slack / missing-row no-ops · failure terminals report · replay doesn't re-post · never-raises.

## Honest limits
- **Not verified against live Slack** — the send path is mocked in tests. The posting primitive (`send_message_detailed`) is the same one #1649 already uses in production, but an end-to-end run needs a real workspace.
- **Opt-in by construction**: nothing reports until an agent passes `execution_id` when delegating *and* the channel has consent enabled.

Related to ent#224 · needs #1761

### DIFF
--- a/src/backend/models.py
+++ b/src/backend/models.py
@@ -253,6 +253,11 @@ class ParallelTaskRequest(BaseModel):
     resume_session_id: Optional[str] = None  # Claude Code session ID to resume (EXEC-023)
     inject_result: Optional[bool] = False  # If true and self-task, inject result as message in originating chat session (SELF-EXEC-001)
     files: Optional[List[WebFileUpload]] = None  # File attachments (#364)
+    # ent#224: the CALLER's current execution id. When agent A delegates to B,
+    # B inherits A's originating channel/thread from this row, so B's completion
+    # can be reported back to the Slack thread the work actually came from.
+    # Optional and fail-open — absent means "no channel context to inherit".
+    parent_execution_id: Optional[str] = None
 
 
 # ============================================================================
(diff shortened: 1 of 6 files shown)


dolho:

Issue #1259, opened 2026-06-18
Epic: Execution correctness & test health


vybe:
**Objective:** Green, trustworthy CI + no silent execution/telemetry loss. Focused epic carved from the ready backlog (2026-06-18 prioritization pass). Do the test-health items first — you can't verify any deck-clear without trustworthy tests.

## Sub-issues (all currently `status-ready`)
- #1134 — e2e smoke drift: Operations consolidation (#1109) broke operating-room smoke; ensure-default modal races api-keys spec
- #1103 — Unit suite: 15 pre-existing failures unmasked by collection-abort fix (quarantined)
- #767 — CB probe executions left open until backend restart inflate failure duration on timeline
- #1201 — Agent-side timeout (504) drops execution telemetry — cost/context/tool-calls lost on timed-out runs
- #799 — SUB-003 auto-switch has no per-agent lock — concurrent 429s race the restart
- #1160 — Migration runner: DROP-rebuild data-loss window and no cross-process serialization

**Suggested order:** #1134 → #1103 (restore trustworthy CI) → #1160 (data-loss window) → #767 → #1201 → #799 (execution correctness).

Issue #804, opened 2026-05-12
bug(tests): security conftest setdefault never overrides global REDIS_BACKEND_PASSWORD='test' sentinel


AndriiPasternak31:
## Summary

The three Redis ACL acceptance tests in `tests/security/test_redis_network_isolation.py` fail against any live stack with `b'NOAUTH Authentication required.\n\n'`. Root cause is a fixture-precedence bug: the global `tests/conftest.py` `setdefault`s `REDIS_BACKEND_PASSWORD="test"` at import time, and `tests/security/conftest.py` then tries to override it from `.env` but also uses `setdefault` — which is a no-op when the key already exists. The bogus `"test"` value sticks, so `redis-cli --user backend -a test` is sent to Redis, which (correctly) rejects it.

The Redis ACL itself is healthy. Confirmed in the phoenix worktree against a clean stack: `docker run --rm --network trinity-platform-network redis:7-alpine redis-cli -h redis --user backend -a $REAL_PWD PING` → `PONG`. The bug is purely in the test fixture wiring.

## Context

Introduced by commit `6841e72b` — *fix(security): lock down Redis — auth + ACL + network split (#589)* — on 2026-05-04. Both halves of the bug landed in that one commit:

1. `tests/conftest.py:30` — `_os_589.environ.setdefault("REDIS_BACKEND_PASSWORD", "test")` (runs first, at global pytest import, poisons the env)
2. `tests/security/conftest.py:25` — `os.environ.setdefault(key, value)` for the `.env` load (runs later, can't override because `setdefault` is a no-op when the key already exists)

Follow-up `3d15b5fa` (#777 — *skip-when-unset guard for REDIS_BACKEND_PASSWORD (#764)*) added a per-test `backend_password` fixture with `pytest.skip("REDIS_BACKEND_PASSWORD not set")`, but the guard only skips on **absence** — the bogus `"test"` value satisfies the guard, so tests run with wrong creds and fail.

This means the Redis ACL acceptance tests have not actually validated the ACL since #589 merged. They've been a green silence in the "skip" path or a red noise in the "live stack" path, never the green truth they were written to be.

Failing tests (all 3, against a live stack):
- `test_platform_container_can_authenticate`
- `test_backend_acl_blocks_flushall`
- `test_backend_acl_blocks_config_get`

Pytest traceback header reveals the smoking gun: `backend_password = 'test'`.

## Acceptance Criteria

[...]

PR #1779, opened 2026-07-26
fix(tests): resolve the integration Redis target once, and fail instead of skip (#1775)

### LINKED ISSUE #1775: bug: integration suite runs only 6 of 83 tests when collected as a directory — 57 silent Redis skips
## Summary

`pytest integration` collected as a whole directory runs **6 of 83 tests**. 77 skip, 57 of them because Redis is reported unreachable at `localhost:6379` — even when Redis is up and `REDIS_URL` correctly points at the live instance. pytest still **exits 0**, so every caller that trusts the exit code reads this as a green suite.

Run individually, the same tests pass. This is a collection-order / module-import interaction, not a Redis outage.

## Evidence

Against a healthy stack (Redis up and healthy, published on an ephemeral host port; backend healthy; independently confirmed with `redis.ping()` from inside the backend container):

```
# whole directory, REDIS_URL correctly pointed at the live Redis
$ pytest integration -q -rs
6 passed, 77 skipped, 2 deselected in 0.66s

# the SAME env, one module alone
$ pytest integration/test_1560_breaker_lifecycle.py -q -rs
4 passed, 1 skipped in 0.17s          # the 1 skip is a real feature gate
                                      # (TRINITY_LIFECYCLE_TEST_AGENT unset)
```

Skip-reason breakdown for the directory run:

| count | reason |
|---|---|
| 41 | `Redis unavailable: Error 61 connecting to localhost:6379. Connection refused.` |
| 16 | `Redis unreachable at redis://backend:***@localhost:6379: Error 61 ... Connection refused.` |
| 5 | `Container agent-testfix not found — Phase 2 tests need a recreated testfix` |
| 4 | `Agent testfix not visible via API (404)` |
| 11 | assorted legitimate feature gates (`TRINITY_*_TEST_AGENT` unset, no PostgreSQL, alembic missing) |

The `localhost:6379` in those messages is the giveaway: the harness supplied an ephemeral port, not 6379.

## Likely mechanism

At least five integration modules carry the same import-time fallback:

- `integration/test_1557_autonomy_inbound.py`
- `integration/test_1560_breaker_lifecycle.py`
- `integration/test_circuit_breaker.py`
- `integration/test_dispatch_breaker.py`
- `integration/test_monitoring_service.py`

e.g. `tests/integration/test_1560_breaker_lifecycle.py:64-71`:

```python
# Honor a pre-set REDIS_URL for sibling stacks.
if "REDIS_URL" not in os.environ:
    _PASSWORD = _env_value("REDIS_BACKEND_PASSWORD")
    ...
    os.environ["REDIS_URL"] = f"redis://backend:{_PASSWORD}@localhost:6379"
```

This **mutates `os.environ` at import time**, so whichever module is collected first fixes `REDIS_URL` for the entire pytest session. Consumers that read `os.environ["REDIS_URL"]` in a fixture (`test_circuit_breaker.py:89`, `test_monitoring_service.py:191`) then get the hardcoded `:6379` regardless of what the harness exported, and skip on connection refused.

The `:6379` literal is also wrong for any sibling/parallel stack by construction — those publish Redis on an ephemeral port precisely so concurrent runs don't collide.

## Impact

Silent, not loud. Because the tests `pytest.skip()` rather than fail, pytest exits 0 and the suite reads as green:

- Any local gate or CI job that scores the integration suite on its **exit code** is currently certifying branches on roughly **7% of that suite** without saying so.
- The failure mode is invisible in the summary line — `6 passed` looks like a pass, not like 77 tests that never ran.

Same class as a standalone-only test that errors inside a full collection: correctness depends on collection order, which nothing in the suite pins.

## Suggested fix

1. **Stop mutating `os.environ["REDIS_URL"]` at module import.** Resolve Redis config in a fixture (or one shared `conftest` hook) so collection order cannot poison the session.
2. **Drop the hardcoded `localhost:6379` fallback**, or derive it from the same discovery the harness uses. A fixed port is wrong for every sibling-stack and parallel-worktree run.
3. **Make an unreachable Redis fail rather than skip** when the harness explicitly declared a Redis target (i.e. `REDIS_URL` was set). A skip there is lost coverage, not an absent optional dependency — and it is what makes this bug silent.

Item 3 is the one that stops this class of problem recurring: as long as infrastructure failures degrade to skips, any future breakage will read as green too.

## How this was found

Our local pre-push verification harness was changed to classify every pytest skip and treat infrastructure-caused skips as failures rather than counting them as benign. The first real run under that rule turned these 57 silent skips into a red stage.

AndriiPasternak31:
## Summary

`pytest tests/integration` collected **as a directory** ran **6 of 83 tests** — 57 skipped claiming Redis was unreachable when it was in fact up — and **pytest exited 0**. Any caller scoring the suite on its exit code certified a branch on ~7% of it. Run per-module, the same tests passed, so this was a collection-order interaction, not a Redis outage.

Two independent collection-time global-state mutations, not one. The issue (and the first draft of the plan) saw only the first.

**M1 — the clobber.** `test_monitoring_service.py` wrote `os.environ["REDIS_URL"]` **unconditionally at module scope**. pytest imports every test module during *collection*, so that single write re-pointed the harness-supplied target for the whole session. Deleted rather than guarded: a module that ERRORs later in its import has *already* executed the write, so the write must not exist at import time at all.

**M2 — the root-conftest sentinels.** `tests/conftest.py` (#589) `setdefault`s a dummy `REDIS_URL` plus `"test"` passwords at global import, then preloads backend modules — and `src/backend/config.py:85` snapshots `REDIS_URL = os.getenv(...)` at *module* scope. Three consequences: `"REDIS_URL" in os.environ` is unconditionally true by the time any child conftest runs (so it cannot distinguish "the harness told us" from "the root conftest defaulted"); `setdefault` on the passwords is a no-op once `"test"` is in place, so the sentinel must be **deleted** before real `.env` values can be overlaid; and re-pointing `os.environ` alone leaves `config.REDIS_URL` frozen to the dummy — which fails the breaker Redis client *open* and turned 7 `test_monitoring_service` assertions into `assert 0 == 2`, a failure mode that looked like a product bug and wasn't.

`tests/security/conftest.py` (#804) had already solved the sentinel-delete-then-overlay dance one directory over, with a comment naming the trap. This PR mirrors that pattern rather than inventing a resolver.

## Sentinel proof

Sentinel `REDIS_URL=redis://backend:SENTINELPW@127.0.0.1:6390`, read at `pytest_collection_finish` after `--collect-only`. This is what makes the one-line root cause credible — it isolates the single guilty module.

**Before**

| Proof | Collection scope | `os.environ["REDIS_URL"]` | `config.REDIS_URL` | `agent_client` |
|---|---|---|---|---|
| **A** | all 14 modules | `…@localhost:6379` — **clobbered** | sentinel | **standalone copy** |
| **B** | all **minus** `test_monitoring_service.py` | sentinel — **survives** | sentinel | real |
| **C** | `test_monitoring_service.py` alone | `…@localhost:6379` — **clobbered** | sentinel | **standalone copy** |

B is the discriminator: removing one module restores the target, which pins M1 to that file and clears the other four (they are correctly guarded and innocent).

**After**

| Proof | `os.environ["REDIS_URL"]` | `config.REDIS_URL` | `agent_client` real? |
|---|---|---|---|
| **A** | sentinel — survives | sentinel — re-pointed | ✅ |
| **B** | sentinel — survives | sentinel — re-pointed | ✅ |
| **C** | sentinel — survives | sentinel — re-pointed | ✅ |

`__file__` is **not** a usable discriminator for the `agent_client` column — the standalone copy is built by `spec_from_file_location` from the same path, so both report the same file. The working check is identity against the parent package attribute: `sys.modules["services"].agent_client is sys.modules["services.agent_client"]`.

## Security — the skip/fail messages leaked the raw Redis password

Three sites (`test_1557:93`, `test_1560:140`, `test_dispatch_breaker:73`) interpolated the **raw** `REDIS_URL` — password included — into pytest messages, and **no masking existed anywhere under `tests/`**. The `***` in the issue body was the reporter hand-redacting. This PR converts those skips into *failures* — louder, and retained in CI logs — so masking is a prerequisite, not a nicety. Trinity mandates masked credentials in logs, and this is a public repo.

Masking is applied at every message site. Two further leak paths were found in review and closed — both cases where the masker was correct but a raised exception routed *around* it:

1. **`mask_redis_url` could throw on the input it exists for.** `urlsplit` is lazy: it accepts `redis://u:pw@host:notaport` and only raises when `.port` is read. Those reads sat outside the `try`, so a malformed URL escaped as a `ValueError` — and since every caller passes the raw URL as an argument, pytest rendered the caller's frame with the password in it. Every parse step is now inside the guard.
2. **`redis_client` raised from inside an `except` block.** `pytest.fail(..., pytrace=False)` suppresses its own traceback but not an implicitly chained one — pytest renders "During handling of the above exception…" plus the original frames, and for a URL *parse* error those frames hold the unmasked URL as a local. The failure is now recorded and re-raised after the handler has exited.

Resulting messages keep everything that identifies a misconfiguration and nothing that authenticates:

```
Redis unusable at redis://backend:***@127.0.0.1:6399 (caller-supplied REDIS_URL): ConnectionError: ...
Redis unusable at redis://backend:***@localhost:6379 (derived from .env + localhost:6379): ConnectionError: ...
```

## Changes

| File | Change |
|---|---|
| `tests/conftest.py` | capture `CALLER_SUPPLIED_REDIS_URL` **before** the sentinels are installed — the root conftest is the only place that can know caller intent |
| `tests/integration/conftest.py` | **new** — resolve the target once, delete the `"test"` sentinels then overlay `.env`, re-point `sys.modules["config"].REDIS_URL`, reset the breaker client caches, expose `redis_url` / `redis_required` / `redis_client` |
| `tests/integration/redis_target.py` | **new** — `mask_redis_url` + URL construction, the single `localhost:6379` literal |
| `test_monitoring_service.py` | delete the unconditional write (M1); Layer D (below); use the shared fixtures |
| `test_1557` / `test_1560` / `test_circuit_breaker` / `test_dispatch_breaker` | drop 5 private `.env` readers and their import-time writes; use the shared fixtures; mask |
| `tests/run-integration.sh` | export a real `REDIS_URL` so both harnesses behave identically |
| `tests/unit/test_1775_no_import_time_env_mutation.py` | **new** — AST guard + masker unit tests |
| `docs/memory/learnings.md` | +1 entry |

**No product code. Zero files under `src/`.**

**Fail-vs-skip policy.** If the harness *declared* a target (caller-supplied `REDIS_URL`, or `TRINITY_REDIS_REQUIRED=1`), unreachable ⇒ **FAIL**. If we derived it ourselves, unreachable ⇒ **SKIP**. A bare developer run with no Redis must still skip — an earlier design keyed on `"REDIS_URL" in os.environ`, which M2 makes unconditionally true, and would have turned every laptop run red.

**Static guard.** AST-based and **module-scope only**. A regex "no occurrence anywhere" guard trips on the legitimate *fixture-time* `os.environ.setdefault` in `test_postgres_backend.py:78`, which runs at test time and is not the bug. Deliberately no `sys.modules` assertion — `tests/lint_sys_modules.py` (#762) already bans that repo-wide with a baseline.

**Layer D** (separate commit, so it reverts independently — it restructures module loading in the file this fix most depends on). `test_monitoring_service.py` installed `sys.modules["services.agent_client"]` at module scope, loading a *second generation* of the module during collection; which generation a module got depended purely on collection order. The existing linter misses it because the file declares the sanctioned `_STUBBED_MODULE_NAMES` + `_restore_sys_modules` escape hatch — but that fixture is **function-scoped**, so its first snapshot is taken *after* collection already installed the copy, preserving the pollution instead of undoing it. Fixed with a class-scoped autouse `monkeypatch.setitem`. Deliberately **not** moved into `tests/integration/conftest.py`: that is linter-legal but *broadens* the blast radius, handing the copy to the five breaker modules that currently get the real module.

## Test plan

Baseline was taken **before any edit** (`.plan/`, worktree-local) so a post-fix red is attributable. It also verified premise P4 — that the 57 skipped tests actually pass once they run — rather than assuming it.

> **Reviewer note on coverage.** CI's pytest job runs `pytest unit/` only — it does **not** run `tests/integration/`. So a green CI does not exercise the resolver, the fail-vs-skip policy, or the five migrated modules; the only thing this PR adds that CI runs is the AST guard in `tests/unit/`. That is why the local proofs below and the `/verify-local` run (which boots a stack and runs integration) are the real evidence here, not the check marks on this PR.

- [x] Sentinel proof A/B/C — survives full-directory `--collect-only`; `config.REDIS_URL` re-pointed; `agent_client` real in all three
- [x] Baseline, reachable target: **72 passed / 13 skipped**, zero Redis-caused skips, identical across all three harness shapes
- [x] FAIL branch — caller-supplied `REDIS_URL` at a dead port ⇒ **exit 1**, masked
- [x] SKIP branch — no `REDIS_URL`, no Redis ⇒ **exit 0**, `SKIPPED`, masked (the bare-laptop regression case)
- [x] No-leak — the real `.env` password appears **0 times** in either run's captured output
- [x] `tests/unit/test_1775_no_import_time_env_mutation.py` — 19 passed; guard fires on a planted import-time write, stays green on `test_postgres_backend.py`
- [x] `tests/lint_sys_modules.py` — 203 violations in 60 files, **no new violations** (baseline allows 240)
- [x] `ruff check` clean on all changed files
- [x] `/verify-local` (`--skip-agent`, no `docker/base-image` change) — **RESULT: PASS**, whole run green: unit **4923 passed / 17 skipped / 0 failed**; integration stage against a freshly built + booted stack: **70 passed, 13 skipped, 0 failed** (2 deselected = the pre-existing `test_webhook_rate_limit` registry entries). That reconciles exactly with the pre-fix baseline's reachable-target run of 72 passed / 13 skipped, and the 13 remaining skips are the legitimate non-Redis gates (AC8) — **zero Redis-caused skips**, versus 57 before.
- [x] CI — 18 checks pass, 0 failures (incl. `gitleaks` and the `sys.modules` pollution lint)

The 7 pre-existing `test_monitoring_service` failures recorded in the baseline (`assert 0 == 2`) are M2's config-snapshot bug and are **fixed** by this PR's re-point, not caused by it.

**Rebased onto `dev` tip (`e9eb40da`).** An earlier `/verify-local` run on this branch went red on `tests/unit/test_1474_read_boundary_z.py::test_schedules_summary_last_run_at_normalized`, which this PR does not touch. Root cause was **not** this branch: that test pinned a literal timestamp (`2026-07-06`) while `get_agent_schedules_summary` reads a 168-hour window, so the literal aged out of the window and the row's `last_run_at` came back `None` — a time bomb, fixed upstream in #1783/#1784 by deriving the fixture timestamp from `now − 1h`. This branch simply predated that commit by one. Rebasing onto the current `dev` tip picks up the fix and the test is green again. Recorded here because the pre-edit baseline had filed this failure as "pre-existing on dev", which was the right conclusion for the wrong reason.

## Follow-ups — flagged deliberately, not filed

- The two `tests/integration/test_webhook_rate_limit.py` entries in `known-false-fails.txt` are **the same bug class as this issue** — a hardcoded endpoint versus the environment-supplied one. Out of scope here, but they are a real fix rather than a permanent registry entry, and could come **off** the registry.
- `_has_stubbed_module_names_helper` in `tests/lint_sys_modules.py` only checks that a file *declares* the `_STUBBED_MODULE_NAMES` + `_restore_sys_modules` pair — it never verifies the restore covers the keys written, or that it runs at a scope able to undo them. That file-level laundering is what hid Layer D. Tightening it is a repo-wide gate change across 60 baselined files.
- #1775 carries no `complexity-*` label and is not linked as a sub-issue of **#1259** (*Execution correctness & test health*), whose outcome — "green CI + no silent execution/telemetry loss" — is verbatim this bug.

## Docs

Bug-fix doc tier (Rule #4): `docs/memory/learnings.md` **+1 entry**. No `requirements.md`, `architecture.md`, or feature-flow change — this is test infrastructure and no section governs `tests/` layout.

Fixes #1775


vybe (approved):
Validated via /validate-pr: test-harness-only fix for the 6-of-83 silent-skip collection bug (#1775). Both import-time mutations addressed (M1 module-scope REDIS_URL clobber deleted, M2 root-conftest sentinel handling made distinguishable), skips converted to failures so the exit code is honest, and the new tests/unit/test_1775_no_import_time_env_mutation.py guard statically prevents recurrence. mask_redis_url never throws on the input it exists to redact — right hardening for a public repo. Security scan clean.

PR #1781, opened 2026-07-26
Release: v0.8.5

vybe:
Release v0.8.5 — cumulative changes since v0.8.0.

**Payload**: 773 commits · 619 files (+97.7k/−14k) · 78 public + ~17 private-tracker issues.
**Full curated notes**: `docs/releases/0.8.5.md` (in this PR — lands on main with the tag).

## Highlights

- **Skills as a first-class platform capability**: full-directory skill packages + frontmatter contract (ent#183/#182), skill-runner agent (ent#139), exposable A2A skills (ent#180)
- **Shared sessions (rooms)** with mention-wake turn engine + Sessions view (ent#169/#170)
- **Agent display names** across the platform (#1639–#1643, ent#181)
- **Two-tier telemetry**: Tier-1 local default-on (ent#184), Tier-2 opt-in sharing (#1723, ent#12)
- **Ephemeral ghost agents** (ent#69) · **Voice replies v2** (ent#117)
- **First-run**: default system seed (ent#124), tokenless public templates (ent#123), per-user GitHub PATs (ent#162, #1574)
- **Async orchestration**: task-completion events (#1578), self-reminders (#1296), re-delivery governor lever (#1402)
- **Fleet hygiene**: git bloat bounded (#1595/#1596), volume lifecycle (#1581, #1664–#1671), retention blast-radius guard (#1638/#1644/#1709)
- **Trinity Helper MCP server** (#1459)

⚠️ **Upgrade notes** (full detail in `docs/releases/0.8.5.md`): unflagged skill-package format change (ent#187 decision — additive, backward-compatible); **base-image rebuild + agent recreate required** for ent#123/#1595/#1596/#1574; `deploy_system` now defaults to continue-on-error (ent#125 — pass `strict: true` for the old behavior).

---

Closes #1077
Closes #1142
Closes #1164
Closes #1167
Closes #1296
Closes #1310
Closes #1402
Closes #1449
Closes #1459
Closes #1474
Closes #1481
Closes #1483
Closes #1484
Closes #1485
Closes #1486
Closes #1487
Closes #1488
Closes #1525
Closes #1526
Closes #1533
Closes #1540
Closes #1557
Closes #1558
Closes #1559
Closes #1560
Closes #1561
Closes #1562
Closes #1574
Closes #1575
Closes #1577
Closes #1578
Closes #1581
Closes #1582
Closes #1593
Closes #1595
Closes #1596
Closes #1598
Closes #1600
Closes #1601
Closes #1607
Closes #1609
[...]

PR #1784, opened 2026-07-26
fix(tests): test_1474 time-bomb fixture — seed relative to now (#1783)

### LINKED ISSUE #1783: bug(tests): test_1474_read_boundary_z schedules-summary test is a time bomb — fixed 2026-07-06 fixture rotted out of the 168h window on 2026-07-13
`test_1474_read_boundary_z.py` seeds `schedule_executions` with the fixed literal `NAIVE = "2026-07-06T11:00:00.207634"`. `get_agent_schedules_summary(agent, 168)` computes `last_run_at` **in-window** (the last-run subquery shares `base_where` with the 168h cutoff — deliberate #1115 semantics), so from 2026-07-13 the seeded row falls outside the window, `last_run_at` comes back `None`, and `test_schedules_summary_last_run_at_normalized` fails.

Invisible until the v0.8.5 release PR (#1781) because:
- dev pushes don't run the pytest matrix (lint only),
- intra-dev PRs fail the test on **both** base and head → the regression diff ignores it as pre-existing,
- the release PR is the first **base=main** comparison, where main lacks the test entirely → reported as "new failure introduced by HEAD" in all 3 seeds.

Fix: derive `NAIVE`/`INSTANT` relative to `now (UTC) − 1h` so the row stays in-window forever; the naive-string shape (no `Z`) that the test exists to pin is unchanged.

Found cutting v0.8.5 (skill step 7: the release PR is the first full unit-suite run over the payload).

vybe:
Fixes #1783

Test-only, one file. The fixed `2026-07-06` fixture literal rotted out of the 168h schedules-summary window on 2026-07-13; the failure was invisible on dev (matrix runs PR-only; intra-dev PRs fail it on both diff sides) and first fired on the v0.8.5 release PR (#1781) as a head-only failure in all 3 seeds. Seeding `now − 1h` keeps the row in-window forever while preserving the naive no-`Z` string shape the test exists to pin.

Unblocks the v0.8.5 cut — once this lands, #1781 re-runs clean.

### DIFF
--- a/tests/unit/test_1474_read_boundary_z.py
+++ b/tests/unit/test_1474_read_boundary_z.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 
 import sqlite3
 import sys
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 
 import pytest
@@ -48,8 +48,11 @@ def _restore_sys_modules():
 
 
 # A naive scheduler-written timestamp (no 'Z') and its aware equivalent.
-NAIVE = "2026-07-06T11:00:00.207634"
-INSTANT = datetime(2026, 7, 6, 11, 0, 0, 207634, tzinfo=timezone.utc)
+# Derived from now − 1h, NOT a fixed literal: the schedules-summary read is
+# 168h-windowed, so a literal date rots out of the window and the test becomes
+# a time bomb (#1783 — fired first on the v0.8.5 release PR).
+INSTANT = datetime.now(timezone.utc).replace(microsecond=207634) - timedelta(hours=1)
+NAIVE = INSTANT.replace(tzinfo=None).isoformat()
 
 
 def _assert_utc_normalized(value: str):


AndriiPasternak31 (approved):
Approved — validated by execution on both sides, not by reading.

**Root cause confirmed in code.** `get_agent_schedules_summary` builds `base_where = and_(se.agent_name == agent_name, se.started_at > cutoff)` and the last-run subquery `_last_subq` reuses that same predicate, so `last_run_at` is in-window by design (#1115). A fixed `2026-07-06` fixture against a 168h window therefore returns `None` from 2026-07-13 on, and `_assert_utc_normalized` fails at `assert value is not None`. The defect is the fixture, not the read path — correct call to fix it here.

**Reproduced locally (py3.12, `tests/requirements-test.txt`):**
- `dev` tip: `1 failed, 5 passed` — `test_schedules_summary_last_run_at_normalized` FAILED (bomb is live today)
- this branch: `6 passed`

**The fix holds at the boundary.** `NAIVE` keeps the naive no-`Z`/no-offset shape the file exists to pin; the pinned `microsecond=207634` is load-bearing — a `microsecond=0` would make `isoformat()` drop the fractional part and silently change the seeded string shape. Round-trip `fromisoformat(normalized) == INSTANT` is exact. And since cutoff and row stay ~167h apart, their date components can never collide, so the naive-vs-`Z` lexicographic comparison (the Invariant #16 trap this file guards) stays correct. It cannot rot again.

**Scope/isolation:** one file, +6/−3, tests only. `NAIVE`/`INSTANT` are module-local (no cross-file refs), and there's no freezegun/time-freezing in `conftest.py` for the module-level `datetime.now()` to fight with. Tiered-docs rule: bug fix → commit message only, no architecture/requirements/feature-flow debt.

CI: all 16 real checks green; `regression diff` (the actual gate) passes across all 3 seeds.

**Follow-up, deliberately not in this PR:** `tests/unit/test_1296_reminders.py` uses the fixed *future* literal `2026-08-01T10:00:00Z` as a valid `fire_at` in ~8 places — same bug class pointed the other way, stale in ~6 days if any of those paths validate the instant against `REMINDER_MIN_DELAY_SECONDS`/the max window rather than just parsing ISO. Worth its own issue; shouldn't hold up the v0.8.5 unblock.

PR #1786, opened 2026-07-26
chore: sync dev with release v0.8.5

vybe:
Post-release resync: records the v0.8.5 squash (`b8cf94ca`) as a parent of dev so the branches share history again. Tree is byte-identical to dev — no content changes.

**Merge with a merge commit (NOT squash)** — squashing would drop the recorded parent and main would count as diverged again at the next cut (the pre-flight Step 1 check).


AndriiPasternak31 (approved):
Validated via `/validate-pr`. Approving — with one merge-mechanics condition that has to hold or the PR is self-defeating.

**Verified the claim.** Three-dot diff against current `dev` is empty — this genuinely adds no content, it only records `b8cf94ca` as a parent. I also confirmed the reconciliation is still needed: `git merge-base --is-ancestor origin/main origin/dev` is still false, so the branches are still diverged. `dev` has moved 13 commits since this was opened, which is harmless for a merge commit (the merge takes both sides).

**⚠️ Must be merged with a merge commit, NOT squash.** This is the whole point of the PR and it runs against the repo default — a squash flattens the history and drops `b8cf94ca` as a parent, so `main` counts as diverged again at the very next release cut and the Step 1 pre-flight check fires exactly as it did this time. Whoever lands this needs to override the merge method deliberately; if auto-merge is armed on the default strategy it should be disarmed first.

No other findings — nothing to scan for secrets, docs, or tests in a zero-content PR.

PR #1798, opened 2026-07-27
fix(ui): surface the autonomy gate in the Schedules tab (#1796)

### LINKED ISSUE #1796: Schedules on a non-autonomous agent are silently skipped — UI shows 'enabled' with a live next-run countdown and writes no execution row
## Summary

`autonomy_enabled` defaults to `0` for every newly created agent, and the scheduler silently skips schedules for non-autonomous agents. The behavior is intentional and documented — but **nothing observable tells the user it is happening**, and the UI actively suggests the opposite: the schedule shows `enabled: true` with a live "Next: in N minutes" countdown that keeps advancing, and **no execution row is ever written**.

The result is the worst kind of first-run experience: a user creates their first schedule, watches the countdown reach zero, and nothing happens — with no error, no failed execution, and no hint anywhere in the product.

Verified on v0.8.5 (`b8cf94ca`), fresh install.

## Repro

1. Create an agent (any template) — `autonomy_enabled` is `0` by default (`db/schema.py:77`).
2. Create a schedule on it: `POST /api/agents/{name}/schedules` with `*/10 * * * *`. It returns `enabled: true` and a correct `next_run_at`.
3. Wait for the fire time.

**Observed** — the job runs and is discarded, in the scheduler container log only:

```
08:20:00 apscheduler.executors.default: Running job "p1-scout:p3-hourly (trigger: cron[…minute='*/10'], next run at: 2026-07-27 08:30:00 UTC)"
08:20:00 scheduler.service: Schedule pc-Zd66mTa_rTmZ3qfQ-Vg skipped: agent p1-scout autonomy is disabled
08:20:00 apscheduler.executors.default: Job "…" executed successfully
```

Meanwhile the API keeps reporting a healthy, enabled, upcoming schedule:

```
GET /api/agents/p1-scout/schedules/{id}
  enabled: True | last_run_at: 2026-07-27T08:14:54Z | next_run_at: 2026-07-27T08:30:00Z

GET /api/executions?agent=p1-scout&triggered_by=schedule
  schedule-triggered rows: 0
```

`last_run_at` is stale (it reflects an earlier *manual* trigger, not the skipped cron), `next_run_at` advances normally, and the executions list stays empty forever.

## Why the user cannot figure this out

- `components/SchedulesPanel.vue` contains **zero** references to autonomy. It renders `Next: {{ formatRelativeTime(schedule.next_run_at) }}` and an "Overdue" badge, so the schedule looks live and healthy.
- Because `next_run_at` advances each cycle, even the "Overdue" badge never appears.
- No execution row is created, so Executions / Overview / analytics show nothing at all — not a failure, not a skip.
- The only mention of autonomy in the whole schedules UI path is a toast *after* you toggle it (`views/AgentDetail.vue:552`).

The dependency **is** documented (`docs/user-docs/automation/scheduling.md:13` — "Schedules will not fire if autonomy is off"), so this is not a correctness bug. It is an observability gap: the product state contradicts the documented rule, and a user who hasn't read that line has no path to discovering it.

## Suggested direction

Any one of these would close the gap:

- Show a warning in the Schedules tab (and in the create-schedule flow) when `autonomy_enabled` is false — e.g. "Autonomy is off — these schedules will not fire", with the toggle inline.
- Record a `skipped` execution row when a fire is discarded for autonomy, so Executions shows *something*. There is precedent: the template pre-check path (SCHED-COND-001) already records a skipped execution for an empty pre-check result.
- Suppress or annotate `next_run_at` in the API response when the agent is non-autonomous, so clients cannot render a countdown for a job that will not run.

The first is the smallest and most direct fix for the newsletter-launch window.

dolho:
## What

Makes the autonomy gate visible in the Schedules tab. **No scheduler or backend change** — the skip behavior is documented and intentional (`docs/user-docs/automation/scheduling.md:13`); the problem is that nothing observable said it was happening.

## Why

`autonomy_enabled` defaults to `0` on every new agent. With it off, the scheduler fires each job on schedule and immediately discards it:

```
08:20:00 apscheduler.executors.default: Running job "p1-scout:p3-hourly …"
08:20:00 scheduler.service: Schedule pc-Zd66… skipped: agent p1-scout autonomy is disabled
08:20:00 apscheduler.executors.default: Job "…" executed successfully
```

Meanwhile the UI showed a healthy, enabled schedule with a live `Next: in N minutes` countdown, `next_run_at` advancing every cycle, and **no execution row anywhere** — not a failure, not a skip. `SchedulesPanel.vue` had zero references to autonomy, so there was no path from the symptom ("my schedule never ran") to the cause.

A/B on a fresh 0.8.5 instance, same schedule and agent, nothing else changed:

| autonomy | result |
|---|---|
| off (08:20 fire) | `skipped: … autonomy is disabled`, 0 execution rows |
| on (08:30 fire) | `Schedule p3-hourly dispatched … execution_id=JNvPBhSlWeqWZv8OPROBxQ`, 1 execution row |

## Changes

- **`SchedulesPanel.vue`** — warning banner when autonomy is off *and* at least one schedule is enabled, explaining that runs are skipped silently and no execution is recorded, with an inline **Enable autonomy** button that emits to the parent's existing toggle.
- **`SchedulesPanel.vue`** — in that state the per-schedule `Next: …` countdown is replaced with `Will not fire — autonomy off`, so no row promises a run that cannot happen. The Overdue branch is unchanged for the normal case.
- **`AgentDetail.vue`** — passes `:autonomy-enabled="!!agent.autonomy_enabled"` and wires `@enable-autonomy` to the existing `toggleAutonomy` (which already updates `agent.autonomy_enabled` locally, so the banner clears without a refetch).

Design notes:
- Counts **enabled** schedules only — a disabled schedule wouldn't fire either way, so warning about it is noise.
- `autonomyEnabled` prop defaults to `true`, so any parent that doesn't pass it can never render a false "will not fire" warning.

## Verification

- `agent.autonomy_enabled` confirmed present in the live `GET /api/agents/{name}` payload, so the prop is never silently `undefined` (which would have inverted the banner).
- Production frontend image builds clean from this branch: `✓ 2713 modules transformed`, exit 0.
- Both new strings confirmed present in the emitted bundle (`assets/AgentDetail-*.js`).
- **Not visually verified** — I had no browser tooling in this session, so the banner's rendering and dark-mode appearance have not been eyeballed. Logic, compilation and bundle inclusion are verified; a quick look at the tab with autonomy off would be worth it before merge.

## Alternatives considered

The issue also proposed recording a `skipped` execution row (precedent: the SCHED-COND-001 pre-check path) or suppressing `next_run_at` in the API. Both are larger and touch the scheduler/API contract; this UI-only change closes the discoverability gap with the least risk for the release window. Either backend option remains open as a follow-up.

Related to #1796

Fixes #1796

### DIFF
--- a/src/frontend/src/components/SchedulesPanel.vue
+++ b/src/frontend/src/components/SchedulesPanel.vue
@@ -17,6 +17,35 @@
       </button>
     </div>
 
+    <!-- #1796: autonomy is the master gate for every schedule on this agent.
+         With it off the scheduler fires each job on time and discards it
+         without writing an execution row, so the tab would otherwise look
+         completely healthy while nothing ever runs. -->
+    <div
+      v-if="!autonomyEnabled && enabledScheduleCount > 0"
+      class="rounded-md border border-amber-300 dark:border-amber-700 bg-amber-50 dark:bg-amber-900/20 p-4"
+    >
+      <div class="flex items-start">
+        <svg class="w-5 h-5 mr-3 mt-0.5 flex-shrink-0 text-amber-600 dark:text-amber-400" fill="none" stroke="currentColor" viewBox="0 0 24 24">
+          <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M12 9v2m0 4h.01M5.07 19H19a2 2 0 001.75-2.96l-6.93-12a2 2 0 00-3.5 0l-6.93 12A2 2 0 005.07 19z" />
+        </svg>
+        <div class="flex-1">
+          <p class="text-sm font-medium text-amber-800 dark:text-amber-200">
+            Autonomy is off — {{ enabledScheduleCount === 1 ? 'this schedule will not fire' : `these ${enabledScheduleCount} schedules will not fire` }}
+          </p>
+          <p class="mt-1 text-sm text-amber-700 dark:text-amber-300">
+            Autonomy mode is the master switch for scheduled work on this agent. While it is off, each run is skipped silently and no execution is recorded.
+          </p>
+          <button
+            @click="$emit('enable-autonomy')"
+            class="mt-3 inline-flex items-center px-3 py-1.5 border border-transparent text-sm font-medium rounded-md text-white bg-amber-600 hover:bg-amber-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-amber-500"
+          >
+            Enable autonomy
+          </button>
+        </div>
+      </div>
+    </div>
+
     <!-- Create/Edit Form Modal -->
     <div v-if="showCreateForm || editingSchedule" class="fixed inset-0 z-50 overflow-y-auto">
       <div class="flex items-center justify-center min-h-screen px-4">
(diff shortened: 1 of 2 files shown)


vybe (approved):
Validated via /validate-pr: frontend-only surfacing of the documented autonomy gate (#1796) — banner + per-schedule chip replace the false 'Next: in N min' promise; enable-autonomy emit wired to the existing AgentDetail.toggleAutonomy (verified at AgentDetail.vue:523). Prop defaults to true so parents not passing it never show a false warning. A/B evidence in the PR body. Security scan clean. Note: no automated UI test in the diff — acceptable for a presentational fix; covered manually per the A/B log. Added 'Fixes #1796' for status auto-promotion.

PR #1803, opened 2026-07-27
fix(create): reject an unresolvable local: template instead of creating an empty agent (#1793)

### LINKED ISSUE #1793: Nonexistent local: template silently creates an empty agent (200) instead of failing — github: path correctly 400s
## Summary

`POST /api/agents` with a `local:` template that does not exist returns **200 Created** and provisions a running container with **no template content at all** — no `CLAUDE.md`, no `template.yaml`, no skills. The caller gets a normal-looking agent object and no indication that the template they asked for was never found.

A nonexistent `github:` template is rejected correctly, so the two source types behave inconsistently.

## Repro

```bash
curl -X POST -H "Authorization: Bearer $T" -H 'Content-Type: application/json' \
     -d '{"name":"p1-nope","template":"local:does-not-exist"}' \
     http://localhost:8000/api/agents
```

**Actual — 200:**

```json
{"name":"p1-nope","type":"business-assistant","status":"stopped","port":2228,
 "container_id":"33995259fd81…","template":"local:does-not-exist",
 "runtime":"claude-code","base_image_version":"0.8.5"}
```

The `type` field is a fabricated default; nothing in the response says the template was missing.

**Resulting agent — an empty shell that reports itself running:**

```
$ curl .../api/agents/p1-nope/info
{"has_template":false,"agent_name":"p1-nope","template_name":"local:does-not-exist",
 "message":"No template.yaml found - this agent was created without a template","status":"running"}

$ docker exec agent-p1-nope ls /home/developer
.bashrc  .cache  .claude  .claude-agent  .config  .local  .mcp.json  .profile
.ssh  .tmp  .trinity  content  mcp-servers          # no CLAUDE.md, no template.yaml
```

## Contrast — the `github:` path gets it right

```bash
-d '{"name":"p1-badrepo","template":"github:abilityai/definitely-not-a-real-repo"}'
→ 400 {"detail":"Repository 'abilityai/definitely-not-a-real-repo' was not found or is private.
        If it is private, add your GitHub token in Settings or ask an admin to configure the platform token."}
```

That is the correct shape: fail fast, explain, create nothing.

## Impact

The agent starts, shows `running` in the fleet list, and answers chat — as a generic assistant with zero instructions, skills, or MCP config. The user discovers this only by talking to it and noticing it has no idea what it is supposed to do, having already spent a container and a data volume.

The most likely way to hit this is **not** the UI dropdown but the MCP `create_agent` tool, which takes a free-text template id — exactly the surface used by the "Claude Code writes the agent, Trinity runs it" workflow, and exactly where a stale or hallucinated template id shows up. A renamed or newly-`hidden` template produces the same silent downgrade for anyone with a saved script.

## Suggested direction

Resolve `local:` templates before provisioning and 400/404 on a miss, mirroring the `github:` path. If templateless creation is a deliberate capability, it should require an explicit opt-in (e.g. an empty/omitted `template`) rather than being the silent fallback for an unresolvable one.

dolho:
## What

`POST /api/agents` with a `local:` template that doesn't resolve now returns **404 `UNKNOWN_LOCAL_TEMPLATE`** instead of creating a templateless agent and reporting 200.

## Why

`_resolve_local_template` looked for `template.yaml` under the curated root then the deploy-local root, and when neither had it, simply returned an empty `template_data`. Creation continued and provisioned a **running container with no CLAUDE.md, no template.yaml and no skills**:

```
$ curl -X POST … -d '{"name":"p1-nope","template":"local:does-not-exist"}'
200 {"name":"p1-nope","type":"business-assistant","status":"stopped", …}

$ curl …/api/agents/p1-nope/info
{"has_template":false,"message":"No template.yaml found - this agent was created without a template","status":"running"}

$ docker exec agent-p1-nope ls /home/developer
.bashrc  .cache  .claude  .config  .mcp.json  .ssh  .trinity  content  mcp-servers    # no CLAUDE.md
```

`type: business-assistant` there is a fabricated default — nothing in the response indicated the requested template was never found. The agent shows `running` in the fleet list and answers chat as a generic assistant with no idea what it is for, having already consumed a container and a data volume.

The `github:` path already handled this correctly (`400`, "Repository … was not found or is private"), so the two source types disagreed.

The most exposed caller is the MCP `create_agent` tool, which takes a free-text template id — precisely the surface used by the "Claude Code writes the agent, Trinity runs it" workflow, and where a stale or hallucinated id shows up. A renamed or newly-`hidden` template silently downgraded any saved script the same way.

## Change

One guard clause in `_resolve_local_template`, raising before any side effect. It sits in the template-resolution phase, so a rejected create leaves **no container, volume, MCP key or ownership row**.

The diff looks larger than it is: the load block is dedented one level now that the existence check is a guard clause. No logic inside it changed.

## Behavior preserved

- **Templateless creation still works** by omitting `template` — the fallback is now opt-in rather than the silent result of an unresolvable name.
- **Invalid/traversal names keep their existing 400** `INVALID_LOCAL_TEMPLATE_NAME` (covered by a test, so #1793 can't mask it).
- **deploy-local (#950) unaffected** — `deploy_local_agent_logic` already validates the uploaded archive contains a `template.yaml` before writing it to the deploy-local root, so a deployed template always resolves.

## Verification

`tests/unit/test_1793_unknown_local_template.py` — 5 cases, all passing:

```
5 passed, 15 warnings in 0.18s
```

- unknown name → 404 `UNKNOWN_LOCAL_TEMPLATE`, message names the template
- directory exists but has no `template.yaml` → 404 (this is the shape that produced the empty agent)
- resolvable template still loads and still projects `type`/`runtime` onto the config — parametrized over **both** roots
- traversal name still 400, not 404

Live check against a running 0.8.5 instance (patch applied to the running backend, which bind-mounts its source):

```
POST {"name":"p1793-bogus","template":"local:does-not-exist"}  → HTTP 404
  {"detail":{"error":"Local template 'does-not-exist' was not found.","code":"UNKNOWN_LOCAL_TEMPLATE"}}
containers matching agent-p1793-bogus: 0

POST {"name":"p1793-ok","template":"local:scribe"}             → HTTP 200
```

The instance was restored to stock afterwards.

Related to #1793

Fixes #1793

### DIFF
--- a/src/backend/services/agent_service/crud.py
+++ b/src/backend/services/agent_service/crud.py
@@ -672,7 +672,11 @@ async def _reserve_git_instance(
 def _resolve_local_template(config: AgentConfig) -> tuple[dict, Optional[dict]]:
     """Load a `local:`-prefixed template's `template.yaml` (curated catalog then
     deploy-local store, #950). Mutates `config` runtime/type/resources/tools/
-    mcp_servers fields. Returns `(template_data, template_shared_folders)`."""
+    mcp_servers fields. Returns `(template_data, template_shared_folders)`.
+
+    Raises `HTTPException(404, UNKNOWN_LOCAL_TEMPLATE)` when the name resolves
+    to no `template.yaml` under either root (#1793) — this previously returned
+    an empty dict and the caller provisioned a templateless container."""
     template_data: dict = {}
     template_shared_folders = None
     # Local template - strip "local:" prefix. Look in curated catalog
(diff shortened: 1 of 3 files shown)


dolho:
### CodeQL fix (edd854fb)

The first push failed CodeQL with **2 new high-severity `py/path-injection` alerts**:

```
src/backend/services/agent_service/crud.py:703  Uncontrolled data used in path expression
src/backend/services/agent_service/crud.py:718  Uncontrolled data used in path expression
```

Those are the `template_yaml.exists()` and `open(template_yaml)` calls — **not** code this PR introduced. They are false positives: `raw_name` passes a regex allowlist (`[a-zA-Z0-9][a-zA-Z0-9_.-]*`, no `..`, no separators) and the joined path is `resolve()`d and checked with `is_relative_to(root)` before any filesystem access. `_safe_local_template_path` documents exactly this as "the pattern CodeQL recognises as a `py/path-injection` barrier".

They surfaced because the original guard-clause version dedented the entire load block, which moved both expressions onto new lines and re-fingerprinted alerts already dismissed on `dev`.

Fix: same behavior, expressed as `if/else` rather than a guard clause. The load block is now byte-identical to `dev` and the two flagged expressions keep their original position, so no new alert fingerprints are produced. The diff is purely additive (23 insertions, 1 deletion — the docstring line).

Trade-off worth naming: a guard clause reads better than `if/else` with the raise at the bottom. I chose the shape that doesn't re-trigger dismissed FPs; if you'd rather have the guard clause, it needs those two alerts dismissed once and this can be flipped back.

Re-verified after the restructure: the 5 `test_1793_unknown_local_template.py` cases pass, 35 neighbouring template/deploy tests pass, and `tests/lint_sys_modules.py` reports no new violations.

vybe (approved):
Validated via /validate-pr: 404 UNKNOWN_LOCAL_TEMPLATE raised from _resolve_local_template's else branch; verified the call site (crud.py:1810 via _resolve_template) sits OUTSIDE the docker try-block, so the 4xx propagates cleanly with zero side effects — matches the github: path contract. Named regression tests (test_1793_*, 4 cases incl. hollow-dir shape and traversal-400 preservation) + characterization fixture fix. Security scan clean. Added 'Fixes #1793' to the body for status auto-promotion.

PR #1807, opened 2026-07-27
fix(reminders): surface the autonomy hold instead of failing silently (#1806)

### LINKED ISSUE #1806: Self-reminders on a non-autonomous agent are silently held — never armed, no log, no doc, stuck at 'pending' past fire_at
## Summary

A self-reminder set on an agent whose `autonomy_enabled` is `0` is **never armed by the scheduler**. It sits at `status: pending` with a `fire_at` in the past, `fire_attempts: null` and `error: null`, and there is no log line anywhere saying why. Nothing in the product or the docs connects reminders to autonomy.

Same family as #1796 (schedules), but with **less** observability: a skipped schedule at least logs `skipped: agent <name> autonomy is disabled`. A held reminder produces no scheduler output at all, because it never enters the job store.

Mitigating: the reminder is **held, not lost** — enabling autonomy makes it fire past-due within one reconcile cycle.

Verified on v0.8.5 (`b8cf94ca`), fresh install. This is one of the headline v0.8.5 features ("Agents can remind themselves").

## Repro — clean A/B

Two agents, identical 60s reminder, set in the same second. Only `autonomy_enabled` differs.

```
POST /api/agents/p1-scout/reminders   {"message":"…","delay_seconds":60}   # autonomy ON
POST /api/agents/p1-corbin/reminders  {"message":"…","delay_seconds":60}   # autonomy OFF
```

Both accepted as `pending`, both `fire_at = 10:43:51Z`. Neither response hints that one of them cannot fire.

Result:

```
10:43:53Z  scout(autonomy ON)=fired    corbin(autonomy OFF)=pending
10:44:08Z  scout(autonomy ON)=fired    corbin(autonomy OFF)=pending
```

The corbin row, well past its fire time:

```
status: pending | fire_at: 2026-07-27T10:43:51.215559Z | attempts: None | error: None
```

## The scheduler never sees it

Only one reminder job was ever armed — the whole scheduler log for that window:

```
10:43:39 apscheduler.scheduler: Added job "SchedulerService._execute_reminder" to job store "default"
10:43:51 apscheduler.executors.default: Running job "SchedulerService._execute_reminder …"
10:43:51 scheduler.service: Reminder rem_9cb920…e4d5 fired for agent p1-scout (execution_id=pdjDMT_4SGuQsuYmdIvJIQ)
```

`get_active_reminders` filters `autonomy_enabled=1`, so corbin's reminder is invisible to `_reconcile_reminders` and no `DateTrigger` is ever registered. There is no skip to log.

## Recovery works

Enabling autonomy on the same agent, with no other change:

```
10:45:02Z corbin reminder = pending
10:45:47Z corbin reminder = fired        # past-due fire, ~1 reconcile cycle
```

So the data is intact and the design is defensible — an operator pausing an agent should not have it wake itself up. The defect is purely that nothing says so.

## Why this is worse than #1796

| | schedules (#1796) | reminders (this) |
|---|---|---|
| scheduler log | `skipped: … autonomy is disabled` | nothing — never armed |
| user-facing signal | none | none |
| documented | yes — `docs/user-docs/automation/scheduling.md:13` | **no** — `agent-reminders.md` has zero mentions of autonomy |

An operator debugging a stuck reminder has no log, no error field, no doc, and an API that says `pending`.

## Not affected

Task-completion event wakes (`agent.task.*` → subscriber) **do** land with the subscriber's autonomy off — that path is an HTTP task dispatch, not a scheduler path. Verified: source task completed, subscriber's event-triggered execution count went 1 → 2 with autonomy disabled. So the gap is specific to the two scheduler-driven features.

## Suggested direction

Pick whichever matches the #1796 outcome, so the two behave alike:

- Reject at create time (`409`/`422`) when the target agent is non-autonomous, naming autonomy — the strongest option, since the agent asking for the reminder can then react.
- Or accept but return and surface a `held: true` / `blocked_by: "autonomy"` field on the reminder, and show it in the UI's reminder list.
- At minimum: log the hold on the reconcile pass, and document the dependency in `docs/user-docs/automation/agent-reminders.md` the way `scheduling.md` already does.

dolho:
## What

Makes the autonomy hold on self-reminders visible. **The gate is not changed** — a paused agent still must not wake itself. Only the silence is fixed.

## Why

`get_active_reminders` filters `ao.autonomy_enabled = 1`, so a reminder on a paused agent is never armed. No job is created, so there is no skip to log — and the API reported a healthy-looking row:

```
status: pending | fire_at: 2026-07-27T10:43:51.215559Z | attempts: None | error: None
```

Clean A/B on a fresh 0.8.5 instance, two agents, identical 60s reminder set in the same second, only `autonomy_enabled` differing:

```
10:43:53Z  scout(autonomy ON)=fired    corbin(autonomy OFF)=pending
10:44:08Z  scout(autonomy ON)=fired    corbin(autonomy OFF)=pending
```

Worse than the sibling schedules case (#1796) on every observability axis: schedules at least log `skipped: … autonomy is disabled`, and `scheduling.md` documents the dependency. Reminders had neither.

## Why not reject at create

Considered and rejected. Chat is **not** autonomy-gated, so an agent can be mid-conversation with a user who says "remind me in an hour". A 409 would fail a legitimate request, and queue-until-unpaused is genuinely useful — the reminder does fire past-due once autonomy returns (verified: held reminder fired ~1 reconcile cycle after enabling). The problem was never the holding; it was not saying so.

## Changes

| Layer | Change |
|---|---|
| `models.py` | `autonomy_hold` on `ReminderSummary` (inherited by `Reminder`), **derived at read time** — no column, no dual-track migration. Defaults `False`. |
| `routers/reminders.py` | `_autonomy_hold()` applied to create, list, cancel. Terminal reminders never flagged and skip the lookup. List resolves autonomy **once per page** (all rows share one agent — a per-row resolve would be a pointless N+1). Fail-open on a read error. |
| `src/scheduler` | log a **count** of held reminders per reconcile pass, only when non-zero, best-effort. The only place that can report them, since they never enter the job store. |
| MCP `set_reminder` | explicit `warning` sentence when held — not just the boolean, since the caller is an LLM about to tell a user "done" — plus the dependency in the tool description (Invariant #13). |
| `docs/user-docs/automation/agent-reminders.md` | documents the dependency, mirroring `scheduling.md`. |

## Verification

**Unit** — `tests/unit/test_1806_reminder_autonomy_hold.py`, 11 cases, all passing: held for `pending`/`firing`, not held when autonomy is on, never flagged for `fired`/`cancelled`/`failed` (asserting the lookup is *not even attempted*), fail-open on a read error, precomputed autonomy short-circuits the per-row lookup, the shared response shaper carries the flag, and both models default to `False`.

**Live**, patched into a running 0.8.5 instance (its backend source is bind-mounted):

```
CREATE on paused agent   → status: pending | autonomy_hold: True
CREATE on active agent   → status: pending | autonomy_hold: False
LIST on paused agent     → fired   | autonomy_hold: False      # terminal, correctly not flagged
                           pending | autonomy_hold: True
```

Then toggling autonomy on, without touching the row:

```
pending | autonomy_hold: False   <- flipped, proving it is derived live, not frozen at create
```

The instance was restored to stock afterwards and verified clean.

**MCP server** compiles (`tsc` via the image build, exit 0).

## Not verified

The scheduler log line is unit-and-inspection level only — the scheduler image is not bind-mounted on my test stack, so I did not rebuild it to watch the line appear. The query is the exact complement of `get_active_reminders`' predicate and is wrapped best-effort.

## Follow-up worth its own issue

A reminder held for days fires **immediately** on unpause, and several would fire at once. "Remind me in an hour" arriving three days late may be worse than never arriving. Deliberately out of scope here — flagging it rather than widening this fix before the release.

Related to #1806

Fixes #1806

### DIFF
--- a/docs/user-docs/automation/agent-reminders.md
+++ b/docs/user-docs/automation/agent-reminders.md
@@ -11,6 +11,7 @@ There is no web UI for reminders. They are a backend and MCP primitive that agen
 - **Reminder** — A one-shot, durable, self-scheduled task. An agent sets a reminder on **itself** with a message and a fire time. At that time, Trinity runs a normal execution of the agent with that message.
 - **Self-only** — An agent can set, list, and cancel only *its own* reminders. A reminder against a sibling agent is rejected (`403`).
 - **Durable** — Reminders survive a restart. The fire time is stored, not held in memory, and re-armed on boot. A reminder due during a restart fires when the service comes back rather than being lost.
+- **Autonomy-gated** — Reminders only fire while the agent's **autonomy mode** is on, the same master switch that governs [schedules](scheduling.md). Autonomy is **off by default on a newly created agent**.
 
 ### Reminder vs. Loop vs. Schedule
 
@@ -36,6 +37,19 @@ Optional per-reminder overrides — `model`, `timeout_seconds`, and `allowed_too
 **Listing** returns pending reminders by default (soonest fire first); pass a status filter of `all` to see fired, cancelled, and failed reminders too.
 
 **Cancelling** a still-pending reminder stops it from firing. Cancelling an already-cancelled reminder is a no-op success; a reminder that has already fired, is mid-fire, or has failed cannot be cancelled (`409`). An unknown reminder id returns `404`.
+
+### Autonomy must be on
+
+A reminder set on an agent whose autonomy mode is **off** is accepted and **held**: it stays `pending`, is never armed, and does not fire — even once its fire time passes. Turn autonomy on and it fires past-due within about a minute. Nothing is lost, but nothing happens either.
+
+Because autonomy defaults to off on a new agent, this is the usual reason a first reminder "never fires".
+
+How to tell:
+
+- The create response and `list_reminders` set **`autonomy_hold: true`** on any reminder in this state, and the `set_reminder` tool result adds an explicit `warning`. An agent should relay that to the user rather than reporting the reminder as scheduled.
+- The scheduler logs a count each reconcile pass, e.g. `2 reminder(s) held: their agent's autonomy is disabled.`
+
+Chat is *not* autonomy-gated, so an agent can be mid-conversation, accept "remind me in an hour", and still be unable to fire it. Check autonomy when a user asks for a reminder.
 
 **Where fired reminders appear** — a fired reminder is a normal execution row (with `triggered_by: "reminder"`), so it appears on the Executions page, in per-execution detail, and as its own **Reminders** category in the agent's analytics timeline — never folded into Scheduled.
 
--- a/src/backend/models.py
+++ b/src/backend/models.py
@@ -2165,6 +2165,11 @@ class ReminderSummary(BaseModel):
     created_at: str
     fired_at: Optional[str] = None
     cancelled_at: Optional[str] = None
+    # #1806: derived at read time (NOT a column) — true when this reminder is
+    # still live but its agent has autonomy off, so the scheduler will not arm
+    # it. Without this a held reminder is indistinguishable from a healthy one:
+    # `pending` with a fire_at that quietly slides into the past.
+    autonomy_hold: bool = False
 
     class Config:
         from_attributes = True
--- a/src/backend/routers/reminders.py
+++ b/src/backend/routers/reminders.py
@@ -59,9 +59,36 @@ def _self_gate(current_user: User, name: str) -> None:
         )
 
 
+# #1806: statuses the autonomy gate can still act on. A terminal reminder is
+# never "held" — it already fired, was cancelled, or gave up.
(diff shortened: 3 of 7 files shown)


vybe (approved):
Validated via /validate-pr: derived autonomy_hold flag (read-time, no column — correctly avoids a dual-track migration), all three surfaces in sync (router+models, MCP tool with an explicit LLM-facing HELD warning, scheduler once-per-pass held count). db.get_autonomy_enabled verified present (db/agent_settings/autonomy.py:53). Gate behavior unchanged; fail-open on settings-read failure; terminal rows skip the lookup; list path avoids N+1. 7 named regression tests + user-docs update. Security scan clean. Added 'Fixes #1806' for status auto-promotion.

PR #1812, opened 2026-07-27
fix(scheduler): actually enforce freeze_schedules_if_sync_failing (#1808)

### LINKED ISSUE #1808: freeze_schedules_if_sync_failing is never enforced — schedules keep firing while sync is failing (safety toggle is inert)
## Summary

`freeze_schedules_if_sync_failing` (#389) is a **safety toggle that is never enforced**. It is settable, persists, reads back `true`, and the backend even computes the freeze decision correctly — but **nothing consumes that decision**. Schedules keep firing against an agent whose git sync is broken.

This is the inverse of #1796/#1806: not "silently disabled", but "silently not enforced". Worse in kind, because the operator has explicitly asked for a safety behavior and been told they have it.

## Proof — full chain

**1. The toggle works.**

```
PUT /api/agents/p1-corbin/git/freeze-schedules-if-failing  {"enabled":true}
→ {"agent_name":"p1-corbin","freeze_schedules_if_sync_failing":true}

GET .../git/freeze-schedules-if-failing
→ {"agent_name":"p1-corbin","freeze_schedules_if_sync_failing":true}
```

**2. The backend computes the decision correctly.** With the agent's sync state set to failing (`last_sync_status='failed'`, `consecutive_failures=5`):

```
GET /api/internal/agents/p1-corbin/sync-health-status
→ {"freeze_schedules_if_sync_failing":true,"sync_failing":true,
   "should_freeze":true,"consecutive_failures":5}
```

**3. The schedule fires anyway.** A `* * * * *` schedule on that same agent, with `should_freeze: true` in force:

```
11:13:23Z  schedule created
11:13:38Z  schedule-triggered executions: 0
11:13:53Z  schedule-triggered executions: 0
11:14:08Z  schedule-triggered executions: 1   ← fired
```

## Root cause

`routers/internal.py:262` exposes the decision and documents its consumer:

> `#389: lightweight read used by the dedicated scheduler before dispatching. Returns both the per-agent freeze_schedules_if_sync_failing flag and whether the current sync state would trip it. **The scheduler multiplies the two to decide whether to skip the fire.**`

The scheduler does no such thing. Searching the whole repo:

```
$ grep -rn "sync-health-status" --include=*.py --include=*.ts --include=*.vue --include=*.sh .
src/backend/routers/internal.py:262:@router.get("/agents/{agent_name}/sync-health-status")
```

**The endpoint's only occurrence is its own definition.** No caller anywhere. Independently:

```
$ grep -rniE "sync_state|sync_health|consecutive_failures|sync_failing" src/scheduler/*.py
(no matches)
```

The scheduler never looks at git sync state at all, and `task_execution_service` has no freeze check either. Everything between the column and the fire decision exists — only the call is missing.

## Why it matters

The feature exists so an agent stops doing autonomous work while its repo is broken — sync conflicts, diverged history, a failed push loop. Those are exactly the states where letting a GitHub-native agent keep running scheduled work compounds the damage: it keeps mutating a workspace whose persistence is failing, and #389's own operator alert (`sync_failing` at ≥3 failures) fires while the schedules the operator thought they'd frozen carry on.

An operator who enabled this believes they are protected. Nothing in the API, UI, or logs contradicts them.

## Also worth noting

There is **no UI control** for this toggle — the only mention in the frontend is a comment in `SettingsPanel.vue:14` referencing the endpoint path. So today it is API-only *and* inert.

## Suggested direction

Either wire it up or withdraw it — the current middle state is the harmful one:

- **Wire it**: have the scheduler call `/api/internal/agents/{name}/sync-health-status` before dispatching a cron fire and skip when `should_freeze` (recording a `skipped` execution row, so the skip is visible — please don't repeat the #1796 silent-skip shape).
- **Or withdraw it**: remove the endpoints and column, so nobody can enable a protection that does not exist.

If it gets wired, the skip needs to be observable — a frozen schedule with no execution row and no signal would land straight back in the #1796 bucket.

dolho:
## What

Wires up `freeze_schedules_if_sync_failing` (#389), which until now was stored, reported as enabled, and never enforced.

## Why

The backend already computed the decision:

```
GET /api/internal/agents/p1-corbin/sync-health-status
→ {"freeze_schedules_if_sync_failing":true,"sync_failing":true,
   "should_freeze":true,"consecutive_failures":5}
```

Nothing consumed it. Across the entire repo:

```
$ grep -rn "sync-health-status" --include=*.py --include=*.ts --include=*.vue --include=*.sh .
src/backend/routers/internal.py:262:@router.get("/agents/{agent_name}/sync-health-status")
```

One line — its own definition. The endpoint's docstring names the scheduler as its consumer ("the scheduler multiplies the two to decide whether to skip the fire"); the scheduler never referenced sync state at all.

Meanwhile `docs/user-docs/faq/troubleshooting.md:23` tells users it works: *"If the agent has the freeze schedules if sync failing flag on and its git sync has failed 3+ times in a row, the scheduler skips firing until sync recovers."*

**Before**, live: flag on, sync failing (5 consecutive), `should_freeze: true` → cron fired anyway at 11:14.

**Direct DB read, not the HTTP endpoint.** The issue suggested calling `/sync-health-status`. I read the predicate straight from the database instead: the scheduler already does exactly that for every other gate (autonomy, `enabled`, reminders), and adding an HTTP hop to the fire path buys nothing but a new failure mode. The endpoint is now redundant — removing it is a separate call, so I left it.

**Fail-open, deliberately.** Any error in the check (missing table on an older DB, read failure) fires the schedule. A freeze-on-error would silently halt every schedule in the fleet the moment the query broke — strictly worse than the bug being fixed. There's a test for this specific path.

**The skip records an execution row** — unlike the autonomy branch, which deliberately doesn't. That branch's comment explains why: autonomy-off is a static config gate on a default-OFF fleet, so a row per tick would flood `schedule_executions`. A sync freeze is different: it needs BOTH an opt-in toggle AND 3+ consecutive real failures, so the volume is bounded, and the row is precisely the signal an operator needs. Without it this fix would land straight back in the #1796 silent-skip bucket. It reuses the existing max_instances audit path, whose reason strings are now parameters (defaults unchanged, so that caller is byte-identical in behavior).

**Cron only.** A manual trigger still fires — the operator is explicitly asking. Same carve-out the autonomy gate already makes.

**Threshold duplication.** `SYNC_FAILURE_FREEZE_THRESHOLD = 3` mirrors the inline `>= 3` in `internal.py`. A test asserts the two stay in step rather than letting them drift silently.

## Verification

**Unit** — `tests/unit/test_1808_sync_freeze_enforcement.py`, 9 cases, all passing: freeze at/above threshold, fire below it, fire when not opted in regardless of how broken sync is, fire when sync is healthy, fire with no sync-state row, fire for an unknown agent, **fail-open on a broken query**, and the threshold-parity check.

**Live, end to end** — rebuilt the scheduler image onto a running instance and re-ran the exact scenario that failed before:

```
11:34:00 WARNING scheduler.service: Schedule Pz0Bvzz1… skipped: agent p1-corbin git sync is
                 failing and freeze_schedules_if_sync_failing is enabled
11:34:00 INFO    scheduler.service: Recorded skipped execution QIDcMtN908oYQZvDXUw9Rg
```

and the executions list shows a `skipped` row. The instance was restored to stock afterwards and verified clean.

One wrinkle worth knowing for anyone reproducing: the sync-health poller rewrites `agent_sync_state` every 60s, so an injected failing state gets reset — the condition has to be held across a fire boundary.

## Note on the test's import

`src/scheduler` uses relative imports and its own top-level `database` module, which collides with the backend's on the pytest path. The test imports it through the repo root (`src.scheduler.database`) — no `sys.path` or `sys.modules` mutation, so `tests/lint_sys_modules.py` stays green.

Related to #1808

Fixes #1808


vybe:
Pushed `1ad280d8` during validation: `should_freeze_schedules` read rows positionally (`row[0]`), but the PG path (#300) yields `RealDictCursor` mapping rows with no positional access — the KeyError got swallowed by the fail-open except, leaving the freeze silently inert on every PostgreSQL deploy (the exact inert-toggle class this PR fixes). Switched to named access (the module convention per `get_connection`'s docstring) and added a mapping-only-row regression test that fails against the positional version. Everything else validated clean — approving once CI is green.

vybe (approved):
Validated via /validate-pr: Fixes #1808 closing keyword ✅ (P2/type-bug), targets dev, security greps clean, no schema/config change, fail-open design correct (freeze-on-error would be worse than the bug), gate placement + skipped-row audit trail + next_run advance all consistent with the existing autonomy/max_instances patterns. Full pytest matrix ×3 seeds + regression diff green incl. the PG dict-row regression test pushed during review; test file re-run locally (10 passed, new test verified to fail against the unfixed code).

Issue #1809, opened 2026-07-27
bug: a rebuilt agent base image is never picked up — no image-drift predicate, so the v0.8.5 "stop/start-with-recreate" upgrade step is a no-op


AndriiPasternak31:
## Summary

A rebuilt `trinity-agent-base` image is **never** picked up by an agent restart. `needs_recreation` in `start_agent_internal` compares eight config surfaces — none of them the image — so `POST /api/agents/{name}/start` on a running agent takes the no-op path and the container keeps running the **old** base image indefinitely.

This makes the v0.8.5 upgrade instruction incorrect. `docs/releases/0.8.5.md` tells operators:

> **Operator action**: run `./scripts/deploy/build-base-image.sh`, then recreate agents (a config-drift recreate or **stop/start-with-recreate** picks up the new image).

There is no such thing as "stop/start-with-recreate" — stop+start reuses the existing container. An operator who follows the release notes exactly ends up with a fleet still on the old image while believing the upgrade landed, and the release notes list what that silently costs: no PAT-free public-template clone (trinity-enterprise#123), no git-maintenance ownership (#1595/#1596), no preinstalled `gh` CLI (#1574), no pull-worker scaffolding.

Observed on a live 23-agent instance upgrading v0.8.0 → v0.8.5: after `build-base-image.sh` succeeded, every agent still resolved to the old image id, and a plain `start` changed nothing.

## Root cause

`services/agent_service/lifecycle.py:287-296`:

```python
needs_recreation = (
    not shared_folder_match or
    not check_public_folder_mount_matches(container, agent_name) or
    not check_api_key_env_matches(container, agent_name) or
    not check_github_pat_env_matches(container, agent_name) or
    not check_resource_limits_match(container, agent_name) or
    not check_full_capabilities_match(container, agent_name) or
    not check_guardrails_env_matches(container, agent_name) or
    not check_agent_auth_token_env_matches(container, agent_name)
)
```

All eight predicates live in `agent_service/helpers.py` (plus `file_sharing.py`); none reads `container.attrs["Image"]`. The only `Image` reference in the whole module is the *write* side of the recreate:

```python
# lifecycle.py:413 — reads the TAG to build the replacement, never compares
image = old_config.get("Image", "trinity-agent-base:latest")
```

That line is also why the fix is cheap: `Config.Image` is the **tag** (`trinity-agent-base:latest`), so once a recreate is triggered for *any* reason it already resolves to the freshly built image. The only missing piece is the trigger.

[...]

Issue #1788, opened 2026-07-27
Docs advertise two install paths that fail on a fresh machine (install.sh one-liner, quickstart.sh) — repoint to start.sh


dolho:
## Summary

The one-line installer advertised at the top of the README — `curl -fsSL .../install.sh | bash` — **fails on a fresh machine** and leaves no running Trinity. It dies immediately after the (multi-minute) base-image build with a raw Compose interpolation error and no guidance.

Reproduced on v0.8.5 (`b8cf94ca`), clean host, no prior `.env`.

## Repro

```bash
curl -fsSL https://raw.githubusercontent.com/abilityai/trinity/main/install.sh | bash
```

## Actual

```
[OK] Base image built
[INFO] Starting Trinity services...
error while interpolating services.redis.environment.[]: required variable REDIS_PASSWORD is missing a value: REDIS_PASSWORD must be set in .env (generate openssl rand -hex 24)
```

`install.sh` has `set -e`, so it exits right there — no "installation failed" message, no next step. **Zero containers are created.** The user has just waited out a full base-image build for nothing.

## Root cause

`install.sh` bypasses `scripts/deploy/start.sh` and calls `docker compose up -d` directly (install.sh:226).

`start.sh` is the only thing that generates the required secrets. `.env.example` ships them all blank, and `install.sh` fills only 2 of 7:

| Variable | `.env.example` | filled by `install.sh` | filled by `start.sh` |
|---|---|---|---|
| `SECRET_KEY` | blank | ✅ | ✅ |
| `ADMIN_PASSWORD` | blank | ✅ | ✅ (`--unattended`) |
| `REDIS_PASSWORD` | blank | ❌ | ✅ |
| `REDIS_BACKEND_PASSWORD` | blank | ❌ | ✅ |
| `CREDENTIAL_ENCRYPTION_KEY` | blank | ❌ | ✅ |
| `INTERNAL_API_SECRET` | blank | ❌ | ✅ |
| `AGENT_AUTH_SECRET` | blank | ❌ | ✅ |

`docker-compose.yml` uses `${REDIS_PASSWORD:?…}` / `${REDIS_BACKEND_PASSWORD:?…}` (#589), so the missing Redis passwords are a hard stop. The other three would degrade credential encryption (#1159 agent auth, CRED-002) even if boot were reached.

Verified post-run `.env`:

```
SECRET_KEY=<set>
ADMIN_PASSWORD=<set>
[...]

PR #1815, opened 2026-07-27
fix(version): report the code that is running, not the image it was built in (#1810, #1814)

### LINKED ISSUE #1810: bug: /api/version still reports "unknown" without start.sh — the "unknown" sentinel is truthy, so #993's VERSION-file fallback is unreachable
## Summary

`GET /api/version` returns `"version": "unknown"` (and `components.backend`, `components.agent_server`, `components.base_image` with it) on any image built without the `VERSION` build arg — i.e. any `docker compose -f docker-compose.prod.yml build` that doesn't go through `scripts/deploy/start.sh`.

#993 anticipated exactly this operator and added two independent mitigations. Both are present in the tree, and **neither works**, because the `"unknown"` sentinel is a truthy string that short-circuits the fallback chain before the file is ever consulted.

Observed upgrading a live instance v0.8.0 → v0.8.5:

```json
{ "version": "unknown", "components": { "backend": "unknown",
  "base_image": "trinity-agent-base:unknown" }, "git_commit": "unknown" }
```

The instance could not report which build it was running — during an upgrade, which is precisely when you need it.

## Root cause

The intended 3-tier resolution (`src/backend/main.py:1282-1299`):

```python
# Version resolution order (#993):
#   1. VERSION env var — build-stamped from git (e.g. "0.9.0+g4c640b6e"), …
#   2. VERSION file — curated semver, mounted in dev / copied in image.
#   3. "unknown" — neither present.
version = os.getenv("VERSION") or None
if not version:
    version_paths = [Path("/app/VERSION"), …]
    …
version = version or "unknown"
```

Tier 2 is unreachable. `docker/backend/Dockerfile:19-25`:

```dockerfile
ARG VERSION=unknown
…
ENV VERSION=${VERSION} \
```

`ENV` is unconditional, so the env var is **always set** in the image — to the literal string `"unknown"` when no build arg was passed. `os.getenv("VERSION")` therefore returns `"unknown"`, which is truthy, so `or None` does not fire, `if not version:` is False, and the `/app/VERSION` file is never read.

Both mitigations #993 added are consequently dead code in the only case they were written for:

- `docker/backend/Dockerfile:114` — `COPY ../../VERSION /app/VERSION`, whose own comment reads *"prod compose doesn't [mount it], so without this COPY the version field falls back to `unknown`"*. The COPY lands; the reader never gets there.
- `docker/backend/Dockerfile:17-18` — *"Defaults to `unknown` so local builds and operators who skip start.sh still get a well-typed response."* Well-typed, but the response is `unknown` rather than the curated semver sitting in the image one path away.

This is a sentinel-vs-absence bug: `ARG`+`ENV` cannot express "unset", so the sentinel must be filtered on read.

## Suggested fix

Treat the sentinel as absence at the single read site:

```python
_version_env = os.getenv("VERSION")
version = _version_env if _version_env and _version_env != "unknown" else None
```

Tier 2 then resolves `0.8.5` from the baked file, and an operator who skips `start.sh` gets the curated semver without the git suffix — a correct, honest, partial answer instead of nothing.

Scope note: the same `ARG=unknown → ENV` shape applies to `GIT_COMMIT`, `GIT_BRANCH`, `BUILD_DATE`, `GIT_COMMIT_SUBJECT`, `GIT_COMMIT_TIMESTAMP`. Those have no file fallback (git metadata genuinely isn't in the image), so `"unknown"` is the right answer for them and they need no change — but it's worth a comment saying so, since the asymmetry is what makes this bug easy to reintroduce.

## Acceptance criteria

- [ ] `docker compose -f docker-compose.prod.yml build` (no build args) → `/api/version` reports the `VERSION` file contents, not `"unknown"`
- [ ] Building **with** the build arg still wins and keeps the git suffix (`0.8.5+g<sha>[.dirty]`)
- [ ] `components.backend` / `agent_server` / `base_image` track the resolved value
- [ ] Regression test covering `VERSION=unknown` in env **plus** a present `/app/VERSION` — the exact combination that is broken today

## Environment

Trinity `0.8.5` (`e9eb40da`), prod compose, built via `docker compose build` without `start.sh`.

### LINKED ISSUE #1814: start.sh never rebuilds platform images on an in-place upgrade — prod applies no new code, and the 0.8.5 release notes omit 'docker compose build'
## Summary

`start.sh` runs `docker compose up -d` and **never rebuilds platform images**. Following the v0.8.5 release notes exactly therefore produces two different silent outcomes depending on which compose file you run:

- **Default / dev compose** — the new code *does* run (backend source is bind-mounted) and migrations apply, but `GET /api/version` keeps reporting the **old release**, because the version payload comes from build-time ARGs baked into an image that was never rebuilt. The operator has no way to confirm the upgrade landed, and the obvious check tells them it didn't.
- **Prod compose** — there is **no source bind mount**, so an in-place `git checkout` changes nothing at all. Old code keeps running, no migrations, `start.sh` prints "Ready ✅", health is green.

Found while running a full v0.8.0 → v0.8.5 in-place upgrade.

## Evidence (dev path — empirically run)

Upgrade performed exactly as `docs/releases/0.8.5.md` instructs: `git checkout` the release, `./scripts/deploy/build-base-image.sh`, `./scripts/deploy/start.sh`. Exit 0, health green.

The code *is* 0.8.5:

```
schema_migrations: 83 → 99      tables: 65 → 68      agent_reminders table created
GET /api/agents/{n}/reminders     → 200   (0.8.5-only)
GET /api/agents/{n}/label         → 200   (0.8.5-only)
GET /api/settings/retention       → 200   (0.8.5-only)
```

The reported version is not:

```
GET /api/version
  version   : 0.8.0+gb3b7078b
  backend   : 0.8.0+gb3b7078b
  base image: trinity-agent-base:0.8.0+gb3b7078b
```

That last line is the clearest tell: the base image **had just been rebuilt to 0.8.5**, and `/api/version` still names the 0.8.0 one — because the string is baked into the *backend* image at build time, not read live. Container env confirms it:

```
$ docker inspect trinity-backend --format '{{range .Config.Env}}…'
GIT_COMMIT=b3b7078b095f728c94fef20b3f62988a2c95e554     # the 0.8.0 commit
```

## Evidence (prod path — reasoned from config, not executed)

I did not stand up a prod stack, so this half is from the compose files rather than a run:

- `docker-compose.yml:261` → `- ./src/backend:/app` (source bind-mounted → checkout updates running code)
- `docker-compose.prod.yml` → **no `src/backend` mount**; `backend.build` bakes the code into the image

So on prod the same sequence leaves the old code running with no signal. Worth confirming with a real prod bring-up before acting on it.

## Why nothing catches this

- `scripts/deploy/start.sh:321` is `docker compose up -d` — no `--build`.
- `start.sh:410` prints the rebuild hint, but only as conditional troubleshooting text: *"Just pulled new code? If services fail with ModuleNotFoundError or the UI shows 'Disconnected'…"*. In this upgrade nothing failed, so the condition never fires.
- `docs/DEPLOYMENT.md:178-200` documents `docker compose build`, but frames it as the fix for a **crash-looping worker** caused by a *new dependency*, and states the deliberate decision not to auto-detect staleness (#557). The silent variants — stale version metadata, and prod-no-op — aren't covered by that framing.
- **`docs/releases/0.8.5.md` never mentions `docker compose build`.** Its only operator actions are "rebuild the base image" and "recreate agents".

## Impact

The newsletter drives upgrades as much as fresh installs. An operator who follows the release notes:

- on dev, cannot verify the upgrade and is actively told they are still on 0.8.0 — the likely reaction is to retry or roll back a *successful* upgrade;
- on prod, silently keeps running 0.8.0 — including everything 0.8.5 fixes, and #1799's log loss;
- in both cases also keeps the old agent base image, because #1809 makes the "recreate agents" step a no-op too.

Together with #1809, following the release notes exactly upgrades neither the platform (prod) nor the fleet.

## Suggested direction

- Add `docker compose build` (or `up -d --build`) to the release-notes upgrade section — smallest fix, no code.
- Make `/api/version` prefer a live source (the `VERSION` file / running commit) over stale baked ARGs so it cannot report a version that isn't running. Overlaps #1810, which is about the same payload being `unknown` when the ARGs are absent — the shared root cause is that this endpoint can only ever describe the image, not the code.
- The `scripts/deploy/upgrade.sh` already promised in DEPLOYMENT.md (backup + rebuild + start + verify) would close all of it.

## Verified good in the same run

Recording this so the upgrade path isn't assumed broken end-to-end — on the dev/SQLite path the actual upgrade is sound: every table preserved with zero row loss (`agent_ownership`, `agent_schedules`, `schedule_executions`, `chat_sessions/messages`, `agent_sessions/messages`, sharing, public links, MCP keys, subscriptions), all four agents still running, chat working, and Session-tab `--resume` continuity surviving the version change (a codeword set on 0.8.0 was recalled on 0.8.5).

dolho:
## What

Closes two ways `GET /api/version` reported a version that wasn't running. Both were observed on the same real v0.8.0 → v0.8.5 in-place upgrade, and they're opposite failures of the same resolver, so they're fixed together.

| | #1810 | #1814 |
|---|---|---|
| `VERSION` env holds | the literal sentinel `"unknown"` | a real, correct-for-that-image value (`0.8.0+gb3b7078b`) |
| Cause | `ARG VERSION=unknown` + unconditional `ENV`; truthy sentinel makes the file tier unreachable | image never rebuilt, so the baked value describes the previous build |
| Reported | `"unknown"` | `0.8.0` while 0.8.5 was executing |

Worth being explicit: **#1810's own suggested patch does not fix #1814.** Filtering the sentinel leaves `_version_env = "0.8.0+gb3b7078b"` — truthy and not the sentinel — so tier 1 still wins and the answer is still wrong. That's why this needed a rethink of the precedence rather than a one-line guard.

## The rethink

The two sources describe **different things**, and the old comment ("Env-first means dev and prod agree") assumed they describe the same thing:

- **env `VERSION`** — stamped at image build. Describes the **image**.
- **`VERSION` file** — bind-mounted live in dev (`docker-compose.yml:262` → `./VERSION:/app/VERSION:ro`), `COPY`'d into the image in prod (`docker/backend/Dockerfile:114`). In dev it describes the **code**.

So:

- `"unknown"` is treated as absence, not a value → #1810's operator gets the curated semver that was already sitting in the image.
- When the file and env **disagree**, the file wins (it's the live one) and the baked value is surfaced as a new **`image_version`** field, so drift is visible rather than smoothed over.
- When they **agree**, the env value wins so the richer `+g<sha>` provenance is preserved. Comparison is on the semver base, since only the env value carries the suffix.
- In prod both come from the same build, so they agree by construction and this branch never fires.
- An unreadable `VERSION` file can no longer 500 the endpoint operators reach for when something is already wrong.

`image_version` is additive and nullable. The `git_*` fields still describe the image — now labelled as such in a comment, since when `version != image_version` they are *not* describing `version`.

## Docs (#1814's other half)

`docs/releases/0.8.5.md` never mentioned `docker compose build`; its only operator actions were "rebuild the base image" and "recreate agents" (and per #1809 the second is a no-op too). Added an upgrade section covering:

- prod compose has **no** `src/backend` bind mount → an in-place checkout leaves the old code running, while `start.sh` reports ready and health passes;
- the default compose runs the new code but keeps stale image metadata;
- the rebuild command, and how to verify via `version` vs `image_version`.

**`start.sh` is deliberately unchanged** — the shell scripts stay as they are, per the same call made on #1788.

## Verification

**Unit** — `tests/unit/test_1810_1814_version_truth.py`, 6 cases, all passing: sentinel falls through to the file, absent env falls through, stale image reports the running code *and* keeps the drift visible, a matching build keeps the git suffix, no-stamp-no-file is still honestly `"unknown"`, and env-only (prod-style) is unchanged.

**Live** — applied to the actual upgraded instance that produced #1814, i.e. 0.8.5 code running inside a 0.8.0 image:

```
before:  version: 0.8.0+gb3b7078b     (no way to tell it was wrong)

after:   version      : 0.8.5              ← code actually running
         image_version: 0.8.0+gb3b7078b    ← stale image, now visible
         backend      : 0.8.5
         git_commit   : b3b7078b           (describes the image)
```

The instance was reverted to stock afterwards and verified clean. That the code really was 0.8.5 is independently established: migrations 83 → 99, `agent_reminders` created, and the 0.8.5-only `/reminders`, `/label` and `/settings/retention` routes all answering 200.

## Not covered

`start.sh` still doesn't rebuild, so the underlying staleness remains possible — this makes it *visible and documented* rather than preventing it. The `scripts/deploy/upgrade.sh` promised in `DEPLOYMENT.md` (backup + rebuild + start + verify) is the real fix; #1814 can stay open for it if you'd rather track it that way.

Related to #1810
Related to #1814

Fixes #1810
Fixes #1814

### DIFF
--- a/docs/releases/0.8.5.md
+++ b/docs/releases/0.8.5.md
@@ -188,6 +188,32 @@ the change is additive, and Phases 2-3 of the skill epic (#139 placement/skill-
 runner, #178 unified exposable-skills config) build on this contract, so a flag
 would have a deliberately short life._
 
+### Upgrading in place: rebuild the platform images (#1814)
+
+`./scripts/deploy/start.sh` runs `docker compose up -d` — it does **not** rebuild
+platform images. Checking out a new release and re-running `start.sh` therefore
+does not, on its own, put the new platform code into service:
+
+- On **`docker-compose.prod.yml`** there is no `src/backend` bind mount, so the
+  containers keep running the code baked into the previous build. `start.sh`
+  still reports the stack ready and health still passes.
+- On the **default compose** the backend source *is* bind-mounted, so the new
+  code does run — but the image is still the old one, so anything derived from
+  the image build (the `git_*` provenance fields on `GET /api/version`) keeps
+  describing the previous release.
+
+Rebuild the platform images as part of the upgrade:
+
+```bash
+git checkout v0.8.5
+docker compose build            # or: docker compose build backend frontend mcp-server scheduler
+./scripts/deploy/start.sh
+```
+
+Verify with `GET /api/version`: `version` reports the code in service, and
+`image_version` reports the build it is running inside. **If those two differ,
+the image is stale** — rebuild before continuing.
+
 ### Base-image rebuild + agent recreate required for several v0.8.5 features
 
 Ten files under `docker/base-image/` changed this release. A running fleet on an
--- a/src/backend/main.py
+++ b/src/backend/main.py
@@ -1279,24 +1279,56 @@ def _build_version_payload(
     import os
     from pathlib import Path
 
-    # Version resolution order (#993):
-    #   1. VERSION env var — build-stamped from git (e.g. "0.9.0+g4c640b6e"),
-    #      wired through docker-compose backend.build.args + start.sh.
-    #   2. VERSION file — curated semver, mounted in dev / copied in image.
-    #   3. "unknown" — neither present.
-    # Env-first means dev (bind-mount) and prod (build-arg) agree for the
-    # same commit instead of diverging on the file-mount being absent.
-    version = os.getenv("VERSION") or None
-    if not version:
-        version_paths = [
-            Path("/app/VERSION"),  # In container (mounted)
-            Path(__file__).parent.parent.parent / "VERSION",  # Development
-        ]
-        for version_file in version_paths:
+    # Version resolution (#993, corrected by #1810 + #1814).
+    #
+    #   env VERSION  — build-stamped from git ("0.9.0+g4c640b6e"), wired through
+    #                  docker-compose backend.build.args + start.sh. Describes
+    #                  the IMAGE.
+    #   VERSION file — curated semver. Bind-mounted live in dev
+    #                  (docker-compose.yml `./VERSION:/app/VERSION:ro`), COPY'd
+    #                  into the image in prod. In dev it describes the CODE.
+    #
+    # #1810: `ARG VERSION=unknown` + an unconditional `ENV` means the env var is
+    # always set — to the literal string "unknown" when no build arg was passed.
+    # That sentinel is truthy, so the old `or None` never fired and the file tier
+    # was unreachable for exactly the operator it was written for. Treat the
+    # sentinel as absence.
+    #
+    # #1814: `start.sh` never rebuilds platform images, so after an in-place
+    # upgrade the baked env still names the PREVIOUS build while the
+    # bind-mounted source runs the new code. The two then disagree, and the
+    # file — being live in dev — is the one describing what is actually
+    # executing. Prefer it, and keep the baked value as `image_version` so the
+    # drift is visible rather than silently smoothed over. In prod both are
+    # baked from the same build, so they agree and this branch never fires.
+    _SENTINEL = "unknown"
+
+    env_version = (os.getenv("VERSION") or "").strip()
+    if env_version == _SENTINEL:
+        env_version = ""
+
+    file_version = ""
+    version_paths = [
+        Path("/app/VERSION"),  # In container (mounted in dev, COPY'd in prod)
+        Path(__file__).parent.parent.parent / "VERSION",  # Development
+    ]
+    for version_file in version_paths:
+        try:
             if version_file.exists():
-                version = version_file.read_text().strip()
+                file_version = version_file.read_text().strip()
                 break
-    version = version or "unknown"
+        except OSError:
+            # An unreadable VERSION file must never 500 this endpoint — it is
+            # the endpoint operators reach for when something is already wrong.
+            continue
+
+    # Compare on the semver base: the env value carries a `+g<sha>` suffix the
+    # curated file does not.
+    image_version = env_version or None
+    if env_version and file_version and file_version != env_version.split("+")[0]:
+        version = file_version
+    else:
+        version = env_version or file_version or _SENTINEL
 
     git_commit = os.getenv("GIT_COMMIT", "unknown")
     git_commit_short = git_commit[:8] if git_commit != "unknown" else "unknown"
@@ -1317,6 +1349,12 @@ def _build_version_payload(
             "base_image": f"trinity-agent-base:{version}"
         },
         "runtimes": ["claude-code", "gemini-cli", "codex"],
+        # #1814: the version baked into the running image. Equal to `version`
+        # on a normally-built stack; DIFFERENT means the image is stale
+        # relative to the code being executed (an in-place upgrade without
+        # `docker compose build`), and the git_* fields below describe that
+        # older image, not `version`. None when the image carries no stamp.
+        "image_version": image_version,
         "build_date": os.getenv("BUILD_DATE", "unknown"),
         "git_commit": git_commit,
         "git_commit_short": git_commit_short,
--- a/tests/unit/test_1810_1814_version_truth.py
+++ b/tests/unit/test_1810_1814_version_truth.py
@@ -1,0 +1,108 @@
+"""
+Regression for #1810 + #1814 — `/api/version` must never name a version that
+isn't running.
+
+Two distinct ways the old resolver got it wrong, both observed on a real
+v0.8.0 → v0.8.5 upgrade:
+
+  #1810  `ARG VERSION=unknown` + an unconditional `ENV` means the env var is
+         ALWAYS set — to the literal string "unknown" when no build arg was
+         passed. That sentinel is truthy, so `os.getenv("VERSION") or None`
+         never fell through and the VERSION-file tier was unreachable for
+         exactly the operator it was added for (#993).
+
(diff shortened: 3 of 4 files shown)


vybe:
The `regression diff` red was real — 7 head failures on all 3 seeds — but both causes were test-harness, not resolver logic. Pushed `c3cdb5b7`:

1. **`test_1810_1814_version_truth.py` did `import main`**, which executes main.py's full import graph (opentelemetry, slack_sdk, …) — deliberately absent in the unit venv. All 6 tests died on `ModuleNotFoundError: opentelemetry.exporter`. Switched to the exec-slice loader `test_926_version_endpoint.py` already uses for exactly this reason (and which the builder's own docstring documents as the contract for keeping it stdlib-only).
2. **`test_926::test_version_payload_prefers_version_env` still pinned the pre-#1814 env-first contract** this PR deliberately replaces (`assert '0.8.5' == '0.9.0+gdeadbeef'`). Re-pinned to the new contract — file wins on disagreement, baked value surfaces as `image_version` — with the disagreeing env stamp derived from the repo VERSION file so the test can't rot on the next version bump.

Resolver change itself validated clean (precedence rethink is sound, sentinel-as-absence + drift surfacing both covered). Will approve/merge on green.

vybe (approved):
Validated via /validate-pr: explicit Fixes #1810 / Fixes #1814 added to the body (auto-promotion + release auto-close), issues labeled (P3+P2/type-bug), targets dev, security greps clean, no packaging/config impact. The precedence rethink is sound: sentinel-as-absence fixes #1810 without regressing prod, file-wins-on-disagreement fixes #1814 while surfacing drift via image_version instead of hiding it, env-wins-on-agreement keeps the +g<sha> provenance. Regression-diff red was test-harness only, fixed in c3cdb5b7; full matrix ×3 seeds + regression diff now green. Suggestion (non-blocking): architecture.md's /api/version catalog row could mention the new image_version field at the next docs pass.

Issue #1028, opened 2026-06-02
refactor: split oversized backend routers & services (settings, public, agent crud, git, agent_client)


vybe:
## Summary

A cluster of backend routers and services exceed the size/complexity thresholds (critical + high) per a `/refactor-audit` scan (2026-06-02). None are on the hot orchestration path (tracked separately), so these are lower-risk structural splits — but several are large enough to no longer fit in one AI context window.

## Context

These files accumulate handlers/logic over many features and have grown past the point where they're reviewable or AI-maintainable as a unit. Splitting them by domain follows the three-layer pattern (Invariant #1) and route-ordering rule (Invariant #4).

## Findings (critical/high)

**Oversized files:**
| Logical lines | File |
|--------------:|------|
| 1,196 | `src/backend/routers/settings.py` |
| 876 | `src/backend/routers/public.py` |
| 834 | `src/backend/main.py` |
| 831 | `src/backend/services/git_service.py` |
| 891 | `src/backend/services/agent_client.py` |
| 793 | `src/backend/routers/ops.py` |

**Long / complex functions:**
| Function | Lines | Cx | Location |
|----------|------:|---:|----------|
| `create_agent_internal` | 714 | 111 | `src/backend/services/agent_service/crud.py:116` |
| `lifespan` | 414 | 46 | `src/backend/main.py:307` (startup wiring) |
| `deploy_local_agent_logic` | 306 | 40 | `src/backend/services/agent_service/deploy.py:301` |
| `public_chat` | 268 | 33 | `src/backend/routers/public.py:404` |
| `system_agent_terminal` | 265 | — | `src/backend/routers/system_agent.py:243` |
| `handle_terminal_session` | 262 | — | `src/backend/services/agent_service/terminal.py:58` |
| `send_session_message` | 231 | — | `src/backend/routers/sessions.py:513` |
| `recreate_container_with_updated_config` | 210 | 39 | `src/backend/services/agent_service/lifecycle.py:336` |
| `_render_forensic` | 183 | 66 | `src/backend/services/canary_alerts.py:252` |
| `check_network_health` / `perform_health_check` | 185 / 177 | — | `src/backend/services/monitoring_service.py:198 / 580` |

## Acceptance Criteria

- [ ] `routers/settings.py` split per settings domain (Slack transport / resources / feature flags / MCP URL), preserving route-registration order
- [ ] `services/agent_service/crud.create_agent_internal` decomposed (it is the longest function in the codebase)
[...]

PR #1818, opened 2026-07-27
fix(recovery): stop recreate_missing_container rebuilding a divergent agent (#1811)

### LINKED ISSUE #1811: bug: #1559 recreate_missing_container rebuilds a divergent container — drops the encrypted-data mount and TEMPLATE_NAME, mints a new MCP key each call
## Summary

`recreate_missing_container` (added by #1559 to unblock soft-delete recovery) rebuilds an agent's container spec **from scratch** rather than from the creation path, and the reconstruction is missing pieces that `create_agent_internal` always sets. An agent that goes through recovery therefore comes back subtly different from an identical agent that never did — silently, with no warning and no diff surfaced to the operator.

Observed on a live instance while upgrading v0.8.0 → v0.8.5. One agent (`receptionist`) had no container and so took this path; the other 22 went through `recreate_container_with_updated_config`. Diffing the before/after `docker inspect` of the recovered one:

```
ENV KEYS LOST : ['TEMPLATE_NAME']
MOUNTS BEFORE : ['encrypted-data', 'agent-receptionist-workspace', …]
MOUNTS AFTER  : ['agent-receptionist-workspace']
```

The workspace volume is correctly preserved (that part of #1559/#1665 works). The gaps are elsewhere.

## Divergences from `create_agent_internal`

**1. The `encrypted-data:/data` mount is dropped.**

`services/agent_service/crud.py:1277-1279` gives every agent this mount:

```python
volumes = {
    str(config_path): {'bind': '/config/agent-config.yaml', 'mode': 'ro'},
    str(credentials_path): {'bind': '/config/credentials.json', 'mode': 'ro'},
    'encrypted-data': {'bind': '/data', 'mode': 'rw'},
}
```

`lifecycle.py:919-925` (`recreate_missing_container`) builds `base_volumes` with the workspace only. Low impact on the instance I hit this on — `/data` was empty in both a recovered and a non-recovered agent, so nothing appears to write there today — but it is an unintended fleet inconsistency, and it means any future use of `/data` would fail *only* on previously-recovered agents. Worth confirming whether the mount is still wanted at all; if it is dead, the honest fix may be deleting it from `crud.py` rather than adding it here.

**2. `TEMPLATE_NAME` is not restored.**

Set at creation (`crud.py:911`) as `config.template`, e.g. `github:Abilityai/cleon-reception@main`. Two consumers:

- `docker/base-image/startup.sh:341` — `elif [ -n "${TEMPLATE_NAME}" ] && [ -d "/template" ]` gates local-template initialisation
- `docker/base-image/agent_server/routers/info.py:182` — `"template_name": os.getenv("TEMPLATE_NAME", "")` feeds `/info`

`recreate_missing_container` reads `_template` out of the volume's `template.yaml` into a local `template_name` and uses it **only** for the `trinity.template` label (`lifecycle.py:906`), never putting it back in `env_vars` (`lifecycle.py:857-867`). So a recovered agent reports an empty `template_name` and skips the local-template branch in `startup.sh`.

**3. A fresh agent-scoped MCP key is minted on every call** (`lifecycle.py:884-891`, `description="recovery-recreate"`).

Deliberate — the old key's plaintext is unrecoverable, only the hash is stored — but it means recovery is **not idempotent**: each invocation leaves another active key row behind. On the instance above the key table had grown to 50 active keys. Given #1745 has just made agent deletion revoke MCP keys, the same hygiene question applies to repeated recoveries. At minimum the superseded key should be deactivated when the new one is minted.

## Why this is easy to miss

The path only runs for an agent whose container is **gone** — soft-delete recovery, or an operator/GC removing a container out of band. Everything else goes through `recreate_container_with_updated_config`, which copies `Config.Env` and `Mounts` off the old container and so cannot drift this way. The two functions are 400 lines apart in the same module and share no spec.

## Suggested fix

Shortest correct version: add the missing mount and re-set `TEMPLATE_NAME` from the value already read off `template.yaml`.

Structurally better: have both recreate paths and `create_agent_internal` derive their container spec from **one** shared builder, so "what a Trinity agent container looks like" has a single definition. #1484 already broke `create_agent_internal` into fenced phase helpers (`_apply_github_env`, `_create_agent_container`, …) — reusing those from the recovery path is the natural follow-up, and `crud.py` module-splitting is already tracked in #1028.

A cheap regression guard, independent of either fix: a test asserting the env-key set and mount set produced by `recreate_missing_container` are supersets of what `create_agent_internal` produces for the same agent config. That fails today and would have caught all three.

## Acceptance criteria

- [ ] A recovered agent's mount set matches a freshly created agent's (modulo the transient creation-time config/credential binds)
- [ ] `TEMPLATE_NAME` is restored, so `/info` reports `template_name` and `startup.sh:341` behaves identically
- [ ] Repeated recovery does not accumulate active MCP keys (supersede or deactivate the prior key)
- [ ] Parity test comparing the two spec builders

## Environment

Trinity `0.8.5+ge9eb40da`, prod compose, SQLite. Recovery triggered by removing a stopped agent's container and calling `POST /api/agents/{name}/start`.

dolho:
## What

All four acceptance criteria from #1811, verified live by reproducing the reporter's exact trigger (remove a stopped agent's container, `POST /api/agents/{name}/start`).

Two of the three fixes are **not** what the issue proposed, for reasons below.

## 1. `encrypted-data:/data` — removed from creation, not added to recovery

The issue asked: *"Worth confirming whether the mount is still wanted at all; if it is dead, the honest fix may be deleting it from `crud.py` rather than adding it here."*

It is dead, and it is worse than dead:

- Nothing under `docker/base-image/` references `/data` — the Dockerfile only `mkdir`s it. The reporter's own observation (empty in both agents) matches.
- The volume name is a **literal**, not per-agent. So a single volume was mounted **rw into every agent simultaneously** — a cross-agent read/write surface in a product whose premise is per-agent container isolation. Nothing writes there *today*; that is not a guarantee.

Removed at the source. Both paths now agree, and the surface is gone. The volume itself isn't deleted, so anything historically written to it remains on the host.

## 2. `TEMPLATE_NAME` — the suggested fix alone doesn't work

The issue proposed re-setting it "from the value already read off `template.yaml`". I did that first and verified it live: **still empty**.

`_read_template_yaml_from_volume` returns the workspace `template.yaml`, which carries `name:` and `type:` — never `_template`. So `tmpl.get("_template")` is empty in practice, which is also why `trinity.template` came back blank. The original id (`local:scout`, `github:Org/repo@main`) existed **only** in the destroyed container's env and label.

`_reconstruct_template_id` rebuilds it from what survives:

- GitHub-native agent → persisted `agent_git_config` (`github:{repo}@{working_branch}`)
- otherwise → `local:{name}` from the volume's `template.yaml`
- nothing identifying → `""`, the honest answer and no worse than before

Being explicit: this is a **reconstruction, not the original string**. A github agent created without a branch suffix gets one back. The authoritative fix is persisting the template id at creation, which needs a schema change — happy to file that as a follow-up.

## 3. MCP keys — deactivate *before* mint

Order is load-bearing: `_deactivate_agent_keys_in_txn` flips **every** agent/connector-scoped key for the agent, so running it after the mint would deactivate the key just issued. Reuses #1745's in-txn primitive via a new standalone-transaction accessor (`deactivate_agent_mcp_keys`), best-effort so credential bookkeeping can never block a recovery. There's a test pinning the ordering.

## 4. Parity guard

`tests/unit/test_1811_recovery_container_parity.py` — the "cheap regression guard" the issue asked for. Source-level (AST) rather than behavioural: driving both builders end-to-end needs the full docker/db mock harness, while the actual failure mode ("someone adds an env var to one path and not the other") is visible in the source and cheap to check every run. It asserts creation's env keys are covered by recovery, `TEMPLATE_NAME` among them, neither path mounts the shared volume, and the deactivate-then-mint order.

It earned its keep immediately: it flagged `AGENT_GUARDRAILS` and `AGENT_TOOL_STALL_LIMIT_S` as divergent. Both turned out to be genuinely supplied at runtime by `_apply_persisted_auth_env` (lifecycle.py:491, 501) — I had guessed at those names in the allowlist instead of reading them, and the test caught my error, not the code's. Now recorded with verified names and line references.

## Verification

**Unit** — 44 passed: the 4 new parity cases plus the full #1484 create-agent characterization suite (unaffected by the mount removal). `tests/lint_sys_modules.py` reports no new violations.

**Live**, reproducing the reporter's trigger on a real instance:

```
before recovery:  TEMPLATE_NAME=local:scout   active agent keys: 1

after recovery (fixed):
  TEMPLATE_NAME = local:scout
  trinity.template = 'local:scout'
  mounts: agent-up-beta-workspace -> /home/developer     (no encrypted-data)
  active agent keys: 1        ← was 1, would have been 2
  log: "Deactivated 1 superseded MCP key(s) for up-beta before recovery mint (#1811)"
```

A **second** recovery left the key count at 1, confirming the operation is now idempotent. The instance was reverted to stock afterwards and verified clean.

## Not done

The issue's structural suggestion — one shared container-spec builder across `create_agent_internal` and both recreate paths — is the real cure and is out of scope here; #1028 already tracks the `crud.py` split that would enable it. This PR fixes the three concrete divergences and adds the guard that makes the next one fail loudly instead of silently.

Fixes #1811

### DIFF
--- a/src/backend/database.py
+++ b/src/backend/database.py
@@ -561,6 +561,10 @@ class DatabaseManager:
     def delete_agent_ownership(self, agent_name: str):
         return self._agent_ops.delete_agent_ownership(agent_name)
 
+    def deactivate_agent_mcp_keys(self, agent_name: str) -> int:
+        """Deactivate an agent's agent/connector-scoped MCP keys (#1811)."""
+        return self._agent_ops.deactivate_agent_mcp_keys(agent_name)
+
     def purge_agent_ownership(self, agent_name: str):
         return self._agent_ops.purge_agent_ownership(agent_name)
 
--- a/src/backend/db/agents.py
+++ b/src/backend/db/agents.py
@@ -395,6 +395,18 @@ class AgentOperations(
             .values(is_active=1 if active else 0)
         ).rowcount or 0
 
+    def deactivate_agent_mcp_keys(self, agent_name: str) -> int:
+        """Deactivate this agent's agent- and connector-scoped MCP keys (#1811).
+
+        Standalone-transaction wrapper over `_deactivate_agent_keys_in_txn`, for
+        callers outside a delete/recover transaction — specifically the recovery
+        recreate, which mints a fresh key on every invocation (the old key's
+        plaintext is unrecoverable) and previously left every superseded row
+        active. Returns the number of rows flipped.
+        """
+        with get_engine().begin() as conn:
+            return self._deactivate_agent_keys_in_txn(conn, agent_name, active=False)
+
     def delete_agent_ownership(self, agent_name: str) -> bool:
         """Soft-delete the agent ownership row (Issue #834 Phase 1a).
 
--- a/src/backend/services/agent_service/crud.py
+++ b/src/backend/services/agent_service/crud.py
@@ -1272,10 +1272,19 @@ async def _build_volume_mounts(
     # /home/developer lives on the container writable layer (overlayfs),
     # auto-reclaimed by container removal. Volumes exist to survive
     # recreate, and ghosts never recreate.
+    # #1811: the `encrypted-data:/data` mount was removed rather than copied
+    # into the recovery path. It was dead AND unsafe:
+    #   * nothing in the agent image ever touched /data — the Dockerfile only
+    #     `mkdir`s it, and no code in docker/base-image references it;
+    #   * the volume name was a LITERAL, so a single volume was mounted rw into
+    #     every agent at once — a cross-agent read/write surface in a product
+    #     whose premise is per-agent isolation. Unused today is not a guarantee.
+    # Removing it here (instead of adding it to recreate_missing_container)
+    # makes both paths agree and closes the surface. The volume itself is not
+    # deleted, so anything historically written to it remains on the host.
     volumes = {
         str(config_path): {'bind': '/config/agent-config.yaml', 'mode': 'ro'},
         str(credentials_path): {'bind': '/config/credentials.json', 'mode': 'ro'},
-        'encrypted-data': {'bind': '/data', 'mode': 'rw'},
     }
     if not config.ephemeral:
         await _workspace_volume_mount(config, volumes)
--- a/src/backend/services/agent_service/lifecycle.py
+++ b/src/backend/services/agent_service/lifecycle.py
@@ -775,6 +775,28 @@ def _workspace_volume_name(agent_name: str) -> str:
         return f"agent-{agent_name}-workspace"
 
 
+def _reconstruct_template_id(agent_name: str, tmpl: dict) -> str:
+    """Best-effort rebuild of an agent's template id after its container is gone (#1811).
+
+    Returns "" when nothing identifying survives — the honest answer, and the
+    same value the caller had before, so this can only improve the result.
+    """
+    try:
+        git_cfg = db.get_git_config(agent_name)
+    except Exception:  # noqa: BLE001 — never block recovery on this
+        git_cfg = None
+
+    repo = getattr(git_cfg, "github_repo", None) if git_cfg else None
+    if repo:
+        branch = getattr(git_cfg, "working_branch", None) or getattr(
+            git_cfg, "source_branch", None
+        )
+        return f"github:{repo}@{branch}" if branch else f"github:{repo}"
+
+    name = (tmpl.get("name") or "").strip()
+    return f"local:{name}" if name else ""
+
+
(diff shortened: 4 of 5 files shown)


vybe (approved):
Validated via /validate-pr: real #1811 fix with a named regression suite (AST drift-guard for env parity, mount-removal assertion, key-deactivation ordering). Verified the removed encrypted-data mount has no other consumers in the tree; full unit suite green across 3 seeds with zero regressions. Approving to land as a release-day bug fix.

Issue #1817, opened 2026-07-27
refactor: per-agent start lock — complete the concurrent-start race hardening from #1809


webmixgamer:
## Summary
#1809 shipped minimal recreate-race hardening: `container_remove` tolerates `NotFound` and a 409 name-conflict on the run adopts the winner's container. That converts the loser's 500 into a graceful success, but two concurrent starts can still both enter the recreate sequence (double stop/remove work, interleaved injections).

The full fix is a per-agent start lock serializing `start_agent_internal`, following the existing Redis SETNX+TTL precedent (`agent:data_op:{name}` in `routers/agent_data.py`).

## Why deferred from #1809
New lock infrastructure on a hot path has its own failure modes (stuck lock blocks starts; TTL sizing; Redis-down behavior; wait-vs-409 contention semantics) — deliberately kept out of a P2 bug fix. See the #1809 plan's decision trail.

## Design notes
- Key: `agent:start:{name}` (register in `services/agent_runtime_state.py` per the #1560 keyspace-parity rule — `tests/unit/test_1560_agent_redis_key_parity.py` fails CI on unregistered `agent:*` keyspaces)
- Fail-open on Redis down (start must never be blocked by lock infrastructure)
- TTL ≥ worst-case recreate duration (stop timeout + run + readiness poll)
- Second caller: prefer short wait-then-adopt over an immediate 409 ("start" is idempotent from the caller's view)

## Acceptance criteria
- [ ] Concurrent starts of the same agent serialize; exactly one performs a needed recreate
- [ ] Lock failure modes fail open (never blocks a start when Redis is down; no stuck-lock lockout beyond TTL)
- [ ] Keyspace registered in `agent_runtime_state.py` with parity-test coverage

Refs #1809

webmixgamer:
Heads-up from #1860 (PR #1912): `lifecycle.restart_agent_internal(name, *, stop_timeout=30)` now exists as the canonical stop→cold-start helper, consumed by `restart_fleet`. Its docstring names this issue as the future home of the per-agent start lock — when the lock lands there, the remaining inline copies of the stop→`start_agent_internal` shape can consolidate onto it in the same pass: `routers/system_agent.py` (restart/reinitialize), `routers/systems.py` (restart_system loop), and `services/subscription_auto_switch.py` (which additionally holds its own #799 process-local lock and uses stop timeout 10 — parameterized via `stop_timeout`).

AndriiPasternak31:
[...]

Issue #1816, opened 2026-07-27
bug: trinity-system never adopts a rebuilt base image — ensure_deployed short-circuits on 'already running'


webmixgamer:
## Summary
`system_agent_service.ensure_deployed()` returns `action: none` the moment the `trinity-system` container reports `running` — no drift predicates are evaluated. Since the system agent runs with `unless-stopped` semantics and is never manually cycled, it is the most-stale agent in every fleet: after a `build-base-image.sh`, **trinity-system keeps the old image indefinitely**, across backend restarts, unless an operator explicitly stops and starts it.

#1809 fixed image adoption for the per-agent start path (`start_agent_internal`, cold-start-gated ninth predicate). The system-agent bootstrap path was explicitly left out of that fix's scope.

## Where
- `src/backend/services/system_agent_service.py` — `ensure_deployed()` short-circuit (~lines 86-95)
- The fix should reuse `check_base_image_matches` (`services/agent_service/helpers.py`, #1809) rather than fork the comparison, and respect the same principle: never kill a system agent mid-work — adopt at a safe boundary (e.g. detect at bootstrap, recreate only when stopped, or log + surface).

## Acceptance criteria
- [ ] After a base-image rebuild, trinity-system adopts the new image at the next safe opportunity (backend boot with the container stopped, or an explicit start cycle)
- [ ] A running trinity-system is never image-recreated mid-operation without an explicit stop
- [ ] Adoption path reuses the #1809 predicate (no forked comparison logic)

Refs #1809

AndriiPasternak31:
/claim — picking this up (AndriiPasternak31). Worktree + plan in progress; will open a draft PR with `Fixes #1816`.

PR #1820, opened 2026-07-27
fix(agents): rebuilt base image never picked up — adopt on cold start via ninth recreate predicate (#1809)

### LINKED ISSUE #1809: bug: a rebuilt agent base image is never picked up — no image-drift predicate, so the v0.8.5 "stop/start-with-recreate" upgrade step is a no-op

webmixgamer:
## Summary
- `needs_recreation` never checked the image, so after `build-base-image.sh` a plain stop/start silently kept every agent on the old image — the v0.8.5 release-note upgrade step described a mechanism that did not exist (reproduced live before the fix; issue #1809 has the full repro).
- Adds a ninth, **cold-start-gated** predicate: `check_base_image_matches` compares the container's resolved image ID against what its **own** `Config.Image` reference resolves to now. Fail-open on every unreadable state (WARNING-logged; `ImageNotFound` would fail the recreate identically), INFO log on drift. Async via a new `docker_utils.image_get` executor wrapper — the first predicate doing a daemon round-trip, evaluated lazily (only when no other predicate fired) and **only for cold starts**: start-on-running stays an idempotent no-op (MCP ensure-running calls, the SUB-003 auto-switch restart, `restart_system` all start running agents), and ephemeral ghosts are excluded (volume-less — a recreate would destroy their workspace).
- Recreate hardening that the new frequency makes routine: `container_remove` tolerates `NotFound` and a 409 name-conflict adopts the concurrent winner's container; the `trinity.base-image-version` label is refreshed from the adopted image (both AgentStatus readers prefer the container label — the UI otherwise reports the old version while running the new image).
- Observability: `recreated` + `recreate_reason` (`config_drift` | `image_drift`) surfaced in the start response and the `agent_lifecycle:start` audit event.
- Docs: the 0.8.5 release note is corrected as a point-in-time correction and now states explicitly that **Operating-Room fleet restart does NOT adopt a rebuilt image**; architecture + feature-flow docs updated; learnings-ledger entry added.

## Changes
- `src/backend/services/docker_utils.py` — async `image_get` wrapper
- `src/backend/services/agent_service/helpers.py` — `check_base_image_matches`
- `src/backend/services/agent_service/lifecycle.py` — lazy cold-start wiring, ephemeral guard, label refresh, race tolerance, response surfacing
- `src/backend/routers/agents.py` — audit details + response fields
- `tests/unit/test_1809_image_drift_recreate.py` — 19 tests (predicate matrix + wiring source-pins)
- `tests/unit/{test_start_agent_skip_inject,test_agent_readiness_probe,test_subscription_auto_switch_no_cred_import,test_monitoring_router_signatures}.py` — enumerated import-stubs gain the new names
- `docs/…` — release-note correction, architecture, 2 feature flows, learnings, CSO diff report (0 findings)

## Test Plan
- [x] New suite green: `pytest tests/unit/test_1809_image_drift_recreate.py` (19 tests)
- [x] Full unit suite matches the merge-base failure-profile exactly: 4931 passed, 1 pre-existing env failure (`Europe/Kiev` tzdata case — fails identically at `e9eb40da`)
- [x] Live e2e on a running stack: (1) start on RUNNING agent with drifted tag → no-op; (2) cold start with drift → recreated, `reason: image_drift`, version label refreshed, workspace volume intact; (3) stop/start with matching image → no churn; (4) tag restored → adopts back
- [x] `/review` clean (2 informational, fixed) · `/cso --diff` — 0 findings (report in `docs/security-reports/`)

## Follow-ups (filed)
- trinity-enterprise#252 — drain-aware rolling fleet image upgrade + out-of-date badge (Ops fleet-restart still bypasses adoption)
- #1816 — `trinity-system` never adopts (ensure_deployed short-circuit)
- #1817 — per-agent start lock (full race hardening)

Fixes #1809


AndriiPasternak31 (approved):
## `/validate-pr` — APPROVE (no blockers)

CI 17/17 green (6 pytest runs across base+head × 3 seeds, regression-diff, prod-image-smoke, CodeQL, gitleaks, schema-parity). Security greps clean; no new top-level backend module, no new `os.getenv`, no infra changes. `Fixes #1809` is same-repo, so the issue auto-promotes to `status-in-dev` on merge.

Verified against source rather than the diff alone: `import docker` is present (`lifecycle.py:11`), so the new `except docker.errors.*` clauses resolve; `image` is assigned at `:440` and passed through `validate_base_image` (SEC-172) before the new label-refresh reads it; `was_already_running` was already computed at `:285` ahead of the predicate block; and the recreate re-bakes `CLAUDE_CODE_OAUTH_TOKEN` from the DB at `:479`, so the now-more-frequent recreates do not regress the #1089 hot-rotated token when the writable layer is wiped. Mechanism is sound — fail-open throughout, lazily evaluated, correctly gated.

Merging now. Four non-blocking findings for follow-up:

**1. The comment justifying the cold-start gate names two paths that are themselves cold starts.** `lifecycle.py:300-304` (and the PR body) cite "the SUB-003 auto-switch restart, `restart_system`" as start-on-running no-ops. Both stop first — `subscription_auto_switch.py:310-311` and `routers/systems.py:236-239` each do `container_stop` → `start_agent_internal`. So after any `build-base-image.sh`, a `restart_system` call and a SUB-003 fallback restart **will** image-recreate. The outcome is benign (the recreate re-bakes env from the DB, which is what auto-switch wants) and the gate still protects the genuine case — the start endpoint hit on a running agent — but the stated blast radius is understated and the comment will mislead the next reader. The release note's separate claim about the *Operating Room* fleet restart (`routers/ops.py`, raw stop+start) not adopting is accurate and distinct.

**2. The ephemeral guard fails toward recreating a ghost.** `lifecycle.py:311-315` — if `db.get_agent_ephemeral_info` raises, `_eph_gate = None`, the `is_ephemeral` test is false, and the image check proceeds, potentially recreating a volume-less ghost and destroying its workspace mid-budget. That inverts the fail-open philosophy `check_base_image_matches`'s own docstring states ("any unreadable state returns True"). Low probability and enterprise-gated, but the safe direction is one line.

**3. Feature-flows index not updated.** `agent-lifecycle.md` and `async-docker-operations.md` both changed, but `docs/memory/feature-flows.md` gets no **Recent Updates** row — the convention every recent entry follows.

**4. Wiring coverage is source-pin-only.** The 10 predicate tests are genuinely good (drift, all three fail-open branches, the ID-pinned tautology, version-pinned tag isolation). But nothing drives `start_agent_internal` with a drifted container to assert it recreates with `recreate_reason == "image_drift"` — the regexes would still pass if the block became unreachable. `test_start_agent_skip_inject.py` already has the harness and only got a default-True stub.

Trivial: `tests/registry.json` lost its trailing newline.

Issue #1640, opened 2026-07-15
feat: set an agent's display label at creation time (#964 follow-up 2/5 — residual)


vybe:
## Summary

Give owners a way to actually set an agent's display name: an owner-only endpoint plus the two input surfaces (agent creation and rename). Without this the column from the schema issue exists but nothing can fill it.

Follow-up 2 of 5 from the design decision on #964.

## Context

Today `src/frontend/src/components/CreateAgentModal.vue` has the only agent-name input — it's labelled "Agent Name" but its placeholder is a slug, which is exactly the confusion #964 was reported for. `src/frontend/src/components/AgentHeader.vue` has the only rename affordance and edits the slug. Both need to handle two distinct fields and say which is which.

## Acceptance Criteria

- [ ] `PUT /api/agents/{name}/display-name` added — owner-only (`OwnedAgentByName`), accepts a nullable string, `NULL`/empty clears back to the fallback chain
- [ ] Length cap enforced with a **named validation error** (not a generic 422 blob) on over-length input
- [ ] Unicode handling + trim/normalization policy decided and enforced (leading/trailing whitespace trimmed; control chars rejected)
- [ ] Uniqueness policy decided and documented — display names are **not** required unique (the slug already guarantees uniqueness); if that changes, state why
- [ ] Setting a display name is audit-logged
- [ ] `CreateAgentModal.vue` gains an optional display-name field; existing name input relabelled to "Slug" / "Identifier" with helper text explaining it can't change casually
- [ ] `AgentHeader.vue` rename affordance edits **both** fields, clearly labelled; slug validation unchanged
- [ ] Display name is optional everywhere — an agent with none behaves exactly as today (graceful fallback, no dead state)
- [ ] MCP tool for setting it (or an explicit decision to keep this UI/REST-only, recorded on the issue)
- [ ] `docs/memory/architecture.md` API Endpoints table updated

## Technical Notes

Endpoint slots naturally beside the other per-agent owner-only settings (`agent_config.py` family). Follow the three-layer pattern (Invariant #1) and the `OwnedAgentByName` dependency (Invariant #8) — note the enumeration-safety rule: uniform 404, never a 404-then-403 split.

Per the #964 decision, `display_name` is **not** part of the slug rename cascade — renaming the slug leaves the display name intact. The rename UI should make that obvious rather than implying the two are coupled.

[...]

PR #1822, opened 2026-07-27
fix(label): reject a mistyped label body instead of silently clearing (#1821)

### LINKED ISSUE #1821: PUT /agents/{name}/label silently clears the label (200) when the body uses an unknown field name
## Summary

`PUT /api/agents/{name}/label` returns **200 and silently clears the label** when the request body uses any field name other than `label`. Because `label=None` *means* "clear it", an unrecognised body is indistinguishable from a deliberate clear, and the caller gets a success response for the opposite of what they asked.

Found while testing the v0.8.5 display-label feature — I hit it by sending `display_label` (the DB column name) instead of `label`.

## Repro

```
label before                          : Alpha (Research Lead)
PUT {"display_label":"Typo Name"}  →  HTTP 200
label after                           : None
PUT {}                             →  HTTP 200
```

The response body even reflects the wipe rather than the request:

```json
{"agent_name":"up-alpha","label":null,"display_name":"up-alpha"}
```

## Root cause

`AgentLabelUpdate` (`models.py:158`) is a plain Pydantic model, so unknown fields are ignored (`extra="ignore"` by default) and `label` falls back to its `None` default — which the handler correctly treats as "clear".

The clear-on-None semantic is fine and documented. The problem is that it is reachable by accident: a typo, an older client, or a caller who guessed the field name from the `display_label` DB column / the `display_name` response field, all get a silent wipe with a success code.

## Impact

Low: one string is lost, and re-setting it takes a second. Raising it because the shape is the same silent-success pattern as #1796/#1806/#1808 — the API reports success for an action it did not perform — and the fix is one line.

Note the field naming makes the mistake easy: the column is `display_label`, the response field is `display_name`, and the accepted request field is `label`.

## Suggested fix

Either reject unknown fields:

```python
class AgentLabelUpdate(BaseModel):
    model_config = ConfigDict(extra="forbid")
```

so a typo 422s instead of wiping, or require the field explicitly and give clearing its own affordance (`DELETE`, or an explicit `{"label": null}` that must be present in the payload). The first is a one-liner and keeps the existing clear semantics.

## Environment

v0.8.5 code (`b8cf94ca`), dev compose, SQLite.

dolho:
## What

`PUT /api/agents/{name}/label` no longer returns 200 and silently wipes the label when the body uses a field name it doesn't recognise.

## Why

`label=None` means "clear", and the model ignored unknown fields with `label` defaulting to `None` — so a malformed body and a deliberate clear were the same request. I hit this by accident while testing the feature, sending `display_label` (the DB column name):

```
label before                        : Alpha (Research Lead)
PUT {"display_label":"Typo Name"} → HTTP 200
label after                         : None
```

The naming makes it easy: the column is `display_label`, the response field is `display_name`, only `label` is accepted.

## Change

- `model_config = ConfigDict(extra="forbid")` — an unknown field 422s.
- `label` becomes **required-but-nullable** — `{}` also 422s, since it carried the same ambiguity and was never a documented way to clear.

Clearing is unchanged in meaning and still supported; it just has to be explicit.

## Compatibility

- The frontend already sends `{ label }` on every call (`stores/agents.js:393`), including when clearing.
- No MCP tool sets labels.
- Behaviour change worth naming: `PUT {}` now 422s where it used to clear. That path was indistinguishable from a typo, which is the bug.

## Verification

**Unit** — `tests/unit/test_1821_label_update_strictness.py`, 7 cases: unknown field rejected, empty body rejected, explicit null still clears, blank string still clears, setting works, the ent#181 normalizer still runs, and control characters are still rejected (strictness must not bypass the existing validator).

One test of mine was wrong on the first pass — I asserted the normalizer collapses internal whitespace; it only trims and NFC-normalizes. Corrected to the real behaviour rather than changing the code to match my assumption.

**Live**, reproducing the issue exactly:

```
PUT {"display_label":"Typo"}   -> 422   (was 200 + silent wipe)
PUT {}                         -> 422
label survived                 : Alpha (Research Lead)
PUT {"label":null}             -> 200
label now                      : None
```

Instance reverted to stock afterwards.

Fixes #1821

### DIFF
--- a/src/backend/models.py
+++ b/src/backend/models.py
@@ -5,7 +5,7 @@ import os
 import re
 import unicodedata
 
-from pydantic import BaseModel, EmailStr, Field, SecretStr, field_validator, model_validator
+from pydantic import BaseModel, ConfigDict, EmailStr, Field, SecretStr, field_validator, model_validator
 from typing import Any, Dict, List, Literal, Optional, Union
 from datetime import datetime
 from enum import Enum
@@ -262,8 +262,18 @@ class AgentLabelUpdate(BaseModel):
     `label=None` (or blank) clears it, and the agent renders under its slug
     again. Presentation only: the slug never moves, which is the entire point —
     a slug rename re-keys ~20 tables and strands the agent's volumes (#1664).
+
+    #1821: `label` is REQUIRED-but-nullable, and unknown fields are rejected.
+    Clearing is a legitimate operation expressed as an explicit null, but with
+    an ignored-extras model and a `None` default it was also what you got from
+    any body the server did not recognise — so `{"display_label": "..."}` (an
+    easy mistake: `display_label` is the DB column and `display_name` the
+    response field) returned 200 and silently wiped the label. Both an unknown
+    field and an empty `{}` now 422 instead of destroying data.
     """
-    label: Optional[str] = None
+    model_config = ConfigDict(extra="forbid")
+
+    label: Optional[str]
 
     @field_validator("label")
     @classmethod
--- a/tests/unit/test_1640_display_label_at_creation.py
+++ b/tests/unit/test_1640_display_label_at_creation.py
@@ -92,9 +92,23 @@ def test_agentconfig_blank_label_clears(M):
 def test_label_update_shares_policy(M):
     assert M.AgentLabelUpdate(label="  Hi  ").label == "Hi"
     assert M.AgentLabelUpdate(label="").label is None       # clear
-    assert M.AgentLabelUpdate().label is None
+    assert M.AgentLabelUpdate(label=None).label is None     # clear, said explicitly
     with pytest.raises(Exception):
         M.AgentLabelUpdate(label="bad\ttab")
+
+
+def test_label_update_requires_the_label_field(M):
+    """#1821: an omitted or misnamed `label` no longer means "clear".
+
+    It used to, which made a typo'd field name — `display_label`, the DB column,
+    right next to the `display_name` response field — indistinguishable from a
+    deliberate clear: the unknown key was dropped, `label` defaulted to None, and
+    the PUT wiped the label while answering 200. Clearing is still supported and
+    unchanged in meaning (see the two cases above); it just has to be stated."""
+    with pytest.raises(Exception):
+        M.AgentLabelUpdate()                              # empty body
+    with pytest.raises(Exception):
+        M.AgentLabelUpdate(display_label="Typo Name")     # the actual #1821 trigger
 
 
 def test_not_unique_policy(M):
--- a/tests/unit/test_1821_label_update_strictness.py
+++ b/tests/unit/test_1821_label_update_strictness.py
@@ -1,0 +1,65 @@
+"""
+Regression for #1821 — a mistyped label body must not silently clear the label.
+
+`AgentLabelUpdate` ignored unknown fields and defaulted `label` to None, and
+None *means* "clear". So a body the server did not recognise was
+indistinguishable from a deliberate clear:
+
+    PUT {"display_label": "Typo Name"}  ->  200, label wiped
+
+That mistake is easy to make: the DB column is `display_label`, the response
+field is `display_name`, and only `label` is accepted.
+
+Clearing stays supported — it just has to be said explicitly.
+"""
+
+from __future__ import annotations
+
+import pytest
+from pydantic import ValidationError
+
+
+def _model():
+    from models import AgentLabelUpdate
+
+    return AgentLabelUpdate
+
+
+def test_unknown_field_is_rejected_not_treated_as_a_clear():
+    """The exact bug: the field name from the DB column must not wipe the label."""
+    with pytest.raises(ValidationError):
+        _model()(display_label="Typo Name")
+
+
+def test_empty_body_is_rejected():
+    """`{}` was also a silent clear — ambiguous with a malformed request."""
+    with pytest.raises(ValidationError):
+        _model()()
+
+
+def test_explicit_null_still_clears():
+    """Clearing is legitimate — it just has to be explicit."""
+    assert _model()(label=None).label is None
+
+
+def test_blank_string_still_clears():
+    """Documented clear-by-blank behaviour is unchanged (normalizer maps to None)."""
+    assert _model()(label="   ").label is None
+
+
+def test_setting_a_label_still_works():
+    assert _model()(label="Alpha (Research Lead)").label == "Alpha (Research Lead)"
+
+
+def test_label_is_still_normalized():
+    """The ent#181 normalizer must still run — strictness must not bypass it.
+
+    It trims and NFC-normalizes; it does NOT collapse internal whitespace.
+    """
+    assert _model()(label="  Alpha Lead  ").label == "Alpha Lead"
+
+
+def test_control_characters_are_still_rejected():
+    """The normalizer's own guard must survive the stricter model config."""
+    with pytest.raises(ValidationError):
+        _model()(label="bad\nlabel")


vybe (changes requested):
This PR requires the following change before merge:

- [ ] Update `tests/unit/test_1640_display_label_at_creation.py::test_label_update_shares_policy` to the new contract. Line 95 asserts `AgentLabelUpdate().label is None` — the empty-body-means-clear behavior this PR deliberately outlaws — so the `regression diff` job fails deterministically across all 3 seeds (`4978 pass / 1 fail` on head vs `4972 / 0` on base).

The fix itself is right (the #1821 silent-clear on an unknown field name is a real bug and `extra="forbid"` + required-but-nullable is the correct shape), but it changes a contract that an existing test pins, and only the new test file was updated. Merging as-is would land a permanently red test on dev and pollute every future PR's regression-diff base.

Please update the old test to the new contract (empty body → `ValidationError`) and request re-review.

dolho:
Fixed in 096eee38 — thanks, that was a clean catch.

`test_1640_display_label_at_creation::test_label_update_shares_policy` line 95 (`AgentLabelUpdate().label is None`) is gone. It's replaced by the case it was actually standing in for — `label=None`, the explicit clear, which still works and is the behaviour that mattered — and the two inputs this PR now rejects moved into their own test: the empty body, and `display_label="Typo Name"`, i.e. the literal #1821 trigger, which had no coverage in that file before.

```
pytest tests/unit/test_1640_display_label_at_creation.py tests/unit/test_1821_label_update_strictness.py
23 passed
```

Full unit suite is running; I'll post the count when it lands.

On the nightly bot's merge-conflict note above: stale. `origin/dev` is an ancestor of this branch as of now, the API reports `mergeable: true`, and no merge was needed.

Re-review please.

dolho:
Full unit suite on the branch after 096eee38:

```
4909 passed, 16 skipped, 4 failed  (15:47)

```

`test_1640_display_label_at_creation` is green — the blocker you flagged is gone. The 4 remaining are pre-existing and environmental (`ModuleNotFoundError: No module named 'tzdata'` for the first; the other three reproduce on an untouched checkout at the same base), and they're the same set showing on #1837 and #1838.

dolho:
Review item was already fixed in `096eee38` (pushed the day after the review); the branch has now also been rebased onto current `dev`, so the tip is `2435e58a`.

**The item.** `test_1640_display_label_at_creation::test_label_update_shares_policy:95` asserted `AgentLabelUpdate().label is None` — exactly the empty-body-means-clear behaviour this PR outlaws. Replaced with the explicit-clear case it was standing in for (`label=None`), and the two now-rejected inputs split into their own test:

```python
def test_label_update_requires_the_label_field(M):
    with pytest.raises(Exception):
        M.AgentLabelUpdate()                              # empty body
    with pytest.raises(Exception):
        M.AgentLabelUpdate(display_label="Typo Name")     # the actual #1821 trigger
```

Worth noting the second case had no coverage in that file at all — the misnamed-field trigger was only tested in the new file, so the contract's own test module didn't pin the bug it exists to describe.

**Checked whether one site was all of them**, rather than just doing the named fix. Every `AgentLabelUpdate` reference in the repo: the model (`models.py:158`), the router (`routers/agents.py:920`), and the two test files — both now on the new contract. No other consumer pins the old default.

**Frontend compatibility re-verified rather than re-asserted.** The PR body claims the UI always sends `{ label }`, but the claim has a sharp edge worth being explicit about: `JSON.stringify({label: undefined})` produces `{}`, which under this PR now 422s — so "always sends `{ label }`" would be *false* if the clear path passed `undefined`. It doesn't. `AgentHeader.vue:725` emits `trimmed || null`, so a clear serializes to `{"label":null}` and still works. Traced the whole path (`AgentHeader` → `AgentDetail:629` → `stores/agents.js:393`).

**Verification after the rebase** (47 commits of `dev` had landed since the green run, so the old CI pass was stale): `test_1640` + `test_1821` = 23 passed; wider label/models/agent-config selection = 91 passed. CI re-running on the rebased tip.

Re-requesting review.

obasilakis (approved):
Re-validated after the requested change landed.

vybe's single blocking item is addressed by 2435e58a: `test_1640_display_label_at_creation.py::test_label_update_shares_policy` line 95 no longer asserts `AgentLabelUpdate().label is None` — it now asserts the explicit-clear case (`label=None`) it was standing in for, and the two newly-rejected inputs (empty body, the misnamed `display_label` field) moved into their own named test. The `regression diff` job is green across all 3 seeds: base 5735 pass / 1 fail vs head 5743 / 1 — the same single pre-existing failure on both sides (`test_1069_voip_call_path_param`, a FastAPI `get_flat_dependant` ImportError already red on dev), so **no new failures** and +8 net passing tests. All 18 checks green.

Contract change verified safe. Every caller sends an explicit `{label: <string|null>}`: the only writer is `stores/agents.js:391` `setAgentLabel`, reached solely from `AgentDetail.vue:629` via `AgentHeader.vue:725` `emit('set-label', trimmed || null)` — the clear path coerces `''` to an explicit `null`, never `undefined`, so it never serializes to `{}`. No MCP tool, CLI command, script, agent-server path, or enterprise-submodule caller touches the endpoint, and nothing constructs `AgentLabelUpdate` programmatically outside tests. Docs already prescribed the strict form — `requirements/core-agent.md:60` specifies `{label: string|null}` and `managing-agents.md:65` documents clearing as `{"label": null}` or a blank string — so this brings the code into conformance rather than changing what was documented. Both fixes match #1821's own suggested remedies.

Approving; not dismissing the prior review.

dolho:
@vybe the blocking item is addressed in `2435e58a`: `test_1640_display_label_at_creation.py::test_label_update_requires_the_label_field` now pins the new contract — `M.AgentLabelUpdate()` (empty body) and `M.AgentLabelUpdate(display_label="Typo Name")` (the actual #1821 trigger) both `raises(Exception)`, and the empty-body-means-clear assertion is gone. `regression diff` is green across all seeds and every check passes. Ready for re-review.

dolho:
Rebased onto `dev` (import conflict in `models.py` — dev added `Any`/`Union` to the `typing` line while this branch added `ConfigDict` to the `pydantic` line; resolved as the union of both).

On the requested change: it was already addressed in `2435e58a` ('test(label): update the #1640 policy test to the #1821 contract'), pushed after the review — the review just was never re-requested, so the PR has been reading as blocked since. Concretely, `test_label_update_shares_policy` no longer asserts the outlawed empty-body behaviour:

```python
def test_label_update_shares_policy(M):
    assert M.AgentLabelUpdate(label="  Hi  ").label == "Hi"
    assert M.AgentLabelUpdate(label="").label is None       # clear
    assert M.AgentLabelUpdate(label=None).label is None     # clear, said explicitly
```

`label=None` is an *explicit* clear, which the new contract still allows; the omitted/misnamed case the PR outlaws is pinned separately by `test_label_update_requires_the_label_field`.

Verified after the rebase: 23 passed across both label suites, 97 passed across everything matching `label or 1640 or 1821 or models`. CI should now report — it had no `refs/pull/1822/merge` to build while the PR was conflicting.

@vybe ready for re-review.

vybe (approved):
Approving — my single blocking item is resolved, verified in the diff rather than taken from the PR body.

`test_1640_display_label_at_creation.py::test_label_update_shares_policy` line 95 no longer asserts `AgentLabelUpdate().label is None`. It now asserts `AgentLabelUpdate(label=None).label is None` — the explicit-clear case the old line was really standing in for — and the two inputs this PR newly rejects (empty body, and the `display_label` typo that is the actual #1821 trigger) moved into their own named `test_label_update_requires_the_label_field`. That is the right split: the old assertion conflated "clear" with "unparsed", which is the bug.

The fix shape is correct. `extra="forbid"` plus required-but-nullable removes the ambiguity at its source rather than special-casing the one field name that happened to collide, and the normalizer guard is retained — `test_label_is_still_normalized` and `test_control_characters_are_still_rejected` pin that the stricter model config does not bypass the ent#181 validator, which is the obvious way this change could have quietly regressed.

The `PUT {}` → 422 behaviour change is real and correctly called out. It is safe: the only writer is `stores/agents.js` `setAgentLabel`, reached from `AgentHeader.vue` via `emit('set-label', trimmed || null)`, which coerces `''` to an explicit `null` and never serializes to `{}`. `requirements/core-agent.md` already specified `{label: string|null}`, so this brings the code into conformance with what was documented rather than changing the contract.

Verified locally at `644d7f88`: **23 passed** across both files. All required checks green.

I patched the body from `Related to #1821` to `Fixes #1821` before merging — without a closing keyword `.github/workflows/issue-status-on-merge.yml` skips the promotion and #1821 strands in `status-in-progress`. Same-repo, so it will auto-promote now.

Also noting the honesty in the write-up: correcting the test to the normalizer's real behaviour (trim + NFC, no internal-whitespace collapse) rather than changing the code to match the assumption is the right instinct.

PR #1824, opened 2026-07-27
fix(agents): reject an invalid local: template and ship the missing local:default (#1759)

### LINKED ISSUE #1759: bug: absent local: template silently creates a blank agent (no error)

AndriiPasternak31:
## Summary

**This PR was overtaken mid-flight and has been rebased around the winner.** #1759 and #1793 are the same defect filed twice; #1793 was fixed by @dolho in **#1803** and merged to `dev` hours before this PR opened. Rather than re-litigate, this branch **concedes dev's contract wholesale** and now carries only what #1803 did *not* fix.

| absent `local:` template | |
|---|---|
| **404 `UNKNOWN_LOCAL_TEMPLATE`** | dev (#1793) — **kept verbatim**, this PR no longer touches it |

The full history is in the collision comment below; the short version is that the 400-vs-404 disagreement was resolved in dev's favour and this PR dropped its own not-found gate entirely.

---

## What this PR actually delivers

Four things, all verified absent from `dev`:

**1. `LOCAL_TEMPLATE_INVALID` — the other half of the bug.** #1803 closed the *absent* case. A **present but malformed** `template.yaml` still hit the bare `except Exception: logger.warning` (`crud.py:723-724` on dev) and produced the **identical** observable outcome — blank agent, HTTP 200 — through a different line. Now 400 `LOCAL_TEMPLATE_INVALID` for an unreadable / unparseable / non-mapping document, in the same pre-side-effect band. The broad `except` survives, narrowed to *field-level* degradation, and its warning now names the template and the agent (it previously named neither).

**2. ★ `config/agent-templates/default/` — fixing a live regression #1803 shipped.** There is no `default/` template in the tree, but `local:default` is referenced **89×/12 files**, including **46× in `tests/test_systems.py`** (19 of them `dry_run: False` deploys) and by `export_manifest`. #1803's new 404 broke every one of them.

**Nothing catches it.** `backend-unit-test.yml:116` collects `tests/unit/` only; the root-level suites are live-stack files no workflow runs; and the one *unit* test using `local:default` (`test_ent125_resilient_system_deploy.py:68,84`) mocks `create_agent` entirely. Two tests are red on `dev` right now — the 404 preempts `normalize_cpu`'s 400 (`crud.py:1841` vs `:910-917`):

- `test_systems.py:1139` `test_strict_deploy_aborts_with_original_status` — asserts 400, gets 404
- `test_systems.py:1075` `test_partial_deploy_continues_and_reports` — flips `partial` → total failure (its *good* agent dies too)

Shipping `default/` turns both green. `test_1759_template_root_parity.py::test_every_shipped_template_resolves_through_the_create_gate` is the guard that would have caught this.

**3. Source-run root fallback.** dev's `_LOCAL_TEMPLATE_ROOTS` is two hard-coded container-absolute paths with no fallback, so **outside a container the gate 404s every `local:` create** — dev shells, source-run CI, unit tests. #1803 could only paper over that by pointing the roots at a tmp `scout` fixture inside the #1484 harness. `_curated_templates_root()` fixes it properly, at **both** seams that read the roots.

**4. Two latent bugs of the same class, one seam over.**
- `os.getenv("HOST_TEMPLATES_PATH", default)` → `os.getenv(...) or ...`: an **empty** value made `Path("") / name` collapse to a bare name, which Docker reads as a **named volume** — an empty one mounted at `/template`. Same silent-blank-template bug.
- `export_manifest`'s `agent.get('template', 'local:business-assistant')` was unreachable dead code: Blank Agents carry `"template": None` (key *present*), and `.get` returns its default only when the key is **absent**. So every Blank Agent exported `template: null`, which `SystemAgentConfig.template` (non-`Optional`) rejects on redeploy. Fixed with `or`. Pre-existing broken round-trip, not a regression.

---

## ★ The constraint that shaped the resolution — do not flatten the `if/else`

#1803's own commit `edd854fb2` reverted a guard-clause refactor for this reason:

> CodeQL flagged two new py/path-injection alerts … They surfaced only because the guard-clause refactor **dedented the whole load block**, moving those two expressions onto new lines and **re-fingerprinting alerts that were already dismissed on dev**.

Confirmed via the API: `py/path-injection` **#238 / #239 / #240** are `dismissed: false positive` on `crud.py`. This branch had made exactly that dedent — it was the mechanical cause of the merge conflict.

The resolution therefore keeps dev's `if template_yaml.exists(): … else: raise` shape and adds the new bands **inside** it. `if template_yaml.exists():` and `with open(template_yaml) as f:` are **byte-identical to dev**, at the same indentation. Verified: the deletion set of `git diff origin/dev -- crud.py` contains neither line, nor any of the `_safe_local_template_path` / `template_yaml` expressions. A comment in the code and a note in `template-processing.md` now say so, so the next person doesn't re-derive it a third time.

---

## Corrected: dev's error message states something false

The message live on `dev` tells the caller:

> `templates marked hidden are not deployable`

That is **not true**. `hidden` is filtered only in the **listing** (`template_service.py:287`); `_resolve_local_template` never reads the flag, and `template_service.py:272-278` documents hidden templates as explicitly creatable by id. This PR ships `config/agent-templates/default/` as `hidden: true`, so the message actively misinforms about the very template being added.

Message text corrected. **Status code and error code unchanged** — `404` / `UNKNOWN_LOCAL_TEMPLATE`.

---

## Also fixed: 11 characterization tests red on every Mac

Pre-existing on `dev`, **not** caused by this PR, one line. `test_1484`'s tmp template root came from an unresolved `tempfile.mkdtemp()`; `/var` is a symlink to `/private/var` on macOS, so `_safe_local_template_path`'s `(root / name).resolve()` yields `/private/var/…` and `is_relative_to(root)` fails — **11 of 40** characterization cases 400 `INVALID_LOCAL_TEMPLATE_NAME`. Linux CI has no such symlink, which is why #1803 shipped green.

Fixed with `.resolve()` because these tests characterize the exact function this PR edits — without it there is no trustworthy local signal on the change. Verified identical 11 failures on pristine `origin/dev` before touching anything.

---

## Design notes a reviewer will otherwise re-litigate

**`config/agent-templates/default/` deliberately declares no `resources:`.** The resolver applies a template's `resources` **over** the caller's request, before `normalize_cpu`. ent#125's resilient-deploy test uses `resources: {cpu: "3"}` on a `local:default` agent as its deliberate pre-side-effect failure vector; a `resources:` block would neutralise it and flip that test from `partial` to `deployed`. Same reasoning for every other field the resolver mutates from.

**`_repo_local_templates_dir()`'s `parents[4]` is now guarded, not commented.** In the container layout `/app/services/agent_service/crud.py` has only parents 0–3, and this is evaluated at **import** time — an `IndexError` would stop the backend booting, not degrade one request. The branch introduced this helper (dev's roots are two literals), so shipping a newly-added crash-on-import path as a comment was not defensible.

**`config/agent-templates` is bind-mounted, never `COPY`'d** (`docker-compose.yml:263`, `docker-compose.prod.yml:269`; the Dockerfile copies `process-docs`/`process-templates`/`manifests` but not this). The new template reaches a deployment **only via the host checkout** — an image-only upgrade on a stale checkout still 404s `local:default`, because Docker auto-creates the missing bind source as an *empty* directory. **Operators must update the checkout, not just the image.** Belongs in the release notes for whichever cut carries this.

**The web UI cannot trigger either reject.** `CreateAgentModal.vue` degrades an unknown template to `''`, and `template=""`/`None` never enter the `local:` branch. API / MCP / manifest only.

**One identical message whichever root missed, echoing no path.** Deploy-local templates (#950) are named after **agent** names, so a root-distinguishing or path-echoing error would let a `creator`-role caller probe whether another user's deploy-local agent exists (#186). Pinned by `test_t1b` / `test_t1c` / `test_t6b`.

---

## Changes

| File | What |
|---|---|
| `services/agent_service/crud.py` | `LOCAL_TEMPLATE_INVALID` band inside dev's `if/else` · narrowed `except` · `parents[4]` IndexError guard · corrected 404 message text · repo-relative curated root at both seams · `os.getenv(...) or …` |
| `services/system_service.py` | `export_manifest` `or` fix + warning naming inferred agents |
| `config/agent-templates/default/*` + `README.md` | The missing template (hidden, no `resources:`) |
| `tests/unit/test_1759_*.py` (3) | Absent-case tests retargeted to 404 and kept for their **pre-side-effect proof** (no container/volume/MCP key/ownership row/slot) — which `test_1793_*` does not cover, since it calls the resolver directly |
| `tests/unit/test_ent125_*.py` | Hand-built fixture retargeted to 404 — it constructs its own `HTTPException`, so drift here passes **green** while asserting a dead contract |
| `tests/unit/test_1484_*.py` | The macOS `.resolve()` fix |
| `tests/registry.json` | The 3 new files registered |
| `docs/memory/*` | Contract flipped to 404 throughout; `learnings.md` attribution corrected + a process entry on parallel duplicate fixes |

**`tests/unit/test_1793_unknown_local_template.py` is deliberately untouched and still green** — that is the proof the concession is real.

---

## Verification

**Unit tier** (what CI runs): **`5033 passed, 17 skipped, 0 failed`**. Includes `test_1793` (untouched), `test_1484` (40/40), the three `test_1759`, `test_ent125`, `test_ent124`, `test_local_templates_listing`. The `test_1474` failure noted in the previous body is gone — dev fixed it meanwhile in `e9eb40da`.

**CI: 20 pass, 3 skipping** (e2e/report/auto-merge — expected for a non-UI draft). Notably `Analyze (python)` passes with **zero open `py/path-injection` alerts** on `crud.py` — the dismissed fingerprints survived, which is what the `if/else` preservation was for.

**`/verify-local`: PASS** — backend build + `import main` smoke inside the built image (the #1033 gate), boot, health.

**Root-level live-stack suites** — the real blast radius, which **no CI job runs**: `tests/test_systems.py` + `tests/test_subscriptions.py` + `tests/test_agent_lifecycle.py` → **`79 passed in 13:13`**.

**★ Before/after on the regression.** `config/agent-templates` is a bind mount and the absent-template path is byte-identical on both sides, so moving `default/` aside reproduces `dev`'s state exactly, isolating the one variable:

| `test_partial_deploy_continues_and_reports` + `test_strict_deploy_aborts_with_original_status` | |
|---|---|
| **without** `config/agent-templates/default/` (i.e. `dev` today) | **2 failed** |
| **with** it (this PR) | **2 passed** |

**Live probes** against a real booted stack:

```
local:nope-1759     -> 404 UNKNOWN_LOCAL_TEMPLATE   (no filesystem path in the body)
local:badyaml-1759  -> 400 LOCAL_TEMPLATE_INVALID   (unparseable)
local:emptyyaml     -> 400 LOCAL_TEMPLATE_INVALID   (empty file -> safe_load returns None)
local:default       -> 200, /home/developer/CLAUDE.md == "# Default Agent"
```

**Merge state**: `git merge-tree origin/dev HEAD` is clean.

### Not verified by execution — stated plainly

- **The agent runtime.** `/verify-local`'s base-image stage failed on `curl https://deb.nodesource.com/setup_20.x` returning **HTTP 403** — an external service, not code. This diff touches **zero** files under `docker/`, so it was re-run with `--skip-agent` rather than treating a third-party 403 as a blocker. (Independent of this PR: that Dockerfile step is an unpinned external installer and can break any build at any time.)
- `/verify-local`'s `integration: OK (4s)` is **not** coverage — that stage environment-skips most of the suite. Please don't read it as such.
- The deploy-local root (`/data/deployed-templates`) is exercised by the probes above and by unit tests, but not against a real deploy-local *agent*.
- Linux / native dockerd (all runs macOS / Docker Desktop) and PostgreSQL.

Fixes #1759


AndriiPasternak31:
## ⚠️ Collision found at ship time — a competing fix for this same defect is already on `dev`

This PR opens `CONFLICTING`. The conflict is not incidental churn: **the same bug was filed twice and fixed twice, in parallel.**

- **#1793** — *"Nonexistent `local:` template silently creates an empty agent (200) instead of failing"* — filed separately, fixed by @dolho in **PR #1803**, **merged to `dev` today** (`acf699c0`, 15:26 +0300). Issue is `status-in-dev`.
- **#1759** (this PR) — same defect, planned and built over the preceding day. Issue is still `status-in-progress`.

**This is a human decision, not a merge. Nothing has been resolved here.**

### Conflicting files (from a read-only `git merge-tree` against `origin/dev@4d743fdb`)

| File | Cause |
|---|---|
| `src/backend/services/agent_service/crud.py` | Two different `else:` branches added at the same site in `_resolve_local_template` |
| `docs/memory/learnings.md` | Ordinary append-point collision (two unrelated `dev` entries) — trivial |

`dev` also advanced **33 commits** since this branch's base `f7aff466`. This PR was **deliberately not rebased**: the `/verify-local` run it rests on was performed against `f7aff466`, and rebasing would invalidate that evidence without re-running it.

### The two fixes are not interchangeable

`#1793` fixes the reported symptom. This PR fixes the symptom **plus five adjacent holes** the plan enumerated. Most consequentially, the two ship **different error contracts**:

| | #1793 — merged to `dev` | #1759 — this PR |
|---|---|---|
| Contract | **404 `UNKNOWN_LOCAL_TEMPLATE`** | **400 `LOCAL_TEMPLATE_NOT_FOUND`** |
| Absent template rejected | ✅ | ✅ |
| Empty / non-dict / unparseable `template.yaml` | ❌ still swallowed by `except Exception: logger.warning` → same blank agent, same 200 | ✅ 400 `LOCAL_TEMPLATE_INVALID`; bare `except` narrowed to `(OSError, yaml.YAMLError)` |
| Gate live in source-run backends / dev shells | ❌ no repo-relative root fallback. Instead `_load_crud` in `test_1484` was pointed at a tmp `scout` fixture — so the gate is satisfied by test scaffolding, not by a root that exists in the environments people actually run | ✅ repo-relative `parents[4]` fallback, applied to **both** seams |
| `/template` bind seam | ❌ untouched — create can load correct metadata while starting with no `/template` bind | ✅ widened, plus `os.getenv("HOST_TEMPLATES_PATH") or …` (an **empty** value made `Path("") / name` an empty **named volume** — the identical silent-blank-template bug, one seam over) |
| `config/agent-templates/default/` | ❌ not shipped — so the `local:default` id used by ~50 test refs and ~35 doc lines still resolves to nothing | ✅ shipped (hidden, deliberately no `resources:`) |
| `export_manifest` `template: null` round-trip | ❌ untouched | ✅ fixed (`or`, not a `.get` default) |
| ent#125 `failed[]` behaviour asserted | ❌ | ✅ `status_code == 400`, reason text, `status == "partial"` |
| Docs | none | `requirements/core-agent.md`, 2 feature flows + index, `learnings.md`, onboarding |
| Tests | 130 lines (`test_1793_*`) | 39 tests across 3 files, plus the ent#125 extension |

### One substantive defect in the merged text, worth a look regardless of which fix wins

The message now live on `dev` tells the caller:

> `templates marked hidden are not deployable`

**That is not true, and the codebase says so at `template_service.py:244-249`:** `hidden` is surfaced *and not filtered* by `get_local_template()` specifically so *"a single-id resolve — `get_local_template()` / **creation-by-id** — still works"*. Only the **listing** (`get_local_templates`) excludes it. Hidden templates are fully deployable by id — this PR's own `config/agent-templates/default/` is `hidden: true` and deploys, which is the whole basis of its D2 decision.

So the merged error message misinforms the caller on the one question they just got it wrong about.

### Options, for the record — pick one

1. **Supersede.** Rebase this branch onto `dev`, drop the now-redundant absent-template `else`, and keep the delta: `LOCAL_TEMPLATE_INVALID`, the root fallback, the bind seam, `default/`, the exporter fix, docs, and the corrected message. Re-run `/verify-local` (the current evidence does not survive a rebase). Retitle toward "harden".
2. **Replace.** Revert `acf699c0` on `dev` and land this PR whole. Cleanest single contract, but reverts a colleague's merged, reviewed work.
3. **Close this PR**, keep #1793's fix, and file the residue (`LOCAL_TEMPLATE_INVALID`, root fallback, bind seam, `default/`, exporter) as follow-ups.

Option 1 looks right, but the **404-vs-400 contract** is a call for a human — as is deciding whether #1759 and #1793 should be de-duplicated on the tracker.

*Filed as a PR comment, not as an issue — agents do not open issues on this repo.*

AndriiPasternak31:
## Resolved — option 1 taken, dev's 404 conceded

Following up on the collision comment above. The decision was **concede `dev`'s contract**: this branch dropped its own not-found gate entirely and is now purely additive on top of #1803. `tests/unit/test_1793_unknown_local_template.py` is **untouched and still green** — that is the proof the concession is real, not a re-litigation.

Merged `origin/dev` (not rebased — 11 commits replayed through a *semantic* conflict is strictly riskier, and the PR squash-merges anyway). **`mergeable: MERGEABLE`.**

### The constraint that dictated the shape of the resolution

@dolho's commit `edd854fb2` on #1803 says it plainly:

> CodeQL flagged two new py/path-injection alerts … They surfaced only because the guard-clause refactor **dedented the whole load block**, moving those two expressions onto new lines and **re-fingerprinting alerts that were already dismissed on dev**.

This branch had made exactly that dedent — which is *why* it conflicted. So the resolution keeps `dev`'s `if template_yaml.exists(): … else: raise` verbatim and layers the new band **inside** it. `if template_yaml.exists():` and `with open(template_yaml) as f:` are byte-identical to dev at the same indentation; I verified by reading the whole deletion set of `git diff origin/dev -- crud.py` rather than trusting the intent.

**It held.** `Analyze (python)` passes and there are **zero open `py/path-injection` alerts** on `crud.py` — the dismissals (#238/#239/#240) kept their fingerprints. Both the code and `template-processing.md` now carry a note saying not to flatten it, so this isn't rediscovered a third time.

### Two things found while resolving that are worth a separate look

**1. The error message on `dev` states something false.** It tells the caller `templates marked hidden are not deployable`. `hidden` is filtered only in the **listing** (`template_service.py:287`); `_resolve_local_template` never reads the flag, and `template_service.py:272-278` documents hidden templates as explicitly creatable by id. This PR ships a `hidden: true` template, so the message misinforms about the exact id being added. Corrected here — **status and error code unchanged**.

**2. 11 of 40 `test_1484` characterization cases are red on every Mac, on `dev` today.** #1803's fixture passes an unresolved `tempfile.mkdtemp()` as the template root; `/var` is a symlink to `/private/var` on macOS, so `_safe_local_template_path`'s `(root / name).resolve()` yields `/private/var/…` and `is_relative_to(root)` rejects it — every `local:` case 400s `INVALID_LOCAL_TEMPLATE_NAME`. Linux CI has no such symlink, which is why it shipped green. Confirmed identical on pristine `origin/dev` before touching anything. One-line `.resolve()` fix included, because those tests characterize the exact function this PR edits — without it there's no trustworthy local signal.

### Verification

**CI: 20 pass, 3 skipping** (e2e/report/auto-merge — expected for a non-UI draft). All 6 pytest seed jobs green.

**Local unit tier: `5033 passed, 17 skipped, 0 failed`.** The `test_1474` failure quoted in the old PR body is gone — `dev` fixed it meanwhile in `e9eb40da`.

**`/verify-local`: PASS** — backend build + `import main` smoke inside the built image (the #1033 gate), boot, health, integration.

**Live probes** against a real booted stack:

```
local:nope-1759     -> 404 UNKNOWN_LOCAL_TEMPLATE   (no filesystem path in the body)
local:badyaml-1759  -> 400 LOCAL_TEMPLATE_INVALID   (unparseable)
local:emptyyaml     -> 400 LOCAL_TEMPLATE_INVALID   (empty file -> safe_load returns None)
local:default       -> 200, /home/developer/CLAUDE.md == "# Default Agent"
```

The two `LOCAL_TEMPLATE_INVALID` cases are the half `dev` does not fix — there they still create a blank agent at HTTP 200.

**Root-level suites — the blast radius no CI job runs.** `test_systems.py` + `test_subscriptions.py` + `test_agent_lifecycle.py` against the live stack: **`79 passed in 13:13`**.

**★ Before/after on the regression.** `config/agent-templates` is a bind mount and the absent-template path is byte-identical on both sides, so moving `default/` aside reproduces `dev`'s state exactly, isolating the one variable:

| `test_partial_deploy_continues_and_reports` + `test_strict_deploy_aborts_with_original_status` | |
|---|---|
| **without** `config/agent-templates/default/` (i.e. `dev` today) | **2 failed** |
| **with** it (this PR) | **2 passed** |

Mechanism: the 404 fires at template resolution (`crud.py` `_resolve_template`), which is **earlier** than `normalize_cpu`'s 400 (`_stage_config_files`). So `local:default` failing to resolve preempts the deliberate `cpu: "3"` failure vector — `test_strict_deploy` gets 404 where it asserts 400, and `test_partial_deploy` flips from `partial` to a total failure because its *good* agent dies too. This is what makes it a regression rather than a cosmetic gap, and it is invisible to CI: `backend-unit-test.yml:116` collects `tests/unit/` only, and the one *unit* test touching `local:default` mocks `create_agent` outright.

### Stated plainly — what is NOT verified

- **The agent runtime.** `/verify-local`'s base-image stage failed on `curl https://deb.nodesource.com/setup_20.x` returning **HTTP 403** — an external service, not code. This diff touches **zero** files under `docker/`, so I re-ran with `--skip-agent` rather than treat a third-party 403 as a blocker. Independent of this PR: that Dockerfile step is an unpinned external installer and can break any build at any time.
- `/verify-local`'s `integration: OK (4s)` is **not** coverage — that stage environment-skips most of the suite. Please don't read it as such.
- The deploy-local root (`/data/deployed-templates`) is exercised by the probes above and by unit tests, but not against a real deploy-local *agent*.
- Linux / native dockerd, and PostgreSQL.

### Note for whoever cuts the release

`config/agent-templates` is **bind-mounted, never `COPY`'d** into the backend image (`docker-compose.yml:263`, `docker-compose.prod.yml:269`). So `config/agent-templates/default/` reaches a deployment **only via the host checkout** — an image-only upgrade on a stale checkout still 404s `local:default`, because Docker auto-creates the missing bind source as an *empty* directory and the mount silently succeeds. **Operators must update the checkout, not just the image.**

AndriiPasternak31:
Ready for review. CI is green (20 pass / 3 skipping) and `mergeable`.

**@dolho** — this PR sits directly on top of your #1803, so you're the right first reviewer. Three things specifically for you:

1. **I conceded your contract.** The absent-template case stays **404 `UNKNOWN_LOCAL_TEMPLATE`**, unchanged. This branch had built a competing 400 and dropped it entirely. `tests/unit/test_1793_unknown_local_template.py` is untouched and still green.

2. **I changed your error message text** (status + code unchanged). It said `templates marked hidden are not deployable`, but `hidden` is filtered only in the listing (`template_service.py:287`) — `_resolve_local_template` never reads the flag, and `template_service.py:272-278` documents hidden templates as creatable by id. This PR ships a `hidden: true` template, so the message would have misinformed about the exact id being added. Happy to revert the wording if you disagree with the reading.

3. **I fixed your `test_1484` fixture** — one line. `tempfile.mkdtemp()` is unresolved, and `/var` → `/private/var` on macOS, so `is_relative_to(root)` rejects every `local:` case: **11 of 40 characterization tests are red on any Mac**, on `dev` right now. Linux CI has no such symlink, which is why it went green. Confirmed identical on pristine `origin/dev` before touching it.

Your `edd854fb2` note about the guard-clause dedent re-fingerprinting the dismissed `py/path-injection` alerts was the single most useful thing in this whole resolution — this branch had made exactly that dedent, which is *why* it conflicted. The fix keeps your `if/else` byte-identical and layers the new band inside it. `Analyze (python)` passes with zero open path-injection alerts on `crud.py`.

**@vybe** — the parts worth your eye are the two judgement calls, both explained in the comment above:
- **Scope**: this PR now carries only what #1803 didn't (the `LOCAL_TEMPLATE_INVALID` band, the source-run root fallback, the bind-seam fix, `export_manifest`) plus `config/agent-templates/default/`, which fixes a **live regression on `dev`** — proven before/after, 2 failed → 2 passed.
- **Release ordering**: `config/agent-templates` is bind-mounted, never `COPY`'d, so an image-only upgrade on a stale checkout still 404s `local:default`. That belongs in the release notes for whichever cut carries this.

Not verified, stated plainly: the **agent runtime**. `/verify-local`'s base-image stage failed on `deb.nodesource.com` returning HTTP 403 — external service, not code, and this diff touches zero files under `docker/`. Re-run with `--skip-agent`.

Also flagging separately, not this PR's to fix: that Dockerfile step is an **unpinned external installer** (`curl … | bash`), so any base-image build can break at any time on someone else's outage or rate limit.

vybe:
Maintainer push 4d86139: reverted the `.claude` submodule pointer back to `de7fc716` (trinity-dev main tip). The branch had bumped it to `3dbde4bf`, which only exists on an unmerged trinity-dev branch — pinning the public submodule there breaks `git submodule update` for every clone if that branch rebases/deletes. No other change; validation of the fix itself passed (/validate-pr), will merge once checks re-run green.

vybe (approved):
Validated via /validate-pr: pre-side-effect 400/404 gates with disclosure discipline, curated-root fallback so the gate is live outside containers, local:default template ships, export_manifest round-trip fixed, extensive named regression tests (t1–t10 + root parity + export). Stray .claude submodule bump reverted in 4d86139 (pointed at an unmerged trinity-dev commit). All checks green post-fix. Approving.

Issue #1771, opened 2026-07-26
Run edge-case analysis & property testing (P-38) on the highest-risk core areas


vybe:
## Summary

Apply the new edge-case analysis & property-testing process (core-team `/edge-cases` skill, catalog candidate **P-38**) to the highest-risk areas of the codebase: enumerate edge cases per target (boundary-value analysis + the incident-traced bug-class checklist), generate parametrized pytest + Hypothesis property tests for the uncovered ones, and verify honestly — branch coverage everywhere, a scoped mutation-testing gate (`--mutate`) on the destructive guards.

## Context

Target selection is risk-based (churn × complexity ∩ git fix-density ∩ property-testing suitability ∩ blast radius), and the shortlist mirrors real incident history: destructive retention prunes (#1638, #1644), naive/aware + ISO-format timestamp bugs (#1474, #476), NaN cost-accumulator poisoning (#1155), CAS status races (#1082), replay side-effects (#1083). Method lineage: Anthropic's agentic property-based testing workflow (understand → propose properties → generate → reflect → report).

## Targets (ranked)

1. `src/backend/services/cleanup_service.py` prune predicates + `src/backend/services/retention_guard.py` — **run with `--mutate`**; a surviving mutant here is a data-loss bug (#1638/#1644 class)
2. Timestamp helpers: `src/backend/utils/helpers.py` (`iso_cutoff`, `parse_iso_timestamp`) + `src/scheduler/utils.py` — round-trip + format-mixing properties (#1474, #476, architecture Invariant #16)
3. `src/backend/db/schedules/` — CAS status writers (precondition-guarded terminal transitions) + analytics aggregation (NULL-skip averages, UTC bucketing)
4. `src/backend/services/loop_service.py` — budget/deadline/no-progress arithmetic with hostile-float strategies (#1155–#1157)
5. `src/backend/services/credential_paths.py` — allow/deny policy; bonus property: byte-identical parity with the vendored agent-server copy on every input
6. `src/backend/services/skill_packaging.py` — frontmatter parse + archive member vetting (agent-authored input trust boundary)
7. `src/backend/services/task_execution_service.py` — `apply_result` envelope normalization + status mapping (pure parts only; replay/CAS semantics)

## Acceptance Criteria

- [ ] Edge-case matrix produced per target (enumerated / already covered / newly covered / unspecified)
[...]

PR #1825, opened 2026-07-27
test(retention): edge-case + property coverage for the retention blast-radius guard (#1771)

AndriiPasternak31:
**Part of #1771 (target 1 of 7).** Refs abilityai/trinity#1771 — deliberately **not** a
closing keyword: targets 2–7 remain outstanding and other slices reference the same issue.

Test-only edge-case + property coverage (P-38 `/edge-cases`) for the **destructive retention
prunes** — the #1638/#1644 data-loss class, where a wrong window silently hard-DELETEs an
un-configured install's history and ships green because a test asserted the destructive default.

**Targets:** `src/backend/services/retention_guard.py` (all of it) ·
`src/backend/db/schedules/retention.py` (prune predicates, bounded counts, the three drain
accessors) · exactly 4 functions of `cleanup_service.py` (`_read_retention_settings`,
`_log_prune`, `_guard_allows`, `_after_guarded_prune`).

**AC#4 holds — zero product-code change.** `git diff origin/dev...HEAD --stat -- src/` is
**empty** — re-verified against current `dev` after the §4 dependency harmonization.
No real product bug was found, and none was manufactured. Two defects that *are* real
(a false docstring and an observability gap) are **pinned as observed behaviour or enumerated**,
never silently fixed — fixing them is product code, which AC#4 forbids.

---

## 1. AC#1 deliverable — completed edge-case matrix

Legend: **C** = already covered on `dev` · **N** = newly covered here · **U** = UNSPECIFIED
(reported, deliberately not tested). Files: `E` = `test_1771a_retention_edges.py`,
`P` = `test_1771a_retention_properties.py`.

### A — `evaluate()` decision logic

| # | Case | Status | Covered by |
|---|---|---|---|
| A2 | `floor=0` does not default to 1000 | **N** (was partial C) | `E::TestEvaluateBoundaries::test_threshold_boundary[A2-*]`, `P::test_decision_table_matches_an_independent_oracle` `@example(floor=0)` |
| A3 | floor=100 with exactly 100 candidates | **N** | `E::...[A3-schedules-at-floor]`, `P` `@example(available=100, floor=100)` |
| A4 | one over threshold, no ack | **N** | `E::...[A4-*]`, `P` `@example(MAX+1, acked=False)` |
| A5 | candidates == threshold exactly | **N** | `E::...[A5-rows-at-threshold]`, `P` `@example(MAX)` |
| A6 | threshold−1 / threshold+1 | **N** | `E::...[A6-*]` |
| A7 | 0 candidates at every floor | **N** (was C for one floor) | `E::test_zero_candidates_never_trips_any_floor` |
| A8 | `count_fn` raises → refuse, `-1`, `count_failed` | C + **N** | `test_1644::test_count_error_refuses`; now also `P::...oracle` (`count_raises=True`) |
| A9 | non-numeric `count_fn` return raises `TypeError` | **N** (observed, not endorsed) | `E::TestEvaluateRaisesContract::test_non_numeric_count_raises_out_of_evaluate` → **Finding 2** |
| A10 | NaN count, no ack → refuse | **N** | `E::test_nan_count_without_ack_refuses` |
| A10b | NaN count, ack present → **allowed** | **N** | `E::test_nan_count_with_ack_is_allowed` |
| A11 | negative count is allowed through | **N** | `E::test_negative_candidate_count_is_allowed_through` |
| A12 | ack lookup raises → refuse | C + **N** | `test_1644::test_ack_lookup_error_refuses`; now also `P::...oracle` (`ack_raises=True`) |
| A13 | acked while over threshold | C + **N** | `test_1644::test_acknowledged_proceeds`; `P` `@example(acked=True)` |
| A14 | `count_fn` receives exactly `threshold + 1` | **N** | `E::test_count_fn_receives_exactly_threshold_plus_one`, `P::...oracle` (asserts `seen_limits`) |
| A15 | `evaluate` performs no direct DB write | **N** | `P::test_evaluate_never_writes_to_the_database` |
| A16 | full decision table == independent oracle | **N** | `P::test_decision_table_matches_an_independent_oracle` (all 5 exits) |
| A17 | negative floor refuses regardless of data | **N** | `E::test_negative_floor_refuses_regardless_of_data`, `P` `@example(floor=-1)` |

### B — ack lifecycle

| # | Case | Status | Covered by |
|---|---|---|---|
| B1 | round-trip at arbitrary window | **N** (C only for w∈{5,30}) | `P::test_ack_round_trips_at_any_window` |
| B2/B3/B4 | wrong window / wrong key / single-use | C + **N** | `test_1644::TestAcknowledgement::*`; generalized in `P::test_an_ack_authorizes_exactly_one_{window,setting}` |
| B5 | consume with no ack present | **N** | `E::test_consume_without_an_ack_is_a_no_op` |
| B6 | stored value with whitespace | **N** | `E::test_stored_value_is_stripped_before_comparison` |
| B7 | `"030"`/`"+30"`/`"30.0"`/`""`/`"thirty"` | **N** | `E::test_non_canonical_stored_values_do_not_acknowledge` |
| B8 | `w=True` vs `w=1` | **U** | unreachable from the router (Pydantic coerces to int); fail-safe direction |
| B9 | window binding over arbitrary pairs | **N** | `P::test_an_ack_authorizes_exactly_one_window` |
| — | ack key lives under the blocklisted prefix | **N** | `E::test_ack_key_is_namespaced_under_the_blocklisted_prefix` |

### C — `announce_refusal` / `_last_refused` / `note_allowed` / `_alarm_id`

| # | Case | Status | Covered by |
|---|---|---|---|
| C1 | first refusal → ERROR + exactly one alarm | **N** | `E::test_first_refusal_logs_error_and_raises_exactly_one_alarm`, `P::test_alarm_count_equals_the_number_of_transitions` |
| C2 | repeat → INFO, no second alarm | **N** | `E::test_repeat_refusal_drops_to_info_and_raises_no_second_alarm`, `P` `@example` |
| C3 | same key, narrowed window → fresh | **N** | `E::test_a_narrowed_window_is_a_fresh_transition`, `P` `@example` |
| C4 | `note_allowed` re-arms ERROR | **N** | `E::test_note_allowed_rearms_the_error_level`, `P` `@example` |
| C5 | alarm raises → refusal unchanged | C + **N** | `test_1644::test_alarm_failure_never_flips_the_refusal`; `E::test_alarm_failure_does_not_propagate` |
| C6 | memo written BEFORE the alarm try → a failed alarm never retries | **U** | **Finding 3** — observability gap; the fix is product code, forbidden by AC#4 |
| C7 | settings read raises → `source == "unknown"` | **N** | `E::test_settings_read_failure_degrades_source_and_never_propagates` |
| C8 | **SECURITY** whole-item allowlist + no value echo | **N** | `E::test_alarm_payload_carries_counts_and_identifiers_only`, `P::test_alarm_payload_never_grows_new_fields` |
| C9 | `_alarm_id` natural key | **N** | `E::test_alarm_id_is_the_natural_key_and_expiry_is_null` |
| C10 | `expires_at is None` | **N** | same as C9, and `P::test_alarm_payload_never_grows_new_fields` |
| — | `note_allowed` clears only its own key | **N** | `E::test_note_allowed_only_clears_its_own_key` |
| — | alarm hosts on the uncreatable sentinel | **N** | `E::test_alarm_hosts_on_the_uncreatable_sentinel_agent` |

### D — `_read_retention_settings` (body executed nowhere on `dev`)

| # | Case | Status | Covered by |
|---|---|---|---|
| D1–D4, D6 | valid / invalid / None / negative / spaced | **N** | `E::TestRetentionSettingsCoercion::test_raw_value_coercion` (12 params) |
| D5 | unicode digits accepted by `int()` | **N** (observed) | `E::test_unicode_digits_are_accepted_by_int` |
| D7 | absurdly large window stays a positive int | **N** | `E::test_absurdly_large_window_stays_an_int_and_disables_nothing` |
| D8 | totality: 4 non-negative ints, never raises | **N** | `P::test_reader_is_total_and_never_negative` |
| — | the 4 windows are returned in a fixed ORDER | **N** | `E::test_returns_the_four_windows_in_a_fixed_order` |

### E — `_log_prune` level rule

| # | Case | Status | Covered by |
|---|---|---|---|
| E1–E3 | 0 / trickle / cap−1 / cap / cap+1 | **N** | `E::TestLogPruneLevels::test_level_escalates_at_the_chunk_boundary` (5 params) |

### F — predicate parity + drain semantics

| # | Case | Status | Covered by |
|---|---|---|---|
| F1/F2 | count == prune, fixed example | C | `test_1644::TestPredicateParity::*` |
| F3 | **property**: parity over arbitrary row sets | **N** | `P::test_row_count_equals_what_the_prune_deletes`, `P::test_log_count_equals_what_the_log_prune_nulls` |
| F4 | all four terminal states counted AND pruned jointly | **N** | `E::test_all_four_terminal_states_are_counted_and_pruned_together`, `P` `@example` |
| F5 | terminal row with `completed_at IS NULL` | **N** | `E::test_terminal_row_with_null_completed_at_is_excluded_by_both_sides`, `P` `@example` |
| F6 | `completed_at` exactly == cutoff (strict `<`) | **N** | `E::test_a_row_exactly_at_the_cutoff_survives` (frozen cutoff — see §5) |
| F7 | non-terminal never touched | C + **N** | `test_1644`; `P` `@example(_NON_TERMINAL)` |
| F8 | `execution_log IS NULL` excluded | C | `test_1644::test_execution_log_count_matches_prune` |
| F9 | `_bounded_count` == `min(candidates, limit)` | C + **N** | `test_1644::test_count_is_bounded_by_limit`; `P::test_bounded_count_returns_min_of_candidates_and_limit` |
| F10 | `retention_days <= 0` / `limit <= 0` → 0 | C + **N** | `test_1644`; `E::test_non_positive_chunk_size_prunes_nothing` |
| F11 | full drain at small `chunk_size` | **N** | `E::test_prune_execution_rows_drains_fully_with_a_small_chunk_size`, `P` `@example(chunk=1, 25 rows)` |
| F12 | naive vs `Z`-suffixed `completed_at` | **U** | boundary with slice **b** (`iso_cutoff` is b's target) — enumerated, not tested here |

### G — structural invariants

| # | Case | Status | Covered by |
|---|---|---|---|
| G1/G2 | threshold direction + not settings-backed | C | `test_1644::TestGuardIsNotItselfA1638::*` |
| G3 | every `RETENTION_OPS_KEYS` window has exactly one `_guard_allows` site | **N** | `E::test_every_retention_window_has_a_guard_call_site` (AST, strict set-equality both directions, **no hardcoded count**) |
| G4/G5 | sentinel uncreatable + windows ⊆ keys | C | `test_1644::TestCrossModuleInvariants::*` |
| G6 | **SECURITY** ack prefix blocklisted from generic `PUT /api/settings/{key}` | **N** | `E::test_ack_prefix_is_blocklisted_from_the_generic_settings_endpoint` |
| G7 | per-sweep floors pinned by DIRECTION | **N** | `E::test_per_sweep_floors_are_pinned_by_direction_not_by_value` |
| G8 | no floor exceeds the global ceiling | **N** | `E::test_no_per_sweep_floor_exceeds_the_global_ceiling` |

### H — `cleanup_service` guard adapters

| # | Case | Status | Covered by |
|---|---|---|---|
| H1 | allowed → `note_allowed`, True, no announce | **N** | `E::test_allowed_verdict_clears_the_memo_and_permits_the_prune` |
| H2 | refused → announce, False, no `note_allowed` | **N** | `E::test_refused_verdict_announces_and_blocks_the_prune` |
| H3 | `_after_guarded_prune` with no ack | **N** | `E::test_after_guarded_prune_is_a_no_op_without_an_ack` |
| H4 | consume raises → ERROR, never re-raised | **N** | `E::test_after_guarded_prune_never_raises_into_a_completed_sweep` |
| — | `floor` is forwarded to `evaluate` | **N** | `E::test_floor_is_forwarded_to_evaluate` |
| — | ack consumed when present | **N** | `E::test_after_guarded_prune_consumes_a_present_ack` |

### Totals

| | count |
|---|---|
| Enumerated | **62** |
| Already covered on `dev` (C, incl. C+N) | **18** |
| Newly covered by this slice (N) | **41** |
| UNSPECIFIED — reported, not tested (U) | **3** (B8, C6, F12) |

**146 tests** across the two files (71 example tests incl. parametrization + 12 Hypothesis
properties: 9 × 200 examples, 3 × 30 examples).

---

### `mutmut` is structurally unusable in this repo (reusable finding)

`mutmut==3.6.0` was installed and driven through four configurations. It generates mutants
correctly but **cannot resolve this repo's module paths**: it derives the module name from the
file path (`src/backend/services/retention_guard.py` → `backend.services.retention_guard`)
while `tests/unit/conftest.py` puts `src/backend` **directly** on `sys.path`, so the tests
import `services.retention_guard`. mutmut detects the mismatch itself and stops:

> `Stopping early, because tests recorded trampoline hits but none match any mutant key.`
> `Recorded keys: ['services.retention_guard.x_evaluate']`
> `Expected keys: ['backend.services.retention_guard.x_evaluate']`

This is structural, not a config error, and affects **any** future mutmut use here — flagged as
a follow-up, since P-38 mandates `--mutate` on subsequent targets.

The plan's sanctioned fallback (a hand-applied mutant list) was used, but made **complete**
rather than cherry-picked: a small AST mutator enumerates every mutation point with a standard
operator set (comparison swaps, boolean-operator flips, `not`-removal, numeric/string/boolean
constant mutation), applies exactly one per variant, and runs the scoped suite.

**Two harness controls run before every gate**, because a broken harness reports false kills:
(1) the pristine source must pass; (2) an **unmutated `ast.unparse` round-trip** must also pass
— **without this second control every mutant falsely reads as "killed"**, and the gate would be
meaningless. Both PASS for both targets. `__pycache__` is cleared around every apply/restore
(the same-byte-length hazard), and `git status` is clean after each run.

### Recorded run

| Target | Mutants | Killed | Survived | Coverage of run |
|---|---|---|---|---|
| `services/retention_guard.py` | 59 | **48** | 11 | complete (every mutant executed) |
| `db/schedules/retention.py` | 73 | **52** | 21 | complete (every mutant executed) |

Both improved by looping back to implementation: guard **41→48**, predicates **38→52**.
Thirteen killing tests were added as a direct result of the gate.

### Survivor dispositions — **two rows corrected since the recorded run**

The recorded survivor table was wrong in two rows. Both are corrected here; the recorded
`killed` counts above therefore **understate** the true kill count on this HEAD. I am not
restating a revised total, because the gate was not re-run end-to-end after the corrections —
the two rows were verified individually.

| Group | Count | Target | Disposition |
|---|---|---|---|
| Log/alarm **prose** string literals (message fragments, `logger` format strings, alarm `title`) | 11 | guard | **Accepted survivor.** Wording is not a behavioural contract. The log **level** (E1–E3), the alarm **count** (C1–C4), the routing **values** (`type`/`priority`/`alert_type`) and the payload **key set** (C8) are each asserted independently. Pinning prose breaks on any wording improvement and catches no bug. |
| `limit <= 0` / `chunk_size <= 0` → `< 0`, and the sibling `or`→`and` | — | predicates | **Equivalent mutants.** With `limit == 0` the early return is bypassed but the query becomes `LIMIT 0`, which returns no rows — and the chunked pruners then `break` on an empty id list. Observable behaviour is identical, so no test can distinguish them. |
| Default **argument values** (`chunk_size=500` ×3, `limit=5000` ×1) | 4 | predicates | **Accepted survivor.** Every production call site passes an explicit value (`RETENTION_CHUNK_SIZE_PER_CYCLE`), and the F3/F11 properties prove *any* positive chunk drains the full candidate set. Pinning `500` would be pinning a destructive default — **exactly the #1638 anti-pattern this slice exists to prevent** (a test asserting a destructive default is what let #1638 ship green). |
| ~~`limit <= 0` → `<= 1`~~ | — | predicates | **CORRECTION — now a deterministic KILL, removed from the survivor list.** An added `@example(available=5, limit=1)` on the bounded-count property (`test_1771a_retention_properties.py:829`) kills it at seeds **0 and 777**. |
| ~~`retention_days <= 0` → `<= 1`~~ | — | predicates | **CORRECTION — mislabelled; already killed** by `E::test_window_is_converted_to_hours_by_exactly_24[days=1]` (`test_1771a_retention_edges.py:1036`). |

---

## 3. Verification evidence

```
cd tests && ../.venv-1771a/bin/python -m pytest \
    unit/test_1771a_retention_edges.py unit/test_1771a_retention_properties.py \
    -q -p no:randomly -p no:warnings
→ 146 passed in 11.55s
```

**Branch coverage** (`--cov=services.retention_guard --cov=db.schedules.retention --cov-branch`,
run with the 3 pre-existing neighbours): `191 passed`.

| Module | Stmts | Miss | Branch | BrPart | Cover |
|---|---|---|---|---|---|
| `services/retention_guard.py` | 66 | 0 | 10 | 0 | **100%** |
| `db/schedules/retention.py` | 90 | 0 | 26 | 0 | **100%** |
| TOTAL | 156 | 0 | 36 | 0 | **100%** |

Zero partial branches — no conditional in either target has an uncovered edge.

**Full unit suite on HEAD:** `1 failed, 5049 passed, 17 skipped`. The single failure —
`test_1474_read_boundary_z.py::test_schedules_summary_last_run_at_normalized` (`assert None is
not None`) — was **reconfirmed on a pristine `f7aff466` worktree**, so it is pre-existing and
unrelated to this branch (see follow-ups).

**Order-independence:** `pytest-randomly` is deliberately absent from `requirements-test.txt`,
so it was installed **locally only** to reproduce CI's 3-seed shuffle. The retention
neighbourhood passes at seeds `1`, `12345`, `99999` — **237 passed** each. This confirms the
`_last_refused` module-global isolation holds under shuffling.

`python3 tests/lint_sys_modules.py` → `OK: 203 violation(s) in 60 file(s); baseline allows 240 —
no new violations.`

### Destructive-test isolation — proven in **both** directions

The highest-consequence risk in this slice is a property's per-example `DELETE` running against
a **real** database (`db/engine.py`'s `resolve_database_url()` prefers `DATABASE_URL`) — a
destructive accident inside a task about destructive-path safety. The fixture `delenv`s
`DATABASE_URL` + `TEST_POSTGRES_URL` and asserts the resolved URL is the tmp file **before** any
DELETE. Verified both ways:

- **Positive:** with a decoy DB of 7 marked rows and hostile `DATABASE_URL` / `TEST_POSTGRES_URL`
  exported into the shell, the suite ran **12 passed with all 7 decoy rows intact**.
- **Negative:** with the fixture's `delenv` deliberately neutered, the suite **refused at setup**
  — `AssertionError: REFUSING to run destructive property tests: the engine resolves to …, not
  this test's temp database.` — rather than proceeding.

---

## 4. `hypothesis==6.161.5` — an exact pin, harmonized across the three slices

`hypothesis` was absent from every requirements file, and all three concurrent #1771 slices
need it. **All three now carry one byte-identical comment block + pin at one fixed position**
in `tests/requirements-test.txt` — immediately after `pytest-cov>=6.0.0`. Block
sha256 `bf4d409f804f6ba414dc94a8334bd40bc2817daa2b42e24f6704debe9b549acd`, verified equal in
each slice. Consequence: the three PRs **merge cleanly into `dev` in any order**, and the file
keeps exactly **one** `hypothesis` line whichever lands first — the second and third slices
contribute an identical hunk, not a competing one.

This harmonization was applied deliberately, after the earlier per-slice variants were found to
conflict **with each other** (never with `dev`): each slice had written its own comment wording,
and the insertion positions differed. Both positions fell **inside git's 3-line context window**,
so the hunks overlapped and any second slice to merge would have hit a conflict — and a careless
resolution could have left the file with two `hypothesis` lines. Collapsing to a single canonical
block at a single anchor removes that failure mode by construction rather than by merge-time care.

The pin itself is unchanged from the original slice proposals, and this harmonization touches
**only** the comment text and anchor — no test file, and no `src/` file.

It is the **only `==` among 34 `>=` floors** — a conscious deviation from the file's convention.
Rationale: Hypothesis's example generation and shrinking change between releases, and the #1771
property suites run under a `derandomize=True` profile whose frozen input set is stable only per
version. A floating floor makes a failure non-reproducible across machines and could surface as a
head-only failure in CI's base-vs-head regression diff — the one thing the derandomized profile
exists to prevent. A floor would also auto-adopt unreviewed releases of a package with an
entry-point/plugin surface.

**Honest caveat:** 6.161.5 was uploaded **2026-07-25**, so it had ~1 day of soak when pinned —
effectively zero. Mitigating facts: 0 OSV advisories for `hypothesis` 6.161.5 or its
`sortedcontainers` dependency, and it is a **test-only** dependency that never reaches a runtime
image (`tests/requirements-test.txt` is referenced by no Dockerfile and no compose service).
A reviewer who wants a longer-soaked version should say so — bumping the pin is a one-line change
(that would need re-applying to all three slices to keep them byte-identical).

**Also worth stating for a reviewer who knows the precedent** (and recorded in the in-file comment
so it survives beyond this PR): `pytest-randomly` is deliberately kept **out** of this file
(`backend-unit-test.yml:93-99`) because pytest auto-discovers installed plugins and it would
silently randomize every local run. Hypothesis also ships an auto-loaded plugin, but unlike
`pytest-randomly` it does **not** alter collection order or the behaviour of tests that don't use
`@given` — it only adds `--hypothesis-*` options and a statistics reporter.

---

## 5. Known risk this work **increased** — PostgreSQL collation

`E::test_a_row_exactly_at_the_cutoff_survives` builds its one-tick-older control row as
`frozen[:-1] + "0"` — swapping the trailing `Z` (0x5A) for `"0"` (0x30) to get a same-length
string that sorts strictly before the frozen cutoff. This relies on `'0' < 'Z'`.

The construction is **pre-existing**, but strengthening the assertion to `counted == pruned == 1`
(a positive control, so a predicate matching *nothing* can no longer pass via `0 == 0`) made it
**load-bearing** where it was previously decorative. It is correct under SQLite's byte ordering
and **unverified under a non-C PostgreSQL collation**.

Collation-robust fix, for whoever touches this next: subtract one microsecond from the *parsed*
timestamp and re-emit a valid ISO-Z string, instead of byte-poking the suffix. Not done here —
it is a behavioural change to a passing test, and PostgreSQL is not exercised by any workflow
(`TEST_POSTGRES_URL` is unset everywhere; test ids show `[sqlite]`).

---

## 6. CI note — `schema-parity` will run in full

This PR triggers the **required** `schema-parity` check purely via its
`tests/requirements-test.txt` path entry (`schema-parity.yml:44,90`) **despite touching no DDL**.
Expect it to execute the full parity check rather than self-skip; it should be green.

---

## 7. Follow-ups — **flagged, not filed** (agents don't open issues here)

1. **`architecture.md` says "7 window-driven destructive prunes" / "`RETENTION_OPS_KEYS` carries
   all 7 windows" — there are 8** since #1296 (`f3559506`). Not fixed here: `architecture.md` is a
   3-way conflict magnet across the concurrent #1771 slices. **Mitigated structurally instead** by
   matrix row G3, which pins set-equality via AST and hardcodes neither 7 nor 8.
2. **`retention_guard.evaluate`'s "NEVER raises" docstring (`:153`) is false** for a non-numeric
   `count_fn` return — `candidates <= threshold` (`:183`) sits **outside** the try that wraps
   `count_fn`. Production-unreachable (all 8 `count_fn`s confirmed to return `int(...)`); safety
   today comes from each sweep's outer try/except, not from the contract. The
   `cleanup_service.py:219-221` comment leans on the same false claim. Pinned as observed
   behaviour, not xfail'd (AC#4 forbids the fix).
3. **★ `db/agents.py::count_soft_deleted_agents_past_retention` is the ONE guarded count invoked
   at `limit == 1` in production.** `FLOOR_AGENTS = 0` and `evaluate` calls `count_fn(threshold + 1)`.
   A `limit <= 1`-shaped regression there returns `0` → `0 <= 0` → **allowed** → **unacknowledged
   Docker-volume destruction (#1581)** — the **fail-OPEN** direction. This slice's survivor
   disposition for the `limit`-guard mutants is sound for `db/schedules/retention.py` (its smallest
   production limit is 101) but **does not generalise** to the agents accessor, which is out of
   this slice's scope. Worth its own coverage.
4. **A failed alarm permanently suppresses its own retry.** `announce_refusal` writes
   `_last_refused[key]` at `:247`, **before** the `create_operator_queue_item` try at `:268` — so
   if the alarm write fails, the memo already says "seen" and no later cycle retries it. The
   refusal is still logged at ERROR that cycle, so it is not silent, but the durable
   operator-visible alarm is lost until a restart or a window change. The fix (move the memo write
   after a successful alarm) is product code.
5. **`mutmut` is unusable in this repo** as-is (§2). Worth a shared fix — a `conftest`-level module
   alias, or a documented `also_copy` + `sys.path` recipe — since P-38 mandates `--mutate` on
   future targets.
6. **`feature-flows/cleanup-service.md` Related Files table** (lines 571-576) indexes test files
   for this flow and would conventionally gain the two new files. Not added: the plan's
   docs-delta decision was "none", reaffirmed at orchestration.
7. **Module-import-time `init_database()` runs migrations against whatever `DATABASE_URL` is
   exported, for ANY unit test file** — reproduced with `test_1644`, which predates this branch.
   So "the tests are isolated" is true of the **DML, not the DDL**: §3's isolation proof covers
   the destructive row operations, not schema creation. Pre-existing and repo-wide.
8. **The aged-out `test_1474` fixture** (the one pre-existing full-suite red, §3) belongs to no
   #1771 slice's fence and needs an owner.

---

## 8. Stated plainly — what was **not** verified by execution

- An independent adversarial **tautology-audit subagent was dispatched and never returned**
  (API limit). **Its findings are NOT reflected here** — every property-vs-tautology judgement in
  this PR is the author's own audit. A reviewer should treat property-test tautology as
  un-second-voiced. (The *plan* did receive two independent adversarial reviews — Codex `gpt-5.5`
  and a Claude subagent, 15 findings, all dispositioned — but that was the plan, not the tests.)
- **PostgreSQL was never exercised** (`TEST_POSTGRES_URL` unset in every workflow; test ids show
  `[sqlite]`). See §5.
- **`/verify-local` was deliberately not run** — Andrii's explicit call, justified by: zero `src/`
  changes (mechanically proven); `tests/requirements-test.txt` is referenced by **no Dockerfile and
  no compose service** (verified by two independent `/cso` passes), so the images rebuild
  byte-identical; and the load-bearing unit tier ran in full (§3).
- **gitleaks is not installed locally** — only a pattern grep of the diff was run (clean). CI
  `secret-scan.yml` is the authority.

## 9. Docs delta — none, deliberately

RETENTION-GUARD-001 (`requirements/lifecycle-observability.md:51-71`) already specifies this
behaviour; the slice adds **no capability**, so Trinity's tiered rule puts a test-only PR at
"descriptive commit message only". The one tempting `architecture.md` edit is follow-up 1 above,
mitigated structurally by G3 rather than by a doc edit that would 3-way-conflict with the
concurrent slices.

### DIFF
--- a/tests/requirements-test.txt
+++ b/tests/requirements-test.txt
@@ -4,6 +4,19 @@ pytest-asyncio>=0.24.0
 pytest-timeout>=2.3.0
 pytest-html>=4.1.0
 pytest-cov>=6.0.0
+# Property-based testing (#1771 / P-38 /edge-cases). EXACT pin, not a `>=` floor,
+# deliberately breaking this file's convention: Hypothesis's example generation and
+# shrinking change between releases, so a floating floor makes a derandomized-seed
+# failure non-reproducible across machines and CI runs. Three concurrent #1771
+# slices add this identical line — keeping it byte-identical keeps the merge
+# trivial in any order.
+#
+# Unlike pytest-randomly (deliberately kept OUT of this file, backend-unit-test.yml:93-99,
+# because pytest auto-discovers installed plugins and it would silently randomize
+# every local run), Hypothesis's auto-loaded plugin does not alter collection order
+# or the behaviour of tests that don't use @given — it only adds --hypothesis-*
+# options and a statistics reporter.
+hypothesis==6.161.5
 httpx>=0.28.0
 websockets>=13.0
 python-dotenv>=1.0.1
--- a/tests/requirements-test.txt
+++ b/tests/requirements-test.txt
@@ -4,18 +4,25 @@ pytest-asyncio>=0.24.0
 pytest-timeout>=2.3.0
 pytest-html>=4.1.0
 pytest-cov>=6.0.0
-# Property-based testing (#1771 / P-38 /edge-cases). EXACT pin, not a `>=` floor,
-# deliberately breaking this file's convention: Hypothesis's example generation and
-# shrinking change between releases, so a floating floor makes a derandomized-seed
-# failure non-reproducible across machines and CI runs. Three concurrent #1771
-# slices add this identical line — keeping it byte-identical keeps the merge
-# trivial in any order.
+# Property-based testing (#1771 / P-38 `/edge-cases`). Used by
+# tests/unit/test_1771*_properties.py.
 #
-# Unlike pytest-randomly (deliberately kept OUT of this file, backend-unit-test.yml:93-99,
-# because pytest auto-discovers installed plugins and it would silently randomize
-# every local run), Hypothesis's auto-loaded plugin does not alter collection order
-# or the behaviour of tests that don't use @given — it only adds --hypothesis-*
-# options and a statistics reporter.
+# EXACT pin, deliberately breaking this file's `>=` convention: Hypothesis's
+# example generation and shrinking change between releases, and the #1771
+# property suites run under a `derandomize=True` profile whose frozen input set
+# is stable only per version. A floating floor would make a failure
+# irreproducible across machines and could surface as a head-only failure in
+# CI's base-vs-head regression diff — the one thing the derandomized profile
+# exists to prevent.
+#
+# Unlike pytest-randomly (deliberately kept OUT of this file — see
+# backend-unit-test.yml:93-99 — because pytest auto-discovers installed plugins
+# and it would silently randomize every local run), Hypothesis's auto-loaded
+# plugin changes neither collection order nor the behaviour of tests that don't
+# use @given; it only adds --hypothesis-* options and a statistics reporter.
+#
+# Every #1771 slice adds this block byte-identically at this exact position, so
+# the slices merge cleanly in any order and the file keeps ONE hypothesis line.
 hypothesis==6.161.5
 httpx>=0.28.0
 websockets>=13.0
--- a/tests/unit/test_1771a_retention_edges.py
+++ b/tests/unit/test_1771a_retention_edges.py
@@ -421,12 +421,82 @@ class TestRefusalAlarm:
     def test_alarm_id_is_the_natural_key_and_expiry_is_null(self, guard_db):
         """C9 + C10. `expires_at` must stay None: `mark_operator_queue_expired`
         flips any pending row past `expires_at` to `expired` fleet-wide every 5s,
-        which would silently retire the alarm."""
+        which would silently retire the alarm.
+
+        The id is compared against a LITERAL format, not against `_alarm_id()`
+        itself. Asserting `item["id"] == _alarm_id(...)` is a tautology — both
+        sides move together, so it survives any change to the id's shape. The
+        mutation gate caught exactly that: mutating the `"retention-guard-"`
+        prefix left the original assertion green. The format is load-bearing:
+        `create_item` is INSERT ... ON CONFLICT DO NOTHING on `id`, so the natural
+        key is what makes re-emission idempotent (one alarm per setting+window).
+        """
         _RG.announce_refusal("execution_row_retention_days", "rows", 5, _verdict())
         _agent, item = guard_db.queue_items[0]
 
-        assert item["id"] == _RG._alarm_id("execution_row_retention_days", 5)
+        assert item["id"] == "retention-guard-execution_row_retention_days-5"
         assert item["expires_at"] is None
+
+    def test_alarm_is_raised_as_a_critical_alert(self, guard_db):
+        """The alarm's ROUTING and VISIBILITY fields, not just its key set.
+
+        `priority` decides how loudly a refused mass-deletion surfaces to an
+        operator; downgrading it to `low` is a silent re-run of the #1638
+        failure mode (the signal exists but nobody sees it). `type` decides
+        which queue surface renders it, and `alert_type` is the discriminator
+        consumers filter on. The C8 payload test pins the key SET; these are the
+        VALUES, which the mutation gate showed were unasserted.
+        """
+        _RG.announce_refusal("execution_row_retention_days", "rows", 5, _verdict())
+        _agent, item = guard_db.queue_items[0]
+
+        assert item["type"] == "alert"
+        assert item["priority"] == "critical"
+        assert item["context"]["alert_type"] == "retention_blast_radius"
+        assert item["context"]["setting_key"] == "execution_row_retention_days"
+        assert item["context"]["window_days"] == 5
+
+    def test_window_source_distinguishes_a_db_row_from_a_code_default(self, guard_db):
+        """The provenance field whose ABSENCE made #1638 invisible.
+
+        A `code-default` window is one a future edit to `OPS_SETTINGS_DEFAULTS`
+        can move under the operator's feet; a `db-row` window is one they chose.
+        Reporting the wrong one on a refusal alarm points the investigation at
+        the wrong cause. Both branches asserted — the mutation gate showed only
+        the `db-row` side was covered.
+        """
+        _RG.announce_refusal("execution_row_retention_days", "rows", 5, _verdict())
+        _agent, item = guard_db.queue_items[0]
+        assert (
+            item["context"]["window_source"] == "code-default"
+        ), "no setting row exists, so the window came from the code default"
+
+        guard_db.settings["health_check_retention_days"] = "7"
+        _RG.announce_refusal("health_check_retention_days", "health", 7, _verdict())
+        _agent, item = guard_db.queue_items[1]
+        assert item["context"]["window_source"] == "db-row"
+
+
+class TestVerdictImmutability:
+
+    def test_guard_verdict_cannot_be_mutated_after_the_decision(self):
+        """`GuardVerdict` is a FROZEN dataclass, and that is a safety property.
+
+        The verdict is handed to `announce_refusal` and returned through
+        `_guard_allows` to the sweep that is about to delete rows. If it were
+        mutable, any of those hops — or a future one — could flip `allowed` from
+        False to True after the decision was made, turning a refusal into a
+        prune with no re-evaluation. `test_1644:170` asserts `v.allowed is False`
+        after a failed alarm, which only proves anything BECAUSE the object is
+        frozen; the mutation gate showed `frozen=True` -> `frozen=False` survived
+        the whole suite.
+        """
+        import dataclasses
+
+        v = _RG.GuardVerdict(False, 5000, 1000, "over_threshold")
+        with pytest.raises(dataclasses.FrozenInstanceError):
+            v.allowed = True
+        assert v.allowed is False
 
     def test_alarm_hosts_on_the_uncreatable_sentinel_agent(self, guard_db):
         _RG.announce_refusal("execution_row_retention_days", "rows", 5, _verdict())
--- a/tests/unit/test_1771a_retention_edges.py
+++ b/tests/unit/test_1771a_retention_edges.py
@@ -599,6 +599,41 @@ class TestRetentionSettingsCoercion:
         )
         assert _CS._read_retention_settings() == (11, 22, 33, 44)
 
+    def test_a_missing_row_falls_back_to_that_keys_code_default(
+        self, cleanup_db, monkeypatch
+    ):
+        """The un-configured install — the DEFAULT state, and #1638's blast radius.
+
+        Every other test in this class overrides `get_setting_value` to IGNORE the
+        `default` it is handed, so nothing asserted where that default comes from.
+        The reader passes `OPS_SETTINGS_DEFAULTS.get(key, "0")`
+        (cleanup_service.py:174); replacing that lookup with a literal, or reading
+        it under the wrong key, hands every install-with-no-row a window nobody
+        chose — the #1638 mechanism one layer above the constant it moved.
+
+        The expectation is READ from `OPS_SETTINGS_DEFAULTS` at runtime, never
+        transcribed: pinning the numbers here would BE the anti-pattern this file
+        exists to avoid (learnings.md:101-103). The DIRECTION those defaults may
+        move is pinned by `test_1638_retention_upgrade_safety.py`, which is where
+        that belongs. This test pins only the WIRING.
+        """
+        from services.settings_service import OPS_SETTINGS_DEFAULTS
+
+        # Returning the handed-in `default` is exactly "there is no system_settings
+        # row for this key" — `db.get_setting_value` is a get-or-default accessor.
+        monkeypatch.setattr(
+            cleanup_db, "get_setting_value", lambda key, default=None: default
+        )
+        assert _CS._read_retention_settings() == tuple(
+            int(OPS_SETTINGS_DEFAULTS[key])
+            for key in (
+                "execution_log_retention_days",
+                "execution_row_retention_days",
+                "health_check_retention_days",
+                "agent_reports_retention_days",
+            )
+        )
+
     def test_unicode_digits_are_accepted_by_int(self, cleanup_db, monkeypatch):
         """D5 — OBSERVED, NOT ENDORSED. `int()` accepts non-ASCII decimal digits,
         so an Arabic-Indic '١٢' resolves to a real 12-day window. Harmless (the
@@ -681,22 +716,26 @@ class TestGuardAdapters:
 
     def test_refused_verdict_announces_and_blocks_the_prune(self, monkeypatch):
         """H2. Returning True here is the unrecoverable failure — it is the
-        difference between 'refused' and '95.7% of the table is gone'."""
+        difference between 'refused' and '95.7% of the table is gone'.
+
+        The WHOLE forwarded argument tuple is asserted, not just the setting key.
+        `announce_refusal(setting_key, label, window_days, verdict)` is positional,
+        so transposing `label` and `window_days` would key the transition memo on a
+        label string and stamp the alarm's `context.window_days` with prose — a
+        wrong blast-radius report on the one row an operator reads to decide.
+        """
         calls = []
-        monkeypatch.setattr(
-            _RG,
-            "evaluate",
-            lambda *a, **k: _RG.GuardVerdict(False, 9999, 1000, "over_threshold"),
-        )
+        refused = _RG.GuardVerdict(False, 9999, 1000, "over_threshold")
+        monkeypatch.setattr(_RG, "evaluate", lambda *a, **k: refused)
         monkeypatch.setattr(
             _RG, "note_allowed", lambda key: calls.append(("allow", key))
         )
         monkeypatch.setattr(
-            _RG, "announce_refusal", lambda *a: calls.append(("refuse", a[0]))
+            _RG, "announce_refusal", lambda *a: calls.append(("refuse", a))
         )
 
         assert _CS._guard_allows("k", "label", 5, lambda limit: 9999) is False
-        assert calls == [("refuse", "k")]
+        assert calls == [("refuse", ("k", "label", 5, refused))]
 
     def test_floor_is_forwarded_to_evaluate(self, monkeypatch):
         """A dropped `floor` would silently promote the agent sweep (floor 0, every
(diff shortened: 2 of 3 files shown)


AndriiPasternak31:
Follow-up tracking: the two findings this PR deliberately does **not** fix (AC#4 forbids product-code change) are now filed, so they survive the merge:

- #1833 — `retention_guard.evaluate`'s `:153` "never raises" docstring is false; `candidates <= threshold` (`:183`) sits outside the try wrapping `count_fn` (matrix row A9 / follow-up 2).
- #1834 — a failed alarm permanently suppresses its own retry; `_last_refused[key]` is written at `:247`, before the `create_operator_queue_item` try at `:268` (matrix row C6 / follow-up 4).

Still **flagged but unfiled** from §7 of the body: the `count_soft_deleted_agents_past_retention` fail-open risk (follow-up 3), the `architecture.md` 7-vs-8-windows drift (1), `mutmut` unusability (5), the `feature-flows/cleanup-service.md` Related Files row (6), and import-time `init_database()` DDL isolation (7). Follow-up 8 (the aged-out `test_1474` fixture) is **already resolved on dev** by e9eb40da — it predates this branch's base `f7aff466`, which is why it showed as a local red.

vybe (approved):
Validated: test-only slice of #1771 (retention blast-radius guard). Security scan clean, full pytest matrix green in CI, commit hygiene good. Bare #1771 ref is correct (shared parent issue across three slices — no auto-close wanted).

PR #1826, opened 2026-07-27
test(timestamps): edge-case and property coverage for the ISO helpers and their scheduler mirror

AndriiPasternak31:
**Refs abilityai/trinity#1771** — Part of #1771 (target **2 of 7**). Deliberately **not** a closing keyword: targets 4–7 remain open and three concurrent slice PRs reference this issue. The reviewer must bump `status-in-dev` manually if/when the whole issue lands.

Applies `/edge-cases` (P-38) to the timestamp helpers: `src/backend/utils/helpers.py` (`utc_now_iso`, `iso_cutoff`, `to_utc_iso`, `parse_iso_timestamp`) and the vendored `src/scheduler/utils.py` mirror (`utc_now_iso`, `to_utc_iso`, `parse_scheduler_ts`) — the #476 / #1474 / Invariant 16 lineage.

**Test-only: `git diff origin/dev...HEAD --stat -- src/` is empty** (AC4). Zero product-code change.

## Read this first — the three non-test files, and why a test-only PR touches them

**1. `docs/memory/architecture.md` — a one-clause correction to Invariant 16.** This is the only substantive non-test edit, and an unexplained doc diff on a test PR is exactly what stalls a review, so it leads.

The invariant claimed `src/scheduler/utils.py` is a "vendored **byte-parity** mirror … regenerate from the backend copy, like `failure_classifier.py`". **That claim is false.** #1713 (`15d69d5a`) already established it: the two copies are functionally identical but **textually divergent** — `to_utc_iso` uses an early return where the backend uses `if/else`. #1713 corrected the source docstring and left architecture.md stale, so the doc both contradicted the code *and* prescribed a verification method ("regenerate and diff") that provably cannot work on precisely this slice's subject. Anyone trusting the doc would have concluded the mirror was verifiable by `diff` and stopped there.

The clause now says **behavioural** mirror, states that the contract is agreement on **output** rather than on source text, and names the tests that actually enforce it (`test_1713_scheduler_utils_parity.py`, already on `dev`, plus this PR's two suites). Correcting a factually wrong invariant is in scope for a PR whose entire subject is that mirror. The diff is that one clause and nothing else.

**2. `.gitignore`** — one line, `.hypothesis/`: the example database this PR's new test dependency writes into the working tree.

**3. `tests/requirements-test.txt`** — the two test-only dependencies. See [Dependencies](#dependencies).

No requirements change (this PR adds no capability, endpoint or behaviour) and **no feature-flow change** (the helpers' behaviour is untouched) — stated explicitly rather than skipped.

---

## What this verifies — and what it does not

The properties show the `to_utc_iso` formatter is **order-preserving** — the **function-level prerequisite** of Invariant 16.

This is **not** "Invariant 16 verified". Invariant 16 is a **column-level** property (every writer of a column emits the same shape). Two known column-level divergences are recorded and **deliberately untested**:

- **G5** — `agent_schedules.next_run_at` is mixed-format by design (scheduler writes `Z`, backend writes an offset). The invariant itself calls this out as safe because that column is only ever parse-compared in Python, never lexicographically in SQL. Asserting uniformity there would be a *wrong test*, not a found bug.
- **F8** — `agent_activities.created_at` is `TEXT DEFAULT CURRENT_TIMESTAMP` (`db/schema.py:477`; Postgres `to_char(...,'YYYY-MM-DD HH24:MI:SS')`, `:1696`), so a default-inserted row stores the space-separated, `Z`-less shape into a column Invariant 16 assumes is ISO-Z. Needs its own issue.

**P1 (mirror parity) is a future-regression guard, not a discovery tool.** The two bodies are currently identical operations, so no input can distinguish them *today*; the value is that a future edit to one copy breaks the test. The one real asymmetry (**E6**, below) is a domain edge excluded by the strategy bound and documented rather than swept.

## Edge-case matrix

| | Count |
|---|---|
| Enumerated | **59** |
| Already covered by existing suites | **13** |
| **Newly covered here** | **38** |
| UNSPECIFIED (reported, deliberately not asserted) | **6** |
| Documented-deliberate (asserting them would be a wrong test) | **2** |

The 6 UNSPECIFIED rows are each unreachable or version-flaky, and each says why in the file:

- **A10** year < 1000 — `strftime('%Y')` zero-padding below year 1000 is platform-libc behaviour; measured 4-digit on darwin 3.12/3.13/3.14, **not verified on the glibc CI image**. Strategy is lower-bounded at `1000-01-02` (not `1000-01-01`: `to_utc_iso(datetime(1000,1,1, tz=+05:00))` → `'0999-12-31T19:00:00.000000Z'`).
- **A11** aware datetime within one UTC offset of `datetime.max` → `OverflowError` on `astimezone`. A genuine domain edge; no Trinity column can hold it.
- **A12** "semi-naive" input (`tzinfo` present, `utcoffset()` returns `None`) — the single host-TZ-dependent path in `to_utc_iso`. Unreachable (`timezone.utc`, `pytz`, `zoneinfo` and Pydantic all return a real offset). An assertion here would be flaky across developer machines — a textbook bad test.
- **A15** `datetime` subclass with overridden `strftime`/`astimezone` — no Trinity call site constructs one.
- **C8** float `hours` (type hint says `int`) — no caller passes a float.
- **D12** hour-24 (`'…T24:00:00Z'`) — parses on 3.14, raises `ValueError: hour must be in 0..23` on 3.12 **and** 3.13. CI is 3.11, prod images are 3.13 — any assertion is version-flaky.

Plus one recorded asymmetry that is *why* a bound exists rather than a bug hidden by one:

- **E6** — an offset-bearing string near year 1 or 9999 parses on the **backend** (offset kept as-is) but raises `OverflowError` on the **scheduler** (which must `astimezone`). Verified with `'0001-01-01T00:00:00+01:00'` and `'9999-12-31T23:59:59-01:00'`. Unreachable in Trinity; it is the reason the ISO-zoo generator is year-bounded to 1001–9997, and the bound is documented rather than silent.

---

## ★ A real bug found — F7, shipped as a strict xfail with NO product fix

`test_P7_validator_acceptance_implies_the_parser_can_parse` ships as `@pytest.mark.xfail(strict=True, …)` per AC4. **No product code is changed here.**

**The divergence.** `ReminderCreate._check_fire_at_iso` (`src/backend/models.py:2128`) normalises with `v.replace("Z", "+00:00")` — replacing **every** `Z` in the string — while `parse_iso_timestamp` (`src/backend/utils/helpers.py:116`) strips only a **trailing** `Z`. A `Z` in the middle therefore passes validation and then fails the parser.

Four confirmed inputs (pinned as `@example`, so the xfail is deterministic under the derandomised profile rather than dependent on a lucky draw):

```
2026-01-15T10Z:30:00
2026-01-15T10:30Z:00
2026-01-15 10Z:30:00
2026-01-15 10:30Z:00
```

**Why it is reachable, not a curiosity.** `reminder_service._resolve_fire_at` (`:38-39`) calls `parse_iso_timestamp(data.fire_at)` with **no try/except**, under a comment asserting the very invariant this property falsifies ("fire_at already validated ISO-8601 by ReminderCreate"). `routers/reminders.py:120-137` catches `except HTTPException:` **only**, so the bare `ValueError` reaches FastAPI → **500 where a 400 belongs**. `fire_at` is **agent-supplied** through the MCP `set_reminder` tool (`z.string().optional()`), i.e. the value arrives from outside the platform.

**Amplifier (code-read; needs a live backend to confirm).** The same escape skips `idempotency_service.fail(idem)`, stranding the `in_flight` claim. The key is derived over the **raw** input under the 24h default TTL, so an identical retry gets **409 "still being processed" for up to 24 hours** instead of a second 500.

**Severity for this public repo:** reachable only by an authenticated, self-gated agent key. No auth bypass, no data exposure, no unbounded resource use — a correctness/UX defect on an agent-facing surface, not a security issue.

**Not fixed here** (AC4). The fix is a product change: guard the call, or align the two normalisations.

---

## F3 — decision recorded: the HAZARD-labelled passing pin ships as-is

`iso_cutoff` silently returns a **future** cutoff for negative input, and raises `OverflowError` for huge input. That is pinned as a passing test carrying an explicit inline `# HAZARD — observed behaviour, NOT a contract; see finding F3` label plus the same statement in its docstring, so nobody can cite the green test to reject a fix.

Both positions, honestly:

- **For a strict xfail:** AC4 says a real bug ships as a `strict=True` xfail. The finding calls F3 *reachable* and then declines to xfail it, which weakens the rule to "real bug, unless inconvenient".
- **Against (the shipped position):** the defect is not in `iso_cutoff` — it is the unvalidated `hours: int = 24` at `routers/agents.py:178`, flowing to `db/schedules/stats.py:42,97`. `GET /api/agents/execution-stats?hours=-1` ⇒ a future cutoff ⇒ silently wrong stats; `?hours=10**9` ⇒ `OverflowError` ⇒ 500. That router is a file **this slice is fenced out of**, and `iso_cutoff` has no validation contract to xfail *against* — an xfail here would assert a contract the function was never written to have (a wrong test) or import a router this PR must not touch.

The counter-position was **recorded and not refuted**. Andrii's explicit call at review: **keep the hazard pin**; carry the follow-up text below.

The retention family is *not* exposed: `if retention_days <= 0: return …` guards every prune/count site. The unguarded reachable path is a **stats** endpoint, not a prune.

---

## ★ Non-vacuity — proven, not asserted

A green property test that cannot fail is worse than no test. The suite was mutation-checked: **8 mutants injected, 8 killed, 0 survivors.**

Two attributions were corroborated independently, because "the suite went red" is not proof that the *intended* test caught it:

- **M2** — make `iso_cutoff` emit a space-separated cutoff (**the actual #476 bug**) → **kills P8**. P3 correctly *survives* M2, which proves P8 is not redundant with P3.
- **M6** — desync the scheduler mirror from the backend copy → **kills P1**, which is exactly what a future-regression guard is for.

**M8 is the one worth reading.** `src/scheduler/utils.py::utc_now_iso` was a declared slice target that **nothing in the slice actually called**. A mutant making it silently emit the SQLite space shape **SURVIVED the pre-review suite** — 83 passed, 0 failures, a real hole hiding behind a green run. The review closed it; scheduler `utils.py` branch coverage went **95% → 100%**.

## A vacuity trap caught at plan time

These are the **first `zoneinfo` users in `tests/unit/`**. `st.timezones()` draws real IANA zones deliberately — DST transitions, quarter-hour offsets and historical LMT offsets are what stop the ordering property from degenerating into "a zero-padded format sorts correctly". But `zoneinfo` reads the **system** database first and falls back to the `tzdata` package; with neither, `available_timezones()` returns **0** and the property would go **vacuously green rather than red**.

Hence `tzdata>=2024.1` (Apache-2.0, **zero dependencies**, test-only) and the deliberate removal of a try/except fixed-offset fallback — the failure mode is now loud.

**Honest caveat: the "bare `ubuntu-latest` has zero zones" premise is UNPROVEN.** The failure mode was reproduced with `TZPATH` emptied, not by demonstrating the runner image lacks a system tz database. The dependency is insurance either way; `tests/requirements-test.txt` is referenced by no Dockerfile and no compose service, so it cannot reach a shipped image.

## Honest limits

- **Hypothesis over 20,000 examples could not reach a fold-ambiguous pair.** P4's DST-fold protection comes from the `@example` pins, **not** from the search.
- **Python 3.10 `fromisoformat` semantics are assumed, not asserted** — no 3.10 interpreter was available. 3.11 substantially extended the parser (offsets carrying seconds; fractional parts other than exactly 3 or 6 digits). CI is 3.11 and prod images are 3.13, so version-sensitive behaviour is deliberately left unasserted and stated in the file header.
- The three CI seeds randomise test *ordering*, not Hypothesis input generation — under the `ci` profile they exercise identical inputs for these two files. They add value to the rest of the unit suite; they add none here.

---

## A broken command in the plan — reported, not silently patched

The plan's verification step 3 used `--cov=../src/backend/utils/helpers`, which emits `CoverageWarning: module-not-imported` and **measures nothing**: both modules are loaded via `spec_from_file_location` under synthetic names, so coverage never matches them by module path. The working form is the **directory** shape:

```bash
cd tests && ../.venv/bin/python -m pytest unit/test_1771b_timestamp_helpers_*.py \
  --cov=../src/backend/utils --cov=../src/scheduler --cov-branch --cov-report=term-missing
```

Real figures:

- `src/scheduler/utils.py` — **100%** line and branch.
- `src/backend/utils/helpers.py` — **100% of the four in-scope helpers** (lines 29–127, `BrPart 0`). The 31% whole-file number is four unrelated non-timestamp helpers this slice does not target.

## Dependencies

Two test-only lines in `tests/requirements-test.txt` — a file referenced by **no Dockerfile and no compose service**, so neither can reach a shipped image.

- **`hypothesis==6.161.5` — an exact pin, deliberately.** The properties run under a `derandomize=True` CI profile whose frozen input set is stable only per Hypothesis version; an open lower bound would let a new release change the generated inputs and produce a **head-only** red in the base-vs-head unit regression diff — the one thing the derandomised profile exists to prevent. It is the only `==` among 34 `>=` floors, chosen because a floor auto-adopts unreviewed releases of a package with a pytest-plugin surface.
- **`tzdata>=2024.1` — a floor, deliberately not a pin.** These properties are zone-agnostic format/ordering invariants: a tz-database revision changes *which* zones are drawn but cannot change whether the property holds. An exact pin would freeze a stale IANA database — a real hazard for any future test that *does* assert real-world DST behaviour. (`hypothesis[zoneinfo]` declares tzdata as a floor for the same reason.) **This line is unique to this slice** — 1771a and 1771c do not need it, and it sits separately after `payments-py>=1.0.0`, so it cannot collide with anything.

### Concurrent merge — harmonised, no conflict expected

All three #1771 slices (**1771a, 1771b, 1771c**) now carry the **byte-identical** hypothesis comment block and pin, inserted at **one fixed position: immediately after `pytest-cov>=6.0.0`**.

That harmonisation is the fix for a real problem, not a formality. Earlier revisions each added the line at a *different* offset (1771a around line 7, 1771c around line 3) — close enough to fall inside git's 3-line context window, so the slices' hunks overlapped and conflicted **with each other**, not with `dev`. With a single shared position and identical bytes, git sees the second and third merges as an already-applied hunk: the slices merge cleanly **in any order**, no rebase-and-drop dance is needed, and the file ends with exactly **one** `hypothesis` line.

---

## Why `/verify-local` was skipped

Explicit call, with the reasoning on the record: **zero `src/` changes**, no image-reachable inputs (`tests/requirements-test.txt` is referenced by no Dockerfile or compose service), and the unit tier ran in full instead:

- `tests/unit/` — **1 failed, 4986 passed, 17 skipped, 1 xfailed**, identical across `--randomly-seed` 12345 / 67890 / 99999. The single failure was the pre-existing `test_1474` fixture rot (**F9** below), red on every branch including pristine `dev` at the time — **since fixed on `dev` by #1784**, so it should no longer appear.
- `tests/scheduler_tests/` — **224 passed**. This tier `/verify-local`'s unit stage **does not reach at all**, so running it is strictly more coverage than the gate would have given.
- This slice's own files — **83 passed, 1 xfailed** (the xfail is F7).

---

Agents do not create issues in this repo. Ready-to-paste, in priority order:

1. **F7 — `set_reminder` returns 500 instead of 400 on a mid-string `Z`, and strands the idempotency claim for 24h.** The real bug above. `ReminderCreate._check_fire_at_iso` (`models.py:2128`) and `parse_iso_timestamp` (`helpers.py:116`) normalise `Z` differently; `reminder_service._resolve_fire_at` (`:38-39`) is unguarded and `routers/reminders.py:120-137` catches only `HTTPException`. Fix: guard the call (→400) **and** ensure `idempotency_service.fail(idem)` runs on that path. Regression test already present as the strict xfail — flip it to a passing test with the fix.
2. **F3 — `hours: int = 24` at `routers/agents.py:178` is unvalidated.** `?hours=-1` ⇒ a *future* cutoff ⇒ silently wrong stats (no error, no log); `?hours=10**9` ⇒ `OverflowError` ⇒ 500. Flows to `db/schedules/stats.py:42,97`. Fix: `Query(24, ge=1, le=…)`. The hazard pin in this PR documents the helper's true behaviour and must not be read as blessing it.
3. **F1 — `parse_iso_timestamp`'s docstring is wrong.** It says "Returns: timezone-aware datetime in UTC"; it actually returns a datetime carrying **the input's** offset. Harmless today (every call site either compares instants, which is offset-correct, or pipes into `to_utc_iso`, which converts) but it is a trap for the next `assert dt.tzinfo is timezone.utc`. Docstring-only fix, deliberately not made here (AC4 — and a docstring edit invites a "while I'm here" behaviour change).
4. **F6 — `services/docker_service.py:109,203` parse `labels["trinity.created"]` unguarded.** A malformed Docker label 500s `GET /api/agents` — the whole fleet list. Trinity writes the label itself, but an operator, a hand-crafted container or an older build can supply anything.
5. **F8 — `agent_activities.created_at` default-inserts the space-separated shape** into a column Invariant 16 assumes is ISO-Z (`db/schema.py:477`; Postgres `:1696`). Reads are normalised by `_norm_ts`, so display is correct; the exposure is any *SQL lexicographic* filter on that column — the exact #476 trap.
6. **F9 — ~~time-bomb fixture in the `test_1474` family~~ — RESOLVED on `dev`, no owner needed.** Recorded here for the audit trail, and with a correction: the affected file is **`tests/unit/test_1474_read_boundary_z.py`**, not the `tests/scheduler_tests/test_1474_timestamp_z_suffix.py` this list originally named. Its hard-coded `NAIVE = "2026-07-06…"` fell outside its own rolling 168h window on 2026-07-13 and went red on every branch. Fixed on `dev` by **#1783 → PR #1784** (`e9eb40da`), which derives the constant from `now − 1h` exactly as proposed. Nothing left to do.
7. **Parser fragmentation** — roughly 15 hand-rolled tolerant ISO parsers across the codebase, each re-deriving "strip `Z`, maybe add `+00:00`" slightly differently. F7 is what that class of duplication produces. Worth a consolidation pass onto the two canonical helpers.

---


AndriiPasternak31:
Follow-up tracking: the strict-xfail finding is now filed as #1831, so it survives the merge.

Worth recording on the issue and here — the four pinned `@example` inputs were re-measured across interpreters rather than assumed:

| Python | inputs reproducing |
|---|---|
| **3.13 (prod image, `docker/backend/Dockerfile:1`)** | **4 / 4** |
| 3.12 | 4 / 4 |
| 3.14 | 2 / 4 — `fromisoformat` tightened |

So the finding is fully valid on the runtime that ships. If Trinity ever moves to 3.14 the example set narrows to the two `10:30Z:00` shapes but stays non-empty, so the `strict=True` xfail keeps holding rather than flipping to XPASS.

vybe (approved):
Validated: test-only slice of #1771 (timestamp helpers + scheduler mirror) plus a legitimate architecture.md correction — the Invariant #16 byte-parity claim was false; now stated as behavioural parity enforced by test_1713_scheduler_utils_parity + the #1771b suites. .gitignore .hypothesis/ addition appropriate. Gitlink verified untouched.

PR #1828, opened 2026-07-27
fix(agents): default base image to UTC + ISO-Z file timestamps (#1795)

### LINKED ISSUE #1795: Agent base image hardcodes TZ=America/New_York — agents run 4h off UTC and file timestamps surface as naive local time
## Summary

`docker/base-image/Dockerfile:12` sets `ENV TZ=America/New_York`, so **every agent container runs on US Eastern local time** regardless of where the instance or its operator is. The image's own timezone files disagree with it (`/etc/localtime` → `Etc/UTC`, `/etc/timezone` → `Etc/UTC`), so the container is internally inconsistent: `date -u` is right, everything local is 4–5 hours off.

Verified on v0.8.5 (`b8cf94ca`), `trinity-agent-base:0.8.5`.

## Evidence

Baked into the image, not the deployment:

```
$ docker run --rm --entrypoint sh trinity-agent-base:0.8.5 -c 'echo "TZ=[$TZ]"; date +"%F %T %z"; ls -l /etc/localtime | tail -1'
TZ=[America/New_York]
2026-07-27 04:16:21 -0400
lrwxrwxrwx 1 root root 27 /etc/localtime -> /usr/share/zoneinfo/Etc/UTC
```

Inside a running agent, while real UTC is `08:16:21`:

```
$ docker exec agent-p1-scout date
Mon Jul 27 04:16:50 EDT 2026

$ docker exec agent-p1-scout python3 -c "import datetime; print(datetime.datetime.now())"
2026-07-27 04:16:50.166431
```

Host, backend and agent all agree on UTC — only the agent's *local* time is shifted:

```
host UTC : 2026-07-27T08:16:06
backend  : 2026-07-27T08:16:06
agent -u : 2026-07-27T08:16:06
agent    : 2026-07-27T04:16:06 -0400
```

## User-visible impact

**1. Wrong timestamps in the UI.** Agent-server file metadata is naive local time and is served as-is. A file written seconds earlier reports a time four hours in the past:

```
GET /api/agents/p1-scout/files
  …"modified":"2026-07-27T04:06:58.549706"      # no offset, actually EDT

GET /api/agents/p1-scout/credentials/status
  ".env":{"exists":true,"size":34,"modified":"2026-07-27T04:15:42.543220"}
```

Both are naive strings with no zone marker, so any consumer treating them as UTC (which the rest of the platform is — Invariant #16 mandates ISO-Z) renders "modified 4 hours ago" for a file that was just written. For an operator in UTC+3 the displayed skew is 7 hours.

**2. Agents reason in the wrong timezone.** Anything an agent does with local time — "today's report", daily rollups, "what time is it", deadline arithmetic, timestamps it writes into its own memory/output files — is on US Eastern. On a platform whose core loop is scheduled autonomous work, an agent that disagrees with both its operator and its own scheduler about what day it is will produce quietly wrong output. Schedules themselves are unaffected (they run backend-side with an explicit `timezone` field), which makes the mismatch harder to notice.

## Note

This is not a regression I can attribute to 0.8.5 — the `ENV TZ` line looks long-standing. Raising it now because it is squarely in the "common things a person does" path (open the Files tab, ask an agent to write a dated note) and the fix is one line.

## Suggested direction

Set `ENV TZ=Etc/UTC` in the base image so local == UTC and matches `/etc/localtime`, optionally allowing an operator override via an env var. Separately, the agent-server file/credential metadata would be more robust emitted as ISO-Z UTC (Invariant #16) rather than naive local strings, so a future TZ change cannot shift what the UI displays.

webmixgamer:
## Summary
- The agent base image baked `ENV TZ=America/New_York`, so **every agent container ran on US Eastern** regardless of operator location — disagreeing with its own `/etc/localtime` (`Etc/UTC`), the backend, and the scheduler.
- Agent-server file/credential/dashboard `modified` fields were emitted as **naive local** strings; consumers treating naive ISO as UTC (Invariant #16) rendered a just-written file "hours ago" (e.g. 4h for a UTC viewer, 7h for UTC+3).

## Changes
- **A** — `docker/base-image/Dockerfile`: `ENV TZ=Etc/UTC` (local == UTC; overridable at deploy via `-e TZ=...`).
- **B** — `agent_server/utils/helpers.py`: new `iso_z_from_mtime()` → canonical **ISO-Z UTC**, independent of the container TZ.
  - Applied at all 6 file-metadata sites: `routers/files.py` (×4), `routers/credentials.py`, `routers/dashboard.py`; dropped the now-dead `datetime` imports.
- `tests/unit/test_1795_agent_timestamp_utc.py`: helper correctness + **TZ-independence** + Dockerfile static guard.
- Doc-sync: `docs/memory/feature-flows/skills-on-agent-start.md` snippet updated to match.

## Test Plan
- [x] Unit assertions pass (helper output, TZ-independence, Dockerfile guard) — verified standalone (stdlib helper runs on 3.9).
- [x] `/verify-local` Docker stages green: base image builds, `import agent_server` succeeds **inside the built image**, and a real agent **boots healthy** from the rebuilt image.
- [x] `/cso --diff` clean (no new surface/auth/injection; Invariant #5 parity intact).
- [ ] CI full unit + integration suite (local run was `incomplete` — no Python ≥3.10 on the dev box; environmental, not a defect).

## Notes
- Base-image change: existing agents adopt UTC on **recreate**, not restart (`ENV` is baked at build).
- Frontend consumers (`new Date()`/`parseUTC`) render the zone-marked value correctly; backend proxies `modified` verbatim (no change needed).

Fixes #1795

### DIFF
--- a/docker/base-image/Dockerfile
+++ b/docker/base-image/Dockerfile
@@ -9,7 +9,12 @@ ARG VERSION=latest
 LABEL trinity.base-image-version="${VERSION}"
 
 ENV DEBIAN_FRONTEND=noninteractive
-ENV TZ=America/New_York
+# Default to UTC so container-local time == UTC and matches /etc/localtime
+# (Etc/UTC). Agents, the backend, and the scheduler all reason in UTC
+# (Invariant #16); a non-UTC default made agents run hours off and file
+# timestamps surface as naive local time (#1795). Override at deploy time
+# with `docker run -e TZ=...` if a specific zone is required.
+ENV TZ=Etc/UTC
 ENV TRINITY_BASE_VERSION="${VERSION}"
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
--- a/docker/base-image/agent_server/routers/credentials.py
+++ b/docker/base-image/agent_server/routers/credentials.py
@@ -5,7 +5,6 @@ import os
 import base64
 import binascii
 import logging
-from datetime import datetime
 from pathlib import Path
 from typing import List
 
@@ -25,6 +24,7 @@ from ..models import (
 from ..state import agent_state
 from ..services.trinity_mcp import inject_trinity_mcp_if_configured
 from ..utils.credential_sanitizer import refresh_credential_values
+from ..utils.helpers import iso_z_from_mtime
 # Second-layer credential-path policy (Invariant #5) — byte-identical vendored
 # copy of src/backend/services/credential_paths.py (#11).
 from ..credential_paths import is_allowed_credential_path
@@ -242,7 +242,7 @@ async def get_credentials_status():
             files_status[filename] = {
                 "exists": True,
                 "size": stat.st_size,
-                "modified": datetime.fromtimestamp(stat.st_mtime).isoformat()
+                "modified": iso_z_from_mtime(stat.st_mtime)
             }
         else:
             files_status[filename] = {"exists": False}
--- a/docker/base-image/agent_server/routers/dashboard.py
+++ b/docker/base-image/agent_server/routers/dashboard.py
@@ -6,10 +6,11 @@ Reads and serves dashboard configuration from dashboard.yaml file.
 import logging
 from pathlib import Path
 from typing import Optional, Dict, Any, List
-from datetime import datetime
 
 from fastapi import APIRouter
 import yaml
+
+from ..utils.helpers import iso_z_from_mtime
 
 logger = logging.getLogger(__name__)
 router = APIRouter()
@@ -227,7 +228,7 @@ async def get_dashboard():
 
         # Get file modification time
         stat = dashboard_path.stat()
-        last_modified = datetime.fromtimestamp(stat.st_mtime).isoformat()
+        last_modified = iso_z_from_mtime(stat.st_mtime)
 
         # Ensure defaults
         if 'refresh' not in config:
--- a/docker/base-image/agent_server/routers/files.py
+++ b/docker/base-image/agent_server/routers/files.py
@@ -4,12 +4,13 @@ File browser endpoints.
 import logging
 import mimetypes
 import shutil
-from datetime import datetime
 from pathlib import Path
 
 from fastapi import APIRouter, HTTPException
 from fastapi.responses import PlainTextResponse, FileResponse
 from pydantic import BaseModel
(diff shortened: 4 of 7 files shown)


AndriiPasternak31 (approved):
Validated via `/validate-pr`. Clean PR — approving.

**Traceability** — `Fixes #1795` resolves to an open P2 `type-bug` `theme-reliability` issue, so the merge-time status promotion will fire. **Security/packaging** — no secrets, emails, or public IPs in the diff; no new top-level `src/backend/*.py` and no new `os.getenv()`, so neither the #1033 Dockerfile-COPY nor the #1056 compose-wiring gap applies. **Docs** — bug fix, so commit message is the bar; the `skills-on-agent-start.md` snippet sync is a nice extra. **Tests** — `test_1795_agent_timestamp_utc.py` covers helper correctness, TZ-independence, *and* a static Dockerfile guard, which is the part that actually stops this regressing.

The two-part diagnosis is right, and it matters that both halves shipped together: flipping `ENV TZ` alone would have left the naive `datetime.fromtimestamp(...).isoformat()` timestamps wrong for any operator who overrides `TZ` at deploy, and fixing only the timestamps would have left agents reasoning in Eastern.

Two non-blocking notes:

**1. The helper is duplicated.** `iso_z_from_mtime` is defined in `agent_server/utils/helpers.py` and re-declared as `_iso_z_from_mtime` in `agent_server/routers/files.py`. The rationale is documented — `tests/unit/test_ent183_skill_packages.py` loads `files.py` by path with no package context, so a relative import breaks the load — but that's a test loader dictating production structure, and it leaves two copies where only one is covered by the helper test. If they ever drift, `files.py` (4 of the 6 call sites) is the copy that drifts silently. Fixing the loader to import with package context would let both sites share the one helper.

**2. Worth a release note.** Agents move from `America/New_York` to UTC on **recreate**, not restart. Combined with the #1809 cold-start adoption predicate, an operator gets this the next time an agent cold-starts — so any agent skill or prompt doing local-time reasoning shifts 4–5h with no obvious trigger. Cheap to call out in the notes; expensive to debug in the field.

Neither blocks. 18/18 checks green, `/verify-local` Docker stages run by the author including a real agent boot from the rebuilt base image.

Issue #254, opened 2026-04-03
feat: Canonical agent development workflow documentation and tooling


vybe:
## Problem

The agent development workflow (create → develop → deploy → sync → operate) is tribal knowledge spread across multiple plugins and skills. New developers don't know the canonical path.

## Solution

Define and implement the canonical agent development workflow based on the Abilities plugins:

### The Workflow
1. **Create**: `create-agent` — scaffolds Trinity-compatible agent with skills/playbooks
2. **Develop**: `create-playbook` / `adjust-playbook` — build and iterate on agent skills locally
3. **Prepare**: `trinity-onboard` — makes agent Trinity-compatible, creates GitHub repo
4. **Deploy**: `trinity-sync` — syncs local agent to remote Trinity instance via Git
5. **Configure**: `credential-sync` — pushes local credentials to remote agent
6. **Operate**: `trinity-remote` / `trinity-schedules` — run skills, manage schedules from local Claude Code

### Deliverables
- [ ] Document the canonical workflow as a skill or guide within the Abilities plugin
- [ ] Ensure each step has clear error messages pointing to the next step
- [ ] Add a `trinity-workflow` or `trinity-quickstart` skill that walks users through the full flow
- [ ] Each plugin step validates prerequisites (e.g., "repo exists" before deploy)

## Context

From weekly sync: the team walked through this flow and agreed it should be the standard path. Currently works but requires knowing which plugins to call in which order.

Issue #253, opened 2026-04-03
feat: Auto-create GitHub repos during agent creation


vybe:
## Problem

Agent creation and deploy require a GitHub repository, but Create Agent doesn't create one. Users must manually create a repo, which breaks the flow and causes deploy failures when forgotten.

## Solution

Auto-create a private GitHub repository for the user during agent creation:

1. During `create-agent` flow, automatically create a private GitHub repo under the user's account (or a configured org)
2. Initialize the repo with the scaffolded agent files
3. Record the repo URL in the agent's CLAUDE.md and Trinity config
4. Ensure subsequent `trinity sync` and `trinity onboard` work seamlessly

## Context

This closes the gap in the agent creation → deploy pipeline. Currently users must manually create repos, which is error-prone and causes the file-based deploy fallback (which is broken).

## Acceptance Criteria

- [ ] `create-agent` skill auto-creates a private GitHub repo
- [ ] Repo is initialized with agent scaffold files and pushed
- [ ] Agent CLAUDE.md references the repo
- [ ] Deploy and sync work without additional manual steps

vybe:
Cross-link: the **fork** path of the new GitHub-repo import wizard (private `abilityai/trinity-enterprise#15`, epic #13) reuses this auto-create-repo primitive. Worth sequencing together.

vybe:
Readiness triage (groom 2026-07-03): closing as aged/superseded — incubating ~14 months, never started. Repo provisioning for GitHub-native agents is now largely covered by `initialize_github_sync` + `github_service` repo creation. Reopen if a concrete need for auto-creation at agent-creation time resurfaces.

Issue #128, opened 2026-03-14
Startup recovery for regular task executions


vybe:
## Summary

When the backend crashes or restarts, regular task executions (scheduled, chat, MCP-triggered) stuck in `running` status are not recovered on startup. They remain in `running` until the cleanup service marks them as failed after 120 minutes. This wastes capacity slots and leaves stale state in the UI.

## Context

Process Engine executions already have startup recovery via `ExecutionRecoveryService` in `executions.py` — this issue extends the same concept to regular `schedule_executions`.

### What Exists Today

| Mechanism | Scope | Gap |
|-----------|-------|-----|
| `ExecutionRecoveryService` | Process Engine executions only | Does NOT cover regular task executions |
| `CleanupService` (every 5 min) | Marks `running` > 120 min as failed | 120 min is too long for crash recovery |
| Redis slot TTL | Auto-expires slots after `timeout + 5min` | Doesn't update DB execution status |

### What Happens After a Crash

1. Backend restarts
2. `schedule_executions` rows with `status='running'` from before the crash still exist
3. Agent containers may have been restarted too — process registry is empty
4. The execution is effectively orphaned but still shows as "running" in the UI
5. Cleanup service will eventually mark it failed — but only after **2 hours**
6. During this time, the capacity slot may also be orphaned in Redis

## Proposed Solution

On backend startup (in `main.py` lifespan handler, after existing recovery):

1. Query `schedule_executions` WHERE `status = 'running'`
2. For each execution:
   a. Check if the agent container is running (`docker inspect`)
   b. If container is running, check agent's process registry: `GET /api/executions/{id}/status`
   c. If execution is NOT found on agent → mark as `failed` with error: `"Execution orphaned — recovered on backend restart"`
   d. If container is not running → same treatment
   e. If execution IS found on agent → leave it alone (still legitimately running)
3. Release any associated capacity slots for recovered executions
4. Log recovery summary (count recovered, count still running)

## Files to Change

- `src/backend/main.py` — Add call to regular execution recovery in lifespan handler
- `src/backend/services/cleanup_service.py` — Add `recover_orphaned_executions()` method (or new service)
- `src/backend/db/schedules.py` — Add query for running executions with agent cross-check

## Acceptance Criteria

- [ ] On startup, `running` executions are checked against agent process registries
- [ ] Orphaned executions (not found on agent) are marked `failed` with descriptive error
- [ ] Capacity slots are released for recovered executions
[...]

PR #1835, opened 2026-07-28
fix(templates): stop a malformed `credentials:` block emptying the catalog (ent#128 PR-A)

AndriiPasternak31:
PR-A of **[Abilityai/trinity-enterprise#128](https://github.com/Abilityai/trinity-enterprise/issues/128)** — closes **AC #5** on its own. The declaration standard (AC #1–4) is PR-B and re-gates after this merges.

> Cross-repo `Fixes` does not fire, so this only **references** the issue. Close it manually at the release cut.

## Why now

Three live bugs on `dev`, each reproduced by execution on `4d743fdb` before any fix. They need no schema decision, which is why they were split out ahead of the standard itself.

### 1. One malformed template 500s the entire catalog

`_build_local_template:252` did `data.get("credentials", {}).get("mcp_servers", {}).keys()`. The `try/except` wrapped only the YAML load and the `isinstance` guard checked only that `data` was a dict, so six distinct shapes escaped uncaught through `get_local_templates` → `routers/templates.py` → **HTTP 500, empty catalog, every template gone**:

| Trigger | Result |
|---|---|
| `credentials:` is a list / string / scalar | `AttributeError: 'list' object has no attribute 'get'` |
| `credentials:` present but **null** | `AttributeError` — the `.get(..., {})` default applies only when the key is *absent* |
| `credentials.mcp_servers` is `[]` / null / str | `AttributeError: … has no attribute 'keys'` |
| **`_build_template` (GitHub) — no `isinstance` guard at all** | `AttributeError` on list, str **and int** |
| deeply nested YAML | `RecursionError` — a `RuntimeError`, so it escapes `except yaml.YAMLError` entirely |

The GitHub path is the worst of them: it is the *untrusted* source (`_fetch_template_yaml` returns `safe_load(...) or {}`, and a top-level list is truthy), and it had none of the protection the local path had.

### 2. A string `env_file` silently corrupts the agent's `.env`

Forgetting the list dash is an ordinary authoring typo:

```yaml
credentials:
  env_file: "OPENAI_API_KEY"     # meant: - OPENAI_API_KEY
```

`generate_credential_files:633` iterated it **character by character**:

```
.env:  O=\nP=\nE=\nN=\nA=\nI=\n_=\nA=\nP=\nI=\n_=\nK=\nE=\nY=
```

Fifteen single-letter variables, the real credential **never written**, and **no error, no warning, no crash**. The agent boots and its MCP server fails at first use with nothing pointing at the cause. Per "zero silent failures" this outranks the crash above — a crash is visible, this is not.

### 3. 🔴 `config_files[].path` is an arbitrary-file-write primitive

`generate_credential_files` built `files[file_path]` from an author-controlled `path`; `_stage_config_files` then did `mkdir(parents=True)` + `open(w)` with no normalization. Executed:

```
path='/tmp/absolute-pwned.txt'      -> writes /tmp/absolute-pwned.txt   (absolute wins the join)
path='../../../../../tmp/pwned.txt' -> escapes the staging directory
```

Reachable from untrusted input: `deploy_local_agent_logic` accepts a base64 `tar.gz` from **any authenticated user**, lands it in `/data/deployed-templates`, and `_resolve_local_template` feeds it straight to `generate_credential_files`.

## The fix

Four tolerant readers in `template_service.py` are now the only way into the block, with **two deliberately opposite contracts**:

| Helper | Contract |
|---|---|
| `credential_shape_errors(block)` | Named errors. Never raises. Absent / null / `{}` = zero credentials, **not** an error |
| `credential_mcp_server_names(block)` | `[]` for any odd shape at either level |
| `credential_env_file_names(block)` | Non-empty strings only — backs the `.env` writer |
| `CredentialDeclarationError` | HTTP-free domain error, raised by the **write** path only |

- **Read paths never raise.** Degrade the derived field, attach `credential_errors` to the entry, log **one** WARNING naming the template id — and the template **still lists**. `get_local_templates` additionally fences each per-template build, so a future unguarded field can't regress the property.
- **The write path fails loud.** Validate first, raise, and the only caller maps it 1:1 to **400 `INVALID_CREDENTIAL_DECLARATION`** (Invariant #1 — services hold no HTTP concerns). The call sits *before* the docker `try/except`, so it is not flattened to a 500.
- **`config_files[].path`** rejected at the parse boundary **and** re-checked at the write sink (`crud._safe_cred_file_path` → 400 `INVALID_CREDENTIAL_FILE_PATH`) — validate-at-boundary-and-at-sink. Two steps, mirroring `_safe_local_template_path`: allowlist the raw string, then resolve and assert `is_relative_to(root)`.

## Deliberate non-changes

- **`credentials.env_file` stays a names-only list.** The enriched per-variable declaration lands under its own top-level key in PR-B, precisely so an already-deployed Trinity reading a newer template is structurally untouched — no floor version required.
- **MCP-server precedence is unchanged**, so cornelius still reports `['aistudio','trinity']` rather than its three real servers. That flip (and the never-rendering `required_credentials` badge) is PR-B — this PR adds no behaviour change beyond crash-avoidance.
- Absent / null / `{}` remain a valid zero-credential contract: the ent#124 starter trio ships exactly that, and a commented-out block must not acquire a spurious warning.

## Verification

- **`tests/unit/test_ent128a_catalog_resilience.py` — 33 pass here, 31 fail on `origin/dev`.** The two headline cases fail with the exact bug signatures: `AttributeError: 'list' object has no attribute 'get'`, and the literal `O=\nP=\nE=\nN=` output. That failure-on-dev is the proof these are real bugs, not defensive padding.
- Includes a **parity test over the shipped bundle**, so a malformed *bundled* template fails CI instead of a user's fresh install.
- **Full unit suite: 5699 passed, 14 skipped, 2 xfailed, 0 failed** on the rebased branch (base `04d11cf2`). All six CI `pytest` jobs (3 seeds x base/head) pass.
- **`/verify-local`: `status: pass`** - image build + in-image `import main` (8s), boot + health (12s), and **70 integration passed / 13 skipped** against the booted stack. Run with `--skip-agent`, justified: this touches zero `docker/base-image/**` files. Honest caveat: that run predates the rebase onto `04d11cf2`. The PR's own delta is byte-identical across the rebase (same file set, same per-file numstat, 1029 diff lines both sides), but the base moved 15 commits - including #1871's compose changes - so the image-level result has not been re-verified post-rebase.
- **CodeQL: gate green - the 2 `py/path-injection` alerts ([#253](https://github.com/Abilityai/trinity/security/code-scanning/253), [#254](https://github.com/Abilityai/trinity/security/code-scanning/254)) are dismissed as verified false positives.** The tainted term is **`root`, not the template-declared path**: the SARIF source is `routers/agents.py:436` (the `config` request body) -> `config.name` -> `crud.py:1066` `Path(f"/tmp/agent-{config.name}-creds")`, passed in as `root`. **That construction line is unchanged from `dev`** - the PR introduces no new exposure; adding the guard only moved `.resolve()` onto new lines, which is what scored it as "new alert in changed code". `config.name` is sanitized on all four routes into the sink: `create_agent_internal:2052` writes `sanitize_agent_name` back onto `config.name` (so the `/tmp` join reads the sanitized value); `deploy.py:442` sanitizes, then `get_next_version_name` appends only `-<int>`; `system_service.py:131` validates `^[a-z0-9][a-z0-9-]*[a-z0-9]$` and `resolve_agent_names:259` composes `{system}-{short}` from two validated halves; the Cornelius seeder passes a hardcoded name with sanitization on. `sanitize_agent_name` (`helpers.py:244`) maps every character outside `[a-zA-Z0-9_.-]` to `-`, so no path separator survives and a bare `..` strips to `""` -> 400 at `crud.py:2054`; verified by execution over 411 payloads (traversal, percent-encoded, unicode solidus, NUL, every byte 0-255) with **zero** cases where the resolved directory was not a single component directly under `/tmp`. **Precedent:** the identical taint at the identical construction site is already dismissed on `dev` (#110-#114, `crud.py` 994/999/1000/1016/1020, same rationale), and `dev`'s `crud.py:778` carries an in-source note that this block re-fingerprints these alerts on any refactor (#1793 reverted its own guard-clause change for exactly this). Control proving it is the tainted *value* and not the guard *form*: `_safe_local_template_path` uses the byte-identical `resolve()` + `is_relative_to(root)` pattern ~50 lines above and is **not** flagged - its `root` is a module-level constant. `dev` already carries 73 open alerts from this same query.
- `tests/lint_sys_modules.py` clean (new tests use the `monkeypatch.setitem` form).
- Smoke against the real 15-template bundle: a dropped-in malformed template lists with its named error, zero collateral on the other 15.

## Known follow-ups (intentionally out of scope)

- **`docker/base-image/agent_server/routers/info.py:191`** has the *same* live crash on `GET /api/template/info` (Info tab + brain-orb route guard). Left for PR-B on purpose: it lives in the agent base image, so fixing it here would drag a base-image rebuild into this PR and invalidate the `--skip-agent` verification envelope.
- `crud.py:739` — the same reach-through, swallowed by a broad `except` that silently loses the agent's `runtime:` / `shared_folders:` config alongside it.
- `config_files` is **hardened here, not documented**; the plan's settled decision is to remove the feature in PR-B rather than publish a file-write primitive in a contract aimed at third-party authors. Zero of 25 bundled templates use it. Wants a 👍 from @vybe before deletion, since it is a public behaviour change.
- YAML-level hardening (alias-bomb serialization amplification, duplicate-key silence) is pre-existing and platform-wide; the gate settled it as its own security issue, not part of #128.

## Docs

`docs/memory/feature-flows/template-processing.md` (new "Malformed `credentials:` resilience" section + the response-schema field), `architecture.md` catalog line, `feature-flows.md` index row.

**No DB change → the dual-track migration rule (#9) does not apply.**


vybe (approved):
Validated via `/validate-pr` — approving. Ran `tests/unit/test_ent128a_catalog_resilience.py` locally at `c07afab7`: **33 passed**. All 22 CI checks green, and I confirmed **zero open code-scanning alerts** on the head ref, so the two high-severity `py/path-injection` flags on `_safe_cred_file_path` are genuinely resolved by the two-step barrier rather than dismissed.

What I verified beyond the test run:

- **The read/write contract split is the right call and is actually implemented that way.** Read paths (`credential_mcp_server_names`, `credential_env_file_names`) degrade to `[]` for a null / list / string / scalar block at either level, so one malformed template costs itself its credential metadata and nothing else — the catalog still lists. The write path raises `CredentialDeclarationError`, which is correct: generating an agent's `.env` from a declaration nobody can parse is precisely the silent corruption being closed.
- **`CredentialDeclarationError` is HTTP-free and mapped 1:1 at its only caller** (`crud.py::_stage_config_files` → named 400 `INVALID_CREDENTIAL_DECLARATION`). That respects Invariant #1 rather than reaching for `HTTPException` inside a service.
- **The `_build_template` guard closes the worse half.** The GitHub path is the untrusted source and had no `isinstance` guard at all, while `_build_local_template` always did — and `_fetch_template_yaml` returns `safe_load(...) or {}`, so a top-level list is truthy and sails through. Fixing only the local path would have left the exploitable one live.
- **The path guard is defence-in-depth at both ends** — rejected at the parse boundary by `_config_files_shape_errors` and re-checked at the write sink by `_safe_cred_file_path`. Since `deploy_local_agent_logic` lets any authenticated user upload a template archive, `config_files[].path` really was an arbitrary-file-write primitive; the allowlist-then-resolve shape mirrors the existing `_safe_local_template_path` precedent instead of inventing a new one.
- **Present-but-null is handled**, which is the subtle one — `.get("credentials", {})` applies its default only when the key is *absent*, so a bare `credentials:` with nothing under it yielded `None` and blew up every downstream `.get()`.

Docs are updated in-diff (`architecture.md`, `feature-flows.md`, and the `template-processing.md` flow), which is what a behaviour change at this boundary warrants.

Two process notes, neither blocking:

- The cross-tracker reference to `abilityai/trinity-enterprise#128` is the expected shape here and deliberately carries no closing keyword, since this is PR-A closing **AC #5 only** — PR-B still owns AC #1–4. So I am **not** bumping ent#128 to `status-in-dev`; it correctly stays `status-in-progress` until the declaration standard lands. The manual close at the release cut is already called out in the PR body.
- The stale nightly-suite comment about a merge conflict no longer applies — the branch is `MERGEABLE` against current `dev`.

PR #1837, opened 2026-07-28
fix(seed): converge the Cornelius flag on a 409 only when the agent exists (#1790)

### LINKED ISSUE #1790: Cornelius seeder marks itself seeded on a volume-conflict 409 — Cornelius never created, never retried
## Summary

`cornelius_agent_service` treats **any** HTTP 409 from the create path as "Cornelius already exists" and durably sets `cornelius_seeded=true`. But the create path also raises 409 for an entirely different condition — the #1671/#1664 unclaimed-data-volume refusal. When that fires, Cornelius is **never created and never retried**, on an install where the flag says it was seeded.

Observed on v0.8.5 (`b8cf94ca`), fresh install, empty DB, no `cornelius` agent and no `agent_ownership` row.

## Evidence

Backend log, fresh install:

```
services.cornelius_agent_service  Brain Orb enabled by default (seeded brain_orb_enabled=true)
services.cornelius_agent_service  Cornelius seed: Cornelius already exists (409) — marking seeded
```

Resulting state — flag set, agent absent:

```
sqlite> select key from system_settings where key like '%seed%';
cornelius_seeded = true
sqlite> select count(*) from agent_ownership where agent_name='cornelius';
0
```

The only thing present was a leftover `agent-cornelius-workspace` volume with no owning row.

## Root cause

`services/agent_service/crud.py:303` raises `HTTPException(status_code=409, …)` for the volume-claim refusal:

> Data volume 'agent-X-workspace' already exists and no agent claims it, so its contents would silently become this agent's home directory. Refusing to adopt another agent's leftover data.

`services/cornelius_agent_service.py:148-154` branches on the status code alone:

```python
# 409 = the agent already exists (another worker won, or a prior partial
# run). Treat as success: converge the flag so we stop retrying.
status_code = getattr(e, "status_code", None)
if status_code == 409:
    db.set_setting(_SEEDED_FLAG, "true")
    self._seed_brain_orb_flag()
    return self._skip(result, "already_exists", "Cornelius already exists (409) — marking seeded")
```

Two distinct conditions share one status code. "Another worker won" is genuinely converged; "a stale volume blocked creation" is a hard failure that is now unrecoverable, because the flag suppresses every later attempt (and `cornelius_seeded=="true"` additionally forces a NOT-fresh verdict in `system_seed_service.ensure_first_run_seeded()`).

## Contrast — the sibling seeder gets this right

`system_seed_service` deliberately does **not** set its flag when 0 agents are created, so it retries on the next boot. Verified: after clearing the blocking volumes and restarting the backend, the starter fleet deployed correctly on the retry pass. Cornelius has no equivalent recovery.

## Repro

1. Create a Trinity instance, let Cornelius seed, then `docker compose down -v` (destroys the DB; agent volumes are **not** compose-managed and survive).
2. Re-run `./scripts/deploy/start.sh` and complete setup.
3. The leftover `agent-cornelius-workspace` triggers the 409 → `cornelius_seeded=true`, no Cornelius.

This is the ordinary "let me start over" flow, so it is reachable without any exotic setup.

## Suggested direction

Distinguish the two 409s before converging the flag — e.g. a typed exception or an error code on the volume-claim refusal — and treat the volume conflict as retryable (or surface it as an operator alert) rather than as success.

dolho:
## What

The Cornelius seeder converged its durable `cornelius_seeded` flag on **any** 409 from the create path. Now it converges only when something actually claims the name.

Related to #1790

## Why it mattered

Three of the create path's 409s mean *nothing was created*:

| 409 source | Meaning | Old handling |
|---|---|---|
| `_check_name_availability` → "Agent already exists" | genuinely there (race backstop) | converge ✅ |
| `_guard_leftover_workspace_volume` (#1667) | unclaimed leftover volume — **refused** | converge ❌ |
| `is_volume_base_reserved` (#1664) | volumes owned by a renamed agent — **refused** | converge ❌ |
| `FORK_DESTINATION_IN_USE` | destination repo bound — **refused** | converge ❌ |

On the bottom three the flag suppressed every later attempt, so the install stayed permanently Cornelius-less while reporting seeded — and `system_seed_service.ensure_first_run_seeded()` then read the install as not-fresh. Recovery meant hand-editing `system_settings`.

Reachable via the ordinary "start over": `docker compose down -v` destroys the DB but **not** the agent volumes (not compose-managed), so re-setup hits the leftover `agent-cornelius-workspace`.

## Approach

Verify by fact, not by status code. On a 409 the handler asks whether anything claims the name; converges only then, otherwise leaves the flag unset, returns `create_blocked`, logs the cause, and the next boot retries. The common blocker is reclaimed unattended by the #1581 orphan-volume sweep, so the retry converges with no operator action.

Two deliberate choices worth reviewing:

- **Shared predicate, not a typed exception.** The issue suggested typing the #1667 refusal. Instead the existence check inside `_check_name_availability` is extracted as `crud.agent_name_is_taken` and reused by the seeder. That covers *every* non-existence 409 — including the two the issue didn't name and any added later — and it keeps `detail` a plain string, so no 409 response shape changes for the UI. Sharing beats copying here: a duplicated predicate would drift the moment a new claim source lands in the create path.
- **The probe fails toward retry.** If the check itself errors we report absent. The errors aren't symmetric: a wrong "absent" costs one redundant create attempt that 409s again next boot; a wrong "present" burns the durable flag and is unrecoverable without editing the DB. That asymmetry *is* the bug.

Precedent: `lifecycle.py` (#1809) already handles a Docker 409 the same way — catch, then confirm the container is really there before adopting.

Note one preserved behaviour: a container-only leftover with no ownership row still counts as taken, because that's what the create path itself refuses on. It converges as before.

## Verification

```
$ pytest tests/unit/test_ent107_cornelius_seed.py -q
18 passed
```

New cases, run against the **pre-fix** handler (fix stashed, tests kept) — both fail, so they pin the actual defect:

```
FAILED test_409_with_no_agent_does_not_burn_flag
FAILED test_409_presence_probe_failure_fails_toward_retry
2 failed, 16 passed
```

Full unit suite:

```
4993 passed, 16 skipped, 4 failed
```

All 4 failures are pre-existing and environmental — `ModuleNotFoundError: No module named 'tzdata'` (test_1472) and 3 in `test_admin_email_login.py` — reproduced identically on an untouched checkout at the same commit. Crud-adjacent files (`test_1484`, `test_1664`, `test_1581`, `test_fork_to_own`, `test_1646`) pass 144/144, covering the predicate extraction.

Not done: no operator-queue alert for a blocked seed. Retry + an ERROR log naming the cause is the fix for the reported defect; an alert is a separate call.

Fixes #1790


vybe (approved):
Validated via /validate-pr: shared claim predicate (no drift), fail-toward-retry asymmetry is right, regression tests named to #1790 cover both 409 branches + probe failure + all claim sources, feature-flow doc updated. Approving.

Issue #1534, opened 2026-07-09
Epic: Agent Reports — richer creation, export, scale, read-back & search


vybe:
## Summary

Extend the Agent Reports subsystem (#918) from "an agent can publish a structured JSON report" into a complete reporting surface: agents know how to report without hand-holding, reports export to the formats humans actually use (Excel/PDF), the pipeline holds up under large datasets, agents can read back what they previously reported, and users can search/filter across the report history.

## Context

Reports today (#918): an agent calls the MCP `report` tool → `POST /api/agents/{name}/reports` → `agent_reports` table → Reports tab + fleet Reports view + thin `agent_report` WS trigger. That's a solid create-and-display primitive, but several gaps keep it from being a first-class feature:

- Agents don't reliably use it — nothing in the platform system prompt tells them the tool exists or when to reach for it.
- Reports are trapped as on-screen JSON/cards — no way to hand a stakeholder an Excel or PDF.
- The 256 KB payload cap + single-blob storage don't scale to large tabular datasets.
- An agent can't see its own reporting history to avoid duplicate/contradictory reports.
- Users can't search or filter the accumulating report history.

This epic groups the five follow-ups as separate, independently-shippable issues.

## Children

1. Inject basic reporting (MCP `report`) guidance into the agent platform prompt
2. Download & render reports in popular formats (Excel, PDF)
3. Make reporting handle large datasets (payload scale, pagination, streaming)
4. Let agents read back their old reports via MCP
5. Search & filter reports (fleet + per-agent)

## Acceptance Criteria

- [ ] All five child issues created and linked as sub-issues
- [ ] Each child independently shippable (no hard ordering dependency beyond the shared table/endpoints)
- [ ] Enhancements stay OSS-core, extending the existing public Reports surface (#918)

## Technical Notes

Anchor surfaces (all OSS/public): `src/backend/routers/reports.py`, `services/report_service.py`, `db/reports.py`, `src/mcp-server/src/tools/reports.ts`, `services/platform_prompt_service.py` (prompt injection), `components/reports/` (frontend renderers), table `agent_reports`. See architecture.md → "Agent Reports (#918)".

Issue #1104, opened 2026-06-09
feat(mcp): respond to / resolve Operator Queue items over MCP (#1101 follow-up)


vybe:
## Summary

Add MCP write tool(s) to **resolve** Operating Room (operator queue) items — answer questions, approve/deny approval requests, and (optionally) cancel pending items — so an agent or external Claude Code client can close the loop on alerts it can already *read* via #1101. Today the operator queue's write surface is REST-only (`POST /api/operator-queue/{id}/respond`, `POST /api/operator-queue/{id}/cancel`); the only MCP write tool in the OPS-001 family is `send_notification` (raise an alert). There is no MCP tool to respond to or resolve an existing item.

## Context

#1101 shipped the **read** half of the operator-queue MCP surface (`list_operator_queue`, `get_operator_queue_item`, `src/mcp-server/src/tools/operator_queue.ts`) and explicitly deferred the write/respond half to a follow-up ("Write actions (respond / cancel) are intentionally deferred to a follow-up — this surface is read-only").

This is that follow-up. Read-only is the less actionable half: an agent can now *see* a sibling's pending approval or question but cannot act on it without a human at the web UI or a hand-curled REST call. Exposing the respond/resolve surface over MCP completes the raise → observe → respond round-trip and finishes Invariant #13 (MCP = third surface in sync) for OPS-001.

## Acceptance Criteria

- [ ] `respond_to_operator_queue` MCP tool added to `src/mcp-server/src/tools/operator_queue.ts` — proxies `POST /api/operator-queue/{id}/respond` with `response` (required) and optional `response_text`
- [ ] (Optional, decide in review) `cancel_operator_queue_item` — proxies `POST /api/operator-queue/{id}/cancel`; consider leaving out of v1 given wider blast radius
- [ ] Access control honored for writes: agent-scoped keys may only resolve items for `{self} ∪ permitted` agents — reuse the existing `checkAgentAccess` gate in `operator_queue.ts` (the backend resolves an agent-scoped key to its owner and does NOT apply `agent_permissions`, so the agent-to-agent gate must live in the MCP layer, same as the read tools)
- [ ] Resolve the item's `agent_name` before gating (fetch via `get`/list if the caller only supplies an `id`) so the permission check has a target agent
- [ ] State-machine errors surfaced cleanly: the backend rejects respond/cancel on a non-`pending` item (HTTP 400) — return a structured `{ error }` rather than throwing
[...]

PR #1838, opened 2026-07-28
feat(reports): agent-facing report lifecycle — prompt, read-back, filters, scale, export (#1534)

### LINKED ISSUE #1535: Inject reporting (MCP report tool) guidance into the agent platform prompt
## Summary

Teach agents to use the reporting capability by injecting a short description of the MCP `report` tool — what it is, when to use it, and the payload/`display_hint` shapes — into the platform system prompt. Today nothing tells an agent the tool exists, so reports only get published if the agent's own CLAUDE.md happens to mention them.

## Context

Part of the Agent Reports epic. The `report` MCP tool (#918) is fully functional but under-used because it's not surfaced to agents by default. A concise, runtime-aware blurb in `platform_prompt_service` makes reporting a first-class, discoverable behavior across the fleet without per-agent template edits.

## Acceptance Criteria

- [ ] Platform system prompt includes a brief `report` tool description (purpose + when to reach for it)
- [ ] Covers the payload shapes per `display_hint` (`table`/`kpi`/`markdown`/`timeline`/`json`) at a glance
- [ ] Runtime-aware — respects the Codex bare-`trinity` vs Claude `mcp__trinity__` tool-name distinction (per architecture.md multi-runtime notes)
- [ ] Kept short enough not to bloat the prompt / context budget
- [ ] No behavior change for agents whose templates already instruct reporting (additive)

## Technical Notes

`services/platform_prompt_service.py` (`get_platform_system_prompt`/`compose_system_prompt`). Reference the tool contract in `src/mcp-server/src/tools/reports.ts`. Keep it edition-agnostic (OSS-core).

### LINKED ISSUE #1536: Download & render reports in popular formats (Excel, PDF)
## Summary

Let users download a report rendered in the formats stakeholders actually consume — Excel (`.xlsx`) and PDF — from the Reports tab / fleet Reports view, in addition to the current on-screen JSON/card rendering.

## Context

Part of the Agent Reports epic. Reports (#918) are currently only viewable in-app (typed renderers in `components/reports/` by `display_hint`). A report of leads, KPIs, or a weekly summary often needs to leave the platform as a shareable artifact. Export should map the existing `display_hint` shapes to sensible documents: `table` → spreadsheet rows/columns; `kpi`/`markdown`/`timeline` → a formatted PDF.

## Acceptance Criteria

- [ ] Download a single report as `.xlsx` (tabular payloads → real cells, not a JSON dump)
- [ ] Download a single report as PDF (formatted per `display_hint`, not raw JSON)
- [ ] Export honors the existing access-control boundary (only accessible agents' reports)
- [ ] Graceful handling when a payload shape doesn't map cleanly to the target format (fallback, not a 500)
- [ ] Export path is streaming/size-aware (coordinate with the large-dataset issue)

## Technical Notes

New export endpoint(s) under the reports router (e.g. `GET /api/reports/{id}/export?format=xlsx|pdf`), plus a download affordance in `components/reports/`. Consider server-side generation (openpyxl / a PDF lib) vs client-side. OSS-core.

### LINKED ISSUE #1537: Make reporting handle large datasets (payload scale, pagination, streaming)
## Summary

Make the reporting pipeline hold up when a report carries a large dataset. Today the payload is a single JSON blob capped at 256 KB (413 over cap) stored in one TEXT column — fine for KPIs, but it fails for a report of thousands of rows.

## Context

Part of the Agent Reports epic. As agents report real tabular results (lead lists, scan results, transaction histories), the 256 KB cap and single-blob storage/transfer become the ceiling. This issue covers scaling the create, store, list, and render paths — without breaking the small-report common case.

## Acceptance Criteria

- [ ] Reports well beyond today's 256 KB cap are supported (raise/stream the cap, or chunk/paginate the payload) with a defined new ceiling
- [ ] List endpoints stay metadata-only (no payload) so the Reports tab is unaffected by large payloads (preserve the `ReportSummary` vs full-payload split)
- [ ] Detail/render path paginates or lazy-loads large tabular payloads instead of shipping the whole blob at once
- [ ] Create path validates/streams large bodies without pinning a backend coroutine or blowing memory
- [ ] Retention sweep still handles large rows efficiently (chunked delete already exists)
- [ ] Backward compatible with existing small reports

## Technical Notes

`db/reports.py` storage model (consider off-row / paginated storage for tabular payloads), `services/report_service.py`, `routers/reports.py` (`REPORT_PAYLOAD_MAX_BYTES`). Dual-track migration if the schema changes (SQLite `db/migrations.py` + Alembic). Coordinate with the export issue (streaming). OSS-core.

### LINKED ISSUE #1538: Let agents read back their old reports via MCP
## Summary

Give agents a read path over their own report history via MCP, so an agent can see what it previously reported — to avoid duplicate/contradictory reports, resume a series, or diff against last period. Today the `report` tool is write-only from the agent's side.

## Context

Part of the Agent Reports epic. Agents publish reports but can't read them back through MCP, so they have no memory of their own reporting history within a stateless run. A scoped read tool (self + accessible agents, per existing access rules) closes that loop and mirrors the read surface already exposed to the UI.

## Acceptance Criteria

- [ ] New MCP tool(s) to list an agent's reports (metadata) and fetch a report's payload by id
- [ ] Self-scoped by default via the agent's auth context; agent-scoped keys gated to `{self} ∪ permitted` (mirror the operator-queue/read-tool pattern)
- [ ] List returns summaries (type/title/created_at/period), detail returns the payload — same split as the REST surface
- [ ] Filterable by `report_type` and time window at minimum
- [ ] Reuses the existing access-controlled REST endpoints (no new tenant-boundary logic in the MCP layer)

## Technical Notes

Extend `src/mcp-server/src/tools/reports.ts` with read tools over `GET /api/reports` / `GET /api/reports/{id}` (already access-controlled). Three-surface sync (Invariant #13). Pairs naturally with the search/filter issue. OSS-core.

### LINKED ISSUE #1539: Search & filter reports (fleet + per-agent)
## Summary

Let users search and filter the accumulating report history — by `report_type`, agent, time window, and free-text over title (and ideally payload) — across the fleet Reports view and the per-agent Reports tab.

## Context

Part of the Agent Reports epic. As reports accumulate (retention default 90 days), the flat list becomes hard to navigate. Search/filter turns the Reports surface into something you can actually mine — "show me all `prospector.leads_found` from last week", "find the report that mentioned X".

## Acceptance Criteria

- [ ] Filter reports by `report_type`, `agent_name`, and time window
- [ ] Free-text search over title (payload full-text search is a stretch goal)
- [ ] Works on both the fleet Reports view and the per-agent Reports tab
- [ ] Respects the access boundary (`accessible_agent_names` / admin = all)
- [ ] Backed by indexed queries (existing `idx_agent_reports_type` / `_created`; add indexes as needed) — no full-table scan on the hot path

## Technical Notes

Extend `GET /api/reports` (`routers/reports.py` + `db/reports.py`) with filter/search params; wire the frontend store + `components/reports/` list UI. Consider FTS for payload search as a follow-up. Coordinate with the MCP read-back issue (shared query params). OSS-core.

dolho:
Epic #1534, five children in one PR. Landing them together is deliberate — see **Scope** below.

Closes #1535
Closes #1536
Closes #1537
Closes #1538
Closes #1539

## What

| # | Child | Change |
|---|---|---|
| #1535 | Prompt guidance | Teach agents the `report` tool in the platform prompt: when a result belongs in a report rather than chat, the five payload shapes, and read-back-before-refiling |
| #1538 | Read-back | MCP `list_reports` / `get_report` — reports stop being write-only from the agent's side |
| #1539 | Per-agent filters | `hours` + `search` on the per-agent list (the fleet list had them since #918) |
| #1537 | Scale | `REPORT_PAYLOAD_MAX_BYTES` 256 KiB → 5 MiB + `GET /api/reports/{id}/rows` paginated row reader |
| #1536 | Export | `GET /api/reports/{id}/export?format=xlsx\|pdf` via a new `services/report_export.py` |

## Scope

**This is five issues, not one.** The earlier description claimed prompt-text only, which was wrong and is corrected here — a reviewer prepared for ~1.3 KB of prompt text was instead handed a new service module and two endpoints, which is how the auth-wiring gate below slipped past a human read.

They ship together because they are one coherent epic slice: #1535 tells agents to publish reports, and #1537/#1538/#1539/#1536 are what makes that instruction honest (a payload big enough to matter, readable back, findable, exportable). Each commit is clean and independently revertible if that judgement turns out wrong.

The `+18% / prompt-only` framing from the previous body is withdrawn — that is not the trade being made.

## Review round 1 — @AndriiPasternak31

**1. CI red on a real regression** ✅ fixed in `a02a9f59`

`test_1310_auth_wiring::test_no_inline_auth_gates_in_routers` flagged `export_report` and `get_report_rows`. The uniform-404 design was right; the wiring was three byte-identical copies of the gate — the duplication the guard proxies for. Took the suggested shape: one `_report_or_404(report_id, current_user)`, all three routes call it, allowlist is one entry instead of three. The allowlist meta-test moved with it (its synthetic function must carry the allowlisted name or it stops asserting anything). Green on seeds 1/2/3.

**2. Prompt ships a 20×-wrong ceiling** ✅ fixed in `a02a9f59`

Correct, and correct about the consequence — it partly cancels #1537. Took the stronger option: the block now **interpolates** `REPORT_PAYLOAD_MAX_BYTES` instead of restating it, so it cannot drift again, plus the seventh drift guard (asserts the figure matches, asserts `256 KB` is absent).

Sweeping for the two stale doc echoes turned up a **third live agent-facing surface** carrying the same number, which wasn't in the review: the MCP `report` tool's own `payload` description (`tools/reports.ts:106`). A tool description is read by the model exactly as the prompt is, so it was the same bug in the same class. Fixed. Docs updated where they state *current* behaviour (architecture.md ×3, agent-reports.md ×3, lifecycle-observability.md ×2); the before/after table, the 201-byte measurement narrative and `models.py`'s "raised from 256 KiB" are history and stay.

**3. Description was for a different PR** ✅ fixed — this body

Kept as one PR. Title and body now say five issues plainly and the prompt-only framing is gone.

**4. `wip:` title, no closing keyword** ✅ fixed — title updated, five `Closes` above

**5. `request: Request = None` → `Optional[Request]`** ❌ **not applied — the suggestion is wrong**

FastAPI special-cases the **bare** `Request` annotation as an ASGI injection. Wrapping it in `Optional` loses that, so FastAPI tries to build a Pydantic field for it and the module fails to import:

```
fastapi.exceptions.FastAPIError: Invalid args for response field!
Hint: check that typing.Optional[starlette.requests.Request] is a valid Pydantic field type.
```

Verified by making the change and running the suite — **9 tests failed** on module import. Reverted, with a comment recording why so it isn't re-suggested. The annotation is a framework contract here, not a nullability claim.

## Verification

- `test_1310_auth_wiring` — 12 passed (was 1 failed), seeds 1/2/3
- Reports suites — `test_1535` / `test_1536` / `test_1537` / `test_1539` / `test_918_report_broadcast` + auth wiring: **34 passed, 1 skipped**
- `npx tsc --noEmit` on the MCP server: clean


dolho:

```
4996 passed, 16 skipped, 4 failed  (15:33)

```

All 4 are pre-existing and environmental, not from this change — the first is `ModuleNotFoundError: No module named 'tzdata'`, and all 4 reproduce identically on an untouched checkout at the same base commit (same set observed on #1837's branch).

One honesty note on sequencing: that 15-minute run started just before a whitespace-only tweak to the code example in the prompt block (indenting the call arguments independently of the call name, so stripping the Codex prefix doesn't misalign them). The five suites that actually read `PLATFORM_INSTRUCTIONS` were re-run against the committed HEAD afterwards — 80 passed.

dolho:
## Added: child 2 of the epic — agent read-back (#1538)

Kept on this branch rather than a second PR, per the "same PR" call. They're one behaviour: the prompt block tells an agent to publish, and this tells it what it already published. Shipping them apart would have meant a prompt that says *"continue the series"* pointing at a write-only surface.

### What

`list_reports` (metadata; filters `agent_name`/`report_type`/`hours`/`search`, paged) and `get_report` (payload by id), proxying the **existing** access-controlled endpoints. No new endpoint, no new table, no new tenant-boundary logic.

### The gate the backend can't apply

An agent-scoped key resolves to its **owner**, so the backend scopes a read to everything the owner can see — wider than the calling agent's permits. The tools narrow a broad listing to `{self} ∪ permitted` (the #1104 rule `list_operator_queue` established) and re-check the owning agent on `get_report`.

One deliberate detail worth reviewing: **a denied `get_report` returns the backend's own `Report not found` shape**, not a distinguishable 403. `GET /api/reports/{id}` answers 404 precisely so an id can't be probed for existence; returning "exists but forbidden" for agent keys would undo that choice one layer up.

Write is untouched and stays self-gated — reading another agent's reports never widens what you can write.

### Prompt

The #1535 block gains one sentence pointing at read-back, and the Codex orientation lists both new tool names so the prefix-stripped variant stays complete. Block is now **1600 chars**, still under the 2000 cap the #1535 test pins.

### Verification

```
node --import tsx --test src/tools/reports.test.ts
# tests 9   # pass 9   # fail 0
```

Nine cases drive the real `execute()` against a fake client: the pure filter, broad-listing narrowing for an agent key, **no** narrowing for a user key (with `getPermittedAgents` throwing, to prove it isn't consulted), scoped denial, self-listing, route selection (per-agent vs fleet), and the not-found-shape check on a denied `get_report`. Plus `tsc --noEmit` clean, and 14 passed across the prompt suites.

Driven live against a running instance:

```
list_reports  -> count: 1 | {"id":"0bff81d2…","agent_name":"report-probe-1535","report_type":"agent.weekly_status",…}
summary has no payload field: true
get_report    -> payload keys: [ 'markdown' ]
```

### One edge worth knowing (pre-existing, not from this PR)

That live report belongs to an agent I'd since deleted. Reports are KEEP-policy, so the row outlives the agent: the **fleet** list happily returns it, while `GET /api/agents/{name}/reports` answers `404 Agent not found`. So a caller can see a report in a broad listing and get a 404 scoping to its agent. It's #918 + soft-delete interacting, not something these tools introduce — flagging it rather than quietly widening the tool to paper over it. Happy to file it if you want it tracked.

### Docs

architecture.md (MCP catalog row 1→3 tools + a read-back bullet on the Agent Reports block), requirements FR-8, `feature-flows/agent-reports.md`, feature-flows changelog.

---

Epic #1534 status: children 1 (#1535) and 2 (#1538) are here. Remaining: search/filter (#1539), export (#1536), scale (#1537).

The `wip:` prefix on the title is yours — leaving it as-is; drop it when you're happy with the scope.

AndriiPasternak31 (changes requested):
## `/validate-pr` — Request changes

Two blockers, both concrete and cheap to fix, plus a scope question that decides how this gets reviewed at all.

First the credit, because it's earned: the docs on this branch are the best I've read in a while — FR-7 through FR-11 in `requirements/lifecycle-observability.md`, architecture.md updated per-surface, the feature flow refreshed, and the rationale is *measured* (201-byte average payload as the argument for 5 MiB rather than a vibe). The cross-surface drift guards in `test_1535_report_prompt_guidance.py` are exactly the right instinct: pinning the prompt's shape table to the MCP `z.enum` and to the renderer's dispatch keys is how you stop a silent-fallback bug. Security scan is clean, the Dockerfile pins are mirrored into `tests/requirements-test.txt`, lazy imports → 503-with-rebuild-hint instead of a router-import crash is the right call for #1814, and the MCP read gate correctly mirrors the #1104 `{self} ∪ permitted` rule.

---

### 1. ❌ CI is red on a real regression, not a flake

`test_1310_auth_wiring::test_no_inline_auth_gates_in_routers` fails on HEAD under **all three** seeds and passes on BASE under all three. Deterministic. Reproduced against the head file with the guard's own AST predicates:

```
INV-8 inline gate -> export_report()     reports.py:235
INV-8 inline gate -> get_report_rows()   reports.py:304
INV-8 inline gate -> get_report()        reports.py:330   # allowlisted
```

The two new routes use the `if not report or not db.can_user_access_agent(...): raise HTTPException(404)` shape, which is the exact pattern Invariant #8's guard exists to catch. Only `("reports.py", "get_report")` is allowlisted.

To be clear — **the design is right**: uniform-404 so an export URL can't become the id oracle the detail route refuses to be. The guard is complaining about the *wiring*, not the intent. Cleanest fix is one private helper rather than three allowlist entries:

```python
def _report_or_404(report_id: str, current_user: User) -> dict:
    """Fetch + gate. 404 (not 403) so a report id stays unprobeable."""
    report = db.get_report(report_id)
    if not report or not db.can_user_access_agent(current_user.username, report["agent_name"]):
        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Report not found")
    return report
```

All three routes call it, the allowlist becomes `("reports.py", "_report_or_404")`, and the duplicated gate collapses from three copies to one — which is the actual risk the guard is proxying for.

### 2. ❌ The prompt block ships a number that is wrong on merge

`platform_prompt_service.py:63`

```
payload={...},    # max 256 KB serialized
```

`models.py:358`

```python
REPORT_PAYLOAD_MAX_BYTES = 5 * 1024 * 1024  # 5 MiB
```

Same PR. So on merge every agent, on every turn, is told the ceiling is **20× below** what the platform actually accepts.

This isn't cosmetic — it partly cancels #1537. The stated purpose of the raise is "the wall the first real table would hit"; an agent that believes the wall is at 256 KB will pre-aggregate a 1 MB lead list down before publishing, and the raise buys nothing. It's also precisely the silent-divergence class this PR's own test file was written to defend against: six drift guards, and none of them pins the size.

Fix: correct the number, and add the seventh guard — assert the byte figure in the block matches `REPORT_PAYLOAD_MAX_BYTES`. Interpolating the constant into the block is even better, since then it can't drift again.

Two stale echoes of the same number to sweep while you're there: architecture.md (`unindexed LIKE over a 256 KB blob`) and requirements FR-9 (`a LIKE over a 256 KB TEXT blob`).

### 3. ⚠️ The description is for a different PR than the branch

The body says:

> **Not in scope** — Reports remain write-only from the agent's side (#1538), unsearchable (#1539), un-exportable (#1536) and capped at 256 KB (#1537) — the other four epic children.

The branch contains all four. Seven commits, 25 files, +2278/−35: a new `services/report_export.py` (238 lines), two new REST endpoints, two new MCP tools, the cap raise, five frontend files, three additional test modules. `report_export.py` does not exist on `dev` — none of this is inherited from the merge commit.

A reviewer reading that description prepares to review ~1.3 KB of prompt text and is instead handed a new service module and two new endpoints. That's how the auth-wiring gate in #1 slips past a human: it's in the half of the PR the description says isn't there.

The docs already discuss all five as separate FRs and each commit is clean and independently revertible, so splitting into the five PRs the epic already has issues for is close to free — `git cherry-pick` per commit onto five branches. If you'd rather land it as one, that's a legitimate call given they're one coherent epic slice, but then the title and body need to say so, and #1535's "prompt-only, 1.3 KB, +18%" framing has to go — the trade being made is no longer the one the body argues for.

My preference is the split: #1535 is a prompt-text change whose whole review is "are these five shapes right", while #1537 is a 20× storage-ceiling decision. Those want different reviewers and different scrutiny.

### 4. ⚠️ Title is still `wip:` and there's no closing keyword

`Related to #1535` is a bare reference. `.github/workflows/issue-status-on-merge.yml` only promotes on `Fixes` / `Closes` / `Resolves`, so on merge all five issues strand in `status-in-progress` and get swept up by the next `/groom`. Needs `Closes #1535` (plus the others, or one per PR if split). And drop the `wip:` — the PR is marked ready for review.

### 5. Minor

`routers/reports.py:62` — `request: Request = None` should be `Optional[Request]`. FastAPI injects it regardless so the behaviour is fine, but the annotation currently says a non-optional type has a `None` default.

---

### Validation summary

| Category | Status | Notes |
|---|---|---|
| Commit messages | ✅ | conventional, one per issue, clean |
| PR size | ⚠️ | 25 files / +2278 — five issues in one PR (#3) |
| Issue link | ⚠️ | bare `#1535`, no closing keyword (#4) |
| Requirements | ✅ | FR-7…FR-11, thorough |
| Architecture | ✅ | per-surface, incl. the drift-pinning rationale |
| Feature flows | ✅ | `agent-reports.md` + index |
| Security scan | ✅ | no secrets / emails / IPs / creds |
| Build packaging | ✅ | new module under `services/` (copied wholesale); pins added to Dockerfile + test reqs |
| Config packaging | ➖ | no new `os.getenv` |
| Test adequacy | ✅ | 4 new modules + MCP unit tests; non-happy-path covered (413, 400, 503, 404) |
| **CI** | ❌ | **1 deterministic new failure across 3 seeds (#1)** |
| **Cross-surface consistency** | ❌ | **prompt says 256 KB, code says 5 MiB (#2)** |

**Blocking:** #1 and #2. **Needs a decision before re-review:** #3.

Happy to re-review as soon as the guard is green — and if you split, ping me and I'll take the five in sequence.

dolho:
@AndriiPasternak31 — both blockers fixed in `a02a9f59`, scope and title corrected in the body. Thanks for the read; #2 in particular was a real cancellation of #1537, not a typo.

**1 — auth wiring.** Took your `_report_or_404` shape verbatim. Three byte-identical copies of the gate was the actual risk, and the guard was right to complain about wiring rather than intent. Allowlist is one entry now. One thing worth flagging for anyone touching that guard later: the allowlist meta-test `test_allowlist_suppresses_intentional_404` uses the allowlisted function *name* in its synthetic source, so it has to move with the rename or it silently stops testing the allowlist. Green on seeds 1/2/3.

**2 — the ceiling.** Interpolated rather than corrected, per your "even better" — the block substitutes `REPORT_PAYLOAD_MAX_BYTES` at import, so a literal can't come back. Seventh guard added.

Sweeping your two stale doc echoes turned up **a third live surface with the same number that wasn't in the review**: the MCP `report` tool's own `payload` description (`tools/reports.ts:106`, "max 256 KB serialized"). Same class as the prompt bug — a tool description is read by the model exactly as the prompt is — so the fix was incomplete without it. Fixed.

**3 — scope.** Keeping it as one PR; the body now says five issues plainly and the "prompt-only, 1.3 KB, +18%" framing is withdrawn. Your point stands that the old description is how the auth gate slipped a human read, and that was the real damage. If you'd still rather take them in sequence, say so and I'll split — the commits are one-per-issue and cherry-pick cleanly.

**5 — not applied, and I think the suggestion is wrong.** `Optional[Request]` breaks FastAPI: the **bare** `Request` annotation is special-cased as an ASGI injection, and wrapping it makes FastAPI build a Pydantic field for it, so the module fails at import:

```
fastapi.exceptions.FastAPIError: Invalid args for response field!
Hint: check that typing.Optional[starlette.requests.Request] is a valid Pydantic field type.
```

I made the change before rejecting it — 9 tests fail on import. Reverted with a comment recording why, so it doesn't get re-suggested. Happy to be shown otherwise if you've seen it work.

Ready for re-review.

AndriiPasternak31 (approved):
Approving. Round-1 findings are all resolved, and #5 was correctly rejected — the bare `Request` annotation is a FastAPI ASGI-injection contract, and the comment recording that is the right outcome so it isn't re-suggested.

Spot-checking the invariants I expected this to break, it doesn't: `_fleet_conditions` is shared by both list routes (Inv #1/#2), the window uses `iso_cutoff()` rather than `datetime('now')` (Inv #16), `/reports/stats` still precedes `/reports/{report_id}` (Inv #4), the Codex prompt strip is a global `.replace("mcp__trinity__", "")` so the new block is covered for free, and the export download goes through the shared `api` client with a static guard pinning it (Inv #7). The `_report_or_404` consolidation is the right answer to the auth-wiring guard — one gate, one allowlist entry, uniform 404 preserved. Non-happy-path coverage is well above average: 404-not-403, 503-not-500, unparseable Content-Length, offset-past-end, non-tabular 400, reportlab markup escaping, MCP fail-closed.

## Follow-up (not blocking this PR)

`export_report` runs the builders on the event loop — `routers/reports.py:268` and `:271` call `build_xlsx`/`build_pdf` synchronously inside an `async def` handler, so they execute on the loop rather than in FastAPI's threadpool.

I measured it against the real builders at the new ceiling rather than guessing:

| payload | builder | wall time |
|---|---|---|
| 5.10 MiB / 72,000 rows | `build_xlsx` | **1.57 s** |
| same, capped at `PDF_MAX_ROWS` | `build_pdf` | 0.24 s |

The PDF cap does its job; the xlsx path is the uncapped one. On container CPU that is plausibly 3-5 s, and `/export` has no rate limit (`rate_limiter.enforce` is only in `create_report`), so concurrent exports serialize into N x that stall on the worker — blocking heartbeat processing, WS dispatch and every other request on it.

Not blocking, because the expensive input does not exist yet: the fleet's reports average 201 bytes, which is precisely why #1537 raised the cap. This is a latent characteristic of a path whose worst-case payload nobody has published. Fix is one line when someone picks it up:

```python
content = await asyncio.to_thread(report_export.build_xlsx, ...)
```

## Minor

- `docs/memory/feature-flows.md` — the index row credits only #1535/#1538 ("children 1-2"); #1536/#1537/#1539 ship in the same PR and aren't in it.
- #1536 carries only `status-in-progress` — no priority or type label, so it's invisible to `/roadmap` ordering.
- `db/reports.py` — `like = f"%{search}%"` passes `%`/`_` through as wildcards. Parameterized so not injection, and pre-existing from #918, but worth escaping while the code is open.

On the five-issues-in-one-PR question: keeping it as one PR is the right call given the children only make sense together, and the body now says so plainly. The commits are cleanly separable if that judgement turns out wrong.

PR #1836, opened 2026-07-28
fix(ui): elastic Dashboard stats bar — degrade telemetry instead of clipping it (#1830)

dolho:
## What

Makes the Dashboard header row elastic. The stats cluster becomes a size container and its contents degrade in a defined order as the leftover width shrinks, instead of being clipped mid-element under the controls cluster.

Related to #1830

## The bug

Row is `[stats cluster | controls cluster]`: the left side was `min-w-0 overflow-hidden` (clip-only, no `flex-1`), the right side `flex-shrink-0`. Once the content stopped fitting, nothing gave — the telemetry meters were cut mid-element and their boxes ran under Tags / owner / Grid-Timeline / time-range.

Measured on a live stack (Playwright, timeline mode, 1-agent fleet — the collision threshold moves up on fleets with the owner filter and more tags, which is the ~1170px reported in the issue):

| width | overlap | clipped mid-element |
|---|---|---|
| 900px | **22px** (`Disk 33%` ⨯ `Timeline` button) | `stat-label` (the "Disk" word, rendered as "D") |
| 1024px+ | 0 | – |

## The fix

- `.stats-cluster` gets `flex-1 min-w-0` + `container-type: inline-size` (name `statsbar`) — it is now the only cluster that gives ground, and its leftover width is the query axis.
- Ladder, widest → narrowest: sparklines → Disk → Mem → whole telemetry → `messages (Nh)` → `working now`. The agent count always survives.
- Keyed on the **container**, not the viewport — the controls cluster is not a constant (grid mode adds Tidy up / Reset; Tags chip and owner filter only render on some fleets), so a viewport breakpoint would be right on one fleet and wrong on the next.
- Thresholds sit ~25–30px above the **measured** intrinsic width of the level each one gates (full 791 → no-sparks 663 → no-disk 522 → no-mem 394 → no-telemetry 308 → no-messages 182 → agents-only 71), leaving headroom for wider live values (three-digit fleets, `100.0/128G`). Setting one below its level reintroduces the clip — noted in the code comment.
- `HostTelemetry` owns its leading `·` separator now (so it leaves with the meters) and each metric carries a `title`, so the value stays reachable once its meter is hidden.

No JS, no ResizeObserver, no new dependency — `@container` is plain CSS through Vite (Chrome 105+ / Safari 16+ / Firefox 110+).

## Acceptance criteria

[...]

PR #1839, opened 2026-07-28
fix: promote post-v0.8.5 bug fixes to main (#1795, #1830, #1790, #1793, #1759)

### LINKED ISSUE #1759: bug: absent local: template silently creates a blank agent (no error)

### LINKED ISSUE #1790: Cornelius seeder marks itself seeded on a volume-conflict 409 — Cornelius never created, never retried

### LINKED ISSUE #1793: Nonexistent local: template silently creates an empty agent (200) instead of failing — github: path correctly 400s

### LINKED ISSUE #1795: Agent base image hardcodes TZ=America/New_York — agents run 4h off UTC and file timestamps surface as naive local time

### LINKED ISSUE #1819: bug: rename_agent misses 8 agent_name tables (agent_sessions, portal sessions/messages, Slack bindings, A2A) — history stranded under the old slug
## Summary

`rename_agent` re-keys `agent_name` across ~20 tables but **omits `agent_sessions` and `agent_session_messages`**. Renaming an agent therefore strands its entire Session-tab history under the old slug: the rows survive, but nothing can reach them, because the old agent name no longer exists.

Found while testing the v0.8.5 display-label / rename feature on a real instance.

## Repro

Agent `up-alpha` with one Session-tab session (4 messages), then `PUT /api/agents/up-alpha/rename` → `up-alpha-renamed`.

Every other table follows the rename; these two do not:

```
table                       old name   new name
agent_sessions              1          0     <-- stranded
agent_session_messages      4          0     <-- stranded
chat_sessions               0          1
chat_messages               0          15
agent_schedules             0          1
schedule_executions         0          13
agent_activities            0          13
```

User-visible result — the session that worked minutes earlier:

```
POST /api/agents/up-alpha-renamed/sessions/{id}/message
→ "Session not found"
```

The rename itself reports success and warns only that a restart is needed. Nothing mentions session history.

## Root cause

`db/agent_settings/metadata.py::rename_agent` is a hand-written sequence of explicit `update(<table>)` statements, plus `EXTRA_AGENT_REFS` (`db/agent_cleanup.py`) for entitled-module tables. `agent_sessions` and `agent_session_messages` appear in **neither** list — an omission, not a decision. There is no shared registry the two paths derive from, so nothing failed when the Session tab shipped its tables.

Note the code already articulates exactly why this matters, in the comment on the reports re-key immediately above:

```python
# Reports (#918) — re-key so the renamed agent keeps its report
# history and the old name doesn't leave orphaned rows a reused
# name could inherit (cross-tenant disclosure).
```

Both halves of that reasoning apply verbatim to the session tables.

## Impact

**Primary (confirmed):** the renaming user silently loses access to their own Session-tab history for that agent — including `--resume` continuity, which is the tab's whole point. The data is not destroyed (rows remain under the old slug) and renaming back would restore access, but nothing tells the user any of that.

**Secondary (not confirmed — flagging, not asserting):** the "orphaned rows a reused name could inherit" concern the reports comment raises. `agent_sessions` rows carry `user_id`/`user_email` and message content. If the old slug is later reused by a different owner's agent, the stranded rows sit under that name. Per-user filtering on the list endpoints should prevent another *user* seeing them, but the interaction with admin-visible surfaces and with `session_cleanup_service` (which decides which JSONLs to reap by comparing `agent_sessions.cached_claude_session_id` for an agent name) is worth someone checking properly.

## Suggested fix

Add both tables to the rename cascade. The durable fix is to derive rename and delete-cascade from **one** registry rather than two hand-maintained lists — `AGENT_REFS`/`EXTRA_AGENT_REFS` already exist for cleanup, and `tests/unit/test_agent_cleanup_parity.py` is CI-blocking for the cleanup side; an equivalent parity test for rename would have caught this and will catch the next table.

## Environment

v0.8.5 code (`b8cf94ca`), dev compose, SQLite. Rename via `PUT /api/agents/{name}/rename`.

## Verified working in the same test

Recording these so the feature isn't assumed broken generally — everything else the release notes promise for labels and renames held:

- **Display label**: set/read correctly, and all four advertised invariants survived it — slug route, MCP tool name (`chat_with_up_alpha`), the existing schedule, and a live webhook token (202).
- **Rename**: old slug 404s, new slug 200s, schedules/chat/executions/activities all re-keyed, webhook token still fires (202), MCP tool name follows (`chat_with_up_alpha_renamed`), the workspace volume is correctly preserved under the pre-rename base with `volume_base_name` pinned to `up-alpha` (#1664 working as designed), the display label survives, and chat works after the documented restart.

### LINKED ISSUE #1830: bug: Dashboard stats bar — right controls overlap host telemetry below ~1170px effective width
## Summary

On the Dashboard, the compact header row (agents / working now / messages / CPU / Mem / Disk telemetry on the left; Tags filter, owner filter, Grid/Timeline toggle, time-range selector on the right) does not degrade gracefully at narrow effective viewport widths. Below ~1170 CSS px the right-hand controls visually overlap the host-telemetry meters — at 900 px the Tags chip sits on top of the Disk meter with a measured 29 px overlap. The row has no wrap, scroll, or progressive-hide behavior, so the two clusters collide instead of compressing.

This is the same "inelastic flex row" defect class as the NavBar overlap fixed in #1789 (PR #1794), one row lower. It is pre-existing — present both before and after the #1794 fix (that PR touched only `NavBar.vue`).

Note that "narrow effective width" does not require a small screen: a wide monitor at 150–175 % browser zoom lands in exactly this range, which is how it was observed.

## Context

Found while validating the #1789 fix on an entitled build. A Playwright width sweep of the Dashboard (900 / 1024 / 1170 / 1366 / 2000 CSS px) confirmed the nav row is fixed, but surfaced this sibling overlap in the stats bar:

- 1366 px: clean
- ≤ ~1170 px: Tags chip begins crowding/overlapping the Disk meter
- 1024 px: Tags chip visibly renders on top of the Disk meter bar
- 900 px: 29 px measured overlap (Disk parent box vs Tags button box); `document.scrollWidth == viewport`, so the overflow is invisible rather than scrollable

Structure (`src/frontend/src/views/Dashboard.vue`, header row): the left stats cluster is `flex items-center min-w-0 overflow-hidden` wrapping a `whitespace-nowrap` span group that includes `HostTelemetry`; the right controls cluster is `flex items-center space-x-2 flex-shrink-0`. Despite the left cluster's `overflow-hidden`, the telemetry meters still paint under the right cluster at narrow widths — worth checking how `src/frontend/src/components/HostTelemetry.vue` renders its meter bars (fixed-width bars / absolutely positioned fills may escape the clip or the clip boundary itself may land mid-meter with no visual affordance).

## Acceptance Criteria

- [ ] At every viewport width from 640 px up, no element of the right controls cluster (Tags, owner filter, Grid/Timeline, time range, refresh) overlaps any element of the left stats cluster
- [ ] Telemetry degrades gracefully as width shrinks (e.g. progressively hide Disk → Mem → CPU meters or collapse to compact values) instead of being clipped mid-meter with no affordance
- [ ] No page-level horizontal overflow introduced (`document.scrollWidth == viewport` stays true)
- [ ] Playwright e2e width sweep covering the stats bar, following the pattern of `src/frontend/e2e/navbar-overflow.spec.js` (#1789), pinning zero link/control–telemetry collision at the swept widths
- [ ] Both Grid and Timeline dashboard modes covered (the row is shared)

## Technical Notes

- Reference fix for the same defect class: #1789 / PR #1794 (`min-w-0` + elastic overflow strategy on the nav row; e2e spec `navbar-overflow.spec.js`)
- Relevant files: `src/frontend/src/views/Dashboard.vue` (header row markup), `src/frontend/src/components/HostTelemetry.vue` (meter rendering)
- On fleets with more stats populated (owner filter present, larger message counts) the left cluster is wider, so the collision threshold moves up accordingly — reproduce with the owner dropdown visible for the worst case

vybe:
## What

Promotes the post-v0.8.5 bug fixes validated and merged to `dev` today onto `main`, following the established hotfix pattern (#1801/#1802). Cherry-picks of the squashed dev commits, in merge order:

| Fix | Issue | dev PR | Commit |
|---|---|---|---|
| Agent base image defaults to UTC + ISO-Z file timestamps | Closes #1795 | #1828 | d6b8557f |
| Elastic Dashboard stats bar — degrade telemetry instead of clipping | Closes #1830 | #1836 | cac5f91e |
| Cornelius seed flag converges on 409 only when the agent exists | Closes #1790 | #1837 | 1d919317 |
| Reject an unresolvable `local:` template instead of creating an empty agent | Closes #1793 | #1803 | 46ccaf0b |
| Reject an invalid `local:` template + ship the missing `local:default` | Closes #1759 | #1824 | 86f8493e |
| Rename cascade derived from `AGENT_REFS` — rename no longer strands 23 tables (Session-tab history, reminders, channel bindings) | Related to #1819 (rename half only — the dark-mode half ships with ent#170) | #1851 | 8efe943c |

#1803 is included because #1759/#1824 builds directly on its contract (the two halves of the same defect — see the learnings entry this PR carries). #1851 is a P1 found on a real v0.8.5 instance exercising the rename feature that is already on `main`; deliberately **not** `Closes #1819` — the issue stays open for its dark-mode half.

## Conflicts resolved

First four picks applied conflict-free. The #1824 pick had two metadata conflicts, both from dev-only history that is deliberately **not** promoted:
- `tests/registry.json` — dropped the incoming `test_1809_image_drift_recreate.py` entry (that test ships with #1820, which stays on `dev`); kept the three #1759 entries. JSON validated.
- `docs/memory/learnings.md` — kept only #1824's two new entries (path-disclosure guard pitfall, duplicate-bug process entry); the #1775/#1809 entries stay with their unpromoted PRs.
- The #1851 pick applied conflict-free (2 files: `db/agent_settings/metadata.py` + its regression test; no registry.json entry to carry — parity with `dev`).

## Verification (local, against this branch)

- All promoted unit suites: `test_1759_*` ×3, `test_1793_unknown_local_template`, `test_1795_agent_timestamp_utc`, `test_ent107_cornelius_seed`, `test_local_templates_listing` — **86 passed**.
- `test_1484_create_agent_characterization.py` — **40 passed** (the interim 4-pick state showed a macOS-only tmpdir-symlink harness artifact identical to `origin/dev`; #1824's harness changes resolve it).
- Unfiltered conflict-marker grep across the tree: clean. No submodule pointer changes in any pick.
- #1851 pick: `test_1819_rename_cascade_parity.py` — **4 passed** against this branch (the parity test derives its expectations from `main`'s own `AGENT_REFS`, so it validates the cascade against the promoted schema, not `dev`'s); adjacent `test_agent_cleanup_parity.py` + `test_1664_renamed_agent_volume_safety.py` — **42 passed**. Conflict-marker grep re-run after the pick: clean.

## Release note ordering

#1828 changes the agent **base image** — deployments pick it up after the next `build-base-image.sh` run. The auto-adoption path (#1809/#1820, cold-start image-drift recreate) is on `dev` only, so on `main` builds agents pick up the rebuilt image on their next recreate.


vybe:
Validated via /validate-pr: all five cherry-picks match their squashed dev commits, conflict resolutions check out (dev-only registry/learnings entries correctly excluded), security greps clean, required checks green, and HOST_TEMPLATES_PATH wiring already exists on main so the config-packaging check passes. Blocked only on branch protection — one approving review needed (@dolho @AndriiPasternak31), then this merges clean with a merge commit preserving the five picks.

vybe:
Added the #1851 pick (rename-cascade fix, #1819 rename half) — a P1 found on a real v0.8.5 instance, so it belongs in this promotion. Applied conflict-free; `test_1819_rename_cascade_parity` (4) + `test_agent_cleanup_parity`/`test_1664` (42) pass against this branch. @AndriiPasternak31 @dolho — re-review covers 6 picks now.

AndriiPasternak31 (approved):
Approved as a **promotion PR** — verified the content is already-reviewed work, not new code.

Verification performed:
- **Patch-id comparison against the `dev` counterparts**: 5 of 6 commits are byte-identical cherry-picks — #1795/`70adb690`, #1830/`6bf208d9`, #1790/`ce9ce328`, #1793/`acf699c0`, #1819/`8efe943c`.
- The one divergence (#1759, `86f8493e` vs `e54ce3a9`) is confined to `tests/registry.json` — the test catalog differs only because `main` lacks the #1809/#843 entries present on `dev`. **All source files are byte-identical** (`crud.py` +213, `system_service.py` ±33, `models.py` ±2 in both).
- 22/22 substantive checks green, including all three `pytest (head)` seeds.

Two notes for the record, neither blocking:
1. **Title undercounts the contents** — it lists 5 issues but the branch carries 6 commits; #1819 (P1 rename-cascade) is included and unlisted. Worth adding so release notes and issue-closing don't miss it.
2. **Base-image adoption gap** — this promotes the #1795 TZ change to `docker/base-image/Dockerfile` but not #1809 (`28ce2253`, adopt a rebuilt base image on cold start), which stays on `dev`. On `main`, existing agents will not pick up UTC even after a base-image rebuild — the same no-op that undercut the v0.8.5 upgrade step. Consider including #1809 in this promotion.

PR #1843, opened 2026-07-28
fix(context): read the answering model's window from modelUsage, not the first entry (#1840)

### LINKED ISSUE #1840: bug: Sonnet 5 / Fable 5 record a 200K context window on most runs — the safe floor wins whenever the runtime omits modelUsage.contextWindow
## Summary

An agent running **Claude Sonnet 5** (a 1M-context model) records `context_max = 200000` on most executions, and `1000000` on some — same agent, same model, same endpoint, consecutive runs. The context-usage percentage shown to the user is therefore overstated ~5× most of the time, and the Executions list is internally inconsistent row-to-row.

This contradicts the v0.8.6 release note: *"Context-usage now reflects each model's real window instead of a flat 200K."*

## Evidence

Five consecutive `POST /api/agents/evt-worker/task` calls, all `"model": "claude-sonnet-5"`, agent runtime `claude-code`, base image 0.8.0, nothing else changed:

| run | message | agent-reported `metadata.context_window` | stored `context_max` |
|-----|---------|------------------------------------------|----------------------|
| 1 | `S5-OK` | — (not captured) | **200000** |
| 2 | `S5-CTX` | **1000000** | **1000000** |
| 3 | `R1` | 200000 | 200000 |
| 4 | `R2` | 200000 | 200000 |
| 5 | `R3` | 200000 | 200000 |

Run 2's raw turn payload carried the runtime value:

```json
"modelUsage": {"claude-sonnet-5": {"contextWindow": 1000000, "maxOutputTokens": 64000, ...}}
```

and its metadata agreed (`context_window: 1000000`, `recovered_from_jsonl: false`). Runs 3–5 reported `200000` in metadata with `recovered_from_jsonl: false` as well — so this is not the JSONL-salvage path.

Resulting rows (`GET /api/executions?agent=evt-worker`):

```
claude-sonnet-5 | ctx_max 200000  | Reply with exactly: R3
claude-sonnet-5 | ctx_max 200000  | Reply with exactly: R2
claude-sonnet-5 | ctx_max 200000  | Reply with exactly: R1
claude-sonnet-5 | ctx_max 1000000 | Reply with exactly: S5-CTX
```

## Where it comes from

The backend is faithful — `task_execution_service.py:1822` writes `metadata.get("context_window") or DEFAULT_CONTEXT_WINDOW`, and the value it stores always matches what the agent reported. The flip originates **agent-side**: when the runtime's result event carries `modelUsage[...].contextWindow`, the real 1M lands; when it doesn't, the seed from `services/model_context.py::resolve_context_window` is what survives — and that floors every bare `claude-*` id at 200K:

```python
("claude", DEFAULT_CONTEXT_WINDOW),   # 200_000
("sonnet", DEFAULT_CONTEXT_WINDOW),
("fable",  DEFAULT_CONTEXT_WINDOW),
```

That floor is deliberate and correct for models with both a 200K and a 1M variant (`opus`, `sonnet-4-6`), where over-reporting would hide an imminent compaction wall. The module's own comment names the gap:

> Sonnet 5 / Fable 5 are unconditionally 1M (no 200K variant exists), but we still keep the uniform safe floor here and rely on the runtime value to show 1M for them

The measurement above shows the runtime value is present on a minority of turns, so "rely on the runtime value" does not hold in practice for exactly the two models the release headlines.

## Impact

- A Sonnet 5 / Fable 5 agent shows ~5× its true context usage (35,977 tokens rendered as 18% of 200K rather than 3.6% of 1M) in the Executions list, Overview `context_avg`, and anywhere the percentage is derived.
- The direction of the error is the one the safe-floor design was built to avoid *causing*: it makes a healthy agent look near its compaction wall, which invites needless model switches or context resets.
- Same agent, adjacent rows disagree — so the number reads as unreliable rather than conservative.

## Suggested direction

Give the two 1M-only ids their real window in the fallback catalog, keeping the floor for the genuinely ambiguous families:

```python
("claude-sonnet-5", EXTENDED_CONTEXT_WINDOW),
("claude-fable-5",  EXTENDED_CONTEXT_WINDOW),
("sonnet-5", EXTENDED_CONTEXT_WINDOW),
("fable-5",  EXTENDED_CONTEXT_WINDOW),
# ...then the existing bare-family floors
```

The floor exists for *unknown or dual-variant* tiers; a model with no 200K variant cannot be over-reported. Order matters — the specific ids must precede the `("claude", …)` / `("sonnet", …)` prefixes, which currently swallow them.

Worth deciding alongside: whether the agent should also emit a stable `context_window` per turn rather than only when `modelUsage` is populated, since the run-to-run flip is visible regardless of what the fallback returns.

## Repro

1. Create an agent, `PUT /api/agents/{name}/task` with `{"message": "Reply with exactly: OK", "model": "claude-sonnet-5"}`.
2. Repeat 4–5 times.
3. `GET /api/executions?agent={name}` — `context_max` alternates between `200000` and `1000000`.

dolho:
## What

Reads the context window of the model that **answered** the turn out of `modelUsage`, instead of whichever entry happened to come first.

Related to #1840

## The bug is not where the issue says it is

I filed #1840 blaming the fallback catalog and proposed giving Sonnet 5 / Fable 5 a 1M entry. That diagnosis was wrong — the raw payloads show the runtime reported 1M on **every** run; it was never read.

```
run   metadata.context_window   contextWindow values present in the result event
R1    200000                    ['200000', '1000000']
R2    200000                    ['200000', '1000000']
R3    200000                    ['200000', '1000000']
S5    1000000                   ['1000000']
```

`modelUsage` carries one entry **per model the turn touched**. Claude Code bills side work (tool-permission classification, title generation) to a cheap Haiku:

```json
{"claude-haiku-4-5-20251001": {"contextWindow":  200000, "inputTokens": 522},
 "claude-sonnet-5":           {"contextWindow": 1000000, "inputTokens": 2}}
```

and the parser took the first one:

```python
for _, model_data in model_usage.items():
    if "contextWindow" in model_data:
        metadata.context_window = model_data["contextWindow"]
    break  # Use first model found
```

Dict order is insertion order, and the side model is recorded first whenever one runs — so the reported window tracked *whether a side model happened to fire*, not the model in use. The single clean run was just the turn where none did. Second defect in the same five lines: the `break` is outside the `if`, so a first entry lacking `contextWindow` ends the scan before a usable one is reached. `jsonl_recovery.py` repeats the pattern on the #678 salvage path.

**Scope is wider than the title.** This was never Sonnet-5-specific — any turn with a side model reported the side model's window, so a 1M-configured Opus agent has been mis-reporting identically.

## Approach

`pick_context_window` lands in the vendored catalog module (re-synced byte-identically, Invariant #5) and is used at all three sites. It matches the entry to `metadata.model_name` — the id from the latest assistant message, which precedes the result event in both parsers — by exact key, then by the entry's own `canonicalModel`, then falls back to the sole entry when the turn is unambiguous.

**It returns `None` rather than guessing.** If the answering model isn't identifiable, the caller keeps its seeded catalog value. The tempting alternative — take the largest window — is precisely the direction #1521 decision A exists to prevent: a 1M side model beside a 200K main model would hide an imminent compaction wall. So the safe floor is untouched, no model id is promoted, and decision A stands as written. The module docstring gains a pointer explaining that reading the runtime value *correctly* is this helper's job.

## Verification

```
pytest tests/unit/test_model_context_catalog.py tests/unit/test_model_context_parity.py
53 passed
```

10 new cases, built from the verbatim live payload: side-model-first, order-independence, the larger-side-window guard (must **not** inflate a 200K main model), unidentifiable-model-keeps-the-seed, single-entry, the break-outside-`if` scan, plus direct unit coverage of the selection rule including junk input.

Reverted the parser to its pre-fix form with the tests in place — **3 failed**, including the exact regression and the dangerous-direction guard:

```
FAILED TestMultiModelUsageSelection::test_side_model_does_not_win
FAILED TestMultiModelUsageSelection::test_larger_side_window_never_inflates_a_smaller_main_model
FAILED TestMultiModelUsageSelection::test_entry_without_context_window_does_not_abort_the_scan
```

Full unit suite result posted below.

## Deployment note

The seed and the parser both live in the **agent image**. Existing agents keep the old behaviour until the base image is rebuilt (`./scripts/deploy/build-base-image.sh`) and their containers recreated — worth calling out in release notes, since the symptom is user-visible and the fix looks inert until then.

Fixes #1840


dolho:

```
5056 passed, 16 skipped, 4 failed  (14:17)
```

The 4 are the environment's pre-existing set (`ModuleNotFoundError: No module named 'tzdata'` in test_1472, plus 3 in test_admin_email_login) — they reproduce on an untouched checkout at the same base and appear identically on #1845/#1847. Worth noting the admin-email-login three trace to a `bcrypt has no attribute '__about__'` version mismatch that also shows in the backend boot log; that's an environment pin to fix separately, not a test defect.

vybe (approved):

Issue #1655, opened 2026-07-16
refactor(ci): validate every bundled template.yaml declares only what it ships


obasilakis:
## Summary

There is **no CI validation of `template.yaml` anywhere** (`grep` over `.github/workflows/*.yml` for `template.yaml` / `template_service` / `compatibility` returns zero hits). The only declaration-vs-file cross-checks that exist — compatibility checks `X-003` (skills) and `X-004` (mcp_servers) — run against a **live agent's workspace** at runtime and are advisory/non-blocking, so they cannot fail a build.

That gap is the root cause of #1646: the bundled Cornelius template was a vendored snapshot of `Abilityai/cornelius` that took `template.yaml` + `CLAUDE.md` + `.gitignore` but not `.claude/` or `resources/local-brain-search/`. It advertised ~24 skills, 9 sub-agents and a FAISS semantic tier the bundle could not serve, and **shipped through a release** (v0.8.0) with nobody noticing — on the agent that is auto-seeded for every fresh install (trinity-enterprise#107).

#1646 fixes that one instance and regression-tests it. This issue guards the **class**.

## Context

`tests/unit/test_1646_cornelius_template_claims.py` (from #1646) already encodes the rules, deliberately written as "every declaration resolves to a file" rather than pinning a list. It is scoped to the Cornelius bundle only. The work here is to generalize it across `config/agent-templates/*/` and wire it to a workflow.

Worth knowing: `skills`, `sub_agents` and `capabilities` have **no functional effect at creation** — `crud.py:566-609` reads only `type`, `resources`, `tools`, `credentials.mcp_servers`, `runtime`, `shared_folders`. They are display metadata (rendered by `InfoPanel.vue` from `/api/agents/{name}/info`, which echoes `template.yaml` verbatim and unvalidated), which is exactly why drift is invisible until a user clicks something.

## Suggested approach

A CI job over every `config/agent-templates/*/template.yaml` asserting:

- [ ] `skills:` → `.claude/skills/<name>/SKILL.md` exists in that bundle (the `X-003` rule, moved pre-merge)
- [ ] `sub_agents:` → `.claude/agents/<name>.md` exists — **no check exists for this today, at runtime or otherwise**, and `InfoPanel.vue:131-160,383-391` renders sub_agents as a *clickable delegation prompt*, so a phantom entry is a button that resolves to nothing
[...]

PR #1844, opened 2026-07-28
fix(cornelius): seed from the public upstream repo instead of a vendored snapshot (#1656)

### LINKED ISSUE #1656: bug: Cornelius seed vault's MOCs link to notes that don't ship — 22 of 23 wikilinks are dead
## Summary

The bundled Cornelius seed vault (`config/agent-templates/cornelius/Brain/`) ships **2 MOCs whose wikilinks point almost entirely at notes that are not in the bundle**:

| File | Wikilinks | Dead |
|---|---|---|
| `Brain/03-MOCs/MOC - Knowledge Base.md` | 10 | **10 (100%)** |
| `Brain/03-MOCs/MOC - Master Navigation.md` | 13 | **12** |

Dead targets include `[[MOC - Decision-Making and Judgment Under Uncertainty]]`, `[[MOC - Cognitive Biases and Behavioral Economics]]`, `[[MOC - Learning, Memory and Expertise]]` and 8 more — 11 MOCs that don't exist in the bundle.

`Brain/README.md` tells a brand-new user: **"Start here: Open [[MOC - Knowledge Base]] to explore the seeded notes"** — and every single link in that file is dead.

## Why this matters

Cornelius is auto-seeded on every fresh install (trinity-enterprise#107), so this is the out-of-box "explore your second brain" experience. The literal first instruction leads to a page of broken links.

## Root cause (same class as #1646)

The bundle is a vendored snapshot of the public `Abilityai/cornelius` repo. The prose and MOCs came across; the note corpus they describe did not. `Brain/` ships **16 `.md` files** (mostly READMEs and `_SAMPLE` templates) while the vault text described a ~1,000-note library. Identical to #1646, where `template.yaml`/`CLAUDE.md` were vendored but `.claude/` and `resources/local-brain-search/` were not.

#1646 fixed the *prose* claims (the removed "Vault Statistics" block, the slash-command references). It deliberately did **not** invent a note corpus — that's a product decision, not a bug fix.

## The decision this needs

What should the seed vault actually be? Roughly:

1. **Ship the MOC targets** — bundle the ~11 referenced MOCs (+ enough permanent notes to make them non-empty) so the seeded graph is real and explorable.
2. **Trim the MOCs to what ships** — cut dead links so the 16 shipped notes form a small but honest graph. Cheapest; the vault becomes a scaffold, not a library.
3. **Synthesize a small coherent demo vault** — purpose-built notes + MOCs that agree with each other.

Note the seed graph rendered by the Brain Orb (`resources/agent-visualization/data.json`, 514 nodes) does **not** correspond to the 16 notes on disk either, so the orb and the vault currently describe different knowledge bases. Whichever option is chosen should reconcile the two.

## Acceptance Criteria

- [ ] No dead wikilinks in the shipped seed vault (or a documented, deliberate exception)
- [ ] `Brain/README.md`'s "Start here" path leads somewhere real on a fresh install
- [ ] The vault's prose describes the corpus that actually ships (no phantom counts)
- [ ] The Brain Orb graph and the on-disk vault describe the same knowledge base
- [ ] A test guards it — extend `tests/unit/test_1646_cornelius_template_claims.py`, which already parses the vault (add a dead-wikilink check)

## Related

- #1646 (same vendoring class; fixed the prose, left the corpus)
- #1655 (CI validation of bundled templates — the systemic guard)
- trinity-enterprise#107 (auto-seeded Cornelius — why this is a first-impression bug)

obasilakis:
## Summary

The bundled `config/agent-templates/cornelius/` was a point-in-time snapshot of `Abilityai/cornelius` that drifted from its own prose: **22 of 23 MOC wikilinks dead**, and a `Brain/README.md` advertising "~1,000 interlinked notes" over a **19-file** vault whose literal "Start here" link went nowhere. Cornelius is auto-seeded on every fresh install, so this was the out-of-box "explore your second brain" experience.

This seeds from **`github:Abilityai/cornelius`** instead — an anonymous, source-mode clone with **no PAT**, on the tokenless public-repo path (trinity-enterprise#123).

## Why now, and why not just re-vendor

The dynamic-github create branch used to hard-require a system PAT even for a public repo. That was **the documented reason the bundle existed** (it's quoted in the old module docstring, and in trinity-enterprise#123's own body). It shipped in `00f68db8` and the constraint is gone.

The obvious fix — vendor the missing 1,067 files — would satisfy every AC **and arm the third instance**. This is the same failure as #1646 (declared ~24 skills / 9 sub-agents / a FAISS tier the bundle didn't carry) on the same bundle: a snapshot drifting from the prose describing it. The fix and the bug have the same shape, so "vendor it correctly this time" is the trap. Seeding from source retires the class.

It also makes `semantic_search` **real**: `resources/local-brain-search/` ships upstream but was never vendored, which is why #1646 had to trim the claim permanently.

## No offline fallback — deliberately

A fallback can only fire on a transient clone failure, and on that path it seeds the known-degraded copy **and** burns the durable `cornelius_seeded` flag — permanently stranding the install on the worse artifact. Leaving the flag unset and retrying next boot is strictly better, and is what `cornelius_agent_service` already does (it deliberately does not set the flag on failure).

There is also no air-gapped install to protect. Every `air-gap` string in the repo is about *telemetry tolerating a blocked POST* — one such comment sits directly below the seed call in `routers/setup.py`, which is how it got misread as a seeder constraint (including by me, in this issue's thread). Building the agent base image alone requires apt / npm / Go downloads, and an agent with no route to the Anthropic API is inert regardless.

## Acceptance criteria

- **No dead wikilinks, no phantom corpus claims** — satisfied by construction: no vault ships
- **Orb graph and vault describe the same KB** — all **514** exported nodes resolve against upstream `Brain/`, zero unresolved (the export was generated from it)
- **AC#5 relocates** — a local test cannot assert a vault that arrives over the network at provision time. `test_1646_cornelius_template_claims.py` retires with its subject; validating the upstream template belongs to #1655, which is where a freshness guard was always the right home

## Changes

- `CORNELIUS_TEMPLATE` → `github:Abilityai/cornelius`. `AgentConfig.source_mode` defaults `True`, which the tokenless path requires (`_gate_tokenless_request` 400s a non-source-mode tokenless request) — **now pinned by test**, since the seeder relies on the default rather than setting it
- Delete `config/agent-templates/cornelius/` (34 files, ~26k lines)
- Retire `test_1646_cornelius_template_claims.py` (240 lines; subject deleted)
- Docs: `architecture.md`, `feature-flows.md` + `cornelius-default-agent.md`, `requirements/core-agent.md`, and a resolution note on the #1646 `learnings.md` entry recording why re-vendoring was the trap

## Testing

Full unit suite green: **5035 passed, 15 skipped, 0 failures** (`-p no:randomly`).

> Note for reviewers: `pytest-randomly` is active by default, and `tests/unit/test_subscription_auto_switch_pingpong.py` has **pre-existing** order-dependent flakiness — a fixture calls `importlib.reload` on a module an earlier test drops from `sys.modules`, so it errors on some seeds on `dev` too. Unrelated to this PR; worth its own issue.

## Follow-ups (not in scope)

- **#1655** — template freshness/validation is the systemic guard, and now the home for the upstream wikilink check
- **trinity-enterprise#239** — the v0.8.5 manifest audit already cites #1656; Cornelius now falls under its "public, anonymously clonable" AC
- Worth measuring on a real fresh install: clone latency for the 1,067-file / 3.9 MB repo at first boot. Seeding is a fire-and-forget background task that never blocks setup, so this is a latency question, not a correctness one

Closes #1656


obasilakis:
## Clone latency measured — the open question in the PR body is closed

Run inside a throwaway container from the real agent base image (`docker run --rm --entrypoint bash trinity-agent-base:latest`), so it reflects the environment `startup.sh` actually clones in, not my laptop:

| Run | Full anonymous clone |
|---|---|
| 1 | 2925 ms |
| 2 | 2539 ms |
| 3 | 2393 ms |

**~2.5 s.** Seeding is a fire-and-forget background task that never blocks setup, so this is invisible to a first-run operator. No concern.

`startup.sh` takes the **full**-clone path here (L159/161 — clone into a temp dir, then merge into `/home/developer`), not the `--depth 1` path at L290, so the number above is the relevant one.

## Correction to the PR body

I wrote "1,067-file / 3.9 MB repo" — that was `Brain/` **markdown only**. The repo is **69 MB**:

| | |
|---|---|
| `.git` | 29 MB |
| `resources/` | 17 MB (incl. the FAISS semantic stack) |
| `docs/` | 15 MB |
| `Brain/` | 5.5 MB |
| working tree total | 69 MB, 1,232 files |

So each seeded Cornelius carries ~69 MB on its workspace volume, against roughly 2 MB for the deleted bundle. That is the cost of the vault being real instead of a 19-file stub advertising 1,000 notes — but it is a ~35× increase in per-agent seed footprint and should be stated accurately rather than buried.

Nothing here changes the approach; flagging it so the tradeoff is explicit at review.

AndriiPasternak31 (approved):
## `/validate-pr` — #1844

Validated against the methodology gates, and I independently verified the load-bearing premises against the real upstream rather than taking the PR body's word for them. The argument holds: **seeding from source is the right call here**, and the "just re-vendor it correctly" alternative really is the trap you name — it's the third instance of the same drift on the same artifact.

### What I verified (not just read)

| Premise | Method | Result |
|---|---|---|
| `Abilityai/cornelius` is public + anonymously clonable | credential-free `git ls-remote` (`GIT_CONFIG_GLOBAL=/dev/null`) | ✅ public, not archived, resolves |
| Tokenless path admits this create | `_gate_tokenless_request` (`crud.py:421-449`) | ✅ source-mode passes; your `source_mode is True` pin is genuinely load-bearing |
| Upstream carries every load-bearing surface | depth-1 clone + inspect | ✅ `capabilities: [brain-orb]`, `.trinity/brain-orb/{scopes,scope,search,action}`, `resources/local-brain-search/`, `.claude/`, 1,067 `Brain/*.md` |
| Orb renders out of the box | traced the chain | ✅ — but not by the mechanism the doc states, see below |
| Seed graph ↔ vault agree (AC#4) | parsed `data.seed.json` | ✅ 514 nodes, matches your claim |
| Security / packaging | full diff scan | ✅ no secrets, emails, IPs, env or credential files; no new top-level `src/backend/*.py`; no new `os.getenv` ⇒ no Dockerfile-COPY (#1033) or compose (#1056) gap |

Docs are thorough — `architecture.md`, `feature-flows.md` + the flow doc, `requirements/core-agent.md`, and the `learnings.md` resolution note. `Closes #1656` is same-repo, so `status-in-dev` auto-promotion will fire. 43 files is 34 deleted-bundle files; not a real split candidate.

### 1. Fix before merge (1 line, no re-review needed)

`docs/memory/feature-flows/cornelius-default-agent.md` carries this line over **unchanged** from the bundle description:

> `resources/agent-visualization/data.json` — a **pre-generated seed graph** so the orb renders immediately on first visit

Upstream does not ship `data.json` — it's gitignored there (`.gitignore:45`, "the live data.json is gitignored (re-export churn must not land in auto-sync commits)"). What ships is `data.seed.json`, and `.trinity/setup.sh:25-27` copy-once creates `data.json` on first boot:

```sh
if [ -f "$VIZ_DIR/data.seed.json" ] && [ ! -f "$VIZ_DIR/data.json" ]; then
    cp "$VIZ_DIR/data.seed.json" "$VIZ_DIR/data.json"
```

The **outcome** is still correct — I traced it end to end: source-mode clone at `startup.sh:289-292` runs before the `~/.trinity/setup.sh` invocation at `startup.sh:523-526`, so `data.json` exists by the time the orb is reachable. But the stated mechanism is wrong, and the dependency that replaced it (Trinity executing the template's `setup.sh`) is now load-bearing for first-visit render and is not mentioned anywhere in the doc. That's worth correcting on its own terms, and doubly so in this PR — a line describing an artifact that isn't there is the exact failure mode you're retiring.

Suggested replacement:

> - `resources/agent-visualization/data.seed.json` — a pre-generated 514-node seed graph. `data.json` is gitignored upstream (re-export churn must not land in auto-sync commits); the template's `.trinity/setup.sh` copy-once creates it at container start (`startup.sh:523`, after the clone), so the orb renders on first visit before the agent has produced its own export.

### 2. PR-body figure is ~10× low

"the 1,067-file / 3.9 MB repo" — 1,067 is the `Brain/*.md` count. The repo tracks **1,232 files**, checkout is **~39 MB** (67 MB with history; source-mode uses `--depth 1`, so ~39 MB transferred). Since that bullet exists specifically to flag clone latency for measurement, the number worth measuring against is 39 MB. Still fine against the 512 MB `/tmp` tmpfs the shallow clone stages in, and Cornelius requests 4g — no resource concern, just a wrong figure in the follow-up.

### 3. Flagging for explicit sign-off, not asking you to change it

**The out-of-box experience is now governed by an unpinned external ref.** `github:Abilityai/cornelius` has no `@branch`/tag, so every fresh install clones `main` at whatever state it is in, and Trinity CI cannot see that repo. You argue this deliberately and route the guard to #1655 — I agree the trade is favourable (the bundle was reviewable and still drifted **twice**), but it is a genuine transfer of first-impression risk to a repo outside this one's review surface, and #1655 is open and unscheduled. If you want the benefit without the exposure, `github:Abilityai/cornelius@<tag>` costs nothing today and turns "whatever main is" into a reviewed bump. Your call — flagging it so the decision is recorded rather than implicit.

**Test-adequacy gate, for the record:** #1656's AC#5 asked for a test guard extending `test_1646_cornelius_template_claims.py`; this PR deletes that file (240 lines) and relocates the AC. The reasoning is correct — a local unit test cannot assert a vault that arrives over the network at provision time — and you do add a real contract pin in its place. Net position is that no automated guard now covers #1646's or #1656's ACs until #1655 lands. Accepting that, but it should be a conscious hand-off rather than a silent one.

**Release-note ordering (narrow):** the tokenless clone needs a base image ≥ ent#123 — `architecture.md` records that old images silently skip it. Fresh installs build the base image, so the main path is safe. The exposure is an existing install whose `cornelius_seeded` flag is still unset (a prior deferral/failure) that pulls backend code but runs a stale `trinity-agent-base`: the retry would clone into an image that skips the tokenless path. Worth one line at release.

### Recommendation

**APPROVE.** The core change is correct, well-argued, and verified sound end to end; the no-offline-fallback reasoning (never burn the durable flag on a transient failure) is right and matches what `cornelius_agent_service` already does. Item 1 is a one-line doc correction — land it with this PR, no re-review needed from me.

PR #1845, opened 2026-07-28
fix(systems): resolve templates in the deploy dry run (#1841)

### LINKED ISSUE #1841: bug: system-deploy dry run passes a manifest with an unresolvable template — the preview clears what the real deploy 404s on
## Summary

`POST /api/systems/deploy` with `dry_run: true` returns `status: "valid"` and **zero warnings** for a manifest whose agent references a template that cannot resolve. The same manifest, deployed for real, fails that agent immediately. So the preview promised by the v0.8.6 note — *"Preview a manifest with a dry run before committing"* — clears a manifest it should flag.

## Evidence

Manifest (one resolvable template, one that does not exist):

```yaml
name: sysprobe
agents:
  good:
    template: local:dd-intake
  bad:
    template: local:this-template-does-not-exist
```

**Dry run — clean bill of health:**

```json
{
  "status": "valid",
  "agents_to_create": [
    {"name": "sysprobe-good", "template": "local:dd-intake"},
    {"name": "sysprobe-bad",  "template": "local:this-template-does-not-exist"}
  ],
  "warnings": [],
  "failed": []
}
```

**Real deploy of the identical manifest — the template is a hard 404:**

```json
{
  "status": "partial",
  "agents_created": ["sysprobe-good"],
  "failed": [{
    "name": "sysprobe-bad",
    "reason": "Local template 'this-template-does-not-exist' was not found. ...",
    "status_code": 404
  }],
  "warnings": ["Partial deploy: 1 agent(s) failed to create: ['sysprobe-bad']. Re-deploying this manifest will create suffixed duplicates of the already-created agents — fix the cause and create the missing agents individually."]
}
```

Best-effort and `strict: true` both behave correctly (strict aborts with `failed_at` + `created`, so the caller can see what side effects landed). The defect is confined to the preview.

## Why it matters

The partial-deploy warning states the recovery cost plainly: a re-deploy of the same manifest produces **suffixed duplicates** of the agents that already succeeded, so the fix is manual, per-agent. Dry run is exactly the control that should prevent paying that cost — and the failure it misses is the cheapest, most common one to make (a typo'd or renamed template id). A user who does the responsible thing, previews, and sees `valid` learns nothing that the deploy won't teach them the expensive way.

The gap widened with #1793/#1759: an unresolvable `local:` template is now a hard 404 at create time rather than silently producing an empty agent. Validation-time resolution is the matching half of that change.

## Suggested direction

Have `validate_manifest` (or the dry-run branch in `services/system_service.py::deploy_system`) resolve each agent's `template` and report unresolvable ones — as `warnings` at minimum, or as a populated `failed` list so the dry-run and real-deploy shapes match and a caller can diff them.

Both pieces already exist: `resolve_agent_names` runs in the dry-run path, and the create path owns template resolution. It needs no new lookup, only that the check moves ahead of the create loop.

Worth deciding during triage: whether a `github:` template should be probed too (network call at validation time) or whether resolution stays local-only for `local:` ids, which is where the cheap typo lives.

## Repro

1. `POST /api/systems/deploy` with the manifest above and `{"dry_run": true}` → `status: "valid"`, no warnings.
2. Same manifest, `{"dry_run": false}` → `status: "partial"`, `failed[0].status_code == 404`.

dolho:
## What

The deploy dry run now resolves each `local:` template, so a manifest that cannot deploy stops reporting `valid`.

Related to #1841

## Before / after, same manifest

```yaml
name: sysprobe
agents:
  good: {template: local:dd-intake}
  bad:  {template: local:this-template-does-not-exist}
```

**Before** — `status: "valid"`, `warnings: []`, `failed: []`. **After**, on a live instance:

```json
{
  "status": "invalid",
  "failed": [{
    "name": "sysprobe2-bad",
    "short_name": "bad",
    "template": "local:this-template-does-not-exist",
    "reason": "Local template 'this-template-does-not-exist' was not found. Check the id against GET /api/templates — ...",
    "status_code": 404
  }],
  "agents_to_create": ["sysprobe2-good", "sysprobe2-bad"]
}
```

## Why this and not a doc note

Recovery from a partial deploy is manual. The deploy path's own warning spells it out — re-running the manifest creates **suffixed duplicates** of the agents that already succeeded — so the fix is per-agent by hand. The preview is the control that avoids paying that, and the failure it was blind to (typo'd or renamed template id) is the cheapest mistake available. #1793/#1759 had just made an unresolvable `local:` template a hard 404 at create time; validation-time resolution is the matching half of that change.

## Approach

The preflight calls the **create path's own** `_resolve_local_template` on a throwaway `AgentConfig`, instead of re-deriving "does this template exist". Two consequences worth reviewing:

- The preview cannot drift from the deploy. The reason string and status code an operator sees in a dry run are produced by the same code that will produce them for real — `test_preflight_reason_matches_what_the_real_deploy_reports` asserts that equality directly rather than hard-coding an expected message.
- The throwaway config matters: that resolver **mutates** the object it is handed (type/resources/tools/runtime from `template.yaml`), so the manifest's own config is never passed to it.

The import is lazy, matching the existing `_default_create_agent_fn` seam — a module-level import would close a cycle.

**`github:` is deliberately not probed.** Validating it means a network call to GitHub with the platform PAT on a preview endpoint: slow, rate-limited, and a new outbound call on a path that had none. So a dry run still cannot promise a github-template manifest deploys, and a test pins that the local resolver is never reached for one. Worth a second opinion during review — the alternative is an opt-in `?probe_remote=true`.

**`status` gains `invalid`** for a preview with blockers, matching the deploy path's own `partial` / `failed` vocabulary rather than claiming success beside a populated `failed[]`. `agents_to_create` still lists the full plan, so the preview shows what was asked for and what blocks it. The response-model comment and the MCP `deploy_system` tool description are updated (Invariant #13).

## Verification

```
pytest tests/unit/test_ent125_resilient_system_deploy.py
19 passed
```

4 new cases; reverted the preflight with the tests in place → **2 failed** (the flagged-failure case and the preview/deploy equality case).

One fixture change to flag: the ent125 `env` fixture now stubs the resolver seam. The preflight reads the real catalog root (`/agent-configs/templates`), which does not exist under pytest, so without the stub every pre-existing dry-run test would have started failing — the same "new contract, old test" trap that blocked #1822.

Full unit suite posted below.

Fixes #1841

### DIFF
--- a/src/backend/models.py
+++ b/src/backend/models.py
@@ -622,7 +622,8 @@ class SystemDeployFailure(BaseModel):
 class SystemDeployResponse(BaseModel):
     """Response from system deployment."""
     # "deployed" (all created) | "partial" (some failed) | "failed" (none created)
-    # | "valid" (dry_run) — trinity-enterprise#125
+    # | "valid" (dry_run, will deploy) | "invalid" (dry_run, blockers in `failed`,
+    # #1841) — trinity-enterprise#125
     status: str
     system_name: str
     agents_created: List[str]  # Final agent names created
--- a/src/backend/services/system_service.py
+++ b/src/backend/services/system_service.py
@@ -781,6 +781,51 @@ def _default_create_agent_fn():
     return create_agent_internal
 
 
+def _preflight_template(
+    final_name: str, short_name: str, template: Optional[str]
+) -> Optional[SystemDeployFailure]:
+    """Can this agent's template resolve? Returns a failure, or None if it can.
+
+    Reuses the CREATE path's own resolver rather than re-deriving "does this
+    template exist", so the preview cannot drift from the deploy: the reason
+    string and status code a caller sees here are produced by the same code that
+    will produce them for real (#1841).
+
+    Scope, deliberately:
+      * ``local:`` — resolved (a filesystem read, no side effects). This is
+        where the cheap typo lives, and where #1793/#1759 made an unresolvable
+        id a hard 404 instead of a silent blank agent.
+      * ``github:`` — NOT probed. Validating it means a network call to GitHub
+        with the platform PAT on a preview endpoint; slow, rate-limited, and a
+        new outbound call on a path that had none. A dry run therefore still
+        cannot promise a github-template manifest deploys.
+      * no template — valid by construction (creates a bare agent by design).
+    """
+    if not template or not template.startswith("local:"):
+        return None
+
+    # Lazy import: crud imports service-layer modules, so a module-level import
+    # here would close a cycle (same reason `_default_create_agent_fn` is lazy).
+    from models import AgentConfig
+    from services.agent_service.crud import _resolve_local_template
+
+    try:
+        # A throwaway config: `_resolve_local_template` mutates the object it is
+        # given (type/resources/tools/runtime from template.yaml), which is why
+        # the manifest's own config is never handed to it.
+        _resolve_local_template(AgentConfig(name=final_name, template=template))
+    except Exception as e:  # noqa: BLE001 — mirrors the create loop's catch
+        reason, status_code = _failure_reason(e)
+        return SystemDeployFailure(
+            name=final_name,
+            short_name=short_name,
+            template=template,
+            reason=reason,
+            status_code=status_code,
+        )
+    return None
+
+
 def _failure_reason(exc: Exception) -> Tuple[str, Optional[int]]:
     """Normalize an agent-create exception into a (reason, status_code) pair.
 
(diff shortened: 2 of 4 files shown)


dolho:

```
5042 passed, 16 skipped, 4 failed  (27:26)

```

Same 4 environmental failures this repo shows on an untouched checkout: the first is a missing `tzdata` module, the other three trace to a `bcrypt has no attribute '__about__'` version mismatch that also appears in the backend boot log. Neither is touched by this PR — identical set on #1843 and #1847.

vybe (approved):

PR #1847, opened 2026-07-28
fix(skills): reap directories the package prune empties (#1842)

### LINKED ISSUE #1842: bug: skill-package prune leaves empty directories behind when a package drops a subdirectory
## Summary

When a skill package drops a subdirectory between versions, the platform prunes the **files** correctly but leaves the now-empty **directory** in the agent. Repeated reorganisations accumulate empty dirs inside `~/.claude/skills/<skill>/` that no version of the package ever references again.

Minor, but it is the one place where "pruned cleanly on update" (v0.8.6 release note) isn't literally true.

## Repro

Verified on a live agent (`evt-boss`), library at `/data/skills-library`.

**v1** — package with three files:

```
.claude/skills/pkg-probe/SKILL.md
.claude/skills/pkg-probe/scripts/hello.sh
.claude/skills/pkg-probe/reference/notes.md
```

Assign + `POST /api/agents/{name}/skills/inject` → delivered correctly (executable bit preserved, script runs in-container):

```
/home/developer/.claude/skills/pkg-probe/.trinity-skill.json
/home/developer/.claude/skills/pkg-probe/SKILL.md
/home/developer/.claude/skills/pkg-probe/reference/notes.md
/home/developer/.claude/skills/pkg-probe/scripts/hello.sh
```

**v2** — `rm -rf reference/`, add `tools/extra.py`, commit, re-inject. Files are right, and the manifest is right:

```json
{"manifest": [".claude/skills/pkg-probe/SKILL.md",
              ".claude/skills/pkg-probe/scripts/hello.sh",
              ".claude/skills/pkg-probe/tools/extra.py"],
 "version": "33081b336de4533590ba1a73c2fe39ebd6485f47"}
```

But the directory survives:

```
$ find ~/.claude/skills/pkg-probe -type d
/home/developer/.claude/skills/pkg-probe
/home/developer/.claude/skills/pkg-probe/reference     ← empty, from v1
/home/developer/.claude/skills/pkg-probe/scripts
/home/developer/.claude/skills/pkg-probe/tools

$ ls -la ~/.claude/skills/pkg-probe/reference
total 8
drwxr-xr-x 2 developer developer 4096 .
drwxr-xr-x 5 developer developer 4096 ..
```

## Cause

`skill_packaging.compute_prune` diffs manifests, and a manifest holds file paths only:

```python
stale = sorted(set(prev) - set(new_manifest))
```

So `reference/notes.md` is deleted and `reference/` is never considered. Nothing is wrong with the diff — the omission is that no step reaps directories left empty by it.

## Why it is worth a small fix

- A skill that instructs "read `reference/notes.md`" against a stale agent now hits an **empty directory** rather than a missing one. `ls reference/` succeeding with no contents is a more confusing failure than a clean `No such file or directory`.
- Empty dirs accumulate silently across every reorganisation; nothing ever removes them, including a full re-inject.

## Suggested direction

After the prune deletes files, `rmdir` (non-recursive, empty-only) each parent directory of a pruned path, bottom-up, stopping at `.claude/skills/<skill_name>/`.

This preserves the existing safety property that makes the current code correct — `compute_prune`'s comment is explicit that the platform "only ever deletes files IT wrote", and the previous manifest is untrusted agent-side state, so every candidate is prefix-confined. An empty-only `rmdir` keeps that: it cannot remove a directory holding runtime artifacts a skill's own scripts generated (`__pycache__`, downloaded models) or anything the agent authored, because such a directory is by definition not empty. The prefix confinement must be kept on the directory candidates too.

## Not a bug (checked while testing)

Re-injecting an unchanged package via `POST .../skills/inject` reports `status: "injected"`, not `"unchanged"`. That is intentional — the REST endpoint passes `force=True` as a repair action, and the version-skip path is `force=False` on agent start. Noting it here so the next person doesn't re-file it.

dolho:
## What

The package prune now removes directories it empties, instead of leaving them behind.

Related to #1842

## Live before / after

Same package, `reference/` dropped between v1 and v2:

```
before                                    after
.../pkg-probe/.trinity-skill.json         .../pkg-probe/.trinity-skill.json
.../pkg-probe/SKILL.md                    .../pkg-probe/SKILL.md
.../pkg-probe/reference        ← empty
```

## Why bother, at P3

The failure it produces is the confusing kind. A stale agent running a skill that says *"read `reference/notes.md`"* hits an **empty directory** rather than a clean `No such file or directory` — `ls` succeeds and shows nothing. And nothing ever cleans them up, including a full re-inject, so they accumulate with every reorganisation.

## Approach

Climb from each pruned file toward the skill root, removing only what is already empty.

**`os.rmdir` is the safety property, not an implementation detail.** It refuses a directory holding anything, so runtime artifacts a skill's own scripts wrote (`__pycache__`, downloaded models) and agent-authored files keep their directory alive — preserving exactly the guarantee `compute_prune` documents, that the platform only ever deletes what it wrote. It also refuses a symlink (`NotADirectoryError`), so a doctored tree can't redirect it at something else, and the path prefix stays confined to `.claude/skills/<name>/` the same way `compute_prune` confines its own candidates. The climb stops at the skill root, which stays installed.

**Placement:** it rides in `_finalize_injected_dirs` rather than next to the per-file DELETEs, because "is this directory now empty?" has to be asked of the filesystem — one in-agent exec answers it for every pruned path at once, where the remote file API would cost a round trip per directory. Only paths whose DELETE actually succeeded are passed, so a failed prune never triggers a climb.

## Verification

```
pytest tests/unit/test_ent183_skill_packages.py tests/unit/test_183_skill_packages_backward_compat.py
64 passed
```

5 new cases that **execute** the rendered script against a tmp home — a mocked assertion on the argument list would not catch a bad climb:

- directory emptied by the prune is removed
- directory still holding `run.sh` + `__pycache__/x.pyc` is kept (`rmdir == 0`)
- multi-level climb removes `b/` then `a/` and stops at the skill root
- out-of-prefix paths (`workspace/data.csv`, `../../etc/passwd`) are ignored
- a symlink where the package dir used to be is not followed, target survives

Removed the reap with the tests in place → **2 failed**.

Also updated `test_fallback_skills_queue_no_chmod_paths`, which unpacked the old 3-arg `_finalize_injected_dirs` signature. Caught it in the run rather than in review — same trap that blocked #1822.

Verified live end to end on a running instance: v1 injected with `reference/notes.md`, v2 dropping the directory re-injected, `reference/` gone where it previously survived.

## Not in scope

`unmanaged_dir_overwritten` (a skill directory with no `.trinity-skill.json`) still prunes nothing at all, by design — the platform can't tell which files it wrote. Unchanged here.

Fixes #1842

### DIFF
--- a/src/backend/services/skill_service.py
+++ b/src/backend/services/skill_service.py
@@ -640,24 +640,35 @@ print(json.dumps({{
         return result
 
     async def _finalize_injected_dirs(
-        self, agent_name: str, skill_names: List[str], exec_paths: List[str]
+        self,
+        agent_name: str,
+        skill_names: List[str],
+        exec_paths: List[str],
+        pruned_paths: Optional[List[str]] = None,
     ) -> Optional[Dict[str, Any]]:
         """Post-restore finalization in ONE exec: chmod +x (git modes are
-        dropped by restore's write_bytes), per-skill .gitignore lines, and
-        untracking already-committed matches.
+        dropped by restore's write_bytes), per-skill .gitignore lines,
+        untracking already-committed matches, and reaping directories the prune
+        just emptied (#1842).
 
         The gitignore step is load-bearing (strategy review F1): the 15-min git
         auto-sync deliberately COMMITS .claude/, so without per-name ignore
         lines every injected package lands in the agent's GitHub repo — the
         #1595/#1596 fleet bloat class. Only platform-injected names are
         ignored; agent-authored Playbooks keep committing.
+
+        The empty-dir reap rides here rather than next to the per-file DELETEs
+        because it needs to ASK the filesystem whether a directory is now empty
+        — one in-agent exec answers that for every pruned path at once, where
+        the remote file API would need a round trip per directory.
         """
         script = f"""
 import json, os, stat, subprocess
 names = {json.dumps(sorted(set(skill_names)))}
 exec_paths = {json.dumps(sorted(set(exec_paths)))}
+pruned_paths = {json.dumps(sorted(set(pruned_paths or [])))}
 home = os.path.expanduser("~")
-out = {{"chmod": 0, "gitignore": False, "untracked": False, "errors": []}}
+out = {{"chmod": 0, "gitignore": False, "untracked": False, "rmdir": 0, "errors": []}}
 for rel in exec_paths:
     p = os.path.join(home, rel)
     try:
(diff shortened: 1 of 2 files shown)


dolho:

```

```

The failures are the environment's standing set (missing `tzdata`; three in `test_admin_email_login` that trace to a `bcrypt has no attribute '__about__'` version mismatch visible in the backend boot log). Same set on #1843 and #1845, and on an untouched checkout.

vybe (approved):

PR #1851, opened 2026-07-28
fix(rename): derive the rename cascade from AGENT_REFS, not a second hand list (#1819)

dolho:
## What

`rename_agent` now derives its table list from `db.agent_cleanup.AGENT_REFS` — the registry the delete path already used — instead of maintaining a second, hand-written one.

Related to #1819 (rename half; see "Not in this PR" below)

## The bug is bigger than the title

The issue reports two stranded tables. Measured against this schema, the rename list had fallen **23 tables** behind the registry:

```
registry (AGENT_REFS): 41 tables
rename (hand-written): 19 tables

not re-keyed by rename: agent_sessions, agent_session_messages, agent_reminders,
agent_loops, agent_notifications, agent_events, agent_event_subscriptions,
operator_queue, access_requests, agent_sync_state, agent_compatibility_results,
agent_dashboard_cache, public_user_memory, nevermined_agent_config,
nevermined_payment_log, subscription_rate_limit_events, slack_channel_agents,
slack_active_threads, telegram_bindings, whatsapp_bindings, voip_bindings,
voip_call_logs, enterprise_connectors
```

So renaming an agent also stranded its reminders, its operator-queue items, and its Telegram / WhatsApp / VoIP / Slack bindings — not just Session-tab history.

**`cascade_rename` already existed**, registry-driven and complete, with **zero callers**. It was written for exactly this and never wired in. That is the same dead-mechanism shape as `register_agent_owned_table` (noted in the issue thread).

## Why this is safe to switch

Checked before replacing anything: **every table the hand-written path touched is already in the registry** — the hand list was a strict subset, so nothing loses coverage. `cascade_rename` skips absent tables (`_table_exists`) and handles the multi-column refs (`agent_permissions.source_agent`/`.target_agent`, the event-subscription pair) that the old code spelled out twice.

`agent_ownership` stays hand-written here: it is the row being renamed, and it carries the #1664 `volume_base_name` pin that must move in the same statement.

## One deliberate behaviour change

The old code re-keyed **every** `mcp_api_keys` row carrying the agent's name. The registry declares two scope-filtered refs (`scope='agent'`, `scope='connector'`), so a user- or system-scoped row no longer follows a rename.

That is the point rather than a side effect — `cascade_delete` already used exactly those filters, and honouring them in rename is what makes the two paths one rule. Those scopes carry `agent_name` as provenance, not ownership. Pinned by `test_scope_filtered_keys_narrow_deliberately` so it stays a decision.

## Tests

`tests/unit/test_1819_rename_cascade_parity.py` — 4 cases, **all four fail against the pre-fix code**:

| test | guards |
|---|---|
| `test_rename_rekeys_every_registered_table` | seeds one row per registered table against the **real schema**, renames, asserts nothing is left behind. Reports what it could not seed, so coverage can't silently shrink |
| `test_session_history_specifically_follows_the_rename` | the reported symptom, with realistic session + 4 message rows |
| `test_scope_filtered_keys_narrow_deliberately` | the mcp_api_keys narrowing above |
| `test_rename_does_not_maintain_its_own_table_list` | structural — a future `update(<table>)` inside `rename_agent` fails CI. This is the drift that caused the bug; the behavioural test alone wouldn't catch a partial re-add |

```
pytest test_agent_cleanup_parity test_1664_renamed_agent_volume_safety \
       test_181_agent_display_label test_1819_rename_cascade_parity
65 passed
```

**Verified live** on a dev instance: an agent with 1 session and 10 reports renamed and back, with `agent_sessions`, `agent_session_messages`, `agent_reports`, `agent_events`, `agent_event_subscriptions`, `operator_queue` and `mcp_api_keys` all showing **0 rows under the old name** afterwards (before the fix, sessions stayed behind).

## Not in this PR

**The rooms dark-mode defects** added by the second comment (`NewRoomDialog.vue`, `RoomsRail.vue`, `Sessions.vue` backdrop). Those files exist only on `feature/ent170-sessions-ui` — the rooms components are not on `dev` — so they cannot land on this base. They need their own PR against that branch.

Also unaddressed, and worth its own issue: `EXTRA_AGENT_REFS` / `register_agent_owned_table` still has no callers, so entitled-module tables join neither cascade. This PR keeps the existing `EXTRA_AGENT_REFS` pass intact but cannot populate it.

### DIFF
--- a/src/backend/db/agent_settings/metadata.py
+++ b/src/backend/db/agent_settings/metadata.py
@@ -13,23 +13,8 @@ from ..engine import get_engine
 from ..tables import (
     agent_ownership,
     agent_sharing,
-    agent_schedules,
-    schedule_executions,
-    chat_sessions,
-    chat_messages,
-    agent_activities,
     agent_permissions,
-    agent_shared_folder_config,
     agent_git_config,
-    agent_skills,
-    agent_tags,
-    agent_public_links,
-    mcp_api_keys,
-    agent_health_checks,
-    agent_dashboard_values,
-    monitoring_alert_cooldowns,
-    agent_shared_files,
-    agent_reports,
 )
 
 
--- a/src/backend/db/agent_settings/metadata.py
+++ b/src/backend/db/agent_settings/metadata.py
@@ -92,26 +92,16 @@ class MetadataMixin:
         """
         Rename an agent by updating all database references.
 
-        This updates agent_name in all tables that reference it:
-        - agent_ownership (primary)
-        - agent_sharing
-        - agent_schedules
-        - schedule_executions
-        - chat_sessions
-        - chat_messages
-        - agent_activities
-        - agent_permissions (source and target)
-        - agent_shared_folder_config
-        - agent_git_config
-        - agent_skills
-        - agent_tags
-        - agent_public_links
-        - mcp_api_keys
-        - agent_health_checks
-        - agent_dashboard_values
-        - monitoring_alert_cooldowns
-        - agent_shared_files
-        - agent_reports
+        Re-keys `agent_name` in `agent_ownership` (here) and in EVERY table
+        registered in `db.agent_cleanup.AGENT_REFS` (via `cascade_rename`),
+        plus any entitled-module table registered at runtime through
+        `register_agent_owned_table`.
+
+        Deliberately NOT a hand-maintained list of tables (#1819): the previous
+        docstring enumerated 19 and the code matched it, while the registry had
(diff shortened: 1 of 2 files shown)


dolho:
rooms-before
rooms-rail-after
rooms-after

vybe (approved):
Validated via /validate-pr: registry-derived rename cascade reviewed (deliberate mcp_api_keys scope narrowing is pinned by test), security greps clean, new parity suite + agent_cleanup parity pass locally on the combined worktree. Merging.

Issue #1848, opened 2026-07-28
bug: MCP key row metadata misaligns when a key has never been used


vybe:
## Summary

On **Settings → MCP Keys**, the per-key metadata line (key prefix · owner · created · last used · request count) renders as a single non-wrapping flex row. When a key has never been used (`last_used_at` is null, `0 requests`), the row's items compress and the **Created timestamp wraps mid-value** (the `AM`/`PM` suffix drops to its own line), and the metadata no longer lines up column-wise with the neighboring rows that do have a *Last used* value. Small cosmetic glitch, but it makes the key list look broken at a glance.

## Context

Noticed during the v0.8.5 release screenshot session while capturing the MCP Keys tab in dark mode at a 1440px viewport: two adjacent key rows with usage render their Created/Last-used values in what reads as tidy columns, while a never-used key row renders its Created timestamp inline and awkwardly wrapped.

## Acceptance Criteria

- [ ] A key row with `last_used_at = null` renders its metadata cleanly: no mid-timestamp wrap (no orphaned `AM`/`PM`), regardless of viewport width down to ~1280px
- [ ] Rows with and without a *Last used* value are visually consistent (either true column alignment or a uniform inline flow — pick one; today it's an accidental mix)
- [ ] Long timestamps don't force the trailing "N requests" item to misalign between rows
- [ ] Dark and light mode both verified

## Technical Notes

The metadata row is in `src/frontend/src/components/settings/McpKeysTab.vue` (~lines 91–108): a `flex items-center space-x-4` div of inline `<span>`s with no `flex-wrap`/`shrink` control and a conditional `v-if="key.last_used_at"` span in the middle. Likely fixes: add `flex-wrap gap-x-4 gap-y-0.5` + `whitespace-nowrap` on each span (uniform inline flow), or move Created/Last-used into a small fixed-width grid so rows align. A `whitespace-nowrap` on the timestamp spans alone would already kill the orphaned-`AM` wrap.

PR #1852, opened 2026-07-28
fix(ui): dark-mode contrast in the rooms surface (#1819)

dolho:
The rooms half of #1819. The dev-based PR (#1848) cannot carry it: `RoomsRail.vue`, `NewRoomDialog.vue` and the rest of `components/rooms/` exist only on `feature/ent170-sessions-ui`, so this targets that branch instead.

## What was wrong

Four components inherit body colour instead of declaring their own. In dark mode that resolves to near-black text on a near-black surface.

| Component | Symptom |
|---|---|
| `RoomsRail.vue` | Session **name** invisible. Its subtitle and counts declare `text-gray-400`, so they stayed readable — the rail looked like blank rows with metadata under them. |
| `NewRoomDialog.vue` | "New session" heading, agent-picker names, and the Name / Topic / Max messages / Max cost / TTL fields. Typed values were the worst of it — you could not read what you had entered. |
| `ParticipantsRail.vue` | Heading + participant identities. **Not named in the issue.** |
| `RoomComposer.vue` | Textarea text + placeholder. **Not named in the issue.** |

The last two carry the same omission and sit one strip and one textarea from the dialog. Fixing the reported three and leaving these would have shipped the same bug in the same view, so they are in.

Second half of the report — the backdrop — moves from `bg-black/40` to `bg-black/60 backdrop-blur-sm`. At 40% the page behind stayed legible through the scrim and the transcript empty-state copy read as stray text floating beside the dialog (visible at the right edge of the before shot). The blur separates the layers without hiding the page, so the modal still reads as an overlay rather than a new screen.

## Fix

Every control declares text + placeholder colour for both themes rather than relying on inheritance. No structural, layout or logic change — `git diff` is class attributes plus one comment.

`Sessions.vue` is **not** touched: the backdrop lives inside the dialog, so the fix lands there. That also keeps this clear of in-flight work on that file.

## Verification

Live dark mode on local dev, admin/admin, `localStorage['trinity-theme']='dark'` + `prefers-color-scheme: dark`, 6 rooms seeded.

| | Path |
|---|---|
| Before | `/tmp/claude-1000/-home-oleksii-Desktop-abilityai-trinity/6e6de766-392c-4cd1-bece-cab147416521/scratchpad/rooms-before.png` |
| After (dialog) | `…/scratchpad/rooms-after.png` |
| After (rail, dialog closed) | `…/scratchpad/rooms-rail-after.png` |

Before shows all four defects at once: unreadable rail names, a barely-visible heading, dark-on-dark `60` / `—` / `24`, and the "session" fragment bleeding through the backdrop. After shows each readable. The rail needs its own shot because the heavier scrim correctly dims it while the dialog is open.

## Scope

Front-end only, `src/frontend/src/components/rooms/` (4 files, +17/-13). No backend, no schema, no API, no test change — this is a Tailwind class defect with no behaviour to assert on.

Related to #1819

### DIFF
--- a/src/frontend/src/components/rooms/NewRoomDialog.vue
+++ b/src/frontend/src/components/rooms/NewRoomDialog.vue
@@ -1,20 +1,24 @@
 <template>
   <div class="fixed inset-0 z-40 flex items-center justify-center p-4" @click.self="$emit('close')">
-    <div class="absolute inset-0 bg-black/40"></div>
+    <!-- #1819: 40% was translucent enough that the page behind (the empty-state
+         copy in particular) read as stray text floating beside the dialog. A
+         heavier scrim plus a blur separates the layers instead of hiding the
+         page, so the modal still reads as an overlay. -->
+    <div class="absolute inset-0 bg-black/60 backdrop-blur-sm"></div>
     <div class="relative w-full max-w-lg rounded-2xl bg-white dark:bg-gray-900 shadow-xl border border-gray-200 dark:border-gray-800 max-h-[90vh] flex flex-col">
       <div class="px-5 py-4 border-b border-gray-200 dark:border-gray-800">
-        <h2 class="text-base font-semibold">New session</h2>
+        <h2 class="text-base font-semibold text-gray-900 dark:text-gray-100">New session</h2>
         <p class="text-xs text-gray-400 mt-0.5">Pick agents to work a topic together. You can @mention any of them once the room opens.</p>
       </div>
 
       <div class="px-5 py-4 space-y-4 overflow-y-auto">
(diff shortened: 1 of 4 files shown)


vybe (approved):
Validated via /validate-pr: pure Tailwind class-attribute diff (explicit text/placeholder colors for both themes) + scrim change, zero logic changes. Security sweep clean (no secrets/emails/IPs/env/infra files). All checks green, base=feature/ent170-sessions-ui is correct — rooms components exist only there. Note: body references dev-based PR as #1848; the actual sibling is #1851 (1848 is an unrelated issue).

PR #1861, opened 2026-07-28
fix(ci): detect frontend changes via the paginated files API, fail open (#1782)

### LINKED ISSUE #1782: bug(ci): frontend-e2e 'changes' path-filter hard-fails on PRs >300 files — every release PR shows a red X
The `changes` job in `frontend-e2e.yml` detects frontend changes via:

    gh pr diff <n> --repo Abilityai/trinity --name-only

GitHub's diff API refuses PRs whose diff exceeds **300 files** (HTTP 406 `PullRequest.diff too_large`), so the job exits 1 and the whole workflow reports failure — with `e2e`/`report` skipped, not run.

First hit: release PR #1781 (v0.8.5, 619 files). Every future release PR (routinely 500+ files) will show the same red X. Introduced with the #1526 rework (the job predates >300-file PRs reaching it).

**Fix options**
- Use the paginated files API instead: `gh api repos/{owner}/{repo}/pulls/{n}/files --paginate --jq '.[].filename'`
- Or checkout + `git diff --name-only "origin/${{ github.base_ref }}"...HEAD`
- Either way, treat "cannot determine changed files" as `frontend=true` (fail-open into running the suite), not job failure.

Not a required check on main, so v0.8.5 shipped past it — but the red X on a release PR is noise we shouldn't normalize.

AndriiPasternak31:
## Problem

The `frontend-e2e` `changes` job detected frontend edits with `gh pr diff --name-only`. That
calls the **diff** API, which hard-fails with **HTTP 406** above 300 files. Every release-sized
PR therefore did two bad things at once:

1. posted a spurious **red X** on the PR, and
2. left `frontend` **unset**, so the e2e suite was **skipped — not run** — on exactly the
   largest, riskiest PRs.

GitHub's own 406 body names the fix: *"Consider using 'List pull requests files' API."*

## Change

- Detect changed files via `gh api repos/{repo}/pulls/{n}/files --paginate` (the files API,
  which paginates) instead of the diff API.
- **Fail open.** If the list can't be fetched *or can't be proven complete*, emit
  `frontend=true` and `exit 0`. This job decides whether e2e runs, so "unable to decide" must
  mean **run**, and must never mean a red X.
- **Prove completeness exactly**, by comparing the list length against the PR's own
  `changed_files`. An exact comparison catches both truncation at GitHub's 3000-file ceiling
  and a pagination that died mid-way but still exited 0 — the latter is silently missed by a
  `>= 3000` heuristic.
- Move `${{ }}` interpolation into `env:` (`PR_NUMBER`, `REPO`) rather than splicing
  expressions into the shell.
- Keep `|| needs.changes.result == 'failure'` on the e2e job as defence in depth.

### One deliberate non-change

An earlier revision also carried `|| needs.changes.result == 'cancelled'`. **That was removed**,
and the reasoning is recorded in `docs/memory/learnings.md`. GitHub cancels a run by
re-evaluating each job's `if` and keeping the job whenever it returns true — that is precisely
what `always()` means. This workflow sets `cancel-in-progress: true` for PRs, so the dominant
producer of a `cancelled` result is a *superseded push*. The clause would have turned every
superseded push into a full 25-minute Docker-stack + Playwright run whose failure posts the very
red X this issue exists to remove. A cancelled run has been superseded; the replacement run
evaluates from scratch. `failure` is kept — that one is a genuine "ran and lost" case.

`container-security.yml` detects changed files with `git diff` instead. It already checks out the
repo to build images; this job must not (a clone turns a ~5s gate into ~40s). The two
implementations differ **on purpose** — there's a comment saying so.

## Verification

`/verify-local` cannot exercise a GitHub Actions workflow, so verification was done by extracting
the step's **shipped bytes** straight out of the YAML and executing them against the real GitHub
API, asserting on the two things the workflow actually consumes: the step's **exit code** and the
contents of **`$GITHUB_OUTPUT`**. The pre-fix step was extracted from `origin/dev` and run through
the identical harness for a symmetric before/after.

**13/13 cases pass.**

| # | Case | Exit | `GITHUB_OUTPUT` |
|---|------|------|-----------------|
| 1 | **BEFORE** — `gh pr diff` on PR #1781 (603 files) | `1` ❌ | *empty — no verdict* |
| 2 | **AFTER** — files API on PR #1781 (603 files) | `0` | `frontend=true` |
| 3 | fail-open: nonexistent PR (404) | `0` | `frontend=true` |
| 4 | fail-open: invalid `GH_TOKEN` (401) | `0` | `frontend=true` |
| 5 | fail-open: nonexistent repo (404) | `0` | `frontend=true` |
| 6 | backend-only PR #1843 (7 files, 0 frontend) | `0` | `frontend=false` |
| 7 | frontend PR #1852 (4 files, all frontend) | `0` | `frontend=true` |
| 8 | mixed PR #1839 (39 files, 3 frontend) | `0` | `frontend=true` |
| 9 | **BEFORE** on small PR #1843 — control | `0` | `frontend=false` |
| 10 | completeness guard: truncated at 3000 ceiling (3500 declared / 3000 returned) | `0` | `frontend=true` |
| 11 | completeness guard: pagination died mid-way, exit 0 (603 / 200) | `0` | `frontend=true` |
| 12 | completeness guard: pagination died mid-way, exit 1 | `0` | `frontend=true` |
| 13 | completeness guard **control**: counts agree, all backend → must *not* fail open | `0` | `frontend=false` |

Case 1 is the reported bug reproduced exactly: red X **and** no verdict. Cases 10–13 use a `gh`
shim on `PATH`, since no real PR can be made to truncate on demand; case 13 is the control that
stops the guard from passing vacuously. Cases 3–5 confirm the fail-open branch is reached by
genuinely different failures (bad PR, bad credentials, bad repo), not one code path three times.

`actionlint` reports the **same 2 pre-existing shellcheck findings** before and after, both in
other jobs — this change adds none.

### Live run on this PR

The harness above runs locally, so it cannot exercise the Actions runner itself — job-level `if`
evaluation, `needs` wiring, `$GITHUB_OUTPUT` plumbing between jobs, or the `GITHUB_TOKEN`
permission set. That needs a real run. This PR produced one
([run 30384069438](https://github.com/Abilityai/trinity/actions/runs/30384069438)):

- `changes` — **success in 5s**, **0 annotations** (so the fail-open warning was *not* emitted).
- Step log: `changed files: 3` — the new code path ran (that line exists only in the fix) and the
  completeness check agreed with the PR's `changed_files`.
- `e2e` — **skipped**, correct: this PR touches no `src/frontend/**`.

That proves, on the runner: the job is green (the red X is gone), the files-API path works with
the workflow's own `GITHUB_TOKEN` and `pull-requests: read` scope, and the verdict propagates
through `needs` to gate `e2e`.

### What is still *not* proven on the runner

This PR is 3 files, so its live run exercises neither the **>300-file path** nor the
**fail-open branch** — both are covered only by the local harness against the real API (cases
1–5, 10–13). The next release-sized PR is the first live exercise of the >300-file path.

The `cancelled`-clause removal is not covered by any automated case here. It is derived from
documented Actions cancellation semantics and validated by reading
`concurrency.cancel-in-progress` in this same file; the reasoning is written up in
`docs/memory/learnings.md` so a future reader doesn't "helpfully" re-add it.

Closes #1782


vybe (approved):
Validated: fixes #1782. Files-API pagination with an exact completeness proof (count vs changed_files), fail-open semantics on every undecidable path, and a correct rationale for excluding 'cancelled' from the always() guard. Closing keyword present so #1782 auto-promotes.

Issue #1332, opened 2026-06-23
fix(activities): cancelled execution's activity row recorded as FAILED (no ActivityState.CANCELLED)


AndriiPasternak31:
## Context

Surfaced during the #679 review/implementation (plumbing the operator-cancel signal into the agent task-runner reply). #679 makes a cancelled turn write `TaskExecutionStatus.CANCELLED` on the `schedule_executions` row correctly, but its **activity** row is still closed as `FAILED`.

## Problem

A CANCELLED execution flows through the failure-style branch of the terminal applier, which closes the activity as `ActivityState.FAILED`:

- `src/backend/services/task_execution_service.py:1449-1454` — `apply_result` failure-style branch passes `status=ActivityState.FAILED` for any non-SUCCESS terminal, including CANCELLED.
- `src/backend/models.py` `class ActivityState` — only `STARTED` / `COMPLETED` / `FAILED`; there is **no `CANCELLED`** state.

So a user-initiated cancel is recorded as a **FAILED activity**, even though the execution row itself is correctly `CANCELLED`.

## Impact (observability only — no money/data-correctness bug)

- Activity timeline, Collaboration Dashboard, Overview "needs-attention", and activity-based health surfaces conflate operator cancels with genuine agent failures.
- The terminate handler does log a *separate* `EXECUTION_CANCELLED` activity (`routers/chat.py` `terminate_agent_execution`), so a cancel signal exists in the timeline — but the execution's **own dispatch activity** is still FAILED.
- Execution-row analytics (terminal-based success rate over `schedule_executions.status`) are unaffected — CANCELLED is excluded correctly there. The divergence is only in activity-derived views.

## Why deferred from #679

This is cross-cutting and out of scope for #679's cancel-labeling:
- adds an `ActivityState` enum value,
- the frontend has **no `cancelled` activity state** today (rendering in `AgentNode.vue` / `ReplayTimeline.vue` / `ChatPanel.vue` assumes started/completed/failed),
- needs a check of any analytics / canary surface that buckets on `activity_state`.

## Proposed approach (confirm during /autoplan)

1. Add `ActivityState.CANCELLED` in `models.py`.
2. In `apply_result`, pass `ActivityState.CANCELLED` (not FAILED) when `envelope.status == CANCELLED` — or skip the activity-FAILED close for cancels.
3. Thread the new state through frontend activity rendering (color/label) so a cancel reads as a cancel, not a failure.
[...]

Issue #767, opened 2026-05-10
bug(timeline): CB probe executions left open until backend restart inflate failure duration on timeline


vybe:
## Summary

When the circuit breaker fires probes against an agent, each probe creates a `schedule_executions` record. If the probe is rejected (CB open), the record is never closed — it stays in a non-terminal state until the next backend restart, which runs startup cleanup and marks it `failed`. The result is a large red block on the execution timeline spanning the entire window from probe creation to backend restart, giving a misleading picture of agent activity.

## Component

Backend / Circuit Breaker / Execution Timeline (Frontend)

## Priority

P3

## Observed Behavior

1. Circuit breaker opens for an agent
2. CB fires probe executions — each creates a `schedule_executions` row
3. Probes are rejected (CB open), but their DB records are never closed
4. Records sit in non-terminal state (`running`, `pending_retry`, etc.)
5. On next backend restart, startup cleanup closes them with `error = "Execution orphaned — recovered on backend restart"`
6. `completed_at` is set to the restart time, not the actual probe end time
7. Timeline displays a large red failure block spanning the full gap (e.g. 40 minutes) instead of a brief probe event

## Root Cause

CB probe executions are not closed at the point of rejection. The backend creates the execution record before dispatching to the agent, but when the CB rejects the call, no corresponding `completed_at` / `failed` update is written back to `schedule_executions`. The record is only resolved as a side-effect of backend restart orphan cleanup, which stamps the current wall-clock time as `completed_at` — not the actual failure time.

## Reproduction Steps

1. Trigger enough consecutive agent failures to open the circuit breaker
2. Observe CB firing probe executions in scheduler logs
3. Let the CB remain open for several minutes without a backend restart
4. Check `schedule_executions` — probe records will be in non-terminal state
5. Restart the backend
6. View the agent's execution timeline — probe records now show as long-duration failures

## Suggested Fix

When the circuit breaker rejects a call, immediately close the associated execution record with `status = 'failed'`, `completed_at = now()`, and a descriptive error such as `"Rejected by circuit breaker (open state)"`. This ensures the timeline accurately reflects a brief probe attempt rather than a multi-minute failure span.

Alternatively, CB probe executions could be marked with a distinct status (e.g. `cb_rejected`) so the frontend can render them differently (e.g. as a small marker rather than a full red block).

## Related

- Circuit breaker implementation: `src/backend/services/agent_client.py`
[...]

PR #1863, opened 2026-07-28
fix(execution): the CAS-won terminal owns the paired activity close (#1804)

### LINKED ISSUE #1804: bug: recovery paths write execution terminals without closing the paired activity — agent shows as running for up to 2h, then logs a fabricated 120-min failure
## Summary

Every terminal writer **outside** the dispatching coroutine writes the terminal to `schedule_executions` but never closes the paired `agent_activities` row. The activity stays `activity_state='started'` until the generic 120-minute sweep closes it — so the Dashboard Timeline renders the agent as still working for up to 2 hours after the execution is terminal, and then renders it as a **~120-minute failure** with a fabricated `duration_ms` that is never corrected.

Observed on a live instance: two `chat_start` activities left open for >80 min against executions that had been terminal `failed` for 70 min. Same class as #45 (tool-call activities orphaned) and #767 (CB probes inflating timeline duration) — both fixed at a single producer, neither changed the ownership model.

## Root cause

The activity id is a **local variable in the dispatching coroutine**:

```python
# services/task_execution_service.py:1107
activity_id = await activity_service.track_activity(..., related_execution_id=execution_id, ...)
```

and the close is gated on that same coroutine winning the CAS:

```python
# services/task_execution_service.py:816-826  (_write_terminal_and_gate)
won = db.update_execution_status(execution_id=..., status=status, ...)   # CAS
...
if won and activity_id:                                                   # <- the gate
    await activity_service.complete_activity(...)
```

So the activity is closed only when (a) the original coroutine is still alive AND (b) it wins the CAS. Two independent ways that fails:

**(a) Process death.** `execute_task` has no `finally` that closes the activity — the close happens exclusively through this applier. If the backend process dies mid-turn, `activity_id` dies with it and nothing else knows the row exists. (With `--reload` in dev, every file save that lands mid-execution mints one.)

**(b) CAS loss.** A late in-process writer that loses the CAS to a recovery path hits `won=False` and skips line 825. Note the asymmetry — the SUCCESS applier *does* handle CAS loss (`task_execution_service.py:1841 if not won:` closes with `activity_state_for_terminal(...)`, #1332); `_write_terminal_and_gate` has no equivalent branch.

**And the writer that wins the CAS never takes over the responsibility.** Both recovery paths stop at the row:

```python
# services/cleanup_service.py:2031-2048  (watchdog orphan / auto-terminate)
updated = db.mark_execution_failed_by_watchdog(execution_id, combined_error)
await capacity.release_if_matches(agent_name, execution_id)   # slot: released
await self._broadcast_watchdog_event(...)                     # WS: broadcast
#                                                               activity: never closed

# services/cleanup_service.py:2383-2387  (startup restart recovery)
db.mark_execution_failed_by_watchdog(..., "Execution orphaned — recovered on backend restart")
await capacity.release(agent_name, execution["id"])
#                                                               activity: never closed
```

The bulk sweeps `_sweep_stale_executions` (`cleanup_service.py:602`) and `_sweep_no_session_executions` (`:631`) have the same gap.

**The helper already exists and is two methods away.** `db/activities.py:183` `get_open_activity_id_for_execution()` — filtered to `chat_start|schedule_start` + `state='started'` — was written for exactly this. It has precisely two callers, both #1083/#429-specific: `_close_stale_slot_activity` (`cleanup_service.py:1549`) and `_close_reaped_activity` (`:579`). None of the writers above call it.

## Impact

1. **Timeline shows phantom in-progress work.** `ReplayTimeline.vue:670` — `const isInProgress = event.status === 'started'`. (The agent tile is unaffected: `agent_service/stats.py:202` derives `activityState` from recency, not `activity_state`.)
2. **Permanently wrong duration.** The backstop `db/activities.py:331` sets `duration_ms = now − started_at`, so a 15-minute run closes as a 120-minute failure. Nothing recomputes it. This is exactly the misleading-timeline symptom of #767, one layer down.
3. **Self-heals slowly, not correctly** — `ACTIVITY_STALE_TIMEOUT_MINUTES = 120` (`cleanup_service.py:56`) via `_sweep_stale_activities` (`:703`).

## Reproduction

1. Trigger a long execution (`POST /api/agents/{name}/schedules/{id}/trigger`).
2. While it runs, restart the backend (or save a file under `--reload`).
3. Startup recovery / the watchdog marks the execution `failed`.
4. `SELECT * FROM agent_activities WHERE activity_state='started'` — the `chat_start` row is still open, `related_execution_id` pointing at the now-terminal execution.
5. Dashboard Timeline renders the agent as in-progress until `started_at + 120min`, then as a 2-hour failure bar.

## Suggested fix

Move the responsibility from "whoever holds the local `activity_id`" to "whoever wins the CAS":

1. Call `get_open_activity_id_for_execution()` + `complete_activity()` inside both `_recover_execution` implementations (`cleanup_service.py:1991`, `:2378`) and after the two bulk sweeps (rows are already collected in `collect_failed` for the #1714 event emit — the same list can drive the activity close).
2. Add the missing CAS-loss branch to `_write_terminal_and_gate` (`:825`), mirroring the SUCCESS applier at `:1841`.
3. Preferably fold the lookup into the terminal-write contract so a recovery path cannot land a terminal without it — otherwise producer #4 repeats this. See the #429 comment: the pull sink (`pull_coordination_service.apply_task_result`) and the lease reaper's re-queue branch already have the same gap, and #429's success criterion (`grep mark_stale_* → zero hits`) would remove the 120-min backstop without adding an owner.

## Prior art

- #45 — tool-call activities created but never completed
- #767 — CB probe executions left open, inflating timeline failure duration

AndriiPasternak31:
Fixes #1804

## Problem

Every terminal writer **outside** the dispatching coroutine wrote the terminal to `schedule_executions` and walked away. The paired `agent_activities` row stayed `activity_state='started'` — the Dashboard Timeline rendered the agent as still working (`ReplayTimeline.vue`) until the generic 120-minute sweep closed it with a fabricated `duration_ms = now − started_at` that nothing ever recomputes. A 15-minute run was permanently recorded as a ~120-minute failure.

The close was gated on holding the `activity_id` **local** *and* winning the CAS **in the same coroutine**, so it failed two independent ways: process death (no `finally`, the id dies with the coroutine) and CAS loss (`_write_terminal_and_gate` had no lost-CAS branch, unlike the SUCCESS applier).

Third appearance of the class after #45 (tool-call activities) and #767 (CB probes) — both patched a single producer and left the ownership model alone.

## Fix — move the responsibility to whoever wins the CAS

**One owner:** `activity_service.close_execution_activity(...)` (+ a sync `spawn_close_execution_activity` wrapper for the pull sink), structurally the twin of `event_dispatch_service.spawn_task_terminal_event`. It maps the terminal via the shared `models.activity_state_for_terminal` (#1332 — never a second mapping), delegates to `complete_activity` so the `agent_activity` WS broadcast survives, and is fail-open: it runs after a committed terminal and can never affect it.

**The close is itself a CAS.** `db.complete_activity` now returns `ActivityCloseOutcome` (`UPDATED`/`ALREADY_CLOSED`/`NOT_FOUND`) — one bool cannot answer both "did this row exist" (`routers/internal.py` 404s on it) and "did anything change" (the broadcast gate), and once idempotent no-op closes are designed behaviour the two answers diverge routinely. "The CAS winner owns it" only holds if a *second* closer trying is safe.

**The lattice is the authority ordering, stated literally** (`started < failed < {completed, cancelled}`): an authoritative COMPLETED accepts `started|failed` (so the #1083 late-SUCCESS-after-lease-expiry path may upgrade a provisional FAILED); CANCELLED/FAILED accept `started` only. The COMPLETED arm is deliberately **tighter** than the execution row's own `!= CANCELLED` — see the review note below. The **lookup** is widened to agree (`include_failed=True` for authoritative terminals); a lookup narrower than the write would make the whole fix inert.

**Split by cardinality:** single-row recovery paths use the per-row helper (WS broadcast preserved); the two bulk sweeps use `db.close_open_activities_for_executions` (set-wise, one transaction, chunked). `_close_bulk_swept_activities` is a **sibling** of `_emit_bulk_terminal_events`, never folded into it — that method short-circuits on `has_task_terminal_subscribers() is False`, which would skip the close on every install with no event subscribers.

**Wired writers:** `_write_terminal_and_gate` (won **and** lost), `apply_result` (both branches), both backend-shutdown `CancelledError` handlers, watchdog + startup `_recover_execution` (the latter's CAS bool was previously *discarded*), the two bulk sweeps, the lease reaper (park → FAILED, re-queue → **CANCELLED**: a superseded attempt is not a failure), the pull sink, and `terminate_execution`.

The parity guard (`test_1804_terminal_activity_parity.py`) is anchored on **terminal writes**, not on #1578 event emission — the two backend-shutdown handlers write a terminal and emit nothing, which is exactly how they hid.

## Review findings fixed in-branch

1. **The lattice inherited the upstream's permissive edge — the bug it exists to prevent.** The predicate was specified as an *ordering* but implemented as an *exclusion* copied literally from the execution CAS (`!= 'cancelled'`), and those disagree exactly on the reflexive case: `!= 'cancelled'` matches `'completed'`. A second COMPLETED close re-dated `completed_at`/`duration_ms`/`error` — a 15-minute activity became ~9h, i.e. the #1804 symptom arriving through #1804's own fix, reachable from the lost-CAS branch (which passes an explicit `activity_id`, so the narrower lookup cannot shield it). The literal copy was *faithful*; the execution row survives that edge only because the #1083 callback's replay short-circuit stops the duplicate upstream — a protection the activity's call sites don't have.
2. **A CAS loss to a successful row** stamped `"superseded by ..."` onto a clean success and rendered the enum repr (`str, Enum`, the #1578 footgun). Now `error=None` on an authoritative SUCCESS close, `.value` for the label.
3. **`_recover_execution` returning the CAS bool changed what a `False` means** to `recover_orphaned_executions`, which counted every `False` as `errors`. A lost CAS is RELIABILITY-005 working as designed, so it's now partitioned into `cas_lost`; genuine exceptions self-count.

## Observability

`CleanupReport.activities_closed_on_recovery`. `stale_activities` should trend to ~0 while this picks up the volume — a non-zero `stale_activities` now means a producer is still unowned. The 120-minute backstop is demoted to *a backstop for the unclaimed* and moved to run **after** `_sweep_stale_slots` (it used to run one line before the reaper that legitimately closes activities, so within a single cycle the duration fabricator could beat a real closer). #429 deletes the backstop entirely — a contract survives that, per-site patches would not.

## Scope note

`db/schedules/cleanup.py::mark_no_session_executions_failed` now returns the **CAS-won** count rather than the candidate count (it previously over-reported `report.no_session_executions` by counting rows whose guarded UPDATE lost). Its sibling `mark_stale_executions_failed` already did this — a deliberate, visible value change.

## Verification

- `tests/unit/` full suite, **5143 passed / 14 skipped** (alphabetical order and `--randomly-seed=12345`); 58 new tests across 3 files.
- `tests/lint_sys_modules.py`: no new violations (`test_reset_preserve_state_guardrails.py` 9 → 0).
- Requirements-structure guard: clean. No schema change, no migration, no feature flag.

**Pre-existing flake, not from this branch:** under some orderings `test_1081_physical_meter.py::{test_max_respects_506_clamp,test_available_floors_at_zero_over_ceiling}[sqlite]` fail with `assert 5 == 2` — its `set_ceiling` monkeypatch doesn't reach `clamp_to_ceiling`. Attribution was done by replaying the *literal* failing order on pristine `origin/dev` (seeds aren't comparable across trees once the collected set changes): **dev fails identically**. This branch only shifts which seeds surface it. Worth a separate issue.


vybe (approved):
Validated via /validate-pr. Fixes #1804 with a closing keyword. The close-lattice CAS (db/activities.py::_close_predicate) mirrors the execution-row CAS with a correctly-argued tightening (no COMPLETED->COMPLETED re-date), close_execution_activity is the single owner with tri-state outcome semantics (ALREADY_CLOSED as designed no-op), and include_failed=True is applied exactly on the late-SUCCESS callback path. Named #1804 regression suites (R1-R5, chain, parity) present; docs complete; full pytest matrix + pg-migrations green; security scan clean. Will resolve the expected docs/routers conflicts against dev after #1867 lands, then merge.

Issue #78, opened 2026-03-06
Payment settings: Update button disabled after changing Plan/Agent ID, API key disappears


vybe:
## Bugs

Two related issues in the Nevermined payment settings panel (`NeverminedPanel.vue`):

### 1. Changing Plan ID or Agent ID does not enable the Update button

The `isFormValid` computed property (line 264) requires **all** fields to be non-empty, including `nvm_api_key`:

```javascript
const isFormValid = computed(() => {
  return form.value.nvm_api_key &&
    form.value.nvm_environment &&
    form.value.nvm_agent_id &&
    form.value.nvm_plan_id &&
    form.value.credits_per_request >= 1
})
```

But the API key is never loaded back from the server (it's encrypted server-side), so `nvm_api_key` is always empty after load. This means **any** field change still leaves the button disabled until the user re-enters the API key.

### 2. API key disappears and must be re-entered on every update

After a successful save, the API key is explicitly cleared (line 301):

```javascript
form.value.nvm_api_key = '' // Clear after save
```

Combined with the fact that `loadConfig()` never populates the API key field (lines 278-282), users must re-enter their API key every time they want to update **any** setting — even unrelated ones like Plan ID or credits.

## Expected Behavior

- Changing Plan ID, Agent ID, environment, or credits should enable the Update button without requiring the API key to be re-entered
- The API key field should show a masked placeholder (e.g., `••••••••`) when a key is already configured, indicating one is set
- Only require the API key field when initially configuring or explicitly changing the key

## Suggested Fix

1. Make `nvm_api_key` optional in `isFormValid` when a config already exists (`config.value` is non-null)
2. On the backend, treat an empty/missing `nvm_api_key` in the update payload as "keep existing key"
3. Show a placeholder in the API key field when a key is already stored (e.g., `sk-****...****`)
4. Track dirty state — only enable Update when at least one field actually changed from its saved value

## Files

- `src/frontend/src/components/NeverminedPanel.vue` — form validation and API key handling
- `src/backend/routers/nevermined.py` — backend config update endpoint

PR #1869, opened 2026-07-29
feat(auth): portal_delegate key scope + single-route fence (ent#163)

dolho:
The **OSS half** of delegated portal identity (`trinity-enterprise#163`). Core owns the scope, its containment, and the admin-only mint; the entitled module owns the endpoint that decides *whether* a given email may be delegated — the same split as `PORTAL_SESSION_SCOPE` (#78) and `users.suspended_at` (#995).

Pairs with `trinity-enterprise#164`. **Land this first** — the enterprise endpoint reads `User.portal_delegate`.

## What and why

A licensee wants to run their own customer portal with their own IdP, using Trinity as the agent backend. Their server holds one API key and needs to say *"this request is for bob@example.com"*.

**Why a new scope instead of reusing `scope='user'`:** the capability lets its holder act as another person and read their conversations. Riding it on an ordinary user key would silently turn **every** user key into a fleet-wide impersonation key.

**The fence is the load-bearing part.** A delegate key resolves to the **key owner**, exactly like every other MCP key — so unfenced it would simply *be* an admin's credential handed to a third party. `get_current_user` confines it to a single `(method, path)` at the one auth entry point, mirroring the connector fence (ent#46): centrally, not in the portal router, because the many endpoints that do inline access checks resolve this principal to the owner and would otherwise treat it as that human.

Deliberately **one route, not a prefix** — the minted portal session (not this key) drives the portal surface afterwards, so the key never needs breadth, and a prefix would silently grant every portal endpoint added in future.

**Minting is admin AND human-only.** `assert_admin` alone is insufficient: it rejects connector principals but not agent-scoped ones, and an agent key resolves to its owner *carrying the owner's role*, which on a default admin-owned install passes a bare role check (trinity-ops-agent#232). The db layer independently refuses any scope outside `{user, portal_delegate}`, so `agent`/`connector`/`system` keys — which carry an agent binding and are minted by their own paths — can never be forged through this endpoint (validate at the boundary **and** at the sink).

## Incidental fix

`create_mcp_api_key` now sets `is_active=1` explicitly instead of relying on a column default that exists in only one of the two schema sources: `schema.py` declares `is_active INTEGER DEFAULT 1`, but `db/tables.py` (the Core/Alembic source) declares a bare `Column("is_active", Integer)`. A table built from the metadata yields `NULL`, and `validate_mcp_api_key` treats falsy `is_active` as revoked — so every minted key would be born invalid. Harmless on both live schema paths; surfaced because it broke a test built from the metadata. Flagging it as a latent `tables.py`↔`schema.py` parity gap worth a separate sweep.

## Tests — `tests/unit/test_163_portal_delegate_scope.py` (19)

**Structural:** the fence is exactly one route (not a prefix), `User.portal_delegate` defaults to `False` (absence must mean "no"), the db creatable-set excludes agent-bound scopes.

**Behavioural**, driving the real `get_current_user` with a delegate key owned by an **admin** (the worst case — if the fence leaks, the caller *is* an admin principal):
- accepted on the exchange route, and still resolves to the owner (which is why the endpoint branches on the flag, not the user)
- **403 on** the fleet list, the user list, key minting, the portal surface itself, and the right path with the wrong method
- an ordinary `user` key is untouched on the very paths the delegate key is refused

Verified the behavioural half **fails (5 tests) with the fence disabled**, so it guards something.

## Verification

- `test_163_portal_delegate_scope.py` — 19 passed
- auth-surface suites (`test_1310_auth_wiring`, `test_1310_auth_consolidation`, `test_186_enumeration_uniformity`, `test_models_centralized`) — **103 passed**
- wider `-k "mcp or key or auth or portal"` selection — **692 passed, 4 skipped**
- enterprise-docs guard run locally — clean (the doc addition names the generic primitive only, never the module)

## Scope

`dependencies.py` (scope + fence), `models.py` (`User.portal_delegate`), `db_models.py` + `db/mcp_keys.py` (optional scope at mint), `routers/mcp_keys.py` (admin+human gate), `architecture.md` (scope table row + the DDL comment). No migration — `mcp_api_keys.scope` is an existing free-text column.

Related to trinity-enterprise#163

### DIFF
--- a/docs/memory/architecture.md
+++ b/docs/memory/architecture.md
@@ -1267,7 +1267,7 @@ CREATE TABLE mcp_api_keys (
     is_active INTEGER DEFAULT 1,
     user_id INTEGER NOT NULL,
     agent_name TEXT,                 -- non-null for agent-scoped keys
-    scope TEXT DEFAULT 'user',       -- user | agent | system
+    scope TEXT DEFAULT 'user',       -- user | agent | system | connector | portal_delegate
     FOREIGN KEY (user_id) REFERENCES users(id)
 );
 ```
@@ -1969,6 +1969,7 @@ Enforced at the **MCP server layer** (`src/mcp-server/src/tools/`), not the back
 | `user` | Owner/admin/shared checks | Owner/admin/shared checks |
 | `agent` | Explicit permission list (`agent_permissions`) | Resolves to owner user; ownership/sharing checks only |
 | `system` | **Bypasses all checks** | Resolves to owner user (system agent owner) |
+| `portal_delegate` | n/a (not an MCP tool principal) | **Fenced to a single route** — may only exchange an asserted end-user email for a portal session; every other path 403s (ent#163) |
 
 ### 7. External Credentials (Agent → External Services)
 
--- a/src/backend/db/mcp_keys.py
+++ b/src/backend/db/mcp_keys.py
@@ -90,10 +90,25 @@ class McpKeyOperations:
             scope=scope
         )
 
+    # ent#163: the scopes this endpoint may mint. `agent`, `connector` and
+    # `system` are deliberately absent — they are bound to an agent and are
+    # minted by their own code paths; accepting them here would let a caller
+    # forge an agent principal with no agent behind it.
+    _USER_CREATABLE_SCOPES = ("user", "portal_delegate")
+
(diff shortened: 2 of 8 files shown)


vybe (approved):
OSS seam for ent#163: portal_delegate key scope + single-route fence. Scope validated in dependencies.py at the single auth entry point (connector-fence pattern), unit test present (test_163_portal_delegate_scope.py), architecture.md updated. Security greps clean.

PR #1872, opened 2026-07-29
fix(validation): restore VALIDATE-001 + independent-referee seam (ent#277)

dolho:
The **OSS half** of cross-model validation (`trinity-enterprise#277`), plus a pre-existing bug I had to fix to get there.

Pairs with `trinity-enterprise#278`. **Land this first.**

## 1. VALIDATE-001 has never worked

`services/validation_service.py` calls `db.update_business_status(...)` and `db.create_validation_execution(...)` on the DatabaseManager facade. `ScheduleCleanupMixin` defines both and `ScheduleOperations` composes it — but **nothing re-exported them on the facade**, so `validate_execution` raised `AttributeError` on its first line, every time.

```
db.update_business_status:      False
db.create_validation_execution: False
```

It failed **invisibly**, which is why it survived: `POST /api/internal/.../validate` answers `{"status": "accepted"}` immediately and runs validation in a background task whose blanket `except Exception` reduces the crash to one log line. An operator who enabled `validation_enabled` on a schedule got an accepted request, no verdict, no operator-queue item, and a `business_status` that was never even set to `pending_validation` — precisely the silent failure validation exists to prevent.

Same class as the #1539 facade-delegation bug, and invisible to any test that mocks `db` wholesale. Fixed with both delegations, keyword-forwarded so a later signature change cannot rebind positionally.

## 2. The referee seam

The existing auditor runs as the **same agent**: same model, same provider, same container, workspace and tools reachable. Useful for "does the file really exist", but not independent — it shares the failure mode it is meant to catch, since a provider retuning the model drifts the auditor along with the executor.

`register_referee()` / `get_referee()` let an entitled module supply a referee that judges the same execution with a **different** model, called by the backend with only the transcript as input. That independence is the point (ent#205's core invariant), not the model change by itself.

Contract, all pinned by tests:

| Property | Why |
|---|---|
| Nothing registered ⇒ behaviour identical to before | OSS default is off |
| A verdict **replaces** the same-agent pass | Two validations double cost and latency, and on disagreement there is no principled tie-break — the whole argument for the referee is that the same-agent auditor is the one sharing the drift |
| The referee creates **no execution row** | Consumes no slot, no agent capacity; can't be mistaken for the agent's own work |
| Returns `None` **or raises** ⇒ degrade to the same-agent pass | An execution that already succeeded must never be failed because its validation was unavailable |
| A failing verdict still reaches the operator queue | AC: surfaced, never silently consumed |

`execution_id` is passed to the referee so a module can correlate its stored verdict — an identifier, not a capability; the referee still gets no workspace, tools or session. The contract test asserts the exact kwarg set, so widening it again is deliberate.

## Verification

- `tests/unit/test_277_referee_seam.py` — **7 passed**, including a regression test asserting both facade methods exist and that args forward by keyword
- The degradation tests prove fall-through by observing the original path being entered, not by asserting an absence

## Scope

`services/validation_service.py` (seam + fork), `database.py` (two missing delegations). No schema change, no migration, no new config.

Related to trinity-enterprise#277

### DIFF
--- a/src/backend/database.py
+++ b/src/backend/database.py
@@ -904,6 +904,54 @@ class DatabaseManager:
         return self._agent_ops.set_max_backlog_depth(agent_name, depth)
 
     # =========================================================================
+    # Business validation (delegated to db/schedules/cleanup.py) — VALIDATE-001
+    # =========================================================================
+    #
+    # These two were MISSING from the facade (found while building ent#277).
+    # `ScheduleCleanupMixin` defines them and `ScheduleOperations` composes it,
+    # but nothing re-exported them here — so every call in
+    # `services/validation_service.py` raised AttributeError on the first line
+    # of `validate_execution`.
+    #
+    # The failure was invisible: the internal endpoint answers "accepted"
+    # immediately and runs validation in a background task whose `except
+    # Exception` turns the crash into one log line. An operator who enabled
+    # `validation_enabled` got an accepted request, no verdict, no operator-queue
+    # item, and a `business_status` that was never even set to
+    # `pending_validation` — the very silent-failure mode validation exists to
+    # prevent. Same class as the #1539 facade-delegation bug: a mixin method
+    # nothing re-exported, which no test using a mocked `db` can see.
+    # Keyword-forwarded so a future signature change can't rebind positionally.
+
+    def update_business_status(
+        self,
+        execution_id: str,
+        business_status: str,
+        validation_execution_id: str = None,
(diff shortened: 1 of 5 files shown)


vybe (approved):
OSS seam for ent#277: restores VALIDATE-001 wiring + independent-referee seam (validation_service + CrossModelValidationPanel). Unit test present (test_277_referee_seam.py). Security greps clean; no config/packaging changes.

Issue #1874, opened 2026-07-29
ops: reclaim existing oversized container logs across deployed instances (follow-up to #1871)


webmixgamer:
## Summary

Follow-up to #1871, which bounds Docker's `json-file` logs going forward (`10m` × 3 for platform services via the compose `x-logging` anchor, and for agent containers via `AGENT_LOG_CONFIG`).

**Capping does not shrink what is already on disk.** `log_config` and the compose `logging:` block are applied at container **create** time, so:

- Platform services adopt the cap on the next `docker compose up`.
- Existing **agents** adopt it only when they are **recreated** — a plain restart is not enough.
- Either way, the bytes already written to `/var/lib/docker/containers/<id>/<id>-json.log` stay until that container is replaced.

The instance that motivated #1871 was carrying **1.4 GB** of unrotated container logs, with `trinity-vector` alone at ~468 MB accumulated since 2026-05-05. That reclaim is an operational task across deployed instances, not code in the #1871 PR — it was scoped out there deliberately and tracked here.

## Scope

Ops tooling / runbook to, per instance:

1. Report current usage — `sudo du -sh /var/lib/docker/containers` plus the top offenders by file size.
2. Confirm which containers still show `{"Type":"json-file","Config":{}}` (uncapped) via `docker inspect --format '{{json .HostConfig.LogConfig}}'`.
3. Reclaim, preferring the non-destructive route: recreate the container (platform services via `docker compose up -d`; agents via a cold start, which triggers the existing recreate path) so the new log file starts capped. Truncation (`: > <logfile>`) is the fallback where a recreate is undesirable — note it needs root and does not change the container's uncapped config.
4. Verify post-state: every running container reports a non-empty `LogConfig`, and the data root has dropped.

## Acceptance criteria

- [ ] A documented procedure exists in the ops runbook (or an ops-agent skill) covering report → confirm → reclaim → verify
- [ ] Applied to all deployed instances, with before/after `du -sh /var/lib/docker/containers` recorded
- [ ] Post-check: no running container reports an empty `LogConfig.Config`

## Out of scope

- The code fix itself (#1871)
- A host-level `/etc/docker/daemon.json` — defense-in-depth for the host, tracked separately if wanted; #1871 makes Trinity correct regardless of host config

Refs #1871

obasilakis:
Moving this to the ops tracker: **Abilityai/trinity-ops-agent#296**.

[...]

Issue #1432, opened 2026-07-03
bug: Vector docker_logs busy-loops on Docker Desktop — pegs the VM at ~4 cores, GB/day log churn; ship file-source local override


vybe:
## Summary

On Docker Desktop (macOS confirmed; Windows very likely — same VM-based architecture), the `trinity-vector` container's `docker_logs` source enters a reconnect busy-loop that pegs the Docker Desktop VM at ~4 cores (~386% CPU observed) within seconds of `./scripts/deploy/start.sh`, and re-ingests full container log history hundreds of times per second (platform log grows GB/day). Native Linux dockerd is unaffected, so production deployments are safe — but **every developer or evaluator who clones Trinity and runs the stack on a Mac hits this immediately**. It's a first-touch experience killer: fans max out and the laptop becomes unusable while the person is deciding whether to adopt.

## Root cause (confirmed, independently reproduced)

Interaction bug between two upstream components, exposed by Trinity's default dev config:

1. **Vector's `docker_logs` source** reconnects immediately whenever a log follow-stream ends. There is no retry/backoff or reconnect knob — reconnect-on-stream-end is hardcoded (verified across Vector 0.43 and 0.56).
2. **Docker Desktop's virtualized log relay** closes a `follow=1` stream right after flushing backlog instead of transitioning to follow. Probed directly against the in-VM dockerd socket:

| Request shape | Behavior |
|---|---|
| `follow=1&since=now&tail=0` (no backlog) | held open >6s ✅ (normal long-poll) |
| `follow=1&since=0&tail=all` (with backlog) | delivered full history (384–688 KB) then **closed at 4–17 ms** ❌ |

Vector treats stream-end as watch-end and instantly re-watches → the observed `Stopped watching`/`Started watching` loop (hundreds/sec in `docker logs trinity-vector`). Each reconnect re-dumps the container's entire log history through the virtualized log path; that re-serialization is what pegs dockerd/the VM, and re-ingesting history every cycle is why the platform log hits GB/day.

**Supporting evidence:**
- Vector running → VM host process at 386% CPU; Vector stopped → 3–7%. Vector's own cgroup CPU was only ~6.7% — the cost is in dockerd/the VM, not Vector itself.
- Quiet containers with large backlogs loop hardest (scheduler, redis were the top cyclers).
- `exclude_containers: [trinity-vector]` doesn't help (the loop targets other containers' backlogs); Vector version bump doesn't help.
[...]

Issue #858, opened 2026-05-15
bug: First-time setup token silently lost — print() block-buffered, breaks fresh installs


AndriiPasternak31:
## Summary

The first-time setup flow described in `docs/memory/architecture.md` is **silently broken** on fresh installs of `dev`. The setup token printed in `main.py:335-341` never reaches `docker logs`, so operators cannot complete the `/api/setup/admin-password` flow described in `routers/setup.py:54-78`. Every fresh Trinity deploy hits this on first boot.

This blocks the `trinity:deploy` skill (deploy plugin v2.2.0), which instructs operators to read the token from `docker compose logs backend`.

## Root cause

1. `docker/backend/Dockerfile` is missing `ENV PYTHONUNBUFFERED=1`. The scheduler Dockerfile already has it (see `docker/scheduler/Dockerfile`), but the backend Dockerfile drifted and lost it.
2. With `PYTHONUNBUFFERED` unset and stdout connected to a Docker log pipe (not a TTY), Python defaults to **block-buffered stdout** (~8KB). The lifespan's `print()` output (setup token plus ~30 other status prints) sits in the buffer and never flushes during normal operation.
3. Diagnostic confirmation inside the running container:
   ```
   $ docker exec trinity-backend env | grep PYTHONUNBUFFERED
   (empty)
   $ docker exec trinity-backend python -c "import sys; print('isatty:', sys.stdout.isatty()); print('line_buffering:', sys.stdout.line_buffering)"
   isatty: False
   line_buffering: False
   ```

## Repro

1. Fresh Trinity install (no existing DB, `setup_completed != 'true'`).
2. `./scripts/deploy/start.sh`.
3. `docker compose logs backend | grep -i "setup token"` → **nothing**.
4. Operator opens UI → setup screen demands a token printed in server logs → **token is invisible** → onboarding deadlock.

Other affected `print()` calls in the same lifespan (all silently lost): `Found N existing agent containers` (line 346), `System agent: ...` (line 355), `Using dedicated scheduler service` (line 369), `Log archive service started` (line 374), `Audit retention service started` (line 381), `DB vacuum service started` (line 388), and ~20 more.

## Fix (verified locally on this branch)

**Two layers, both required.**

### Layer 1 — Dockerfile (catches every `print()` call)

`docker/backend/Dockerfile`:
```diff
 FROM python:3.11-slim

 WORKDIR /app

+# Unbuffered stdout/stderr — otherwise Python block-buffers print() to the
[...]

PR #1875, opened 2026-07-29
fix(infra): bound container logs — json-file max-size/max-file everywhere (#1871)

### LINKED ISSUE #1871: bug: container logs are unbounded — no json-file max-size anywhere (compose, agent SDK, or daemon); filled the disk and wedged dockerd on 2026-07-27
## Summary

Trinity sets **no Docker log configuration anywhere** — not in `docker-compose.yml`, not in `docker-compose.prod.yml`, not on SDK-created agent containers, and not via a host `daemon.json` that any provisioning writes. Docker's default `json-file` driver has no `max-size` and no `max-file`, so every platform service and every agent container grows a `*-json.log` under `/var/lib/docker/containers/` forever.

Nothing fails while this happens. Then the Docker data root's filesystem hits 100%, and everything fails at once: partial writes leave zero-filled holes, dockerd can no longer parse its own logs (`invalid character '\x00' looking for beginning of value`), the daemon wedges, and the dashboard reports `0/0 agents`. That is what happened on 2026-07-27.

This is the **root cause** that #1799 was found next to but did not fix. #1799 fixed Vector's write permissions and the prod-compose duplicate-env validation failure — both discovered while diagnosing that same wedge. Neither bounds container log size.

## Why this was missed

Trinity has a working log-retention system, but it points at the **wrong set of files**. There are two copies of every log line on disk:

| | Path | Owner | Retention |
|---|---|---|---|
| Vector's aggregate | `/data/logs/*.json` | Trinity | `log_archive_service.py` (`LOG_DIR = Path("/data/logs")`, `LOG_RETENTION_DAYS`, daily archive+delete) |
| Raw Docker logs | `/var/lib/docker/containers/<id>/<id>-json.log` | Docker | **none** |

`grep -rn "var/lib/docker/containers" src/backend/` returns nothing. `docs/QUERYING_LOGS.md` documents rotation for the Vector copy, which makes the raw set feel handled. Three log-adjacent issues have shipped (#1432 churn, #1478 retention perms, #1799 Vector perms) and all three touch the Vector copy only.

## Evidence — a production instance (verified 2026-07-29)

```
$ sudo cat /etc/docker/daemon.json      → NO /etc/docker/daemon.json
$ sudo docker info                       → Driver=json-file  Ver=29.3.0  OS=Ubuntu 22.04.5 LTS
$ ls /etc/logrotate.d/ | grep docker      → (no docker logrotate rule)
$ sudo docker ps -q | xargs -I{} docker inspect --format '{{json .HostConfig.LogConfig}}' {} | sort | uniq -c
     21 {"Type":"json-file","Config":{}}      ← all 21 running containers, no cap

$ sudo du -sh /var/lib/docker/containers  → 1.4G
```

Largest files, and growth rates:

| Container | Size | Created | Rate |
|---|---|---|---|
| `trinity-vector` | 490,496,983 B (468 MB) | 2026-05-05 | ~5.8 MB/day |
| `trinity-frontend` | ~80 MB | | |
| `agent-A` | 66,642,128 B | 2026-07-20 | ~7 MB/day |
| `agent-B` | 66,321,308 B | 2026-07-20 | ~7 MB/day |
| `agent-C` | 66,280,944 B | 2026-07-20 | ~7 MB/day |

`trinity-vector` — whose entire job is aggregating everyone else's logs into the rotated copy — has the single largest unrotated file on the box.

Ubuntu/Debian `docker.io` and Docker CE ship no `daemon.json`, and our instance provisioning never writes one, so **every deployed instance is in this state** unless an operator hand-wrote one out-of-band.

**Honest scope note:** the instance measured is at 67% disk use, and 1.4 GB of logs is not what drove its two prior emergency disk resizes. This is a slow monotonic leak that never self-limits, not currently the dominant disk consumer. It is worth fixing because it has no ceiling, not because it is on fire today.

### 1. `docker-compose.yml` and `docker-compose.prod.yml` — add a shared anchor

Add above `services:` in **both** files:

```yaml
x-logging: &default-logging
  driver: json-file
  options:
    max-size: "${CONTAINER_LOG_MAX_SIZE:-10m}"
    max-file: "${CONTAINER_LOG_MAX_FILE:-3}"
```

Then add `logging: *default-logging` to each service.

- `docker-compose.yml` (9): `backend`, `frontend`, `redis`, `mcp-server`, `scheduler`, `logs-init`, `vector`, `otel-collector`, `postgres`
- `docker-compose.prod.yml` (9): `backend`, `frontend`, `redis`, `scheduler`, `logs-init`, `vector`, `mcp-server`, `cloudflared`, `otel-collector`

### 2. `src/backend/services/agent_service/capabilities.py` — agent containers

Compose's `logging:` **cannot** reach agent containers: they are created through the Docker SDK, not compose. This is the load-bearing half — the measurements above show agents are the fleet's fastest log writers.

Add `AGENT_LOG_CONFIG` beside the existing `AGENT_TMPFS_MOUNT`, following the same creation-time pattern (`os` and `re` are already imported at lines 18–19):

```python
_AGENT_LOG_MAX_SIZE_DEFAULT = "10m"
_AGENT_LOG_MAX_FILE_DEFAULT = "3"
_AGENT_LOG_MAX_SIZE_RE = re.compile(r"^\d+[kmg]$")
_AGENT_LOG_MAX_FILE_RE = re.compile(r"^[1-9]\d*$")


def _resolve_agent_log_opt(env_var: str, pattern: re.Pattern, default: str) -> str:
    raw = (os.getenv(env_var) or "").strip().lower()
    return raw if pattern.match(raw) else default


AGENT_LOG_CONFIG: dict[str, object] = {
    "type": "json-file",
    "config": {
        "max-size": _resolve_agent_log_opt(
            "AGENT_LOG_MAX_SIZE", _AGENT_LOG_MAX_SIZE_RE, _AGENT_LOG_MAX_SIZE_DEFAULT
        ),
        "max-file": _resolve_agent_log_opt(
            "AGENT_LOG_MAX_FILE", _AGENT_LOG_MAX_FILE_RE, _AGENT_LOG_MAX_FILE_DEFAULT
        ),
    },
}
```

Validation is deliberately fail-safe: a malformed value (empty, `10` with no suffix, `0` files) falls back to the bounded default rather than yielding an unbounded config. A typo must never silently disable rotation — same principle as #1638.

### 3. Thread it through all three container-create sites

Defined once in `capabilities.py` so the three paths cannot drift (same rationale as `AGENT_TMPFS_MOUNT`):

- `src/backend/services/agent_service/crud.py` — `_create_agent_container` (create path)
- `src/backend/services/agent_service/lifecycle.py` — recreate path; **this is what retrofits existing agents**, since `log_config` is creation-time
- `src/backend/services/system_agent_service.py` — `_create_system_agent`

Each imports `AGENT_LOG_CONFIG` from `.capabilities` and passes `log_config=AGENT_LOG_CONFIG` to the container-create call.

### 4. `.env.example` — document the knobs

Currently absent. Add `CONTAINER_LOG_MAX_SIZE`, `CONTAINER_LOG_MAX_FILE`, `AGENT_LOG_MAX_SIZE`, `AGENT_LOG_MAX_FILE` with their defaults, or they are inert-by-obscurity (the #1039 packaging-gap class).

## Rollout note

`log_config` is **creation-time**, exactly like `AGENT_TMPFS_MOUNT`. Existing agents keep their unbounded config until they are **recreated** — a plain restart is not enough. Platform services pick it up on the next `docker compose up`. Deployed instances also need their existing oversized files reclaimed once (recreate, or truncate), since capping does not shrink what is already on disk.

## Acceptance criteria

- [ ] `x-logging` anchor + `logging: *default-logging` on all 9 services in each compose file
- [ ] `AGENT_LOG_CONFIG` in `capabilities.py` with fail-safe validation
- [ ] `log_config` passed at all three create sites (create / recreate / system agent)
- [ ] Four env knobs documented in `.env.example`
- [ ] Verify: `docker inspect <agent> --format '{{json .HostConfig.LogConfig}}'` shows `max-size`/`max-file` on a freshly created agent **and** on a recreated pre-existing one
- [ ] Verify: `docker compose -f docker-compose.prod.yml config` still validates
- [ ] Fleet follow-up filed in ops tooling for reclaiming existing files across instances

## Out of scope

Setting a host-level `/etc/docker/daemon.json` during provisioning. That belongs in ops provisioning and would be defense-in-depth; this issue makes Trinity correct regardless of host configuration.

webmixgamer:
## Summary

Docker's `json-file` driver ships with **no** `max-size` and **no** `max-file`, so every platform and agent container log grew forever under `/var/lib/docker/containers/`. Nothing fails while it happens — then the Docker data root hits 100%, dockerd can no longer parse its own logs, and the whole fleet wedges at once (the 2026-07-27 incident).

Trinity's existing retention (`log_archive_service`, `LOG_RETENTION_DAYS`) governs only **Vector's aggregate copy** at `/data/logs`. The raw Docker copy had no owner at all.

Two halves, because compose cannot reach agents:

| | Surface | Knobs |
|---|---|---|
| **Platform services** | one shared `x-logging` anchor (the repo's first) on all **9** services in **both** compose files | `CONTAINER_LOG_MAX_SIZE` / `CONTAINER_LOG_MAX_FILE` |
| **Agent containers** | `AGENT_LOG_CONFIG` in `capabilities.py`, beside its sibling `AGENT_TMPFS_MOUNT`, threaded into the 3 create sites | `AGENT_LOG_MAX_SIZE` / `AGENT_LOG_MAX_FILE` |

Agents are SDK-created, so compose's `logging:` can never reach them — and they're the fleet's fastest log writers (~7 MB/day each, measured). `lifecycle.py`'s site is the shared tail of **both** `recreate_container_with_updated_config` and `recreate_missing_container`, so one constant covers every retrofit path.

## Design notes

**Validation is fail-safe in _both_ directions.** A format-only check is insufficient: `1000g` parses cleanly while effectively *removing* the cap — the exact failure the constant exists to prevent. So magnitude is bounded too (1g/file, 10 files) and zero is rejected. An explicitly-set rejection is reported via `print(flush=True)`, **not** `logger.*`, because `AGENT_LOG_CONFIG` resolves at **import** time, before `lifespan` calls `setup_logging()` (#858); an unset variable stays quiet.

**Default `10m` × 3 = 30 MB/container** (~4 days for a busy agent), deliberately in line with the aggregate's `LOG_RETENTION_DAYS=5`. The raw copy is *secondary* — Vector's `/data/logs` remains the primary queryable one.

**Rotation is lossless for live collection** — verified zero-gap over ~10 forced rotations on all three paths: the production `docker_logs` source, the #1432 `file` source, and the underlying daemon follow endpoint. Only post-hoc `docker logs` history shortens, and the platform's own log endpoint defaults to `tail=100` (~590x headroom).

**Creation-time semantics.** Platform services adopt on the next `docker compose up`; existing agents adopt on **recreate**, not restart. Capping does not shrink what is already on disk — fleet reclaim is tracked in #1874.

## The CI guard

`tests/unit/test_1871_log_config_parity.py` exists for the `learnings.md` (2026-07-10) *"the create path is never one call site"* class. It:

- scans **both** docker-py creation entry points — `containers_run()` **and** `containers.create()`. Covering only the first was the guard's own first blind spot, caught in review: `deploy.py` already used the second
- uses the **AST**, not a regex, so a *mention* of `containers_run(` in a comment is never a call site (`lifecycle.py`'s own comments do exactly that, and a regex draft flagged them)
- meta-tests its own machinery — planted violations for both call forms, unrelated `.create()` on volumes/networks, pre-filter-covers-every-call-form, allowlisted-paths-still-exist — so it cannot rot into a no-op

## Test plan

- [x] `pytest tests/unit/test_1871_container_log_rotation.py tests/unit/test_1871_log_config_parity.py` — 62 pass
- [x] No regressions across all 12 test files touching container creation (incl. `test_1484_create_agent_characterization.py`'s 40 tests and both #1816 suites) — 247 pass; 3 pre-existing harness failures reproduce identically on merge-base
- [x] `docker compose config` and `docker compose -f docker-compose.prod.yml config` both validate; env override honored
- [x] **Fresh agent create** → `{"Type":"json-file","Config":{"max-file":"3","max-size":"10m"}}`
- [x] **Recreated pre-existing agent** → same, with workspace volume intact (`clone_status: ok`, git HEAD unchanged)
- [x] Guard proven to catch a real regression at all 3 sites (line numbers correct)

## Docs

`architecture.md` (Container Security) · `requirements/infrastructure.md` §8.5c (LOG-001…006) · `feature-flows/container-capabilities.md` (new Container Log Rotation section) · `feature-flows/vector-logging.md` (rotation interaction) · `feature-flows/agent-lifecycle.md` · `learnings.md`

## Out of scope

- Host-level `/etc/docker/daemon.json` — ops provisioning, defense-in-depth; this makes Trinity correct regardless of host config
- Reclaiming existing oversized files across deployed instances — **#1874**

Closes #1871


obasilakis (approved):
Reviewed (/review + /cso daily, diff-scoped). No critical or informational blockers.

**Verified beyond the diff:**
- `log_config` dict form is a real docker-py contract — `HostConfig(log_config={'type':...,'config':...})` renders `{'Type': 'json-file', 'Config': {'max-size': '10m', 'max-file': '3'}}`, and `log_config` is in `RUN_HOST_CONFIG_KWARGS`.
- **Create-site coverage is complete.** All 10 `containers_run`/`containers.create` sites in `src/backend` accounted for: 3 durable agent creates capped; 6 ephemeral (`remove=True` — the alpine chown helpers in crud/lifecycle and the template.yaml read at lifecycle.py:995); `deploy.py:154` allowlisted with a real `finally: transient.remove(force=True)`. `src/scheduler/` creates no containers, so its exclusion is accurate.
- **Compose coverage is complete.** Rendered both files: dev 9/9 services carry `logging` (incl. `postgres` under its profile), prod 9/9. Both `docker compose config` invocations validate.
- 62/62 new tests pass locally; full suite green in CI across 6 base/head × seed runs.

**Security (diff-scoped, 8/10 gate): 0 findings.**
- A09 was the only plausible class — capping raw logs at 30 MB does not weaken the evidence chain. The authoritative trail is the append-only `audit_log` table (365-day floor, untouched here); the queryable copy is Vector's `/data/logs`, fed by the API-based `docker_logs` source that streams live and is indifferent to file rotation. Only redundant post-hoc `docker logs` history shortens, to ~4 days — deliberately aligned with `LOG_RETENTION_DAYS=5`. Net effect is a **reduction** in an availability failure mode (fleet-wide dockerd wedge), not a reduction in auditability.
- The operator-supplied values never reach a shell — they terminate in a docker-py HostConfig dict, regex-validated and magnitude-bounded, with `%r` on the raw value in the rejection path so a newline-bearing value can't forge log lines.
- No auth, network, Redis-ACL, dependency, or CI surface touched.

Two things I want to single out as genuinely good rather than merely correct: bounding **magnitude** and not just format (`1000g` parses cleanly while removing the cap — a format-only check would have shipped a control that a fat-finger silently disables), and the `.get()` on the unit table with the comment explaining that a future regex widening would otherwise `KeyError` at *import* time and crash-loop boot. That is the right instinct for a fail-safe constant.

Residual, non-blocking: `_TRANSIENT_CREATE_ALLOWLIST` is file-scoped, so a *second*, durable create added to `deploy.py` would be exempted silently. The docstring is honest about this and the file has one call site today — worth a call-line-scoped allowlist only if that file grows.

Issue #1258, opened 2026-06-18
Epic: First-run & self-host robustness — the OSS front door


vybe:
**Objective:** Clear the bug deck that breaks the OSS self-hoster's first 10 minutes — install, agent creation, and agents staying alive. Focused epic carved from the ready backlog (2026-06-18 prioritization pass).

**Audience:** OSS community / self-hosters — this is the front door.

## Sub-issues (all currently `status-ready`)
- #858 — First-time setup token silently lost (block-buffered `print`) — breaks fresh installs
- #1131 — Non-root backend can't reach `docker.sock` on macOS Docker Desktop — agents silently vanish (#874 regression)
- #689 — Duplicate mount-point error on agent restart with custom_mounts
- #1231 — `/tmp` tmpfs (100 MB) fills and silently breaks autonomous git commits — make size configurable
- #1197 — Agent creation crashes with opaque ValueError on non-integer CPU in template resources
- #958 — Build Info dialog displays 'unknown' for every field in local development
- #953 — Freshly deployed agents report `M .gitignore` against origin/main

**Suggested order:** #858 → #1131 → #1197 (install / create path) → #689 → #1231 (runtime survival) → #958 → #953 (noise/cosmetic).

webmixgamer:
## Inert-lever audit: documented `.env` knobs that don't reach the backend

> **CORRECTED 2026-07-29** — the original version of this comment listed `TRINITY_GIT_BASE_URL` / `TRINITY_GIT_API_BASE` as "fully inert everywhere." **That was wrong** — they are working as designed (see *Retracted* below). Two detector flaws caused it. Corrected counts and findings follow; the canary finding survived re-verification and was strengthened.

Found while fixing #1871 — which shipped **this exact class** and was caught by `/validate-pr` §4.9 before merge (PR #1875): `AGENT_LOG_MAX_SIZE` / `AGENT_LOG_MAX_FILE` were documented in `.env.example` and read via `os.getenv()`, but never wired into `backend.environment:` in either compose file, so both levers silently did nothing. Fixed there.

Having written the detector, I ran it across the backend on `dev@a8418a2f`.

### Method

A var is a **documented inert lever** when it is (1) read by the backend via `os.getenv()`, (2) documented **and uncommented** in `.env.example`, and (3) absent from `backend.environment:` in a compose file that actually deploys it. Prod compose launches **standalone** — verified: **no `env_file:` on any service in either file**, so the explicit `environment:` list is the only path in.

[...]

Issue #751, opened 2026-05-10
fix: backend Dockerfile missing COPY for canary/ package (a4eec13 breaks startup)


vybe:
## Summary

Commit `a4eec13` (feat(#411): canary invariant harness Phase 1) added `src/backend/canary/` but did not add a corresponding `COPY` line to `docker/backend/Dockerfile`. The backend container fails to start with:

```
File "/app/routers/canary.py", line 21, in <module>
    from canary import INVARIANTS
ModuleNotFoundError: No module named 'canary'
```

## Root Cause

`docker/backend/Dockerfile` copies source directories explicitly. The new `canary/` package was not added:

```dockerfile
# Existing copies:
COPY ../../src/backend/routers /app/routers/
COPY ../../src/backend/services /app/services/
COPY ../../src/backend/db /app/db/
# Missing:
COPY ../../src/backend/canary /app/canary/   ← not present in a4eec13
```

## Fix

Add to `docker/backend/Dockerfile` after the existing COPY block:

```dockerfile
COPY ../../src/backend/canary /app/canary/
```

## Impact

construct instance attempted to deploy `a4eec13`, backend failed to start, rolled back to `7b95c46`. No data loss.

## Reproduction

```bash
git checkout a4eec13
docker compose build --no-cache backend
docker compose up -d backend
docker logs trinity-backend   # ModuleNotFoundError: No module named 'canary'
```

dolho:
Closing — already fixed by commit `62cc7495` ("fix(docker): add missing COPY for canary/ package in backend Dockerfile (#768)"). Same bug, different PR number. The current `docker/backend/Dockerfile` (line 57) has `COPY ../../src/backend/canary /app/canary/` and the chmod block (lines 65–66) includes the package. Reopen if the breakage came back.

dolho:
Correction on the earlier comment — premature close. Fix `62cc7495` is merged to **dev** only, not yet to main. Reopening + relabeling `status-in-dev` per the team rule that issues close only on main merge (i.e. at release cut).

vybe:
Shipped in **v0.6.0** (released via #998).

Issue #411, opened 2026-04-19
Research and implement orchestration invariant catalog + continuous canary harness


vybe:
## Summary

Design and build a continuous **invariant-based testing harness** for Trinity's orchestration layer, running against staging/dev. Goal: prove orchestration invariants hold 24/7 under real traffic and adversarial conditions — catching bugs like #378/#403 (phantom stale-slot failures), #407 (subprocess reaping), and #129 (orphaned executions) *before* users see them.

Full design: [`docs/testing/orchestration-invariant-catalog.md`](../blob/main/docs/testing/orchestration-invariant-catalog.md)

## Motivation

Recent orchestration bugs all violated obvious-in-hindsight invariants:

- #378/#403 — *"a terminal execution state must be immutable"*
- #407 — *"every subprocess must be reapable on every exit path"*
- #129 — *"every `running` execution maps to either an agent registry entry or a cleanup action within one cycle"*
- #219/#226 — *"slot TTL ≥ execution timeout"*

Unit tests don't catch these because they live in cross-component state — Redis × SQLite × Docker × agent process registry. A continuous harness that checks invariants on a live staging instance would have caught all of them.

## Proposed approach

See the full catalog in `docs/testing/orchestration-invariant-catalog.md`. Summary:

**Framework layers** (bottom-up):
1. Invariant library (~55 invariants across 14 subsystems, each a pure `(snapshot) → ViolationReport`)
2. Snapshot collector (SQL + Redis + Docker + agent registry, roughly simultaneous)
3. Scenario runner (pytest-style, exercises flows + asserts invariants)
4. Continuous canary (reads-only subset every 5–15 min, alerts on violation)
5. Chaos layer (opt-in, container kill / Redis partition / socket lag)

**Tiering:** Tier A (must hold always) vs. Tier B (must reconcile within SLA T). Severity: 🔴 critical / 🟡 major / 🟢 minor.

## Starting subset (Phase 1 — 10 invariants, ~80% coverage)

| ID | Invariant | Reason |
|----|-----------|--------|
| S-01 | Slot–row bijection | Core orchestration |
| S-02 | No overbooking | Capacity |
| E-01 | Terminal-state closure | No stuck executions |
| E-02 | No phantom reversal | #378/#403 |
| E-06 | No completed-but-not-reported | #129 |
| B-01 | Queue-status coherence | Backlog integrity |
| B-02 | No queued without slots-full | Drain liveness |
| L-03 | Delete cascades | Dangling references |
[...]

PR #1876, opened 2026-07-29
fix(canary): wire CANARY_ENABLED/_SLACK_WEBHOOK_URL into prod backend env

### LINKED ISSUE #1878: bug: canary harness cannot be enabled on any deployed instance (CANARY_ENABLED absent from prod compose)
## Summary

`CANARY_ENABLED` and `CANARY_SLACK_WEBHOOK_URL` are wired into `docker-compose.yml` only. **Every deploy path uses `docker-compose.prod.yml`** (`.github/workflows/deploy-dev.yml`, `scripts/deploy/gcp-deploy.sh`), and prod compose launches **standalone** — no base-compose merge and no `env_file:` on any service — so the explicit `environment:` list is the only path into the container.

A `CANARY_ENABLED=1` in the deployed `.env` (which `gcp-deploy.sh` does write) is interpolated into nothing and silently dropped. **The CANARY-001 harness cannot be enabled on any deployed instance**, and its Slack alert sink is unconfigurable, so even a hand-enabled canary would be a silent sink.

Epic #411 was closed `COMPLETED` on 2026-07-28 at 53/53 sub-issues while this was true.

## Evidence

```
$ CANARY_ENABLED=1 VOIP_ENABLED=true docker compose -f docker-compose.prod.yml config
  CANARY_ENABLED           -> None    *** never reaches container ***
  CANARY_SLACK_WEBHOOK_URL -> None    *** never reaches container ***
  VOIP_ENABLED             -> 'true'  (control: wired, so it reaches)
```

The positive control confirms the method.

## Was the omission deliberate? No

`.env.example` says *"Set to 1 on staging/dev … Production stays 0"*, which reads like intent — but:

1. **The intent was unachievable from day one.** `deploy-dev.yml` was created 2026-04-28 already using prod compose; the canary landed 2026-05-10, 12 days later. The deploy path did not change underneath it.
2. **The same commit shipped another packaging gap** — #751, missing `COPY` for `canary/` in the backend Dockerfile, which crash-looped startup.
3. **"Production stays 0" is about the default**, which the fix preserves — unset still resolves to `CANARY_ENABLED='0'`.
4. **Convention is unambiguous**: prod compose already carries eight default-OFF opt-in flags (`VOIP_ENABLED`, `WORKSPACE_ENABLED`, `OTEL_ENABLED`, `DISPATCH_ASYNC`, …). The VOIP block even states that omitting it *was* the prod packaging gap (#1056).

## Acceptance criteria

- [x] Both vars injected into `docker-compose.prod.yml` `backend.environment:`, mirroring `docker-compose.yml`
- [x] Unset resolves to `CANARY_ENABLED='0'` / `CANARY_SLACK_WEBHOOK_URL=''` — no behaviour change for existing installs
- [x] `docker-compose.prod.yml` and the prod+enterprise overlay both validate
- [x] Regression guard so a revert fails CI rather than silently reverting the harness (`tests/unit/test_canary_env_prod_parity.py`)

## Context

Found by the inert-lever audit on #1258, written while fixing the same class in #1871. That audit flagged 11 candidates; hand-verification retracted 2 as false positives (`TRINITY_GIT_*`, correctly handled by the `docker-compose.gitea.yml` overlay). **This is the only one that proved to be a genuine, user-visible defect.**

Fixed by PR #1876.

Refs #1258
Refs #411

webmixgamer:
Fixes #1878

## Summary

**The canary invariant harness could not be enabled on any deployed instance.**

`architecture.md` documents it as *"Disabled by default; `CANARY_ENABLED=1` on staging/dev"* — but every deploy path uses `docker-compose.prod.yml` (`.github/workflows/deploy-dev.yml`, `scripts/deploy/gcp-deploy.sh`), and that file never passed the two knobs into `backend.environment:`.

Prod compose launches **standalone** — no base-compose merge, and no `env_file:` on any service in either file — so the explicit `environment:` list is the only path in. A `CANARY_ENABLED=1` in the remote `.env` (which `gcp-deploy.sh` does write) was interpolated into nothing and silently dropped.

Net effect: **epic #411 was closed `COMPLETED` on 2026-07-28 at 53/53 sub-issues**, while the harness it delivered was un-turnable-on on the very instance it exists to watch — and its Slack alert sink was unconfigurable, so even a hand-enabled canary would have been a silent sink. `config/canary-fleet.yaml` targets a running instance's `/api/systems/deploy`, so this was never a laptop-only path.

## Change

Two lines in `docker-compose.prod.yml` `backend.environment:`, mirroring `docker-compose.yml` verbatim. No code change.

## Verification

Same test that proved the bug, with a positive control so the method itself is validated:

```
$ CANARY_ENABLED=1 CANARY_SLACK_WEBHOOK_URL=… VOIP_ENABLED=true \
    docker compose -f docker-compose.prod.yml config

before:  CANARY_ENABLED -> None    CANARY_SLACK_WEBHOOK_URL -> None    VOIP_ENABLED -> 'true'
after:   CANARY_ENABLED -> '1'     CANARY_SLACK_WEBHOOK_URL -> 'https://…'  VOIP_ENABLED -> 'true'
```

- [x] **Additive-safe** — unset resolves to `CANARY_ENABLED='0'` / `CANARY_SLACK_WEBHOOK_URL=''`, so existing installs see no behaviour change
- [x] `docker compose -f docker-compose.prod.yml config` validates
- [x] `docker compose -f docker-compose.prod.yml -f docker-compose.prod.enterprise.yml config` validates (the enterprise overlay layers on prod, so it inherits the fix)
- [x] dev/prod parity now complete for both vars — and these are the **only** two `CANARY_*` vars the backend reads
- [x] No other compose file needs it — `.gitea.yml` is an unrelated overlay; `.override*.yml` and `.sibling.yml` define no `backend` service

## Provenance

Found by an inert-lever audit posted on #1258, written while fixing the **same class** in #1871 (where `/validate-pr` §4.9 caught `AGENT_LOG_MAX_*` before merge). This is the #1039/#1056 packaging-gap class: a documented `.env` lever that silently does nothing.

That audit initially flagged 11 candidates; hand-verification retracted 2 as false positives (`TRINITY_GIT_*`, correctly handled by the `docker-compose.gitea.yml` overlay). **This is the only one that turned out to be a genuine, user-visible defect** — the remaining items are dev-compose parity and a docs deletion, both noted on #1258 rather than filed.

Deliberately no closing keyword: #1258 is an epic and #411 is already closed, so neither should be auto-closed by this PR.

## Not doing here

No CI guard. A naive "every `os.getenv` must be in compose" check fails on 56 pre-existing vars, most legitimately code-default-only. The tractable predicate — *read **and** uncommented in `.env.example` **and** not forwarded by any compose file **including overlays*** — is viable once #1258's short list is burned down; see the audit comment for why the overlay and comment-line cases must both be handled.

Refs #1258
Refs #411

---

## "Was the omission deliberate?" — investigated, no

Fair challenge, since the originating commit's own wording sounds like intent. `.env.example` (added by the same commit `a4eec13`) says:

```
# CANARY INVARIANT HARNESS (Optional, staging/dev)
# Set to 1 on staging/dev to run the 5-min check loop. Production stays 0.
```

Four independent lines of evidence say the *omission from prod compose* was an oversight, not enforcement:

**1. The intent was unachievable from day one.** `.github/workflows/deploy-dev.yml` was created **2026-04-28** already using `docker-compose.prod.yml`. The canary landed **2026-05-10** — 12 days later. So "set to 1 on staging/dev" could never work on the deployed staging/dev instance; the deploy path did not change underneath it. Reading the omission as deliberate enforcement would mean the feature was dead on arrival everywhere, which is incoherent for an epic that also shipped a load-generator fleet manifest and a Slack alert sink.

**2. The same commit shipped another packaging gap.** #751 — *"backend Dockerfile missing COPY for canary/ package (a4eec13 breaks startup)"* — the very same commit omitted a required `COPY`, crashing the backend on boot. That PR demonstrably had packaging blind spots; two is unsurprising.

**3. "Production stays 0" is about the default, and the fix preserves it.** The variable being *present with an OFF default* is not the same as enabled. With both unset, the resolved prod config is still `CANARY_ENABLED='0'` and `CANARY_SLACK_WEBHOOK_URL=''` — verified above. Production users see no canary activity unless they deliberately opt in, exactly as written.

**4. Codebase convention is unambiguous.** `docker-compose.prod.yml` already carries **eight** default-OFF opt-in flags: `WORKSPACE_ENABLED`, `VOIP_ENABLED`, `MCP_AGENT_CHAT_PULL_ENABLED`, `REDELIVERY_GOVERNOR_ENABLED`, `OTEL_ENABLED`, `PUBLIC_ACCESS_REQUESTS_ENABLED`, `DO_NOT_TRACK`, `DISPATCH_ASYNC`. A default-OFF flag belongs in prod compose *with* its OFF default.

**And this exact question has already been adjudicated in this file.** The VOIP block, two lines above where the canary vars now sit:

```
# VoIP telephony (VOIP-001, #1056) — opt-in, default OFF. Default-OFF means
# the master switch MUST reach the container or voip_available stays false;
# omitting it here was the prod packaging gap (same class as #1039 LOG_*).
```

Identical situation, resolved as a packaging gap. This PR applies the same resolution.

---

## Added after `/validate-pr`

**Tracking (W2)** — filed **#1878** as a sub-issue of #1258 and switched to a closing keyword. Without it nothing transitioned `status-in-progress → status-in-dev`, so `/release` (which builds notes from in-dev issues) would have shipped a user-visible fix silently.

**Regression guard (W1)** — `tests/unit/test_canary_env_prod_parity.py`. The fix is two compose lines that nothing asserted: delete them and every test still passed, which is precisely how the original omission survived ~2.5 months and an epic close-out. The guard pins, for **both** compose files and **both** knobs:

- the injection line exists (absence ⇒ the lever is inert on every deploy)
- it defaults to **OFF** (presence must not mean enabled — a future `:-1` would start a 5-min watcher loop *and its synthetic agent fleet* on every install)
- dev and prod agree (the drift **is** the bug)
- plus a meta-test proving the matcher fails on pre-fix content, so a typo can't make the assertions vacuously true

Verified it actually bites: with the fix reverted in-memory, the matcher drops from 1 injection line to 0 and the guard fails. Shape mirrors `test_1076_voice_model_config.py`, the existing precedent for guarding a compose injection against a silent revert.

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -247,6 +247,19 @@ services:
       # packaging-gap class).
       - TELEMETRY_CONTAINER_STATS_TTL=${TELEMETRY_CONTAINER_STATS_TTL:-10}
       - TELEMETRY_DOCKER_POOL_SIZE=${TELEMETRY_DOCKER_POOL_SIZE:-16}
+      # Canary invariant harness (CANARY-001 / Issue #411). Mirrors
+      # docker-compose.yml; without these lines the .env knobs are inert in prod
+      # (the #1039 packaging-gap class). Load-bearing here specifically: the
+      # harness is documented as "CANARY_ENABLED=1 on staging/dev", and the dev
+      # instance deploys via THIS file (.github/workflows/deploy-dev.yml,
+      # scripts/deploy/gcp-deploy.sh) — so omitting them made the watcher
+      # un-enableable on the very instance it exists to watch.
+      # Default 0 — production users see no canary activity.
+      - CANARY_ENABLED=${CANARY_ENABLED:-0}
+      # Slack alert sink for canary green→red transitions. URL is the
+      # credential — leaking it lets anyone post to that one channel.
+      # Unset = canary cycles run silently (still persists violations).
+      - CANARY_SLACK_WEBHOOK_URL=${CANARY_SLACK_WEBHOOK_URL:-}
       # SSH access host override (read by services/ssh_service.py). Mirrors
       # docker-compose.yml; without this line the .env value is inert in prod (the
       # #1039 packaging-gap class) and SSH host detection falls back to localhost.
--- a/tests/unit/test_canary_env_prod_parity.py
+++ b/tests/unit/test_canary_env_prod_parity.py
@@ -1,0 +1,123 @@
+"""Regression guard: the canary env knobs must reach the backend in BOTH composes.
+
+Bug: `CANARY_ENABLED` / `CANARY_SLACK_WEBHOOK_URL` were wired into
+`docker-compose.yml` only. Every deploy path uses `docker-compose.prod.yml`
+(`.github/workflows/deploy-dev.yml`, `scripts/deploy/gcp-deploy.sh`), and prod
+compose launches **standalone** — no base-compose merge and no `env_file:` on any
+service — so the explicit `environment:` list is the only path into the
+container. A `CANARY_ENABLED=1` in the deployed `.env` was interpolated into
+nothing and silently dropped, leaving the CANARY-001 harness (#411)
+un-enableable on the very instance it exists to watch.
+
+This is the #1039/#1056 packaging-gap class: a documented `.env` lever that
+silently does nothing. It has now recurred five times (#1056 `VOIP_*`,
+trinity-enterprise#31 `LOG_*`, #1039, #1871 `AGENT_LOG_*`, and this), so the
+fix is pinned rather than trusted — deleting the compose lines must fail CI,
+not pass silently.
+
+Shape mirrors `test_1076_voice_model_config.py::test_compose_default_is_non_empty`,
+the existing precedent for guarding a compose injection against a silent revert.
+
+Lives under tests/unit/ so the CI unit job (`cd tests && pytest unit/`) collects
+it — a guard must run where the bug would regress.
+"""
+from __future__ import annotations
+
+from pathlib import Path
+
+import pytest
+
+# tests/unit/<this file> → parent=tests/unit, .parent=tests, .parent=repo root
+_REPO_ROOT = Path(__file__).resolve().parent.parent.parent
+
+_COMPOSE_FILES = ["docker-compose.yml", "docker-compose.prod.yml"]
+
+# (var, the OFF default the injection must carry)
+_CANARY_VARS = [
+    ("CANARY_ENABLED", "0"),
+    ("CANARY_SLACK_WEBHOOK_URL", ""),
+]
+
+
+def _injection_lines(compose_file: str, var: str) -> list[str]:
+    """Uncommented `- VAR=${VAR:-…}` lines for `var` in `compose_file`."""
+    text = (_REPO_ROOT / compose_file).read_text(encoding="utf-8")
+    return [
+        ln
+        for ln in text.splitlines()
+        if f"{var}=${{{var}:-" in ln and not ln.strip().startswith("#")
+    ]
+
+
+@pytest.mark.parametrize("compose_file", _COMPOSE_FILES)
+@pytest.mark.parametrize("var,_default", _CANARY_VARS)
+def test_canary_var_is_injected(compose_file, var, _default):
+    """Both canary knobs must be injected in BOTH compose files.
+
+    The prod file is the load-bearing one: it is what every deploy path uses,
+    so an omission there makes the knob inert on every real instance while
+    still working on a developer's laptop — which is exactly how this shipped
+    unnoticed for ~2.5 months.
+    """
+    lines = _injection_lines(compose_file, var)
+    assert lines, (
+        f"{compose_file}: no `{var}=${{{var}:-…}}` injection under "
+        f"backend.environment. Without it the .env lever is INERT on any deploy "
+        f"using this file (the #1039/#1056 packaging-gap class)."
+    )
+
+
+@pytest.mark.parametrize("compose_file", _COMPOSE_FILES)
+@pytest.mark.parametrize("var,default", _CANARY_VARS)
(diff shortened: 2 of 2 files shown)


obasilakis (approved):
Reviewed via /review + /cso — no critical or security findings.

**Independently verified** (not taken from the PR body):
- Backend genuinely reads both vars — `canary_service.py:130`, `canary_alerts.py:134`
- Prod compose has no `env_file:` on any service, so the explicit `environment:` list really was the only path in
- Post-fix resolution: `CANARY_ENABLED -> '1'`, webhook URL reaches the container
- Inert by default: unset resolves to `'0'` / `''` — no behaviour change for existing installs
- prod and prod+enterprise overlay both validate
- Guard bites: reverting the two compose lines fails 3 tests; `backend-unit-test.yml:116` collects `tests/unit/`

**Security note.** This makes a Slack webhook URL — a bearer credential — settable in production for the first time, so the leak path matters now. It is already hardened: `slack_service.py:346-351` returns `type(e).__name__` rather than `str(e)` specifically because httpx embeds the request URL in exception messages, and the URL appears in no log line or API response. Given Vector persists all logs to disk, that guard is load-bearing and intact. Payload contents (agent names, execution ids, counts) carry no credentials or user content; the `backlog_metadata`-touching invariants (E-04/G-04) have no `_render_forensic` branch and fall through to a generic message — a completeness gap that fails in the safe direction.

**Non-blocking nits:**
1. `_injection_lines` scans the whole file while the failure message claims "under backend.environment" — a line moved to the wrong service would pass. Narrow case; worth tightening if this file is touched again.
2. Scope is correctly limited to the canary vars, with the broader inert-lever sweep tracked on #1258.

LGTM.

Issue #182, opened 2026-03-26
SEC: Agent Subscription Management by Shared Users (BOLA)


vybe:
## Pentest Finding 3.2.5 — Medium (CVSS 4.8)

**CVSS Vector**: [CVSS:4.0/AV:L/AC:L/AT:N/PR:L/UI:N/VC:N/VI:L/VA:L/SC:N/SI:N/SA:N](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:L/AC:L/AT:N/PR:L/UI:N/VC:N/VI:L/VA:L/SC:N/SI:N/SA:N)

**Location**: `PUT /api/subscriptions/agents/{name}`, `DELETE /api/subscriptions/agents/{name}` — `routers/subscriptions.py` lines 184, 218

## Impact

A user with shared read access to an agent can:
- Assign any subscription (Claude API OAuth token) to that agent
- Clear its subscription assignment
- Both operations trigger a container restart

This enables hot-swapping Claude credentials, DoS by clearing the subscription, and cross-account subscription abuse (unexpected billing/rate-limit consumption).

## Root Cause

`routers/subscriptions.py` line 184 uses the permissive `can_user_access_agent` check (owners + admins + shared users) instead of `can_user_share_agent` (owners + admins only). The endpoint docstring states "Owner access required" but the implementation enforces only "any access". Same issue at line 218 for DELETE.

## Remediation

- Replace `db.can_user_access_agent` with `db.can_user_share_agent` (owner/admin only) in both PUT and DELETE handlers
- Audit all other mutation endpoints for the same authorization predicate confusion
- Add test asserting non-owner shared users receive HTTP 403 on subscription mutation endpoints

## References

- [OWASP API: Broken Object Property Level Authorization](https://owasp.org/API-Security/editions/2023/en/0xa3-broken-object-property-level-authorization/)

_Source: UnderDefense Web Pentest Report, March 2026_

Issue #178, opened 2026-03-26
SEC: WebSocket Accepts Unauthenticated Connections


vybe:
## Pentest Finding 3.2.1 — Medium (CVSS 6.9)

**CVSS Vector**: [CVSS:4.0/AV:L/AC:L/AT:N/PR:N/UI:N/VC:H/VI:N/VA:N/SC:N/SI:N/SA:N](https://www.first.org/cvss/calculator/4.0#CVSS:4.0/AV:L/AC:L/AT:N/PR:N/UI:N/VC:H/VI:N/VA:N/SC:N/SI:N/SA:N)

**Location**: `ws://host/ws` — `main.py` lines 355–364

## Impact

The WebSocket endpoint accepts TCP upgrade and establishes a connection without any authentication. An unauthenticated attacker can receive real-time platform activity broadcasts including:
- Agent status changes, task progress, activity feed events
- Agent names, activity types, task identifiers
- PII leaks (e.g., `shared_with` email addresses in `agent_shared` events)
- Process execution details and error messages

## Root Cause

`main.py` lines 355–364 call `websocket.accept()` before performing any authentication check.

## Steps to Reproduce

A PoC script (`ws_vuln_tests.py`) demonstrates that connecting without auth yields real-time events:
- `agent_activity` events with model names, execution IDs, message previews
- `agent_shared` events with PII (`shared_with: user2@user.com`)
- `process_event` events with process names and execution details

## Remediation

- Reject WebSocket upgrade requests that do not provide a valid bearer token before calling `websocket.accept()`
- If browser clients cannot send Authorization headers during WebSocket upgrade, implement a short-lived ticket pattern: exchange a JWT for a one-time WebSocket ticket via an authenticated POST, then use the ticket in the WS URL with a 30-second TTL

## References

- [WebSocket Security Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/WebSocket_Security_Cheat_Sheet.html)
- [CWE-306: Missing Authentication for Critical Function](https://cwe.mitre.org/data/definitions/306.html)

_Source: UnderDefense Web Pentest Report, March 2026_

PR #1877, opened 2026-07-29
feat(ui): Skills management surface — unhide and rebuild the Skills tab (ent#235)

dolho:
Implements `trinity-enterprise#235`. **OSS-core, ungated** — confirmed with the issue author before building: every file here is already public, the endpoints are already ungated, and the paid piece (`skill_runner`, ent#139) plus exposure curation (#178) are both explicitly out of scope.

## The gap

The skills machinery shipped across three planes (#182 distribute/place/expose, #183 package injection with a per-skill result contract) and **nothing rendered any of it**. The tab was excluded from `visibleTabs` per requirements §22.2 ("component preserved for potential admin-only access"), assignment was REST/MCP-only (§21.3), and #183's statuses and named warnings had no consumer. A user could not browse the library, see what an agent had, or assign anything.

## What lands

**Tab unhidden** for owners/admins on non-system agents, matching the other management tabs; `OverflowTabs` absorbs it.

**`stores/skills.js`** — a domain store (Invariant #6). Worth flagging: the old panel called `axios` directly with a hand-built auth header, silently bypassing the shared client every other call relies on. Everything now goes through `api` (Invariant #7).

**Library browse** with the §21.6 contract: description, automation, `user_invocable`, declared `requires`, multi-file count, size, and the git tree SHA as version. Dependencies are shown **before** assignment, because they're exactly what later becomes a `missing_binary:*` warning.

**Assignment** via the existing `PUT .../skills` with bulk save, plus dirty/reset so a half-made selection is recoverable.

**Honest injection status — the load-bearing part.** #183 reports `injected | unchanged | fallback | failed` with named warnings. The panel renders the verdict per skill and translates tokens into what they mean for *this* agent (`missing_binary:jq` → "`jq` is not installed in this agent — the skill may not run"). `fallback` renders as **"partial" in amber, never a green tick** — an explicit AC. Injection results are held separately from assignment in the store precisely so durable assignment state can't be painted with a stale success.

**Manual sync** (`force=True` repair action) with in-flight state; a 409 from `SkillInjectionBusy` is reported as "already running", not a generic failure.

**No dead empty states** — the store computes one discriminator (`library_unconfigured` / `library_empty` / `none_assigned`) so the panel can't invent a fourth. Unconfigured routes an admin to Settings and tells a non-admin to ask one.

**Stopped agent** renders persisted assignment with Sync disabled and the reason in the tooltip, rather than offering an action that would fail.

## Verified against the live instance

Not just rendered — exercised:

```
library:      3 skills, contract fields rendered
bulk assign:  PUT .../skills → ['haiku', 'word-count'] persisted
agent start:  200
Sync now:     {"haiku": {status: injected, files_written: 2},
               "word-count": {status: injected, files_written: 2}}
UI:           version SHAs + `synced` badges + "Last sync 29/07/2026, 18:06:40"
```

The stopped-agent path was confirmed first (Sync disabled with tooltip), then the running path after starting the agent.

## Not in scope

Exposure curation (#178) — what an agent advertises outward — is deliberately untouched, per the issue. The Settings library panel already reports sync status/last-synced/skill count, so it needed no change; say the word if you'd like it extended further.

### DIFF
--- a/src/frontend/src/components/SkillsPanel.vue
+++ b/src/frontend/src/components/SkillsPanel.vue
@@ -30,7 +30,7 @@
           Skills come from a git repository shared across the fleet. Once it's configured,
           every agent can be assigned skills from it.
         </p>
-        <router-link v-if="isAdmin" to="/settings?tab=skills"
+        <router-link v-if="isAdmin" to="/settings?tab=agents"
                      class="mt-3 inline-block px-3 py-1.5 rounded-lg bg-action-primary-600 hover:bg-action-primary-700 text-white text-sm">
           Configure the library
         </router-link>
--- a/src/frontend/src/components/SkillsPanel.vue
+++ b/src/frontend/src/components/SkillsPanel.vue
@@ -1,324 +1,260 @@
 <template>
-  <div class="p-6">
-    <!-- Skills Assignment Section -->
-    <div class="space-y-4">
-      <!-- Header -->
-      <div class="flex items-center justify-between">
-        <div>
-          <h3 class="text-lg font-medium text-gray-900 dark:text-white">Assigned Skills</h3>
-          <p class="mt-1 text-sm text-gray-500 dark:text-gray-400">
-            Select skills to enable for this agent. Skills are injected when the agent starts.
-          </p>
-        </div>
-        <div class="flex gap-2">
-          <button
-            v-if="agentStatus === 'running'"
-            @click="injectSkills"
-            :disabled="injecting || assignedSkills.length === 0"
-            class="inline-flex items-center px-3 py-1.5 border border-transparent text-sm font-medium rounded-md text-white bg-action-primary-600 hover:bg-action-primary-700 disabled:opacity-50 disabled:cursor-not-allowed"
-          >
-            <svg v-if="injecting" class="animate-spin -ml-1 mr-2 h-4 w-4" fill="none" viewBox="0 0 24 24">
-              <circle class="opacity-25" cx="12" cy="12" r="10" stroke="currentColor" stroke-width="4"></circle>
-              <path class="opacity-75" fill="currentColor" d="M4 12a8 8 0 018-8V0C5.373 0 0 5.373 0 12h4zm2 5.291A7.962 7.962 0 014 12H0c0 3.042 1.135 5.824 3 7.938l3-2.647z"></path>
-            </svg>
-            {{ injecting ? 'Injecting...' : 'Inject to Agent' }}
-          </button>
-          <button
-            @click="saveAssignments"
-            :disabled="saving || !hasChanges"
-            class="inline-flex items-center px-3 py-1.5 border border-transparent text-sm font-medium rounded-md text-white bg-status-success-600 hover:bg-status-success-700 disabled:opacity-50 disabled:cursor-not-allowed"
-          >
-            <svg v-if="saving" class="animate-spin -ml-1 mr-2 h-4 w-4" fill="none" viewBox="0 0 24 24">
-              <circle class="opacity-25" cx="12" cy="12" r="10" stroke="currentColor" stroke-width="4"></circle>
-              <path class="opacity-75" fill="currentColor" d="M4 12a8 8 0 018-8V0C5.373 0 0 5.373 0 12h4zm2 5.291A7.962 7.962 0 014 12H0c0 3.042 1.135 5.824 3 7.938l3-2.647z"></path>
-            </svg>
-            {{ saving ? 'Saving...' : 'Save' }}
-          </button>
-        </div>
+  <div class="space-y-5">
+    <div class="flex items-start justify-between gap-4">
+      <div>
+        <h2 class="text-lg font-semibold text-gray-900 dark:text-gray-100">Skills</h2>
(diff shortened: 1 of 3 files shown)


dolho:
## /review — self-review

**Branch:** `feat/235-skills-ui` → `dev` · **3 files, +386/−282** (diffed against the merge-base, not the base tip)

**Scope: CLEAN.** One commit, one issue. All 9 ACs traced; the two that were not actually met are below. *(A `src/backend/enterprise` line appears in a naive `git diff dev` — that is working-tree noise from a sibling worktree, not in the commit. Verified with `git diff <merge-base> HEAD`.)*

---

### Critical — both fixed in `7ac03171`

**[C1] The "Configure the library" CTA was a dead link** (Confidence 10/10)

```html
<router-link v-if="isAdmin" to="/settings?tab=skills">Configure the library</router-link>
```

There is no `skills` tab. Settings has `general · access · integrations · mcp-keys · agent-permissions · security · sso · agents · retention · activation`, and the Skills Library config lives under **`agents`** (`Settings.vue`, `v-if="activeTab === 'agents'"`).

So the single call-to-action offered to an admin looking at an unconfigured library went nowhere — inside the *"no dead empty states"* AC this panel was built to satisfy. Now points at `agents`.

Related: I asserted in the PR body that the Settings panel already reported sync status / last-synced / skill count, **without checking**. It does report all three (`skill_count`, `commit_sha`, `last_sync`) — but I had the tab wrong, and checking is what would have caught that.

**[C2] Every library error rendered as "empty library"** (Confidence 9/10)

```js
api.get('/api/skills/library').catch(() => ({ data: [] })),
```

The intent was to tolerate an unconfigured library. The effect was to swallow **all** errors — a 500, a timeout, an auth failure — and render *"The library is configured but has no skills yet"*: a confident, wrong empty state pointing the operator at the wrong problem, which is the failure mode this panel is supposed to remove.

Fixed by fetching the list only when `status.configured` is true. The known empty state now comes from the status read; anything else surfaces as a real error.

---

### Clean

**XSS** — no `v-html`; skill names/descriptions are `{{ }}`-interpolated (they originate from an admin-controlled git library, but are not trusted into markup either way). **Invariant #6/#7** — new domain store, all HTTP via the shared `api` client; this PR *removes* the old panel's direct `axios` + hand-built auth header. **Auth** — tab gated on `can_share && !isSystem`, matching the sibling management tabs; no new endpoint. **Stale-state** — injection results are held separately from assignment, so a durable assignment can't inherit a stale success (verified in the store shape, and the "not synced from this screen yet" line covers the no-result case). **Empty-state completeness** — one discriminator in the store, so the panel cannot invent a fourth branch.

### Informational — not fixed

**[I1] `SkillMeta` drops a passed `class`** (6/10) — the inline functional component renders its own `class` and doesn't merge fall-through attrs, so `class="mt-1"` at one call site is ignored. Cosmetic spacing only; verified the component itself renders correctly on the live instance (`213 B` / `277 B`).

**[I2] `watch(() => store.assigned, resetDraft, {deep:true})`** (5/10) — if `assigned` ever changed while a user had unsaved ticks, the draft would be silently reset. No polling or WS feed touches this store today, so it can't fire; worth remembering if one is added.

---

### Verification

Re-ran the live probe after the fixes: tab present, 3-skill library renders, save button present, and the earlier end-to-end still holds — bulk assign → agent start → **Sync now** → `{haiku: injected/2 files, word-count: injected/2 files}` with badges, version SHAs and last-sync line rendering from that response.

**Critical: 2 (fixed) · Informational: 2 · Scope: clean**

vybe (approved):
Skills management surface (ent#235): frontend-only (SkillsPanel + skills store + AgentDetail tab wiring), no backend change. Live-probe verification in PR comments covers browse/assign/sync end-to-end. Security greps clean.

obasilakis:
Heads-up from ent#237 (multi-source skills library) so we don't collide, @dolho — **I have deliberately not touched `SkillsPanel.vue`**, since you're rewriting it here.

Two things that are now available for the rebuild whenever it suits you (backend already shipped on `feature/ent-237-multi-source-skills`):

`GET /api/skills/library` entries carry two new fields:

```jsonc
{
  "name": "pdf-export",
  "source_name": "Acme internal skills",   // which repo it came from
  "shadowed_by": [                          // lower-precedence repos ALSO shipping this name
    { "source_id": "src_…", "source_name": "Trinity Community Skills" }
  ]
}
```

- `source_name` is a **name only, never a URL** — deliberate, because that endpoint is reachable by agent-scoped keys and a private repo's URL is itself sensitive. Source URLs live on the admin-only `GET /api/skills/sources`.
- `shadowed_by` non-empty means another repo also ships that skill name and its copy is **unreachable**. ent#237 AC#4 requires the conflict be visible — *"conflict surfaced in the UI, never a silent overwrite"* — so a badge/tooltip on the row would be the natural home for it. Zero urgency; it's additive and the field is simply ignored until something renders it.

Also FYI: I've split the admin side into `components/SkillSourcesPanel.vue` + `stores/skillSources.js` rather than adding to `stores/skills.js`, specifically so this PR's file isn't in my diff. Different domain too — yours is per-agent assignment, mine is which repos the platform syncs from.

No action needed to merge this PR; I'll rebase ent#237 onto it.

PR #1882, opened 2026-07-30
fix: clamp negative duration_ms at every execution writer

### LINKED ISSUE #1832: bug: negative duration_ms from cross-process clock skew reaches the Overview analytics chart unguarded
## Summary

`update_execution_status` persists an **unguarded** `completed_at − started_at`. When the finalizing process's clock is behind the one that wrote `started_at`, the stored `duration_ms` is **negative**, and it flows **unguarded** into `get_agent_analytics` → the Agent Detail Overview chart.

Found by #1771 slice c (PR #1827), shipped there as a `strict=True` xfail plus a companion characterization test pinning today's behaviour. No product fix in that PR (its AC forbids product-code change).

## Minimal reproduction

Row in `running` with `started_at = now + 300s`, then `update_execution_status(id, "success")`:

```
duration_ms = -299997
get_agent_analytics → {'avg': -299997, 'p95': -299997}
```

## Why this is not a toy input

`started_at` and `completed_at` are written by **different processes**:

- the standalone `src/scheduler/` container repeats the same unguarded subtraction at `src/scheduler/database.py:514-516`
- the backend runs `--workers 2`

so the two timestamps can legitimately come from different clocks. No NTP guarantee makes this impossible.

## Why "canary G-03 already covers it" is wrong

G-03 **detects** the skew and files it at severity `minor`. Nothing **prevents** the poisoned metric, and the analytics consumption path is entirely unguarded — so a negative duration silently reaches a user-facing chart regardless of whether G-03 fires.

## Suggested fix

Clamp at the write (`max(0, delta)`) in **both** writers — `db/schedules/executions.py` and `src/scheduler/database.py:514-516` — since fixing only the backend leaves the scheduler path live. Consider whether a clamped-to-zero row should be distinguishable from a genuine sub-millisecond execution.

> **Note for whoever fixes it:** the xfail at `tests/unit/test_1771c_schedules_cas_edges.py:307` is `strict=True`. When a clamp lands it flips to **XPASS and fails the suite**. That is the intended alarm telling you to delete the marker — not a broken test.

Found via #1771 (PR #1827).

Kayvan-Zahiri:
Fixes #1832

## Problem

`started_at` and `completed_at` are written by **different processes** — the backend runs `--workers 2` and the standalone `src/scheduler` container is a separate image — so an unguarded `completed_at - started_at` goes negative whenever the finalizing process's clock trails the one that opened the row. The value flowed unguarded into `get_agent_analytics` and out to the Agent Detail Overview chart. Canary G-03 detects the skew at severity `minor` but prevents nothing.

## Fix

Clamps with `max(0, ...)` at the write, so no reader has to know.

The issue named two writers. There are **three** — the second scheduler site finalizes `process_schedule_executions` with the same unguarded subtraction, so fixing only the two named ones would have left the bug live on that path:

| Writer | |
|---|---|
| `src/backend/db/schedules/executions.py` | `update_execution_status` |
| `src/scheduler/database.py` | `schedule_executions` finalizer |
| `src/scheduler/database.py` | `process_schedule_executions` finalizer ← not named in the issue |

## The xfail, as promised

`test_1771c_schedules_cas_edges.py::test_a6_clock_skew_must_not_persist_negative_duration` is `strict=True`, so it flips to XPASS and fails the suite the moment a clamp lands — exactly as its reason predicted. The marker is removed and it now asserts the contract directly.

Its companion `test_a6_current_behaviour_negative_duration_is_persisted` pinned the **pre-fix** value (`< 0`) and would also have failed. It now pins the clamped value (`== 0`) and is renamed to say so.

## Tests

New `tests/unit/test_1832_duration_clamp.py`:

- the reported reproduction (row opened 300s in the future) clamps to `0`
- a real 5s duration is **not** flattened — the clamp only touches negatives
- every terminal transition (`success` / `failed` / `error`), since the subtraction is shared
- the consumption path: `get_agent_analytics` reports no negative `avg` / `p95`
- an **AST guard** asserting no writer assigns `duration_ms` unguarded

The AST guard covers the two `src/scheduler` writers, which ship in a separate image these unit tests cannot import, and it catches a fourth call-site added later — which is how this bug survived being fixed in one place. It accepts a literal `0` (the `skipped` rows) or a `max()` wrapper, and there is a test asserting the guard itself fails on unclamped source rather than passing vacuously.

```
52 passed
```

(9 new, plus the 43 existing in the CAS-edges file whose xfail this retires.)

## Trade-off

A skewed row clamps to exactly `0`, making it indistinguishable from a genuine sub-millisecond execution. The issue raises this. Preserving the distinction needs a separate column or a sentinel — a schema change, and with dual-track migrations that is well beyond a p2 reliability fix. Recorded in the test module docstring so the decision is discoverable rather than assumed.

### DIFF
--- a/src/backend/db/schedules/executions.py
+++ b/src/backend/db/schedules/executions.py
@@ -366,7 +366,12 @@ class ScheduleExecutionsMixin:
 
             started_at = parse_iso_timestamp(row["started_at"])
             completed_at = parse_iso_timestamp(utc_now_iso())
-            duration_ms = int((completed_at - started_at).total_seconds() * 1000)
+            # started_at and completed_at are written by different processes
+            # (backend --workers 2, and the standalone scheduler container), so
+            # clock skew can make this subtraction negative. Clamp at the write
+            # rather than at every reader — get_agent_analytics consumes
+            # duration_ms unguarded (#1832).
+            duration_ms = max(0, int((completed_at - started_at).total_seconds() * 1000))
 
             values = {
                 "status": status,
--- a/src/scheduler/database.py
+++ b/src/scheduler/database.py
@@ -567,7 +567,10 @@ class SchedulerDatabase:
 
             started_at = parse_scheduler_ts(row["started_at"])
             completed_at = datetime.utcnow()
-            duration_ms = int((completed_at - started_at).total_seconds() * 1000)
+            # Clock skew between the writer of started_at and this process can
+            # make the subtraction negative; clamp so a poisoned duration never
+            # reaches the analytics chart (#1832).
+            duration_ms = max(0, int((completed_at - started_at).total_seconds() * 1000))
 
             cursor.execute("""
                 UPDATE schedule_executions
@@ -1217,7 +1220,10 @@ class SchedulerDatabase:
 
             started_at = parse_scheduler_ts(row["started_at"])
             completed_at = datetime.utcnow()
-            duration_ms = int((completed_at - started_at).total_seconds() * 1000)
+            # Clock skew between the writer of started_at and this process can
+            # make the subtraction negative; clamp so a poisoned duration never
+            # reaches the analytics chart (#1832).
+            duration_ms = max(0, int((completed_at - started_at).total_seconds() * 1000))
 
             cursor.execute("""
                 UPDATE process_schedule_executions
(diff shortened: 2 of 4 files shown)


vybe (approved):
Validated via /validate-pr. Ran test_1832_duration_clamp.py + test_1771c_schedules_cas_edges.py locally at 919ad4c5: 52 passed. Clamp-at-write in all three writers (incl. the process_schedule_executions finalizer the issue missed), strict-xfail correctly retired into a direct contract assertion, AST source guard covers the scheduler image + future call-sites and is itself tested against vacuous pass. The clamp-to-0 vs sub-millisecond ambiguity trade-off is recorded. LGTM.

PR #1883, opened 2026-07-30
feat(skills): library lifecycle automation — auto-sync, fleet re-inject, removal-on-unassign (abilityai/trinity-enterprise#236)

obasilakis:
## Summary

Closes the three skills-lifecycle gaps documented as "Not Built" in requirements §21.1/§21.4. All three default **OFF/no-op** — a zero-config install behaves exactly as before.

- **Removal-on-unassign** — unassigning a skill now actually removes the injected package, via the existing manifest-prune primitive. Wired into both the single `DELETE` and the bulk `PUT` (the primary UI/MCP path, which drops skills far more often).
- **Start-path reconciliation** — how a removal reaches an agent that was stopped when it happened. The assignment row is gone by then, so it's reconciled (diff the agent's platform-managed skill dirs against the assignment set), not replayed. No tombstone table, no migration.
- **Scheduled auto-sync + fleet-wide re-inject** — a leader-locked backend service that pulls the library on a timer and, when the commit actually changed, pushes updates to running agents at bounded concurrency.
- **Durable sync status** — prerequisite for the "failures surface honestly" AC: the old `_last_sync` was per-process, so under `--workers 2` the panel showed a stale timestamp and could never show an error at all.

## Design decisions worth knowing

**Reconciliation, not tombstones.** Every removal route converges (single DELETE, bulk-PUT shrink, direct DB edit) with no schema change. It runs after injection and **also at zero assigned skills** — that's exactly the "unassigned the last skill" case an early return would strand forever.

**Blast-radius guard.** A reconcile proposing >10 removals for one agent refuses wholesale and alarms. A wiped `agent_skills` table is indistinguishable from a legitimate mass-unassign, and keeping files is the recoverable direction (#1638/#1644 discipline; the alarm carries counts, never skill names).

**Backend-hosted, not the standalone scheduler.** The sweep must reach agent containers; the scheduler sits on `trinity-platform` only.

**Only platform-written files die.** The delete set is the previous injection's own manifest, so agent-authored files and runtime artifacts survive; `os.rmdir` refuses a non-empty directory, so a skill dir holding agent files stays.

## Hardening found by `/review` and `/cso --diff` on this branch

Four issues were found and fixed here, not deferred:

1. **PAT could reach durable admin-visible state.** `sync_library`'s outer handler passed a raw `str(e)` into a new `system_settings` row rendered in Settings, and the authenticated remote URL is an argument in the `subprocess.run` command list. Now scrubbed at the write (canary G-04 class).
2. **No lock on the shared clone.** Scheduled + manual sync both run `git fetch` + `git reset --hard` on `/data/skills-library`. Two admins colliding was never realistic; a scheduled loop landing on a manual click is routine. Cross-worker lock added; contention returns **409** and writes no status row, so a contended click can't paint "Last sync failed".
3. **Bulk-PUT TOCTOU.** `previous - new` is computed before the lock is taken, so concurrent PUTs could each hand the other's newly-assigned name to the remover. Assignments are now re-read inside the lock and a still-assigned skill is refused (#1445 pattern).
4. **The automation on-switch was gated by role only.** `PUT /api/settings/skills-library` enables an unattended fleet-wide write of `SKILL.md` files — which Claude executes as *instructions*, not data. `assert_admin` answers "what role", never "is this a human": an agent-scoped MCP key resolves to its owner carrying the owner's role. Now `reject_agent_principal`. **Third occurrence of the trinity-ops-agent#232 class** (#1644, #1816).

⚠️ The remaining half of that chain — `skills_library_url` writable by an agent-scoped key through the generic settings PUT, with the SSRF guard checking only the *host* — is **pre-existing and still open**, filed as `abilityai/trinity-enterprise#293` (P1). It's filed privately rather than here because publishing an unfixed exploit chain on a public repo is the wrong trade. That issue asks for a structural fix rather than a fourth per-endpoint patch.

## Changes

| Area | Files |
|---|---|
| New service | `services/skills_sync_service.py` |
| Removal + reconcile + durable status | `services/skill_service.py`, `services/skill_packaging.py` |
| Wiring | `routers/skills.py`, `services/agent_service/lifecycle.py`, `main.py` |
| Settings surface | `routers/settings.py`, `services/settings_service.py`, `models.py`, `views/Settings.vue` |
| Docs | requirements §21.7, `architecture.md`, `skill-injection.md`, `skills-library-sync.md`, `learnings.md` |
| Tests | `tests/unit/test_ent236_skills_lifecycle.py` (58), `test_telegram_webhook_backfill.py` (stub kept current) |

## Test Plan

- [x] New suite: `pytest tests/unit/test_ent236_skills_lifecycle.py -q` → **58 passed**
- [x] Full unit suite vs a pristine `origin/dev` baseline, three orderings:
  - default order — **5573 passed / 0 failed** (baseline 5525 / 0)
  - seed 12345 — **5575 passed / 0 failed**
  - seed 777 — **5566 passed / 9 errors**, the same 9 errors reproduced on pristine `origin/dev` with that seed (pre-existing order-dependence in `test_1089_*`)
- [x] Design-token check: `node scripts/check-design-tokens.mjs` → all references resolve
- [x] Vendored-parity tests re-run green (`skill_packaging.py` is backend-only, Invariant #5 unaffected)
- [ ] Manual: enable auto-sync at 300s with a real library, confirm a commit change sweeps running agents and a no-op pull does not
- [ ] Manual: unassign a skill on a stopped agent, start it, confirm the package is gone and CLAUDE.md no longer lists it

Fixes abilityai/trinity-enterprise#236


vybe (approved):
Validated via /validate-pr: requirements §21 + architecture + feature flows updated, ent#236 tuning vars wired into both compose files and .env.example, no secrets, blast-radius guard on reconcile, full CI green (6 pytest runs). Cross-tracker close — will set status-in-dev on ent#236 manually.

PR #1885, opened 2026-07-30
fix(security): redact credentials from reaped process cmdlines

dolho:
Fixes abilityai/trinity-enterprise#292

Fixes a credential-in-logs defect reported privately. **P0.**

## The bug

The agent-side orphan sweeper logged the full command line of every process it killed, unsanitized. Trinity writes git remotes with the PAT embedded in the URL, so every reaped `git remote-https` wrote a **live GitHub PAT in plaintext** into the agent container log — and from there into Vector's persisted archives and any snapshot of the log volume.

**Log level was never a mitigation.** `main.py` sets `basicConfig(level=INFO)` and Vector routes agent logs by container class with no level filter, so INFO and WARNING persist identically. Downgrading the line would not have stopped it; only redaction does.

## The rule is shape-independent, deliberately

`credential_sanitizer` already carried value patterns for `ghp_`, `github_pat_`, `sk-ant-`, `xoxb-` and friends — but those only match tokens whose prefix we already know, and the report showed that failing twice:

- an audit grepping `oauth2:` cleared an agent that was carrying credentials in the **`x-access-token:`** form
- both classic `ghp_` (~40 chars) and fine-grained `github_pat_` (~93) were live, so any fixed-length assumption is wrong too

So the new rule keys on **URL structure**, not token shape: `://[^/@\s]+@` → `://***@`. That covers both userinfo prefixes, both token classes, and formats that don't exist yet.

## One rule, one place

Rather than add a sweeper-local regex, it's folded into `sanitize_text()` — so **every existing caller gains it**, not just the sink that leaked — with `sanitize_cmdline()` as the explicit entry point for anything derived from argv.

`routers/git.py::_redact_url_userinfo` (added by #1595 for this *same credential* in git stderr) now delegates to the shared helper. That duplication is exactly why the argv sink was missed when the stderr sinks were fixed; a second copy is how the next one gets missed too.

**Sanitize before the length cap.** Truncating first can slice a token mid-value and leave a partial secret that no pattern then matches.

Also corrects a swapped comment in the pattern table — `ghp_` is the *classic* PAT and `github_pat_` the *fine-grained* one, not the reverse.

## Tests — 12

Every userinfo × token combination from the report, including an **unknown future format** (the case a shape-based rule misses); benign `@` left untouched, because an over-broad redactor gets switched off and then nothing is redacted; sanitize-before-cap ordering; a **structural guard** that fails if any `_read_cmdline` result reaches a log without the helper; and a guard that `routers/git.py` hasn't re-derived the regex.

Verified 2 of them fail with the fix reverted. **548 passed** across the sanitizer/orphan/credential suites (5 pre-existing collection errors from a missing `hypothesis` dep, unrelated).

## Explicitly NOT fixed here

**The PAT is still in argv.** It remains readable via `/proc/<pid>/cmdline` to any co-resident process in the container. Closing that needs a credential helper or `http.extraHeader` **plus a migration that rewrites the remotes already baked into existing containers** (#1264) — redaction cannot reach those. This PR is the half that protects the installed base immediately, without a per-agent migration; the argv half needs its own change and should not block this one.

Operator remediation (rotate the exposed tokens, purge affected archives) is also out of scope for a code change — and worth noting that **rotation alone doesn't stop the leak** while a container's baked remote still holds the old token, and **purging alone doesn't help** while snapshots retain it.

### DIFF
--- a/docker/base-image/agent_server/routers/git.py
+++ b/docker/base-image/agent_server/routers/git.py
@@ -24,6 +24,8 @@ from ..utils.registered_run import run_registered
 from .files import _read_persistent_state
 from .snapshot import build_snapshot, restore_from_tar
 
+from ..utils.credential_sanitizer import redact_url_userinfo
+
 logger = logging.getLogger(__name__)
 router = APIRouter()
 
@@ -560,13 +562,12 @@ def _run_auto_sync_once(home_dir: Path) -> Dict:
 def _redact_url_userinfo(text: str) -> str:
     """Strip credentials from URLs in git output (#1595 review).
 
-    Git error lines print the full remote URL — for PAT-embedded remotes that
-    is `https://x-access-token:ghp_…@github.com/…` — and both the sync-state
-    error summary and the push-collision operator-queue entry persist and
-    surface that output. Redact userinfo everywhere git output leaves the
-    process.
+    Kept as a name so the two git-stderr call sites read unchanged, but the rule
+    itself now lives in `utils.credential_sanitizer` (ent#292): the same
+    credential leaks through argv as well as stderr, so one regex serving both
+    is what stops the next sink from re-deriving its own.
     """
-    return re.sub(r"(://)[^/@\s]+@", r"\1***@", text)
+    return redact_url_userinfo(text)
 
 
 def _summarize_git_error(raw: str) -> str:
--- a/docker/base-image/agent_server/utils/credential_sanitizer.py
+++ b/docker/base-image/agent_server/utils/credential_sanitizer.py
@@ -55,8 +55,8 @@ SECRET_VALUE_PATTERNS = [
     r'sk-[a-zA-Z0-9]{20,}',           # OpenAI API keys
     r'sk-proj-[a-zA-Z0-9\-_]{20,}',   # OpenAI project keys
     r'sk-ant-[a-zA-Z0-9\-_]{20,}',    # Anthropic API keys
-    r'ghp_[a-zA-Z0-9]{36,}',          # GitHub PAT (fine-grained)
-    r'github_pat_[a-zA-Z0-9_]{22,}',  # GitHub PAT (classic)
+    r'ghp_[a-zA-Z0-9]{36,}',          # GitHub PAT (classic, ~40 chars)
+    r'github_pat_[a-zA-Z0-9_]{22,}',  # GitHub PAT (fine-grained, ~93 chars)
     r'gho_[a-zA-Z0-9]{36,}',          # GitHub OAuth token
     r'ghs_[a-zA-Z0-9]{36,}',          # GitHub App token
     r'ghr_[a-zA-Z0-9]{36,}',          # GitHub refresh token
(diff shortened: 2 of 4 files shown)


dolho:
## /review — self-review

**3 source files + 1 test file, +272/−9** against the merge-base. Scope: clean, one issue.

### [C1] The fix had reintroduced the exact duplication the issue forbade — fixed in `b2a541ea`

The issue is explicit: *"Prefer promoting one of them into a shared `utils/` helper applied at every argv/env log exit **over a sweeper-local regex**"*.

Chasing the CI collection failure, I inlined `_URL_USERINFO_RE` into `orphan_sweep` to make redaction independent of an import. That left the same security regex defined **twice** — and my own test asserted the principle for `routers/git.py` while I violated it one file over.

Two copies of a security pattern drift, and **that drift is this issue**: #1595 fixed the same credential in git stderr with a local regex, the argv sink kept its own behaviour, and it was missed for two months.

Fixed without losing the guarantee: the rule lives solely in `credential_sanitizer`; `orphan_sweep` imports it **lazily** (so no import-time dependency can break collection), and when the helper genuinely cannot be reached the cmdline is **dropped** rather than logged raw. Diagnostics degrade; the credential cannot leak. Added a guard asserting the pattern has exactly one definition across both files.

### Clean

**Redaction correctness** — every userinfo × token combination from the report, plus an unknown future format; benign `@` untouched. **Ordering** — sanitize before the length cap, so truncation can't slice a token mid-value and leave a partial secret; pinned by a test. **Class sweep** — a structural guard fails if any future `_read_cmdline` result reaches a log without the helper. **Never raises** — redaction failure can't break a sweep. **Import safety** — CI-style collection verified clean at 5624 tests.

### Honest note on the churn

Five commits for a one-line leak, and three of them were me fixing my own breakage:

1. `sys.modules` pollution in the test — caught by lint
2. a bare relative import breaking a flat-load path — caught by CI, **after I saw it locally and dismissed it as environmental**
3. the flat fallback resolving to the *backend's* different `credential_sanitizer` (no `sanitize_cmdline`), aborting collection entirely
4. a byte-equality assertion that depends on `GITHUB_SERVER_URL` being set on a runner

Each was real, and each was found by CI rather than by me. The pattern worth naming: **a local test run and a CI run are different environments**, and I twice reasoned about that difference instead of reproducing it. Reproducing CI's exact invocation (`cd tests && pytest unit/ -p randomly --randomly-seed=…`) found in seconds what two rounds of theorising did not.

**Critical: 1 (fixed) · Scope: clean**

vybe (approved):
Validated via /validate-pr. Ran tests/unit/test_292_cmdline_credential_leak.py locally at 3f7f8802: 17 passed. Shape-independent URL-userinfo redaction folded into sanitize_text (every caller covered, not just the leaking sink); sanitize-before-cap ordering pinned; fail-closed drop when the sanitizer is unreachable; one-definition structural guards prevent the #1595-style drift that caused this. Added the qualified 'Fixes abilityai/trinity-enterprise#292' to the body for traceability — cross-tracker, so status bump + close are manual. LGTM.

Issue #919, opened 2026-05-23
feat(mcp): introspect agent pipelines via standardized `~/.trinity/pipelines/` + `~/.trinity/pipeline-state/`


vybe:
## Summary

Expose agent-defined, long-running pipelines (multi-stage DAGs run by the agent itself) through Trinity's MCP surface, by standardizing two read-only file paths inside agents and adding two thin MCP tools that read them. No new backend logic, no DB, no platform-level DAG engine — Trinity stays a fleet orchestrator while agents own pipeline execution and recovery.

## Context

Trinity-compatible agents increasingly run multi-stage internal pipelines — perception → incubation → synthesis → publish → measure → mutate, or similar shapes. Today these live as ad-hoc skills + schedules + event subscriptions inside each agent. There is no canonical way for Trinity (or another agent) to ask "what's the state of agent X's pipeline?" without knowing that agent's internal file layout.

The conclusion from the design discussion: the agent already has every primitive it needs to **manage** its pipeline (schedules CRUD, events, operator queue, pre-check hook, dashboard.yaml). What's missing is a **standardized read surface** so Trinity and other agents can observe pipelines uniformly.

The full design will ship as a plugin skill (`agent-dev:add-pipeline`) in `abilityai/abilities` once this Trinity surface lands. That plugin will scaffold the `~/.trinity/pipelines/` and `~/.trinity/pipeline-state/` layout into any compliant agent.

## Acceptance Criteria

- [ ] **Standardized file convention** documented in `docs/TRINITY_COMPATIBLE_AGENT_GUIDE.md`:
  - `~/.trinity/pipelines/<pipeline_id>.yaml` — pipeline definitions (DAG, stages, transitions, escalation policy)
  - `~/.trinity/pipeline-state/<pipeline_id>/<instance_id>.json` — runtime state (current stage, attempt, health, blockers, open escalations)
  - JSON Schema for `state.json` shipped at `docs/schemas/agent-pipeline-state.schema.json`
  - YAML Schema for `pipeline.yaml` shipped at `docs/schemas/agent-pipeline.schema.json`

- [ ] **MCP tool `list_agent_pipelines(agent_name)`** in `src/mcp-server/src/tools/agents.ts` (or new `pipelines.ts`):
  - Lists `~/.trinity/pipelines/*.yaml` via existing `agent_files` router
  - Returns `[{ pipeline_id, name, stages_count, instances_count, health_summary }]`
[...]

PR #1886, opened 2026-07-30
fix(systems): harden manifest YAML against expansion bombs and duplicate keys (#1884)

### LINKED ISSUE #1884: Harden parse_manifest against YAML anchor/alias expansion and duplicate keys
## Context

`services/system_service.py::parse_manifest` uses `yaml.safe_load`, which prevents
arbitrary object construction but does **not** prevent:

1. **Anchor/alias expansion** — a few-hundred-byte manifest (a "billion laughs" /
   YAML bomb) can expand to gigabytes and pin a worker. The `MANIFEST_MAX_BYTES`
   size cap added in trinity-enterprise#126 bounds the *input*, not the expansion,
   so it does not close this.
2. **Duplicate keys** — `safe_load` silently keeps the **last** duplicate. A
   manifest with two `agents:` blocks deploys only the second one, with no signal.
   That is a real footgun now that trinity-enterprise#126 ships a hand-editable
   YAML textarea in the UI.

## Why now

`src/backend/models.py` (`SystemDeployRequest`) carries a comment stating this is
"tracked as a follow-up" — this issue is that tracker. The
trinity-enterprise#126 plan named it in its security section and deliberately
deferred it as out of scope.

## Prior art in this repo

MCP `src/mcp-server/src/tools/pipelines.ts` (#919) already implements exactly
"size cap + duplicate-key guard + alias guard" for the agent-published pipeline
YAML, and `docs/memory/architecture.md` cites it as the standard for this class.
Mirror that rather than inventing a second approach.

## Severity

Low. Both reachable paths (`POST /api/systems/deploy` and the ent#126
bundled-manifest catalog) require an authenticated **`creator`**, so this is
privileged-user DoS / silent misconfiguration, not an unauthenticated attack. Filed
for completeness because the mitigation is cheap and the precedent already exists.

## Acceptance criteria

- [ ] `parse_manifest` rejects a manifest whose alias expansion exceeds a bounded
      node/expansion budget, with a named 400 (not a timeout or an unnamed 500)
- [ ] `parse_manifest` rejects (or at minimum warns on) duplicate top-level and
      per-agent keys rather than silently keeping the last
- [ ] Both behaviours mirror `pipelines.ts` (#919) rather than a second approach
- [ ] Tests cover a billion-laughs payload and a duplicate-`agents:` manifest
- [ ] The "tracked as a follow-up" comment in `models.py` is updated to point here

## Notes

Surfaced by the pre-landing review of trinity-enterprise#126 (`/cso --diff`), where
it was hard-excluded from the audit itself as authenticated DoS.

dolho:
Closes #1884.

`parse_manifest` used bare `yaml.safe_load`. Mirrors the house standard set by the MCP pipeline reader (#919) — **size cap + duplicate-key guard + alias guard** — rather than inventing a second approach. PyYAML has no equivalent of the JS `yaml` library's `uniqueKeys` / `maxAliasCount`, so both are implemented in a `SafeLoader` subclass.

## A measured correction to the issue's premise

The issue describes a few-hundred-byte manifest expanding to gigabytes and pinning a worker. **That does not reproduce on PyYAML.** It memoises constructed objects per node, so a billion-laughs document yields a shared-reference DAG:

```
levels=4  src=237B  parse=0.002s  peak=19KB  aliases-shared=True
levels=6  src=329B  parse=0.003s  peak=18KB  aliases-shared=True
levels=8  src=421B  parse=0.013s  peak=20KB  aliases-shared=True
```

The stated DoS isn't reachable through `safe_load` alone. The guard ships anyway — the DAG re-expands the instant anything walks or serialises it (`json.dumps`, a deep copy, echoing the manifest back), and guarding the parser is cheaper than auditing every future consumer — but **the duplicate-key half is the one with immediate user-visible impact**, and that ordering is now written down rather than left implied.

## …which changed the guard's shape

A `maxAliasCount`-style budget is the wrong rule here. The classic bomb uses **~45 aliases** across a few levels and would sail under any count-based limit, because the blow-up is *multiplicative per level*, not linear in aliases.

So the budget counts **expansion cost** — the number of logical nodes the document would have with every alias resolved — accumulated at compose time, so it's spent before the document is materialised. Verified behaviour:

| input | result |
|---|---|
| bomb, 3–4 levels | parses (small structure, no DoS) |
| bomb, 5–8 levels | **refused** `manifest_alias_budget_exceeded` |
| `<<: *defaults` reuse across 3 agents | **parses** — the guard must not break the feature |

## Duplicate keys

Rejected at **every mapping depth**, not just the two levels the issue names — a duplicated `template:` inside one agent is the same silent last-wins bug one level down. The error names the offending line, because "duplicate key" without a location in a 200-line manifest is worse than the silent bug.

## Two assumptions in the issue that don't hold on `dev`

Neither the *"tracked as a follow-up"* comment in `models.py` nor `MANIFEST_MAX_BYTES` exists here — trinity-enterprise#126 hasn't landed in the public repo. So the size cap is **introduced** by this change (completing the three-part mirror), and the `models.py` pointer is **added** rather than updated. Flagging rather than silently reinterpreting the AC.

## Compatibility

`ManifestError` subclasses `ValueError`, so both call sites (`deploy_manifest`, `system_seed_service`) keep working unchanged. The deploy path additionally surfaces the code, giving the **named 400** the AC asks for instead of a timeout or an unnamed 500.

## Tests — 10

Billion-laughs refused by name; a bomb with <100 aliases still refused (pinning *why* the budget isn't count-based); legitimate anchor reuse still parses; duplicate `agents:` and duplicate `template:` both refused; the error names the line; oversize refused; `ManifestError` is a `ValueError`; a normal manifest unaffected; malformed YAML still reported as a parse error rather than mislabelled as a bomb.

**Verified 7 fail with the guards bypassed.** 204 passed across the system/manifest suites; CI-style collection clean at 5617 tests (the 4 errors are `hypothesis`-dependent files, absent locally, installed in CI).

### DIFF
--- a/src/backend/models.py
+++ b/src/backend/models.py
@@ -627,6 +627,10 @@ class SystemManifest(BaseModel):
 
 class SystemDeployRequest(BaseModel):
     """Request to deploy a system from YAML manifest."""
+    # #1884: the raw YAML is parsed by `system_service.parse_manifest`, which
+    # applies the size / expansion-cost / duplicate-key guards mirroring the MCP
+    # pipeline reader (#919). Not validated here — a Pydantic field cannot see
+    # YAML structure.
     manifest: str  # Raw YAML string
     dry_run: bool = False
     # trinity-enterprise#125: abort on the first agent-create failure (legacy
--- a/src/backend/services/system_service.py
+++ b/src/backend/services/system_service.py
@@ -32,17 +32,140 @@ logger = logging.getLogger(__name__)
 _REASON_MAX_LEN = 500
 
 
+# --- Manifest YAML hardening (#1884) -------------------------------------
+#
+# `yaml.safe_load` blocks arbitrary object construction but not two other
+# things, and the MCP pipeline reader (#919) already set the house standard for
+# this class — "size cap + duplicate-key guard + alias guard". This mirrors that
+# rather than inventing a second approach; PyYAML simply has no equivalent of
+# the JS `yaml` library's `uniqueKeys` / `maxAliasCount` options, so both are
+# implemented here.
+
+# Bounds the INPUT. Deliberately not sufficient on its own: alias expansion is
+# bounded by the budget below, because a few hundred bytes can expand to
+# gigabytes ("billion laughs") while sitting comfortably under any size cap.
+MANIFEST_MAX_BYTES = 256 * 1024
+
+# Bounds the EXPANSION COST, not the alias count.
+#
+# A naive `maxAliasCount` is the wrong shape for a bomb: the classic
+# billion-laughs uses ~45 aliases across a few levels and would sail under any
+# count-based budget, because the blow-up is MULTIPLICATIVE per level, not
+# linear in aliases. So this budget is the total number of logical nodes the
+# document would have if every alias were written out — which is exactly the
+# quantity that explodes.
+#
+# Measured, not assumed: PyYAML itself does NOT expand aliases into copies (it
+# memoises constructed objects per node, so a bomb yields a shared-reference
+# DAG at ~20 KB, not gigabytes). The DoS the issue describes is therefore not
+# reachable through `safe_load` alone. The guard stays because the DAG re-expands
+# the moment anything walks or serialises it — `json.dumps`, a deep copy, or an
+# echo of the manifest back to a client — and a parser that gains this guard is
+# cheaper than auditing every future consumer.
+MANIFEST_MAX_EXPANDED_NODES = 100_000
+
+
+class ManifestError(ValueError):
+    """A manifest the platform refuses to parse.
+
+    A distinct type so the router can answer a NAMED 400 instead of letting a
+    bomb surface as a request timeout or an unnamed 500 — which is the
+    difference between "your manifest is malformed" and "Trinity is broken".
+    """
+
+    def __init__(self, code: str, message: str):
+        self.code = code
+        super().__init__(message)
+
+
+class _HardenedManifestLoader(yaml.SafeLoader):
+    """SafeLoader + an expansion-cost budget + duplicate-key rejection."""
+
+    def __init__(self, stream):
+        super().__init__(stream)
+        self._expanded_nodes = 0
+        self._anchor_cost: dict = {}
+
+    def compose_node(self, parent, index):
+        # An alias contributes the FULL logical size of what it points at, so a
+        # pyramid of anchors costs what it would cost written out.
+        if self.check_event(yaml.events.AliasEvent):
+            anchor = self.peek_event().anchor
+            cost = self._anchor_cost.get(anchor, 1)
+            self._expanded_nodes += cost
+            if self._expanded_nodes > MANIFEST_MAX_EXPANDED_NODES:
+                raise ManifestError(
+                    "manifest_alias_budget_exceeded",
+                    f"Manifest expands to more than {MANIFEST_MAX_EXPANDED_NODES:,} "
+                    "nodes once its YAML aliases are resolved. This is the shape of "
+                    "an expansion bomb; if the manifest is genuine, write the "
+                    "repeated block out instead of anchoring it.",
+                )
+            return super().compose_node(parent, index)
+
+        before = self._expanded_nodes
+        self._expanded_nodes += 1
+        event = self.peek_event()
+        anchor = getattr(event, "anchor", None)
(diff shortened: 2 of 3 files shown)


dolho:
## /review — self-review

**3 files, +325/−5** against the merge-base. Scope: clean, one issue, one commit plus this review's additions.

### Critical: none

The substantive risk in a parser guard is a **false positive**, and the highest-stakes one wasn't covered. Both gaps below are test coverage, fixed in `9caa4646`.

### [I1] Nothing checked the manifests Trinity actually ships — fixed

`config/manifests/default-system.yaml` is deployed automatically on a fresh install (`system_seed_service`). A false positive there isn't a rejected request — it's a **broken installation**, on first boot, before anyone has touched anything.

Verified all three bundled manifests parse (`acme-consulting` 3 agents, `default-system` 3, `research-network` 2) and pinned it, so a future tightening of the budget can't brick first-run seeding silently.

### [I2] Every bomb test was a deep pyramid — fixed

The expansion budget is shape-agnostic by construction, but nothing proved it against a **wide-and-shallow** bomb (two levels, huge fan), which a level-based intuition misses. It does refuse one — 400-fan × 2 levels, ~2.8 KB source — but that was luck rather than coverage until now.

### Clean

**Injection/parse safety** — still `SafeLoader`, no construction surface added. **Compatibility** — `ManifestError` subclasses `ValueError`, so both call sites (`deploy_manifest`, `system_seed_service`) are unchanged; verified by test. **Named 400** — the deploy path surfaces the code, so a refusal is actionable rather than a timeout or unnamed 500. **Not over-broad** — legitimate `<<: *defaults` reuse across agents still parses. **Misclassification** — malformed YAML reports as a parse error, not as a bomb. **Depth** — duplicate keys caught at every mapping level, not just the two the issue names. **Collection** — CI-style run clean at 5617 tests.

### Worth restating for the reviewer

This PR **contradicts its own issue's premise, deliberately and with measurements**. The issue asserts a few-hundred-byte manifest can expand to gigabytes and pin a worker; PyYAML memoises constructed objects per node, so it produces a shared-reference DAG at ~20 KB and milliseconds. The guard still ships (the DAG re-expands on any walk or serialisation) but **the duplicate-key half is the one with immediate user-visible impact**, and the PR says so rather than inheriting the framing.

That measurement is also what changed the guard's shape — a `maxAliasCount`-style limit would have passed the classic bomb, since it uses ~45 aliases and the blow-up is multiplicative per level.

**Critical: 0 · Informational: 2 (both fixed) · Scope: clean**

dolho:
## CodeQL review item — handled in `fe417062`

**Alert:** `py/unsafe-deserialization` on `yaml.load(..., Loader=_HardenedManifestLoader)` (`system_service.py:150`) — *"Deserialization of user-controlled data"*.

**Verdict: false positive**, dismissed as such with the justification recorded on the alert. The rule keys on the call shape and does not resolve the subclass; a custom loader is the only way to add the guards this PR exists for, and `yaml.load` + a `SafeLoader` subclass is the canonical way to do that.

But I did not want to answer a security alert with an assertion, so the property is now **pinned rather than argued**:

**Structural** — `yaml.SafeLoader` is in the MRO and none of `FullLoader` / `UnsafeLoader` / `Loader` / `CLoader` is. If anyone re-parents the loader, the build fails instead of a reviewer having to spot it.

**Behavioural** — the payloads the rule exists to stop are refused:

```
MRO includes SafeLoader: True
is a Full/Unsafe loader:  False

!!python/object/apply:os.system ["echo pwned"]              -> refused
!!python/object/apply:subprocess.check_output [["id"]]      -> refused
!!python/name:os.system {}                                  -> refused
!!python/object/new:type [x, !!python/tuple [], {}]         -> refused
```

That last point matters more than the dismissal: this PR is what moved the call off `yaml.safe_load`, so "it's still safe" is exactly the claim that deserves a regression test rather than a comment. 17 tests pass.

vybe (approved):
Validated via /validate-pr. Ran tests/unit/test_1884_manifest_yaml_hardening.py locally at fe417062: 17 passed. Expansion-cost budget (not alias count) is the right shape for the multiplicative bomb; duplicate-key rejection at every depth with line numbers; ManifestError subclasses ValueError so both existing call sites are unaffected; shipped-manifest false-positive test protects first-run seeding; CodeQL alert answered with a pinned SafeLoader-MRO test rather than an assertion. LGTM.

Issue #1754, opened 2026-07-23
[BUG] Dashboard navigation and timeline controls overlap at narrow desktop widths


erichroepke:
## Describe the bug

> Maintainer triage: the intended labels are `type-bug`, `frontend`, `theme-ui-ux`, `priority-p2`, `complexity-medium`, and `status-in-progress`. The contributor account cannot assign repository labels.

At narrow desktop widths around 700–714 CSS pixels, Trinity's Dashboard chrome does not reflow as a coordinated responsive layout:

- The shared top navigation lets `Settings` collide with the WebSocket `Connected` status.
- The global Trinity header has no clear full-width lower divider separating it from Dashboard controls.
- Dashboard statistics, telemetry, and view/filter controls become crowded.
- In Timeline mode, zoom controls, `Active only`, the activity legend, live status, and event count compete for one row; labels wrap internally and the toolbar clips beyond the right edge.

This is one narrow-viewport responsive defect across the Dashboard's stacked navigation and control surfaces.

## To reproduce

1. Open an authenticated Trinity Dashboard from current `dev`.
2. Set the browser viewport to approximately `714 × 863` CSS pixels.
3. Use a maximum-content state where possible: WebSocket connected, Operations badge at `99+`, and build metadata visible.
4. Select `Timeline` and the `24h` range.
5. Observe the `Settings` / `Connected` collision and clipped Timeline controls and legend.

On untouched `dev` (`7ee9a19c2210`), geometric inspection at 714px measured:

- `Settings` / connection-status intersection: `true`
- Replay Timeline toolbar overflow: `140px`
- global header bottom border: `0px`

## Expected behavior

- Global navigation, Dashboard summary controls, and Timeline controls reflow or collapse without overlapping or clipping.
- Every primary route and the text connection state remain reachable at narrow widths.
- The global Trinity header has a continuous lower divider in light and dark themes.
- Horizontal scrolling remains confined to the Timeline canvas.
- Controls preserve their existing behavior and keyboard accessibility.

## Acceptance criteria

- [ ] No document-level or Replay Timeline toolbar overflow at 375, 639, 640, 700, 714, 768, 1024, 1100, 1101, 1279, 1280, and 1440 CSS pixels.
- [ ] `Settings` and connection status never intersect, including with `99+`, build information, Enterprise navigation, and `Disconnected` visible.
- [ ] All primary navigation destinations remain reachable through visible links or an accessible compact menu.
[...]

PR #1887, opened 2026-07-30
fix(ui): widen the Settings panel to fit its own tab strip (#1862)

### LINKED ISSUE #1862: bug: default Settings panel too narrow — components don't fit the available width
## Summary

The default panel of the Settings view is too narrow for its content: some components are wider than the space the panel gives them, so they don't fit cleanly (overflow / clipping / cramped layout). Widen the default Settings panel so all current components fit within it without breaking layout.

## Context

Operator-reported visual defect (2026-07-28): opening Settings on its default panel shows components that need more horizontal room than the panel provides. The fix is sizing/layout, not new functionality — give the panel enough width (or make it responsive to its content) so the existing components render as designed.

_[Fill in when picking up: which specific components overflow on the default panel, plus a screenshot of the current state.]_

## Acceptance Criteria

- [ ] The default Settings panel is wide enough that all its current components fit without overflow, clipping, or awkward wrapping
- [ ] No horizontal scrollbar appears inside the panel at standard desktop viewport widths
- [ ] The widened layout degrades gracefully at narrower viewports (no regression on smaller windows)
- [ ] Other Settings panels/tabs are unaffected or benefit consistently from the same width change

## Technical Notes

- Likely a container width / max-width constraint on the Settings panel layout rather than per-component styling.
- Check whether the width constraint is shared across all Settings tabs — if so, widening once at the container level is preferable to per-panel overrides.

dolho:
Closes #1862.

## Diagnosis (the issue asked for this)

The issue left a placeholder: *"Fill in when picking up: which specific components overflow on the default panel."* Measured at a 1440px viewport:

```
main             = 896px   (max-w-4xl)
main.scrollWidth = 957px   -> 61px of horizontal overflow
overflowing child: nav.-mb-px.flex.space-x-6   (925px inside an 832px box)
```

**The overflowing component is the view's own tab strip.** Ten tabs (General … Activation) need ~925px, so the tab row rendered past the right edge of the content card beneath it. That is the only offender — no input, select, table or form overflows its parent.

## Root cause

Settings was the **only** top-level view constrained to `max-w-4xl` (896px):

| View | max-width |
|---|---|
| Agents, Operations, Templates, ExecutionDetail | `max-w-7xl` |
| **Settings** | **`max-w-4xl`** |

So the fix matches the house convention rather than inventing a bespoke width. `max-w-*` is a ceiling, so nothing changes below it.

## Verified across viewports — before vs after

| width | before | after | |
|---|---|---|---|
| 1440px | 4 horizontal scrollers, main 896px | **0 scrollers**, main 1280px | fixed |
| 1024px | doc scroll 1014 | doc scroll 1009 | slightly better |
| 768px | doc scroll 949 | doc scroll 949 | **identical — no regression** |

Measured by re-running the same probe against `max-w-4xl` and `max-w-7xl` on a live instance, not inferred.

## Honest scope note

**768px still overflows, before and after.** Ten tabs cannot fit a 768px viewport at any container width — that needs the strip to wrap or scroll deliberately, which is the class of #1754 and explicitly not addressed here. This issue is the panel being narrower than the viewport allows; that one is the strip being wider than a small viewport.

## Also

Corrected a comment on the users table justifying trimmed padding as fitting *"the max-w-4xl card"* (#995). That constraint is gone; left as-is the comment would push the next person to preserve a workaround for a limit that no longer exists.

## Screenshots

Before / after at 1440px are attached to the issue thread. The visible difference: the tab strip previously ran past the right edge of the content card; it now sits inside it, and the cards use the full width.

### DIFF
--- a/src/frontend/src/views/Settings.vue
+++ b/src/frontend/src/views/Settings.vue
@@ -2,7 +2,14 @@
   <div class="min-h-screen bg-gray-100 dark:bg-gray-900">
     <NavBar />
 
-    <main class="max-w-4xl mx-auto py-6 sm:px-6 lg:px-8">
+    <!-- #1862: `max-w-4xl` (896px) was too narrow for this view's own tab
+         strip — the ten tabs measure ~925px, so `main` overflowed by ~61px
+         and scrolled horizontally at any desktop width. Settings was also
+         the ONLY top-level view on 4xl; Agents, Operations, Templates and
+         ExecutionDetail all use 7xl. Matching them fixes the overflow and
+         removes the odd one out rather than inventing a bespoke width.
+         `max-w-*` is a ceiling, so narrow viewports are unaffected. -->
+    <main class="max-w-7xl mx-auto py-6 sm:px-6 lg:px-8">
       <div class="px-4 py-6 sm:px-0">
         <div class="mb-8">
           <h1 class="text-3xl font-bold text-gray-900 dark:text-white">Settings</h1>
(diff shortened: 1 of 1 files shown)


vybe (approved):
Validated via /validate-pr: measured diagnosis (tab strip = the only overflowing child), fix aligns Settings with the max-w-7xl convention of every other top-level view, ceiling semantics leave narrow viewports untouched. LGTM.

PR #1890, opened 2026-07-30
fix(security): admin gates reject agent-scoped keys

dolho:
Fixes a privilege-escalation class reported privately. **P1.**

## The bug

An agent-scoped MCP key resolves to its owner **carrying the owner's role**, so on a default admin-owned install every agent's injected `TRINITY_MCP_API_KEY` satisfied every `assert_admin` / `require_admin` gate.

The chain that makes it P1: a prompt-injected agent repoints the platform skills library at an attacker-controlled repo through the generic settings setter; the scheduled sync pulls it; the fleet sweep writes attacker-authored `SKILL.md` files into **every running agent**. A `SKILL.md` is instructions Claude executes, persisted on the durable home volume. One compromised agent becomes all of them.

## Fixed at the gate, not the endpoint

This is the **third** occurrence of the class — trinity-ops-agent#232, #1644 (retention acknowledge), #1816 (system-agent restart) — each previously closed by bolting `reject_agent_principal` onto one more endpoint. Three occurrences means the gate is wrong, not the endpoints.

So `assert_admin` and `require_admin` now reject agent principals. The class closes at once, and a settings key or endpoint added tomorrow inherits the protection without anyone remembering to ask for it.

## Why this is safe — verified, not assumed

| Concern | Finding |
|---|---|
| Do agent flows need an admin gate? | **No.** Heartbeat, structured reports and the #1083 result callback authorize on `current_user.agent_name` self-checks. |
| Does `trinity-system` break? | **No.** `User.agent_name` is set only for `scope == "agent"`, so system-scoped keys never trip the guard. |
| Do MCP tools hit admin endpoints? | **Yes — six do.** The original audit was wrong (it only checked the skills tools against `/skills/library/sync`). Corrected below. |
| Do agents lose legitimate capability? | **No.** `require_role` is deliberately unchanged — MCP `deploy_system` uses `require_role("creator")`. |

That last row is the **grant-vs-use line** from `learnings.md` landing where it should: the endpoint that *uses* a capability may be agent-callable; the endpoint that *grants* one is human-only. Admin is the granting tier.

## Tests — 9, and no `MagicMock` anywhere

Deliberately: a bare `MagicMock` auto-creates a truthy `.agent_name`, so it reads as an agent key and passes for the wrong reason — the trap #1816 recorded and this issue's AC called out. Every principal is an explicit dataclass with every field stated.

Both gates reject an agent principal carrying `role="admin"`; the exploit is driven through the **real** `update_setting` handler; non-regression for a human admin, a system-scoped key, a connector key and a non-admin. **Verified 3 fail with the gate fix reverted.**

## Test-double churn, and why it's honest

The change broke 6 pre-existing tests. All six were incomplete principal doubles that set `connector_agent = None` — churn from when #1310 added the *connector* guard — but omitted `agent_name`: either absent (raising `AttributeError`) or a truthy `MagicMock` reading as an agent key.

I checked each rather than assuming, because 6 failures could equally have meant a real flow broke. They didn't; the doubles were simply one guard out of date. Same convention applied.

**815 passed** across the auth/admin/settings/ops/subscription/retention suites, at failure parity with base (3 pre-existing `bcrypt` environmental failures before and after).

Closes the privately-tracked issue.

---

## Correction: MCP surface fallout (review of this PR)

The claim above was inaccurate. Six MCP tools call admin-gated routes. Re-audited, split by whether the call is a **grant** or a **use**:

### Four intended 403s — deliberate, no change

The grant-vs-use line working as designed. Each hands out or revokes a platform credential, which is exactly what an agent-scoped key must not be able to do:

| Tool | Route | Why it stays closed |
|---|---|---|
| `get_credential_encryption_key` | `GET /api/agents/{name}/credentials/encryption-key` | Pulls the platform master key |
| `get_agent_ssh_access` | `POST /api/agents/{name}/ssh-access` | Already self-documents as admin-only |
| `register_subscription` | `POST /api/subscriptions` | Grants a platform credential |
| `delete_subscription` | `DELETE /api/subscriptions/{id}` | Revokes one |

An agent needing any of these is asking a human to perform a privileged act, and a 403 is the correct answer.

### Two fixed — reads, not grants

These left a half-working workflow, which is worse than a cleanly closed door:

**`trigger_health_check`** (`POST /api/monitoring/agents/{name}/check`) — an ops agent could READ health (`GET /status`, `GET /agents/{name}`, both non-admin) but not REFRESH it. The `require_admin` was redundant regardless: `AuthorizedAgentByName` already restricts the endpoint to callers who own or are shared the agent. Now `get_current_user` + the unchanged per-agent gate.

**`list_subscriptions`** (`GET /api/subscriptions`) — `assign_subscription` / `clear_agent_subscription` / `get_agent_auth` all kept working under owner gates, so an agent could ASSIGN a subscription it could no longer ENUMERATE.

Note this one is **scoped, not un-gated**. Un-gating was the first attempt and it traded one disclosure for another: the payload carries `owner_email` and the full agent-name list of *every* subscription, so an ungated read hands any `role=user` account a fleet-wide owner-email + agent-name enumeration oracle — the disclosure class Invariant #8 exists to prevent, reintroduced by a fix for a different disclosure. Resolution: admin keeps the fleet view, everyone else gets `owner_id`-scoped through a filter already in the accessor's contract. An agent key is excluded from the fleet view by the `agent_name` term *even though it carries the owner's admin role* — which is the whole point of ent#293.

Also corrected a stale "Admin-only" docstring on `get_subscription_usage`, whose gate has always been `get_current_user`.

### Pinned

`tests/unit/test_293_admin_gate_rejects_agent_keys.py` now covers all three resolutions plus the four intended 403s, so the next change to the gate cannot silently re-break them.

One note on how that test is written, since it bit during this fix: the access-gate assertion reads the **source annotation**, not the resolved `Depends`. Sibling monitoring tests override that dependency with a lambda, so a runtime check passes standalone and fails in a full run purely on collection order — caught by running the wider selection rather than the file alone.

### DIFF
--- a/src/backend/dependencies.py
+++ b/src/backend/dependencies.py
@@ -687,11 +687,17 @@ def require_admin(current_user: User = Depends(get_current_user)) -> User:
 
     An agent-scoped key resolves to its owner CARRYING THE OWNER'S ROLE, so on a
     default admin-owned install every agent's injected `TRINITY_MCP_API_KEY`
-    satisfied this gate. That is three consecutive incidents of one class —
+    satisfied this gate. ent#297 traced FIVE occurrences of one class —
     trinity-ops-agent#232, #1644 (retention acknowledge), #1816 (system-agent
-    restart) — each previously closed by bolting `reject_agent_principal` onto
-    one more endpoint. Three occurrences means the GATE was wrong, not the
-    endpoints, so the rejection moves here and the class closes at once.
+    restart), ent#236 and ent#293 (skills-library repointing) — each previously
+    closed by bolting `reject_agent_principal` onto one more endpoint, 18 of
+    them, against 114 admin-gated call sites. Five occurrences means the GATE
+    was wrong, not the endpoints, so the rejection moves here.
+
+    Note this closes the class for THIS gate and `assert_admin`. It does not
+    reach `require_role("admin")`, a third spelling that must not exist — see
+    `require_role`'s docstring for why it stays permissive and the guard that
+    keeps the spelling from coming back.
 
     Safe by construction, verified rather than assumed: the agent-key flows that
     must keep working — heartbeat, structured reports, the #1083 result callback
@@ -727,6 +733,24 @@ def require_role(min_role: str):
 
     Raises:
         HTTPException(403): If user's role is below the minimum required
+
+    **Deliberately does NOT call `reject_agent_principal`** — read this before
+    "fixing" it to match `require_admin`/`assert_admin` (ent#293/ent#297).
+
+    Those two gates reject agent principals because an admin gate is never
+    agent-callable. This one is different: its main consumer is
+    `require_role("creator")` on `POST /api/agents`, and agent-spawned agent
+    creation is a SUPPORTED feature (ent#69 Part 2 — the spawn persists
+    `spawned_by_agent`, auto-grants the parent→child permission edge, and is
+    bounded by `enforce_agent_spawn_scope` + the per-parent spawn rate limit).
+    Adding a blanket rejection here would break ghost spawning outright.
+
+    The corollary, and the trap ent#297 walked into: `require_role("admin")` was
+    a THIRD spelling of an admin gate that this permissiveness left open after
+    the other two were closed. Do not write `require_role("admin")` — use
+    `require_admin`, which is equivalent (`admin` is last in ROLE_HIERARCHY) and
+    carries the agent rejection. `tests/unit/test_293_admin_gate_rejects_agent_keys.py`
+    fails the build if a `require_role("admin")` reappears.
     """
     def _require_role(current_user: User = Depends(get_current_user)) -> User:
         _reject_connector_principal(current_user)
@@ -961,11 +985,17 @@ def assert_admin(current_user: User, *, detail: str = "Admin access required") -
 
     An agent-scoped key resolves to its owner CARRYING THE OWNER'S ROLE, so on a
     default admin-owned install every agent's injected `TRINITY_MCP_API_KEY`
-    satisfied this gate. That is three consecutive incidents of one class —
+    satisfied this gate. ent#297 traced FIVE occurrences of one class —
     trinity-ops-agent#232, #1644 (retention acknowledge), #1816 (system-agent
-    restart) — each previously closed by bolting `reject_agent_principal` onto
-    one more endpoint. Three occurrences means the GATE was wrong, not the
-    endpoints, so the rejection moves here and the class closes at once.
+    restart), ent#236 and ent#293 (skills-library repointing) — each previously
+    closed by bolting `reject_agent_principal` onto one more endpoint, 18 of
+    them, against 114 admin-gated call sites. Five occurrences means the GATE
+    was wrong, not the endpoints, so the rejection moves here.
+
+    Note this closes the class for THIS gate and `assert_admin`. It does not
+    reach `require_role("admin")`, a third spelling that must not exist — see
+    `require_role`'s docstring for why it stays permissive and the guard that
+    keeps the spelling from coming back.
 
     Safe by construction, verified rather than assumed: the agent-key flows that
     must keep working — heartbeat, structured reports, the #1083 result callback
--- a/src/backend/dependencies.py
+++ b/src/backend/dependencies.py
@@ -680,9 +680,30 @@ def require_admin(current_user: User = Depends(get_current_user)) -> User:
     Dependency that requires the current user to be an admin.
 
     Raises:
-        HTTPException(403): If user is not an admin
+        HTTPException(403): If user is not an admin, or is an agent/connector
+            principal.
+
+    ent#293 — an ADMIN gate is never agent-callable.
+
+    An agent-scoped key resolves to its owner CARRYING THE OWNER'S ROLE, so on a
+    default admin-owned install every agent's injected `TRINITY_MCP_API_KEY`
+    satisfied this gate. That is three consecutive incidents of one class —
+    trinity-ops-agent#232, #1644 (retention acknowledge), #1816 (system-agent
+    restart) — each previously closed by bolting `reject_agent_principal` onto
+    one more endpoint. Three occurrences means the GATE was wrong, not the
+    endpoints, so the rejection moves here and the class closes at once.
+
+    Safe by construction, verified rather than assumed: the agent-key flows that
+    must keep working — heartbeat, structured reports, the #1083 result callback
+    — authorize on `current_user.agent_name` self-checks and never touch an admin
+    gate. System-scoped keys are unaffected because `User.agent_name` is set only
+    for `scope == "agent"`, so `trinity-system` still passes.
+
+    This is the grant-vs-use line from `learnings.md`: the endpoint that USES a
+    capability may be agent-callable; the endpoint that GRANTS one is human-only.
     """
     _reject_connector_principal(current_user)
+    reject_agent_principal(current_user)
     if current_user.role != "admin":
         raise HTTPException(
             status_code=status.HTTP_403_FORBIDDEN,
@@ -932,11 +953,31 @@ def get_owned_agent_by_name(
 def assert_admin(current_user: User, *, detail: str = "Admin access required") -> None:
     """Imperative admin gate — parity with the ``require_admin`` Depends form.
 
-    Rejects connector principals first (they are consumption-only), then requires
-    ``role == "admin"``. Raises 403 on failure; returns None on success. Use in a
-    router body where the admin check is inline rather than a ``Depends``.
+    Rejects connector AND agent principals (neither is a human admin), then
+    requires ``role == "admin"``. Raises 403 on failure; returns None on success.
+    Use in a router body where the admin check is inline rather than a ``Depends``.
+
+    ent#293 — an ADMIN gate is never agent-callable.
+
+    An agent-scoped key resolves to its owner CARRYING THE OWNER'S ROLE, so on a
+    default admin-owned install every agent's injected `TRINITY_MCP_API_KEY`
+    satisfied this gate. That is three consecutive incidents of one class —
+    trinity-ops-agent#232, #1644 (retention acknowledge), #1816 (system-agent
+    restart) — each previously closed by bolting `reject_agent_principal` onto
+    one more endpoint. Three occurrences means the GATE was wrong, not the
+    endpoints, so the rejection moves here and the class closes at once.
+
+    Safe by construction, verified rather than assumed: the agent-key flows that
+    must keep working — heartbeat, structured reports, the #1083 result callback
+    — authorize on `current_user.agent_name` self-checks and never touch an admin
+    gate. System-scoped keys are unaffected because `User.agent_name` is set only
+    for `scope == "agent"`, so `trinity-system` still passes.
+
+    This is the grant-vs-use line from `learnings.md`: the endpoint that USES a
+    capability may be agent-callable; the endpoint that GRANTS one is human-only.
     """
     _reject_connector_principal(current_user)
+    reject_agent_principal(current_user)
     if current_user.role != "admin":
         raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail=detail)
 
--- a/src/backend/routers/agents.py
+++ b/src/backend/routers/agents.py
@@ -31,7 +31,7 @@ from models import (
     User,
 )
 from database import db
-from dependencies import get_current_user, decode_token, require_role, AuthorizedAgentByName, OwnedAgentByName, CurrentUser, enforce_agent_spawn_scope
+from dependencies import get_current_user, decode_token, require_role, require_admin, AuthorizedAgentByName, OwnedAgentByName, CurrentUser, enforce_agent_spawn_scope
 from services.docker_service import (
     get_agent_container,
     get_agent_by_name,
@@ -826,7 +826,20 @@ async def force_release_agent(
 @router.post("/{agent_name}/circuit-breaker/reset")
 async def reset_circuit_breaker_endpoint(
     agent_name: AuthorizedAgentByName,
-    current_user: User = Depends(require_role("admin")),
+    # ent#293/ent#297: was `require_role("admin")` — the THIRD spelling of an
+    # admin gate, and the one this PR's first pass missed. `require_role`
+    # rejects connector principals but NOT agent ones, deliberately: agent
+    # -spawned agent creation runs through `require_role("creator")` (ent#69
+    # Part 2). So after `require_admin`/`assert_admin` were closed, an
+    # admin-owned agent's own key still passed HERE — and ent#297 names
+    # circuit-breaker resets in its blast radius. An agent that can re-close its
+    # own dispatch breaker on demand defeats the control built to contain it.
+    #
+    # `require_admin` is equivalent for this case (`admin` is last in
+    # ROLE_HIERARCHY, so ">= admin" is "== admin") and additionally rejects
+    # agent principals. Swapping this ONE call site is the fix; adding the
+    # rejection to `require_role` itself would break ghost spawning.
+    current_user: User = Depends(require_admin),
 ):
     """Force the agent's circuit breakers to closed (#921, #526).
 
--- a/src/backend/routers/monitoring.py
+++ b/src/backend/routers/monitoring.py
@@ -379,12 +379,20 @@ async def get_agent_health_history(
 @router.post("/agents/{agent_name}/check", response_model=AgentHealthDetail)
 async def trigger_health_check(
     agent_name: AuthorizedAgentByName,
-    current_user: User = Depends(require_admin)
+    current_user: User = Depends(get_current_user)
 ):
     """
     Trigger an immediate health check for an agent.
 
-    Admin only. Forces a fresh health check regardless of schedule.
+    Forces a fresh health check regardless of schedule.
+
+    ent#293 review: the `require_admin` here was redundant — `AuthorizedAgentByName`
+    already restricts this to callers who own or are shared the agent — and once
+    admin gates stopped accepting agent-scoped keys it made the surface
+    incoherent: an ops agent could READ health (`GET /status`,
+    `GET /agents/{name}`, both non-admin) but not REFRESH it. Triggering a probe
+    on an agent you can already see is a USE of a capability, not a grant of one,
+    so it takes the same gate as its siblings.
 
     #631 — this endpoint is the documented manual recovery path out of the
     dormant circuit state. Reset the circuit before running the check so the
--- a/src/backend/routers/subscriptions.py
+++ b/src/backend/routers/subscriptions.py
@@ -114,14 +114,29 @@ async def list_subscriptions(
     current_user: User = Depends(get_current_user)
 ):
     """
-    List all subscriptions with their assigned agents.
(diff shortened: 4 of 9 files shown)


vybe (changes requested):
Validated via `/validate-pr`. The security reasoning here is right and the fix is the correct shape — three occurrences of one class means the **gate** was wrong, not the endpoints, and moving the rejection into `require_admin`/`assert_admin` closes it at the root. I verified the two load-bearing safety claims independently and both hold:

- **System-scoped keys are unaffected** — `dependencies.py:360` sets `User.agent_name` only when `scope == "agent"`, so `trinity-system` still passes both gates.
- **Agent-originated flows don't touch an admin gate.** I enumerated all 116 admin-gated routes across all 63 routers and cross-checked every agent-authored path: heartbeat (`agents.py:1182`), the #1083 result callback (`agents.py:1252`), reports (`reports.py:56`), notifications, `emit_event`/`subscribe_to_event`, reminders, `write_user_memory`, `share_file`, and all of `routers/internal.py`. Every one authorizes via a `current_user.agent_name` self-check. None regresses.

Tests pass locally at `7cdc127b` — 113 passed across the new file plus the four touched suites and both `test_1310_auth_*` wiring/consolidation suites. The no-MagicMock discipline in the new test file is the right call (a bare `MagicMock` auto-creates a truthy `.agent_name` and would pass for the wrong reason — the #1816 trap).

**One blocking correction: the claim "Do MCP tools hit admin endpoints? No." is not accurate.**

Six MCP tools call admin-gated backend routes. The audit only covered the skills tools, which is why this was missed. The MCP layer's only visibility gate is `connectorDenied` (`src/mcp-server/src/server.ts:244,279-281`), which filters `scope === "connector"` — **not** `scope === "agent"` — and none of the six has its own scope check. So on a default admin-owned install all six previously succeeded for an agent-scoped key and newly 403:

| MCP tool | Backend route | Gate |
|---|---|---|
| `get_credential_encryption_key` (`agents.ts:649`) | `GET /api/credentials/encryption-key` | `require_admin` — `credentials.py:451` |
| `get_agent_ssh_access` (`agents.ts:669`) | `POST /api/agents/{name}/ssh-access` | `require_admin` — `agent_ssh.py:19` |
| `trigger_health_check` (`monitoring.ts:164`) | `POST /api/monitoring/agents/{name}/check` | `require_admin` — `monitoring.py:379` |
| `register_subscription` (`subscriptions.ts:43`) | `POST /api/subscriptions` | `assert_admin` — `subscriptions.py:60` |
| `list_subscriptions` (`subscriptions.ts:91`) | `GET /api/subscriptions` | `assert_admin` — `subscriptions.py:122` |
| `delete_subscription` (`subscriptions.ts:227`) | `DELETE /api/subscriptions/{id}` | `assert_admin` — `subscriptions.py:198` |

Four of these I read as **intended** consequences that fit the grant-vs-use line squarely, and I'd keep them: a hosted agent pulling the platform master key is an exfiltration surface; `get_agent_ssh_access` already says "Admin only" in its own description; `register_subscription`/`delete_subscription` are human admin grants. Worth stating explicitly in the PR body rather than leaving them as undocumented fallout.

**Two are the actual problem, because they are reads, not grants:**

- **`trigger_health_check`** — an admin-owned ops/monitoring agent running fleet-health routines newly 403s. Its siblings `get_fleet_health` (`GET /api/monitoring/status`) and `get_agent_health` (`GET /api/monitoring/agents/{name}`) are *not* admin-gated and keep working, so the result is an incoherent surface: an agent may read health but not refresh it.
- **`list_subscriptions`** — `assign_subscription`, `clear_agent_subscription` and `get_agent_auth` all stay working (owner/access gates at `subscriptions.py:251,344,396`), so the **write** step of the subscription workflow survives while the **read** step that discovers subscription names breaks. An agent can assign a subscription it can no longer enumerate.

Neither is a grant, so neither is justified by this PR's own rationale — and both leave a half-working workflow, which is worse than a cleanly closed door.

### What I'd like before merge

1. Give `trigger_health_check` and `list_subscriptions` a path that keeps working for an agent-scoped key — either an MCP-layer scope gate that fails cleanly with a comprehensible message, or a non-admin backend path consistent with their already-non-admin siblings. Your call which; the coherence of the surface is the requirement, not the mechanism.
2. Correct the "no MCP tools hit admin endpoints" line in the PR body, and record the four intended 403s as deliberate.
3. A test pinning whichever resolution you pick, so the next gate change doesn't silently re-break it.

To be clear about priority: I'm not disputing the fix or asking you to weaken the gate. The P1 is real and this is the right place to close it. Ship the same change with those two tools coherent and I'll merge it immediately.

dolho:
Review addressed — all three items.

**(1) The two half-working workflows.** Fixed in `ebd5c16`.

`trigger_health_check` → `get_current_user`, keeping `AuthorizedAgentByName`. The `require_admin` was redundant anyway; the per-agent gate was already doing the work.

`list_subscriptions` → **scoped, not un-gated**. I un-gated it first and that was wrong: the payload carries `owner_email` and the full agent-name list of every subscription, so an ungated read hands any `role=user` account a fleet-wide owner-email + agent-name enumeration oracle. That's the Invariant #8 disclosure class, reintroduced by a fix for a different disclosure. Admin keeps the fleet view; everyone else is `owner_id`-scoped via a filter already in the accessor's contract. An agent key stays out of the fleet view on the `agent_name` term despite carrying the owner's admin role — which is the point of the issue.

**(2) + (3) PR body.** Corrected. The "no MCP tools hit admin endpoints" line now says six do and links to a new section recording the four intended 403s (`get_credential_encryption_key`, `get_agent_ssh_access`, `register_subscription`, `delete_subscription`) as deliberate, with the grant-vs-use rationale per tool.

**Pinning.** Four new tests cover all three resolutions plus the four intended 403s.

One thing worth flagging from writing them, since it's the kind of thing that silently rots: my first version asserted the access gate by reading the **resolved** `Depends`. Sibling monitoring tests override that dependency with a lambda, so it passed standalone and failed in a full run purely on collection order (`assert 'get_authorized_agent_by_name' in {'<lambda>'}`). It now reads the source annotation via AST instead. Caught by running the wider selection rather than the file alone — noting it because a structural guard that depends on collection order isn't a guard.

485 passed across the monitoring / subscriptions / credentials / ssh / auth-invariant (#1310, #186) selection.

dolho:
## /review Report — #1890 + #1893 (the ent#297 pair)

**Branches**: `fix/293-admin-gate-rejects-agent-keys` → `dev` (MB `04d11cf2`) · `fix/297-retention-settings-hardening` → `dev` (MB `39f29c64`)
**Files Changed**: 8 (+448/−15) · 5 (+559/−9)
**Scope**: CLEAN both
**Plan Completion (ent#297 ACs)**: 6 done / 1 partial / 0 not done / 1 changed / 0 unverifiable

Self-review, so weighted toward finding problems — my self-review of #1885 earlier this week missed that I'd reintroduced the exact duplication that issue forbade. Four findings, all fixed.

---

### Critical

**[C1] Auth boundary: a THIRD admin-gate spelling left open** (Confidence: 10/10) — **FIXED `bf23a2d7`**

`src/backend/routers/agents.py:808`

```python
@router.post("/{agent_name}/circuit-breaker/reset")
async def reset_circuit_breaker_endpoint(
    agent_name: AuthorizedAgentByName,
    current_user: User = Depends(require_role("admin")),   # <- not require_admin
```

and the reason it stayed open (`dependencies.py:590`):

```python
def _require_role(current_user: User = Depends(get_current_user)) -> User:
    _reject_connector_principal(current_user)      # connector, yes
    # reject_agent_principal — absent
```

**Issue.** #1890 closes `require_admin` and `assert_admin`. There are **three** spellings of an admin gate; `require_role("admin")` is the third, and after the other two were closed it was the only one left accepting an admin-owned agent's own key. Its single call site is `POST /api/agents/{name}/circuit-breaker/reset` — a route **ent#297 names in its own blast-radius list** ("circuit-breaker resets"). An agent that can re-close its own dispatch breaker on demand defeats the control built to contain it.

**The naive fix is wrong**, which is why this needed checking rather than patching: adding `reject_agent_principal` to `require_role` would break `require_role("creator")` on `POST /api/agents`, and agent-spawned agent creation is a *supported* feature (ent#69 Part 2 — `spawned_by_agent` provenance, auto-granted parent→child edge, `enforce_agent_spawn_scope`). It would kill ghost spawning.

**Fix applied**: swap the one call site to `require_admin` — equivalent for the admin case (`ROLE_HIERARCHY = [..., "admin"]`, so `>= admin` is `== admin`) and it rejects agent principals. `require_role`'s docstring now records why it stays permissive, so nobody "fixes" it later and breaks spawning.

Pinned by an AST scan over `routers/`. **AST, not a regex** — the comment on the fixed call site necessarily *contains* the offending string, and my first version flagged its own documentation. That is the second time this session I wrote a source-scanning guard that matched its own explanatory text; it is now the reason `_calls_in_body` exists in this file. Meta-tested by planting a violation in `monitoring.py` and confirming it fires.

---

### Informational

**[I1] #1893 invented a bound instead of adopting the existing one** (Confidence: 9/10) — **FIXED `825afbfd`**

`config.py` had `_DAYS_MAX = 36500` (100y). The enterprise `retention` module validates the **same windows** at `le=3650`:

```python
execution_row_retention_days: Optional[int] = Field(None, ge=0, le=_MAX_DAYS)   # _MAX_DAYS = 3650
```

Two validated write paths, two different contracts for one value: an admin could store a window via the OSS route that the managed panel then refuses to edit — surfaced by its own GET, rejected by its own PUT. Aligned to 3650 with a test pinning it, since the enterprise constant can't be imported (private submodule; OSS must build without it). This is the duplicated-policy-constant drift class from `learnings.md`, committed inside a fix whose theme is "one path".

**[I2] Two false claims in #1893's own prose** (Confidence: 10/10) — **FIXED `825afbfd`**

- *"the seven retention windows"* — there are **eight**. `agent_reminders_retention_days` (#1296) joined `RETENTION_OPS_KEYS` and my prose didn't. Stated in three code comments, the commit message, and the PR body, while my own test asserted `== 8`.
- *"this is the ONLY write path"* — false. `PUT /api/enterprise/retention/config` is a second one. It was **already** typed and range-validated, so no security claim rested on this, but the sentence was wrong.

**[I3] #1890's gate docstrings undercounted the class** (Confidence: 8/10) — **FIXED `bf23a2d7`**

Both said *"three consecutive incidents"*. ent#297 traced **five** (ops-agent#232, #1644, #1816, ent#236, ent#293), against 18 bolt-ons and 114 admin-gated call sites. In a security docstring the count is the argument for why the gate moved, so an undercount weakens the reasoning a future reader inherits.

**[I4] Enterprise panel can't set one of the eight windows** (Confidence: 8/10) — **NOT FIXED, enterprise-side**

`RetentionConfigUpdate` covers 7 of 8; `agent_reminders_retention_days` (#1296) is absent. So the managed retention panel cannot set it, and an entitled operator has no UI for that window. Pre-existing, in the private repo, out of scope for these PRs — flagging for a follow-up.

---

### Clean (verified, cited)

- **AC #4 blocklist is not a bypass.** `DELETE /api/settings/{key}` is untouched, but deleting a retention row reverts to `OPS_SETTINGS_DEFAULTS`, which are the **wide** values (30/90/7/180/30/90/90/90 — read at runtime, not assumed). Deleting fails safe; it is not a way around the 422.
- **No legitimate caller broken.** Checked before blocking: the Settings UI writes retention only through the enterprise endpoint (`Settings.vue:2524`), and the sole UI use of `/ops/config` is the `ssh_access_enabled` toggle (`Settings.vue:3799`), which the bool validator accepts. Driven end-to-end via `TestClient`: the PoC call 422s, a normal key still 200s, the SSH toggle still 200s.
- **All-or-nothing is real, not aspirational** — validation runs in its own loop before any `db.set_setting`, and a behavioural test asserts the good key in a mixed request never lands.
- **`assert_admin`/`require_admin` parity** — both reject agent + connector, both keep human admins and `trinity-system` (`agent_name` is set only for `scope == "agent"`).
- **Enum/value completeness** — `OPS_SETTINGS_VALIDATION` has parity tests against both `RETENTION_OPS_KEYS` and `OPS_SETTINGS_DEFAULTS`, so a new window or ops setting can't ship unvalidated.
- **Credential exposure** — the new audit `details` carries ops config values; none of the 16 `OPS_SETTINGS_DEFAULTS` keys is a secret (thresholds, day counts, one bool), and answering "who shortened retention, to what, when" is the point.
- **Migrations / SQL / concurrency** — no schema, no SQL, no shared mutable state in either diff.

### Summary
- **Critical: 1** — found and fixed (`bf23a2d7`)
- **Informational: 4** — three fixed, one (I4) is enterprise-side and reported
- **Scope: clean**

The one that mattered was C1, and it is worth naming why it survived the first pass: I audited the gates I *changed* rather than enumerating every way an admin gate can be spelled. Same shape as the #1871 lesson in `learnings.md` — a guard for a "more than one call site" class has to start by enumerating the API surface, not the call name you just fixed.

dolho:
@vybe all three requested changes are in (`ebd5c16a`, `bf23a2d7`), plus one extra the review surfaced.

**1 — the two non-grant reads now keep working, coherently:**
- `trigger_health_check` → dropped to `AuthorizedAgentByName` + `get_current_user`, matching its already-non-admin siblings `get_fleet_status`/`get_agent_health`. Triggering a probe on an agent you can already see is a *use*, not a grant.
- `list_subscriptions` → **owner-scoped, not un-gated.** Dropping the gate entirely was the first attempt and it was wrong: the payload carries `owner_email` + the full agent-name list of every subscription, so an ungated read is a fleet-wide enumeration oracle for any `role=user`. It now filters by `owner_id` (admin still sees the fleet), which restores exactly the read the workflow needs — an agent key resolves to its owner and sees the subscriptions it can already assign.

**2 — PR body corrected:** the "no MCP tools hit admin endpoints" line is replaced with "**Yes — six do**", and the four intended 403s (`get_credential_encryption_key`, `get_agent_ssh_access`, `register_subscription`, `delete_subscription`) are documented as deliberate grant-side denials.

**3 — tests pin every resolution:** `test_triggering_a_health_check_is_not_admin_gated`, `test_listing_subscriptions_is_not_admin_gated`, `test_the_restored_read_is_scoped_not_ungated` (asserts the owner filter, so a future re-ungate fails), and `test_the_granting_endpoints_stay_admin_only`.

**Extra from the review pass — a third gate spelling:** `require_role("admin")` was the one admin gate that still accepted agent keys (it rejects connector but not agent principals, deliberately, because agent-spawned creation runs through `require_role("creator")`). Its one admin-gated site was `POST /api/agents/{name}/circuit-breaker/reset` — an agent re-closing its own dispatch breaker defeats the control built to contain it. Fixed at the call site (not the shared helper, which has a legitimate permissive consumer), guarded by `test_no_route_uses_require_role_admin_as_an_admin_gate`. Recorded in `learnings.md`.

All green, pushed at `bf23a2d7`. Ready for re-review.

dolho:
@vybe all three asks were addressed in `ebd5c16a` + `bf23a2d7`, pushed shortly after your review — the review was never dismissed, so the PR has been sitting as blocked since. Recapping against your list, then what I've just done:

**1. The two incoherent tools.** Both fixed, and they needed different answers:

- `trigger_health_check` → `get_current_user` + the unchanged `AuthorizedAgentByName`. The `require_admin` was redundant anyway: the per-agent gate already restricts the endpoint to callers who own or are shared the agent, so probing an agent you can already see is a *use*, not a grant. Coherent with its non-admin siblings now.
- `list_subscriptions` → **scoped**, not un-gated. Un-gating was my first attempt and it was wrong: the payload carries `owner_email` and the full agent list per subscription, so an ungated read hands any `role=user` account a fleet-wide owner-email and agent-name enumeration oracle — reintroducing a disclosure class while fixing a different one. Admin keeps the fleet view; everyone else is `owner_id`-scoped through a filter already in the accessor's contract. An agent key is excluded from the fleet view by the `agent_name` term even while carrying its owner's admin role, which is exactly ent#293's point.

**2. PR body corrected** — the "no MCP tools hit admin endpoints" line is gone; the table now reads *'Yes — six do. The original audit was wrong (it only checked the skills tools).'* The four intended 403s are recorded as deliberate with the grant-vs-use rationale.

**3. Tests** — `test_293_admin_gate_rejects_agent_keys.py` pins all three resolutions plus the four intended 403s. One implementation note: the access-gate assertion reads the **source annotation** rather than the resolved dependency, because sibling monitoring tests override that dependency with a lambda — a runtime check passes standalone and fails in a full run purely on collection order.

`bf23a2d7` additionally closes a third spelling of the same gate, `require_role("admin")`, which the original audit missed.

Just now: rebased onto current `dev` (clean, no conflicts) so CI re-runs against today's base rather than the 30 Jul one — 279 passed locally across `293/297/admin/subscription/monitoring/1310`. (Three `test_admin_email_login` failures in my local run are a passlib/bcrypt version mismatch in my venv; they reproduce on clean `dev` and pass in CI.)

Ready for re-review.

vybe (approved):
Approving — all three blocking items from my previous review are resolved, and verified against the diff rather than the PR body.

**1. Surface coherence.** `trigger_health_check` drops to `get_current_user` while keeping `AuthorizedAgentByName` (the admin gate was redundant with it all along), and `list_subscriptions` is **scoped, not un-gated** — the right call. Un-gating would have traded the agent-key disclosure for a fleet-wide `owner_email` + agent-name enumeration oracle readable by any `role=user`. `owner_id` scoping was already in the accessor's contract, and an agent key lands in the scoped branch *despite carrying `role="admin"`*, which is the whole point.

**2. PR body corrected**, with the four intended 403s recorded as deliberate rather than left as undocumented fallout.

**3. Pinned** — `test_293_admin_gate_rejects_agent_keys.py` covers both gates, the exploit driven through the real `update_setting` handler, non-regression for human admin / system key / connector / non-admin, both restored reads, and the four grants.

**The third spelling is a real find.** `require_role("admin")` at `POST /{name}/circuit-breaker/reset` survived closing the other two gates, and an agent able to re-close its own dispatch breaker defeats the control built to contain it. Keeping `require_role` permissive (ghost spawning runs through `require_role("creator")`, ent#69 Part 2) and swapping the one call site is correct; the AST guard is the right shape, since the failure mode is a *new* route picking the wrong helper — which no behavioural test of existing routes can see. Good catch that both helper tests read the AST rather than grepping source: the comment documenting what not to write contains the offending string.

Verified locally at `a1f60999`: **118 passed** across the new file, both `test_1310_auth_*` suites, and the four touched suites. All required checks green; `pytest (head/base × 3 seeds)` green.

**One follow-up, not blocking this merge.** `docs/memory/architecture.md:1254` (Invariant #8) now contradicts the code:

> **Role ≠ human (#1644, #1816):** `assert_admin`/`require_admin` answer *what role*, never *is this a human* … needs `reject_agent_principal` **in addition to** the role gate

That is precisely the bolt-it-on-per-endpoint pattern this PR eliminates, and it also still describes `assert_admin` as `_reject_connector_principal` + role only. Left as-is it will tell the next contributor to re-implement the thing that just got centralised. The new hard rule (never write `require_role("admin")`) is also documented only in a docstring and a test.

I'm reconciling this in one architecture.md pass across today's merge batch rather than pushing to this branch — eight PRs in flight touch that file, so batching avoids compounding the conflict. Tracking it explicitly; it will not be dropped.

Cross-tracker refs resolve: ent#293 (P1, `status-in-progress`) and ent#297 (P1, `status-in-dev`). GitHub auto-closes same-repo only, so I'll bump both manually after merge.

PR #1497, opened 2026-07-06
feat(canary): execution-row integrity invariants E-03, G-03, E-04, G-04 (Phase 4)

AndriiPasternak31:
## Summary

Adds the second half of the canary **Phase 4** execution-row integrity batch (epic #411) — the two remaining pure single-table predicates over `schedule_executions`, both **read-only**, registry-driven, with **no endpoint / service / DB change**.

| ID | Invariant | Tier | Severity | Guards |
|----|-----------|------|----------|--------|
| **E-04** | Queued rows carry valid backlog metadata | A | major | `status='queued'` ⇒ `queued_at NOT NULL` AND `backlog_metadata` non-NULL + JSON-parseable — the `backlog_service.drain_next` replay contract. A malformed blob raises `JSONDecodeError` and stalls the FIFO. |
| **G-04** | No raw credentials in `backlog_metadata` | A | critical | A queued row's `backlog_metadata` matches no known secret prefix (`sk-`/`ghp_`/`gho_`/`ghs_`/`ghu_`/`github_pat_`/`xoxb-`/`xoxp-`/`AKIA`/`AIza`/`sk_live_`), word-boundary anchored. Rides E-04's collected bytes. |

Both ride the queued-row metadata the collector now captures (`AgentSnapshot.queued_meta`), scoped **strictly to `status='queued'`** so #1449's deferred terminal-row `backlog_metadata` NULL-out can't false-fire E-04. Older-image DDL without the columns → skip the eid (fail-open).

This PR also carries E-03 (completed rows populated) + G-03 (clock sanity), the terminal-row half of Phase 4, over the shared `_collect_terminal_rows` collector.

## Credential safety

E-04/G-04 violations persist to `canary_violations`, so **neither ever echoes the raw `backlog_metadata`**: E-04 reports the failed-predicate reason code (`queued_at_null` / `backlog_metadata_null` / `backlog_metadata_invalid_json`) + ids; G-04 reports the matched pattern NAME + ids only (one violation per row, stops at first match). Every G-04/E-04 test asserts the secret / raw metadata bytes appear **nowhere** in the persisted record.

> The test suite contains deliberately-fake credential exemplars (e.g. AWS's own `AKIAIOSFODNN7EXAMPLE` docs placeholder, a fake `xoxb-`/`ghp_`) — required to exercise a credential-scanning invariant. No real secrets.

## Stacked on #1450 (PR #1495)

[...]

Issue #882, opened 2026-05-18
Canary Phase 2 + 3: add invariants S-02, E-01, E-05, B-01, S-03, B-02, R-01


obasilakis:
## Summary

Phases 2 and 3 of the Canary Invariant Harness (#411), shipped together in PR #884. Follow-up to PR #653 (Phase 1 — S-01, E-02, L-03).

Adds seven invariants in two waves. The split reflects internal complexity — Phase 2 is the four single-source SQL/Redis checks, Phase 3 adds the three "moderate" ones that each need one new piece of plumbing.

## Context

Phase 1 landed the canary library + service + fleet manifest + admin endpoint. The orchestration invariant catalog (`docs/testing/orchestration-invariant-catalog.md`) identifies 12 high-impact invariants for the starting subset; 3 are now in production. This issue covers the next 7. After it lands, the 2 remaining catalog invariants (E-06, G-01) require either HTTP fan-out or one-shot restart probes — different shape, deferred to a future ticket.

### Phase 2 — single-source SQL/Redis (no new sources)

- [x] **S-02 — No overbooking** added: `ZCARD(agent:slots:A) ≤ agent_ownership.max_parallel_tasks`. Critical. Tier A.
- [x] **E-01 — Terminal-state closure** added: `status='running' AND started_at < now() - (timeout_seconds + 300s)` must be 0. Critical. Tier B. Uses `iso_cutoff()` (architecture invariant #16).
- [x] **E-05 — Dispatched rows have session** added: `status='running' AND started_at < now() - 60s ⇒ claude_session_id IS NOT NULL`. Major. Tier B. Guards #106.
- [x] **B-01 — Queue-status coherence** added: per agent, `db.get_queued_count` matches snapshot's independently-collected queued id-list count. Critical. Tier A. Regression guard against future cache layer / status-filter drift.

### Phase 3 — moderate complexity (one new piece of plumbing each)

- [ ] **S-03 — Slot TTL ≥ execution timeout** added: for every member of `agent:slots:A`, the companion `agent:slot:A:{eid}` HASH has `TTL ≥ execution_timeout_seconds + 300s` (`SLOT_TTL_BUFFER`). Critical. Tier A. Guards #226.
[...]

PR #1894, opened 2026-07-30
fix(canary): give every invariant a legible alert, and a guard that keeps it (#1880)

### LINKED ISSUE #1880: Canary: 5 Phase-4 invariants alert with no name, runbook, or detail — including the critical credential-leak check (G-04)
## Summary

`services/canary_alerts.py` carries a per-invariant `_INVARIANT_NAMES` and `_INVARIANT_RUNBOOKS` dict that make a Slack alert interpretable. **Five invariants shipped without entries in either**, so their green→red alerts degrade to a generic fallback that tells the on-call nothing.

Verified against `dev`:

```
registry : B-01 B-02 E-01 E-02 E-03 E-04 E-05 E-06 G-03 G-04 H-01 L-03 R-01 S-01 S-02 S-03
named    : B-01 B-02 E-01 E-02      E-05           H-01 L-03 R-01 S-01 S-02 S-03
missing  : E-03, E-04, E-06, G-03, G-04   (identical set for names and runbooks)
```

All five shipped in Phase 4 (#1077) and E-06 (#1472); the alerting dicts were never extended alongside the registry.

## What the operator actually sees

Three separate degradations stack, because each render path falls back independently:

| Element | Source | Result for a missing invariant |
|---|---|---|
| Header name | `_INVARIANT_NAMES.get(id, id)` | falls back to the **id itself**, so the header prints it **twice** |
| Summary body | `_render_message` | generic `"<ID> fired N violation(s)."` |
| Forensic block | `_render_forensic` | returns `None` → block omitted |
| Runbook hint | `_INVARIANT_RUNBOOKS.get(id)` | `None` → block omitted |

So a firing G-04 pages as:

```
🔴 canary G-04 G-04 (critical): G-04 fired 1 violation(s).
```

No name, no detail, no next step, and the id stuttered.

## Why this matters

**G-04 is `critical` and it is the credential-leak check** — "no raw credentials in `backlog_metadata`". It is arguably the single alert in the catalog most in need of an unambiguous name and a runbook, and it is one of the five that has neither. The other four are `E-03` (major), `E-04` (major), `E-06` (major), `G-03` (minor).

An alarm nobody can act on is close to an alarm that didn't fire. The canary is default-OFF (`CANARY_ENABLED=1`, staging/dev), which is why this hasn't bitten yet.

## Acceptance criteria

- [ ] `E-03`, `E-04`, `E-06`, `G-03`, `G-04` each have an `_INVARIANT_NAMES` entry
- [ ] Each has an `_INVARIANT_RUNBOOKS` entry pointing the on-call at where to start looking (match the existing one-line style; the catalog holds the full prose)
- [ ] `G-04`'s runbook must **not** instruct anyone to read the raw `backlog_metadata` — the invariant deliberately reports only the matched pattern *name*, never the secret, and its violations persist to `canary_violations`
- [ ] A `_render_message` branch for any of these whose violations carry useful structure (E-03/E-04/G-03/G-04 are per-row and do carry `agent_name`, so they can follow the existing S-01/B-01 shape rather than the fallback)
- [ ] **A guard test that fails when a registry id has no name/runbook**, so the next invariant cannot ship with this gap — the root cause is that the registry and the alerting dicts are two hand-maintained lists with nothing tying them together

The last item is the real fix; the five entries are the symptom.

## References

- Found during `/review` of #1879 (H-01, #1813), which added its own entries because H-01's entire purpose is a legible alarm. Deliberately not widened there to keep that PR's blast radius honest.
- Registration requirement now documented in `docs/memory/requirements/infrastructure.md` §31 (Registration bullet), including this known gap.
- Missing invariants shipped in #1077 (E-03/E-04/G-03/G-04) and #1472 (E-06).

webmixgamer:
## Summary

Five registry invariants — `E-03`, `E-04`, `E-06`, `G-03`, `G-04` — shipped with no entry in the per-invariant Slack alert surfaces, so a green→red page rendered:

```
🚨 canary G-04 G-04 (critical): G-04 fired 1 violation(s).
```

No name, no evidence, no next step, and the id stuttered — on the **critical credential-leak check**. Default-OFF is why it hadn't bitten yet; #1876 wired `CANARY_ENABLED` into prod compose two days ago, so it's now one env var away from mattering.

Root cause is a mechanism gap, not diligence: #882 populated every alert surface for its invariants, #1497 and #1472 populated none.

**After:**

```
🚨 G-04 Credential pattern in backlog metadata — critical
2 queued row(s) matched a credential pattern (github_pat, slack_bot_token) across 2 agent(s): prospector, webmaster.
  • *prospector* `exec-9f2a`: matched `github_pat`
  • *webmaster* `exec-71bd`: matched `slack_bot_token`
_Rotate the matched credential first — `backlog_metadata` is stored plaintext and survives until the row is deleted…_
```

## Three things that go beyond the issue's written ACs

Each was surfaced and approved before implementing, not discovered afterwards.

**1 · There are four id-keyed surfaces, not two.** `_INVARIANT_NAMES`, `_INVARIANT_RUNBOOKS`, and the branch chains in `_render_message` **and** `_render_forensic` — all four independently covering the identical 10 of 15 ids. The two the issue didn't name carry the per-row *evidence*, so satisfying AC1/AC2/AC4 as written would have given G-04 a name, a runbook, a count, and no rows.

**2 · The guard had to move to `tests/unit/`.** `tests/test_canary_invariants.py` (3,726 lines, the only file exercising this code) is executed by **no** workflow — every gating job is `cd tests && pytest unit/`. A guard placed beside the code it guards would never have gone red.

**3 · Names come from each invariant module's docstring title, not the catalog.** Catalog ids are not registry ids: catalog `E-06` is the unimplemented #129 check while registry `E-06` is "no overdue `next_run_at`" (#1472, semantically `SCH-03`). `G-04`'s catalog title also promises "/ logs" coverage that doesn't exist. A catalog-sourced name would confidently mislabel a live alert — worse than the honest bare-id fallback. Both stale catalog cross-refs corrected.

## The render fallbacks are a security property, not debt

`_render_forensic`'s `return None` and `_render_message`'s count-only string are *why* an un-rendered invariant can't echo `observed_state` — E-04 and G-04 scrub at the **check**, reporting a reason code / pattern name only. A parity gate creates standing pressure toward the one-line generic `observed_state.items()` dumper that satisfies it forever, and that change is near-invisible in review because it *removes* a special case. Both fallbacks are kept, commented, and pinned by a negative test.

## Also in scope

- **`_mrkdwn_safe()` at the render boundary** (56 sites), collapsing two inconsistent `?`-defaulting idioms into one. Escapes Slack's documented `&`/`<`/`>` set so a name can't forge a live link or `@channel` mention. Injection safety is now local rather than transitive across ~8 sanitize-at-write paths — `retention_guard.py` already writes an `agent_name` that `sanitize_agent_name` could never produce.
- **A `NULL` `agent_name` no longer raises `TypeError`** out of `sorted()` into `canary_service`'s swallowing `except`, which dropped the alert entirely *while still advancing the green→red cursor*, so nothing retried. Swept all ten pre-existing sites, not only the five new branches — so this affects all 15 invariants, not just the 5 the issue names.
- **`requirements/infrastructure.md` §31** said registering an invariant leaves "the service and API surface unchanged" — after this change that produces a red CI. Corrected, along with the `architecture.md` alert-sink block.

## Changes

| File | What |
|---|---|
| `src/backend/services/canary_alerts.py` | 5 ids × 4 surfaces, `_mrkdwn_safe()`, `_format_duration()`, NULL sweep, fallback comments |
| `tests/unit/test_1880_canary_alert_parity.py` | AST bidirectional parity guard — the CI gate |
| `tests/unit/test_1880_canary_alert_rendering.py` | Behavioural: no-crash, not-fallback, Slack limits at scale, hostile-input, no-leak |
| `docs/memory/architecture.md` | Alert-sink contract + restricted-channel operational note |
| `docs/memory/requirements/infrastructure.md` | §31 registration contract |
| `docs/testing/orchestration-invariant-catalog.md` | Catalog-vs-registry id drift footnote |
| `docs/memory/learnings.md` | Durable entry (parallel surfaces; guard placement; heuristics) |
| `docs/security-reports/cso-diff-2026-07-30-*` | `/cso --diff` audit |

## Test plan

- [x] `pytest tests/unit/test_1880_canary_alert_{parity,rendering}.py` — **107 passed**
- [x] `pytest tests/test_canary_invariants.py` — **139 passed, 1 skipped** (no regression)
- [x] `pytest tests/unit -k 'canary or alert or slack'` — **424 passed, 1 skipped**
- [x] `python tests/lint_sys_modules.py` — no new violations
- [x] `pyflakes` clean on all changed files
- [x] **Guard mutation-verified** — six regression shapes each break it, source restored byte-identical:
      drop a name · drop the forensic branch · typo an id · stub a name to silence the gate ·
      make the fallback a generic state dump · remove the mrkdwn escaping (17 failures)
- [x] Rendered payload inspected by hand for G-04 / E-06 / E-03
- [x] Bounded at fleet scale: 5000 violations × 500 agents stays inside Slack's 150/3000/50-block limits

## Security

`/cso --diff` run: **no CRITICAL, no HIGH.** One MEDIUM (information disclosure) remediated in-diff via a documentation control — a rendered G-04 alert points at *where* a live credential sits, and channel membership behind `CANARY_SLACK_WEBHOOK_URL` isn't tied to the Trinity admin role, so `architecture.md` now requires a restricted channel. The central property — that a credential-detection alert can't itself leak the credential — was tested adversarially by an independent verifier and holds; it's now pinned by regression tests rather than resting on reviewer attention. Full report in `docs/security-reports/`.

No schema change, no migration, no new endpoint, no config, no dependency.

Fixes #1880


obasilakis (approved):
Approving.

Verified independently rather than taken on faith:

- **`observed_state` key parity with the checks.** The quiet failure mode here is a render branch reading a key the invariant never emits — it degrades to `?` and reads fine in review. All five match: E-03 (`status`/`started_at`), E-04 (`reason`), E-06 (`schedule_id`/`next_run_at`/`overdue_seconds`), G-03 (`skew_seconds`), G-04 (`matched_pattern`).
- **The un-CI'd-suite claim holds** — `backend-unit-test.yml:116` is `cd tests && python -m pytest unit/`, so `tests/test_canary_invariants.py` is collected by nothing and the guard's placement in `tests/unit/` is load-bearing. Ran it locally since CI won't: **139 passed, 1 skipped** (no regression from the 56 `_mrkdwn_safe` sites), plus **107 passed** on the two new files.
- **The swallow-then-advance path is real.** `canary_service.py:253` catches `Exception` around `emit_transition` and `_write_prev_cycle_at` runs unconditionally after, so a `TypeError` out of `sorted({None, …})` loses the alert *and* advances the cursor — and continuing-red never re-posts. Sweeping all 15 branches rather than only the 5 new ones is the right scope.
- **Rendered a live G-04 payload** with a NULL and a hostile agent name: `<!channel> pwn` → `&lt;!channel&gt; pwn`, NULL → `?`, no stutter, no secret value, header 56 chars.

Two nits, neither blocking, both pre-existing:

1. `_render_message`'s S-02 branch still uses `max(..., key=lambda v: v.observed_state.get("overbooked_by", 0))` while the new E-06/G-03 branches use `or 0`. Safe today (`s02_no_overbooking.py:69` writes `slot_count - cap`, always an int) — just the one asymmetry the sweep left.
2. The `canary_service` comment "Done AFTER notifications so a crash mid-emit doesn't advance the cursor" is true only for an exception escaping `run_cycle`; the inner `except` falsifies it for anything raised inside `emit_transition`. This PR removes the concrete trigger; the shape survives. Follow-up, not this PR.

The three out-of-AC extensions are each justified in the diff and hold up: four surfaces not two (the two the issue missed carry the per-row evidence), guard placement, and docstring-title-over-catalog naming. Keeping both render fallbacks state-free and pinning them with a negative test is the right answer to the pressure a parity gate creates.

Issue #40, opened 2026-02-24
Platform Simplification Mode


vybe:
Hide advanced features (Collaboration Dashboard, Process Engine) behind toggle or role. Simpler default experience.

PR #1899, opened 2026-07-30
feat(templates): the credential declaration standard — `credential_setup:` + two HARD-gate fixes (ent#128 PR-B)

AndriiPasternak31:
# ent#128 PR-B — the credential declaration standard

Closes **AC #1–4** of [abilityai/trinity-enterprise#128](https://github.com/abilityai/trinity-enterprise/issues/128). **AC #5 already shipped** in #1835 (PR-A).

Cross-repo `Fixes` does **not** fire — the issue is referenced, not closed. Close it manually at the release cut, and set `status-in-dev` by hand (the same-repo automation won't promote it).

**Baseline:** rebased onto `origin/dev` @ `c4c83f4a` (2026-08-02), which contains PR-A (`8d59b6c8`, #1835) — so this diff isolates PR-B cleanly.

Rebase resolved two doc conflicts by **conceding dev**: ent#263 renamed `feature-flows/templates-page.md` → `library-page.md` and rewrote it (477 → 168 lines), dropping the `required_credentials` blocks my annotations targeted — the three-way shape table survives in `template-processing.md:183-185`, so nothing was lost. Index + `learnings.md` entries merged in date order.

**2026-08-02 rebase.** Re-based onto `c4c83f4a` after `dev` moved 5 commits ahead. Two conflicts, both resolved by **keeping both sides**: `requirements/core-agent.md` §4.3 (two independent bullets) and `template_service.py` (dev's new `contained_template_dir` block vs this PR's `_build_local_template(…, *, is_bundled)` signature). The commit SHAs cited further down were rewritten by that rebase and have been updated here. The substantive consequence — dev landing #1900 — is written up in the CodeQL section below.

**Size:** 25 files, +3402/−63 — but that splits as **944 lines of production code**, 1995 of tests, 406 of docs, 57 of template fixtures. The single largest file in the diff is a test file.

---

## ✅ Decisions — all resolved by @vybe

[Ruling on ent#128.](https://github.com/Abilityai/trinity-enterprise/issues/128#issuecomment-5143599220) No vetoes; nothing outstanding.

| # | Call | Landed as |
|---|---|---|
| Schema sign-off | **Approved**, incl. tri-state `required` (`unknown` for legacy bare names) — that's what ent#137 propagates | Ships as-is |
| `config_files` | **Keep as shipped** — hardened + `deprecated` in the schema. Don't delete here | Removal follow-up → **#1930** |
| YAML hardening | **File in `trinity-enterprise`** — documents a working amplification attack against live installs; nothing public until the platform-wide caps ship | **trinity-enterprise#314** |
| `${VAR:-default}` | Public `type-bug`, **P1** — breaks the seeded first-run agent | **#1929** |
| OSS-core gating | **Confirmed** | This PR's placement |
| Split the fastapi cap? | **"One PR is fine, don't split"** | Rides here |

The path-traversal CodeQL surfaced (pre-existing, not introduced here) is **#1900** — **since fixed on `dev`** by #1935, and inherited by this branch through the 2026-08-02 rebase. Details in the CodeQL section below.

---

### Two HARD gates change verdicts — these were live false-positives

- **K-002** (HARD) — `c_t015` compared `${VAR}`s against the credentials block's **section names**, never the variable names. A template whose `${VAR}` happened to match a section name passed vacuously; a correctly-declared variable could fail.
- **K-001** (HARD) — `_env_example_vars`' uppercase-only filter could not see a documented **lowercase** variable, so it reported it as undocumented.

Neither needed a spec change: `docs/agent-validation-spec.md` already describes the intended behaviour ("*is listed in `template.yaml` credentials schema*"). The **code** was wrong, not the spec.

### ⚠️ The verdict diff is NOT monotone fail→pass

I want to be explicit because an earlier version of this write-up claimed "only fail→pass", and **that is measured false**. Across **54 fixtures × 66 static checks** (`origin/dev` @ `39f29c64` vs this branch) there are **60 transitions in three classes**:

| Count | Transition | What it is |
|---:|---|---|
| 27 | `fail → pass` | The fix — T-015/K-002 section-name bug, K-001 lowercase reader |
| 27 | `skipped(check_error) → fail` | `run_static` failing **closed**. All 27 on one fixture (`60k-deep-nesting`), whose `hard_count` goes **7 → 15**, i.e. **UP** |
| 6 | `pass → fail` | Vacuous passes becoming true positives |

**The claim I am defending is not "only fail→pass".** It is narrower and checkable:

1. **Zero escapes** from `run_static` on either tree.
2. **Zero dark checks** on the branch (nothing left `skipped`/`check_error`).
3. Every one of the **15 fixtures whose `hard_count` dropped** traces **line-by-line to an intended `fail→pass`** — **none** to a `fail→skipped`. That is the property that matters: no check went quiet to buy a green number.

**Correction to my own brief:** the 6 `pass→fail` split is **2 HARD + 4 SOFT**, not "4 HARD, 2 SOFT". Verified against `spec.py` (`K-002: hard`, `K-001: hard`, `T-015: soft`, `K-003: soft`) and cross-checked against the per-fixture hard/soft deltas:

- `var-is-section-env_file` and `var-is-section-mcp_servers` — **K-002 (HARD)** + **T-015 (SOFT)** each, `hard 8→9`, `soft 12→13`. These are the section-name bug **in the opposite direction**: a `${VAR}` that collides with a section name used to pass vacuously and now correctly fails. Same root cause as the 27 `fail→pass`.
- `env_example-generic-lowercase`, `env_example-lowercase-only-no-comments` — **K-003 (SOFT)**, `soft →+1`. K-001 now *sees* lowercase vars, so K-003 ("comments explain each variable") finally has them to evaluate and correctly reports the undocumented ones.

Evidence + rerunnable harness in `.plan/ent128b-transition-evidence/` (git-excluded — it does **not** travel with the branch, which is why the numbers are pasted here).

### Smaller behaviour changes

- **The "N credentials" badge starts showing non-zero.** It was always wrong at 0 — it read a top-level `required_credentials:` key that **no template defines**. Release-note it.
- Templates declaring **both** `mcp_servers:` and `credentials.mcp_servers` now report the former. This matches what `info.py` already did.
- Lowercase `.env.example` vars are now extracted. **No production caller** — `extract_agent_credentials` is dead code today (V4).
- **No DB change → Rule #9 (dual-track migration) does not apply.**
- `docker/base-image/**` changed (`info.py`), so this needs a **base-image rebuild** to reach agents.

---

### `11251b39` — the tolerant credential reader must not blow up or raise

Two holes in the "never raises, never amplifies" property:

**(a) `credential_shape_errors` was uncapped** while its sibling `normalize_credential_requirements` got the cap — and both feed the **same two surfaces** (the catalog's `credential_errors`, and the `"; ".join(errors)` that becomes the agent-creation **400 body**).

Measured on a 6,738-byte template with one 200-element anchor aliased across 200 servers (re-verified on this rebased branch, not copied forward):

| Tree | Errors | Joined bytes | Amplification |
|---|---:|---:|---:|
| `origin/dev` @ `39f29c64` | 0 | 0 | — |
| this branch, walk-bound removed | 40,000 | 3,635,998 (3.64 MB) | **539.6×** |
| this branch, as shipped | 101 | 8,973 | 1.3× |

`origin/dev` returns **0**, so the amplification is **this branch's own** — introduced and closed within the PR. The cap bounds the **walk**, not just the output; a cap applied after the loops would still have built and joined every string.

**(b) `source_trust not in _SOURCE_TRUST_LEVELS`** is a frozenset membership test, so an **unhashable** value raised `TypeError` **on the guard line itself**. Verified as shipped: `list`, `dict`, `set`, `str`, `None` all pass through and fall back to `github` with a warning.

Both regression tests were confirmed to fail with the fix reverted.

### `93daa1e5` — cap fastapi to prod's 0.115.x line

Out of scope for a credential PR, flagged rather than let ride in silently. **@vybe ruled: "One PR is fine, don't split"** — so it stays.

The case for keeping it: prod pins `fastapi==0.115.6` **exactly**, while the test env floated to `0.140.13` — ~25 minors ahead, i.e. the unit suite was validating against a different FastAPI than the one that runs. CI is green **only on a warm pip cache** keyed on `tests/requirements-test.txt`, so the next edit to that file *for any reason* busts the key, re-resolves, and breaks CI for everyone. Capping is the safe way to bust that cache — the change that invalidates the key is the one that makes re-resolution correct.

Note `<0.141` does **not** work (`0.140.13 < 0.141` is true). The symbol died in `0.140.7` — a **PATCH** release — so no minor-level bound is trustworthy.

---

## CodeQL — resolved, and what it turned up

The first push tripped **1 new HIGH `py/path-injection`** (alert 260). Disposition, since it is worth understanding rather than skimming:

The flagged line was **mine**, but it performed **no I/O** — it was a pure `==` comparison used to pick a trust label (`source_trust` selects `logger.warning` vs `logger.info` and nothing else). It still called `.resolve()` on a path built from a user-supplied template id, which is a genuine tainted-path sink, added to decide a log level.

**Fixed in `5e02fe53` by deleting the sink, not by suppressing the alert.** `is_bundled` is now a required kwarg from whoever knows the provenance: the catalog scan iterates the curated root (bundled by construction), and the by-id lookup decides from the id **string**. Measured against the old predicate across 9 ids: 7 identical, 2 divergent (`'../agent-templates/sage'`, `'a\b'`), both `True → False` — strictly more conservative, never over-granting `bundled`.

**What it turned up is the part worth your attention.** Tracing the taint to its source found a **pre-existing** traversal in `get_local_template` (byte-identical on `dev` at the time, not introduced by ent#128): `_local_templates_dir() / name` with no containment check, reachable by **any authenticated user** via `GET /api/templates/{id:path}`. Verified that `local:..` escaped the templates root.

That was **not fixed here** — filed as **#1900** rather than folded into a credential-declaration PR or dismissed. Nothing was dismissed.

**Update (2026-08-02): #1900 is now fixed on `dev`**, by #1935, which this branch is rebased onto. `get_local_template` routes `name` through `contained_template_dir()` — a name allowlist plus `resolve()` + `is_relative_to` — **before** this PR's label check runs. Two consequences worth knowing at review time:

- **`is_plain_segment` is now redundant** — provably `True` wherever it is reached, since the barrier above rejects every non-plain name first. It is deliberately **kept** as defence in depth: it decides a trust **label**, and `contained_template_dir` is a shared primitive the remote-template-registry work (trinity-enterprise#14) is expected to edit. A label that silently became `bundled` if that barrier were ever widened is the exact failure the `is_bundled` kwarg exists to prevent. It re-adds no tainted-path sink — it reads the id string, never the filesystem.
- **The in-code comment and the commit message were both rewritten.** The merged versions still claimed "deliberately NOT a traversal guard … `name` still reaches the filesystem exactly as it did before ent#128" — true when written, false after the rebase. Left alone, that is the kind of stale comment a later reader takes as a statement about current behaviour.

---

## Verification

Run on the 2026-08-02 rebased branch (`5e02fe53`), at the capped FastAPI:

- `tests/unit/test_ent128{a,b1,b2}*.py` → **209 passed** (b1 = 39, b2 = 131, a = 39 — `a` gained 6 cases when #1935 added to the same file)
- **Full `tests/unit` → 6410 passed, 16 skipped, 0 failed** (6m59s)
- Compatibility spec-sync tests (`test_compatibility_checks.py`) → 36 passed
- dev's own template tests, re-run on top of this branch — `test_1900_template_id_traversal`, `test_1759_local_template_not_found`, `test_1759_template_root_parity`, `test_local_templates_listing` → all green, i.e. the rebase preserved #1900's guarantees rather than merging over them
- Security scan on the diff: no keys/tokens, no real emails, no public IPs, no `.env`, no credential files. Two grep hits are intentional test fixtures: `https://google.com@evil.tld` (the `setup_url` userinfo-phishing validator's own test input) and `STRIPE_API_KEY: 'from the vault'` (the deliberate unambiguous placeholder).
- Packaging: `credential_charset.py` lives under `services/`, which `docker/backend/Dockerfile` copies wholesale — no COPY-list gap (#1033 class). No new `os.getenv()` → no compose/`.env.example` gap (#1056 class).

**Not run:** `/verify-local`. `docker/base-image/**` is touched, so a base-image rebuild + live agent exercise is the right gate before release.

---

## Honest framing

- **AC #1's literal wording is superseded.** Enrichment lands in a **sibling `credential_setup:` key** rather than by extending `credentials:` — because an old Trinity iterating an enriched `env_file` raises `TypeError: unhashable type: 'dict'`. Same outcome, better compatibility. (Posted as an issue comment so the AC isn't read as unmet at close.)
- `credentials:` is **frozen, names-only, forever**. `credential_setup:` **decorates** it and can never declare a variable `credentials:` does not — base-set-plus-overlay, joined **by name**, so the pair cannot drift.
- **The new tests fail on `origin/dev` and pass here** — the proof they encode real defects rather than being written to fit the implementation. Measured, not asserted: copying `test_ent128b1_compat_gates.py` onto a pristine `origin/dev` worktree gives **21 failed / 18 passed**; on this branch all **39 pass**. The failures are exactly the named defects (`test_k002_no_longer_passes_a_reference_to_a_section_name[env_file|mcp_servers|config_files]`, `test_a_raising_check_is_a_failure_not_a_skip`, `test_check_error_cannot_erase_a_hard_finding`, …).
- **AC #4 is a ratchet, not coverage.** The bundle has exactly **one** non-empty declaration (`test-codex`); the other three bundled templates declare `credentials: {}`. So the parity test guards ent#137's curated fleet going forward — it does not prove breadth today. Not overselling it.
- **Cross-reference enforcement is asymmetric, and that's fine.** Drift is **impossible** for bundled templates (the parity test) and **visible** for external ones (`credential_errors` + the compatibility check). Trinity does not gate a third-party repo's CI.
- `normalize_credential_requirements` **never raises and never mutates** — `_build_template` runs unfenced inside `get_all_templates()`, so a raise there empties the whole catalog. That is the property (a) above exists to keep literally true rather than true-by-call-site-audit.


AndriiPasternak31:
Filed the pre-existing traversal this PR's CodeQL alert surfaced (but did **not** introduce): #1900.

CodeQL alert 260 is resolved in `5c0f712f` by **removing the sink**, not suppressing it — the trust label now comes from the caller instead of a `.resolve()` on a user-supplied template id. Nothing was dismissed. Detail in the PR body.

AndriiPasternak31:
Un-drafted per @vybe's call on [ent#128](https://github.com/Abilityai/trinity-enterprise/issues/128#issuecomment-5143599220). Rebased onto `dev`; all 20 checks green.

**`/verify-local`: full PASS** (8/8 stages, agent stage included — `docker/base-image/agent_server/routers/info.py` is touched, so a backend-only run would have been blind to it):

| stage | |
|---|---|
| unit | 6199 passed / 0 failed |
| build + `import main` in the built backend image | pass |
| agent-build + `import agent_server` in the built base image | pass |
| boot + health | pass |
| **agent-exercise** | real agent created → started → `/health` healthy in 3s, `clone_status: ok`, correct network |
| integration | 70 passed (2 known-false-fails deselected from the registry) |

Two timings looked too good, so I checked rather than trusting the banner:
- The 2s base-image build was **not** a stale cache hit — `#35 COPY ./agent_server /app/agent_server` is `DONE`, **not `CACHED`** (layers #6–#34 are cached; the `agent_server` layer rebuilt because `info.py` changed, invalidating #36–#40). The image genuinely contains the change.
- The 3s integration stage ran real tests: `70 passed, 13 skipped, 2 deselected`.

**On the CI blip:** `pytest (head, seed 67890)` first came back red. It was **cancelled by `timeout-minutes: 25`**, not a test failure — `conclusion: cancelled`, zero failed steps, log progressing and passing to 79% with all dots and bursty runner timestamps. `regression diff` went red only because the cancelled job skipped its JUnit upload. Re-ran it: **8m54s, pass** — same code, same seed, 3× faster, so runner variance. This is the failure mode the workflow's own comment documents at the previous 15-min cap.

Worth flagging separately: the suite is now 6199 tests, a normal shard is ~10m, and the cap is 25m — a 2.5× slow runner lands exactly on the ceiling. Not this PR's doing, but that margin is thinning.

**Release-note requirement:** agents only pick up the `info.py` fix after a base-image rebuild.

vybe (approved):
/validate-pr: PASS.

- Base `dev` ✅ · 25 files ✅ · security scan clean (`STRIPE_API_KEY: 'from the vault'` and `google.com@evil.tld` are fixtures/spoof vectors, not credentials) ✅
- No new top-level backend module (`credential_charset.py` sits under `services/`, copied wholesale by the Dockerfile) ✅ · no new `os.getenv()` ✅
- Docs: 2 requirements files + architecture + 3 feature flows + a JSON schema artifact ✅ · 6 test files ✅

Three things stood out as genuinely good:

1. **`_did_not_pass` is a real correctness fix, not cleanup.** `_counts` only counted `status == "fail"`, so a raise inside a HARD check dropped `hard_count` and flipped `overall_status` to `compatible` — a broken agent reporting green. Counting `skipped`+`check_error` as a finding at the sink, *in addition to* `run_static` now returning FAIL, is the right two-layer shape (#1525): if any future path reintroduces the skip, the count still holds.

2. **The input caps are measured, not guessed.** The record cap is applied to the INPUT before the walk because `str()` on an aliased YAML node expands it (443 B → 52 MB measured) — a cap applied after the loop would bound nothing. The separate error cap exists because 400k malformed entries produced 1 capped record and 400,000 uncapped errors, i.e. a 35 MB response built out of the cap meant to prevent it.

3. **The fastapi pin is a CI-integrity fix worth more than its diff.** An unbounded `fastapi>=0.115.0` resolved 0.140.13 while `docker/backend/Dockerfile` pins `fastapi==0.115.6` exactly — the unit suite was validating ~25 minor versions ahead of production. It was green only on a warm pip cache keyed on this very file, so the next unrelated edit would have busted the key and broken CI for everyone. The note that `<0.141` would NOT have worked (the private symbol vanished in a PATCH, 0.140.6 → 0.140.7) is the part that makes tracking prod's line the durable fix rather than a bigger number. Verified 5861 passed / 0 failed at 0.115.14 vs 1 failed at 0.140.13.

This pairs with #1910 (CI on Python 3.11 while every image ships 3.13) — same class, and both worth landing.

⚠️ Cross-tracker: ent#128 needs `status-in-dev` by hand (AC #1–4 here; AC #5 shipped in #1835).

PR #1902, opened 2026-07-31
feat(auth): bulk portal-session revocation primitive + operator client panel

dolho:
Related to trinity-enterprise#281 (the enterprise half is Abilityai/trinity-enterprise#307 — **merge that after this**, it calls the primitive added here).

## What

The OSS half of an operator kill switch for signed-in portal clients: a bulk **revoke-all-sessions-for-email** primitive, plus the Sharing-tab panel that drives it.

A client who signs in through sharing holds a 12-hour portal session. Until now the only ways to end it were natural expiry and a backend restart (`SECRET_KEY` rotation) — which logs out *everyone*.

## The primitive

`dependencies.revoke_portal_sessions_for_email(email)` — edition-agnostic, exactly like the existing portal mint/decode: OSS owns the mechanism, the entitled module decides *when* to call it.

**A per-email cutoff, not a jti list.** `jti` is random per token and nothing indexes email → issued jtis, so "which tokens does this address hold?" would require a write-side index maintained at every mint. One timestamp per email instead:

- O(1) write, O(1) check
- self-expiring at the max session lifetime — a token older than that is expired anyway, so the key stops existing exactly when it stops mattering (same bounded-growth property as the #187 `jti` blacklist, no sweep)
- **covers every mint path by construction**, because they all go through `create_portal_session_token`. That is the failure mode a hand-maintained index has and this one structurally cannot.

Supporting changes:
- `create_portal_session_token` stamps an explicit `iat` so tokens can be dated. Set there, not in `create_access_token`, so no other token type's claim set changes.
- `decode_portal_session` rejects `iat <= cutoff`. `<=` not `<` is deliberate — for a kill switch, a token minted in the same second as the revoke must die.

## Two deliberate fail-direction choices

**A token with no `iat` is treated as revoked — fail CLOSED.** Only sessions minted before this shipped lack one (all expired within `PORTAL_SESSION_EXPIRE_HOURS` of the upgrade), and only for an email an operator actively revoked. Letting a token that cannot be shown to post-date the kill switch survive it is the worse failure.

**The Redis read stays fail-OPEN**, matching #187 and the platform posture. This is why revocation is not the whole feature: the durable half (a DB row) lives in the entitled module and keeps working with Redis down. `revoke_...` returns a bool so the caller can report honestly rather than claim a success the operator would act on.

## Frontend

`PortalClientsPanel.vue` on the Sharing tab — per-client **Log out / Block / Unblock** with current state. Entitlement-gated on `client_portal`, so an OSS build renders nothing.

Three honesty properties, since this is a security control an operator will trust:
- **Block is hidden for non-admins**, not offered and then 403'd (it is admin-only server-side, because its effect is platform-wide).
- **A failed revoke is reported as a failure** — "Could not end the session — their sign-in is still active" — not a green "signed out".
- **No live-session count is shown.** Portal sessions are stateless JWTs with no server-side store, so any number would be a guess. The panel shows `last_active` and when sessions were last cut off.

Copy states the scope plainly, because both actions are global-per-email while the operator is standing at one agent: *"Log out ends this person's live sessions everywhere… To remove someone from this agent only, unshare them above."*

## Verification

```
32 tests (in the enterprise PR, which exercises this primitive end to end): all pass
  — incl. revoke-kills-existing / does-not-kill-later, per-email scoping,
    the no-iat fail-closed case, and honest reporting with Redis down
SFC compile: PortalClientsPanel.vue OK, SharingPanel.vue OK
enterprise-docs-guard pattern over changed public docs + seam files: clean
```

The primitive's tests live in the enterprise PR because that is where the fixtures and the caller are; the OSS functions have no OSS caller until that module registers.

`vite build` in this worktree fails on a **pre-existing** missing `mermaid` dep in the borrowed `node_modules` (`AgentWorkspace.vue`), unrelated to this change — CI does a clean `npm ci`.

## Docs

`architecture.md` — one line in the Redis-keys list covering both `jwt:revoked:{jti}` and the new `portal:revoked_before:{email}`, describing the mechanism and why the cutoff form was chosen. Generic seam only; no enterprise module or table named (guard verified clean).

## Note on the submodule

This PR deliberately does **not** bump the `src/backend/enterprise` gitlink — the enterprise branch is unmerged, and the pointer is `update = none` for OSS clones.


dolho:
## `regression diff` red — pre-existing flake, not this change

`test_1771c_schedules_cas_edges::test_a13_exact_cutoff_row_is_not_expired[sqlite]` failed in **1 of 3** head seeds and **0 of 3** base seeds. A deterministic regression from a code change fails all three.

It is a wall-clock boundary test:

```python
now   = datetime.now(timezone.utc)
exact = (now - timedelta(hours=24)).strftime("%Y-%m-%dT%H:%M:%S") + ".000000Z"
...
assert ops.expire_stale_queued(24) == 1
assert status_of("a13-exact") == "queued"
```

`exact` is built from the wall clock at *setup*; `expire_stale_queued` computes its own threshold at *query* time. If the second ticks between the two, `exact` becomes strictly older than the threshold, gets expired, and the assertion flips. The test is pinning a genuine ≤1s tolerance band, but it samples the clock twice to do it.

Reproduced locally, on this branch, 150 consecutive runs of that test alone:

```
FAILURES: 5 / 150      (3.3%)
```

At 3.3% across 6 runs per CI invocation (3 seeds × 2 sides), roughly 1 in 5 PRs will show this regardless of content — which matches seeing it here once.

This PR's diff is `dependencies.py` (portal token mint/decode + a Redis cutoff helper), one Vue component, one Vue import line, and one line of `architecture.md`. There is no path from any of that to schedule-expiry SQL.

Filed as #1909 and re-ran the failed job.

vybe (approved):
Validated via /validate-pr: focused 4-file diff, architecture.md updated (Redis keys), no secrets, all 23 checks green. Primitive tests arrive with the enterprise caller PR (ent#307), merging next in sequence.

Issue #224, opened 2026-03-30
feat: Agent notifications to Slack with Ops integration (SLACK-NOTIFY)


pavshulin:
## Summary

Route agent notifications and operational events to Slack channels. This expands beyond scheduled task results to cover all notification types: task completions/failures, operator queue items, agent alerts, and custom agent notifications.

## Motivation

Currently agent notifications only surface in the Trinity UI. Operators and stakeholders need to see critical agent events in Slack where they already work — especially Ops queue items (approvals, questions) that require timely human response.

### Notification Types Routed to Slack
1. **Scheduled task results** — completion, failure, with summary (original #224 scope)
2. **Operator queue items** — approvals, questions, alerts from agents needing human input
3. **Business task validation results** — pass/fail from validation sessions (see related issue)
4. **Agent alerts** — cost thresholds, health issues, autonomy state changes
5. **Custom agent notifications** — via `send_notification` MCP tool

### Configuration
- Per-agent Slack channel binding (already exists via SLACK-002)
- Notification type filtering — choose which types route to Slack per agent or globally
- Ops queue items include action buttons (approve/reject) where Slack supports it, or a link to the Trinity Ops UI

### Message Format
- Use Slack Block Kit for rich, structured notifications
- Include agent name, timestamp, status badge, and relevant details
- Link back to Trinity UI for full context

### Integration with Ops
- When an agent posts to the operator queue, the notification is also sent to the configured Slack channel
- Operator can respond via Slack thread (if supported) or click through to Trinity
- Acknowledgment syncs back to operator queue status

## Acceptance Criteria
- [ ] Scheduled task completions/failures post to agent's Slack channel
- [ ] Operator queue items (approvals, questions) post to Slack
- [ ] Notification type filtering is configurable per agent
- [ ] Messages use Block Kit formatting with status badges
- [ ] Links back to Trinity UI are included
- [ ] Agent name displays correctly in Slack notifications

vybe:
Closing as outdated/superseded. The instance-level, easy-integration slice of this — agent OPS notifications routed to a single configured Slack channel via an `OPS_SLACK_WEBHOOK_URL` webhook (mirroring the canary pattern) — is now tracked in #1100. The broader per-agent SLACK-002 channel binding + inbound ack-sync scope here is deferred; if it's revived later it should be reopened as a focused phase-2 issue rather than this catch-all. See also #900 (incident-webhook decision).

PR #1905, opened 2026-07-31
feat(telegram): in-progress status indicator — reaction ack + elapsed-time placeholder (ent#264)

vybe:
Fixes abilityai/trinity-enterprise#264

## Summary

Upgrades the existing `indicate_processing`/`indicate_done` seam into a full in-progress indicator for Telegram: a 👀 reaction ack at dispatch, plus a router-owned elapsed-time driver that ticks a new default-no-op `indicate_progress` hook on the `ChannelAdapter` ABC (Slack/WhatsApp inherit the no-ops unchanged; Slack's ent#228 later implements only the hook). Per-turn state lives on `NormalizedMessage.metadata` — the adapter instance holds nothing. Every terminal path cancels-and-awaits the driver before resolving. A per-binding toggle ships via dual-track migration (SQLite + Alembic `0031`), surfaced in the Telegram channel panel.

## Behavior-change flags

- **Default-ON for existing bindings** — the indicator appears fleet-wide after deploy; the UI toggle is the opt-out (`server_default="1"` populates existing rows at migration time).
- **Access hardening**: `GET /api/agents/{name}/telegram` moved from bare `get_current_user` → `AuthorizedAgentByName` (closes a webhook-secret disclosure — the response's `webhook_url` embeds the webhook secret; arbitrary authenticated readers now get a uniform 404).
- **Review fix**: `GET /api/agents/{name}/telegram/groups` likewise hardened (cross-tenant group-data disclosure — group chat ids/titles/welcome text were readable by any authenticated user).
- **Group ack gate** = @mention/reply OR `trigger_mode == "all"` (observe excluded) — orchestrator-amended after two reviewers flagged the zero-feedback gap.
- **No success-✅** — disallowed for bots; the 👀 reaction is cleared at terminal instead.
- 30s threshold / 60s cadence as class constants (`progress_threshold_seconds` / `progress_interval_seconds`).
- **Documented residuals**: a backend restart strands an in-flight indicator; a first send slower than 10s abandons its placeholder (same class).

## Migration note

Alembic drafted as `0031_telegram_progress_indicator` (down_revision `0030_slack_channel_allow_proactive`). Sibling PR (ent#265) also drafts an `0031` — **whichever merges second renumbers at rebase; never two heads** (#1763 lesson). SQLite track verified applied on a fresh boot (column default `'1'`).

## Test evidence

- 62-test dedicated suite (`tests/unit/test_ent264_telegram_progress_indicator.py`) + 124 neighboring greens.
[...]

PR #1906, opened 2026-07-31
feat(telegram): long-running task reports back to the originating chat on completion (ent#265)

vybe:
Fixes abilityai/trinity-enterprise#265

## Summary

Opens the shipped Slack #224 completion-report seam to **Telegram** — a Telegram-triggered long/delegated task posts its terminal (success/failure) back to the originating chat on completion:

- Persists the deliverable bot identity (`schedule_executions.source_channel_agent`, written at row creation, **transitive through delegation**) so consent + bot token always evaluate against the bot the user actually addressed.
- Telegram delivery leg with **DM consent-by-construction** and a group `allow_proactive` toggle (`telegram_group_configs.allow_proactive`, opt-out "Completion reports" switch in the Telegram panel).
- **Two channel-agnostic fixes that also repair the shipped Slack leg** (reviewers: these are NOT Telegram-only):
  - **D0** — the delegation-inheritance wire was severed: `/task` pre-created rows without channel columns, so delegated reports were latent dead code **on every channel**. Inherited channel context is now persisted on the child row at `/task` row creation.
  - The `apply_result` **failure branch never spawned a report** — agent-reported failures went silent. It now reports on both terminals.
- Plus a **security fix from review**: the provenance guard was keyed on the spoofable `X-Source-Agent` header — now **principal-keyed** (an agent key must BE the parent's executing agent; a human caller must own the agent; a connector key never inherits).

## Behavior / taste flags for reviewers

- **Group consent default = ALLOW** for existing AND new groups (diverges from ent#223's new-deny split; the flip is one default + one line in `get_or_create_group_config`). Three reviewers held three postures — operator's call.
- The proactive group-send endpoint stays **ungated** on the new flag (default-1 makes a future gate a no-op for untouched groups).
- Fan-out = **one report per child** (inherited Slack semantics).
- Failure terminals **always report** (AC-mandated).
- `effect_guard` dedup keyed on the resolved destination; at-most-once bias mirrored from Slack.

## Migration note

ONE Alembic revision **`0031_channel_report_back`** (both columns, `down_revision = 0030`). Sibling PR (ent#264) also drafts an 0031 — **second-merger renumbers; never two heads** (#1763). SQLite track runtime-proven on a fresh boot (both columns + migration row present).

## Test evidence

- 41-test dedicated suite (`tests/unit/test_265_telegram_completion_report.py`); 99 across the 4-suite gate incl. 3 spoof-guard regressions proven red pre-fix.
- 107 blast-radius + 49 adjacent greens.
- `verify-local --skip-agent` PASS (unit 5771 greens, 1 pre-existing env red proven on base; integration 70 passed).
- PG apply defers to the **pg-migrations** CI job — confirm green before merge.

- Gate Telegram proactive group sends on `allow_proactive` (ent#223 coherence).
- Unified adapter outbound-delivery seam.
- `list_telegram_groups` ownership check + MCP `list_channel_groups` `agent_name` passthrough (pre-existing IDOR, payload widened by one field here).
- UTF-16 vs Python char cap headroom (~4000) for astral-emoji reports.


dolho:
## /review Report

**Branch**: `feature/ent265-telegram-report-back` → `dev` (merge-base `5b289996`)
**Files**: 25 (+1858/−84) — new `channel_completion_report` service + execution-path wiring + migration
**Scope**: CLEAN — a task that inherited channel context reports its terminal back to the originating chat/thread, per ent#265.
**CI**: green.

No critical findings. This is the strongest PR of the ent#260–265 set — it handles every trust-boundary concern a "post task output into an external chat" feature has, and I verified each claim against the code rather than the docstring.

### Verified safe (with evidence)

- **The target chat is not attacker-controllable.** Delivery goes to the persisted `schedule_executions.source_channel/_chat_id/_thread` (ent#117) of the *originating* turn — never a request parameter.
- **Content is credential-sanitized, and the ordering is correct.** `_summarize` runs `sanitize_text` **before** truncation, over a **2× window** (`_MAX_REPORT_CHARS * 2`), then truncates the cleaned text. That's the subtle-right choice: sanitizing after truncation could pass a secret split across the cut, and cutting first could leave an unsanitized fragment. The comment explains it; the code does it.
- **At-most-once is real, not just documented.** `async with effect_guard(...)` is actually invoked (`channel_completion_report.py:334`), with `EffectInProgressError` caught (`:356`) — so a re-delivered terminal can't double-post.
- **Fires on the CAS-won terminal only**, both success and the failure path (the D3/AC#2 gap you closed), fire-and-forget + fail-open in `task_execution_service` — so a replay/late #1083 callback can't re-report, and a report failure never breaks the turn.
- **No double-posting of normal channel replies.** Reports only when `triggered_by` is NOT a channel trigger (the execution *inherited* its context, i.e. a delegated/background terminal) — a synchronous channel turn already replies inline and is skipped.
- **Proactive consent is gated per channel.** `_channel_allows_proactive` enforces Slack channel-binding consent and Telegram group config, and Slack suppresses a send to an unbound channel rather than firing one destined to fail. Delivery uses `binding_agent = source_channel_agent or agent_name` — the bot that owns the binding, correct for delegated/fan-out.
- **Cascade ref is registered with the right policy.** `AgentRef("schedule_executions", "source_channel_agent", Policy.KEEP)` — KEEP mirrors the execution-history row (which isn't cascade-deleted with the agent), so no orphan and no wrong-tenant reuse.
- **New toggle endpoint** in `routers/telegram.py` is `OwnedAgentByName` + `reject_agent_principal` (owner + human-only).

### Informational

**[I1] `0031` Alembic head collides with #1905 and #1628** (Confidence: 9/10)
`0031_channel_report_back (down_revision=0030)` shares its parent with `0031_telegram_progress_indicator` (#1905) and `0031_agent_ownership_a2a_exposed` (#1628). The second of the three to merge to `dev` creates two Alembic heads and breaks `alembic upgrade head` (the `pg-migrations` CI job). Whichever lands after the first must re-chain onto the prior `0031` and renumber. Coordination hazard, not a defect against `dev` today. (Same note is on #1905.)

### Clean categories
SQL — new column via parameterless `ADD COLUMN`, dual-track (Alembic `0031` + `migrations.py`); no interpolation. Auth — no new caller-facing trigger surface; the send is a backend fire-and-forget off a terminal it already owns. Concurrency — effect_guard + CAS-won gating cover the #1083 re-delivery race. Error handling — every delivery path is fail-open with a logged suppression reason.

**Critical: 0 · Informational: 1 · Scope: clean.** LGTM on the code; coordinate the `0031` re-chain (I1) at merge.

dolho (approved):

PR #1903, opened 2026-07-31
feat(dashboard): List view replaces the Agents page (timeline · grid · list) (ent#260)

vybe:
Fixes abilityai/trinity-enterprise#260

## Summary

Retires the standalone **Agents page** and delivers its row list as a third **List** mode on the Dashboard (Timeline · Grid · **List**), through the existing `VIEW_MODES` machinery — persistence and unknown-mode degrade come for free.

- **AgentListPanel extraction**: `views/Agents.vue` (1253 lines) deleted; its row list extracted into props-driven `components/AgentListPanel.vue`, mounted by the dashboard chassis. Shared sort comparator extracted to `utils/agentSort.js`.
- **`/agents` → `/?view=list` redirect**: query-preserving (`?onboarding=1` etc. pass through), one-shot and **non-persisting** — the `view` param applies via `setViewMode({persist:false})` and is stripped after apply, so a deep link never rewrites the user's saved mode.
- **Both N+1 mount loops deleted**: rows read `tags` / `read_only_enabled` off the fleet payload — zero per-row HTTP on mount, and read-only state now renders payload-truthfully on stopped agents.
- **NavBar consolidation**: Agents entry removed; Dashboard inherits the agent-detail active-highlight.
- **28-item parity disposition**: every behavior of the old page dispositioned (carried, migrated to chassis controls, or intentionally dropped) — recorded in `docs/memory/feature-flows/dashboard-list-view.md` and requirements §9.9.

Zero backend changes; no new endpoints.

## Flags for the reviewer (auto-decided taste calls)

- `?view=` deep-links do **not** persist the mode (one-shot apply, param stripped).
- Tag + owner filters migrated to the **chassis** filter controls (Clear-all clears both layers).
- System-agent visibility is now **backend-authoritative** (no frontend allowlist).
- Post-delete navigations go to bare `/` (previously `/agents`).
- List adopts the grid's **system-row Run-toggle guard**.
- **NEW** filter localStorage keys (`trinity-dashboard-list-filter-*`) — old Agents-page filters don't resurrect.
- Create Agent is a chassis **header button**, all modes (icon-only ≤ md breakpoint).
- Review fix: **gated fetch write-through** — a quick-tag-filtered fetch no longer clobbers the full-fleet store.

## Test evidence

- `verify-local --skip-agent` **PASS**: unit 5726 green; the 1 red is a pre-existing fastapi-0.141.1 environment failure, proven present on a pristine dev base; integration 70 passed.
[...]

PR #1907, opened 2026-07-31
feat(dashboard): hotkey type-to-filter across timeline · grid · list (ent#261)

vybe:
Fixes abilityai/trinity-enterprise#261

> **⚠️ STACKED PR** — Stacked on #1903 (ent#260). Retarget base to `dev` after #1903 merges; **do not merge before it.**

## Summary

`/` hotkey type-to-filter across all three dashboard modes (Timeline · Grid · List):

- **Hotkey**: document-level `/` keydown with a fresh input-focus guard (editable targets, open modals, chords, IME composition, key-repeat, `defaultPrevented`) + Firefox quick-find `preventDefault`.
- **Predicate**: layered inside the store-level seam — `visibleAgents` splits into pre-query `ownerFilteredAgents` (tag ∘ owner; feeds ALL `convertAgentsToNodes` call sites, incl. the 30s poll that previously rebuilt from the raw list) ∘ the query layer (case-insensitive substring over slug + display label via `agentDisplayName`, #1642).
- **Floating pill** with live "X of Y match" count, Esc hint, and clickable header `kbd` toggle.
- **Chassis-level query-empty overlay** that keeps panes MOUNTED underneath; the onboarding CTA branches are `&& !filterActive`-guarded so the CTA is unreachable under a query.
- **Timeline joins** via ReplayTimeline's `:agents` prop switch onto the same seam.
- **Nothing persisted** — reload/unmount always starts unfiltered. **Zero backend.**

## Release note (verbatim)

> The dashboard timeline now honors the persisted owner filter (previously grid/list only) — users carrying a stale owner filter will see timeline rows narrow on upgrade day.

## Flags

- **Timeline-honors-owner-filter side effect** — one-line reversible (`:agents` prop switch).
- **Floating pill vs header input** — pill chosen; visible whenever a query is open OR applied (an applied-but-hidden filter is the dishonest state).
- **Doc-level Esc backstop** scoped to non-editable targets (Esc inside the list panel's search box or any input/textarea/contenteditable belongs to that control).
- **Enter = blur-and-keep** (filter stays applied, focus released).
- **List-mode panel filter COEXISTS with the chassis query** — strategy voice preferred supersede; overruled per compose-honestly brief (N/M badge suppressed while the chassis query is active so two disagreeing denominators never render simultaneously).
- **Review fixes folded in**: CreateAgentModal z-10 → z-50 (house modal tier); dual empty-state CTA gate in AgentListPanel.

## Test evidence

- `verify-local --skip-agent` **PASS** on the full 15-commit stack (unit: 5726 greens + 1 pre-existing env red proven on base; integration: 70 passed ×2).
- Bundle-level proof of all filter affordances from the built prod image.
- 8-test e2e spec (`src/frontend/e2e/dashboard-type-filter.spec.js`) executes in CI via the `ui` label.

Manual live-stack checks bundle: label-match, owner+query composition, dark mode, Firefox quick-find, de-DE Shift+7, IME, tag-dropdown Esc layering.


dolho:
## /review Report

**Branch**: `feature/ent261-dashboard-type-filter` → `feature/ent260-dashboard-list-tab` (stacked on #1903; merge-base `980206ec`)
**Files**: 13 (+540/−40) — frontend only
**Scope**: CLEAN — `/`-hotkey type-to-filter across timeline · grid · list, per ent#261.
**CI**: green.

No critical findings. The keyboard handling is textbook — it closes every classic type-to-filter footgun, which is the entire risk surface of this kind of feature.

### Verified safe (with evidence)

- **The global keydown listener is properly paired** — `document.addEventListener('keydown', handleDashboardKeydown)` in `onMounted` (`Dashboard.vue:830`), `removeEventListener` in `onUnmounted` (`:839`). No leaked listener across route changes.
- **The `/` handler never steals keystrokes it shouldn't:**
  - Guard 0 — respects `e.defaultPrevented` and ignores `e.repeat` (key-hold).
  - Guard 2 — bails on `ctrl/meta/alt` (don't shadow OS/browser chords).
  - Guard 3 — bails on `e.isComposing` (IME).
  - Guard 4 — bails on `INPUT`/`TEXTAREA`/`SELECT`/`isContentEditable` targets (typing `/` in a form field doesn't open the pill).
  - Guard 5 — bails when a modal is open (onboarding/editor/create).
  - `e.preventDefault()` to block Firefox quick-find.
- **Escape is layered, not destructive.** It's a backstop only while the filter exists, never races a modal's own Esc, defers to an open tag dropdown, and explicitly does **not** clear the chassis filter when Esc is pressed inside *another* editable field — avoiding cross-layer destruction. The pill input's own `@keydown.esc.stop.prevent` means it never reaches this backstop.
- **Substring match, not regex.** There is no `new RegExp`/`.test()` built from `filterQuery` anywhere in the diff — so a query containing `(`, `[`, `\` can't throw or ReDoS. The query is never persisted (`filterQuery` is an accelerator, cleared on unmount).
- **No XSS.** The query renders only through `{{ filterQueryTrimmed }}` (auto-escaped, `Dashboard.vue:386`); no `v-html` on it.
- **Layout-aware.** The `/` check keys on `e.key === '/'` (layout-produced), so `Shift+7` on de-DE still triggers it — a nice correctness detail.

### Informational

**[I1] Stacked PR — merges after #1903** (Confidence: 10/10)
Base is `feature/ent260-dashboard-list-tab`, not `dev`, so this lands only once #1903 does. Inherent to the stack, not a defect — flagging so the merge order is explicit. The diff reviewed here is the ent#261 delta against #1903's tip, not against `dev`.

### Clean categories
SQL / auth / credentials — N/A (frontend only). Reactive state — one mutation path (input writes through the store setter); `filterOpen` is deliberately Dashboard-local so a stale open-empty pill can't resurrect on remount. Focus — `nextTick(() => filterInputRef.value?.focus())` handles both first-render and reopen.

**Critical: 0 · Informational: 1 (stacking) · Scope: clean.** LGTM.

dolho (approved):

PR #1904, opened 2026-07-31
feat(library): rename Templates page to Library — one surface for agent templates + skills (ent#263)

vybe:
Fixes abilityai/trinity-enterprise#263

## Summary

Two-commit pure-`git mv` rename (Templates.vue → Library.vue, then content edits) + a `/library` route with a query+hash-preserving `/templates` redirect. Adds a fleet-level **skills browse** section over the existing `/api/skills` reads — isolated in `stores/skillsLibrary.js`, zero overlap with open PR #1901's files — plus dormant slots for #1901's `source`/`shadowed` fields. e2e anchors + docs updated; stale-doc cleanup (dead `env-template`/`refresh` endpoints, dead skills-management links).

## Flags

- **AC#3 fleet-assignment counts deferred** — fully specified follow-up (plan §6.2); both plan second-voices objected to bundling it here, so the operator owns the renegotiation. The ~40-line severable `GET /api/skills/assignments` endpoint is viable now.
- **AC#4 page-identity interpretation** — rename covers page identity (nav label, route, page title, headings say Library); "template" survives as the asset noun for the agent-template kind. See the [issue comment](https://github.com/Abilityai/trinity-enterprise/issues/263#issuecomment-5141094160).
- **Filter pills dropped** in favor of stacked sections (Agent Templates / Skills).
- **Security hardening in-branch**: `stripUserinfo` rewritten after 5 proven credential-leak shapes (26-case adversarial suite); shared with SkillsPanel.

## Test evidence

- `verify-local --skip-agent` **PASS**
- Bundle-level proof from the built prod image: library route present, redirect works with query+hash, NavBar clean, 5 asset-noun-only 'Templates' hits
- Live `tests/test_templates.py`: **5 passed**
- Unit suite: **5726 greens** (1 pre-existing env red, proven on base)

## Merge-order notes

- Prefers landing **after ent#260**: its Templates.vue:304 repoint becomes a content merge inside Library.vue thanks to the pure-rename commit; second-lander re-points.
- `requirements/skills.md` hunks verified **disjoint** from open PR #1901 (first-merged wins; worst case a context-line rebase).

## Follow-ups (listed, NOT filed)

- Per-source display post-#1901
- AC#3 `GET /api/skills/assignments` (plan §6.2, unblocked)
- Invariant-#7 raw-axios debt
- Unified LibraryItem direction (ent#14/#108)
- `/operating-room` redirect hash gap
[...]

Issue #1430, opened 2026-07-03
Design system v2: shared UI primitives, raw-color ratchet, and written style guide


vybe:
## Summary

Grow the current color-token layer into a proper frontend design system: shared UI primitives (button, input, badge, card, modal shell), an adoption ratchet so raw Tailwind palette usage converges to zero, and a written style-guide doc contributors can actually find.

## Context

Today "the design system" is exactly one layer: semantic color tokens in `src/frontend/tailwind.config.js` (`status-*`, `state-*`, `brand-*`, `accent-*`, `action-*`, from #67), guarded by `src/frontend/scripts/check-design-tokens.mjs` (`npm run check:tokens`, wired into `.github/workflows/frontend-build.yml`). That works, but an audit (2026-07-03) found the gaps:

- **The validator doesn't enforce adoption.** It only checks token→palette aliasing and that used token references resolve. Raw palette classes are legal: ~3,200 semantic-token usages across 99 files vs ~685 raw-palette usages across 48 files (worst: `views/enterprise/Audit.vue`, `components/NavBar.vue`, `components/process/TemplateSelector.vue`, `views/Login.vue`, `components/settings/TwoFactorPanel.vue`, `views/Settings.vue`, `components/GitConflictModal.vue`, the operator/executions panels).
- **Tokens only — no component primitives.** Among ~86 components there is no BaseButton/Input/Badge/Card; buttons and form controls are hand-rolled Tailwind class strings per component. Visual consistency relies on copy-convention, which is why "align X to the design system" keeps recurring as manual work (e.g. the Brain-tab alignment pass).
- **No written definition.** The taxonomy lives in a `tailwind.config.js` comment block; nothing in `docs/` describes it. No ESLint/stylelint exists in the frontend to flag drift at write time.

## Acceptance Criteria

- [ ] **Design reference pass first**: before locking primitive designs, benchmark against well-designed products using the design-reference MCP (Refero — screen/style search) so the primitives look polished, not just consistent; capture chosen references in the style-guide doc
- [ ] Shared primitives exist and are used by at least the core surfaces: `BaseButton` (primary/secondary/danger/ghost × sm/md), `BaseInput`/`BaseSelect`/`BaseToggle`, `BaseBadge` (status-token driven), `BaseCard`, and a modal shell (`ConfirmDialog` refactored onto it) — all dark-mode aware, keyboard/focus-ring correct
[...]

PR #1911, opened 2026-07-31
feat(systems): install a system from a manifest via the UI (trinity-enterprise#126)

AndriiPasternak31:
Refs trinity-enterprise#126

`POST /api/systems/deploy` has shipped for months and was reachable only by
`curl` or the MCP `deploy_system` tool. This exposes it: pick a bundled
manifest, upload a file, or paste YAML → preview → deploy.

> Cross-tracker issue — GitHub auto-closes same-repo only, so ent#126 is closed
> manually at release (working agreement per #1760). The private issue also
> needs `status-in-dev` set by hand after merge.

## Where it lives — @vybe, this is the bit to review

A third stacked `<section id="systems">` on the **Library** page, between Agent
Templates and Skills.

It was originally built as an `?tab=agents|systems` strip on `Templates.vue`.
Your ent#263 (#1904) landed while this sat in draft, renamed that page to Library
and recorded that tab/filter-pill navigation was *deliberately* rejected for it.
Rather than reintroduce a competing model I've conformed to yours — but you own
this page now, so two calls are yours to make:

1. **Does a fleet installer belong on a browse page at all?** Your subtitle says
   "installable assets for your fleet", and a manifest is one — the bundled-manifest
   catalog is a card grid structurally like your template cards. But installing is
   an action, not browsing. Its own route is a reasonable alternative; ent#126's
   premise was only "no 7th NavBar entry".
2. **Ordering.** I put Systems *between* Templates and Skills rather than appending
   it, because Systems and Agent Templates both install agents while Skills
   configures agents that already exist. That reorders your page — say the word and
   it goes last instead.

Both are one-file changes. The section is hidden outright below `creator` (mirroring
`POST /api/systems/deploy`), with its jump anchor gated alongside so the nav can't
point at an unrendered section. `/templates` still redirects with query **and** hash
preserved, which the e2e now pins.

Components in `components/systems/` over a new domain-scoped `stores/systems.js`
(Invariant #6 — deliberately *not* bolted onto `systemViews.js`: a **System** is a
manifest-deployed name-prefix group, a **System View** is a saved tag filter).

## Backend

**Read-only bundled-manifest catalog** — `GET /api/systems/manifests` and
`/manifests/{id}`, both `require_role("creator")` mirroring `POST /deploy`.
Declared above the parameterized routes for **two** Invariant-#4 collisions
(`/{system_name}` *and* `/{system_name}/manifest`). Listing is fail-soft per
file — one bad manifest is listed `valid: false`, never a 500 for the catalog.
Path confinement is layered, and each layer is there for a distinct reason: the
character allowlist does **not** reject `..` (`.` is inside its class), and the
symlink check must run *before* `resolve()`, which has already erased the link.
Reads open once with `O_NOFOLLOW` and `fstat` that descriptor — `config/manifests`
is a host bind mount, so stat-then-read has a real swap window.

**Extended dry-run preview** — gains `permission_edges`, `schedules_preview` and
`system_view_requested`, computed by **pure resolvers shared with the writers**
(`configure_permissions`/`create_schedules` now loop over the same functions), so
the preview cannot drift from what deploy does **on the fields it models** (see the ent#89 note below). Only the backend knows the
resolved `_N`-suffixed names, so this is not derivable client-side. Behaviour
preservation is pinned by characterization tests **captured green before** the
refactor — a resolver-vs-refactored-writer parity test would be tautological, and
the ent#125 suite mocks both writers.

**Resource preflight** — `_preflight_template` now validates **merged** resources
through the create path's own `normalize_cpu`/`normalize_memory`. It previously
checked template shape only: a shipped bundled manifest carried `cpu: 1.0`,
previewed `valid`, then failed 100% of its agents at create. That manifest — a
broken duplicate of the live seed, same `acme` system name — is deleted, and a
table-driven test over the real directory guards the class.

## Honest reporting

The store switches on `status`, never the HTTP code: `partial`/`invalid` are
**200**, and `failed` is **500 with the full report as the body**, which a naive
`catch` discards along with the `failed[]` list. The result reads *"agents
created"*, not *"success"* — `status` describes agent creation only, so a fleet
where every schedule failed and nothing started still reports `deployed`;
`warnings[]` therefore renders prominently. A manifest that sets `prompt:`
(replacing the platform-wide prompt for **every** agent) or carries enabled
schedules is gated behind an explicit acknowledgement, not a banner. A timeout
reports **"outcome unknown — may still be running"** and offers no retry:
cancelling the request does not cancel the serial server-side deploy, and
re-running duplicates every agent that succeeded.

Plain `<textarea>`. The orphaned monaco `YamlEditor.vue` stays unrevived — prod
CSP is `script-src 'self'` with no `unsafe-eval`/`worker-src` while the dev CSP
allows it, so `npm run dev` cannot prove prod.

## Review fixes in this branch

- A hygiene check became the unnamed 500 it existed to prevent: PyYAML is YAML
  **1.1**, so bare `on`/`off`/`yes`/`no` keys parse as booleans and `sorted`
  raised `TypeError` — on manifests that deployed fine before the check existed.
- `max_length` counts characters, the limit is bytes — now byte-exact, so the
  request cap, `st.st_size` and the UI's `file.size` cannot disagree.
- Catalog `reason`s were raw `str(e)`; PyYAML parse errors echo the offending
  source line. Routed through the deploy report's own sanitizing exit.
- The stale-preview hint was unreachable (an edit told you to "Preview first"),
  and `?tags=` silently no-opped for anyone with a view restored from
  localStorage — a dead end for AC #5.

## Merged with #1884

trinity#1884 landed on `dev` mid-review and overlaps this branch directly — it
moves the manifest size cap into `parse_manifest`, next to the alias-budget and
duplicate-key guards this branch's own comment named as the outstanding
follow-up. Resolved in its favour: ent#126's request-model byte validator is
dropped (#1884's cap is already byte-exact), and `MANIFEST_MAX_BYTES` keeps a
**single** definition in `models.py` — the merge had left `system_service` both
importing it and redefining it locally, where the local silently shadowed the
import. Oversize is now a 400 parse failure rather than a 422; the 422
list-detail test was re-pointed at a live producer rather than retargeted at the
400, which would have dropped the store's only `Array.isArray(detail)` coverage
while staying green.

## Merged with ent#89 and ent#314

A second round: `dev` moved 36 commits under this branch. Four conflicts, none
resolvable whole-file.

**ent#89 (#1946)** added a read-then-skip to `create_schedules` so a manifest
schedule does not duplicate one the *template* already materialized at agent
creation — the same block where this branch swapped the inline `ScheduleCreate`
for the shared `_build_schedule_create`. Both are kept: skip first, then the
shared builder. Verified load-bearing by reverting the skip on the merged tree
(3 of `test_ent89_manifest_no_duplicate.py` go red, green again on restore),
because resolving this hunk either way would have looked clean and silently
dropped one side.

**That skip is a deploy-side read `resolve_schedule_previews` cannot make.** A
manifest schedule whose name collides with a template-declared one previews as
"will be created" and is then skipped. The **armed end state is unchanged** — the
template creates a schedule of that name — so what diverges is provenance and the
report's `schedules_created` count, not whether the fleet runs that schedule.
Documented at the resolver and in `architecture.md` rather than half-fixed;
closing it properly means resolving templates inside the preview.

**ent#314 (#1961)** made `utils/safe_yaml.py` the single YAML implementation and
re-added a module-local `MANIFEST_MAX_BYTES` where this branch imports the single
definition from `models.py`. Kept the import, dropped the local — which would have
silently shadowed it, exactly the way the #1884 merge did. AST-verified: zero
module-level assignments, one import. The bundled-catalog readers hand raw bytes
to `parse_manifest`, so they inherit the hardened loader; ent#314's whole-backend
AST consolidation guard passes on the merged tree (54 hardening tests green).

## Verification

- `tests/unit/test_ent126_*.py` + `test_ent89_*.py` — **287 passed, 1 skipped** (9 files: catalog, dry-run
  preview, bundled-manifest smoke, both characterization suites, route-order
  guard in the assembled app)
- Full unit suite on the **merged** tree — **7587 passed, 18 skipped, 1 failed**. The one
  failure is `test_1069_voip_call_path_param`, reproduced identically on a pristine
  `origin/dev` worktree **using the same interpreter** (local FastAPI lacks
  `get_flat_dependant`) — pre-existing, not from this branch
- `npm run build` clean
- `TRINITY_MANIFESTS_DIR` proven to reach the backend container env by
  `docker compose config` render on **both** compose files (prod launches
  standalone — dev-only wiring would not carry over)
- e2e spec `system-install.spec.js` covers the section, the `/templates#systems` legacy redirect, preview, and both transport traps
- New frontend design contract (#1430): components pass with **0 raw non-gray classes, 0 hardcoded hex**; the Dashboard raw-color drift is pre-existing on `dev` (verified against a control worktree)

## Not in scope

Async deploy with a job id (the real fix for the timeout window), a
deployed-systems browser, remote/registry manifest sources (ent#14/#108),
per-agent credential setup (ent#127), and `convergent`/`on_conflict` re-deploy
semantics (still deferred under ent#122).


vybe (approved):
Both calls are yours to keep as built — answering the two questions directly:

**1. Yes, it belongs on Library.** The subtitle says "installable assets for your fleet" and a manifest is the most literally installable asset on the page. The "installing is an action, not browsing" objection applies equally to the Agent Templates section, which already carries a Create Agent action — so the page has always mixed browse and install, and a separate route would split one concept across two surfaces for no gain.

**2. Keep the ordering.** Templates → Systems → Skills is right, and for exactly the reason you give: the first two create agents, Skills configures agents that already exist. Appending it would put the odd one out in the middle.

Conforming to the stacked-section model rather than reintroducing the `?tab=` strip was the right instinct, and gating the jump anchor alongside the section (so nav can never point at an unrendered target) is the detail that usually gets missed.

On the substance:

The path confinement is the part with real attack surface and it is layered correctly, with each layer's *reason* stated rather than assumed — the explicit `..` rejection (the regex genuinely does not reject it, `.` being inside the class), the `is_symlink()` check on the **unresolved** entry before `resolve()` erases the link, and open-once + `fstat` + `O_NOFOLLOW` for a host bind mount where stat-then-read has a real window. The catalog/read parity argument for that symlink layer — that without it "not listed" stops meaning "not served" — is the kind of reasoning that makes a guard survive its next refactor.

Sharing pure resolvers between the preview and the writers is the right way to make the dry-run non-lying, and pinning it with characterization tests **captured green before** the refactor (rather than a tautological resolver-vs-writer parity test) is correct. The ent#89 caveat is honestly scoped.

`_preflight_template` validating **merged** resources through the create path's own normalizers is the real bug fix here — a shipped manifest that previewed `valid` and then failed 100% of its agents is precisely the failure a dry-run exists to prevent.

Config packaging verified: `TRINITY_MANIFESTS_DIR` reaches `.env.example` and **both** compose files, and there's a test for the `${VAR:-}` empty-string case — which is the trap that would otherwise turn every deployment into `Path("")`. Manifests parse through `_load_manifest_yaml`, so the #1884 size/alias/duplicate-key hardening applies. Security scan clean; all four required checks green.

Cross-tracker ref is the expected shape — I'll set `status-in-dev` on ent#126 by hand after merge, and it stays open for manual close at release.

PR #1915, opened 2026-07-31
fix(ui): show a not-found state instead of a blank page for an unknown agent (#1914)

### LINKED ISSUE #1914: bug(ui): navigating to a non-existent agent renders a blank page instead of a not-found message
## Summary

Navigating directly to an agent URL that doesn't resolve — e.g. `http://localhost/agents/does-not-exist` — renders a **completely blank page**. No error, no "agent not found", no way back except the top nav. The route matches, `AgentDetail.vue` mounts, and then nothing is drawn at all.

`AgentDetail.vue` already has an error banner (`v-if="error && !agent"`, line 25) — it just never fires, because the error never reaches the view.

## Context

Hit while navigating to a stale bookmark/URL for an agent that no longer exists. The user is left staring at an empty dark canvas with no indication that anything went wrong.

**Root cause** — `stores/agents.js::fetchAgent` (line 138) swallows the failure:

```js
async fetchAgent(name) {
  try {
    const response = await axios.get(`/api/agents/${name}`, { headers: authStore.authHeader })
    ...
    return this.selectedAgent
  } catch (error) {
    this.error = error.message
    console.error('Failed to fetch agent:', error)   // <-- swallowed, returns undefined
  } finally { this.loading = false }
}
```

It catches, logs to the console, and returns `undefined`. So in `AgentDetail.vue::loadAgent` (line 858) the `catch` block never runs, `error.value` stays `''`, and `agent.value` is `undefined` — making **both** `v-if="error && !agent"` and `v-if="agent"` false. Nothing renders. The only evidence is a console line the user never sees.

**Enumeration-safety constraint (Invariant #8, #186):** the backend deliberately returns a **uniform 404** for both a non-existent agent and one the caller can't access — that differential is an enumeration oracle. The frontend copy must therefore **not** distinguish "this agent doesn't exist" from "you don't have access to it". One neutral message covers both.

## Acceptance Criteria

- [ ] Navigating to `/agents/<unknown-name>` renders a visible not-found state instead of a blank page
- [ ] The message is neutral and does not disclose whether the agent exists (uniform-404 contract, Invariant #8 / #186) — e.g. "Agent not found, or you don't have access to it."
- [ ] The state offers a next action (link back to the Dashboard) — no dead empty state. Deliberately **not** the Agents list: the agent you asked for isn't in that list either, so it's a second dead end for the same question
- [ ] A non-404 failure (network error, 500, expired token) shows a distinguishable "couldn't load" message with a retry, not the not-found copy
- [ ] `stores/agents.js::fetchAgent` no longer silently returns `undefined` on failure — the caller can tell success from failure
- [ ] The existing tolerant caller in `AgentDetail.vue` (the post-stop status poll, ~line 712, which deliberately tolerates transient fetch errors) still tolerates them and does not start throwing
- [ ] Regression test covering the 404 path renders the not-found state (not a blank page)

## Technical Notes

- `src/frontend/src/stores/agents.js` — `fetchAgent` (line 138): the swallow. Either re-throw, or return a discriminated result; re-throwing is cleaner but requires wrapping the ~line 712 caller.
- `src/frontend/src/views/AgentDetail.vue` — `loadAgent` (line 858) and the error banner (line 25). Minimal fix: distinguish 404 from other errors and render a proper not-found panel rather than the generic red banner.
- `src/frontend/src/router/index.js` line 58 — the `/agents/:name` route. Alternative/complementary approach: a `beforeEnter` guard, though the two other agent routes (`/agents/:name/workspace`, `/agents/:name/brain`) already bail to `AgentDetail`, so fixing the view is the higher-leverage single point.
- Worth checking whether sibling views that fetch a named agent (`AgentWorkspace.vue`, `AgentBrainOrb.vue` — each with their own local `fetchAgent`) share the blank-page failure mode. If so, note it here rather than expanding this issue's scope.
- Same swallow-and-return-undefined shape exists elsewhere in `stores/agents.js`; out of scope here, but a candidate for a follow-up refactor.

obasilakis:
Closes #1914

## The bug

Navigating to `/agents/<unknown>` rendered a **completely blank page** — no error, no way back. `AgentDetail.vue` already **had** an error banner (`v-if="error && !agent"`). It just never fired.

## Root cause — the store, not the view

`stores/agents.js::fetchAgent` caught the 404, logged it to the console, and returned `undefined`:

```js
} catch (error) {
  this.error = error.message
  console.error('Failed to fetch agent:', error)   // <-- swallowed, returns undefined
}
```

So `AgentDetail::loadAgent`'s `catch` never ran: `error` stayed `''` and `agent` stayed `null`, which made **both** `v-if="error && !agent"` and `v-if="agent"` false. Nothing rendered. The only evidence was a console line the user never sees.

Returning `undefined` on failure is indistinguishable from a successful empty fetch — that ambiguity is the whole bug.

## The fix

- **`fetchAgent` re-throws.** The one caller that genuinely tolerates a failure — the post-stop status poll in `AgentDetail` (`waitForAgentStatus`, ~L739) — already wraps it in its own `try/catch`, so it keeps polling through transient errors exactly as before. Verified: only two callers of `agentsStore.fetchAgent` exist, both in this file. (`AgentWorkspace.vue` / `AgentBrainOrb.vue` have their own local `fetchAgent` that hits axios directly and deliberately degrades — untouched.)
- **`loadAgent` splits 404 from everything else.** 404 → a not-found panel. Anything else (network blip, 500, expired token) → the generic banner, now with a **Retry**. Retrying a 404 is pointless, so only the retryable branch offers one.
- **The exit is the Dashboard, not the Agents list** — the agent you asked for isn't in that list either, so the list is a second dead end for the same question.

## The copy is deliberately vague, and a test pins that

The backend returns a **uniform 404** for "no such agent" and "you can't see this agent" on purpose — the differential is an enumeration oracle (Invariant #8 / #186). So the panel says:

> `skills-fixture` doesn't exist, or you don't have access to it.

…covering both without claiming which. One of the specs asserts the *absence* of an existence claim (`not.toContainText(/no such agent|does not exist on this instance|was deleted/i)`) so a well-meaning future copy edit can't quietly leak it back.

## Tests

`src/frontend/e2e/agent-not-found.spec.js` — 3 specs, tagged **`@smoke`** so they run in CI on `ui`-labelled PRs. They're `@smoke`-eligible precisely because the agent deliberately does *not* exist: unlike every other AgentDetail spec, they need no fixture agent, no running container, and no Claude dispatch. ~2s, deterministic.

1. unknown agent renders the not-found state (not a blank page) + "Back to Dashboard" navigates
2. the copy doesn't disclose whether the agent exists
3. a non-404 (mocked 500) shows the *retryable* load error, not the not-found state

**Verified failing on the pre-fix code** — stashed the two source files, re-ran, 3/3 failed; restored, 3/3 pass. Existing `smoke.spec.js` still green. Happy path (real agent loads) confirmed unaffected by the re-throw.

## Verification

Run against a live stack via a Vite dev server on the branch, proxying to the local backend:

| State | Result |
|---|---|
| `/agents/skills-fixture` | "Agent not found" panel + Back to Dashboard |
| Click Back to Dashboard | lands on `/` |
| `/agents/acme-scout` | normal agent page, unchanged |
| mocked 500 | red banner + Retry |
| dark + light | both styled |

## Scope note

`stores/agents.js` has the same swallow-and-return-`undefined` shape in other actions (`fetchAgents`, etc.). Left alone here — none of them produce a blank page, and widening this would turn a targeted bug fix into a store-wide refactor. Worth a follow-up.

### DIFF
--- a/src/frontend/e2e/agent-not-found.spec.js
+++ b/src/frontend/e2e/agent-not-found.spec.js
@@ -1,0 +1,73 @@
+import { test, expect } from '@playwright/test'
+
+/**
+ * Agent Detail not-found state (#1914).
+ *
+ * Regression guard for a BLANK PAGE: navigating to /agents/<unknown> rendered
+ * nothing at all. `stores/agents.js::fetchAgent` caught the 404, logged it to
+ * the console, and returned `undefined`, so `AgentDetail.vue::loadAgent`'s catch
+ * never ran — `error` stayed '' and `agent` stayed null, making BOTH the error
+ * banner (`v-if="error && !agent"`) and the agent body (`v-if="agent"`) false.
+ * The only evidence was a console line the user never sees.
+ *
+ * @smoke, deliberately: the whole point is that the agent does NOT exist, so
+ * unlike the other AgentDetail specs this needs no fixture agent, no running
+ * container, and no Claude dispatch. It is fast and deterministic.
+ *
+ * The name below must never resolve to a real agent. `sanitize_agent_name`
+ * strips a leading underscore (see the `_retention-guard` sentinel in
+ * architecture.md), so a name in this shape cannot be created through the API.
+ */
+
+const MISSING_AGENT = '_e2e-agent-1914-does-not-exist'
+
+test.describe('agent detail — not found', () => {
+  test('@smoke unknown agent renders a not-found state, not a blank page', async ({ page }) => {
+    await page.goto(`/agents/${MISSING_AGENT}`)
+
+    const panel = page.getByTestId('agent-not-found')
+    await expect(panel).toBeVisible({ timeout: 10000 })
+    await expect(panel).toContainText(/agent not found/i)
+
+    // No dead empty state — there is a way out (Product Quality Bar). The exit
+    // is the Dashboard, not the Agents list: the agent you asked for isn't
+    // there either, so the list is a second dead end for the same question.
+    const back = panel.getByRole('link', { name: /back to dashboard/i })
+    await expect(back).toBeVisible()
+    await back.click()
+    await expect(page).toHaveURL(/\/$/)
+  })
+
+  test('@smoke not-found copy does not disclose whether the agent exists', async ({ page }) => {
+    // The backend returns a UNIFORM 404 for "no such agent" and "you can't see
+    // this agent" on purpose — the differential is an enumeration oracle
+    // (Invariant #8 / #186). The UI must not reintroduce it by claiming which
+    // case this is. This asserts the copy covers BOTH, and pins the absence of
+    // an existence claim so a future copy edit can't quietly leak it.
+    await page.goto(`/agents/${MISSING_AGENT}`)
+
+    const panel = page.getByTestId('agent-not-found')
+    await expect(panel).toBeVisible({ timeout: 10000 })
+    await expect(panel).toContainText(/doesn't exist, or you don't have access/i)
+    await expect(panel).not.toContainText(/no such agent|does not exist on this instance|was deleted/i)
+  })
+
+  test('@smoke a non-404 failure shows a retryable load error, not the not-found state', async ({ page }) => {
+    // Retrying a 404 is pointless; retrying a 500/network blip is not. The two
+    // failures must stay visually and behaviourally distinct.
+    await page.route('**/api/agents/*', (route) => {
+      const url = route.request().url()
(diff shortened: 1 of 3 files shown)


vybe (approved):
Validated via /validate-pr — clean pass: root cause fixed at the store, both fetchAgent call-sites audited, enumeration-safe 404 copy pinned by test, regression specs verified failing pre-fix and green in CI (e2e ran). Approving and merging in sequence.

Issue #1920, opened 2026-07-31
bug: hand-rolled Redis single-flight locks — 7 divergent copies; system_seed_service releases with an unconditional tokenless delete


webmixgamer:
## Summary

The SETNX single-flight lock idiom is hand-rolled in **seven** places with seven divergent behaviours, and the bug class #1919 fixed in `restart_fleet` (extending/releasing a lock you no longer own) is live elsewhere — worst in `system_seed_service`, whose `_release_lock` does an **unconditional, tokenless `r.delete(_PROVISION_LOCK_KEY)`** (src/backend/services/system_seed_service.py:476-482): it deletes a successor's lock, the severest form of the class. Mitigated today only by first-run-only usage + the existence backstop.

## The seven sites

| Site | Shape |
|---|---|
| `routers/ops.py` | SETNX+token, ownership-checked gate + CAD release (**hardened by #1919** — shares `redis_breaker_util.lock_token_matches`) |
| `services/agent_service/ephemeral.py:119-131,210-218` | byte-identical to pre-#1919 ops.py, incl. the dead `.encode()` branch — first adoption candidate for the shared helper |
| `services/skill_service.py:509-545` | token + check-and-delete, TTL 1800 |
| `routers/sessions.py:180-186,309-383` | async client, Lua CAS release, context manager — most mature |
| `services/monitoring_service.py:957-1006` | leader lease, own-lease-only refresh (#1464) |
| `services/operator_queue_service.py:278-306` | verbatim copy of monitoring (#1632) |
| `services/system_seed_service.py:463-482` | **no token at all; unconditional delete on release** |

## Proposal

1. Fix `system_seed_service._release_lock` to a token + compare-and-delete release (smallest slice, ships alone if needed).
2. Consolidate onto an ownership-checked `SingleFlightLock` primitive in `redis_breaker_util` (`acquire` / `refresh_if_owned` / `release_if_owned`; fail-open preserved; #1919's `lock_token_matches` is the first shared piece). Adopt in `ops.py` + `ephemeral.py` first; leader-lease sites (`monitoring`/`operator_queue`) may stay on their own-lease idiom or adopt a leader variant.
3. Do NOT regress the deliberate per-site properties: fail-open on Redis down, sessions' Lua CAS, leader-lease semantics.

## Context

[...]

PR #1928, opened 2026-07-31
fix: fleet-restart lock — ownership-checked lease refresh + acquire inside try/finally (#1919)

### LINKED ISSUE #1919: bug: fleet-restart lock — lease refresh isn't ownership-checked, and acquisition sits outside the try/finally (#1912 review follow-up)
## Summary

Two hardening gaps in the `ops:fleet_restart` single-flight lock shipped by PR #1912 (#1860), flagged as informational notes in the post-merge review and explicitly blessed as a follow-up rather than a respin:

1. **The per-iteration lease refresh isn't ownership-checked.** `restart_fleet` refreshes its lease with a bare `lock_client.expire(_FLEET_RESTART_LOCK_KEY, ...)` gated only on the local `lock_held` flag (`src/backend/routers/ops.py`, top of the restart loop, ~:294-298). If a single agent's processing ever exceeds the 900s TTL (the unbounded terms: skill injection of large packages, a wedged Docker daemon mid-recreate), the lease expires, a second caller legitimately acquires, and the first loop then extends the *second* holder's lease while both run — the exact interleaving the lock exists to prevent, with no signal that it happened. The release is already compare-and-delete; the refresh should be held to the same standard.
2. **Lock acquisition sits outside the `try/finally`.** `agents = list_all_agents_fast()` (`ops.py:279`) runs between acquire (`:268`) and `try:` (`:289`). Unreachable today — that helper swallows all errors and returns `[]` — but any statement later added in that gap that raises leaks a 900s fleet-restart lockout (409s until TTL expiry, no holder).

## Context

Follow-up from the approving review on https://github.com/Abilityai/trinity/pull/1912 (notes 1 and 2; note 3 was editorial). Both fixes are small and localized to `restart_fleet`; the deliberate design properties to preserve are fail-open on Redis-down and compare-and-delete release.

## Acceptance Criteria

- [ ] The lease refresh is ownership-checked — refresh only when the stored token is still ours (`GET`-compare-then-`EXPIRE`, or a Lua CAS mirroring the `monitoring:leader` own-lease-only refresh precedent, #1464)
- [ ] **Loss is detected, not masked**: when the refresh finds a foreign (or absent) token, the loop logs it and stops processing further agents (a detected lost lease shrinks the dual-run window; a silent CAS alone does not) — the partial-completion audit entry still records what was done
- [ ] `agents = list_all_agents_fast()` moves inside the `try`, so nothing can ever sit between acquire and the `finally` that releases
- [ ] Fail-open behavior unchanged (Redis down ⇒ proceed unlocked), compare-and-delete release unchanged
- [ ] `tests/unit/test_1860_fleet_restart_adoption.py` extended: lost-lease mid-loop (foreign token ⇒ loop stops + logs + audit still written), and the existing lock-lifecycle tests stay green

## Technical Notes

- All changes confined to `src/backend/routers/ops.py::restart_fleet` (+ the test file); the `_FLEET_RESTART_LOCK_KEY`/TTL constants and `redis_breaker_util.get_breaker_redis` client are unchanged
- Precedent for own-lease-only refresh: the monitoring leader lock (`monitoring:leader`, #1464) and the compare-and-delete idiom already used in this endpoint's own release path
- Accepted edge, out of scope: a client-side timeout on the initial `SET` (server applied it, client raised) leaves an ownerless lock until TTL — inherent to fail-open SETNX, self-heals in ≤900s

webmixgamer:
## Summary

- The `ops:fleet_restart` per-iteration lease refresh was a bare `EXPIRE` gated only on a local flag: after a TTL lapse it would extend a **concurrent caller's** lease while both fleet loops ran — the exact dual-run the lock exists to prevent, silently. The refresh is now an **ownership-checked pre-action gate** (GET-compare via the new shared `redis_breaker_util.lock_token_matches`, same predicate as the release so the two can't drift).
- **Foreign token** ⇒ stop the run with honest partial accounting: `summary`/audit gain `processed` + `stopped_early="lease_lost_foreign"`; completed restarts stand; the release compare-and-delete still runs (foreign-safe by construction — skipping it on a false "absent" would self-lockout the fleet for the TTL). **Absent token** (incl. `EXPIRE`→0 in the GET→EXPIRE sliver) ⇒ SETNX **re-acquire** and continue — nothing is racing us, so a destructive half-done run completes instead of aborting; `lease_reacquired` is audited. Refresh Redis errors stay **fail-open** with one throttled warning per run.
- `agents = list_all_agents_fast()` moved **inside the `try/finally`** (pre-initialized for the `finally`-audit read), so nothing can ever sit between lock acquire and release; an abnormal exit audits as `stopped_early="error"` + exception class name only (#1912 exposure rule), distinguishing an aborted run from an empty fleet.
- **TTL 900→2100**, sized above the slowest single agent: the restart path runs skill injection, itself bounded at `skill_service._INJECT_LOCK_TTL_SECONDS` (1800s) — at 900s a mid-agent lapse was arithmetically guaranteed on the slow path, and detection alone would have shipped against a guaranteed trip. Constants are comment-linked so retuning either side re-opens the math visibly.

## Scope notes

- The gate **shrinks** the dual-run window to at most one in-flight agent restart — it cannot eliminate it (per-agent start locks are #1817).
- `stopped_early`/`processed` are consumed by API/audit readers; the MobileAdmin PWA toast reads only a generic message today — client-side follow-up noted in `mobile-admin-pwa.md`.
- The same bug class lives in 6 sibling hand-rolled lock sites (worst: `system_seed_service`'s unconditional tokenless release) — consolidation filed as #1920; this PR deliberately stays scoped to `restart_fleet` + the first shared helper.

## Changes

- `src/backend/routers/ops.py` — ownership gate, re-acquire, TTL resize, acquire-inside-try, partial-run honesty fields (response `summary` + audit `details`)
- `src/backend/redis_breaker_util.py` — `lock_token_matches` (dual bytes/str, documented as belt-and-braces under `decode_responses=True`)
- `tests/unit/test_1860_fleet_restart_adoption.py` — +11 tests (27 total): 2-agent happy-path lease baseline, foreign stop + release-leaves-foreign, absent re-acquire win/lose, refresh-error scoped fail-open with release still deleting, EXPIRE→0 / EXPIRE-raise, first-iteration loss, bytes token, no-token-values-in-logs, listing-failure releases lock + error audit
- Docs: `architecture.md` Redis-catalog bullet, `agent-lifecycle.md` (single-flight section, tests line, changelog row, line refs re-verified), `mobile-admin-pwa.md`, `feature-flows.md` index row, 2 `learnings.md` entries (lease-TTL-vs-slowest-iteration; false-detection-vs-self-safe-cleanup)
- Security: `/cso --diff` PASS, zero findings — report pair committed under `docs/security-reports/`

## Test Plan

- [x] `pytest tests/unit/test_1860_fleet_restart_adoption.py` — 27/27
- [x] Live suite `tests/test_ops.py` against the running stack — 41/41
- [x] Adjacent `test_dispatch_breaker.py` + `test_69_ephemeral_agents.py` — 66/66
- [x] Live validation on a real 4-agent stack: baseline (TTL 2098 observed mid-run, lock released), foreign-takeover mid-run (partial stop, foreign token untouched by release, warning + audit row), absent-lease mid-run (re-acquired, run completed, `lease_reacquired:true` audited), concurrent call → 409, fleet healthy after

Fixes #1919


obasilakis (approved):
Validated via /validate-pr: docs, security, packaging, test adequacy and lock-logic correctness all pass. All 21 CI checks green. Approving.

Issue #1900, opened 2026-07-30
Path traversal in template resolution: `local:<id>` read escape, plus `name:`-derived credential staging that leaks another tenant's .mcp.json


AndriiPasternak31:
## Summary

`template_service.get_local_template()` builds a filesystem path by joining a **user-supplied** template id onto the local-templates directory with no containment check:

https://github.com/Abilityai/trinity/blob/dev/src/backend/services/template_service.py#L1174-L1182

```python
def get_local_template(template_id: str) -> Optional[dict]:
    """Get a single local template by `local:<name>` id."""
    if not template_id.startswith("local:"):
        return None
    name = template_id[len("local:"):]
    template_dir = _local_templates_dir() / name   # <-- `name` is caller-controlled
    if not template_dir.is_dir():
        return None
    return _build_local_template(template_dir)
```

`_build_local_template` then reads `template_dir / "template.yaml"` and returns fields from it (`display_name`, `description`, …).

## Reachability

`routers/templates.py`:

```python
@router.get("/{template_id:path}")
async def get_template(template_id: str, current_user: User = Depends(get_current_user)):
```

- The `:path` converter **permits `/`**, so `..` segments are expressible in the id.
- The gate is `get_current_user` only — **any authenticated user, any role**. No owner/admin check, no `require_role`.

## Verified behaviour

Measured against the real `_local_templates_dir()`:

| id suffix | `is_dir()` | resolves to |
|---|---|---|
| `sage` | True | `<base>/sage` (intended) |
| `..` | **True** | `<repo>/config` — **outside the templates root** |
| `../agent-templates/sage` | True | back inside, via traversal |
| `../../../../etc` | False | `<...>/etc` (didn't exist on my box) |

So the join **does** escape the templates root. The read is then `<escaped-dir>/template.yaml`.

## Impact

Bounded but real: an authenticated low-privilege user can probe for directories outside the templates root and, where such a directory contains a file named `template.yaml` that parses as a YAML mapping, have selected fields of it reflected back in the API response.

It is **not** arbitrary file read — the filename is fixed (`template.yaml`), the content must be a YAML mapping, and only specific keys are echoed. Treat it as **directory-existence probing + narrow content disclosure**, not full LFI.

## Not a regression — pre-existing

`get_local_template` is **byte-identical on `dev`** and was **not** introduced or modified by trinity-enterprise#128 (PR-A #1835 / PR-B #1899).

[...]

Issue #430, opened 2026-04-20
decision: fold process engine agent_task through TaskExecutionService, or delete the engine


vybe:
## Summary

`src/backend/services/process_engine/engine/handlers/agent_task.py` calls `AgentClient.chat()` directly, **bypassing `TaskExecutionService`, `SlotService`, `CapacityManager`, cleanup service, and all other orchestration invariants.** `architecture.md` already documents the engine as "OUT OF SCOPE, dormant."

This is the silent exception to every invariant in the Tier 0–2.5 plan. Every hardening PR that ships has a footnote "…except for process engine."

## Decision required

Two honest options — sitting between them is the expensive choice:

### Option A — Fold `agent_task` handler through `TaskExecutionService`

- Process steps inherit slot/timeout/PID/cleanup behaviour automatically.
- Process-engine becomes a specific kind of trigger, not a parallel universe.
- Cost: ~200–400 LOC refactor in the engine handler.
- Risk: process-engine tests may surface drift (the engine has its own observability).

### Option B — Delete the process engine

- Code exists but is not actively being developed (per CLAUDE.md and `architecture.md`).
- `feature-flows/process-engine/` docs move to historical/.
- Removes ~several thousand LOC of dormant surface.
- Cost: one PR with a big red diff, plus stakeholder confirmation that no one is using it.

## Non-option — status quo

Leaving it as-is means every future invariant (CAPACITY-CONSOLIDATE, CLEANUP-COLLAPSE, #306 push completion) has a silent exception. #291 webhook triggers also can't be designed cleanly until this is resolved.

## Parallel to other Tier 2.5 work

This can proceed in parallel with #306 because it's an architectural decision + refactor with no runtime dependency on push completion. Best to resolve **before** CAPACITY-CONSOLIDATE so the new primitive doesn't have to account for a legacy bypass path.

## Decision owner

@vybe — please weigh in on A vs. B. If A, I can scope the refactor. If B, I'd like a confirmation that no live deployment depends on process engine.

## Context

See `docs/planning/ORCHESTRATION_RELIABILITY_2026-04.md` — Tier 2.5 Simplification.

vybe:
Closed by PR #482 — chose Option B (delete). Process engine removed from main; full history preserved on branch `archive/process-engine`.

PR #1934, opened 2026-08-01
fix(templates): stop shipping the dd-* demo fleet and a dead GitHub repo list as starting points

### LINKED ISSUE #1931: bug(library): the agent catalog still ships 11 dd-* VC-demo templates and a Dec-2025 GitHub repo list as real options
## Summary

The Library page's agent catalog still presents the **11 `dd-*` VC-demo templates** as real, installable options, alongside a **hardcoded GitHub repo list last pushed in Dec 2025 / Jan 2026**. Of the 14 locally-bundled templates a user can see today, **3 are genuine starters** (`sage`, `scout`, `scribe`). The first screen a new operator reaches is ~80% demo fixtures and dead links.

This is the *content* half of the catalog problem. #1513 fixed the *fixture* half and closed; `dd-*` was never in its scope.

## What a user sees today

**Starter Templates (local)** — 14 visible of 25 bundled dirs (`config/agent-templates/`, non-`hidden` only):

- `sage`, `scout`, `scribe` — the real starters (`priority: 20`)
- `dd-intake`, `dd-lead`, `dd-market`, `dd-tech`, `dd-founder`, `dd-traction`, `dd-bizmodel`, `dd-captable`, `dd-competitor`, `dd-compliance`, `dd-legal` — 11 templates with `author: Trinity VC Demo`, a purpose-built due-diligence demo fleet. Last touched 2026-01-30 (`69384bf3`). No `hidden:` flag.

**GitHub Templates** — 6 entries from `DEFAULT_GITHUB_TEMPLATE_REPOS` (`src/backend/config.py:329`): `agent-ruby`, `agent-cornelius`, `agent-corbin`, `ruby-orchestrator`, `ruby-content`, `ruby-engagement`. All last pushed Dec 2025 / 2026-01-03 — the pre-2026 Ruby-era set. `agent-cornelius` also duplicated the bundled `cornelius` until that directory was removed.

## Expected

The catalog offers only templates we would stand behind as a first agent. Demo fleets are available deliberately (a demo manifest, a flag, an explicitly-labelled section) — never mixed into the default browse list.

## Technical notes

- Catalog assembly: `src/backend/services/template_service.py` → `get_all_templates()` returns `local` + `github`.
- `local` = every non-`hidden` directory under `config/agent-templates/`. The `hidden: true` mechanism already exists — #1513 built it and applied it to the 11 fixture/demo/system dirs. Hiding `dd-*` is the same one-line-per-template change.
- `github` = rows in `system_settings.github_templates` when an admin has configured them (`GET/PUT/DELETE /api/settings/github-templates`, `src/backend/routers/settings.py:1278`), **else** the hardcoded `DEFAULT_GITHUB_TEMPLATE_REPOS` fallback. No instance has ever written an override row, so every install sees the stale list.
- Convention guard: `tests/unit/test_local_templates_listing.py` covers the local-listing contract.

## Relationship to the structural fix

This is the tactical cleanup, deliberately separable from — and much cheaper than — the two issues that fix the underlying model:

- `abilityai/trinity-enterprise#14` — remote template registry, so the catalog is a manifest we edit rather than an image we cut a release for.
- `abilityai/trinity-enterprise#137` — curating the public starter fleet that the registry then serves.

Do this one now so the Library stops shipping demo content; the registry work supersedes the mechanism later. Adjacent: #1908 (the three real starters ship no `.gitignore`).

## Acceptance criteria

- [ ] The 11 `dd-*` templates no longer appear in the default Library catalog
- [ ] The `dd-*` demo fleet remains reachable deliberately (demo manifest / explicit opt-in), not deleted outright — it is a working VC demo we still use
- [ ] `DEFAULT_GITHUB_TEMPLATE_REPOS` no longer offers repos that are unmaintained or that 404/duplicate a bundled template — either refreshed to a list we stand behind, or emptied so the section renders its empty state
- [ ] The GitHub Templates section teaches the next action when the list is empty (no dead empty state — quality bar #3)
- [ ] A fresh install's Library shows only templates that produce a working agent (quality bar #1)
- [ ] Convention test updated so a newly-added bundled directory must declare its catalog intent rather than defaulting to visible

vybe:
Fixes #1931

A fresh install's Library advertised **14 local templates — 11 of them the `dd-*` VC
due-diligence demo fleet** — plus **6 GitHub repos last pushed Dec-2025/Jan-2026** that no
install had ever overridden. Both were presented as real starting points. They are not: the
`dd-*` agents only work as a deployed set, and the GitHub list was a bundled default nobody
curated. The catalog now shows **3** starters (`sage`, `scout`, `scribe`) and **0** GitHub
repos.

## What changed

- **`hidden: true` on the 11 `dd-*` templates** via the existing #1513 mechanism — omitted
  from `GET /api/templates`, still creatable by id. The fleet stays deployable as a *set*
  through a bundled manifest.
- **`DEFAULT_GITHUB_TEMPLATE_REPOS = []`**. The constant is *kept*, not deleted — TMPL-001's
  `None`-vs-`[]` fallback, `routers/settings.py`, and the Settings "defaults" badge all
  reference it, and trinity-enterprise#14 will repoint the same seam.
- **Every bundled template directory must now declare `hidden:`** (CI-enforced). The *runtime*
  default is deliberately unchanged — flipping it would turn a forgotten key into a *silent*
  absence.
- **Library + Settings empty states** now name an action the user can actually take.
- **`.gitleaks.toml`: `YOUR_TOKEN` allowlisted** (a `regexes` entry, **not** `paths`). The new
  manifest's DEPLOY comment reuses the sanctioned `Authorization: Bearer YOUR_TOKEN` placeholder
  already in `research-network.yaml:10`, which trips gitleaks' default `curl-auth-header` rule;
  a `regexes` entry suppresses only that placeholder, so `config/manifests/` stays scanned for
  real secrets (a `paths` block is a pre-scan file skip and would not).

## Deliberate decisions (not scope drift)

- **The marketplace-first empty state was an explicit human decision that superseded the
  plan.** The `abilityai/abilities` marketplace and its `create-agent` wizards are presented as
  the **recommended** path; Settings → GitHub Templates + `owner/repo` is **secondary**. The
  plan had these the other way round.
- **`Settings.vue` is AC-#4 completion, not scope drift.** The Library's new hint pointed at a
  destination that itself dead-ended — Settings said *"No templates configured. Add a GitHub
  repo above **or reset to defaults**"* while the Reset button is `:disabled` in exactly that
  state and the shipped default list is now empty. Fixing the origin without the destination
  would have left the dead end one click further in.
- **The manifest was PROMOTED from `docs/demos/vc-due-diligence/system-manifest.yaml`, not
  invented**, and is named `vc-due-diligence.yaml` **on purpose**. Deployed agent names are
  built as `f"{system_name}-{short_name}"` (`system_service.py:382`) and
  `dd-lead/CLAUDE.md` hardcodes its roster as **nine** `vc-due-diligence-dd-*` literals. Any
  other manifest name therefore ships 11 healthy containers whose orchestrator 404s on every
  delegation — **with green CI**. A new guard
  (`tests/unit/test_1931_manifest_roster.py`) catches exactly that. The promoted copy also adds
  the `dd-lead → specialist` permissions the demo left to a manual step and drops three keys
  that were dead or harmful there.
- **Tag-alongs, named so they aren't mistaken for drift:** the `Library.vue` page-level hint
  copy; the ent124 test widening from a fixed path to a glob; the
  `test_real_catalog_surfaces_the_three_starters` rename and its now-inert ordering clause; and
  a third test file (`test_1931_empty_github_defaults.py`) that converts a read-only
  "these two branches are byte-identical" argument into a **running proof**.

### Two intended side-effects

- `GET /api/templates` now makes **zero outbound GitHub calls** on a cold metadata cache.
- The new manifest is **inert data, not a trigger** — `BUNDLED_MANIFEST_PATH` hard-codes
  `config/manifests/default-system.yaml` (`system_seed_service.py:83`), so nothing auto-deploys
  it.

### Measured outcome

| | before | after |
|---|---|---|
| Visible local templates | 14 | **3** (`sage`, `scout`, `scribe`) |
| GitHub repos in the shipped default list | 6 | **0** |
| Bundled dirs omitting `hidden:` | 25 | **0** |

25 bundled template directories total: 3 visible, 22 hidden.

## Verification

- **Full `tests/unit` (verify stage): 6012 passed / 18 skipped**
- **Full `tests/unit` (ship stage re-run): 6010 passed / 18 skipped / 2 failed** — both failures are
  **environmental, not branch-caused**; see "Ship-stage re-run caveats" below.
- 3-file `#1931` set + the 2 modified test files, re-run at ship stage after the docstring fix:
  **69 passed** (`test_1931_catalog_intent.py`, `test_1931_empty_github_defaults.py`,
  `test_1931_manifest_roster.py`, `test_local_templates_listing.py`,
  `test_ent124_default_system_seed.py`)
- `npm run build`: clean
- `/verify-local --skip-agent`: **PASS**; integration **70 passed / 13 skipped / 2 deselected**
- **Image-level proof**: `/app/config/manifests/vc-due-diligence.yaml` is present in the built
  backend image, **6493 B byte-match**, parses to `name = 'vc-due-diligence'`, 11 agents, **zero
  `vc-demo`**.

### Guard mutations all fail RED

| Mutation | Result |
|---|---|
| Drop a `dd-*` `hidden:` key | 3 red |
| Add a visible directory | visible-set pin red |
| Manifest `name:` → `vc-demo` | roster red |
| Break a name inside `.claude/commands/` | roster red |

### Ship-stage re-run caveats (host environment, disclosed)

The ship-stage full-suite re-run was done on a host whose environment differs from the verify
host. Two artifacts, neither branch-caused:

1. **4 files could not be collected** — `test_1771{a,b,c}_*_properties.py` need `hypothesis`,
   which is a *declared* test dep (`tests/requirements-test.txt:26` — `hypothesis==6.161.5`)
   simply **not installed** on this host. Excluded from the re-run; all four are untouched by
   this branch.
2. **2 failed** — `test_1081_physical_meter.py::TestPerAgentMeter::test_available_floors_at_zero_over_ceiling[sqlite]`
   and `::TestDisjointBulkMeter::test_max_respects_506_clamp[sqlite]`. **Shared-state
   interference from a second, unrelated `pytest` process running concurrently on the same host**
   (these tests exercise the fleet-wide `max_parallel_tasks_ceiling` setting, which #506
   deliberately reads uncached from shared `system_settings`). Evidence they are not
   branch-caused:
   - `tests/unit/test_1081_physical_meter.py` is **byte-identical** to `origin/dev`
     (blob `622c4eb4`);
   - both named tests **pass in isolation**;
   - the **whole file passes in isolation** — 14 passed.

   CI runs the suite without a competing process, so this should not reproduce there.

### Pre-existing failure — disclosed, not branch-caused

`tests/unit/test_1069_voip_call_path_param.py::test_flat_path_params_are_agent_name_not_name`
imports `get_flat_dependant` (line 214), which **fastapi 0.141.1 removed**.
`tests/requirements-test.txt:33` floats `fastapi>=0.115.0` with **no upper bound**, so whether
this test passes depends on which fastapi a host resolves.

This branch leaves `tests/unit/test_1069_voip_call_path_param.py`, `src/backend/routers/voip.py`,
**and** `tests/requirements-test.txt` byte-identical to `origin/dev` — verified by blob hash — so
it cannot be branch-caused either way.

> **Honest correction to the hand-off note:** on the host used for this ship run the test
> **passes** (fastapi **0.136.1** resolved, where `get_flat_dependant` still exists). The red is
> real but environment-dependent, not "fails on this host". The unbounded pin is the actual
> defect and is listed as a follow-up below.

- **Delete `docs/demos/vc-due-diligence/system-manifest.yaml`.** It still carries `prompt:`
  (line 46), and `system_service.py:1143` writes that **fleet-wide** via
  `db.set_setting("trinity_prompt", …)`. This PR only adds a `SUPERSEDED` header naming the
  hazard.
- **Pin an upper bound on `fastapi` in `tests/requirements-test.txt`** — the unbounded `>=`
  above is why a green branch can go red on a different host.
- `template-processing.md:560-578` documents a `GITHUB_TEMPLATES = [...]` literal at
  "`config.py:91-164`" that **does not exist** (confirmed: the symbol is absent from a 380-line
  `config.py`). Pre-existing rot, left alone because it is a live #1900 collision file.
- The now-dead `if repo in DEFAULT_GITHUB_TEMPLATE_REPOS:` branch in
  `template_service.py:602` (also #1900's file this run).
- README / onboarding / `MULTI_AGENT_SYSTEM_GUIDE.md` Corbin/Ruby staleness.
- `research-network.yaml` / `acme-consulting.yaml` headers document deploy commands that cannot
  work.
- **`dd-lead` should discover its roster via `list_agents` instead of hardcoding nine names** —
  the root fix for this whole class.
- Process-Engine orphans, dead since #430.
- The fork-to-own browse surface is empty with zero GitHub templates. **Not a regression** — the
  backend fails LOUD with `FORK_TO_OWN_REQUIRED` plus remediation.

## Merge order

Four files collide with sibling PRs. **No rebase is needed today** — `origin/dev` is unmoved at
the base commit. **First-merged wins; the others rebase.**

| File | Note |
|---|---|
| `docs/memory/requirements/core-agent.md` §4.1 | all three PRs; clean 2-bullet insert at the `…since #843.` anchor |
| `docs/memory/feature-flows.md` | index row |
| `docs/memory/feature-flows/template-processing.md` | |
| `tests/unit/test_local_templates_listing.py` | one bottom-anchored hunk at line 312 of 368 |


AndriiPasternak31 (approved):
**APPROVE** — `/review` + `/validate-pr`. Zero critical findings, 3 warnings, none blocking.

I verified the load-bearing claims from source rather than from the PR body, and mutation-tested the guards. Everything holds. One correction below, in the PR's favour.

## Verified independently

| Claim | Method | Result |
|---|---|---|
| `.gitleaks.toml` is a `regexes` entry, not `paths` | read the full `[[allowlists]]` block; grepped every `+`/`-` line | ✅ `'''YOUR_TOKEN'''` appended to `regexes` (`:65`); `paths` (`:50-61`) **byte-untouched**, so `config/manifests/` stays scanned for real secrets. The only added lines naming `config/manifests` are comment text *inside* the regexes entry. `gitleaks` CI passing is the empirical half. |
| `vc-due-diligence.yaml` is inert | grepped every `config/manifests` reference repo-wide; read `_resolve_manifest` | ✅ `BUNDLED_MANIFEST_PATH` (`system_seed_service.py:83`) is a string literal; the resolver reads exactly two sources (env override + that literal). **Zero** `glob`/`iterdir`/`listdir`/`scandir` in the three system modules. Every other reference is packaging (Dockerfile COPY, compose `:ro`). |
| `hidden:` is catalog-only — the fleet stays deployable | traced all read paths | ✅ `get_local_templates()` (`:529`) is the **sole** exclusion point. `get_local_template()`, `crud._resolve_local_template()`, and `system_service._preflight_template` all resolve by directory name and never read the flag. |
| Manifest's name/roster claims | read the cited code | ✅ `base_name = f"{system_name}-{short_name}"` (`system_service.py:383`). `dd-lead/CLAUDE.md` = 13 occurrences, **9 distinct** names — exactly the 9 `permissions.explicit` edges, `dd-intake` correctly absent. The other ten `dd-*/CLAUDE.md`: **0** cross-agent refs, so 9 edges is the complete topology, not an omission. |
| `explicit` = full replacement | read `configure_permissions` | ✅ `:455-459` clears every system agent absent from the map before applying. Comment is accurate. |
| Catalog outcome | audited all 25 dirs on disk | ✅ **VISIBLE = {sage, scout, scribe}**, 22 hidden, **0 undeclared** — matches the PR's table exactly. |
| Tests | ran them | ✅ **69 passed**, reproducing the stated number. |
| Guards aren't vacuous | **mutated the tree, ran, restored** | ✅ Drop `dd-founder`'s `hidden:` → **3 red**. Manifest `name:` → `vc-demo` → **roster red**. Both match the mutation table. Tree restored, 69 green again, `git status` clean. |
| Test hygiene | ran `tests/lint_sys_modules.py` | ✅ 194 violations vs baseline 240 — **no new**. |

Two things the PR under-claims, worth stating: **Invariant #13 needed no code change** — MCP `list_templates` → `client.listTemplates()` → `GET /api/templates` → the same single filter, so all three surfaces converge (only the tool *description* needed editing, which it got). And the **upgrade path is fine** — both compose files bind-mount `./config/agent-templates:/agent-configs/templates:ro`, so existing installs pick up the `hidden:` flags on restart rather than needing a volume reset.

The new empty state's CTA is real, not decorative: `useTemplate({id:'github-custom'})` → `:initial-template` → `form.template` (`CreateAgentModal.vue:347`), and `:447` **explicitly exempts** the sentinel from the "template doesn't exist → reset to blank" guard, with `:365` auto-focusing the repo input.

## Correction: the merge-order table over-warns

I simulated squash-merges in a throwaway worktree rather than predicting from hunk overlap:

| Scenario | Conflicts |
|---|---|
| `dev` + #1935, rebase #1934 | **`docs/memory/requirements/core-agent.md` only** |
| `dev` + #1935 + #1936, rebase #1934 | **`docs/memory/requirements/core-agent.md` only** |

`template-processing.md` **auto-merges**; `feature-flows.md` and `test_local_templates_listing.py` never conflict. The one conflict is a pure adjacent-insert — #1935's `#1900` read-path bullet vs this PR's two bullets at the same §4.1 anchor, zero semantic overlap. Resolution is **keep both, either order**. I resolved it that way, completed all 7 commits clean, and re-ran: **84 passed** (this PR's 69 + #1935's additions to the same files). The two PRs' guards coexist, and #1935's own bullet independently states *"Hidden templates (`hidden: true`) stay resolvable by id: the barrier reads the name, never the flag"* — corroborating this PR's central claim.

Flagging this *down* because an inflated conflict list invites a careless bulk resolution on the one file that actually needs care.

## Warnings (non-blocking)

1. **The superseded demo manifest keeps its live footgun.** `docs/demos/vc-due-diligence/system-manifest.yaml:46` still carries `prompt:`, and `system_service.py:1143` writes it **fleet-wide** via `db.set_setting("trinity_prompt", …)`. The `SUPERSEDED` header names the hazard and the README is repointed — real mitigation — but the header itself says the file "is NOT guarded against drift," and removal is one line. Its sibling claims do check out: `auto_start` has zero reads, and the only `folders` read is per-agent (`:193`). Fine as the filed follow-up; noting the one-line option.

2. **#1936's prose goes stale on rebase.** Its §4.1.1 says "all **14 visible** bundled templates failed the same four HARD checks" — after this PR, 3 are visible, contradicting §4.1 two bullets above. Its *tests* are safe (`GUARDED_TEMPLATES` is an explicit tuple, deliberately not a live `hidden` query, with a comment citing #1931). Docs-coherence fix for whoever lands second.

3. **Issue #1931 carries both `status-in-progress` and `status-ready`.** Auto-promotion writes `status-in-dev` and leaves `status-ready` as drift. Worth dropping before merge.

## Suggestions

- `requirements/security.md` §20.8 doesn't name the new `YOUR_TOKEN` entry. Consistent with precedent (it never listed `your-api-key`/`your-domain\.com` either), so optional.
- `test_1931_manifest_roster.py` mutates `sys.path` at import while its sibling explicitly argues against that, and its `assert paths` runs at **collection** time (called in the `@parametrize` at `:72`) so a missing `config/manifests/` errors rather than skips. Cosmetic.

## Forward-looking

Once #1911 (ent#126) lands, `GET /api/systems/manifests` globs `config/manifests/` and surfaces each manifest as a one-click UI install with a `sets_prompt` warning — so `vc-due-diligence.yaml` stops being merely inert data and becomes a **listed, installable** catalog entry. That reads as AC #2's ideal fulfilment rather than a defect, and I checked the interaction: `_preflight_template` resolves through `_resolve_local_template`, which ignores `hidden`, so all 11 `dd-*` preflight `valid: true` rather than showing up broken. No action — just don't carry "inert" forward as an assumption after #1911.

## Process

All 24 CI checks pass, including `gitleaks`, `prod-image-smoke`, `regression diff`, the sys.modules lint, `e2e`, and 6 pytest runs (3 seeds × base/head). Security greps clean — the only "email" hit is `@pytest.mark.parametrize` and the 5 "IP" hits are GitHub-logo SVG path coordinates. Zero new `os.getenv()` (no config-packaging gap); `config.py` is a modification already covered by `Dockerfile:83`, and the new manifest ships via the wholesale `Dockerfile:109` `COPY ../../config/manifests`. `architecture.md:222` names only `default-system.yaml` for the seed and stays accurate with a second manifest present, so skipping it was right. All 6 ACs trace to a named guard.

The detail I'd single out: `test_real_catalog_surfaces_the_three_starters` noticed its own ordering clause could only go **vacuous, not red**, on the very change it guarded — said so in the docstring, and replaced it with a non-vacuous `priority == 20` assertion. That's the failure mode most PRs ship silently.

**Merge order:** land #1935 first, expect exactly one conflict in `core-agent.md`, resolve keep-both.

Issue #1933, opened 2026-08-01
feat(whatsapp): voice-note transcription — extend the transcription branch beyond Telegram (#468 item 3d)


vybe:
## Summary

WhatsApp voice notes reach the agent as an audio file, never as text. Transcription is gated to Telegram by an explicit channel check in the message router, so a WhatsApp user who sends a voice message gets an agent that cannot read it. This pulls item **3d** out of the #468 umbrella into a standalone, pickable issue.

## Context

`src/backend/adapters/message_router.py` (~line 936):

```python
async def _maybe_transcribe_voice(...):
    """Step 2b: transcribe a Telegram voice note in place."""
    if channel != "telegram":
        return message
```

Two things changed since 3d was written as a speculative bullet inside an umbrella framed as "ship only if demand appears":

1. **Demand appeared.** A live deployment received WhatsApp voice notes from a real user and had to reply that it couldn't hear them.
2. **The channel is asymmetric.** Outbound voice on WhatsApp already exists — an agent can deliver its reply as a playable voice note. Inbound is the remaining gap: the agent can speak but not listen.

Two details in the #468 3d bullet are stale and worth correcting when this is picked up: it specifies the **Whisper API**, but transcription moved to Gemini (`_transcribe_audio_gemini`, model pinned via `GEMINI_TRANSCRIPTION_MODEL` after #1130); and it cites `message_router.py:212`, while the branch now lives at ~936.

**Hard dependency:** WhatsApp media download is currently broken (#1932) — the SSRF allowlist refuses Twilio's media CDN, so `download_file` returns `None` for every attachment. Transcription cannot be built or tested until that lands.

## Acceptance Criteria

- [ ] A WhatsApp voice note is transcribed and the transcript replaces the media placeholder in the agent's prompt, matching Telegram's behavior
- [ ] The transcription branch becomes **channel-generic** rather than gaining a second hardcoded channel name — the guard keys on "this message carries a voice attachment this adapter can fetch", not on a channel string
- [ ] Reuses the shared `_transcribe_audio_gemini(bytes, mime_type)` helper — no WhatsApp-specific transcription implementation
- [ ] Audio is fetched through the existing SSRF-gated `WhatsAppAdapter.download_file` — no new fetch path
- [ ] Graceful fallback: on download failure, transcription failure, or unsupported MIME type, the agent still receives the message with an honest placeholder naming what failed — never a silent drop
[...]

Issue #1315, opened 2026-06-23
feat(whatsapp): outbound media attachments — deliver ChannelResponse.files to WhatsApp users via Twilio MediaUrl


vybe:
## Summary

WhatsApp agents can't send files to users. The WhatsApp adapter's `send_response` only delivers `response.text` and silently drops `ChannelResponse.files` — the channel-agnostic outbound-file field that the Slack adapter (and Telegram) already honor. Today an agent's only way to "send a file over WhatsApp" is to paste a FILES-001 download link into its text reply (no native attachment/preview). This issue wires native Twilio outbound media (the `MediaUrl` param) so a WhatsApp agent can deliver an actual image / PDF / CSV / document inline.

## Context

- **Inbound media already works**: the adapter parses `MediaUrl{N}` from Twilio webhooks into `NormalizedMessage.files` (`src/backend/adapters/whatsapp_adapter.py` ~168–208) and downloads via the SSRF-gated `download_file`.
- **Outbound is the gap.** `ChannelResponse.files: List[OutboundFile]` exists in `src/backend/adapters/base.py`; `src/backend/adapters/slack_adapter.py:145-155` uploads them via `slack_service.upload_file`, and Telegram has a `(file upload)` path. WhatsApp's `send_response` (`whatsapp_adapter.py` ~236–276) never iterates `response.files` — it returns early on empty text and only sends the body.
- Twilio's WhatsApp send takes a **publicly-reachable `MediaUrl`** that Twilio fetches server-side. FILES-001 (#295) already mints public `?sig=` download URLs from agent files (`src/backend/services/agent_shared_files_service.py`, `src/backend/routers/files.py`), so the storage + public-URL plumbing exists to reuse.
- Tracked as a slice of WhatsApp Phase 3 (#468); builds on WHATSAPP-001 Phase 1 (#299) / Phase 2 (#467) and FILES-001 (#295).

## Acceptance Criteria

- [ ] `WhatsAppAdapter.send_response` iterates `response.files` and delivers each as a Twilio `MediaUrl` (alongside the text body), reaching parity with the Slack adapter
- [ ] Each outbound file is persisted to FILES-001 storage and exposed via its public `?sig=` URL (HTTPS, Twilio-reachable); media is delivered **by URL**, not inlined as bytes
[...]

Issue #468, opened 2026-04-23
feat(whatsapp): Phase 3 advanced features — templates, SMS, interactive, voice (WHATSAPP-001 Phase 3, #299 follow-up)


vybe:
## Summary

Umbrella tracking issue for WHATSAPP-001 Phase 3. Phase 1 (#299) shipped direct-message chat. Phase 2 (#467) wires access control. Phase 3 unlocks advanced Twilio capabilities: outbound-first messaging, SMS, interactive UX, voice transcription.

Each bullet below can be split into its own issue when picked up — kept together here so we can evaluate scope and sequencing as one conversation.

## Context

Phase 1 operates inside Twilio's 24-hour customer-service window (an inbound message opens the window; freeform outbound is free within it). This is fine for reactive agents but blocks:

- Scheduled/proactive notifications to WhatsApp users
- Initiating conversations from the agent side
- Rich UX beyond plain text

Phase 3 addresses each with targeted Twilio primitives. No single item here is urgent — ship only if demand appears.

### 3a — Message templates (outbound-first / post-24h messaging)
- [ ] Twilio Content Builder integration: register + reference pre-approved templates
- [ ] Template storage: `whatsapp_templates` table (template_sid, name, category, placeholders schema, approval status)
- [ ] Outbound send helper accepts `template_name` + `variables` → builds payload → routes via Content API
- [ ] Error 63016 ("outside window") now falls back to template send if one is configured
- [ ] UI: list registered templates per agent, show approval status
- [ ] Unit tests for template payload construction

### 3b — SMS on same Twilio binding
- [ ] Reuse `whatsapp_bindings` row (drop `whatsapp:` prefix for SMS routes)
- [ ] New `SmsAdapter` subclassing `ChannelAdapter` OR extend `WhatsAppAdapter` with a `medium` flag — pick whichever is cleaner after review
- [ ] Router logic: `From: +E164` (no prefix) vs `From: whatsapp:+E164` dispatches to appropriate adapter
- [ ] Rate limiting keyed on `twilio_sms:{binding_id}:{phone}`
- [ ] UI: checkbox on Sharing panel "Also accept SMS to this number"
- [ ] Session identity: decide whether SMS + WhatsApp from the same phone share a session (likely yes — same `verified_email`)

### 3c — Interactive messages (buttons, list pickers)
- [ ] Twilio Content API quick-reply buttons and list messages
[...]

PR #1937, opened 2026-08-01
fix(whatsapp): allow Twilio's media CDN as a validated redirect target (#1932)

### LINKED ISSUE #1932: bug(whatsapp): inbound media download always fails — SSRF allowlist rejects Twilio's own media CDN
## Summary

Every inbound WhatsApp attachment fails to download. Twilio serves message media from `mms.twiliocdn.com`, reached via a 302 from `api.twilio.com`. `WhatsAppAdapter.download_file` deliberately does not auto-follow redirects and re-validates the redirect target against an SSRF allowlist that accepts only the `.twilio.com` suffix — so the CDN host is refused and `download_file` returns `None`. The agent receives a `<name> — download failed` placeholder instead of the file.

This is not voice-specific: it breaks **every** inbound media type on WhatsApp — images, PDFs, documents, audio. It has been this way since the adapter shipped (#463, 2026-04-23), so inbound WhatsApp media has never worked.

## Context

`src/backend/adapters/whatsapp_adapter.py`:

```python
_TWILIO_MEDIA_ALLOWED_HOST_SUFFIXES = (".twilio.com",)   # line 44
```

```python
# download_file, ~line 669
async with httpx.AsyncClient(timeout=30.0, follow_redirects=False) as client:
    resp = await client.get(file.url, auth=(binding["account_sid"], auth_token))
    # Twilio sometimes 302-redirects to a signed S3 URL for large media.
    # Follow only if the redirect target is still Twilio-hosted.
    if resp.status_code in (301, 302, 303, 307, 308):
        location = resp.headers.get("location", "")
        if not _is_twilio_media_url(location):
            logger.warning(
                "[WHATSAPP] Refusing off-domain media redirect to host=%s",
                urlparse(location).hostname,
            )
            return None
```

`mms.twiliocdn.com` is a different registrable domain from `twilio.com`, so `_is_twilio_media_url` returns `False` and the download is abandoned. The inline comment ("sometimes… for large media") understates the behavior — Twilio redirects to the CDN for ordinary media too, which is why the failure is total rather than intermittent.

Observed on a deployed instance: repeated `[WHATSAPP] Refusing off-domain media redirect to host=mms.twiliocdn.com` warnings in the backend log, with the agent's retained history showing `media_0.ogg — download failed` for each attachment and the agent replying to the sender that the voice message "didn't come through." Every attachment that instance has ever received failed this way.

**Why it went unnoticed.** The shipping PR (#463) tested the SSRF allowlist with six *spoof* vectors (`eviltwilio.com`, `api.twilio.com.evil.com`, plain-`http://`, garbage) — all negative cases. No test exercised a real Twilio 302, so the legitimate CDN hostname never appeared in test or review. Two issues consequently record inbound media as working: #1315's Context section states "**Inbound media already works** … downloads via the SSRF-gated `download_file`", and #468 item 3d treats `download_file` as a working prerequisite.

**No ops workaround exists** — the allowlist is a hardcoded module constant with no env override.

## Acceptance Criteria

- [ ] An inbound WhatsApp image, document, and audio attachment each land in the agent's workspace as real bytes, verified end-to-end against a live Twilio sender
- [ ] `_TWILIO_MEDIA_ALLOWED_HOST_SUFFIXES` accepts Twilio's media CDN (`.twiliocdn.com`) alongside `.twilio.com`
- [ ] The redirect target is still **validated**, never blindly followed — `follow_redirects=False` plus the explicit allowlist check stays; only the allowlist widens
- [ ] Anti-spoof property preserved: `eviltwiliocdn.com`, `twiliocdn.com.evil.com`, and any non-`https` scheme are still refused — the existing #463 spoof tests keep passing, with CDN-shaped spoofs added
- [ ] New unit test covering the real-world path: `api.twilio.com` → 302 → `mms.twiliocdn.com` → 200, bytes returned
- [ ] Download failure stays non-fatal and honestly reported: the agent still receives the message, and the placeholder says the download failed rather than implying the file arrived
- [ ] Decide explicitly whether the allowlist gets an env override so a future Twilio host change is an ops fix rather than a release — record the decision either way
- [ ] `docs/memory/feature-flows/whatsapp-integration.md` inbound-media section reflects the actual redirect chain
- [ ] The stale "inbound media already works" claim in #1315 is corrected (a comment on the closed issue is enough)

## Technical Notes

- Primary file: `src/backend/adapters/whatsapp_adapter.py` — constant at line 44, `_is_twilio_media_url` at ~151, `download_file` at ~647 (verified on `dev`)
- The suffix match (`host.endswith(s)` against a leading-dot suffix) is already spoof-safe; adding `".twiliocdn.com"` to the tuple preserves that property
- The redirect is followed **without** auth (the signed CDN URL carries its own credentials) — correct as-is, should not change
- Twilio may also serve `media.twiliocdn.com`; a `.twiliocdn.com` suffix rather than a single fixed host covers both
- Prerequisite for WhatsApp voice-note transcription — without it the transcription branch would have nothing to transcribe

## Related

- Introduced by #463 (WHATSAPP-001 Phase 1, #299)
- **Blocks #1933** — WhatsApp voice-note transcription (#468 item 3d)
- Corrects the inbound-media premise asserted in #1315

vybe:
Fixes #1932

## What was broken

Twilio **never** serves inbound WhatsApp media from `api.twilio.com` directly — it 302s **every** attachment to its media CDN (`mms.twiliocdn.com`). The SSRF allowlist only permitted `*.twilio.com`, so `download_file` refused the redirect and returned `None`.

**Every inbound WhatsApp attachment has failed since #463 (2026-04-23)** — roughly three months on a feature that looked shipped. The pre-existing `WARNING` for the refusal surfaced nothing in that entire window (which is why it is now `ERROR`).

## The fix — a **two-tier** allowlist, deliberately narrower than the AC asked for

The AC said "widen the allowlist". Widening the single existing constant would also have widened the **credentialed** hop — the request that carries the tenant's `AccountSid:AuthToken` as HTTP Basic auth. Three independent reviewers (`/autoplan` strategy, `/autoplan` security, and the Gemini second voice) called that ~3× wider than the bug. So the constant splits:

| Constant | Value | Gates |
|---|---|---|
| `_TWILIO_MEDIA_SOURCE_HOST_SUFFIXES` | `.twilio.com` | webhook `MediaUrl{N}` parse gate **+ credentialed hop 1** |
| `_TWILIO_MEDIA_ALLOWED_HOST_SUFFIXES` | `.twilio.com`, `.twiliocdn.com` | validated 30x **redirect targets only** (unauthenticated) |

**AC #2 is satisfied literally** — the AC-named constant `_TWILIO_MEDIA_ALLOWED_HOST_SUFFIXES` still receives `.twiliocdn.com`. The split only means the *credentialed* blast radius stays exactly as narrow as it was before this PR. Verified narrow by grep: exactly three call sites — `:310` (parse) and `:724` (hop 1) use the **narrow** predicate; only `:759` (redirect target) uses the wide one. `test_allowlist_constants_are_the_documented_tiers` pins `SOURCE ⊆ ALLOWED`.

Also in the diff, each a direct consequence of making this path reachable for the first time:

- `follow_redirects=False` **stays** — every hop is re-validated *before* it is issued, nothing is blind-followed. Single-follow became a **bounded budget** (`_MAX_MEDIA_REDIRECTS = 3`) so a future Twilio 302→302 chain degrades to "still works" instead of silently reproducing this exact bug.
- **The redirect hop carries no auth.** The signed CDN URL has its own credentials; forwarding the tenant AuthToken to a CDN host would widen the credential's reach.
- **Transport size cap** (`16 MB`) — Telegram (20 MB) and Slack (10 MB) both cap; WhatsApp's read was unbounded, and this fix makes that unbounded read reachable for the first time.
- **Exception-log hygiene** — `type(e).__name__` only. Some `httpx` exceptions (`UnsupportedProtocol`, `InvalidURL`) embed the URL, and the signed CDN URL is a live ~4-hour bearer capability to the media.

---

## ⚠️ AC #1 is NOT met by this PR — read this before verifying

**Two reasons, and the second one matters more:**

**(a) No live-sender verification was possible in this run.** It needs a real Twilio account with *HTTP Basic Authentication for media* enabled, a publicly reachable webhook URL, and a provisioned WhatsApp sender.

**(b) The AC's premise is partly wrong.** `services/upload_service.py:45-49` `UNSUPPORTED_MIMES` rejects `application/pdf` **and** `audio/*` for **every** channel, enforced channel-agnostically at `upload_service.py:177`. So of the AC's three attachment types, only the **image** can actually land as bytes.

A PDF or a voice note will now be **fetched successfully** by this fix and then surface as **"unsupported format"** — that is a separate, deliberate policy gate, **not a regression from this PR**. Expanding scope into `UNSUPPORTED_MIMES` was explicitly rejected as out of fence.

**How a human should verify on a deployed instance:**
1. Send the agent an **image**. It should reach the workspace, and the backend log should contain **no** `Refusing off-domain media redirect` warning.
2. If you also send a **PDF or a voice note**: reading **"unsupported format"** rather than "download failed" is the **expected, improved** outcome. That is the fix working. **Please do not file it as a regression.**

---

## `*.amazonaws.com` was rejected — and it is worse than it looks

Accounts **without** media auth enabled get redirected to `s3-external-1.amazonaws.com`. That host is **path-style** (`/<any-bucket>/<key>`), so allowlisting it — *even as a single exact hostname* — admits arbitrary attacker-controlled content under an allowlisted name. That is a bypass primitive, not a widening.

The supported configuration is therefore **media-auth-enabled accounts**. An account without it will still fail — with the `Refusing off-domain media redirect to host=s3-external-1.amazonaws.com` log line as the diagnostic, and a new operator-facing troubleshooting table in the user docs that names exactly this case and how to fix it in the Twilio Console.

**This is test-pinned, not prose:** `test_s3_redirect_refused`.

## Env override: **NO** — recorded decision (AC #4)

Rationale, recorded as a comment at the constant so it survives without this PR body:

- Trinity has **zero** env-driven outbound-fetch / SSRF allowlists. (`EXTRA_CORS_ORIGINS` is *inbound* CORS, not an outbound fetch gate; `telegram_media.ALLOWED_DOWNLOAD_HOST` and `url_validation`'s github.com lock are both hard-coded; `base_image_allowlist` is an **admin DB setting with an audit trail**.)
- This gate protects a fetch carrying a **tenant's** credential, while env is a **platform-scope** control. An env knob would let one operator-level actor redirect every tenant's AuthToken.
- If an escape hatch is ever genuinely needed, the correct shape is the admin-DB-setting-with-audit pattern (`agent_service/helpers.py::validate_base_image`) — never env.

---

## Anti-vacuity evidence (this is what makes the suite trustworthy)

`download_file` returns `None` on **every** failure path, so a negative test passes even against a completely unpatched seam. Every negative test therefore asserts the **recorded call log / hop count**, not just `is None`. Four deliberate breaks were introduced and all four were detected (tree restored clean after each):

| Break | Result |
|---|---|
| **A — fixture:** `httpx.Headers` → plain `dict` | **7 red.** Decisively: `test_non_200_after_redirect_returns_none` and `test_redirect_budget_exhausted` went red **purely on hop count** — captured output shows `assert result is None` *passing*, then `assert len(calls) == 2` → `assert 1 == 2`. **Proof the suite is not carried by the all-paths-return-`None` trap.** |
| **B — product:** revert the widening | **12 red** across both files |
| **C — credential:** re-attach `auth=` to the redirect hop | **exactly 2 red** |
| **D — ordering:** move the allowlist check after the fetch | **7 red** — the security ordering is pinned, not incidental |

A `TestInboundMediaHarnessIntegrity` class fails loudly if the DB seam or the `httpx.Headers` fake ever regresses.

**`/cso` CLEAN with adversarial evidence:** a 37-vector corpus against the widened tier with a `urlparse` ↔ `httpx.URL` **cross-check** found no bypass and no exploitable differential (the single disagreement — a leading-space URL — is unreachable from a parsed HTTP header and fails closed). Multi-`Location` join measured on real `httpx.Headers`: attacker-first → denied; twiliocdn-first → httpx still resolves `host='mms.twiliocdn.com'`, so no attacker host is ever contacted.

**A second-voice claim was tested and refuted, twice, by two different workers:** the alleged hop-1 cookie leak to the CDN does not occur — with a hop-1 `Set-Cookie: Domain=.twilio.com`, hop 2 carries `authorization=None` **and** `cookie=None` (httpx enforces RFC 6265 domain matching).

## AC #3 — the #463 spoof tests

The six #463 spoof tests are **byte-identical to `origin/dev`** (verified by per-function diff — all six IDENTICAL), node-ids unchanged. `_is_twilio_media_url`'s name and signature are preserved. `_host_matches_suffixes` takes `suffixes` as a **required positional** — no permissive default that a future call site could silently inherit.

CDN-shaped spoof vectors added on top (`mms.twiliocdn.com@evil.com`, `mms.twiliocdn.com.evil.com`, and the scheme/userinfo family).

---

## Accepted residuals — named so none of them reads as an oversight

- **No private-IP / DNS-rebinding resolution check.** The allowlist is name-based; a hostname under an allowlisted suffix that resolves to a private address is not caught.
- **Port is unconstrained in the matcher** — and that now spans a second registrable domain.
- **Worst-case unauthenticated fetches per attachment go 2 → 4** (the bounded budget). All targets are allowlisted, and `_process_update` is a background task, not a request hold.
- **Files > 16 MB now read "download failed" instead of "exceeds 10 MB limit."** The transport cap sits *above* `upload_service`'s policy caps on purpose so the policy layer keeps producing its honest message; 16 MB is itself above Twilio's own 16 MB WhatsApp media ceiling, so this is the pathological case only.
- **Inherited, not introduced:** `_host_matches_suffixes` keeps the pre-existing `host == s.lstrip(".") or host.endswith(s)` shape, so the widened tier also admits bare `twiliocdn.com`. It is the line the two-tier split rests on.

## Pre-merge operator check (2 minutes)

Confirmation, not a blocker — the bounded budget already covers a chain. Please run against a real account and paste the output:

```bash
curl -sI -u "$ACCOUNT_SID:$AUTH_TOKEN" \
  "https://api.twilio.com/2010-04-01/Accounts/$SID/Messages/$MSG/Media/$ME"
```

Then follow the `Location` once. This confirms the redirect target host and that hop 2 answers `200`.

---

## Corrections to other issues (recorded here as text; no comments were posted)

- **#1315** — its Context claims inbound media already works. Half-true: the **parse** half worked, the **download** half never did, from #463 (2026-04-23) until this PR.
- **#468 item 3d** — treats `download_file` as a working prerequisite. Same correction.

## #1933 relationship

**Unblocked — but only for media-auth-enabled accounts.** `download_file`'s signature and its `None`-on-failure contract are **byte-stable** (verified by diff), so #1933 can build on it unchanged.

> **Correction (review round 2).** The original wording here said "`_maybe_transcribe_voice` runs at `message_router.py:406`, **before** the upload path, so the transcription route does benefit from this fix." **That is wrong for WhatsApp.** `_maybe_transcribe_voice` opens with `if channel != "telegram": return message` and calls `telegram_media.process_voice(bot_token, raw_msg["voice"])` — a Telegram-Bot-API fetch that never touches `adapter.download_file`. There is **no** WhatsApp transcription route today, so a WhatsApp voice note is now fetched successfully and then rejected as `"— unsupported format"`, full stop. #1933 therefore has to *build* that route; it does not inherit one. The claim is corrected in `feature-flows/whatsapp-integration.md` in this branch.

## NEW security finding — follow-up recommendation (deliberately not filed)

**Slack's `download_file` has no SSRF allowlist at all.** `slack_service.py:739` fetches an inbound-event URL with `follow_redirects=True` (`:754`) — no allowlist, no scheme check. Same bug class as this issue, different channel.

Deliberately **not** widened into this PR. The eventual shape is a shared `channel_media_fetch(url, allowlist, auth, max_bytes)` covering WhatsApp / Telegram / Slack uniformly.

## Other follow-ups (listed, not filed)

1. Private-IP / DNS-rebinding resolution check **+** a true streaming size bound for the WhatsApp fetch (the unconstrained-port item folds in here).
2. The `UNSUPPORTED_MIMES` residual — if PDFs should reach agent workspaces, that is a deliberate policy change, not a bug fix.
3. Repo-level: `fastapi>=0.115.0` is a floor, not a pin (see Verification below).

---

## Verification — stated honestly

**`/verify-local` ran with `--skip-agent`.** This was forced by the host: the operator's live dev stack owns the global `trinity-agent-network`, so the full agent stage hard-refuses. **This is a host-environment constraint, not a passing agent stage.** The skipped stages (3: `import agent_server` inside the built base image; 5: real agent `/health`) cover no surface this PR touches — the diff contains no `docker/base-image/` file and no new backend module.

**WAVE-4 result:**

| Stage | Result |
|---|---|
| Build + import smoke | **pass** (18s) |
| Boot + health | **pass** (13s) |
| Integration | **70 passed**, 13 skipped, 2 deselected |
| Unit (full suite) | **1 failed, 6054 passed, 18 skipped, 1 xfailed** |

**The single red is proven pre-existing, not assumed.** `tests/unit/test_1069_voip_call_path_param.py::TestVoipCallPathParam::test_flat_path_params_are_agent_name_not_name` was run from a pristine `git archive origin/dev` @ `8e924526` extracted into a temp tree, using the same verify venv via the harness's own invocation → **identical** `ImportError: cannot import name 'get_flat_dependant'`. The file is byte-identical across this branch and `dev`.

**Repo-level root cause:** `fastapi>=0.115.0` is a **floor, not a pin**, so any fresh venv now resolves 0.141.1, where `get_flat_dependant` no longer exists. At the prod pin `fastapi==0.115.6` the file is `4 passed`. This will red for every contributor building a clean environment and deserves its own issue — **it must not be attributed to this PR.**

**Targeted suite, re-derived first-hand at ship time** (explicit collected counts; **no `-m unit` anywhere** — that flag silently deselects: measured `65 collected / 42 deselected / 23 selected`):

```
pytest tests/unit/test_whatsapp_inbound_media.py \
       tests/test_whatsapp_adapter.py \
       tests/unit/test_whatsapp_outbound_media.py -q
→ 148 collected, 148 passed, 0 deselected   (23 + 83 + 42)

pytest tests/unit/test_whatsapp_inbound_media.py tests/test_whatsapp_adapter.py -q -p no:randomly
→ 106 passed   (seed-independent)
```

## Notes for the reviewer

- **The HMAC citation in the new code comment is symbol-anchored on purpose.** Line numbers in `transports/twilio_webhook.py` drift; the comment names `TwilioWebhookTransport.handle_webhook` and `RequestValidator(...).validate(...)` instead. Please do **not** "restore" line numbers there.
- **The user-doc troubleshooting table is the operator-facing complement** to the `type(e).__name__` log narrowing — the narrowed log is deliberately less informative, and the table is what makes it actionable.
- **Uncommitted by design:** the `.claude/agents/test-runner.md` catalog sync is intentionally left uncommitted (it lives in the private `trinity-dev` submodule) and needs a separate commit there. It is not part of this PR.
- **Merge order:** this PR touches neither `docs/memory/learnings.md` nor `tests/registry.json`, so it has no conflict with the sibling PRs in this batch and can merge in any order relative to them.

---

## Review round 2 (2026-08-03) — merge + reviewer follow-ups

Rebased onto `dev` and applied the reviewer's pre-merge item. Head is now `33c28d64`; PR is **MERGEABLE** again.

**Merged `origin/dev`** (`c55ff804`). One conflict, in `docs/memory/feature-flows.md` — both sides prepended a row to the *Recent Updates* table (#1932 here, #1931 on `dev`). Both rows kept, #1932 above #1931. `whatsapp_adapter.py` and both test files were untouched by the merge (verified by empty `git diff HEAD -- <paths>`), and the #1932 `architecture.md` line survived the auto-merge intact.

**F1 applied — both SOURCE-tier gates now log at ERROR** (`33c28d64`). The reviewer's framing is right and worth restating: #1932's root cause was never the allowlist, it was a fail-closed gate reporting a 100% outage at a level nobody reads. Raising only the redirect gate left the *stricter* half — the one the two-tier split newly made most likely to break on the next vendor change — guarded at WARNING.

Not left as prose, because the level is a one-word revert with a three-month observed detection latency. New `TestSourceTierRejectionsAreVisible` pins both, and is mutation-proven: reverting either `logger.error` REDs the pair on `assert {30} == {40}`, and both go green on restore.

**Both stale doc claims fixed.** `feature-flows/whatsapp-integration.md:17` and `requirements/public-access.md:309` described inbound media as "images/audio/PDFs"; `upload_service.UNSUPPORTED_MIMES` rejects `application/pdf`, `audio/`, `video/` and archives channel-agnostically, so only images (plus text/CSV/JSON) reach a workspace. Chasing that down is what surfaced the `_maybe_transcribe_voice` error corrected above.

**Vendor evidence folded into the user doc.** Media auth has been mandatory on every main account created since **2023-07-31** (Twilio changelog, linked in the doc), so the S3 residual reaches only pre-July-2023 grandfathered accounts. The prerequisite now opens with "most accounts already have this and need no action" instead of "opt-in and off by default" — materially less alarming, and accurate.

### Verification

| Check | Result |
|---|---|
| `test_whatsapp_inbound_media.py` + `test_whatsapp_adapter.py` (`-p no:randomly`) | **108 passed** (106 + 2 new) |
| Same two, random ordering | **108 passed** (seed-independent) |
| \+ `test_whatsapp_outbound_media.py` | **150 passed** |
| Full `tests/unit` (`-m "not slow"`, `-p no:randomly`) | **5897 passed / 18 skipped / 0 failed** |
| Mutation: revert either SOURCE level → WARNING | **2 red** on `{30} != {40}`; green on restore |

The 8 `tests/unit/test_1771*.py` property files are excluded from that count — they fail *collection* with `ModuleNotFoundError: No module named 'hypothesis'` on this host's interpreter (system Python 3.14, no venv). No `1771` file is touched by this branch. Note `test_1069_voip_call_path_param.py` **passed** in this run, so the `fastapi` floor-drift red reported in the original body is environment-dependent, not a property of the branch.

### Still open — both yours, neither needs a code change

- **AC #1** — the operator check against a real Twilio account. The reviewer's cheaper signal stands: send a **PDF**; if the placeholder reads `— unsupported format` rather than `— download failed`, the redirect chain provably worked and bytes came back.
- **AC #9** — the corrections to #1315 and #468 item 3d still live only in this PR body. Not posted; say the word and I'll post them.


AndriiPasternak31 (commented):
Ran `/review` + `/validate-pr` + `/cso --diff` against `8946bf70`.

**Verdict: 0 critical, 0 blocking security findings.** Not approving yet — only because AC #1's operator check hasn't run and AC #9 is still open. Both are yours to close; neither needs a code change. Details below, plus one two-word fix I'd take before merge and some vendor evidence that materially de-risks AC #1.

The diff is genuinely clean and I'd rather say so than manufacture objections. The two-tier split is the right call: the credentialed hop's blast radius is byte-for-byte what it was, and only the unauthenticated redirect-target tier widened.

---

## What I verified independently (rather than trusting the PR body)

**Anti-vacuity — reproduced first-hand.** Since `download_file` returns `None` on every path, I re-ran two of your deliberate breaks rather than taking the table on trust:

| Break | Your claim | My result |
|---|---|---|
| B — revert the widening | 12 red across both files | **12 red** ✅ |
| C — re-attach `auth=` to the redirect hop | exactly 2 red | **exactly 2 red** ✅ |

Tree restored clean after each (`git diff --stat` empty). The hop-count assertion discipline does what you say it does.

**35-vector `urlparse` ↔ `httpx.URL` differential corpus → 0 exploitable bypasses.** I scored by parser *disagreement* (matcher reads host X, socket goes to host Y), since that's the only place a real bypass can hide. Three vectors disagree; all three **fail closed** — httpx raises `InvalidURL` or yields an empty host before any connection, and the raise lands inside your `try` → `return None`:

| Vector | `urlparse` host | httpx | Outcome |
|---|---|---|---|
| `https://mms.twiliocdn\t.com/x` | `mms.twiliocdn.com` (allowed) | `InvalidURL` | no request |
| `https://ｍms.twiliocdn.com/x` (full-width m) | `ｍms.twiliocdn.com` (allowed) | `InvalidURL` | no request |
| `␣https://mms.twiliocdn.com/x` | `mms.twiliocdn.com` (allowed) | empty host | no request |

You named the leading-space case. The tab and full-width cases are two you didn't — both land in the same safe direction, and all three are unreachable from a parsed `Location` header anyway.

Correctly refused: prefix-not-suffix, suffix-append, userinfo (`@` and `%40`), backslash authority, CR/LF/NUL, Unicode dot variants (U+3002 / U+FF0E / U+FF61 in *both* positions), Cyrillic homoglyph (IDNA → `mms.xn--twilicdn-rbh.com`), Turkish dotted-İ, trailing-dot FQDN, `http://`, IP literals incl. link-local metadata and `[::1]`, query/fragment decoys, and both S3 forms.

Multi-`Location` re-measured on real `httpx.Headers`: attacker-first → denied; CDN-first → httpx still resolves `mms.twiliocdn.com`, trailing attacker URL degrades to path bytes. No attacker host is ever contacted. Matches your finding.

**The newly-live downstream.** This is the angle I'd flag as most under-examined, and it holds up. The fix resurrects a path dead since #463, so nothing after `download_file` has been exercised in prod for three months. Traced it:

- `upload_service.py:190` — `actual_size = len(data)`, declared size explicitly advisory. This one matters: `whatsapp_adapter.py:327` hardcodes `size=0`, so a declared-size-trusting sink would have had *no* effective size limit. It doesn't.
- `sanitize_filename()` on every branch; `[^\w.\-()]` strips separators, so a malformed `MediaContentType` can't traverse via the synthesized `media_{i}.{ext}`.
- Magic-byte MIME cross-check rejects declared/detected mismatches outside the image↔image and text↔text families.

---

## AC #1 — better than "unverifiable"

You flagged this as resting on one empirical vendor claim. It's actually **vendor-documented**:

> "Requests to fetch your media will redirect you to a secure URL that is only valid for 4 hours… **The domain of the unsecured media URL is `s3-external-1.amazonaws.com` and secured media URL is `mms.twiliocdn.com`.**"
> — [Twilio: Protect Media Access with HTTP Basic Authentication](https://help.twilio.com/articles/223183748-Protect-Media-Access-with-HTTP-Basic-Authentication-for-Programmable-Messaging)

One line confirming both the fix and the S3 exclusion. And the [July 2023 changelog](https://www.twilio.com/en-us/changelog/upcoming-security-changes-enforcing-http-authentication-for-media) adds the part I'd fold into the docs:

> "Newly-created main accounts will have HTTP Basic Authentication enabled **without the option to disable it**." (effective 2023-07-31)

So media-auth isn't just "the supported configuration" — it's **mandatory for every account created in the last three years**. The S3 residual only reaches accounts grandfathered before July 2023. That's a stronger position than the PR body currently claims, and worth stating that way in the user doc.

Also worth noting: a 2016 report of Twilio's chain landing on `media.twiliocdn.com.s3-external-1.amazonaws.com` in **two** redirects — real-world evidence that your bounded budget (rather than a single follow) was the right call. That host is correctly refused by the matcher (prefix, not suffix); I tested it.

Still run the `curl` — the docs establish Twilio's behaviour, not Trinity's end-to-end delivery. **Cheaper alternative:** send a **PDF**. If the placeholder flips from `— download failed` to `— unsupported format`, the redirect chain provably worked and bytes were returned (`upload_service.py:166` vs `:174`). Your PR body frames the PDF case defensively — it's actually the crispest cheap end-to-end signal available.

---

## One thing I'd fix before merge

**The fix leaves its own root cause half-repaired.** #1932's root cause wasn't the allowlist — it was that a fail-closed security gate logged its rejection at WARNING, hiding a 100% outage for three months. This PR raises the redirect gate to `logger.error` (`:764`), but both SOURCE-tier gates stay at WARNING:

```python
:314   if not _is_twilio_media_source_url(media_url):
:315       logger.warning("[WHATSAPP] Rejecting non-Twilio media URL at parse time: %s", ...)

:728   if not _is_twilio_media_source_url(file.url):
:729       logger.warning("[WHATSAPP] Refusing to download non-Twilio media URL (host=%s)", ...)
```

Before the split these shared one constant and couldn't diverge. The two-tier design makes `MediaUrl{N}` **strictly stricter** than the redirect target — a new asymmetry, guarded by the quieter level. If Twilio ever puts a CDN-form URL in `MediaUrl{N}` (exactly the class of vendor change that caused #1932), every attachment fails again at WARNING. Empirical detection latency for that signature: three months.

`8946bf70`'s message says it covers "every media log line" — these two are the exceptions. Two-word change, no behaviour change.

## AC #9 is still open

#1315 is CLOSED with 2 comments (`/claim` + the bot); no correction posted. The AC says "a comment on the closed issue is enough" — the correction currently lives only in a PR body that nobody reading #1315 will encounter, which is the information loss the AC existed to prevent. Same for #468 item 3d.

## On the Slack finding

Confirmed at `slack_service.py:739`. Agree it's out of scope here, and I'd keep the severity honest — the URL comes from an HMAC-verified Slack event and httpx strips `Authorization` cross-origin, so the bot token doesn't travel. But the trend is the argument for filing it:

| Audit | Cited as | Status |
|---|---|---|
| `cso-2026-06-21.md:159` | `slack_service.py:639` | VERIFIED |
| `cso-2026-07-13.md:31` | `slack_service.py:702` | PERSISTENT |
| today | `slack_service.py:739` | **PERSISTENT — 3rd cycle** |

Six weeks, three audits, no tracker entry — and the citation is drifting as the file grows. A finding that lives only in dated report files is losing its own address.

One footnote I found bleakly instructive: the 2026-06-21 report cited the WhatsApp adapter as the **good** example Slack should copy — *"unlike the Telegram and WhatsApp adapters that validate host."* At that moment this allowlist was rejecting 100% of legitimate media. "Allowlist present" scored as correct without anyone checking it admitted the *legitimate* host.

---

## Minor

- **Stale claim, two places.** `feature-flows/whatsapp-integration.md:17` and `requirements/public-access.md:309` both still summarise inbound as "images/audio/PDFs". You corrected the SSRF clause on `:309` but left the media-types clause your own analysis contradicts (`upload_service.py:45-49` rejects `audio/` and `application/pdf`).
- **16 MB cap doesn't bound memory** — `len(resp.content)` runs post-read. Your comment at `:92-96` says this outright so it's honest; noting only that the constant's name reads like a DoS control. Filed follow-up is the right home.
- **Port unconstrained** — confirmed `https://mms.twiliocdn.com:8443/` is accepted. Named as a residual; needs an attacker to already hold a `*.twiliocdn.com` name.
- Confirmed `--skip-agent` was correct here: no `docker/base-image/` file and no new backend module in the diff. And `test_1069_voip_call_path_param` is pre-existing `fastapi>=0.115.0` floor drift — should not be attributed to this PR.

---

**To convert this to an approval:** the AC #1 check (or the PDF placeholder-flip signal) and AC #9's comment on #1315. The F1 log-level change is a strong suggestion, not a gate.

Process surface is complete — all four memory surfaces plus user docs, closing keyword present, full label set, 23/23 CI green across six pytest seeds, and a named regression test whose non-vacuity I confirmed myself.

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

trinity-ability (approved):
/validate-pr: PASS.

- Base `dev` ✅ · 8 files ✅ · `Fixes #1932` closing keyword present, issue auto-promotes ✅
- Security scan clean. The `api.twilio.com@evil.com` / `mms.twiliocdn.com@evil.com` strings are userinfo-spoof **test vectors**, and `_host_matches_suffixes` reads `parsed.hostname`, which resolves them to `evil.com` — correctly rejected ✅
- No new top-level backend module, no new `os.getenv()` ✅
- Docs: requirements + architecture + 2 feature flows + user doc ✅
- Test adequacy: redirect-chain coverage, CDN-shaped spoof vectors, two-tier allowlist pins — non-happy-path, named to #1932 ✅

Design call worth recording: splitting SOURCE (credentialed hop 1, stays `.twilio.com`) from ALLOWED (unauthenticated redirect targets, adds `.twiliocdn.com`) widens only the gate that had the bug, so the tenant AuthToken's blast radius is unchanged. The redirect budget, the transport size cap, and dropping the exception body (the signed CDN URL is a ~4h bearer capability) are all correct.

Approving from trinity-ability — GitHub does not permit self-approval and the author is vybe.

Issue #1870, opened 2026-07-29
bug: completed turn discarded on error_during_execution — jsonl_recovery never consulted, response lost despite end_turn in the JSONL


obasilakis:
## Summary

When Claude Code returns `error_during_execution`, agent-server records the run as failed with an empty response — even when the turn actually **completed successfully** and the finished assistant text is sitting in the persisted JSONL. `jsonl_recovery._recover_response_from_jsonl` exists for exactly this shape of loss but is never consulted on the error path, so a delivered result is thrown away and a scheduled deliverable silently doesn't land.

## Component

Agent Runtime (agent-server: `headless_executor`, `jsonl_recovery`)

## Priority

P1 — completed work is discarded and reported as a failure. On a retrying schedule it burns a second full execution to redo work that already succeeded.

## Error

```
Execution error: [ede_diagnostic] result_type=user last_content_type=n/a stop_reason=null
```

The row is written `status=failed`, `response` empty — while the JSONL for the same session ends with:

```
<ts>  assistant  stop_reason=end_turn  "Q1 complete: A1=0.27, A2=0.28, B=0.27 → median 0.27.
                                        Q2: all three at 0.62 → median 0.62. Waiting for BDI arms A2 and B."
<ts>  user       <task-notification> …          ← background subagent finished, injected after end_turn
<ts>  user       [Request interrupted by user]   ← follow-on turn interrupted
```

The answer is complete and present. Only the trailing pair makes the CLI classify the session as an error.

## Location

- `docker/base-image/agent_server/services/headless_executor.py` — error dispatch in finalize (~L907–935) and the `execution_error` branch (~L1028): writes the terminal failure without attempting recovery.
- `docker/base-image/agent_server/services/jsonl_recovery.py` — `_recover_response_from_jsonl` is wired only for the "stdout dropped the assistant text" case.

## Root Cause

The finalize path treats `metadata.error_type == "execution_error"` as terminal and unrecoverable. But for these runs the JSONL **is** on disk — `event=jsonl_persistence_auto_enabled timeout_seconds=<n> threshold=600` is logged on every execution (#678 auto-persist for `timeout > 10 min`) — and it contains a complete final assistant message with `stop_reason=end_turn`. Nothing reads it, so the completed response is dropped.

## Reproduction Steps

1. Create a schedule whose skill fans out `Agent` subagents (backgrounded by default), with `timeout_seconds > 600` so JSONL persistence auto-enables.
[...]

Issue #1853, opened 2026-07-28
bug: error_during_execution runs store zero telemetry (no execution_log/session_id/cost) and the JSONL is reaped in 1-7h — failure class is undiagnosable


obasilakis:
## Summary

Executions that end in `error_during_execution` are written to `schedule_executions` with **nothing but a status and an error string** — no `execution_log`, no `claude_session_id`, no `cost`, no `duration` telemetry. The Claude Code JSONL that would otherwise carry the evidence is auto-persisted for these runs (#678) but reaped by `session_cleanup_service` within **1–7 hours**, and the row carries no session id to locate it with. Net effect: this failure class is **undiagnosable after the fact**. The only way we obtained evidence was to poll a live instance and `docker cp` the JSONL out during the window.

## Component

Agent Runtime (`headless_executor`) + Backend (`session_cleanup_service`)

## Priority

P2 — no data loss, but it makes a live failure class impossible to root-cause without hand-built capture tooling.

## Evidence

One schedule on an ops-managed instance, all runs since 2026-07-24:

| status | rows | with `execution_log` | avg log size | with `claude_session_id` | with `cost` |
|---|---|---|---|---|---|
| success | 26 | **26** | **573 KB** | 26 | 25 |
| failed / pending_retry | 6 | **0** | 0 | 0 | 0 |

Successful runs store the entire transcript. Every failed run stores nothing — including the session id that would let an operator find the on-disk JSONL before it is swept.

## Why the JSONL doesn't survive either

1. `headless_executor` **does** request persistence for these runs — `event=jsonl_persistence_auto_enabled timeout_seconds=2700 threshold=600` is logged on every execution (#678, `timeout > _JSONL_PERSIST_THRESHOLD_S`). The file is written.
2. `session_cleanup_service` sweeps **every 6h**, deleting any JSONL older than the **1h** age guard whose UUID is not on an active `agent_sessions` row. Headless UUIDs are never inserted there, so effective retention is **1–7 hours**, not the *"reaps these JSONLs after 24h"* stated in `headless_executor`'s own comment (~L410). That comment is stale and should be corrected regardless of what else changes here.

So the artifact exists for a few hours, the DB row can't point at it, and nothing else recorded the stream.

## What this cost, concretely

[...]

Issue #1831, opened 2026-07-27
bug: mid-string 'Z' in agent-supplied fire_at passes validation then 500s in the parser


AndriiPasternak31:
## Summary

`fire_at` is **agent-supplied** (MCP `set_reminder`, `z.string().optional()`). A string containing a `Z` **in the middle** passes Pydantic validation and then raises out of the parser, producing **HTTP 500 where 400 belongs**.

Found by the property `test_P7_validator_acceptance_implies_the_parser_can_parse` in #1771 slice b (PR #1826), shipped there as a `strict=True` xfail with no product fix (that PR's AC forbids product-code change).

## Root cause — two normalizers that disagree

| | Code | Behaviour |
|---|---|---|
| Validator | `src/backend/models.py:2134` — `datetime.fromisoformat(v.replace("Z", "+00:00"))` | replaces **every** `Z` |
| Parser | `src/backend/utils/helpers.py:117-119` — `if timestamp.endswith('Z'): timestamp = timestamp[:-1] + '+00:00'` | strips only a **trailing** `Z` |

So `2026-01-15T10:30Z:00` normalizes, under the validator, to a string `fromisoformat` accepts — but the parser leaves the mid-string `Z` in place and raises `ValueError`.

## Verified reproduction

Measured directly (not inferred), on the interpreters that matter:

| Python | Inputs reproducing |
|---|---|
| **3.13 (prod image, `docker/backend/Dockerfile:1`)** | **4 / 4** |
| 3.12 | 4 / 4 |
| 3.14 | 2 / 4 — `fromisoformat` tightened |

The four inputs:
```
2026-01-15T10Z:30:00
2026-01-15T10:30Z:00
2026-01-15 10Z:30:00
2026-01-15 10:30Z:00
```
On 3.14 only the two `10:30Z:00` shapes still reproduce. The xfail's example set therefore narrows but stays non-empty if Trinity ever moves to 3.14 — the strict xfail keeps holding.

## Why it reaches the client as a 500

- `reminder_service._resolve_fire_at:38-39` calls `parse_iso_timestamp(data.fire_at)` with **no try/except**, under a comment asserting the very invariant this falsifies ("fire_at already validated ISO-8601 by ReminderCreate").
- `routers/reminders.py:120-137` catches `except HTTPException:` **only**, so the bare `ValueError` propagates to FastAPI.

## Amplifier — needs confirmation against a live backend

[...]

PR #1938, opened 2026-08-01
fix(agent-server): never surface Claude Code's [ede_diagnostic] header as the error cause (#1849)

### LINKED ISSUE #262: fix(cli): URL handling in init, human-readable output, version bump checklist
## Problem

Three CLI usability issues:

### 1. `trinity init` fails on bare domains
Entering `trinity.ability.ai` (without `https://`) causes init to fail and exit. Should auto-prepend `https://` and allow retry on connection failure.

### 2. JSON output by default is not user-friendly
CLI operations return raw JSON by default. For interactive terminal use, human-readable text should be the default, with `--format json` available for scripting.

### 3. No version bump checklist/automation after CLI changes
After modifying the CLI, there's no reminder or automation to bump the version in `pyproject.toml` and update the Homebrew formula.

## Acceptance Criteria

- [ ] `trinity init` accepts bare domains, URLs with/without protocol, auto-adds `https://`
- [ ] On connection failure during init, user gets to retry with a different URL instead of hard exit
- [ ] Default output format is human-readable table/text for interactive use
- [ ] `--format json` flag still available for scripting
- [ ] Document version bump checklist (options presented, not implemented yet for #3)

## Files to Change

- `src/cli/trinity_cli/commands/auth.py` — URL normalization + retry loop
- `src/cli/trinity_cli/main.py` — default format detection (isatty)
- `src/cli/trinity_cli/output.py` — ensure table format works well
- `src/cli/trinity_cli/commands/*.py` — audit default format usage

### LINKED ISSUE #1849: bug: #1673's errors[0] picks Claude Code's internal [ede_diagnostic] header — real error text dropped, resume-not-found self-healing defeated
## Summary

#1673 correctly identified that `error_during_execution` carries its text in `msg["errors"]`, not `msg["result"]`, and fixed `stream_parser.py` to read `errors[0]`. On current Claude Code (observed **2.1.215**), `errors[0]` is **never** a real error — the CLI prepends its own internal diagnostic header, `[ede_diagnostic] …`, which it then **filters out of its own user-facing output**. So the surfaced error text is a meaningless marker, and any real error messages (at `errors[1..]`) are dropped.

## Component

Agent Runtime (agent-server: `stream_parser`) — consequence in Backend (`routers/sessions.py`)

## Priority

P1 — it silently regresses the self-healing #1673 shipped, and corrupts the error text of *every* empty-result failure.

## Evidence

**1. The CLI builds `errors` as `[marker, ...realErrors]`** — from the 2.1.215 bundle (`@anthropic-ai/claude-code/bin/claude.exe`), the emitter:

```js
errors: (() => {
  let Lt = lvt(), sn = ls ? Lt.lastIndexOf(ls) + 1 : 0;
  return [`[ede_diagnostic] result_type=${pi} last_content_type=${Er} stop_reason=${tt}`,
          ...Lt.slice(sn).map((Lr) => Lr.error)]   // ← the real errors
})()
```

A second variant exists: `` `[ede_diagnostic] turn aborted (${Sr}) stop_reason=${Ft}` ``.

**2. The CLI treats it as internal and strips it** before producing anything user-visible — if nothing remains after the filter, it reports *nothing*:

```js
let t = e.errors.filter((r) => !r.startsWith("[ede_diagnostic]"));
if (t.length === 0) return null;
return { type: "system", subtype: "informational", content: vi(t.join(", ")), level: "warning", ... }
```

**3. What agent-server does with it** — `stream_parser.py` (post-#1673, ~L298-302):

```python
errors = msg.get("errors", [])
metadata.error_type = "execution_error"
metadata.error_message = (result_text or (errors[0] if errors else "Execution error"))
```

`result_text` is empty on this path (that is #1673's whole premise), so `errors[0]` — the marker — always wins.

**4. Observed in production.** On an ops-managed instance, a scheduled fan-out task recorded four executions whose stored `error` is, verbatim and in full:

```
Execution error: [ede_diagnostic] result_type=user last_content_type=n/a stop_reason=null
```

Agent-server log for one of them:

```
WARNING [Headless Task] Task <id> produced a result but did not exit — finalizing early, terminating teardown
WARNING stream_parser: Claude Code result is_error=true: type=execution_error,
        message=[ede_diagnostic] result_type=user last_content_type=n/a stop_reason=null
ERROR   [Headless Task] Claude Code execution error: [ede_diagnostic] result_type=user last_content_type=n/a stop_reason=null
```

No cause text survives anywhere — the row's `claude_session_id`, `tool_calls`, `cost` and `response` columns are all empty on this path, so the marker is the *only* thing an operator gets.

## Impact

1. **Every** `error_during_execution` failure is now labelled with a marker instead of a cause. Ops error-taxonomy rules bucket it as uncategorized; the string is unsearchable against this codebase because it exists nowhere in it (`git log -S"ede_diagnostic" --all` → empty), which sends investigators to the wrong repo.
2. **Real errors are discarded.** Anything the CLI put at `errors[1..]` never reaches `error_message`.
3. **#1673's own fix is defeated on current CLI** *(inferred from the emitter above — not reproduced against 2.1.215)*. `routers/sessions.py:651` gates the stale-session fallback on `_is_resume_not_found(result.error)`, a substring match for `"no conversation found"` / `"session not found"`. In #1673's repro, `errors == ["No conversation found with session ID: <uuid>"]`. On a CLI that prepends the marker, that text moves to `errors[1]`, `error_message` becomes the marker, the substring match fails, `clear_cached_claude_session_id()` never runs, and the session re-locks into exactly the permanently-broken state #1673 was filed to fix.

## Suggested Fix

Mirror what the CLI does with its own array — skip the diagnostic entries, keep the rest:

```python
errors = [e for e in msg.get("errors", []) if not str(e).startswith("[ede_diagnostic]")]
metadata.error_message = (
    result_text or (", ".join(errors) if errors else "Execution error")
)
```

Joining rather than taking `[0]` also preserves multi-error results. If the diagnostic detail is worth keeping, log it separately at DEBUG — it should not be the value that reaches `schedule_executions.error` or `_is_resume_not_found`.

Worth adding a regression test with an `errors` array whose first element is a `[ede_diagnostic]` marker and whose second is `"No conversation found with session ID: …"`, asserting the resume fallback still fires.

## Notes

The underlying failure that produced our four rows (a parent turn ending with no assistant content mid-fan-out) is a separate question still under investigation on our side; this issue is only about the misattribution of the error text, which is reproducible from the code paths above regardless of what caused the empty result.

vybe:
Fixes #1849

## ⚠️ Rollout constraint — read before merging

**Agent containers need a base-image rebuild to pick this up.** The only product change is in `docker/base-image/agent_server/services/stream_parser.py`, which is baked into `trinity-agent-base`. Merging this PR changes nothing on any already-running agent until `./scripts/deploy/build-base-image.sh` runs and agents are recreated — same constraint as commit `7a179429`. **A revert needs one too.** Please flag this in the release notes.

---

## What's wrong

#1673 correctly found that `error_during_execution` carries its text in `msg["errors"]`, not `msg["result"]`, and fixed `stream_parser.py` to read `errors[0]`. On current Claude Code (observed 2.1.215), `errors[0]` is **never** a real error — the CLI builds the array as `[<marker>, ...realErrors]` where the marker is its own internal diagnostic header:

```
[ede_diagnostic] result_type=user last_content_type=n/a stop_reason=null
```

…and the CLI's own consumer filters every `[ede_diagnostic]`-prefixed entry out before showing anything to a user. A second variant reads `[ede_diagnostic] turn aborted (…) stop_reason=…`. Upstream: `anthropics/claude-code#82235`.

Three consequences, all reproduced end to end (the issue called #3 "inferred, not reproduced" — it is now demonstrated by driving the real parser → the real 502 detail → the real backend predicate and measuring `_is_resume_not_found == False`):

1. **Every** failure on this path was labelled with a marker that exists nowhere in this codebase — unsearchable, so ops taxonomy bucketed it uncategorized and investigators were sent to the wrong repo (closed #262 is the symptom).
2. **Real causes at `errors[1..]` were dropped.**
3. **#1673's own self-healing was defeated.** `routers/sessions.py::_is_resume_not_found` substring-matches `"no conversation found"` against the 502 detail. With the marker in front, the match fails, `clear_cached_claude_session_id()` never runs, and the session re-locks into exactly the permanently-broken state #1673 was filed to fix.

## The fix

Two private helpers in `stream_parser.py`, applied at **both** `errors[0]` reads:

- `_entry_text(entry)` — whitespace-normalized, dict-unwrapping best-effort text for one entry.
- `_split_claude_errors(raw)` — returns `(real_errors, diagnostics)`, mirroring the CLI's own filter (so it is correct on any CLI version — no version gate).

Applied at the `execution_error` branch **and** the pre-existing `max_turns` branch (#361), which reads the same array from the same emitter and has the same defect. The issue names only the first; the second was found during planning. Both now `", ".join(...)` the surviving real errors instead of taking `[0]`, so a multi-error result is preserved.

**Degenerate all-diagnostics case** — the modal production shape; all four observed rows were marker-only. The diagnostic's `key=value` payload is kept as clearly-labelled context, `Claude Code reported no error detail (diagnostic: …)`, with the unsearchable token stripped and the tail capped at `_DIAGNOSTIC_MAX=180`. Nothing else on this path is recorded (`cost`, `tool_calls`, `response`, `claude_session_id` all empty — see #1853), so discarding it would be a net telemetry loss. The fallback literal is deliberately **not** the bare `"Execution error"`, which `headless_executor`'s `f"Execution error: {…}"` would render as `"Execution error: Execution error"`. The word choice is load-bearing: it must trip none of `is_auth_failure` / `_is_resume_not_found` / `_is_rate_limit_message`, and that is test-pinned.

The diagnostic stays in the existing **WARNING** log at both sites — not DEBUG as the issue suggested, because `agent_server/main.py` pins `logging.basicConfig(level=logging.INFO)` and a DEBUG record would emit nothing at all.

### Four latent defects fixed in the same two expressions

- A malformed `errors` shape (dict, int, bool) made `errors[0]` **raise inside the parser**. The raise is swallowed by the per-line `except Exception`, and since `cost_usd`/`duration_ms` are already set the execution fell through to the #160 `context: fork` placeholder → **HTTP 200, `status=success`**. That is #1673's own bug in a new shape.
- A bare-string `errors` was indexed character-by-character (`"boom"` → `'b'`).
- A blank/`None` entry could join to `""`, re-introducing the empty `error_message` bug #1673 fixed. (Note: the issue's own suggested `", ".join(errors)` has exactly this hole.)
- A leading newline slipped a marker past the prefix test. The prefix test is now `casefold()`-insensitive — deliberately more lenient than the CLI's own case-sensitive filter, since no legitimate cause begins with this token in any case.

## Unclaimed security win

`_entry_text`'s whitespace normalization means **neither the joined message nor the diagnostic tail can carry a newline**. Pre-fix, `errors[0]` reached `logger.warning` raw — so this **closes a CRLF log-forging vector** on a path fed by third-party text. Not the reason for the change, but worth recording.

## ⚠️ Behaviour change — SUB-003 exposure, BOTH branches

Real error text now reaches the backend's `is_auth_failure` substring classifier, which the inert marker previously suppressed. So an `execution_error` can now trigger a SUB-003 subscription switch **and**, via the pre-raise `classify_switch_failure`, one same-`execution_id` retry.

**This covers both branches, not just `execution_error`.** The plan asserted the `max_turns` text was "substring-matched by nothing" — **that claim was refuted at review.** `classify_switch_failure` (`task_execution_service.py:408-433`) returns `"auth"` for **any** `>= 400` response body that trips `is_auth_failure` — a **422 included** — and the `except httpx.HTTPError` handler (`:1701`) calls `is_auth_failure(error_msg)` **independent of status code**. So the `max_turns` 422 detail is matched at the same two sites.

In practice the `max_turns` branch stays inert: a `max_turns` result whose `errors[]` carries real entries is **unobserved**, and a marker-only/absent array still yields the fixed `Task stopped after N turns` string. The difference from `execution_error` is the cost shape — a spurious switch there re-runs a turn that legitimately hit its turn limit, so the misfire costs tokens rather than a fast-failing duplicate.

**The misfire rate: `AUTH_INDICATORS` contains bare `"401"`/`"403"`, and a session UUID trips it ~1.73% of the time — ~1 in 58.** Measured 1.727% over 500,000 `uuid4()`s; independently re-measured at review as 1.738% (8688/500,000, ~1 in 57.6), 0.6σ apart. (`uuid4` pins `'4'` at string index 14, which lifts the rate above a naive estimate.)

**This is cost, not incorrectness.** The duplicate turn fails fast (~2s, $0, per #1673 evidence) and the resume fallback still fires — pinned by `test_resume_fallback_survives_a_403_bearing_uuid`, so the claim is tested rather than asserted. The dispatch circuit breaker is unaffected: it counts `error_code == AUTH`, gated on HTTP 503 on the sync path and mapped to `error_code=None` for 502 by `result_callback._STATUS_MAP` on the async #1083 path.

### Recommended follow-up **before merge** (not filed — this is a request to the reviewer)

Anchor the auth indicators with word boundaries — `\b401\b` / `\b403\b` — in:

- `src/backend/services/failure_classifier.py:9` (`AUTH_INDICATORS`), **and** its **byte-identical vendored mirror** `src/scheduler/failure_classifier.py`, which is parity-tested by `tests/unit/test_904_sigkill_no_false_auth.py::TestBackendSchedulerParity` — both copies must change together or the parity test reds.
- Trigger sites to re-check: `task_execution_service.py:1397` and `:1695`.

This was **deliberately sequenced after this PR**, not skipped: it changes auth classification on **every** path in the fleet, a strictly larger blast radius than this P1 parser fix, and hardening first would serialize the whole batch behind a PR that is not one of the claimed issues. Shipping this as a draft means the reviewer sees the recommendation before anything merges.

## Verification

**Tests added:** `tests/unit/test_1849_ede_diagnostic_filtered.py` — 19 test functions / **23 collected cases**.

The acceptance test drives a marker-first `errors[]` through the real parser and `_finalize_headless_result` into the **real** `routers/sessions.py::_is_resume_not_found` (loaded via `spec_from_file_location`, **not** `from routers.sessions import …`, which raises `ImportError` under CI's randomized seeds). A unit test asserting `metadata.error_message == X` would have stayed green through **both** #1673 and #1849 — that is why the pin spans the whole chain.

**TDD red is behavioural, not an ImportError:** pre-fix → **18 failed, 5 passed**, failing on `_is_resume_not_found(detail) is False`, `KeyError: 0`, `TypeError`, `assert 'b' == 'boom'`. Re-proven at review by reverting `stream_parser.py` to `origin/dev` (18 failed / 5 passed), then restoring clean.

**Run in this ship stage** (affected-file neighbourhood only — the full suite is cited, not re-run):

```
pytest tests/unit/test_1849_ede_diagnostic_filtered.py tests/unit/test_1673_execution_error_not_success.py -q -p no:randomly
  → 34 passed

pytest <6-file stream_parser neighbourhood> -q -p no:randomly
  → 123 passed
  (test_1673_*, test_1849_*, test_claude_code_result_recovery, test_claude_code_session_id_parser,
   test_error_classifier_dict_body, test_model_context_catalog)

cd docker/base-image && python3 -c "import agent_server.services.stream_parser"
  → OK
```

**Full suite (WAVE-4 verify-local, cited not re-run):** unit `1 failed, 6054 passed, 18 skipped, 1 xfailed`. Docker stages all passed: build + import-smoke, boot + health, integration `70 passed`.

The single red is **`tests/unit/test_1069_voip_call_path_param.py::TestVoipCallPathParam::test_flat_path_params_are_agent_name_not_name`** and it is **proven pre-existing, not assumed**: a pristine `git archive origin/dev @ 8e924526` was extracted into a temp tree and run with the same venv via the same harness invocation → identical `ImportError: cannot import name 'get_flat_dependant'`. The file is byte-identical across dev and this branch. **Root cause is repo-level and independent of this PR:** `fastapi>=0.115.0` is a floor, not a pin, so any fresh venv now resolves 0.141.1 where `get_flat_dependant` no longer exists. This will red for every contributor building a clean env and deserves its own issue — please do not attribute it here.

### Verification honesty — what was NOT proven

`verify-local` ran with `--skip-agent` **for a host-environment reason, not because the agent stage passed.** The operator's live dev stack owns the global `trinity-agent-network`, and `origin/dev` is in global mode, so the agent precheck hard-refuses. `--skip-agent` removes exactly stage 3 (`import agent_server` inside the built base image) and stage 5 (real agent `/health`) — the two stages that would cover this PR's only product change.

**Compensations, verified rather than asserted:**
- The skipped stage-3 import smoke was **reproduced manually**: `cd docker/base-image && python3 -c "import agent_server.services.stream_parser"` → OK.
- The **#1033 COPY bug class that stage 3 exists to catch is structurally inapplicable here**: this is an `M` on an *existing* file, with **no new imports** and **no new module** under `docker/base-image/`.

**Residual: no real-agent runtime coverage.** Of the PRs in this batch, this is the one that most deserves a real-agent smoke before merge.

## ⚠️ Merge-order warning

`docs/memory/learnings.md` **will conflict** with the PR for #1831. Both patch the **identical hunk** `@@ -230,3 +230,7 @@` of a 232-line file with the same three trailing context lines — both append a new `## 2026-08-01 — pitfall — …` block at EOF. Git's 3-way merge cannot auto-resolve two insertions at one anchor.

**Merge this PR first** (it is the P1); #1831 then rebases and keeps both blocks. Resolution is mechanical — keep both appended sections, either order — but **must be done by hand. Do not let auto-merge attempt this file.**

## Known residuals — pinned and deliberate

- **Truncation budget.** `headless_executor` applies `err[:300]` **after** the join, so a long first error can push a later resume marker past the budget. Pinned by `test_multi_error_truncation_limit_is_known`.
- **That test pins a KNOWN-BAD state.** If a future truncation fix is correct, `test_multi_error_truncation_limit_is_known` / T17 **must be flipped deliberately** — its failure message says outright that a failure means truncation was FIXED, and an added `assert _CLAUDE_MSG not in detail` stops the pin passing vacuously.
- **Partial forward-compat.** `_entry_text`'s dict unwrap only partially defends against a hypothetical nested marker shape. Guessing at a non-existent API is worse than an honest gap.

## Uncommitted by design

`.claude/agents/test-runner.md` (the `/update-tests` catalog) is **intentionally not updated** — `.claude` is a private submodule, so committing it would put a gitlink to an unpushed commit into this PR. The in-repo index `tests/registry.json` **is** updated. Suggested catalog row for whoever lands it in `trinity-dev`:

```
| test_1849_ede_diagnostic_filtered.py | #1849 | agent-runtime, unit, sessions, regression | [ede_diagnostic] header never surfaced as the error cause; both parser branches; resume-not-found chain pinned end to end |
```

## Follow-ups — listed here, deliberately **not filed**

1. **`\b401\b` word-boundary anchoring** in `AUTH_INDICATORS` (details above). Pre-merge recommendation.
2. **Typed `ExecutionMetadata.error_code`** to retire substring matching. The field **already exists and its own comment says it awaits exactly this.** This cross-surface contract has now broken **twice in six weeks with green CI** (#1673, #1849) — the strongest argument in this PR for doing it.
3. **Widen the `err[:300]` budget** in `headless_executor` (or truncate per-error before the join).
4. **The `max_turns` 422 detail has no length cap at all** — belongs in `headless_executor.py`, out of this fence.
5. **Rate-limit misrouting:** a rate-limit message in `errors[]` with an empty `result_text` misroutes to 502, so neither the 429 path nor SUB-003's `rate_limit` switch fires.
6. **Async 502 mislabelled** `terminal_reason="empty_result"`.
7. **Upstream:** `anthropics/claude-code#82235`.

## Out of scope

**#1870 and #1853** are the same error-handling family, but their files overlap open PR #1718 (`headless_executor.py` / `claude_code.py` / `codex_runtime.py` / `subprocess_*`). **This branch touches neither** — `stream_parser.py` is not in #1718's diff, so there is zero conflict surface here.

## Files changed (6)

| File | Why |
|---|---|
| `docker/base-image/agent_server/services/stream_parser.py` | The fix — two helpers, applied at both `errors[0]` reads |
| `tests/unit/test_1849_ede_diagnostic_filtered.py` | New — 19 tests / 23 cases, incl. the end-to-end chain pin |
| `tests/registry.json` | Registry entry for the new file |
| `docs/memory/feature-flows/parallel-headless-execution.md` | Changelog row + corrected the stale `:363-372` snippet, which quoted pre-#1673 code and misattributed it to `claude_code.py` |
| `docs/memory/feature-flows/session-tab.md` | Hazard clause on the resume fallback — it is carried by third-party free text, and has now broken twice |
| `docs/memory/learnings.md` | The bug-class lesson (see merge-order warning) |

Per CLAUDE.md Rule #4 a bug fix needs a descriptive commit message only; `architecture.md` and `requirements/*` were checked and mention neither `result.errors` parsing nor `stream_parser`, so neither needed an edit.


AndriiPasternak31 (commented):
## Review — `/review` + `/validate-pr`, second pass

Read-only against `origin/dev` @ `8e924526`. Head `5472ce27` unchanged since the first pass, so the earlier content analysis stands; this pass re-derives the two decisions the PR routes to the reviewer, plus the merge-order question.

**Verdict: APPROVE-grade diff. No code change requested.** Everything below is process or framing. Posting as a comment rather than an approval only because three mechanical gates are still open — see the checklist.

Re-verified this pass and holding: both `errors[]` read sites fixed (`stream_parser.py:381` and `:411`; a branch-wide grep finds no third — remaining `get("errors")` hits are `cleanup_service.py:2589` stats and `skill_service.py:1261,1449` finalize). Input is credential-sanitized at both producers before the parser sees it (`headless_executor.py:651-654`, `claude_code.py:366-367`) and again at egress (`:1029`), and `_entry_text`'s `" ".join(str(raw).split())` *removes* a CRLF log-forging vector that previously reached `logger.warning` raw. Invariant #5 untouched — `failure_classifier.py` is byte-identical across `src/backend/services/` and `src/scheduler/` on `dev` (verified by `diff`), and correctly left alone here.

---

### Decision 1 — accept the SUB-003 exposure now, or land `\b401\b`/`\b403\b` anchoring first?

**Recommendation: ship now. File the anchoring as its own P2. Don't gate this P1 on it.**

I reproduced the disclosed figure exactly — **1 in 57.6** over 2,000,000 UUID4s (1.69% for the full realistic message). The number is right. Two further measurements change what to do with it.

**The anchoring works in isolation.** Over 300,000 resume-not-found messages it takes the UUID false-positive rate from 1.78% to **exactly zero**, and loses **zero** true positives across 15 realistic auth strings — including tight forms `(401)`, `[401]`, `code:401,`, `status=401`, `got 403.`. Only no-separator forms (`HTTP401`, `error401`) regress, and those aren't real API error shapes. Clean standalone change.

**But it buys much less than the framing implies.** The disclosure scopes the exposure to the UUID sub-case. The actual change is that *all* real error text now reaches `is_auth_failure`'s bare-substring table on this path, where previously only the marker did — and the marker trips nothing. Against 10 realistic `execution_error` texts:

```
                                        plain  anchored
ENOENT ... open '.config/credentials.json'   AUTH   AUTH   (credentials)
Tool 'Bash' failed: ... /etc/forbidden-path  AUTH   AUTH   (forbidden)
net::ERR_ABORTED at .../oauth/callback       AUTH   AUTH   (oauth)
KeyError: 'authentication_mode'              AUTH   AUTH   (authentication)
git push rejected: ... denied (403)          AUTH   AUTH   (403)
Task stopped after 403 turns                 AUTH   AUTH   (403)
AssertionError: expected 403 but got 200     AUTH   AUTH   (403)
MCP server returned 401 during handshake     AUTH   AUTH   (401)
No conversation found ... 8f2401ab-...       AUTH   ok     <- fixed
File not found: .../401k-report.csv          AUTH   ok     <- fixed

false positives before anchoring: 10/10
false positives after  anchoring:  8/10
```

Anchoring removes **2/10** — the *embedded-digit* subset only. Every delimited numeric still matches (correctly, by its own rule) and so does every word indicator. It's a ~20% mitigation of the exposure, not a fix for it. Sequencing a P1 behind that is the wrong trade.

**And the misfire is cost, not incorrectness** — traced rather than assumed. Pre-raise switch (`task_execution_service.py:1397`) retries with the same stale `resume_session_id`, so it deterministically fails identically; the except handler (`:1683-1699`) passes `error_msg` into the `TerminalEnvelope` **unchanged** with `error_code=None` (gated strictly on 503, so the dispatch breaker is untouched); `routers/sessions.py:651` still matches → `clear_cached_claude_session_id` → cold retry → succeeds. `test_resume_fallback_survives_a_403_bearing_uuid`'s docstring claim holds.

The baseline is the clincher: pre-fix this path had a 0% misfire rate *because it was 100% broken*. You can't keep the 0% without keeping the bug.

**Ask:** reword the "~1.73%" disclosure to name the general classifier exposure rather than the UUID sub-case, and file the anchoring with the 2/10 finding attached so it isn't oversold as the fix for it.

### Decision 2 — does the `err[:300]` residual block closing #1849?

**Recommendation: no. Close #1849 on merge; file the truncation as a follow-up.**

1. **Pre-existing and untouched.** `headless_executor.py:1029,1032` applies `err[:300]` on `dev` today, to every 502 detail. This PR changes what enters the budget, never the budget itself. Not a regression it introduces.
2. **Practically unreachable on the path that matters.** Truncation only defeats self-healing when a >283-char real error *precedes* the resume marker in the same filtered array. `--resume` against a missing JSONL fails at startup, before turn work can accumulate other errors — which is exactly why #1673 observed `errors == ["No conversation found with session ID: <uuid>"]`, a single ~60-char entry. Post-filter that's still one entry, well inside 300. `test_multi_error_truncation_limit_is_known` has to synthesize `"A" * 400` to reach the residual at all.
3. **Pinned loudly and correctly.** The known-bad pin states in its docstring that a failure means the residual was *fixed*, gives the exact edit for that moment, and blocks vacuous passing with `assert _CLAUDE_MSG not in detail`.

#1849's impact #3 was itself flagged in the issue as *"inferred from the emitter above — not reproduced against 2.1.215"*. Closing the inferred defect and pinning the narrow residual satisfies it.

---

### Still open

- **Follow-ups: none filed.** Re-checked today — `AUTH_INDICATORS` and `ede_diagnostic` searches return no follow-up issues, and everything created since 2026-08-01 (#1931–#1933) is unrelated. A PR body isn't a tracker; these evaporate at merge. Minimum bar: the `AUTH_INDICATORS` anchoring and the `err[:300]` budget.
- **`docs/memory/feature-flows.md` "Recent Updates" row missing.** Two flows updated, index untouched. Not mandated by the skill (no *new* flow) but it's the convention (#1450, #1445, #1444 all carry one).
- **#1849 carries both `status-in-progress` and `status-ready`** — contradictory, drop the latter.
- **No real-agent runtime coverage.** Release gate, not merge gate — the PR is honest about it. `stream_parser.py` is baked into `trinity-agent-base`; CI never runs it in a container and `verify-local` ran `--skip-agent`. Worth carrying into release notes: nothing changes on a running agent until `build-base-image.sh` + recreate, **and a revert needs one too**.

### Merge order

Confirmed by `git merge-tree`, not inferred: **this PR merges clean into current `dev`.** #1911, #1913 and #1940 all append at the same `docs/memory/learnings.md` EOF hunk (`@@ -230,3 @@`), so whichever lands first, the other three conflict — in *every* ordering, not just this one. The conflict is confined to that single file and both sides are disjoint appends to an append-only ledger, so resolution is **concatenate, keep both** — no semantic decision. Only real hazard is a careless `--ours`/`--theirs` silently dropping an entry; verify after each with:

```bash
git show dev:docs/memory/learnings.md | grep -c '^## 20'
```

Recommend merging this one first — it's the P1 and it's the only one of the four that's currently conflict-free.

### Pre-merge checklist

- [ ] File the follow-ups (min: `AUTH_INDICATORS` anchoring, `err[:300]` budget)
- [ ] Add the `docs/memory/feature-flows.md` "Recent Updates" row
- [ ] Reword the "~1.73%" disclosure to name the general classifier exposure
- [ ] Drop the contradictory `status-ready` label on #1849
- [ ] Real-agent smoke on a rebuilt `trinity-agent-base` before release
- [ ] Second reviewer per SOC 2 (this is a comment, not an approval)

<sub>Full working notes, including the reproduction scripts behind every figure above, are in my local review record. Happy to paste any of them into the thread.</sub>

AndriiPasternak31 (approved):
## Approving — `dbb1d192` (rebased)

My first pass called this **APPROVE-grade, no code change requested**, and nothing in the diff has
changed since. Converting to an approval.

**I rebased this branch onto `dev` myself** (`5472ce27` → `dbb1d192`). That was our doing, not an
oversight of yours: `dev` moved four times today (#1935 `609f8dff`, #1934 `f5a21624`, #1940
`d632c3a5`, #1936 `c4c83f4a`) and #1940 took the `docs/memory/learnings.md` anchor first, so this
branch inherited a conflict it didn't have this morning.

The resolution was mechanical — an append-only ledger, two disjoint entries, keep-both in merge
order. Proof rather than assertion:

- Entry count `59 → 60` (`grep -c '^## 20'`), i.e. exactly one entry added and none lost
- File set identical before and after the rebase (6 files)
- Of those 6, only `learnings.md`'s diff changed; the other five are byte-identical
- All 6 commits still authored by @vybe
- `test_1849_ede_diagnostic_filtered.py` → **23 passed** on the rebased tree

**One gate I am deliberately not closing myself:** the `docs/memory/feature-flows.md` "Recent
Updates" row is still missing (the PR body doesn't claim it either). Writing it would mean
authoring prose about your fix and then approving the PR containing it — the wrong side of SOC 2
separation of duties. Filed separately instead; see the linked follow-up.

**Remaining from the first pass, none blocking, none code:**

- Follow-ups unfiled: `AUTH_INDICATORS` anchoring, the `err[:300]` budget
- The "~1.73%" disclosure still reads narrower than the general classifier exposure
- #1849 carries both `status-in-progress` and `status-ready`
- Real-agent smoke on a rebuilt `trinity-agent-base` before release — release-time, not merge-time,
  but it does need to happen since this ships in the agent image

Recovery handle if the rebase ever needs undoing: `5472ce27`.

vybe:
Rebased onto `dev`. Three files conflicted, all resolved as unions rather than by picking a side:

- **`tests/registry.json`** — rebuilt from the merge stages (dev's 131 entries + this branch's one = 132, no duplicates, valid JSON).
- **`docs/memory/learnings.md`** — this branch's 2026-08-01 entry and dev's 2026-08-02 block both added at the same point. Kept both; the file is oldest-first, so 08-01 precedes 08-02.
- **`docs/memory/feature-flows/parallel-headless-execution.md`** — this branch's #1849 row against dev's #1870 row. That table is **newest-first**, so dev's 2026-08-02 row now sits above this branch's 2026-08-01 one.

`stream_parser.py` merged clean, which was the part worth checking: #1870 landed on dev while this sat, and it touches the same headless result path. Ran both suites together to confirm they compose — **104 passed** (`test_1849_ede_diagnostic_filtered` + `test_1870_completed_turn_recovery`), and **356 passed** across the wider stream-parser / headless / session selection.

Merging on AndriiPasternak31's existing approval.

PR #1943, opened 2026-08-02
feat(mcp-key): detect, self-heal and rotate the per-agent Trinity MCP key (#1854)

### LINKED ISSUE #1854: feature: first-class agent-scoped MCP keys — scope visibility, regeneration, and validation/enforcement of agent-key usage from agent containers
## Summary

Nothing in Trinity validates (let alone enforces) that an agent container talks to the platform MCP server with its **agent-scoped** key. If the trinity entry in an agent's `.mcp.json` carries a **user-scoped** key, the platform accepts it silently — and the agent then operates with the *user's* identity and reach: `list_agents` returns the user's owned/shared agents, and the agent-to-agent permissions matrix configured in the UI is **silently bypassed** (it only governs the agent identity). There is no UI surface that shows which key an agent uses, what scope it has, or that a mismatch exists.

Requested feature, in three parts:

1. **Make the agent-scoped key a first-class entity.** Today it is a row in `mcp_api_keys` (`scope='agent'`, `agent_name` binding) minted invisibly at agent creation. It should be visible on the agent's settings page (prefix, created, last-used) and **regenerable** there — mint new + revoke old + re-inject into the agent's `.mcp.json`. Regeneration matters because the plaintext is unrecoverable (hash-only storage): once the injected value is lost or overwritten, there is no sanctioned recovery path.
2. **System-level validation.** Detect when MCP requests originating from an agent container authenticate with a non-agent-scoped key (or with another agent's key) and surface it: log it, badge it in the agent UI ("this agent authenticates as user X — permissions matrix not in effect"), optionally raise an operator-queue warning.
3. **Optional enforcement.** A setting to reject user-scoped keys presented from inside the agent network, forcing agent containers onto their own keys.

## Component

MCP Server / Backend (mcp_api_keys) / Frontend (agent settings)

## Priority

P1

## Motivating incident (sanitized)

An operator investigation on a production instance: an agent's permissions matrix granted access to 13 agents, but the agent reported only 9 reachable. Root cause — months earlier the agent's owner had copied their laptop MCP config into the agent workspace, overwriting the injected trinity entry with their personal **user-scoped** key. Consequences observed:

- The agent's "visible agents" set was the owner's owned+shared list, not the permitted list; edits to the permissions matrix had no effect and there was no hint why.
- The agent-scoped key sat unused for months with nothing flagging it.
- Privilege escalation: anyone with access to that agent container could act as the owner across every agent the owner could reach on the instance.
- The workspace volume survives rebuilds, so the misconfiguration persisted through container recreation indefinitely; the key is only injected at creation, with no reconciliation.

## Current state (relevant plumbing that already exists)

- `mcp_api_keys` already has `scope` (`user`/`agent`) and `agent_name` columns — the data model is mostly there; the gaps are UI, validation, and enforcement.
- Agent creation already mints and injects the agent key into `.mcp.json`.
- The per-agent auth machinery from #1159 (`AGENT_AUTH_SECRET`-derived per-agent tokens) and the agent-network/platform-network split may provide the request-origin signal enforcement needs (an MCP request from the agent bridge network, or carrying an agent auth context, must present the matching agent key).

## Suggested implementation sketch

1. **Entity/UI**: agent settings → "Platform access key" card: scope, key prefix, created/last-used timestamps, health indicator ("in use" vs "never used / stale — agent may be using a different key"), and a Regenerate action (revoke + mint + re-inject).
2. **Validation**: MCP server resolves the authenticated key's scope; if the request's origin is an agent container (network origin or agent-auth token) and the key is not that agent's key → structured warning log + UI badge + optional notification.
3. **Enforcement**: config flag (default off for back-compat) that turns the validation into a 403.

## Environment

- Observed on a v0.8.0-era production instance (PostgreSQL backend), but the gap is version-independent by design.

## Related

- `src/backend` MCP key handling (`mcp_api_keys`), agent creation key injection
- #1159 (per-agent in-container auth) — potential origin signal for enforcement

trinity-ability:
Fixes #1854

The agent's own `scope='agent'` Trinity MCP key becomes a first-class, inspectable, self-healing, rotatable credential: a **container config-truth probe**, a **start-time drift predicate** that repairs the fleet unattended, and an **owner-driven rotation** that returns metadata only.

---

### 1. ⚠️ UNVERIFIED BY TOOLING — frontend + TypeScript

There is **no `node_modules` on this host**. Therefore **no `tsc`, no `vite build`, no e2e** ran against:

- `src/frontend/src/components/AgentMcpKeyPanel.vue` (new)
- `src/frontend/src/components/settings/SettingsPanel.vue` (2-line wire-in)
- `src/mcp-server/src/types.ts`, `src/mcp-server/src/server.ts`

**Both were reviewed by reading only.** A reviewer with a working toolchain should compile them.

**Mitigation for the TS change specifically:** it is **type-only** — a scope-union widening (`portal_delegate` added to `McpAuthContext["scope"]`), fully erased at runtime. All `.scope` consumers were grepped (37 in `src/mcp-server/src/`; the count including the pass-through sites cited during review is 40) and **every one is an equality test** (`=== "agent"`, `!== "connector"`, …) or an opaque pass-through (`mcp_scope: authContext?.scope`). There is **no `switch` anywhere in `src/mcp-server/src/`**, no `never`-default exhaustiveness check, and no narrowing assignment. Widening the union therefore cannot change a single runtime branch. Re-verified on this branch:

```
$ grep -rn "switch" src/mcp-server/src/          # → no matches
```

The Vue panel is unverified by anything stronger than reading. It mirrors the sibling `McpExposedPanel.vue` pattern and is wired into `SettingsPanel.vue` exactly as its neighbours are.

### 2. VERIFICATION GAPS — disclosed, not softened

- **`verify-local` did not complete.** The **agent-exercise stage did NOT run**, and integration was **not reached** — the harness could not mint an admin token against its freshly-wiped DB. The same failure reproduced identically on a sibling branch, so it is a **host/harness issue, not this change** — but it is still a gap, and it is the gap that matters most here.
- ⇒ **The headline runtime AC — "the regenerated key actually reaches the RUNNING container's `.mcp.json`" — is UNPROVEN LOCALLY.** The `/review` pass already flagged this exact end-to-end check as **REQUIRED before merge**. It is proven only at the unit level (`env_overrides` carrying both `TRINITY_MCP_API_KEY` and `TRINITY_MCP_URL` through `recreate_container_with_updated_config` on a container that had neither, plus a source-order guard that overrides are applied last).
- **Also unproven locally:**
  - the new `agent:mcp_key_regen:{name}` lock against **real Redis** — two overlapping starts of one drifted agent; a start racing an explicit rotation;
  - the `env_absent` / `env_mismatch` GET path against a **live Docker daemon**;
  - the `verify` probe against a **real running agent** container.
- **What DID pass:** backend build + import-smoke, agent-build + import-smoke, boot + health.

### 3. 🔴 SECURITY — please prioritise this PR

Commit `6a968d25` closes a **LIVE, PRE-EXISTING, fleet-wide key-deletion path**. It is **not introduced by this PR**; it is three feet from the code this PR adds, and leaving it open while adding a guard next to it would have been negligent.

- `db/mcp_keys.py` `revoke_mcp_api_key` / `delete_mcp_api_key` **skip the ownership check entirely** for admins — `if user["role"] != "admin":` at lines 536 and 558.
- `dependencies.py` builds an MCP-key principal that carries the **owner's role**.
- ⇒ On a **default admin-owned install** — which is every out-of-the-box Trinity — an agent using its own `TRINITY_MCP_API_KEY` as a Bearer token could `DELETE /api/mcp/keys/{id}` for **any** id: every sibling's key, the platform orchestrator's system-scoped key, and the operator's own. A one-request fleet-wide auth wipe, unrecoverable (no plaintext is retained).
- Separately, `POST /api/agents/{name}/connector/key` — the one route in that file returning a **live key in plaintext** — had no principal guard, so an agent principal could mint and read a **sibling's** connector credential.

Both were **verified against the code**, not inferred. All three routes now run `reject_agent_principal` + `_reject_connector_principal`, the same pair `POST /api/mcp/keys` already applies before minting a `portal_delegate` key, for the same recorded reason (trinity-ops-agent#232). The commit is deliberately standalone and reviewable without reading the feature.

---

## 4. Behaviour change for existing fleets — RELEASE-NOTE ITEM

**The drift predicate repairs a broken agent on its next start, unattended.** `check_agent_mcp_key_matches` joins the eight existing `check_*_matches` predicates in `start_agent_internal`; when it fires, the recreate mints a fresh agent key and bakes both `TRINITY_MCP_API_KEY` and `TRINITY_MCP_URL` (plus `TRINITY_BACKEND_URL`) so the agent-server's own injection actually runs. `trinity-system` and ephemeral ghosts are exempt. Operators upgrading an existing fleet will see agents self-heal on their next start without anyone asking — this is the intended behaviour, but it is a change and belongs in the release notes.

## 5. Design notes

- **No schema change.** Both migration tracks are untouched — re-confirmed on this branch:
  ```
  $ git diff --name-only origin/dev..HEAD -- src/backend/db/schema.py src/backend/db/tables.py \
        src/backend/db/migrations.py src/backend/migrations/versions/
  (empty)
  ```
- **Superseded keys are DELETED, not deactivated.** `recover_agent_ownership` reactivates every inactive per-agent row, so a deactivated superseded key would be *resurrected* by a soft-delete → recover cycle — i.e. a leaked key coming back to life. Proven durable across that cycle in the tests.
- **The automatic self-heal takes the SAME `agent:mcp_key_regen:` lock as explicit rotation.** Capture → mint → DELETE must be atomic **as a unit**: there is no per-agent start lock, so two concurrent starts both observe the drift, and if the second captures after the first's mint but before its delete, it removes the key the first is about to bake in and the surviving container 401s all four readers (heartbeat, #1083 result callback, #1081 pull worker, the runtime's own MCP client). The heal fails **silent and inert** when it cannot hold the lock (nothing minted, nothing deleted, recreate proceeds on existing env, next start retries) — deliberately the opposite of rotation's loud 503/409, because nobody asked for a heal. It still writes an `agent_key_self_heal` audit row: it mints and deletes credentials fleet-wide with no human in the loop, and an INFO log is not an audit trail.
- **`spawned_by_key_id` is re-pointed with a `!= :current` form.** The `= :old_id` form is a **no-op** and cannot repair children stranded by *earlier* rotations. Reconciled **before** delivery, so `enforce_agent_spawn_scope` never 403s at any step of a rotation.
- **Rotation returns metadata only — never plaintext.** The probe emits **only `sha256(bearer)`** per `.mcp.json` entry; the token and the file body never cross the container boundary. A source guard test pins this so no future edit of the in-container script can start leaking one.

## 6. Scope note — why this is wider than a minimal cut

A narrow version would ship a regenerate button attached to a detector that stays **green during the exact incident it exists to catch**. The detector had to be container-truth (not DB-belief) for the feature to mean anything, and the self-heal had to exist for the fleet to actually recover.

**Parts 2b (request-origin attribution) and 3 (enforcement) are NOT included.** There is genuinely **no origin signal** on the MCP→backend hop today: the MCP server forwards none, and client IP cannot substitute (`172.28.0.0/16` ≠ agent container — the frontend sits on `trinity-agent-network` too, and port 8080 is published on all interfaces). They require *building* a signal, not consuming one. This PR **detects** and **repairs**; it does not **prevent**.

## 7. ⚠️ Enforcement caution for whoever picks up Part 3

The scope space is **FIVE**-valued: `user` / `agent` / `system` / `connector` / `portal_delegate` — and `mcp_api_keys.scope` is **free-text with no CHECK constraint**, so it is open at the top. Any enforcement must be an **explicit allowlist**. A denylist ("reject non-agent") would break the **system agent**, the **connector**, and **portal_delegate**. `reject_non_interactive_principal` (added here) is the allowlist shape: it passes only when `User.mcp_scope is None`, i.e. the caller came through the JWT branch.

## 8. #1848 collision: **ZERO lines of `McpKeysTab.vue`**

Disjoint files. Re-verified after the review fixes:

```
$ git diff --name-only origin/dev..HEAD | grep -i McpKeysTab
(no output)
```

## 9. `tests/registry.json` convention note

Pre-existing MCP-key tests are **unregistered**; the tests added here **are** registered. Flagged so a reviewer who prefers the old convention can revert that one hunk without touching anything else.

## 10. Repo-level defect found during this run (NOT caused by this PR)

`tests/unit/test_1069_voip_call_path_param.py` fails in **any fresh venv** with:

```
ImportError: cannot import name 'get_flat_dependant' from 'fastapi.dependencies.utils'
```

The repo declares `fastapi>=0.115.0` — a **floor, not a pin**. This is **invisible in CI** (base and head are both red, so the diff matrix absorbs it) while breaking every fresh local environment. It passes on this host only because the installed FastAPI happens to be compatible.

---

## Verification performed on this branch

| Check | Result |
|---|---|
| `docker/base-image/**` diff | **empty** — see below |
| Schema / both migration tracks | **empty** — see §5 |
| Unit suite | **6037 passed, 18 skipped** in 372s |
| New + adjacent guards | **53 passed** (`test_1854_agent_mcp_key.py`, `test_agent_readiness_probe.py`, `test_1560_agent_redis_key_parity.py`) |
| Invariant guards | **45 passed** (`test_models_centralized.py`, `test_1310_auth_wiring.py`, `test_186_enumeration_uniformity.py`) |
| Secret / email / IP / credential-file scan | clean |
| Build packaging (Invariant, #1033) | ✅ `docker/backend/Dockerfile` uses `COPY src/backend/*.py` glob; all new files land in wholesale-copied subdirs |
| `/validate-pr` | **APPROVE**, 0 fixes required |

4 hypothesis-based property test modules (`test_1771*`) could not be collected — `hypothesis` is not installed on this host. Pre-existing env gap, unrelated to this change.

### 🔑 Zero `docker/base-image/**` — works on already-deployed agents with NO rebuild

```
$ git diff --name-only origin/dev..HEAD -- docker/base-image/
(empty)
```

This is the load-bearing property of the design. Delivery rides a new `env_overrides` kwarg on `recreate_container_with_updated_config`, applied last — so **every already-deployed agent is repairable and rotatable immediately, with no base-image rebuild and no image-version ordering requirement in the release notes.**

---

## Follow-ups (listed here deliberately — **no issues filed**)

1. Narrow `/api/mcp/keys/{id}` revoke + delete further — a `scope='system'` key still passes both new guards (it sets neither `agent_name` nor `connector_agent`).
2. `recreate_missing_container` mints **without** the rotation lock, and calls `deactivate_agent_mcp_keys`, which over-reaches into `scope='connector'`.
3. The heal's Redis dependency — no Redis ⇒ no self-heal that pass; converges on the next start.
4. `create_mcp_api_key` has the same missing `usage_count=0` default that `create_agent_mcp_api_key` needed fixing for.
5. Silent mint failure at agent creation (the `try/except`-swallowed mint) — the root cause the self-heal compensates for, still unfixed at the source.
6. `git rm --cached .mcp.json` on the repair path — a repaired `.mcp.json` can still be tracked by the agent's own repo.
7. Backend REST ignores `agent_permissions` entirely (enforcement lives only in the MCP server layer).
8. `getPermittedAgents` swallows errors → returns `[]`, which reads as "no permissions" rather than "unknown".
9. Two stale feature-flow docs: `mcp-api-keys.md`, `api-keys-page.md`.
10. `architecture.md` router/service counts are stale — it says 63 routers / 66 services; actual is 69 / 98.

---


trinity-ability:
## CodeQL `py/weak-sensitive-data-hashing` (HIGH) — verified FALSE POSITIVE, not dismissed

Verified rather than waved through, since this PR mints and deletes credentials.

**Alert:** `tests/unit/test_1854_agent_mcp_key.py:545` — *"Sensitive data (password) is used in a hashing algorithm (SHA256) that is insecure for password hashing."*

**The flagged code is a test helper:**
```python
def _digest(token: str) -> str:
    return hashlib.sha256(token.encode()).hexdigest()
```

**Why it is a false positive — four independent reasons:**

1. **No new hashing design is introduced.** It mirrors the platform's pre-existing production function, unchanged on `dev`:
   ```python
   # src/backend/db/mcp_keys.py:77-79
   def _hash_api_key(api_key: str) -> str:
       """Hash an API key for secure storage."""
       return hashlib.sha256(api_key.encode()).hexdigest()
   ```
   Every Trinity MCP key already stores its `key_hash` this way. The alert fires on the *test* replicating the existing convention.
2. **The input is not a password.** It is a 256-bit random API key (`trinity_mcp_` + `secrets.token_urlsafe(32)`). CodeQL's heuristic matched the identifier, not the entropy.
3. **SHA-256 is the correct primitive here.** The bcrypt/scrypt/argon2 recommendation exists because low-entropy human-chosen passwords are brute-forceable, so an expensive KDF raises attacker cost. A 256-bit uniformly random token is not brute-forceable at any hash cost; an expensive KDF would add latency to every request and buy nothing.
4. **It is a test file** — not shipped in the runtime image.

**Not dismissed by automation.** Dismissing a HIGH security alert on a public repo is a governance decision left to a maintainer. Recommendation: dismiss as **"Used in tests"** or **"False positive"** citing reasons 1–3.

*Independent note: switching `mcp_api_keys.key_hash` to a slow KDF would be a genuine platform-wide design change well outside this PR's scope, and is not recommended for high-entropy tokens.*

vybe (approved):
Validated via /validate-pr — APPROVE.

CI 25/25 green (all 4 required checks, 3 pytest seeds x base+head). 98 tests pass locally across the new suite and the invariant guards.

Verified independently rather than taken on trust:
- The standalone security commit 6a968d25 addresses a real defect: db/mcp_keys.py:536,558 skip the ownership check entirely for admins, and an agent-scoped key resolves to its owner carrying the owner's role. No agent-side caller of those routes exists, so the guards break nothing.
- Rotation returns no plaintext (checked all four response models), and the in-container probe emits only sha256 of Trinity-pointing bearers.
- Self-heal is capture-before-mint under the same lock as rotation, fail-inert without the lock, audited with metadata only.
- The drift predicate is env-only (no docker exec), fail-safe on error, converges in one pass, exempts trinity-system and ephemeral ghosts.

The disclosed unproven end-to-end AC is structurally bounded: superseded keys are deleted only after the _new_key_is_live post-condition confirms delivery; a failed delivery preserves the old keys and 500s honestly, and the start-time predicate would heal any residual drift.

Non-blocking notes:
- CodeQL comment on tests/unit/test_1854_agent_mcp_key.py:546 is a false positive — hashing a 256-bit random API key, pre-existing platform design (db/mcp_keys.py:67 on dev). The CodeQL required check passed.
- Feature flow uses the decision-narrative structure rather than the canonical template; consistent with recent house style (51/156 flows, incl. the sibling agent-repo-binding.md).
- Release note: existing fleets will self-heal agent MCP keys on next start, unattended.

PR #1948, opened 2026-08-02
feat(credentials): guided per-agent credential setup checklist (trinity-enterprise#127)

AndriiPasternak31:
Refs trinity-enterprise#127

> **Cross-tracker.** trinity-enterprise#127 is a private-tracker issue and this is a public-repo PR, so a closing keyword would not fire — GitHub auto-closes same-repo only. The reference is deliberately `Refs`, and **the issue is closed by hand at the release cut**. A reviewer merging this should also set `status-in-dev` on trinity-enterprise#127 manually, since `issue-status-on-merge.yml` is same-repo too.

> ⚠️ **Depends on #1899** (`feat(templates): the credential declaration standard — ent#128 PR-B`). This branch is **stacked on it**: `5e02fe53` is an ancestor of this head, so the diff below currently contains #1899's 15 commits as well. **Merge #1899 first**; this PR's own diff is the 22 commits from `955c4931` onward (21 files / +3,821). It consumes `normalize_credential_requirements` and the `credential_setup:` declaration that #1899 introduces — without it there is nothing to join live state against.

---

## What this is

The declaration half (ent#128) says what an agent *needs*. This is the other half: **what is actually set, and where to get what isn't.**

`GET /api/agents/{name}/credential-requirements` joins the ent#128 declaration against live in-container state, and an Agent Detail checklist renders **set / missing / where-to-get-it** per variable, writing through the existing owner-gated inject path.

**No DB table, no migration, no feature flag, no new backend write endpoint.**

## The decisions worth reviewing

**Status comes from a fixed base64 `docker exec` probe, not an agent-server endpoint.** So it works on the entire existing fleet the moment the backend updates — no base-image rebuild. AC #3's audience is the ent#123 tokenless seeded fleet, and a checklist inert for every pre-release agent fails its own audience. The probe returns **key names with a non-empty value, never values**, projected backend-side onto the declared set, so an undeclared operator-added variable is never disclosed.

**"Set" is *defined* as agreement with the agent's own exporter** (`agent_server/routers/credentials.py`), pinned by a parity test — the highest-value test here. Nothing is vendored and there is no agent-server mirror, so no Invariant #5 obligation attaches; the parity test is the whole contract.

**`.env` absent ⇒ `missing`, never `unknown`.** A `github:` agent never gets one. `unknown` means "we could not look".

**`degraded` dominates the empty state unconditionally.** A degraded lookup and a genuinely credential-free agent produce an *identical* empty list, and "Ready" is the one state nobody investigates. An **empty** catalog result counts as degraded too — `get_github_template` returns empty-not-`None` on a failed fetch, which is the expected outcome for the tokenless fleet.

**A fourth state, `declaration_incomplete`,** closes the confidently-wrong green: 12 of 25 bundled templates declare `credentials: {}`. Its rows are advisory, never required, never blocking — `.mcp.json.template` does not become a declaration authority.

**Owner-only AND human-only** (`get_owned_agent_by_name` + `reject_agent_principal`). The read gate must equal the write gate it drives. `AuthorizedAgent` resolves an agent-scoped key to the owner *carrying the owner's role*, which would hand an agent the credential inventory of every sibling its owner can access — the whole fleet on a default admin-owned install. The coarse `/credentials/status` keeps the looser gate because it returns a count and names nothing.

**The exec is bounded three ways** (container-side `timeout`, `asyncio.wait_for`, `S_ISREG` before `open()`). `execute_command_in_container` accepts a `timeout` and then never references it again, and the call runs on the shared 4-slot Docker pool (Invariant #11) — four wedged calls stop the backend's entire Docker layer. It is agent-triggerable: the agent owns `/home/developer/.env`, and `mkfifo` there makes `open()` block forever. `compatibility/collector.py` has the identical hole; it is **not** filed on either tracker, and the code comment now says so plainly instead of claiming a follow-up exists.

**`setup_url` renders a UTS-46 nontransitional host with eTLD+1 emphasis, failing closed to inert text.** The stdlib `str.encode("idna")` codec is IDNA2003 and disagrees with browsers on exactly the deviation set that matters (`faß.de` → `fass.de` vs `xn--fa-hia.de` — a *different registrable domain*). This is a mitigation of the schema's inherited IDN residual, not a closure: it does not catch typosquats or a hostile destination behind a correct host.

**Two latent defects in the existing write path are fixed here**, since the checklist writes through it: a transient read failure wiped every configured credential, and the quote round-trip grew one backslash per submit for every other credential in the file.

## Review-round fixes (`b14d23cb`)

Three defects that each **rendered as a pass** — the class this feature's own `degraded`-dominates rule exists to prevent, arriving by other routes:

1. **A caught normalizer raise read as "Ready — needs no credentials."** The narrow `try/except` was a *detector*, not a fix: it swallowed the exception and left `degraded_reason=None`, so the empty list scored green. The covering test passed throughout — it asserted the exception was caught and stopped there. It now asserts the resulting **state**.
2. **A trailing DNS root dot moved the eTLD+1 emphasis off the attacker.** `accounts.google.com.evil.tld.` resolves identically in every browser but adds an empty label, shifting the right-anchored split so the UI bolded `tld.` and dimmed the true registrant. One author-controlled character, inverting the module's primary defence next to a "paste your API key" input.
3. **The single-flight lock released leases it no longer owned.** The TTL is reachable in *normal* operation (`_REQUEST_TIMEOUT` 20s + the catalog fallback's 10s = exactly `_LOCK_TTL_SECONDS`), so the bare `DELETE` freed another worker's lock and let a third caller probe concurrently — silently. Now a random per-acquisition token released by compare-and-delete (#1919).

Plus: the transient 409 no longer blanks a healthy checklist (`v-if="error"` renders *ahead of* `v-else-if="report"`), and the `setup_url` anchor's `title` is the URL, never the author's label.

## Validation

`/validate-pr` run against this branch. One gate failed and is fixed in `b14d23cb`: the flow doc was missing `## Testing` and `## Related Flows` (present in all five sibling flows I compared).

| Gate | Result |
|---|---|
| Base branch | ✅ `dev` |
| Commit messages | ✅ conventional; 23 commits, all authored by @AndriiPasternak31 |
| Requirements | ✅ `requirements/credentials.md` §3.6 |
| Architecture | ✅ service entries, endpoint row, per-variable-status paragraph |
| Feature flow | ✅ `guided-credential-setup.md` (424 lines) — **`## Testing` + `## Related Flows` added by the gate** |
| Flows index | ✅ changelog row + catalog row |
| Security scan | ✅ no keys/tokens/emails/IPs/`.env`/credential files |
| Build packaging | ✅ no new top-level `src/backend/*.py`; `idna==3.10` pinned in the Dockerfile (transitive-pin/#1033 class) |
| Config packaging | ✅ no new `os.getenv` vars |
| Tests | ✅ **208 unit passed** + 3 live-backend route-wiring smoke |
| Conflicts vs `dev` | ✅ 0, and 0 commits behind |

**Not run:** `/verify-local`. This touches no base image, `startup.sh`, migration, or compose service — but the `idna` pin is precisely the class `/verify-local` exists to catch on the *backend* image, so it is worth one run before this leaves draft.

**Not covered by tests, stated rather than implied:** no Playwright/component test exercises the rendered checklist — the frontend contract is source-anchored, and will not catch a regression that keeps the source pattern while breaking the render. No test boots a real agent container; the probe script is executed against fixture directories, not through `docker exec`.


vybe (approved):
/validate-pr: PASS.

- Base `dev` ✅ · 41 files pre-rebase, drops to ~21 now #1899 has landed ✅ · security scan clean (`sk-ent127-must-not-appear` etc. are **leak-assertion fixtures** — the tests assert they do NOT appear) ✅
- Docs: 2 requirements files + architecture + 5 feature flows ✅ · 12 test files ✅
- Dockerfile change justified: `idna==3.10` pinned explicitly because `setup_url_display` needs real UTS-46 and a transitive pin is one resolution away from vanishing — the #1033 class, correctly anticipated ✅

Four things I verified rather than assumed:

1. **The endpoint gate is right, and the reasoning for making it stricter than its sibling is right.** Owner-only + `reject_agent_principal` + rate-limited, where `/credentials/status` gets by on the read gate. The distinction is real: a count discloses nothing, but this names `STRIPE_SECRET_KEY` per agent and says which are populated — a targeting map. Under the read gate an agent's own injected `TRINITY_MCP_API_KEY` would reach it for every sibling its owner can access, which on a default admin-owned install is the whole fleet.
2. **The `docker exec` probe has no injection surface.** `b64` is base64 of a module-level constant, charset `[A-Za-z0-9+/=]`, and nothing caller-controlled is interpolated — `{t}` is a constant too. The container-side `timeout` plus the outer `asyncio.wait_for` are correctly described as bounding different things (pool thread vs request).
3. **`setup_url_display` fails closed and is honest about its limits.** `display_host is None` → inert text, never a fallback to the raw hostname (which would make a failed check byte-identical to a passed one). The IDNA2003-vs-UTS-46 divergence is real and the `faß.de` → `fass.de` vs `xn--fa-hia.de` example is the reason the stdlib codec cannot be used here. Leading with the registrable domain is the right emphasis, since punycode is irrelevant to `accounts.google.com.evil.tld`.
4. **The Vue component actually honours the rendering contract** the model docstring specifies: zero `v-html`, no `utils/markdown.js` import (both mentions are the ban comment), anchor text bound to `setup_url_display_host` and never `title`, and `:type` masking driven by `row.secret !== false`.

**Conflict resolution (mine), pushed as `372a8565`.** #1899 was squash-merged and #1946 landed after it, so 7 conflicts surfaced — all additive. The three source hunks are ent#89 content this branch simply doesn't have; for `static_checks.py` I took dev's merged form (`logger.error(exc_info=True)` + `_fail`) over this branch's inherited `logger.warning`, because that `_fail` is #1899's own HARD-count fix and taking this side would have regressed it. Docs: architecture is a strict superset on dev (ent#128's own bullet byte-identical both sides); all four changelog rows kept in reverse-chronological order; `credentials.md` §3.6 is this branch's alone.

Verified: zero markers tree-wide, all five touched modules compile, and `git diff origin/dev` reports **zero deletions across all seven** conflicted files — the merge is purely additive.

⚠️ Cross-tracker: ent#127 needs `status-in-dev` by hand — the PR body already flags this, and `Refs` rather than `Closes` is deliberate here.

Issue #1926, opened 2026-07-31
bug: fetch failures render as empty states or vanish silently (design-system p15/p25)


vybe:
## Summary

Across ten audited files, fetch and action failures are either swallowed to `console.error` with zero user-visible feedback, or rendered as the *empty* state — telling the user "nothing here yet" when the truth is "the request failed". Principles 15 (loading ≠ empty ≠ failed) and 25 (errors say what happened, what it means, what to do).

## Context

Found by the 2026-07-31 design-system audit (dev@022710dc). Failed-shown-as-empty is actively misleading: it points users at the wrong remedy (e.g. "No templates found" on a network error suggests creating a template, not retrying). Silently swallowed action failures are worse — the verb appears to succeed.

## Acceptance Criteria

- [ ] Every listed surface distinguishes loading / empty / failed visually (distinct treatment, not shared copy)
- [ ] Failed states name what happened and offer a retry action
- [ ] Action failures (toggle, acknowledge, respond, stop) surface an inline error near the failed control — never console-only
- [ ] Empty states gate on "fetch succeeded and returned zero", never on "list currently empty" during first load
- [ ] Error toasts are not used: errors persist until acknowledged (principle 18)

## Technical Notes

Failed rendered as empty:
- `src/frontend/src/components/process/TemplateSelector.vue:235` (fetch error → "No templates found")
- `src/frontend/src/components/TasksPanel.vue:153` (loadExecutions error → "No tasks yet")
- `src/frontend/src/components/InfoPanel.vue:339` (fetch error mapped to `has_template:false` empty visual)

Loading rendered as empty (first-load window):
- `src/frontend/src/views/Agents.vue:673` (gates on `displayAgents.length===0` only)
- `src/frontend/src/views/Operations.vue:138` ("All caught up" before first poll returns; + :180)
- `src/frontend/src/components/operator/NotificationsPanel.vue:268` (blank body during initial fetch; no failed state either)

Failures swallowed to console (no user feedback):
- `src/frontend/src/views/MobileAdmin.vue:663` (+ :676, :861, :872 — toggles, queue respond, acknowledge)
- `src/frontend/src/components/operator/NotificationsPanel.vue:407` (acknowledge/dismiss/bulk)
- `src/frontend/src/components/SchedulesPanel.vue:1271` (toggle; delete/trigger use native `alert()` at :1253, :1290)
- `src/frontend/src/components/SystemViewEditor.vue:296` (tags fetch — row silently vanishes)
[...]

Issue #1266, opened 2026-06-18
feat(ui): lightweight loading states for Dashboard & timeline so the UI isn't frozen during metrics load


vybe:
## Summary

While the Dashboard and timeline metrics are loading (see #1265 for the underlying latency), the UI renders nothing and looks completely frozen/dead — there is no signal that work is in progress. Add **lightweight loading dynamics** (skeleton placeholders / progressive rendering / subtle spinners) so that during the load the interface clearly communicates "loading" instead of appearing broken.

This is the perceived-performance half of the pair: #1265 makes the data arrive faster; this issue makes the wait feel responsive rather than dead.

## Context

- On 10+ agent fleets the initial metrics load can take 20s+ (#1265). Even after that's optimized, any non-trivial load benefits from clear loading feedback.
- Keep it lightweight — skeletons/shimmers over the metric cards, charts, and timeline rows; no heavy spinner overlays that block the whole page.

## Acceptance Criteria

- [ ] Dashboard metric cards/charts show skeleton placeholders (not blank/empty) while their data is loading.
- [ ] Timeline view shows skeleton rows or a progressive load indicator instead of an empty frozen area.
- [ ] Loading states appear immediately on navigation (don't wait for the slow request to resolve) and clear smoothly when data arrives.
- [ ] No layout shift / jank when skeletons are replaced by real content (reserve space).
- [ ] Dark-mode aware and consistent with existing component styling.
- [ ] Error/empty states are distinct from loading states (a failed load doesn't show an infinite skeleton).

## Technical Notes

- Frontend-only; pairs with backend perf work in #1265.
- Reuse/extend existing chart components where possible (e.g. the Overview panel charts) so skeletons match real layout dimensions and avoid shift.
- Loading flags should live in the relevant Pinia stores (e.g. `src/frontend/src/stores/executions.js`) per the store=domain convention, with views rendering skeleton vs. content off those flags.
- Consider a small reusable skeleton component so Dashboard, timeline, and per-agent panels share one implementation.

Issue #596, opened 2026-04-29
Visual regression infrastructure for design-system migrations (#554/#556 follow-up)


oleksandr-korin:
## Summary

The design-system sweep PRs (#554) claim "visual output is unchanged" — currently we trust this based on palette equivalence math (`status-success` aliases `colors.green` directly). Adding **automated visual regression** would convert this claim into a CI-enforced contract.

The Playwright harness (#556 / PR #571) supports `toHaveScreenshot()` out of the box. The blocker is **cross-platform snapshot rendering**: macOS and Linux render fonts and anti-aliasing slightly differently, so baselines captured on a developer's mac fail in CI's Linux runner.

## What's needed

Pick one of the standard fixes:

### Option A — Capture baselines in CI, commit them
- CI workflow has an explicit `--update-snapshots` mode (manual trigger or label-gated)
- Workflow runs Playwright with `--update-snapshots`, commits the resulting PNGs back to the branch
- Subsequent runs check against those committed baselines
- **Pros**: zero local-dev burden, baselines are always Linux-rendered
- **Cons**: requires write-access from CI back to the branch (token + commit step)

### Option B — Capture locally via Playwright Docker image
- Devs run snapshots inside `mcr.microsoft.com/playwright:latest` to match CI's Linux runtime
- Add a `npm run test:e2e:update:linux` script that wraps Docker
- **Pros**: no CI write permissions needed
- **Cons**: extra step for developers; slower locally

### Option C — Platform-specific snapshot directories
- Use Playwright's `snapshotPathTemplate` to keep separate `snapshots-darwin/` and `snapshots-linux/` dirs
- **Pros**: simple
- **Cons**: doubles maintenance; baselines drift between platforms over time

## Recommended approach

**Option A** — capture in CI, commit back. Most teams converge here; it's the lowest ongoing burden. The CI commit step needs:
- A PAT or `GITHUB_TOKEN` with `contents: write` permission
- A guard to only run on PRs with the `ui` label AND a special opt-in (e.g. `update-snapshots` label)
- Commit message convention so the auto-commits are easy to identify

## Specs to capture once infra is in place

The design-system migrations would each get a corresponding visual spec:

- [ ] `/monitoring` — health summary cards + alerts banner (deferred from #595)
- [ ] `/operating-room` — queue list + queue card colors (#569 follow-up)
- [ ] Each future `#554` slice as it lands

## Acceptance

- [ ] One of A/B/C implemented and documented in `e2e/README.md`
[...]

PR #1950, opened 2026-08-03
fix(ui): render failed fetches as failed, not as empty or nothing (#1926)

dolho:
## Problem

Across the ten audited surfaces, a failed fetch was rendered with the **empty** state's copy, and failed actions were swallowed to `console.error`. Failed-shown-as-empty is actively misleading — it points the user at the wrong remedy:

- "No templates found" on a network error → create a template, when the fix is retry
- "All caught up" before the first poll returns → nobody needs you, when we haven't asked yet
- `InfoPanel` synthesized a `has_template: false` payload in its `catch`, turning *"the request failed"* into a claim about the agent

Swallowed action failures are worse — the verb appears to succeed. The row snaps back to its server value with no explanation, and on `MobileAdmin` there is no console to open.

## Two primitives

| Component | For | Behaviour |
|---|---|---|
| `LoadFailed.vue` | a failed **fetch** | owns the surface where the list would be, shares the loading/empty footprint (principle 4), names the problem, offers the retry |
| `InlineError.vue` | a failed **verb** | sits next to the control that was pressed, persists until dismissed (principle 18), optional retry |

Both are token-only and keep the raw error (status code, server message) behind a "Technical detail" disclosure — headline in user vocabulary, trace on demand (principle 25). Both reuse the existing `utils/apiError.js::apiErrorMessage`, so a FastAPI 422 renders as a sentence rather than a stringified Pydantic array.

## The structural fix: `hasLoaded`

`list.length === 0` cannot express "the fetch succeeded and returned zero" — before the first response it means *loading*, after a rejection it means *failed*. That single overload is what produced both the "loading rendered as empty" and "failed rendered as empty" families. So `stores/operatorQueue.js` and `stores/notifications.js` now expose a `hasLoaded` flag that flips **only on success**, and the consuming components gate their empty branch on it. A background refresh failing *after* a good load deliberately keeps the stale list rather than blanking the surface (principle 13).

## Per-surface

**Failed → distinct:** `process/TemplateSelector.vue`, `TasksPanel.vue`, `InfoPanel.vue`.

**Loading → distinct:** `views/Operations.vue` (both queue tabs), `operator/NotificationsPanel.vue` (which had *no* failed state and rendered a blank body during its first fetch).

**Swallowed verbs → visible:** `views/MobileAdmin.vue` (start/stop, autonomy, queue respond, acknowledge), `operator/NotificationsPanel.vue` (acknowledge/dismiss/bulk), `components/SchedulesPanel.vue` (toggle, plus the two blocking `alert()`s on delete/trigger), `components/ExecutionsPanel.vue`, `components/SystemViewEditor.vue`.

Two of these deserve calling out:

- **`ExecutionsPanel` stop** silently redirected to the detail page on failure. An execution that kept running looked like a deliberate hand-off. It now says the execution is still running, keeps the user in place, and offers the detail page as an explicit next step.
- **The bulk helpers use `Promise.allSettled`**, so they *resolve* even when every item failed — a `try/catch` around them reported success on a total failure. The panel now inspects the settlement and names the count that did not apply ("3 of 5 notification(s) couldn't be dismissed and are unchanged").

## Toasts (principle 18)

The listed instance (`views/Agents.vue:8`) is gone, but the behaviour had moved into the shared `composables/useNotification.js`, which auto-dismissed **every** notification after 3s — errors included. An `error` notification now persists until dismissed, and all four toast hosts render a dismiss control (a persistent toast with no close button would be a dead end). The three files carrying hand-rolled copies of the helper now use the composable. The pending timer is also cleared on a new toast — a second notification used to inherit the first one's countdown and could vanish almost immediately.

## Scope note — two listed spots no longer exist

`views/Agents.vue:673` and `views/Agents.vue:8` were audited on 2026-07-31; ent#260 has since retired that page. Its successor `Dashboard.vue` already ships the loading/failed/empty triad (#1266), and `AgentListPanel.vue`'s filtered-empty card correctly gates on `agents.length > 0`. Nothing to fix there; flagged rather than silently dropped.

## Verification

```
$ npm run build
✓ built in 2.58s          # no errors

$ npm run check:tokens
Design-token check OK: 11 tokens equivalent to source palettes; all references resolve

# raw-color ratchet, measured as before-vs-after on exactly the touched files
BEFORE: {raw_nongray: 216, raw_gray: 1349, hardcoded_colors: 134, semantic_tokens: 705}
AFTER : {raw_nongray: 216, raw_gray: 1370, hardcoded_colors: 134, semantic_tokens: 726}
```

`raw_nongray +0` and `hardcoded_colors +0`: the new components are semantic-token-only, and the MobileAdmin banner composes the existing `.result-error` palette instead of introducing two new hexes (it did, in an earlier pass — the scan caught it). `raw_gray` grows because gray is the contract's non-semantic default.

`e2e/honest-failed-states.spec.js` forces each failure with `page.route(...).abort()` and asserts **two-sided** — the failed state appears AND the empty-state copy does not; asserting only the former would still pass if both rendered. It also covers retry-recovers-into-the-real-state and the "still there after 4s" toast/inline-error persistence check.

**Not run locally:** the e2e needs a live stack and this machine's `trinity-backend` / `trinity-frontend` containers are stopped, so the spec is unexecuted — it needs the `ui` label for `frontend-e2e` in CI. Everything else above was run.

## Docs

`docs/memory/design-system.md` gains a "Failed states" recipe section (the primitives, the `hasLoaded` rule, the no-console-only/no-`alert()` rule, and the `allSettled` trap); `design-system-contract.md` updates principles 15 and 18, the primitives list, and adds two PR self-check boxes.

Related to #1926

### DIFF
--- a/docs/memory/design-system-contract.md
+++ b/docs/memory/design-system-contract.md
@@ -19,7 +19,7 @@ Load this before writing any code under `src/frontend/`. It is the condensed, bi
 
 ## Primitives first
 
-- Compose `BaseButton`, `BaseInput`, `BaseSelect`, `BaseToggle`, `BaseTextarea`, `BaseBadge`, `BaseCard`, the modal shell (`ConfirmDialog`), `OverflowTabs`, and the bounded data table. Never hand-roll a lookalike — identical pixels from a class string is still a defect.
+- Compose `BaseButton`, `BaseInput`, `BaseSelect`, `BaseToggle`, `BaseTextarea`, `BaseBadge`, `BaseCard`, the modal shell (`ConfirmDialog`), `OverflowTabs`, the bounded data table, and the failed-state pair `LoadFailed` (failed fetch) / `InlineError` (failed verb). Never hand-roll a lookalike — identical pixels from a class string is still a defect.
 - BaseButton: 4 variants (primary/secondary/danger/ghost) × 2 sizes (md 13.5 pad 7×14 · sm 12.5 pad 4×10), radius 6px. Disabled = opacity .45. In-flight = 16px spinner + progressive label. Focus ring on all variants.
 - BaseInput/BaseSelect: field bg, border-strong, radius 6px, pad 8×11; focus = accent border + 3px ring; errors name the problem, the fix, and an example — never a bare red border.
 - BaseTextarea: min-height 84px, `resize: vertical` only, mono variant for prompts/config, same focus/error as inputs.
@@ -69,10 +69,10 @@ Data loading:
 14. Loading = "no data yet"; cache hits skip animation; reduced motion honored.
 
 Honest state:
-15. Loading ≠ empty ≠ failed ≠ partial — visually distinct, never optimistic success.
+15. Loading ≠ empty ≠ failed ≠ partial — visually distinct, never optimistic success. An empty state requires a fetch that **succeeded and returned zero** (`store.hasLoaded`), never `list.length === 0`; a failed fetch renders `LoadFailed`, never the empty copy (#1926).
 16. No dead ends: empty states name the purpose and the next action.
 17. Validation errors are named, actionable, with an example; pre-validate client-side where known.
-18. Verbs acknowledge in ~100ms and confirm completion; toasts for completed verbs, never errors.
+18. Verbs acknowledge in ~100ms and confirm completion; toasts for completed verbs, never errors. A failed verb surfaces an `InlineError` next to its control and persists until dismissed — never `console.error` alone, never `alert()` (#1926).
 19. Destructive actions: named verb, restated consequence, safe action focused first.
 22. Times are honest: relative for recency, absolute + timezone on hover/detail.
 23. Keyboard baseline: visible focus, Esc closes overlays, modal focus trap, tab order = visual order.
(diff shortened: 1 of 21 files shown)


dolho:
CI verification of the new e2e spec (the one thing I couldn't run locally — this machine's backend/frontend containers are stopped):

The `e2e` job runs `--grep @smoke` and uses the `github` reporter, which prints only failures — so a passing spec name never appears in the log. Comparing totals against a PR without this spec:

| run | result |
|---|---|
| #1949 (no this spec) | 20 passed · 7 skipped |
| this PR | **21 passed** · 7 skipped |

The +1 is `honest-failed-states.spec.js`'s `@smoke` test — operator-queue fetch aborted, failed state asserted visible, "All caught up" asserted absent — run against the live stack. The other three tests in the file are untagged and therefore local-only, matching the repo's \@smoke-only CI policy (#596).

The `1 flaky` and the cancelled `report` job are both pre-existing/unrelated: the flake is `dashboard-type-filter.spec.js` (also flaky on #1949), and `report` was cancelled by the concurrency group when adding the `ui` label re-triggered the workflow.

dolho:
## Before / after, captured against a live stack

📷 **[Screenshots — PR #1950](https://claude.ai/code/artifact/48789d3a-85d3-4105-b313-09b608bd1d83)**

Each pair is the *same forced failure* on `origin/dev` (left) and this branch (right), both running the real app against the same backend with the named request rejected at the network boundary (`page.route(…).abort()`). Nothing is mocked but the failure itself.

Covered: operator queue, notifications, task history, schedule toggle, plus the two states that had no picture at all before — the failed-refresh strip and the dark theme.

The sharpest one is task history: on `dev` a rejected fetch renders **"No tasks yet — run a task above"** on an agent whose own header reads **24 runs**.

Two more commits since the review pass, both from doing this:
- `2dd73ec1` — the three findings from `/review` (failed-refresh honesty, `LoadFailed`'s `onRetry` prop colliding with the name Vue compiles `@retry` into, stale error left over between schedule verbs)
- `6685adc7` — the retry now sits above the technical-detail disclosure; the screenshot made the wrong order obvious

dolho:
Screenshot from 2026-08-03 12-06-13
Screenshot from 2026-08-03 12-06-07
Screenshot from 2026-08-03 12-05-58
Screenshot from 2026-08-03 12-05-50
Screenshot from 2026-08-03 12-05-42

vybe (approved):
/validate-pr: PASS.

- Base `dev` ✅ · 21 files ✅ · commits conventional ✅
- Security scan clean: no keys/tokens, no real emails, no public IPs, no .env, no credential files ✅
- No new top-level backend module, no new `os.getenv()` — no packaging gap ✅
- Scope is tight: every change traces to #1926, comments cite the principle they enforce
- Test surface: `e2e/honest-failed-states.spec.js` + `ui` label present so `frontend-e2e` runs in CI ✅

⚠️ Non-blocking: the reference to #1926 is bare, not a closing keyword — `issue-status-on-merge.yml` will NOT promote it to `status-in-dev`. Bumping the label by hand at merge.

PR #1954, opened 2026-08-03
fix(ui): move the Agent Detail toast off the header controls (#1953)

### LINKED ISSUE #1953: fix(ui): Agent Detail toast overlaps the agent header controls (Running toggle + Delete)
Saving any setting on Agent Detail pops a toast that lands directly on top of the agent header's top-right control cluster — covering the **Running** toggle and the **Delete** button until it fades.

Reproduced on `v0.8.5` (`223b5570`), dark mode, agent `aero-scout`.

## Repro

1. Open an agent → **More → Settings**
2. Change **Max turns (task)** under Guardrails → **Save**
3. The "Guardrails saved — restart agent to apply" toast renders over the Running toggle and the trash icon

Not guardrails-specific — every `notify()` call in the view does this: rename, autonomy, read-only, subscription change, tag edits, etc.

## Root cause

The toast and the header controls resolve to the same coordinates, and the toast wins on `z-index`.

- `src/frontend/src/views/AgentDetail.vue:18` — toast is `fixed top-20 right-4 z-50` → **80px** from the viewport top, flush right.
- The header's row-1 right cluster (Workspace / `RunningStateToggle` / Delete) also starts at ~80px: `NavBar` `h-16` (64px) + `<main>` `py-2` (8px) + inner `py-2` (8px) = 80px, and `AgentHeader` is the first thing after it.

So `top-20` is exactly the header's first row. Nothing about it is width-dependent — it collides at every viewport size.

## Why not just move it to `bottom-4 right-4`

That's the other convention already in the codebase (`PlaybooksPanel.vue:162`, `PublicLinksPanel.vue:350`), but it trades one overlap for another: `HelpChatWidget` is a **global** 56px FAB at `fixed bottom-6 right-6 z-50` (`App.vue:12`), so `bottom-4 right-4` lands underneath it.

The bottom-right gutter *above* the FAB is the only region on this view with no fixed/absolute competition.

## Scope

- `views/AgentDetail.vue:18` — the reported toast.
- `components/FilesPanel.vue:301` — byte-identical `fixed top-20 right-4 z-50` toast, rendered **inside** Agent Detail (Files tab), so it carries the same collision. Fixed together.

Out of scope, deliberately:

- `views/Agents.vue:10` uses the same `top-20 right-4`, but that view has no header controls there — not a collision, left alone.
- `PlaybooksPanel.vue` / `PublicLinksPanel.vue` at `bottom-4 right-4` overlap the Help FAB. Pre-existing, unreported, separate fix.
- Extracting a shared `<Toast>` component is the real cure for five hand-rolled copies drifting apart, but that is a refactor, not this bug.

## Acceptance criteria

1. Saving guardrails on Agent Detail leaves the Running toggle and Delete button fully visible and clickable.
2. The toast does not overlap the Help chat FAB.
3. Same holds for the Files tab toast.

obasilakis:
Closes #1953

## The bug

Saving anything on Agent Detail popped a toast directly on top of the header's top-right control cluster — the **Running** toggle and the **Delete** button — until it faded.

Not guardrails-specific: every `notify()` on the view (rename, autonomy, read-only, subscription change, tag edits) hit it.

## Why

Both elements resolve to the same coordinates; the toast wins on `z-index`.

- Toast: `fixed top-20 right-4 z-50` → **80px** from the viewport top, flush right.
- Header row-1 right cluster also starts at **80px**: `NavBar` `h-16` (64) + `<main>` `py-2` (8) + inner `py-2` (8), with `AgentHeader` immediately after.

`top-20` *is* the header's first row. Not width-dependent — it collided at every viewport size.

## The fix

Anchor the toast bottom-right: `fixed bottom-24 right-6 z-50`.

Deliberately **not** `bottom-4 right-4` (the other convention in this codebase, `PlaybooksPanel.vue:162` / `PublicLinksPanel.vue:350`) — `HelpChatWidget` is a **global** 56px FAB at `bottom-6 right-6 z-50` (`App.vue:12`), so that would trade one overlap for another. `bottom-24` puts the toast in the FAB's own right gutter with a 16px gap above it, which is the only region on this view with no fixed/absolute competition.

`FilesPanel.vue` carries a byte-identical toast and renders **inside** Agent Detail, so it had the same collision and moves with it.

## Left alone, deliberately

- `Agents.vue:10` and `AgentListPanel.vue:17` use the same `top-20 right-4`, but those views have no header controls there — no collision.
- `PlaybooksPanel` / `PublicLinksPanel` at `bottom-4 right-4` overlap the Help FAB. Pre-existing, unreported, separate fix.
- Extracting a shared `<Toast>` component is the real cure for five hand-rolled copies drifting apart — a refactor, not this bug.

## Verification

```
$ npm run build
✓ built in 2.43s

$ grep -ho "\.bottom-24{[^}]*}\|\.right-6{[^}]*}" dist/assets/*.css | sort -u
.bottom-24{bottom:6rem}
.right-6{right:1.5rem}
```

Tailwind emits both utilities into the bundle (they're literal strings in the `:class` array, so JIT picks them up). Geometry: toast bottom edge at 96px, FAB top edge at 24 + 56 = 80px → 16px clearance.

Frontend-only, two class strings plus comments. No backend, no schema, no API change.

### DIFF
--- a/src/frontend/src/components/FilesPanel.vue
+++ b/src/frontend/src/components/FilesPanel.vue
@@ -295,10 +295,12 @@
       </div>
     </div>
 
-    <!-- Notification Toast -->
+    <!-- Notification Toast. #1953: same anchor as the AgentDetail toast — this
+         panel renders inside Agent Detail, so `top-20 right-4` collided with the
+         same header controls. See views/AgentDetail.vue for the full rationale. -->
     <div v-if="notification"
       :class="[
-        'fixed top-20 right-4 z-50 px-4 py-3 rounded-lg shadow-lg transition-all duration-300',
+        'fixed bottom-24 right-6 z-50 px-4 py-3 rounded-lg shadow-lg transition-all duration-300',
         notification.type === 'success' ? 'bg-status-success-100 dark:bg-status-success-900/50 border border-status-success-400 dark:border-status-success-700 text-status-success-700 dark:text-status-success-300' : 'bg-status-danger-100 dark:bg-status-danger-900/50 border border-status-danger-400 dark:border-status-danger-700 text-status-danger-700 dark:text-status-danger-300'
       ]"
     >
--- a/src/frontend/src/views/AgentDetail.vue
+++ b/src/frontend/src/views/AgentDetail.vue
@@ -12,10 +12,15 @@
           <div class="animate-spin rounded-full h-12 w-12 border-b-2 border-action-primary-500 mx-auto"></div>
         </div>
 
-        <!-- Notification Toast -->
+        <!-- Notification Toast.
+             #1953: anchored bottom-right, NOT `top-20 right-4`. `top-20` is exactly
+             where AgentHeader's row-1 right cluster starts (NavBar h-16 + main py-2 +
+             inner py-2 = 80px), so the toast covered the Running toggle and Delete.
+             `bottom-24` clears the global HelpChatWidget FAB (bottom-6 + h-14 = 80px),
+             which is why this is not the `bottom-4 right-4` used elsewhere. -->
         <div v-if="notification"
           :class="[
-            'fixed top-20 right-4 z-50 px-4 py-3 rounded-lg shadow-lg transition-all duration-300',
+            'fixed bottom-24 right-6 z-50 px-4 py-3 rounded-lg shadow-lg transition-all duration-300',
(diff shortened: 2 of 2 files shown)


vybe (approved):
Validated via /validate-pr. Frontend-only (2 files, +11/-4), `Closes #1953`, no secrets/IPs/env changes, no schema or API surface. Geometry reasoning checks out — `top-20` is literally the header's first row (64+8+8), and `bottom-24` clears the global 56px Help FAB at `bottom-6` with 16px to spare rather than trading one overlap for another. FilesPanel's byte-identical toast moved with it. Approving.

PR #1955, opened 2026-08-03
fix(slack): gate inbound file downloads behind a host allowlist (#1951)

### LINKED ISSUE #1951: fix(slack): download_file has no SSRF allowlist and blind-follows redirects
## Summary

`SlackService.download_file` (`src/backend/services/slack_service.py:739`) fetches an inbound-event URL with **no host allowlist and no scheme check**, and with `follow_redirects=True` (`:755`) — so redirects are blind-followed to any host. The two sibling channels both validate:

| Channel | Host allowlist | Redirects |
|---|---|---|
| Telegram (`telegram_media.py:29,74`) | `api.telegram.org`, exact match | n/a |
| WhatsApp (`whatsapp_adapter.py`, #1932) | two-tier suffix allowlist | `follow_redirects=False`, each hop re-validated |
| **Slack** (`slack_service.py:739`) | **none** | **`follow_redirects=True`** |

Same bug class as #1932, different channel.

## Context

Surfaced while reviewing #1932 (PR #1937). This is the **third consecutive CSO audit** to report it, and the citation has been drifting as the file grows — a finding whose only address is a dated report file is losing that address:

| Audit | Cited location | Status |
|---|---|---|
| `docs/security-reports/cso-2026-06-21.md:159` | `slack_service.py:639` | VERIFIED |
| `docs/security-reports/cso-2026-07-13.md:31` | `slack_service.py:702` | MITIGATED · PERSISTENT |
| PR #1937 review (2026-08-02) | `slack_service.py:739` | PERSISTENT — 3rd cycle |

Six weeks, three audits, no tracker entry. This issue is that entry.

### Severity is deliberately P2, not higher — stated honestly

Two real mitigations, both verified rather than assumed:

1. **The URL is not attacker-supplied from outside.** It comes from `url_private_download` on a Slack event (`slack_adapter.py:557`), and `slack_webhook.py` HMAC-verifies `X-Slack-Signature` against the signing secret before the event is parsed. The trust boundary is Slack itself.
2. **The bot token does not travel.** `httpx` strips the `Authorization` header on cross-origin redirects, so the `Bearer {bot_token}` set at `:754` is dropped before any off-origin hop.

What remains is a genuine defense-in-depth gap rather than a live exploit:

- **The fetch is issued from the backend**, which is on **both** Docker networks (`trinity-platform-network` and `trinity-agent-network`), so an arbitrary-host fetch from here reaches internal services that agents deliberately cannot route to.
- **It is non-blind.** The response bytes are returned to `message_router` and written into the agent workspace as a file, so fetched content lands somewhere a user reads it — a stronger primitive than a blind SSRF.
- **No scheme check**, so `http://` is accepted alongside `https://`.
- The mitigation is entirely *transitive* — it holds only as long as Slack's field stays trustworthy and httpx keeps stripping the header. Neither is a property of Trinity's own code, which is exactly the argument the #1932 two-tier split was built on.

## Acceptance Criteria

- [ ] `SlackService.download_file` validates the URL host against an allowlist before any request is issued, and refuses a non-`https` scheme
- [ ] Redirects are no longer blind-followed: either `follow_redirects=False` with each hop re-validated against the allowlist before it is issued (the #1932 shape), or an equivalent bounded, per-hop-validated follow
- [ ] A refusal logs at **ERROR**, not WARNING — a fail-closed media gate that goes quiet is the root cause of #1932, where a 100% outage hid for three months
- [ ] Regression tests assert the **recorded call log / hop count**, not just `is None` — `download_file` returns `None` on every failure path, so a negative test asserting only the return value passes against a completely unpatched seam (see `tests/unit/test_whatsapp_inbound_media.py` for the pattern)
- [ ] A legitimate Slack file download still succeeds end-to-end (the #1932 lesson: an allowlist that rejects the *legitimate* host scores as "allowlist present" in an audit while being 100% broken)

## Technical Notes

**The eventual shape is one shared helper, not a third bespoke allowlist.** Three channels now implement the same fetch-with-allowlist logic three different ways, one of which is missing. A `channel_media_fetch(url, allowlist, auth, max_bytes)` covering WhatsApp / Telegram / Slack uniformly would collapse them and make the next channel correct by construction. Whether to converge now or fix Slack in place first is an implementation call.

Two adjacent gaps worth folding in if the shared helper lands (both currently apply to WhatsApp too, tracked as accepted residuals on #1932):

- **No private-IP / DNS-rebinding resolution check** — every allowlist here is name-based, so an allowlisted hostname resolving to a private address is not caught. This is the control that actually closes the internal-network reach described above.
- **Size caps bound what is accepted, not what is buffered** — `len(response.content)` runs after the body is already resident (`slack_service.py:760`). A true streaming bound is the fix.

Relevant files:
- `src/backend/services/slack_service.py` (`download_file`, `:739`)
- `src/backend/adapters/slack_adapter.py` (`:228`, `:557` — caller and URL source)
- `src/backend/services/telegram_media.py` (`:29`, `:74` — the exact-match precedent)
- `src/backend/adapters/whatsapp_adapter.py` (the #1932 two-tier precedent)

dolho:
Closes #1951

## Problem

`SlackService.download_file` fetched `url_private_download` with **no host allowlist, no scheme check, and `follow_redirects=True`** — redirects blind-followed to any host. Both siblings already validate; Slack was the one that never did, across three consecutive CSO audits.

| Channel | Host allowlist | Redirects |
|---|---|---|
| Telegram | `api.telegram.org`, exact | n/a |
| WhatsApp (#1932) | two-tier suffix | `follow_redirects=False`, each hop re-validated |
| **Slack (before)** | **none** | **`follow_redirects=True`** |

## Why it mattered despite two mitigations

The issue is honest that the URL is HMAC-verified Slack input and that httpx strips `Authorization` cross-origin. Both hold — and both are **transitive**: one trusts Slack's field, the other trusts httpx's behaviour. Neither is a property of Trinity's code. What *is* ours:

- the fetch is issued from the backend, which is on **both** Docker networks — so an arbitrary-host fetch reaches internal services agents deliberately cannot route to
- the bytes are returned to `message_router` and written into an agent workspace, so this is a **non-blind** primitive, stronger than a blind SSRF

## Fix — the #1932 shape, not a third bespoke allowlist

```python
_SLACK_FILE_SOURCE_HOST_SUFFIXES  = (".slack.com",)                                    # hop 1: carries the bot token
_SLACK_FILE_ALLOWED_HOST_SUFFIXES = (".slack.com", ".slack-edge.com", ".slack-files.com")  # validated 30x targets, no token
_MAX_FILE_REDIRECTS = 3
```

`follow_redirects=False`, every hop re-validated **before it is issued**, https-only, refusals at **ERROR**.

**`*.slack.com` rather than an exact `files.slack.com`, deliberately.** #1932's real lesson is that an allowlist which rejects the *legitimate* host scores as "allowlist present" in an audit while being 100% broken — WhatsApp media was dead for three months that way. A Slack-apex suffix can't break on a `files-N.slack.com` variant and still confines the token to Slack. If Slack ever redirects somewhere outside the tiers, it fails **loudly at ERROR** with the host named, and the fix is one allowlist line.

## Verification

Before/after against the pre-fix module through the same fake transport:

```
BEFORE (origin/dev)  https://169.254.169.254/latest/meta-data/  requests=1  follow_redirects=True  bytes=b'INTERNAL-RESPONSE'
BEFORE (origin/dev)  http://172.29.0.2:5432/                    requests=1  follow_redirects=True  bytes=b'INTERNAL-RESPONSE'
BEFORE (origin/dev)  https://evil.example.com/x                 requests=1  follow_redirects=True  bytes=b'INTERNAL-RESPONSE'

AFTER  (this branch) all three                                  requests=0                         bytes=None
```

`172.29.0.2:5432` is this machine's actual `trinity-postgres` on the platform network — the reach the issue describes, demonstrated rather than asserted.

- **26 new tests**, plus all 355 existing Slack unit tests green
- **Every negative test asserts the recorded call log**, never `is None` alone — `download_file` returns `None` on every failure path and its bare `except Exception` swallows `AssertionError`, so a return-value assertion passes against a *completely unpatched* seam (the AC's point, and `test_whatsapp_inbound_media.py`'s documented lesson)
- The legitimate `files.slack.com` download still succeeds end-to-end, and the CDN-redirect test asserts the second hop carries **no** `Authorization`

One test earns its place specially: `evil-slack-files.com` passes a dotless `endswith("slack-files.com")` check. Every allowlist entry must therefore start with a dot, pinned by `test_every_allowlist_entry_starts_with_a_dot`.

## Acceptance criteria

- [x] Host allowlist validated before any request; non-`https` refused
- [x] No blind redirect following — `follow_redirects=False`, per-hop re-validation, bounded budget
- [x] Refusals log at ERROR
- [x] Regression tests assert the call log / hop count
- [x] A legitimate Slack download still succeeds

## Deliberately not folded in

The shared `channel_media_fetch` helper (would touch two channels that were just fixed — better as its own change with all three test suites as the safety net), the private-IP/DNS-rebinding resolution check, and a true streaming size bound. The latter two are accepted residuals on #1932 and belong with that helper; the size cap here stays honest about bounding what is *accepted*, not what is *buffered*, exactly like both siblings.

Related to #1951


vybe (approved):
Validated via /validate-pr. Closes #1951 (closing keyword added so the issue auto-promotes).

Security fix, no schema/API/config surface. The two flagged strings in my scan are test fixtures (`xoxb-test-token`, `169.254.169.254` as a redirect *target*), not credentials.

What I checked and agreed with:

- **Reuses the #1932 two-tier shape rather than inventing a third bespoke allowlist** — source tier (`.slack.com`, carries the bot token) strictly narrower than the redirect tier, `follow_redirects=False`, every hop re-validated *before* it is issued, bounded budget, https-only.
- **`*.slack.com` over an exact `files.slack.com` is the right call.** #1932's actual lesson is that an allowlist rejecting the legitimate host scores as "allowlist present" in an audit while being 100% broken — WhatsApp media was dead three months that way. A too-narrow list here fails silently; a too-broad-within-Slack one fails loudly at ERROR with the host named.
- **Every negative test asserts the recorded call log, never `is None`.** `download_file` returns `None` on every failure path and its bare `except Exception` swallows `AssertionError`, so a return-value assertion passes against a *completely unpatched* seam. This is the vacuous-negative class from #1932/#1917, and it's handled.
- `test_every_allowlist_entry_starts_with_a_dot` pins the `evil-slack-files.com` dotless-suffix bypass — the kind of guard that stops the next entry from reintroducing it.

The before/after demonstration against the real `trinity-postgres` on the platform network makes the reachability argument concrete rather than theoretical.

Merged `dev` and resolved the `tests/registry.json` append conflict (both entries kept; JSON re-parsed, no new duplicates, touched-file set unchanged).

PR #1957, opened 2026-08-03
test: reconcile ent#89 tests with ent#128's compatibility contract

trinity-ability:
## What

Four unit tests are failing on `dev` right now. This fixes them.

They are not a defect in either PR that produced them: **#1899 (ent#128) and #1946 (ent#89) were each green in isolation**, and the failures only exist in their combination. Nothing in the production code is wrong — the four assertions describe a contract that #1899 deliberately changed, and #1946 was written before that landed.

## Why it reached `dev`

`dev`'s required checks are `Analyze (python)`, `Analyze (javascript-typescript)`, `schema-parity` and `verify-non-root`. **None of them runs a unit test.** The `regression diff` job does, but it reports after the required set goes green — so auto-merge had already landed #1946 by the time the failures were visible.

`regression diff` on the merge commit reported exactly these four, against a base with zero failures:

```
## ❌ New failures introduced by HEAD (4)
- [F] test_compatibility_checks.TestReportDirectionOnAFailingValidator::test_persisted_check_error_does_not_replay_as_clean
- [F] test_compatibility_checks.TestRunStaticLogsItsSwallow::test_a_raising_check_is_logged
- [F] test_ent89_template_schedules.TestCatalogSurface::test_a_malformed_block_does_not_raise_out_of_either_builder
- [F] test_ent89_template_schedules.TestCatalogSurface::test_local_builder_surfaces_normalized_schedules
```

All six `pytest` jobs (base and head × 3 seeds) themselves passed — they emit JUnit XML and let `regression diff` be the gate.

### 1. `test_a_raising_check_is_logged`

`run_static`'s per-check swallow used to record `_skip(..., "check_error")`. #1899 changed it to `_fail(..., {"check_error": ...})` so that a crashed check is actually counted by `_counts` — without that, a raise inside a HARD check dropped `hard_count` and flipped `overall_status` to `compatible` on a broken agent.

The test still read `detail["skip_reason"]`, which no longer exists. It now asserts the FAIL shape, which *is* the fix — only a counted status reaches `hard_count`/`soft_count`.

Worth noting the file was already internally inconsistent: a sibling test in the same class asserts `t018["status"] == "fail"` and was passing.

### 2. `test_persisted_check_error_does_not_replay_as_clean`

This asserted `_counts(...)["soft_count"] == 0` for a persisted `skipped` + `check_error` row — i.e. **it pinned the bug as expected behaviour**. #1899's `_did_not_pass` counts exactly that row as a finding, on purpose.

Expectation flipped to `1`, and the docstring rewritten so the test keeps real teeth: a row persisted by an *older* build, before the swallow started returning `fail`, still must not replay as a clean bill of health on a stopped-agent read.

### 3 & 4. `_build_local_template(...)`

#1899 gave it a required keyword-only `is_bundled` parameter (it selects the credential-metadata trust label, so an operator-uploaded template does not inherit the bundled catalog's label). Both ent#89 call sites now pass it. Orthogonal to the `schedules:` block under test.

## Verification

```
$ pytest tests/unit/test_compatibility_checks.py \
         tests/unit/test_ent89_template_schedules.py \
         tests/unit/test_ent128b1_compat_gates.py -q
225 passed, 1 skipped in 0.48s
```

Full backend unit suite on this branch:

```
6698 passed, 19 skipped
```

Two unrelated files (`test_1081_physical_meter.py`, `test_subscription_auto_switch_pingpong.py`) reported failures/errors in that local whole-suite run **but pass on a pristine `origin/dev` worktree and pass in CI on both base and head** — local suite-ordering artifacts on this host, not touched by this change (the diff is two test files). The four `test_1771*` hypothesis property files were excluded: the dependency is absent from this host.

## Scope

Two test files, 18 insertions / 7 deletions. No production code.

### DIFF
--- a/tests/unit/test_compatibility_checks.py
+++ b/tests/unit/test_compatibility_checks.py
@@ -674,8 +674,11 @@ class TestReportDirectionOnAFailingValidator:
 
     def test_persisted_check_error_does_not_replay_as_clean(self):
         """`_report_from_persisted` recomputes counts from `checks_json`, so a
-        swallowed raise persisted once is replayed as a clean bill of health on
-        every stopped-agent read. This pins the shape of that recompute."""
+        swallowed raise persisted once used to be replayed as a clean bill of
+        health on every stopped-agent read. ent#128 closed that at the sink:
+        `_did_not_pass` counts a `skipped` row carrying `check_error` as a
+        finding, so a row persisted by an OLDER build — before the swallow
+        started returning `fail` — still cannot replay as clean."""
         import services.compatibility as svc
 
         swallowed = {"checks": [{
@@ -687,7 +690,7 @@ class TestReportDirectionOnAFailingValidator:
             "skip_reason": None,
         }]}
 
-        assert svc._counts(swallowed["checks"])["soft_count"] == 0
+        assert svc._counts(swallowed["checks"])["soft_count"] == 1
         assert svc._counts(failed["checks"])["soft_count"] == 1
 
         degraded = svc._report_from_persisted(
@@ -699,7 +702,8 @@ class TestRunStaticLogsItsSwallow:
     def test_a_raising_check_is_logged(self, monkeypatch, caplog):
         """Before ent#89 this swallow left no trace anywhere, for all ~100
         checks — a broken validator reported "healthy" and the logs were
-        silent."""
+        silent. Both halves are now closed: ent#128 makes the result a FAIL so
+        `_counts` sees it, and ent#89 logs it so it is diagnosable."""
         import logging
 
         monkeypatch.setitem(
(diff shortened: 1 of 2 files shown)


vybe (approved):
Approving my own finding, with the reasoning stated so it can be checked rather than taken on trust.

**This is a test-only reconciliation, not a behaviour change.** The diff is two test files, +18/-7, no production code. Each of the four assertions was correct when written and became wrong when #1899 landed:

1. `test_a_raising_check_is_logged` read `detail["skip_reason"]`, which stopped existing when the swallow moved from `_skip` to `_fail`. Asserting the FAIL shape is not a workaround — it is the property #1899 added, since only a counted status reaches `hard_count`/`soft_count`. The same file already had a passing sibling asserting `"fail"`, so it was internally inconsistent.
2. `test_persisted_check_error_does_not_replay_as_clean` asserted `soft_count == 0` for a persisted `skipped`+`check_error` row — it pinned the **bug** as expected behaviour. `_did_not_pass` counts that row deliberately. Flipped to `1` and kept with teeth: a row persisted by an older build still must not replay as clean.
3/4. `_build_local_template` gained a required keyword-only `is_bundled` in #1899; both ent#89 call sites now pass it.

**The process gap is the real finding.** `dev`'s four required checks run no unit tests, so auto-merge landed #1946 before `regression diff` reported. Two independently-green PRs can therefore redden `dev` on contact, with nothing blocking either. Worth considering `regression diff` as a required check — filing that separately rather than in this PR.

Verified: the three affected files → 225 passed / 1 skipped; full `tests/unit` on this branch → 6698 passed. The two unrelated files that misbehaved in my local whole-suite run pass on a pristine `origin/dev` worktree and pass in CI on both base and head, so they are local ordering artifacts, not regressions.

Issue #231, opened 2026-03-30
Trinity CLI tool mirroring MCP server


vybe:
## Summary

Create a `trinity` CLI tool that mirrors all 58+ MCP server tools as shell commands. Agents can call these via Bash instead of MCP, eliminating LLM reasoning overhead for deterministic operations and saving significant tokens.

## Motivation

- MCP tool calls cost ~500-1000 tokens per invocation (LLM reasons about tool selection, parameters, response parsing)
- A Bash call to `trinity agents list` costs ~50 tokens (command + output)
- CLI commands are composable (pipes, scripts, chaining)
- Non-MCP clients (scripts, CI/CD, cron jobs) get first-class access without MCP client libraries

## Architecture

```
Agent (MCP) ──→ MCP Server ──→ Backend API (FastAPI)
                                      ↑
Agent (Bash) ──→ trinity CLI ─────────┘
Human ─────────→ trinity CLI ─────────┘
Script ────────→ trinity CLI ─────────┘
```

The CLI is a thin HTTP client to the existing FastAPI backend, same as the MCP server. Auth reuses MCP API keys as Bearer tokens (already supported).

## Proposed Command Structure

```bash
# Agent management
trinity agents list [--format json|table]
trinity agents get <name>
trinity agents create <name> [--template <tpl>]
trinity agents delete <name>
trinity agents start <name>
trinity agents stop <name>
trinity agents rename <old> <new>

# Chat & communication
trinity chat <agent> "<message>"
trinity chat history <agent>
trinity logs <agent> [--tail N]

# Schedules
trinity schedules list <agent>
trinity schedules create <agent> --skill <skill> --cron "<expr>"
trinity schedules trigger <agent> <schedule-id>
trinity schedules toggle <agent> <schedule-id>

# Skills
trinity skills list
trinity skills get <id>
trinity skills assign <agent> <skill-id>
trinity skills sync <agent>

# Tags
trinity tags list
trinity tags set <agent> <tag1> [tag2...]

# Monitoring
trinity health fleet
trinity health agent <name>

# Events
trinity events emit <type> --payload '{"key": "value"}'
trinity events subscribe <type> --agent <name>
trinity events list

# Executions
trinity executions list [--agent <name>]
trinity executions get <id>

# Systems
trinity systems list
trinity systems deploy <manifest>
trinity systems restart <name>

# Subscriptions & Auth
trinity subscriptions list
trinity subscriptions register --agent <name> --plan <id>

# Notifications
[...]

Issue #39, opened 2026-02-24
Vectorized Episodic Memory


vybe:
Auto-store conversations for 'Have I solved this before?' queries.

oleksandr-korin:
Closing as speculation — one-sentence body ("Auto-store conversations for 'Have I solved this before?' queries.") with no acceptance criteria, no scope. Adjacent territory is actively being explored by #380 USER-MEMORY (P2, in progress) and the per-user memory write surface (MEM-001, #888). Re-open or file a fresh issue if a concrete episodic-memory requirement emerges that isn't covered by those. Reference: backlog lean review on 2026-05-28.

PR #1959, opened 2026-08-03
fix(security): stop exception messages reaching ops/system-agent responses (#1917)

### LINKED ISSUE #1917: bug: exception messages flow into ops/system-agent API responses (py/stack-trace-exposure siblings of #1860)
## Summary

CodeQL flagged `py/stack-trace-exposure` on `restart_fleet` during PR #1912, which fixed its own instance (response carries `HTTPException.detail` or the exception class name only; full message + traceback go to the backend log via `exc_info=True` — see the pattern at `src/backend/routers/ops.py`, the `restart_fleet` error branch). The same raw-`str(e)`-into-response pattern pre-exists at sibling sites that PR deliberately left untouched (minimal-changes rule):

- **Open CodeQL alert [#231](https://github.com/Abilityai/trinity/security/code-scanning/231)**: `src/backend/routers/system_agent.py:83` (ref `main`)
- `src/backend/routers/ops.py:205` — fleet health `"Agent not responding: {str(e)[:50]}"` (truncated, lowest risk)
- `src/backend/routers/ops.py:564` — `stop_fleet` per-agent `"error": str(e)`
- `src/backend/routers/ops.py:751` — `_stop_agent_container` `"error": str(e)` (feeds `emergency_stop` results)
- `src/backend/routers/ops.py:1056` — ops costs `"Failed to fetch metrics: {str(e)}"`
- (grep-worthy: `routers/agents.py` single-agent stop/start raise `HTTPException(500, f"...{str(e)}")` — same class via `detail`)

All these endpoints are admin-gated, so exposure is bounded — but exception messages can embed internals (the #1885 reaped-cmdline and git-stderr-PAT classes, `docs/memory/learnings.md` 2026-07-14), and alert [#39](https://github.com/Abilityai/trinity/security/code-scanning/39) (the `restart_fleet` instance on `main`) demonstrates CodeQL files them as findings.

## Acceptance criteria

- [ ] Each listed site returns platform-authored text or exception class name only; full message + traceback logged server-side (`exc_info=True`) — the PR #1912 pattern
- [ ] CodeQL alert #231 closes on the next `main` analysis (alert #39 closes on its own once #1912 ships in a release — verify, don't re-fix)
- [ ] Tests assert the raw message is absent from responses (leak regression guards, per `tests/unit/test_1860_fleet_restart_adoption.py`)

Found while resolving the code-scanning review on PR #1912 (#1860).

dolho:
## Problem

PR #1912 fixed `restart_fleet` after CodeQL flagged `py/stack-trace-exposure`, and deliberately left the sibling sites alone under the minimal-changes rule. This closes them with the same pattern: the response carries platform-authored text or the exception **class name**; the full message + traceback go to the backend log via `exc_info=True`.

Exposure is bounded — every endpoint here is admin-gated — but exception messages routinely embed internals. Docker and httpx errors lead with the socket path or the internal container host, and `learnings.md` (2026-07-14) records git stderr carrying a **PAT** into operator-visible state. "Admin-only" is a blast-radius argument, not a reason for the string to be there.

## Sites

Located by **content, not the issue's line numbers** — `dev` has moved since it was filed, and every cited line had drifted.

| File | Sites |
|---|---|
| `routers/ops.py` | fleet health probe · `stop_fleet` per-agent · `_stop_agent_container` (feeds `emergency_stop`) · ops costs |
| `routers/system_agent.py` | health probe (**CodeQL alert #231**) · workspace cleanup · template copy · re-initialize 500 · restart 500 · terminal WebSocket error frame |
| `routers/agents.py` | start / stop / logs `HTTPException` details |

**The health probe had `str(e)[:50]`.** Truncation bounds a leak, it doesn't remove one — docker/httpx messages lead with the host, socket path or URL, so the first 50 characters *are* the sensitive part.

**Two sites weren't in the issue** and are the same defect, found by grepping the named files rather than trusting the list: system_agent's re-initialize/restart 500 details, and its terminal WebSocket error frame — the only one here that reaches a **browser** rather than an admin API client.

## Guards — 8 tests, all 8 failing against pre-fix code

Each drives the real router function with a sentinel planted in the exception message, and asserts **both halves**:

```python
assert SENTINEL not in body        # the leak is gone
assert "RuntimeError" in body      # ...and the failure is still diagnosable
```

The second assertion matters: a response that dropped the error entirely would pass a sentinel-only check while making the failure impossible to debug.

Plus a static ban on `str(e)` in the two routers where every occurrence was this defect. `agents.py` is deliberately **excluded** from that ban — it's a 1000+ line router with many `str(e)` uses that never reach a response, so a blanket rule there would be false; its three sites get a targeted window check instead.

### One test was vacuous, and the pre-fix run caught it

My first version of the fleet-health test patched a `get_agent_context_info` that **does not exist**. `raising=False` made that silent, the probe never raised, and the test **passed against the unpatched router** — 7 of 8 failing pre-fix instead of 8. It now patches the real seam (`get_agent_client(...).get_session()`) and asserts `await_count == 1`, so it cannot pass without the probe running. Same vacuous-negative class as #1932/#1951.

## Verification

```
tests/unit/test_1917_stack_trace_exposure.py                         8 passed  (8 failed pre-fix)
+ test_1860_fleet_restart_adoption / test_1816_system_agent_adoption  102 passed
pytest unit/ -k "ops or system_agent or agents"                      275 passed, 2 skipped
```

## Acceptance criteria

- [x] Each listed site returns platform-authored text or class name only; full detail logged with `exc_info=True`
- [ ] **CodeQL alert #231 closes on the next `main` analysis** — cannot be verified from a PR branch; alert #39 closes on its own once #1912 ships in a release (the issue says verify, don't re-fix — I have not touched `restart_fleet`)
- [x] Tests assert the raw message is absent from responses

Closes #1917


vybe (approved):
Validated via /validate-pr. Closes #1917 (closing keyword added so the issue auto-promotes to `status-in-dev`).

Security fix, no schema/API/config surface, no new top-level backend module, no secrets in the diff. Three things I checked specifically and liked:

- **Sites located by content, not the issue's stale line numbers** — and two sites that weren't in the issue (system_agent's re-init/restart 500 details, and the terminal WS error frame, the only one reaching a browser) were found by grepping the named files rather than trusting the list.
- **`str(e)[:50]` correctly treated as a leak, not a mitigation** — docker/httpx messages lead with the socket path or internal host, so the first 50 chars *are* the sensitive part.
- **Both-halves assertion** (`SENTINEL not in body` AND `"RuntimeError" in body`) — a response that dropped the error entirely would pass a sentinel-only check while making the failure undebuggable. The disclosed vacuous first draft (patching a `get_agent_context_info` that doesn't exist, silent under `raising=False`) is exactly the class worth catching, and 8/8 now fail pre-fix.

The static `str(e)` ban scoped to the two routers where every occurrence was this defect — with `agents.py` deliberately excluded and given targeted window checks instead — is the right call over a blanket rule that would be false there.

CodeQL alert #231 can only close on the next `main` analysis; noted, not blocking.

PR #1961, opened 2026-08-03
fix(security): one hardened parser for all author-controlled YAML (trinity-enterprise#314)

dolho:
Closes the live amplification vector in `abilityai/trinity-enterprise#314` (child of epic ent#318). Filed privately because it documents a working attack; the code is OSS-core, so the fix lands here.

## The vector, reproduced on this tree before any code was written

10-way nested anchor bomb terminating in `skills:`, serialized with `json.dumps`:

| levels | source | serialized | amplification | `safe_load` time |
|---:|---:|---:|---:|---:|
| 4 | 298 B | 1.10 MB | 3,700× | — |
| 5 | 357 B | 11.02 MB | 30,882× | — |
| 6 | 416 B | **110.25 MB** | **265,017×** | **0.0011 s** |

The parse is *free* — PyYAML resolves aliases into a small shared-reference DAG. The cost lands entirely on whatever **walks** it: `json.dumps` for `/api/templates`, a deep copy, a `str()` in a warning. **An input-size cap cannot close this**; the input genuinely is small.

Duplicate keys reproduce too — `safe_load` on the issue's fixture returns `{'env_file': ['SECOND_KEY']}`. For a `credentials:` block that means a template can show one set of required credentials to a human reading the file and declare a different set to Trinity.

Reachable by any `creator`-role user via a public repo (ent#123 tokenless), and `_build_template` runs unfenced inside `get_all_templates()` — so one hostile repo is on the catalog path for everyone who lists templates.

## Fix: one loader, two policies

`utils/safe_yaml.py` — size cap + alias policy + duplicate-key rejection, **rejecting rather than truncating**. There were already **three** partial implementations (#1884's manifest loader and two independent `_NoAliasSafeLoader`s) while the catalog path had none — which is precisely how this survived three CSO audits.

| Policy | Used by | Why |
|---|---|---|
| `BUDGET` | manifests, `template.yaml` catalog | bounds expansion **cost**. A naive `maxAliasCount` is the wrong shape — billion-laughs uses ~45 aliases and sails under any count, because the blow-up is multiplicative per level. Measured: refuses level 4+, admits the small honest anchor |
| `REJECT` | skills frontmatter, live agent-writable `template.yaml`, agent workspace YAML | those consumers walk every field. **Preserved deliberately** — relaxing them to a budget would be a security regression shipped as a refactor |

Both are **required arguments with no default**: picking one silently is how a caller ends up with the wrong gate.

**Nine readers wired**, not the four the issue names — `template_service` (remote fetch + 4 local), `system_service`, `skill_packaging`, `credential_requirements_service`, `compatibility/static_checks` (which is also how `dashboard.yaml` is covered), `git_service`, `agent_service/lifecycle`, `agent_service/crud`, `system_agent_service`.

## Backwards compatibility

`ManifestError` is now a `HardenedYamlError` subclass, so the shared loader raises the manifest's own type: `routers/systems.py`'s code-to-400 mapping and every `except ManifestError` are unchanged, and the published codes (`manifest_alias_budget_exceeded`, `manifest_duplicate_key`, `manifest_too_large`) are pinned by test. #1884's *SafeLoader-not-FullLoader* guard was re-pointed at the shared factory rather than deleted along with the class it named.

## Three things this found the hard way

1. **`skill_packaging`'s copy gated at the scanner** (`fetch_alias`) as well as at compose. The shared loader carries both hooks, so consolidating didn't quietly drop the stricter of the two copies.
2. **Both `REJECT` sites catch `yaml.YAMLError`**, which `HardenedYamlError` (a `ValueError`) is not. Left alone, a refused alias would have escaped as a 500 instead of the graceful degradation those contracts promise.
3. **The policy first lived in `template_service`.** Several harnesses stub `services.*` wholesale in `sys.modules`, so `crud.py`'s import became a MagicMock, the parse returned a MagicMock, and the caller's `isinstance(..., dict)` check rejected every local template as "malformed" — 7 failures in `test_1759`. The policy now lives in `utils.*` (import-safe from anywhere), which also avoids a new `crud → template_service` edge.

## The guard test was wrong twice, instructively

- **Draft 1** listed the four files it had just fixed — self-fulfilling. It passed while `agent_service/crud.py` (`template.yaml` on the **creation** path — the exact file this issue is about) was still on bare `safe_load`, along with four others.
- **Draft 2** scanned text and flagged a *docstring quoting* `yaml.safe_load(...)` in `template_schedules.py` as a vulnerability. It now walks `ast.Call` nodes — the trap `learnings.md` (2026-07-10) already records for the `containers_run` guard.

It now sweeps the **whole backend** with `utils/safe_yaml.py` as the only module allowed to call PyYAML directly.

## Verification

- 21 ent#314 tests; the 4 product-level guards fail against pre-fix code
- A fixture-integrity test asserts the bomb *still amplifies* under bare `safe_load`, so the suite cannot pass vacuously if PyYAML changes
- A legitimate small anchor **still parses** — the #1932 lesson: a guard that rejects the legitimate document reads as "hardened" in an audit while being an outage
- 1179 passing across every touched suite; full unit run in flight

## AC status

- [x] Level-6 bomb rejected at parse in any author-controlled YAML
- [x] Duplicate top-level keys are a named error (incl. the issue's `credentials:` case, and nested)
- [x] `_fetch_template_yaml` size-caps the fetched document
- [x] One shared loader; #1884's manifest path migrated onto it
- [x] Regression tests naming this issue
- [ ] Public follow-up filed once the caps ship — by definition after merge

Related to trinity-enterprise#314


vybe (approved):
/validate-pr — approved.

Security scans clean; new module is `src/backend/utils/safe_yaml.py`, under the wholesale `COPY src/backend/utils /app/utils/` (Dockerfile:89), so the #1033 top-level-glob packaging class doesn't apply. No new `os.getenv`. `architecture.md` + `learnings.md` updated. 30 test defs named for the issue; the fixture-integrity test asserting the bomb *still* amplifies under bare `safe_load` is what stops the suite passing vacuously, and the "a legitimate small anchor still parses" case is the right #1932 counterweight. AST-walking source guard with an **empty** allowlist is the correct shape — the two wrong drafts you describe are the whole lesson.

Policy as a required argument with no default, and `ManifestError` subclassing `HardenedYamlError` so `routers/systems.py`'s code→400 map is untouched: both right.

**One residual, non-blocking.** The guard walks `_BACKEND.rglob("*.py")`, so the agent server is outside the sweep — `docker/base-image/agent_server/` still has six bare `yaml.safe_load` calls on author-controlled documents: `routers/info.py:108,173,252` (`template.yaml` — the file this issue is about), `routers/skills.py:66` (skill frontmatter, a REJECT-policy consumer on the backend side), `routers/dashboard.py:205`, `routers/files.py:244`. The backend proxies `/info` and `/dashboard`, so the same alias amplification lands on the same JSON-serialization walk, just inside the container first. Lower-privilege path than the catalog one you close here, and your "sweeps the whole backend" wording is honest about it — but Invariant #5's vendored-mirror pattern (`credential_paths.py`, `model_context.py`) is exactly the shape it needs. Filing it so it doesn't live only in a merged PR body.

Resolved the `learnings.md` append conflict against dev (kept all three entries; #1909's from dev first, then both ent#314 ones). Diff vs dev is exactly the 15 ent#314 files.

Cross-tracker ref, so `status-in-dev` on ent#314 goes on manually after merge.

PR #1962, opened 2026-08-03
fix(tests): de-race the a13 exact-cutoff boundary test with a single clock sample (#1909)

### LINKED ISSUE #1909: bug: test_a13_exact_cutoff_row_is_not_expired is wall-clock flaky (~3.3%), reddens regression diff on unrelated PRs

webmixgamer:
## Summary

- `test_a13_exact_cutoff_row_is_not_expired` sampled the wall clock **twice** — once to build its fixture, once inside `expire_stale_queued` at query time. A second rollover between the two advanced the threshold, made `exact` strictly older, expired the row, and flipped the assertions (`assert 2 == 1`).
- Freezes the clock the implementation reads, so setup and query observe **one instant**. The race is gone by construction, not made less likely.
- Tolerance deliberately **not** widened — this case pins a zero-width boundary (the Invariant-#16 `strftime` vs `utc_now_iso()` format mismatch), so nudging the fixture off it would delete the property the test exists to pin. Recorded as an explicit anti-goal in the docstring.

## Changes

| File | Change |
|---|---|
| `tests/unit/test_1771c_schedules_cas_edges.py` | Single clock sample + patch-took-effect guard + docstring rationale (+34/−4) |
| `docs/memory/learnings.md` | New bug-class entry (+4) |

**No production code is touched.**

## Why the fix isn't "run it more times"

A wall-clock flake's rate is a property of the **machine**, not the code. The issue measured 3.3% (5/150); this branch's baseline measured **0.33%** (1/300) on faster hardware — same test, same code, 10× apart. A green run-loop on a fast box says nothing about a slow shared runner, and by the rule of three you'd need ~900 clean runs just to bound your *own* machine at the rate you were disproving.

So the race is **forced** instead. A ~20-line pytest plugin parks the runner ~1.5 ms before a wall-second boundary, putting the rollover inside the window every time:

```python
# align_boundary.py  —  load with: -p align_boundary  (dir on PYTHONPATH)
import os, time
from datetime import datetime, timezone

TARGET  = os.environ.get("TARGET_SUBSTR", "a13_exact_cutoff")
PARK_US = int(os.environ.get("PARK_US", "998500"))

def pytest_runtest_call(item):
    if TARGET not in item.name:
        return
    while datetime.now(timezone.utc).microsecond < PARK_US:
        time.sleep(0.0002)
```

## Test Plan

| Check | Before | After |
|---|---|---|
| Forced second-boundary (`PARK_US=998500`) | **20/20 fail** | **0/25 fail** |
| Natural runs | 1/300 fail | **0/300 fail** |
| File under the 3 real CI seeds (12345/67890/99999) | — | 43 passed each |
| Full unit suite (seed 12345) | — | **7002 passed, 17 skipped, 0 failed** |

The before/after uses the same harness in both directions, so the test is shown to have power — it catches the bug *and* clears a correct fix, rather than merely not failing.

> Suite numbers were taken on a venv built from `tests/requirements-test.txt` (FastAPI pinned `<0.116`). On an out-of-cap venv, `test_1069_voip_call_path_param` fails with `ImportError: cannot import name 'get_flat_dependant'` — pre-existing on `dev`, unrelated to this change, and an environment artifact rather than a repo bug.

## Sibling audit

The issue's closing note asked whether the same two-clock shape exists elsewhere. It does not:

- `test_1771c_*` L338/L354 — 300 s margin
- `test_1771c_*` L547 — parametrized, tightest margin 24 min
- `test_1771c_*` L601/L617 (`a16`) — already passes `now_iso=` **into** the implementation (single source by construction)
- `test_backlog.py` — 24 h and 9 s margins; documents this as the "time-mock-free pattern"

Only the zero-width boundary case is tight enough to lose the race often enough to notice — and it's the one case the wide-margin pattern structurally cannot help.

Fixes #1909


dolho:
## /review — 0 critical, 2 informational

Reproduced the central claim independently rather than taking it on trust. The fix targets the right seam: `expire_stale_queued` builds its threshold in **Python** (`queue.py:585`, `datetime.now(timezone.utc)` off the module global), not in SQL, so patching `queue_mod.datetime` genuinely controls it. Rebuilt the forced-boundary plugin from the PR body:

| | result |
|---|---|
| Pre-fix, forced boundary | **20 / 20 fail** |
| Post-fix, same harness | **0 / 20 fail** |
| Post-fix, natural | 0 / 60 fail |
| Whole file, 3 CI seeds | 43 passed each |

Sibling audit checks out too: L338/L354 = 300 s, the parametrized sweep's tightest pair is `(0.5 h window, 0.1 h age)` = 24 min, and `a16` passes `now_iso=` into the implementation.

---

### I1 — the patch-took-effect guard does not catch the failure its comment names

```python
monkeypatch.setattr(queue_mod, "datetime", _FrozenDatetime)   # :537
# Patch-took-effect guard: if `queue.py` ever aliases its datetime import, ...
assert queue_mod.datetime.now(timezone.utc) == frozen          # :543
```

I mutated the implementation twice to test it:

**(a) The aliasing case the comment names** — `from datetime import datetime as _dt`:

```
E  AttributeError: <module 'db.schedules.queue' ...> has no attribute 'datetime'
   tests/unit/test_1771c_schedules_cas_edges.py:537
```

Caught — but by `monkeypatch.setattr` itself, at **line 537**. Line 543 never executes. The guard is dead code for its stated purpose.

**(b) The realistic drift** — threshold moves to `iso_cutoff(...)`, so the implementation stops reading the patched name:

- guard on :543 **passes**
- the failure that surfaces is `assert ops.expire_stale_queued(24) == 1` → `assert 2 == 1`
- and only under a forced boundary: **0 / 40 failures on natural runs**

So the flake returns at its original rate with no signal — the exact outcome this PR exists to prevent. `assert queue_mod.datetime.now(…) == frozen` is tautological on the success path: it asserts the name it just set returns what it was set to, never that the implementation *reads* that name.

This drift is more likely than average here — Invariant #16 pushes ISO-Z cutoff comparisons toward `iso_cutoff()`, which is precisely mutation (b).

A guard that survives both mutations, behavioural rather than source-coupled:

```python
# Prove the patched clock is load-bearing: re-freeze an hour later and the same
# row must now expire. If the implementation stopped reading this name, both
# calls agree and this fires — which is the drift the comment is about.
frozen = frozen + timedelta(hours=1)      # rebind before a second call
assert ops.expire_stale_queued(24) == 1   # a13-exact is now strictly older
```

Or the repo's existing source-guard idiom: `assert "datetime.now(" in inspect.getsource(queue_mod.ScheduleQueueMixin.expire_stale_queued)`.

**Not blocking.** The fix is correct today and demonstrated. The defect is the comment promising a protection that isn't there — which is what the next reader will trust.

### I2 — dead branch in the test double

`_FrozenDatetime.now(cls, tz=None)` returns naive when `tz` is falsy. `queue.py` calls `datetime.now(timezone.utc)` at both sites (`:101`, `:585`) and never bare, so that branch is unreachable. Harmless, but untested code inside a test whose point is determinism.

### Clean

No production code, so SQL/auth/credential passes are vacuous. `monkeypatch` is function-scoped and `a13`–`a16` in sequence pass (10 passed) — no leak. `_DB_MODULES` pops `db.schedules` but not `db.schedules.queue`, so the re-imported package binds the same cached submodule the mixin's `__globals__` point at — the patch does reach the method, which the forced-boundary result confirms.

### Not verified

"Full unit suite (seed 12345): 7002 passed" — I ran the affected file under all three CI seeds, not the whole suite. Flagging as unverified rather than assuming.

dolho (approved):

webmixgamer:
## Both addressed in `ccf8ad16` — I1 confirmed by reproducing your mutations

Ran both mutations independently before changing anything. You're right on both, and I1 is worse than "the comment overpromises": the guard was **unfalsifiable**. It asserted that the name I had just `setattr`'d returns what I set it to — never that the implementation reads it.

| Mutation of `expire_stale_queued` | Old guard |
|---|---|
| (a) `from datetime import datetime as _dt` | `AttributeError` at **:537** — `monkeypatch.setattr` itself, one line before the guard. Never executed, exactly as you said. |
| (b) threshold off the module global | **passed**, 0/40 natural failures — while the flake was fully restored at **20/20** forced |

(b) is the one that matters, and your point about Invariant #16 pushing ISO-Z cutoffs toward `iso_cutoff()` is what makes it a likely refactor rather than a hypothetical.

### I1 — behavioural guard

Took your suggestion. Preferred it over the `inspect.getsource` idiom for the reason your own report implies: a source grep asserts a string is present, not that the patched clock is load-bearing — it would pass under (b) too, since `datetime.now(` survives in the aliased/relocated form.

```python
frozen = frozen + timedelta(hours=1)
assert ops.expire_stale_queued(24) == 1
assert status_of("a13-exact") == "failed"
```

Added the status assertion on top of the rowcount so it pins the transition, not just the count.

### I2 — tripwire rather than deletion

Deleting the dead branch loses the signal; a bare `datetime.now()` would then silently receive an **aware** value where naive semantics are expected. Replaced it with `assert tz is not None`, which converts it into a real detector — and it earns its place: mutation (c) below is caught by it alone.

### Mutation matrix (natural runs unless noted)

| Scenario | Result |
|---|---|
| correct code | **0/30** false positives · **0/20** forced-boundary failures |
| (a) aliased import | caught by `monkeypatch.setattr` |
| (b) bypassed name | **30/30 detected** — old guard **0/40**, silent |
| (c) bare `now()` | caught by the I2 tripwire |

Documented one residual honestly in the comment rather than implying the guard is airtight: if the wall second rolls over between the two calls, a bypassing implementation expires the row anyway and the guard passes. ~0.3%, unobserved across 30 runs, and it costs a *detection*, never a false failure.

### Your "not verified" note

Fair to flag it. The full-suite number is real — **7002 passed, 17 skipped, 0 failed** — but it was measured on a venv built from `tests/requirements-test.txt`, not my default one, which had FastAPI 0.141 against the `<0.116` cap and failed `test_1069_voip_call_path_param` on an unrelated `ImportError`. Baselined that on clean `dev` before concluding it wasn't mine.

CI has since gone green on the rebased commit, and post-#1910 that run was the first on **Python 3.13** (`Successfully set up CPython (3.13.14)` in the job log) — 18 pass / 3 skipping, `regression diff` pass.

Whole file green under all three CI seeds (43 each); `test_1771c_schedules_cas_properties.py` + `test_backlog.py` 44 passed.

Issue #1895, opened 2026-07-30
bug: 108 test files sit in tests/ root, where no CI job runs them


webmixgamer:
## Summary

Every CI pytest invocation is scoped to `tests/unit/`. The **108 `test_*.py` files in the `tests/` root are executed by no workflow at all** — they run only from the local `tests/run-core.sh`.

This is not theoretical. It is why #1880 shipped: `tests/test_canary_invariants.py` (3,726 lines) is the only file exercising the canary alert composer, and it has never been a gate. Five invariants — including the critical credential-leak check — shipped with unusable alerts and nothing could notice.

## Evidence

| File | Invocation |
|---|---|
| `.github/workflows/backend-unit-test.yml` | `cd tests && python -m pytest unit/ -m "not slow"` |
| `.github/workflows/backend-unit-nightly.yml` | same, base and head |

`tests/unit/pytest.ini` sets `norecursedirs = ..`, so `unit/` is a closed island — nothing recurses out into the root suite.

## How many could gate? Measured, not estimated

Two intuitive classifications both turned out wrong, so this section records what was actually run.

**Attempt 1 — "does it request a live-instance fixture?"** Gives ~63 integration / ~45 self-contained. **Wrong**: it counts a *local* fixture definition as a live dependency. `tests/test_canary_invariants.py` was misfiled this way — it defines its own `api_client` (see below).

**Attempt 2 — "run each candidate standalone where it sits."** Only **10 of 45 pass**. Also misleading: the failures are almost all `httpcore.ConnectError: Connection refused` raised at setup of the **session-scoped `api_client`** fixture in `tests/conftest.py`, which is pulled in for nearly every root file whether or not the test asks for it. That measures the root conftest, not the test.

**Attempt 3 — copy the file into `tests/unit/` and run it there.** This is the only question that matters, and the answer is **per-file**. A 6-file trial:

| File | Result under `tests/unit/` | Verdict |
|---|---|---|
| `test_watchdog_unit.py` | 43 passed | moves clean |
| `test_proactive_audit_unit.py` | 4 passed | moves clean |
| `test_password_validation.py` | 1 collection error | mechanical fix |
| `test_event_bus.py` | 1 collection error | mechanical fix |
| `test_cli_profiles.py` | 1 collection error | mechanical fix |
| `test_public_links.py` | 15 failed, 14 errors | genuinely needs a live API |

The collection errors are **path-depth assumptions**, not architecture:

```
[...]

PR #1963, opened 2026-08-03
ci: give the live-instance integration suite a nightly CI home (#1896)

### LINKED ISSUE #1896: bug: the live-instance integration suite (~60 files) has no CI home
## Summary

Roughly **60 of the 108 `test_*.py` files in `tests/` root are live-instance integration tests** — they need a running Trinity and cannot run in the current CI, which only executes `tests/unit/`. They gate nothing.

Sibling of #1895, deliberately split. #1895 relocates the tests that *can* pass under `tests/unit/` today; this issue is the remainder, which needs a stack to run at all. Splitting keeps the cheap fix unblocked.

## Why they can't just be moved

`tests/conftest.py` opens with:

> ⚠️ **THIS SUITE MUTATES THE TARGET INSTANCE.** The API fixtures CREATE and DELETE …

The fixtures create and delete real agents against `TRINITY_API_URL` (default `http://localhost:8000`). A session-scoped `api_client` fixture opens the connection, so without a backend they fail at setup with `httpcore.ConnectError: Connection refused`.

That mutation property is also why this is not simply "point CI at the directory": the suite needs an instance it is allowed to destroy, plus cleanup that survives a crashed fixture (the conftest already has a session-scoped reclaim pass for agents a teardown missed).

## This is not greenfield — three workflows already stand up a stack

| Workflow | How |
|---|---|
| `.github/workflows/frontend-e2e.yml` | `scripts/deploy/start.sh` |
| `.github/workflows/container-security.yml` | `scripts/deploy/start.sh` |
| `.github/workflows/build-without-submodule.yml` | `docker compose up` |

`frontend-e2e.yml` is also the model for **gating**: it is path-filtered via a `changes` job and skips on unrelated PRs. A ~60-file mutating integration suite should not run on every push — it wants the same treatment (path filter, label opt-in, or nightly).

## The count is approximate and will settle

~60 is the residue of #1895's classification, and that classification was wrong twice before being measured (see #1895 for the method and the trap). The real boundary is decided per-file as #1895 proceeds: anything that turns out to pass under `tests/unit/` leaves this issue's scope. **Do not treat ~60 as a target.**

## Acceptance criteria

- [ ] A CI job stands up a Trinity stack and runs the live-instance root suite against it, reusing `scripts/deploy/start.sh` rather than a bespoke compose path
- [ ] The job is **gated**, not on every PR — path filter, label opt-in, or nightly. A mutating ~60-file suite on every push is a cost and a flake source; `frontend-e2e.yml`'s `changes` job is the in-repo pattern to copy
- [ ] Failures are attributable: the run produces JUnit XML and feeds the same base-vs-head **`regression diff`** comparison the unit job uses, so a pre-existing failure does not read as a new one
- [ ] The instance is disposable and teardown is guaranteed even on a crashed fixture — the suite deletes real agents, so a leaked instance between runs is a correctness problem, not just cost
- [ ] Runtime is stated and bounded, with a documented decision on what happens when it exceeds it (skip, shard, or move to nightly)
- [ ] `tests/run-core.sh` remains the supported local path — this adds a CI home, it does not replace the local workflow

## Technical notes

- **Sequencing**: land #1895 first. Every file it moves is one fewer here, and the residue is only knowable after it.
- Secrets: the suite needs an admin credential against the throwaway instance. `scripts/deploy/start.sh` generates what it needs; confirm nothing requires a real external key (Slack, Twilio, Gemini) or the affected tests must be skip-marked rather than given credentials in CI.
- Consider whether this belongs as a **nightly** job rather than per-PR. `backend-unit-nightly.yml` already establishes a nightly base-vs-head pattern with the same diff tooling, which would give coverage without adding minutes to every PR.

dolho:
~93 of the `test_*.py` files in `tests/` root need a **running** Trinity — the fixtures create and delete real agents against `TRINITY_API_URL` — and no CI job ran them. They gated nothing.

`integration-nightly.yml` is that home. It composes what already exists rather than inventing anything: `scripts/deploy/start.sh` for the stack (as `frontend-e2e.yml` does), the `discover → test → comment` split and sticky-comment machinery from `backend-unit-nightly.yml`, and `diff-pytest-failures.py` for base-vs-head attribution.

## It did not need to wait for #1895

The issue's Technical Notes open with *"Sequencing: land #1895 first"* — but `tests/run-core.sh:22` already defines this suite **directory-based**:

```bash
pytest -m "not slow" --ignore=unit --ignore=process_engine
```

The workflow uses the same formulation, so every file #1895 relocates leaves this job's scope automatically, with no workflow edit. There is no "~60" target to chase and no file list to drift — which is also the AC that keeps this from rotting.

## Why nightly (AC 2)

The suite **mutates** its target and needs ~3-5 min on top of a ~2-3 min boot, **twice** (base and head). In the merge path that is ~10 min on every backend PR plus a mutating flake surface in front of every merge. Nightly also gets the base-vs-head attribution the ACs require for free — that tooling exists for exactly this shape.

## Three things measured, not assumed — each would have shipped this broken

1. **`--continue-on-collection-errors` is load-bearing.** Three root modules currently fail to *import* (two `git-sync` files wanting `utils.credential_sanitizer`, plus `tests/integration`'s conftest). Without the flag pytest aborts with `Interrupted: 3 errors during collection` and writes **no JUnit XML at all** — the job would have failed every night, for every PR, reporting nothing. With it, the errors land on both sides, cancel in the diff, and a *new* collection error still surfaces. I found this by actually running the collection, not by reading the invocation.

2. **`src/cli` has to be installed.** `test_cli_*.py` imports `trinity_cli`, deliberately kept out of `requirements-test.txt` (an `-e ./src/cli` line breaks GitHub's dependency-graph updater — `tests/setup-env.sh` documents it). Without the on-demand install those modules fail to import and their tests silently never run — precisely the no-coverage this issue exists to end. That accounted for 2 of the original 5 collection errors.

3. **A missing JUnit must fail the job loudly.** `diff-pytest-failures.py` is fail-closed on a missing input, so a stack that never booted would have been reported to the author as *"this PR introduced regressions"* — an infrastructure failure wearing a PR's name. That is #1941's exact class, so the job now checks both XMLs exist and errors out instead of posting a false verdict.

Also inherited from #1941, since this is a second workflow that merges: full-depth fetch (two shallow sides share no ancestor and every PR reads as conflicting), and a conflict verdict gated on actual unmerged paths.

## AC coverage

- [x] **1** — boots a stack via `scripts/deploy/start.sh`, runs the live root suite
- [x] **2** — gated: nightly + `workflow_dispatch` (with a single-PR input), never on `pull_request`
- [x] **3** — JUnit XML on both sides, fed to the same `regression diff` tooling the unit job uses
- [x] **4** — `docker compose down -v` between sides so head never inherits base's agents/volumes/DB; the runner VM is destroyed afterwards regardless, which is what makes the instance disposable
- [x] **5** — 45-min ceiling, expected ~15-20, with the escalation written into the workflow header: shard by marker or cut the PR limit **before** raising the ceiling
- [x] **6** — `tests/run-core.sh` untouched; a guard asserts the two keep defining the same suite

## Guards

13 tests in `test_1896_integration_nightly_workflow.py`, in the shape of `test_1941_nightly_merge_depth.py` — they pin the properties whose absence is invisible until the night it matters, including the two load-bearing flags above and the security split (the PR-code job holds only `contents: read`; the write-scoped `comment` job never checks out PR code).

## What I could not verify

The workflow has never executed — it is nightly and its first real run is after merge. YAML parses, every `run` block passes `bash -n`, the diff script's self-test passes and its CLI contract matches, and the pytest invocation was run locally to the point of collection. The stack boot itself is copied verbatim from `frontend-e2e.yml`, which runs green today. First-run risk is real; `workflow_dispatch` with a `pr_number` input exists so it can be smoke-tested against one PR on demand rather than waiting for 06:30 UTC.

**Known caveat, not introduced here:** the 3 pre-existing collection errors above are real breakage in the root suite. They are exactly what "no CI job runs these" produces, and they belong to the #1895/#1896 cleanup rather than to whichever PR this job runs against first.

Related to #1896

Fixes #1896


dolho:
Rebased onto current `dev` — the PR was conflicting, so there was no `refs/pull/1963/merge` for CI to build and **no checks were reporting at all**. They should run now.

Ready for review.

obasilakis (changes requested):
This is a well-built workflow and the AC coverage is complete — stack via `start.sh` rather than a bespoke compose path, nightly-gated, base-vs-head JUnit attribution, `down -v` between sides, a stated budget with a written escalation ladder, `run-core.sh` untouched. 13 guard tests, all passing. The `#1941` lessons are applied pre-emptively (`fetch-depth: 0`, no `--depth` on the PR-head fetch, the unmerged-path gate), which is good to see land in a second workflow before the first one has even merged. The `#1891` Python pin catching this file at `'3.11'` is that guard doing exactly its job.

Two blocking items, both small.

---

## Blocking 1 — the false ✅ "clean", filed as #2029

Filed separately because it spans three workflows rather than just this PR: **#2029**.

Two paths reach it here. The merge infra branch (`:177`) exits without writing `merge_conflict`; the missing-XML branch (`:278`) exits after `merge_conflict=false` was written but before `steps.diff` runs. Either way `Write status JSON` is `if: always()` (`:303`), `regression` defaults to `false` (`:307`), and the script's final `else` (`:376`) renders:

> ✅ **Live-instance integration suite clean** when this PR is merged into `dev`

`:271-273` states the opposite:

> Fail the job loudly instead; the `comment` step then posts nothing for this PR rather than something false.

It posts something false — the `comment` job is `if: always()` and never inspects `needs.test.result`.

This one is the worst of the three instances. `backend-unit-nightly.yml:301` guards creation with `} else if (status.regression || status.merge_conflict)`, so a false clean there can only overwrite an existing sticky. Here `:396` is a bare `} else {` → `createComment` unconditionally, and this is a new workflow where no PR has a sticky yet. The first bad night posts a fresh green tick to up to 20 PRs.

Fix in #2029 — refuse to write the status file when the verdict is unknown, so the comment job skips that PR.

---

## Blocking 2 — a secret in the job that runs untrusted fork code

`:196`

```yaml
ANTHROPIC_API_KEY: ${{ secrets.E2E_ANTHROPIC_API_KEY || 'placeholder' }}
```

That env is on the step that runs `./scripts/deploy/start.sh` and the pytest suite **from the merged PR tree** — `:234` switches to `head-side` before `boot_stack`, so it is arbitrary shell from the PR, not just test files. `schedule` exposes secrets to fork PRs (unlike `pull_request`), and `discover` enumerates them: **#2009 is an open fork PR targeting `dev` right now.**

This breaks the invariant the header says is inherited verbatim. `backend-unit-nightly.yml`'s only secret is `GITHUB_TOKEN` in `discover` — its test job is genuinely credential-free. This one isn't.

The guard reads as if it covers this but doesn't. `test_the_job_running_pr_code_has_no_write_token` (`:158-161`) asserts only on the `permissions:` block. Token scope is correct; a `secrets.*` reference in the same job is unguarded — which is precisely how this would survive review. Extending that test to assert no `secrets.` other than `GITHUB_TOKEN` appears in the `test` job would close it.

To be straight about severity: **`E2E_ANTHROPIC_API_KEY` does not exist on the repo today** (`gh secret list` shows seven secrets, not that one), so the expression resolves to `'placeholder'` and nothing is currently exposed. The problem is that the workflow is *designed* around that secret existing, and whoever adds it will not re-review this file. Either make the design fork-safe (have `discover` skip `isCrossRepository` PRs) or commit to the placeholder and skip-mark the tests that need a real key — which is the AC below anyway.

---

## Non-blocking

**1. The "pre-merge" stash comment names the wrong branch.** `:122-128` says *"Workspace is unmodified `dev` at this point"*, but the checkout at `:117` has no `ref:` and this repo's default branch is `main`, so a scheduled run copies the trusted diff script from **main**. The security property holds (still trusted, still pre-merge), but the script can be a release behind `dev`'s. `backend-unit-nightly.yml` pins `ref: dev` explicitly — worth matching, and it would make the comment true.

**2. `|| 'placeholder'` silently degrades coverage.** With no key, agent-dependent tests fail on *both* sides and cancel out in the diff, so the job reports clean while a chunk of the suite is inert. #1896's own AC asks to "confirm nothing requires a real external key … or the affected tests must be skip-marked rather than given credentials in CI" — the skip-marking isn't there. Given this issue is specifically about tests that gate nothing, inheriting a silently-inert slice seems worth closing here rather than later. It also interacts with Blocking 2: skip-marking is the fork-safe answer.

**3. Fleet runtime.** 20 PRs at `max-parallel: 3` and ~15-20 min each is roughly 2 hours wall-clock. Fine for a nightly — noting it only because the escalation ladder at `:34-39` is written per-PR and the fleet total is the number that will actually bite.

---

The directory-based membership (`--ignore=unit --ignore=process_engine`, matching `run-core.sh:22`) is the detail that makes this survive #1895, and the note explaining why it therefore didn't need to wait for #1895 is right. `--continue-on-collection-errors` being justified with the actual count of currently-broken imports, rather than added defensively, is the kind of thing that makes this reviewable.

dolho:
@obasilakis — both blocking items are now closed.

**Blocking 2 — the secret in the job that runs untrusted fork code.** Fixed by removing it rather than fencing it:

1. The step commits to the literal `ANTHROPIC_API_KEY: placeholder`, with the reasoning inline so the next person adding a live-LLM key sees why this line isn't the place for it. Tests needing a real key get skip-marked — which is the AC anyway. Since `E2E_ANTHROPIC_API_KEY` doesn't exist on the repo, the expression already resolved to `'placeholder'`, so **nothing about today's runs changes**; what changes is that the workflow is no longer designed around a secret that a future person would add without re-reviewing this file.
2. **`discover` now excludes fork PRs** (`isCrossRepository`), so untrusted code doesn't run in a scheduled job at all — defence in depth for the same exposure, and relevant right now with #2009 open. Skips are logged with `::warning::` rather than dropped silently: a skipped fork is a coverage gap, and a silent skip looks exactly like a pass.

**And the guard that read as if it covered this.** `test_the_job_running_pr_code_has_no_write_token` asserted only on `permissions:` — the token scope was right, and a `secrets.*` reference in the same job was unguarded, which is how this would have survived review a second time. Now: **no `secrets.*` reference of any name** in the test job, plus the fork filter must be present, filtered on, and logged.

```
secret restored ->     1 failed, 17 passed
fork filter removed -> 1 failed, 17 passed
as shipped ->          18 passed
```

**Blocking 1 — the false ✅ "clean"** — fixed in `3405002f`, pushed before this review round (the same shape as the fix in #1952 for #2029): the status step refuses to write when the merge or regression verdict is unknown, and the comment job enumerates `status-pr*.json`, so a PR with no file is skipped and its sticky is left untouched instead of getting a fresh green tick. You were right that this instance was the worst of the three — `createComment` here is unconditional and no PR has a sticky yet, so the first bad night would have posted a green tick to up to 20 PRs.

Merged latest `dev`. Re-review please.

obasilakis (approved):
Both blocking items addressed.

**Blocking 1 (false ✅ clean, #2029)** — the status step refuses to write on an unknown verdict, same shape as the fix in #1952, so the comment job skips the PR rather than rendering a verdict it never computed. The bare `} else { createComment` path can no longer be reached with a defaulted `regression: false`.

**Blocking 2 (secret in the job running PR code)** — closed from both ends, which is better than I asked for. The step commits to the literal `placeholder` with the reasoning inline so the next person adding a live-LLM key doesn't put it here, *and* `discover` filters `isCrossRepository` out of the matrix so untrusted code never runs in a scheduled job regardless. Logging the skipped forks with `::warning::` rather than dropping them silently is the right posture — a silent skip reads as a pass, which is the same class as the false green tick above.

The guard extension is what makes this stick: `test_the_job_running_pr_code_has_no_write_token` asserted only on `permissions:`, which is precisely how a `secrets.*` reference in the same job survived the first review. Asserting no `secrets.*` of any name in the test job, plus fork-filter presence/filtering/logging, closes it. Mutation results in the commit message (secret restored → red, fork filter removed → red, as shipped → 18 passed) are the proof I wanted.

Checks green. Approving.

PR #1964, opened 2026-08-03
refactor(ui): raise dark meta text off the gray-500 floor (#1922)

### LINKED ISSUE #1922: refactor: dark ink-ladder sweep — raise dark-theme meta text off gray-500/600 (design-system p20)
## Summary

Eleven files render dark-theme meta text at gray-500 or darker, breaching the dark ink ladder (style guide §2, principle 20): in dark theme, primary text is gray-100, secondary gray-300, tertiary gray-400 — **gray-500 is the floor**, reserved for disabled states and decoration, never readable meta text. The breaches drop below 4.5:1 AA contrast on gray-800 surfaces.

## Context

Found by the 2026-07-31 design-system audit (dev@022710dc). This is a mechanical sweep — almost every fix is `dark:text-gray-500` → `dark:text-gray-400` — which makes it a good first adoption PR for the ratchet workflow.

## Acceptance Criteria

- [ ] Every occurrence below renders dark-theme meta text at gray-400 or lighter
- [ ] FleetGrid's CSS custom properties (`--gv-faint: #6b7280`, `--gv-ghost: #4b5563`) raised to gray-400/gray-300 equivalents in the dark block
- [ ] No new `dark:text-gray-500`+ meta text introduced (spot-check with the raw-color scanner once wired into CI)
- [ ] Verified visually in dark theme on the affected surfaces

## Technical Notes

- `src/frontend/src/views/enterprise/Audit.vue:610` (+ :658, :842, :847 — `text-gray-500` with no `dark:` override at all)
- `src/frontend/src/components/NavBar.vue:137` (build-info chip)
- `src/frontend/src/views/Login.vue:138` (countdown meta, explicit `dark:text-gray-500`)
- `src/frontend/src/components/operator/QueueCard.vue:24`
- `src/frontend/src/components/operator/QueueItemDetail.vue:66` (`dark:text-gray-600` — below the floor)
- `src/frontend/src/components/operator/NotificationsPanel.vue:226`
- `src/frontend/src/views/Dashboard.vue:86` (tag-count meta in dropdown)
- `src/frontend/src/components/TasksPanel.vue:158` (empty-state helper)
- `src/frontend/src/components/SystemViewsSidebar.vue:75` (agent-count meta)
- `src/frontend/src/components/ReplayTimeline.vue:37` (+ :172, :207, :223)
- `src/frontend/src/components/FleetGrid.vue:603` (CSS vars, meta usage at :721, :813)

dolho:
The dark ink ladder puts primary at gray-100, secondary at gray-300, tertiary at gray-400, with **gray-500 as the floor** — disabled states and pure decoration only. Twelve files rendered readable meta text at or below it: timestamps, counts, empty-state helpers, the build-info chip, the resend countdown, request ids, repo labels. Under 4.5:1 on a gray-800 surface.

## The audit found the loud half; there is a quiet half

`dark:text-gray-500` is greppable. The other breach is a light-theme `text-gray-500` with **no `dark:` override at all** — Audit.vue's loading and empty text (the issue's :610/:658/:842/:847). It reads as neutral in review, but Tailwind reuses the light tertiary verbatim in dark, landing exactly on the floor. Six of those, and a grep for `dark:text-gray-5` cannot see any of them.

## What moved, and what deliberately did not

| | Change | Why |
|---|---|---|
| Readable text | → gray-400 | Tertiary — the dimmest step the ladder allows for text |
| Decoration **below** the floor (separator glyphs, empty-state inbox icon) | gray-600 → gray-500 | Up to the floor, not to gray-400: ornaments should recede |
| Disabled buttons, svg glyphs, icon-only affordances | **unchanged** | Already at the floor, which is exactly where the ladder puts them |

Leaving the last row alone keeps the diff to what the issue describes. Raising icon-only buttons is a judgement call about affordance weight, not an ink-ladder breach.

**HostTelemetry.vue was not on the list.** It is here because its separators render *into* the Dashboard stats bar, beside Dashboard's own — sweeping one and not the other leaves two different inks in one row. I found it by measuring rendered colors, not by reading files: the probe came back with two values where the fix should have left one.

**FleetGrid states its ladder as CSS custom properties**, so no class-level fix reaches it. Dark `--gv-faint` (#6b7280) and `--gv-ghost` (#4b5563) both carry text — the zoom readout, the far-tile placeholder, the empty stat line, and (via AgentTile) the repo label and the dimmed half of a stat line. Both go to gray-400, and therefore **collapse onto `--gv-muted`**. That is the honest consequence of the ladder having exactly one legal step for tertiary text: there is no darker ink left to spend. The light block keeps all three distinct because it has headroom *below* tertiary. Re-separating them in dark needs weight or size, not a darker gray — a design call for #1430, not a color sweep. Flagging it rather than burying it, since the AC read as "gray-400/gray-300" and gray-300 would have made a var named `ghost` render *brighter* than `faint`.

## Enforcement — a ratchet, not a spot-check

AC 3 asks for a spot-check "once the raw-color scanner is wired into CI". `npm run check:tokens` is already wired, so the check goes there as a **ratchet by membership** over the twelve swept files — fleet-wide today would fail on hundreds of pre-existing hits, which is #1430's job:

- **R1** nothing at `dark:text-gray-600`+ — below the floor needs no judgement and has no exceptions
- **R2** no template gray-500+ in a class attribute with no `dark:text-*` companion (the quiet half above)
- **R3** a per-file cap on floor-level ink — lowerable, never raisable
- **R4** FleetGrid's `--gv-*` hex ink, which no class scanner can see

Caps are literals on purpose: a cap regenerated from current source is not a ratchet, it is a rubber stamp. A swept file that moves or vanishes **fails** the check rather than being silently skipped — otherwise the guard retires itself for that file with nobody deciding to. A later sweep adds its files and they can never regress; membership is the whole mechanism, so there is no per-line allowlist to drift.

**Proven both directions.** Against pre-fix source it reports **24 findings**, including all six Audit.vue no-override sites and both FleetGrid vars. Against this branch: clean.

## Visual evidence (AC 4)

Before/after captured on a live stack in dark theme — Dashboard, FleetGrid, operator queue, notifications, agent Tasks, Audit, login countdown. Both passes ran against the **same tree**, differing only by this diff (the sweep stashed for the "before" pass), so nothing else can account for the delta.

Colors were **probed via `getComputedStyle`, not eyeballed**:

| Element | Before | After |
|---|---|---|
| Dashboard stats separators | `rgb(75, 85, 99)` gray-600 | `rgb(107, 114, 128)` gray-500 |
| NavBar build-info chip | `rgb(107, 114, 128)` gray-500 | `rgb(156, 163, 175)` gray-400 |
| FleetGrid `--gv-faint` | `#6b7280` | `#9ca3af` |
| FleetGrid `--gv-ghost` | `#4b5563` | `#9ca3af` |

The separator row is the useful one: **before it returned one value across four elements, after the HostTelemetry fix it returns one again** — the intermediate run returned two, which is what surfaced the twelfth file.

## What I did not verify

The capture harness was a throwaway spec run against a local stack; it is not committed, so these surfaces have no *automated* visual regression. The static guard covers the ink, not the rendering. Light theme is untouched by construction (every edit is either a `dark:`-prefixed class or inside `:root.dark`), and I did not re-shoot light theme to confirm.

Fixes #1922

### DIFF
--- a/docs/memory/design-system.md
+++ b/docs/memory/design-system.md
@@ -67,6 +67,8 @@ The app ships `darkMode: 'class'`. Every primitive and pattern is specced, built
 
 - **Do (dark):** `dark:text-gray-300` for "Last run 12 min ago".
 - **Don't:** `dark:text-gray-500` on anything a user is expected to read.
+- **Also don't:** ship a light-theme `text-gray-500` with *no* `dark:` override. It is not neutral — the light tertiary is reused verbatim in dark, where it lands exactly on the floor. This is the quieter half of the breach and the harder half to catch in review (#1922).
+- **Enforced, per file:** `npm run check:tokens` holds the ladder over the files a sweep has already cleaned (`INK_LADDER_SWEPT` in `scripts/check-design-tokens.mjs`) — nothing below the floor, no missing `dark:` override, and a per-file cap on floor-level ink that may be lowered but never raised. It also reads FleetGrid's `--gv-*` ink, which states its ladder as hex custom properties that no class scanner can see. Fleet-wide enforcement waits on the raw-color ratchet (#1430); each sweep adds its files to the set and they can never regress.
 
 ### Interactive accent (`action-primary`) per theme
 
--- a/src/frontend/scripts/check-design-tokens.mjs
+++ b/src/frontend/scripts/check-design-tokens.mjs
@@ -7,6 +7,8 @@
  *      `dark:*-status-*` reference in the migrated source files uses one of
  *      the defined token names (catches typos that Tailwind would silently
  *      drop).
+ *   3. The dark ink ladder holds in every SWEPT file (#1922) — see
+ *      `checkDarkInkLadder` below.
  *
  * Run via `npm run check:tokens` or directly: `node scripts/check-design-tokens.mjs`.
  */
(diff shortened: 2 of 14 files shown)


dolho:
**Visual evidence — before/after in dark theme:** https://claude.ai/code/artifact/1e59a6d4-17f7-4745-b6f4-147b27f0dd61

Six surfaces side by side (operator queue, fleet grid, dashboard stats bar, audit, agent tasks, notifications), plus the contrast ramp measured against gray-800 and the `getComputedStyle` probe table.

The ratios that motivate the sweep, on gray-800 `#1f2937`:

| shade | ratio | verdict |
|---|---|---|
| gray-300 | 9.96:1 | secondary |
| gray-400 | 5.78:1 | tertiary — passes AA |
| gray-500 | 3.04:1 | floor — large-text only |
| gray-600 | 1.94:1 | fails everything |

vybe (approved):
/validate-pr — all green, approved.

Fully green including `e2e` (auto-triggers on any `src/frontend/` path via the `changes` job — the `ui` label is only a manual opt-in for non-frontend PRs) and `check:tokens`, which is wired in `frontend-build.yml:32`, so the new ink-ladder ratchet is enforced from merge. Security scans clean; refactor, so commit-message-only docs — `design-system.md` updated anyway.

Body rephrased `Related to #1922` → `Fixes #1922` so the issue auto-promotes to `status-in-dev` on merge instead of stranding.

Two non-blocking notes:
- R2 matches `class="…"`, so a single-quoted class attribute is invisible to the ratchet. The swept files are clean today — it's a hole in the guard, not the sweep.
- The `--gv-faint`/`--gv-ghost` collapse in dark is a real consequence you called honestly; worth landing on #1430 rather than living only in this PR body.

Issue #1969, opened 2026-08-03
bug: cron tick suppressed by distributed lock leaves no execution record


vybe:
## Summary

When a cron tick is suppressed by the Redis distributed lock because the same schedule is already running, `_execute_schedule()` logs at INFO and returns without writing anything. No `schedule_executions` row is created, so the suppressed tick is invisible in execution history, the UI and monitoring — it looks exactly as if the tick never fired. Trinity already records this situation correctly for the *other* suppression path (APScheduler `max_instances`), so this is a missing call rather than a design gap.

## Component

Scheduler

## Priority

P2 — no execution is lost incorrectly (suppression is the right behaviour); what is lost is the audit trail that it happened.

## Error

Only trace left behind, at INFO:

```
[INFO] scheduler.service: Schedule {schedule_id} already being executed by another instance
```

No corresponding row in `schedule_executions`. APScheduler still marks the job successful and advances `next_run_at` to the following occurrence.

## Location

- **File**: `src/scheduler/service.py`
- **Line**: 809–822 — `_execute_schedule()`, lock-denial branch
- **Function**: `_execute_schedule()`

```python
lock = self.lock_manager.try_acquire_schedule_lock(schedule_id)
if not lock:
    logger.info(f"Schedule {schedule_id} already being executed by another instance")
    return          # <-- nothing recorded
```

## Root Cause

There are two independent ways a schedule run can be suppressed, and only one of them is audited:

| Path | Mechanism | Audited? |
|---|---|---|
| APScheduler refuses the job | `max_instances=1` → `EVENT_JOB_MAX_INSTANCES` → `_on_job_max_instances()` → `_record_skipped_agent_schedule()` | ✅ writes a `status='skipped'` row + publishes `schedule_execution_skipped` |
| Redis distributed lock denies the run | `_execute_schedule()` lock-denial branch | ❌ bare `return` |

These do not overlap. A **manual** trigger bypasses APScheduler entirely — `_trigger_handler()` runs the execution via `asyncio.create_task`, so APScheduler's per-job instance counter never sees it. When the cron tick for that same schedule then fires, APScheduler happily starts the job (its `max_instances` counter is zero), and the collision is caught one layer down by the Redis lock — precisely the branch that records nothing.

So the common real-world case, *manual trigger shortly before the scheduled time*, is the one case that produces no audit record at all.

## Reproduction Steps

1. Create an agent schedule whose cron fires at a known time.
[...]

Issue #1968, opened 2026-08-03
bug: manual schedule trigger always returns execution_id 'undefined'


vybe:
## Summary

Every manual schedule trigger reports `execution_id: undefined` to the caller, even though the execution starts correctly. The scheduler's trigger handler is fire-and-forget — it spawns the execution as a background task and returns *before* the execution record exists, so no id can be included. The caller therefore cannot correlate its trigger with an execution, poll its status, or fetch its result.

## Component

Scheduler / Backend (schedules router) / MCP Server

## Priority

P2 — the execution runs correctly; only the returned identifier is unusable. Workaround is to poll `list_recent_executions` and guess by timestamp.

## Error

Returned to every MCP `trigger_agent_schedule` caller:

```
Schedule triggered. Execution started with ID 'undefined'.
```

Meanwhile the execution exists and completes normally — it simply was never named back to the caller.

## Location

- **File**: `src/scheduler/main.py`
- **Line**: 152–212 — `_trigger_handler()`; `asyncio.create_task(...)` at 200, response at 205–212 (no `execution_id` key)
- **File**: `src/backend/routers/schedules.py`
- **Line**: 483–489 — return dict passes through five fields, no `execution_id`
- **File**: `src/mcp-server/src/tools/schedules.ts`
- **Line**: 539, 544–545 — reads `result.execution_id`, which is therefore always `undefined`

## Root Cause

`_trigger_handler()` validates the schedule, then fires the execution asynchronously and responds immediately:

```python
asyncio.create_task(self._execute_manual_trigger(schedule_id))

return web.json_response({
    "status": "triggered",
    "schedule_id": schedule_id,
    "schedule_name": schedule.name,
    "agent_name": schedule.agent_name,
    "message": "Execution started in background"
})
```

The execution record is only created later, inside `_execute_schedule_with_lock()` → `db.create_execution()`. At response time no id exists, so none is returned. The backend relays the same five fields, and the MCP tool interpolates the missing key into its success string.

A second consequence of the same ordering: the response is emitted **before** `_execute_manual_trigger()` even attempts the distributed lock. If the lock is denied because the schedule is already running, the trigger silently no-ops while the caller still receives `"status": "triggered"`. A suppressed trigger and a successful one are byte-identical to the caller — the only trace is a scheduler-side WARNING.

## Reproduction Steps

1. Create any agent schedule.
[...]

PR #1974, opened 2026-08-04
fix(scheduler): record who initiated an execution (#1970)

### LINKED ISSUE #1970: bug: scheduler create_execution never populates execution origin columns
## Summary

`schedule_executions` carries five execution-origin columns for audit (`source_user_id`, `source_user_email`, `source_agent_name`, `source_mcp_key_id`, `source_mcp_key_name`), but the scheduler's `create_execution()` never writes them — they are NULL on every scheduler-created row. `triggered_by` records *that* a run was manual, but never *who* triggered it. Once server logs roll, "did anyone trigger this run, and who?" becomes unanswerable from the database.

This is an audit-completeness gap, **not a vulnerability** — no access control depends on these columns.

## Component

Scheduler (database layer) / Backend (schedules router)

## Priority

P2 — executions run correctly; the attribution metadata is simply absent.

## Error

No error is raised. The failure is silent and only visible on inspection:

```
triggered_by         = manual
source_user_id       = NULL
source_user_email    = NULL
source_agent_name    = NULL
source_mcp_key_id    = NULL
source_mcp_key_name  = NULL
```

Observed as NULL across the entire execution history of an actively used schedule (100+ runs, mixed `schedule` and `manual`).

## Location

- **File**: `src/scheduler/database.py`
- **Line**: 373 — `create_execution()`
- **Function**: `create_execution()` — the INSERT column list omits all five `source_*` columns

```python
cursor.execute("""
    INSERT INTO schedule_executions (
        id, schedule_id, agent_name, status, started_at, message, triggered_by,
        model_used, attempt_number, retry_of_execution_id
    ) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
""", (...))
```

## Root Cause

The columns exist in the schema and the backend's own database layer supports them (`src/backend/database.py` accepts `source_user_id` and siblings), but the scheduler is a separate service with its own DB module, and its `create_execution()` was never extended to match. Nor does the signature accept them — there is no parameter to thread a caller identity through:

```python
def create_execution(
    self, schedule_id, agent_name, message,
    triggered_by="schedule", model_used=None,
    attempt_number=1, retry_of_execution_id=None,
) -> Optional[ScheduleExecution]:
```

The identity is available at the point of origin — the backend's trigger endpoint authenticates the caller (user or MCP key) before delegating — but it is dropped at the backend → scheduler hop, which forwards only `schedule_id` in the URL path.

Consequence: for cron-triggered runs NULL is correct and harmless. For `triggered_by='manual'` runs the attribution exists only in MCP-server / backend logs, bounded by log retention, so the durable record cannot answer who initiated a run.

## Reproduction Steps

1. Trigger any agent schedule manually (MCP `trigger_agent_schedule`, or the backend trigger endpoint) as an authenticated user or MCP key.
2. Query the resulting row:
   ```sql
   SELECT triggered_by, source_user_id, source_user_email,
          source_agent_name, source_mcp_key_id, source_mcp_key_name
   FROM schedule_executions ORDER BY started_at DESC LIMIT 1;
   ```
3. Observe `triggered_by = 'manual'` with all five `source_*` columns NULL.

## Suggested Fix

Thread the authenticated caller from the backend trigger endpoint through to `create_execution()`.

1. `src/backend/routers/schedules.py` — forward the caller identity in the POST body to the scheduler's trigger endpoint (currently identity-free):

```python
response = await client.post(
    f"{SCHEDULER_URL}/api/schedules/{schedule_id}/trigger",
    json={
        "source_user_id": current_user.id,
        "source_user_email": current_user.email,
        "source_mcp_key_id": mcp_key.id if mcp_key else None,
        "source_mcp_key_name": mcp_key.name if mcp_key else None,
    },
    timeout=10.0,
)
```

2. `src/scheduler/main.py` — read that payload in `_trigger_handler()` and pass it down through `_execute_manual_trigger()`.

3. `src/scheduler/database.py` — accept the fields and include them in the INSERT:

```python
def create_execution(
    self, schedule_id, agent_name, message,
    triggered_by="schedule", model_used=None,
    attempt_number=1, retry_of_execution_id=None,
    source_user_id=None, source_user_email=None,
    source_agent_name=None, source_mcp_key_id=None, source_mcp_key_name=None,
) -> Optional[ScheduleExecution]:
```

Cron-triggered runs keep passing NULL, which is the correct value for them.

Note this composes with the sibling issue on `execution_id: undefined` — both are fixed most cleanly by creating the execution record synchronously in `_trigger_handler()`, at which point the caller identity is still in scope. Worth sequencing them together.

## Environment

- Observed on a deployed instance running `1d3863d5`
- Confirmed still present at `4d743fdb` (2026-07-27) — `create_execution()` INSERT column list is unchanged

## Related

- `src/scheduler/database.py` — `create_execution()`
- `src/backend/database.py` — backend equivalent, which *does* accept `source_*`
- `src/backend/routers/schedules.py` — manual trigger delegation (identity dropped here)
- Sibling gaps in the same manual-trigger path filed separately (undefined `execution_id`; silent lock-denied cron tick)

dolho:
## Problem

`schedule_executions` has carried five execution-origin columns for audit since AUDIT-001, and `src/backend/database.py` populates them on every path the backend owns. The scheduler is a separate service with its own DB module, and its `create_execution()` listed none of them:

```python
INSERT INTO schedule_executions (
    id, schedule_id, agent_name, status, started_at, message, triggered_by,
    model_used, attempt_number, retry_of_execution_id
) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
```

Nor did the signature accept them, so there was nowhere to put a caller even if one had been forwarded. Every scheduler-created row was written with all five NULL — observed across 100+ runs of an actively used schedule.

`triggered_by='manual'` recorded **that** a human ran something and never **who**. The attribution existed only in backend / MCP-server logs, bounded by log retention, so past a few weeks the durable record could not answer "did anyone trigger this run, and who?".

Not a vulnerability — nothing authorizes on these columns.

## The identity was dropped at three points, not one

| # | Where | What was missing |
|---|-------|------------------|
| 1 | `routers/schedules.py::trigger_schedule` | The delegating `POST` to the scheduler sent **no body at all**. The authenticated caller was in scope right there and never crossed the hop. |
| 2 | `scheduler/main.py::_trigger_handler` | No parameter to receive an identity. |
| 3 | `scheduler/database.py::create_execution` | Nowhere to put it. |

Fixing only #3 — the line the issue points at — changes nothing observable, because nothing upstream would have had anything to pass.

## Fix

An `ExecutionOrigin` value object threaded `backend → _trigger_handler → _execute_manual_trigger → _execute_schedule_with_lock → create_execution`.

One object rather than five parallel parameters at four call depths: five positional siblings is exactly how one of them silently stops being forwarded, which is the failure mode already on display here.

**Cron ticks stay NULL.** Attributing an autonomous fire to, say, the schedule's owner would make the column actively misleading — a blank reads as "unknown", a wrong name does not. Pinned by a test, because it is the half of the behaviour that must *not* change.

## Two more paths the issue's fix would have left blank

Neither is mentioned in the issue, and both would have stayed anonymous after a literal reading of it:

- **Retry** (`triggered_by='retry'`) inherits the original run's origin. A retry has no caller of its own, but it exists only because someone started the original; a chain that drops the initiator makes attempt 1 the only attributable attempt. The read is **fail-open** — an audit lookup must not be able to stop a retry from running.
- **Reminder** (`triggered_by='reminder'`) inherits the provenance #1296 already persisted (`owner_id`, `created_by_email`, `source_agent_name`, `source_mcp_key_id`). This needed those four fields surfaced on the scheduler's `Reminder` model; the mapper guards them the way `_row_to_execution` guards its optional columns, so a table predating them doesn't take the fire path down.

## Hardening the issue didn't ask for

**Boundary validation of the trigger body.** The scheduler's trigger endpoint takes an untrusted JSON body (as it already did for `triggered_by`). `source_user_id` is **dropped rather than coerced** when it is not an int — `bool` IS an `int` in Python, so `True` would have persisted as user `1`: a real account attributed to a run it had nothing to do with. Strings are length-capped and blank-to-`None`, so `""` and `NULL` are not two spellings of "unknown". A malformed body costs attribution, never the run.

**Precedence flipped versus `chat.py`.** The backend prefers the validated `current_user.agent_name` over the raw `X-Source-Agent` header:

```python
"source_agent_name": current_user.agent_name or x_source_agent,
```

Header-first is fine for a collaboration hint; in an audit column it would let an agent-scoped caller pin its run on a sibling agent. The header now fills only the case where it is the sole signal (a user-scoped key, whose `agent_name` is `None`).

**MCP third surface (Invariant #13).** `trigger_agent_schedule` now forwards the same origin headers `chat()` already sends. Without it an MCP-triggered run attributes to the key *owner* but not to which key or which agent fired it — the part that identifies the actor when one human owns many of both. The generic `extraHeaders` hook this required explicitly refuses `Authorization`, so an attribution convenience can't become an auth-substitution surface.

## Compatibility

Backward compatible in **both** rolling-deploy directions:

- old scheduler + new backend → the extra body fields are ignored, `triggered_by` still read as before;
- new scheduler + old backend → a bodyless POST parses to an empty origin, i.e. today's behaviour.

All five parameters are keyword-optional, so an un-migrated `create_execution` call site still inserts a valid row.

## Verification

`tests/unit/test_1970_execution_origin.py` — **27 checks, 25 of which fail against the pre-fix tree**:

```console
$ git stash push -- src/ && pytest unit/test_1970_execution_origin.py -q
25 failed, 2 passed
$ git stash pop && pytest unit/test_1970_execution_origin.py -q
27 passed
```

The 2 that pass either way are deliberate: the cron-stays-NULL guard and the optional-parameter back-compat guard both assert behaviour the fix must preserve, so passing before *and* after is the correct result for them.

The DB assertions run against a real temp SQLite file whose table declares the five columns — a mock would have happily accepted the pre-fix signature, and a table without the columns would let a regressed INSERT pass by having nothing to fill.

Also green: `tsc --noEmit` on the MCP server, and the scheduler-adjacent suites (`test_1808`, `test_schedule_status_observability`, `test_1945`, `test_1713` — 70 tests).

## Acceptance criteria

- [x] `create_execution()` accepts and writes all five `source_*` columns
- [x] The backend forwards the authenticated caller across the scheduler hop
- [x] `_trigger_handler` reads it and passes it down
- [x] Cron-triggered runs keep passing NULL, which is correct for them

## Sibling issues

#1968 (`execution_id: undefined`) and #1969 (lock-denied cron tick leaves no record) are untouched here. The issue notes both are "fixed most cleanly by creating the execution record synchronously in `_trigger_handler()`" — a restructure of the trigger path that this PR deliberately does not attempt, so the attribution fix can land without carrying that risk. The `ExecutionOrigin` object is already in scope at `_trigger_handler`, so it composes with that change rather than competing with it.

Related to #1970

Fixes #1970


dolho:
## /review — self-review (PR #1974, #1970)

**Branch**: `fix/1970-scheduler-execution-origin` → `dev` · 9 files, +834/−23
**Scope**: CLEAN. The issue's suggested fix named 3 files; the diff touches 7 source files, all justified — the two extra non-cron paths (retry, reminder) and the MCP third surface (Invariant #13). No unrelated changes.
**Plan completion**: 4 AC DONE, 0 partial.

Self-review, so discount accordingly. Nothing blocking found.

### [I1] The retry-origin read is a new DB call on the retry path (Confidence: 8/10)

```python
original = self.db.get_execution(original_execution_id)
```

Trading "the retry is unattributed" for "the retry never runs" would be a bad deal, so this is wrapped fail-open with a WARNING. Worth naming because the *first* version I wrote had the `if original` guard but no `try` — a missing row was handled, a failing read was not. Those are different failures.

### [I2] Attribution fields are caller-supplied at the scheduler boundary (Confidence: 7/10)

`/api/schedules/{id}/trigger` has no auth (platform network only), so a caller on that network could post a false `source_user_id`. **Not a new exposure** — `triggered_by` has been caller-supplied on this same endpoint all along — and nothing authorizes on these columns. Mitigated at the boundary: wrong types are dropped rather than coerced (`bool` IS an `int` in Python, so `True` would otherwise persist as real user 1) and strings are length-capped. Recording it explicitly so "audit column" is not later mistaken for "trustworthy identity".

### [I3] The 255-char cap is not derived from the column (Confidence: 6/10)

`source_user_email`/`source_mcp_key_name` are `TEXT` — no DB constraint to match. 255 is a judgement call, not a derived bound. Fine, but it is a magic number.

### Clean categories

- **SQL safety** — the INSERT stays qmark-parameterised; the `_PgCursor` `?`→`%s` rewrite covers the 5 new placeholders.
- **Race conditions** — no new shared state; `ExecutionOrigin` is a frozen-in-practice value object passed by argument, never mutated.
- **Auth boundaries** — no new endpoint. The backend endpoint keeps `AuthorizedAgent`; the three X-\* headers are attribution only, and precedence deliberately prefers the **validated** `current_user.agent_name` over the raw header (the reverse of chat.py) so a caller cannot pin its run on a sibling agent.
- **Credential exposure** — no values logged; the scheduler log line carries the user *id*, not the email, because it lands in Vector-captured platform logs.
- **Enum completeness** — no new enum/status values.
- **Migrations** — none needed; all five columns already exist (AUDIT-001).

### Summary
- Critical: 0
- Informational: 3 — none blocking
- Scope: clean

CI: 18/18 green.

dolho:
## /edge-cases — 3 real bugs found across today's PRs (2 in this one)

Boundary-value analysis + Hypothesis over `ExecutionOrigin.from_payload`, the validation boundary for the scheduler's unauthenticated manual-trigger endpoint. **Both bugs below were in code that already had 100% statement _and_ branch coverage** from `test_1970_execution_origin.py` — full coverage says every line ran, not that every value class was tried.

### [B1] Range was never checked — a payload field could silently lose a run

`from_payload` dropped wrong-**typed** values and capped string **length**, but accepted any `int`. `source_user_id` lands in an INTEGER column, and SQLite **raises** rather than truncating:

```
OverflowError: Python int too large to convert to SQLite INTEGER
```

Verified end-to-end against a real `create_execution`. On this PR's path the exception lands in `_execute_manual_trigger`'s blanket `except` — so the run is logged and **silently lost, after the endpoint already answered `"triggered"`**. (On #1976's path it is caught and 500s.)

The range check is the missing sibling of a guard that was otherwise present — the function already drops unusable input for exactly this reason.

### [B2] `is_empty()` used truthiness — `user_id=0` read as no attribution

`not any((self.user_id, ...))` — `0` is falsy. Hypothesis shrank it to `{"source_user_id": 0}`.

Latent: nothing in `src/` calls it yet. That is exactly why it was worth fixing now, rather than leaving a public helper that lies about a valid value for the first caller to gate on.

### Fixed here

Both, in `8c8f0b1`, with `tests/unit/test_execution_origin_properties.py` — 31 boundary rows + 6 properties (total function, output types match the columns, strings stripped/capped/never-blank, `is_empty` agrees with the fields, parsing idempotent and non-mutating).

Branch coverage of `src/scheduler/models.py`: **100%** (158 stmts, 6 branches, 0 partial) — before *and* after, which is the point.

### Method notes

- Mutation gate not run (`--mutate` is off by default); the coverage-vs-bugs result above already makes the honest case better than a mutation score would.
- `hypothesis==6.161.5` is already a pinned dep in `tests/requirements-test.txt`; the local venv was just stale. No new dependency.

vybe (approved):
Validated via /validate-pr — complete three-surface fix (backend router + scheduler + MCP client/tool), 25/27 regression tests fail pre-fix plus Hypothesis properties, retry inheritance and rolling-deploy compatibility verified both directions, extraHeaders hook refuses Authorization. I added the missing 'Fixes #1970' keyword and resolved the registry.json conflict (rebuild: dev + this PR's two entries); scheduler/service.py auto-merged cleanly against #1975 as predicted. Full CI green post-rebase. Note: stacked #1976 auto-retargets to dev on branch delete. Known follow-up: architecture.md schedule_executions DDL omits the source_* columns (pre-existing drift).

PR #1977, opened 2026-08-04
fix(settings): audit POST /api/settings/ops/reset (#1966)

### LINKED ISSUE #1966: bug: POST /api/settings/ops/reset writes no audit entry — ent#297's audit half stopped one route short
ent#297 / PR #1893 added validation **and** an audit entry (`ops_settings_change`) to `PUT /api/settings/ops/config`. It did not add one to `POST /api/settings/ops/reset`, but both the PR body and the `architecture.md` prose it landed read as though it covered both:

> **Writes are audited.** Neither this endpoint nor `/ops/reset` logged anything, while the generic `PUT /{key}` directly above them does…

> …neither this route nor `/ops/reset` logged anything **before**…

`reset_ops_settings` (`src/backend/routers/settings.py:2661`) is admin-only, deletes rows for every key in `OPS_SETTINGS_DEFAULTS`, and writes no audit entry.

## Scope of what's still unlogged

The retention claim in that prose **does** hold — reset correctly `continue`s over `RETENTION_OPS_KEYS` (#1638), so it cannot shrink a retention window. What it can silently reset with no trace:

- `ssh_access_enabled` — the security-relevant one; resetting it changes whether ephemeral SSH credentials can be minted at all
- `ops_context_warning_threshold` / `ops_context_critical_threshold`
- `ops_idle_timeout_minutes`, `ops_max_execution_minutes`, `ops_alert_suppression_minutes`
- `ops_cost_limit_daily_usd`
- `ops_health_check_interval`
- `ops_log_retention_days`

So the asymmetry ent#297 objected to still exists one route over: the generic `PUT /{key}` audits, `/ops/config` now audits, `/ops/reset` doesn't.

## Fix

Mirror the `/ops/config` call — `platform_audit_service.log(event_type=AuditEventType.CONFIGURATION, event_action="ops_settings_reset", …)` with the `reset` and `skipped` key lists in `details` (keys and counts only; the values are being deleted, and none of these keys is a secret). `request` is already a parameter on the handler, so `actor_ip` / `endpoint` / `request_id` need no signature change.

Either that, or narrow the two prose claims to `/ops/config` — but auditing it is four lines and closes the class rather than documenting around it.

## Acceptance

- [ ] `POST /api/settings/ops/reset` writes a `CONFIGURATION` audit entry naming which keys were reset and which were skipped
- [ ] A test asserts the entry is written (and that retention keys appear under `skipped`, never `reset`)
- [ ] `architecture.md`'s "neither this route nor `/ops/reset` logged anything before" is true after the change

Follow-up to trinity-enterprise#297 · surfaced in review of #1893.

dolho:
## Problem

ent#297 / PR #1893 added validation **and** an audit entry to `PUT /api/settings/ops/config`, and its prose claimed the audit half covered both routes:

> **Writes are audited.** Neither this endpoint nor `/ops/reset` logged anything, while the generic `PUT /{key}` directly above them does…

`/ops/reset` never got one. So the exact asymmetry ent#297 objected to survived one route over:

| Route | Audited before | Now |
|---|---|---|
| `PUT /api/settings/{key}` (generic) | ✅ | ✅ |
| `PUT /api/settings/ops/config` | ✅ (ent#297) | ✅ |
| `POST /api/settings/ops/reset` | ❌ | ✅ |

`reset_ops_settings` is admin-only and deletes a row for every key in `OPS_SETTINGS_DEFAULTS`.

The **retention** claim in that prose does hold — reset correctly `continue`s over `RETENTION_OPS_KEYS` (#1638), so it cannot shrink a retention window. What it could silently revert with no trace is everything else, including **`ssh_access_enabled`** — the setting that decides whether ephemeral SSH credentials can be minted at all.

## Fix

Mirrors the `/ops/config` call, with two deliberate departures from it — neither is a straight copy, and both are the kind of thing that looks like an oversight later if unstated:

**Keys and counts only, no values.** `/ops/config` records the values it set, because "who shortened retention, to what" is the question there. Here every row is being **deleted**, so there is no value worth keeping: the durable fact is which keys reverted to their code default and which were protected.

**Logged unconditionally**, not gated on `deleted` the way `/ops/config` gates on `updated`. There, an empty set means nothing was asked for. Here, an admin pressing reset on already-default settings is a real administrative act — and its absence from the log would be indistinguishable from it never having been attempted, which is this issue's own reporting gap in miniature.

**Distinct action name** (`ops_settings_reset`, not `ops_settings_change`): setting values and deleting rows are different acts, and sharing a name would make them indistinguishable in the very log built to tell them apart.

## One coupling made explicit

The `log()` call sits inside the `try` whose `except` maps to a 500, with **no guard of its own**. A raising sink would therefore turn a completed reset into a 500 the operator retries against already-deleted rows.

That is safe only because `platform_audit_service.log` is documented best-effort and swallows its own errors. Rather than trust the docstring, two tests pin the coupling from both ends: one shows the handler does not guard the call, and one breaks the service's real `db.create_audit_entry` seam and asserts `log()` returns `None` instead of raising. If someone later makes the audit service strict, the second test fails and names this handler as a caller that depends on it.

## Verification

`tests/unit/test_1966_ops_reset_audit.py` — **11 checks, 9 of which fail against the pre-fix tree**:

```console
$ git stash push -- src/ && pytest unit/test_1966_ops_reset_audit.py -q
9 failed, 2 passed
$ git stash pop && pytest unit/test_1966_ops_reset_audit.py -q
11 passed
```

Structural assertions are scoped to the **reset handler's own source slice**, not the file: `/ops/config`'s `ops_settings_change` sits ~30 lines above, and a whole-file grep for "is there an audit call" would have passed against the unfixed tree. (The slice helper also handles the handler being the *last* function in the file — searching unconditionally for a following `def ` raises, which is a harness failure dressed up as a code failure. I hit exactly that on the first run.)

Behavioural coverage drives the real handler with the same `_Admin` dataclass `test_297_ops_settings_validation.py` uses — a bare `MagicMock` has a truthy `.agent_name`, so it reads as an agent key and exercises the wrong branch (the #1816 trap, re-hit in ent#293).

Also green: `test_297_ops_settings_validation`, `test_1638_retention_upgrade_safety`, `test_1644_retention_guard` — **83 passed** across the new file and all three siblings, including a direct assertion that retention keys were not merely *reported* as skipped but genuinely never deleted (a #1638 regression guard).

## Acceptance criteria

- [x] `POST /api/settings/ops/reset` writes a `CONFIGURATION` audit entry naming which keys were reset and which were skipped
- [x] A test asserts the entry is written, and that retention keys appear under `skipped`, never `reset`
- [x] `architecture.md`'s "neither this route nor `/ops/reset` logged anything before" is true after the change

On the third: the sentence became true from the code change alone. I added one clause naming `ops_settings_reset` and the two departures above, so the log's action names are discoverable from the doc rather than only correct in it.

Related to #1966 · follow-up to trinity-enterprise#297 · surfaced in review of #1893

Fixes #1966


dolho:
## /review — self-review (PR #1977, #1966)

**Branch**: `fix/1966-ops-reset-audit` → `dev` · 4 files, +360/−1
**Scope**: CLEAN — the audit call, its test, the registry entry, one `architecture.md` clause.
**Plan completion**: 3 AC DONE.

### [I1] The audit call sits inside the try that maps to a 500, unguarded (Confidence: 8/10)

```python
await platform_audit_service.log(...)   # inside `try:` … `except Exception: raise HTTPException(500)`
```

If `log()` ever raised, a **completed** reset would answer 500 and the operator would retry against already-deleted rows. It cannot today — the service is documented best-effort and returns `None` on failure — so this is safe by a guarantee that lives in another module.

I chose to depend on that rather than add a local guard, because the sibling `/ops/config` call has the identical shape and diverging would be worse than the coupling. Made visible instead: one test shows the handler does not guard, another breaks the service's real `db.create_audit_entry` seam and asserts `log()` returns `None`. If someone later makes the audit service strict, that second test fails and names this caller.

### [I2] `skipped` is `sorted()` in the audit entry but `list()` in the response (Confidence: 6/10)

`details["skipped"] = sorted(RETENTION_OPS_KEYS)` vs the response's pre-existing `list(RETENTION_OPS_KEYS)` — a set, so the response ordering is non-deterministic across runs. Sorting the audit copy is right (a stable record); the response was already like that and I left it. Minor.

### [I3] Values are omitted from the entry — deliberate, and a divergence from the sibling (Confidence: 7/10)

`/ops/config` records the values it set, because "who shortened retention, to what" is its question. Here every row is being **deleted**, so there is no post-state worth recording. Anyone diffing the two entries will notice; that is why it is in the code comment and the doc clause rather than only in my head.

### Clean categories

- **SQL safety** — no new SQL; `db.delete_setting` unchanged.
- **Auth** — `assert_admin(current_user)` unchanged and still first; no new surface.
- **Credential exposure** — keys and counts only, and none of the reachable keys is a secret. Verified against the reset loop's actual key set, not assumed.
- **Retention safety (#1638)** — the `continue` over `RETENTION_OPS_KEYS` is untouched; a test asserts those keys were not merely *reported* as skipped but genuinely never deleted.
- **Enum completeness** — `ops_settings_reset` is a new `event_action` string; `AuditEventType.CONFIGURATION` is existing. Checked for consumers that switch on action names — none; the audit UI filters on `event_type`.

### Summary
- Critical: 0
- Informational: 3 — none blocking
- Scope: clean

CI: 19 pass / 3 skip, green.

vybe (approved):
Validated via /validate-pr — minimal fix mirroring the /ops/config audit pattern; 9/11 regression checks fail pre-fix; the retention-skip contract (#1638/#1644) is untouched and pinned from both ends (retention keys reported under skipped AND asserted never-deleted; audit-swallows-own-failures coupling test-pinned). I added the missing 'Fixes #1966' keyword and resolved the registry.json conflict. Full CI green post-rebase.

PR #1978, opened 2026-08-04
fix(security): put the agent server behind the hardened YAML loader (#1965)

### LINKED ISSUE #1965: bug(security): agent server still parses author-controlled YAML with bare safe_load — outside the ent#314 sweep
`utils/safe_yaml.py` (ent#314, PR #1961) consolidated every author-controlled YAML reader in the backend onto one hardened loader — size cap, alias policy, duplicate-key rejection — and its guard test walks the whole backend with an empty allowlist, so a new bare `yaml.safe_load` there fails the build.

**The agent server was outside that sweep.** `test_no_service_parses_yaml_without_the_shared_loader` walks `_BACKEND.rglob("*.py")` only, and `docker/base-image/agent_server/` still parses author-controlled documents with bare `yaml.safe_load`:

| Site | Document | Backend-side policy for the same document |
|---|---|---|
| `routers/info.py:108,173,252` | `template.yaml` | `BUDGET` (catalog) / `REJECT` (live agent-writable, `credential_requirements_service`) |
| `routers/skills.py:66` | skill frontmatter | `REJECT` (`skill_packaging`) |
| `routers/dashboard.py:205` | `dashboard.yaml` | `REJECT` (via `compatibility/static_checks`) |
| `routers/files.py:244` | `.trinity/persistent-state.yaml` | — |

## Why it matters

The vector ent#314 measured is amplification at **serialization**, not parse: a 416 B level-6 anchor bomb resolves in ~0.001 s and blows up to ~110 MB when something walks the graph. The backend proxies `/info` and `/dashboard` from the agent server, so the same walk happens — just inside the container first, then across the wire.

Lower privilege than the catalog path ent#314 closed (this one needs an already-created agent whose workspace you can write, rather than any `creator`-role user pointing creation at a public repo), which is why #1961 shipped without it. But `template.yaml` inside a running container is exactly the "live agent-writable" document ent#314 assigns the strictest `REJECT` policy on the backend side.

## Shape of the fix

Invariant #5's vendored-mirror pattern, as already done for `credential_paths.py` and `model_context.py`: vendor `utils/safe_yaml.py` byte-identically into `docker/base-image/agent_server/`, add a parity test, wire the six call sites with the policy their backend counterpart uses, and extend the ent#314 AST guard to walk `docker/base-image/` too — otherwise the next agent-server YAML reader lands unguarded exactly the same way.

## Acceptance

- [ ] `safe_yaml.py` vendored into the agent server with a byte-parity test (the `credential_paths.py` shape)
- [ ] All six sites on the shared loader, policy matching the backend counterpart for the same document
- [ ] The ent314 AST guard covers `docker/base-image/` with an empty-by-default allowlist
- [ ] A level-6 bomb in a container's `template.yaml` is refused by `GET /api/agents/{name}/info` with a named error, not a timeout

Follow-up to trinity-enterprise#314 · surfaced in review of #1961.

dolho:
## Problem

ent#314 / PR #1961 put every author-controlled YAML reader in the backend behind one hardened loader — size cap, alias policy, duplicate-key rejection — and its AST guard walks the whole backend with an **empty allowlist**, so a new bare `yaml.safe_load` there fails the build.

That guard walked `_BACKEND.rglob("*.py")` **only**. The agent server ships as its own image and kept parsing the same documents bare:

| Site | Document | Backend policy for the same document |
|---|---|---|
| `routers/info.py` ×2 | `template.yaml` | `REJECT` (`credential_requirements_service`) |
| `routers/skills.py` | skill frontmatter | `REJECT` (`skill_packaging`) |
| `routers/dashboard.py` | `dashboard.yaml` | `REJECT` (`compatibility/static_checks`) |
| `routers/files.py` | `.trinity/persistent-state.yaml` | agent-authored (S4, #383) |

The vector ent#314 measured is amplification at **serialization**, not parse: a 416 B level-6 anchor bomb resolves in ~0.001 s and expands to ~110 MB when something *walks* the graph. The backend proxies `/info` and `/dashboard`, so that walk happens inside the container first and then again across the wire.

Lower privilege than the catalog path ent#314 closed — this needs an already-created agent whose workspace you can write — which is why #1961 shipped without it. But an in-container `template.yaml` is exactly the "live agent-writable" document ent#314 assigns the strictest `REJECT` policy.

## Fix

`safe_yaml.py` vendored **byte-identically** into the agent server (Invariant #5, the `credential_paths.py` shape) with a parity test, since the agent server structurally cannot import `src/backend`. All six sites wired with the policy their backend counterpart uses. The Dockerfile already `COPY`s the whole `agent_server/` directory, so nothing shipped unpackaged (the #1033 class — checked, not assumed).

## Two judgement calls worth your attention

**`/config/agent-config.yaml` gets BUDGET, not REJECT** — deliberately the odd one out. The platform writes it (`agent_service/crud.py`) and bind-mounts it **`mode: 'ro'`**, so the agent cannot author it; it is not the surface this issue is about. REJECT there would risk refusing a *legitimate* document, because `yaml.dump` emits an anchor for any shared object reference — a self-inflicted outage in exchange for no security. The size and duplicate-key guards still apply. Pinned by its own test so the exception is visible rather than looking like a missed site.

**`HardenedYamlError` is a `ValueError`, not a `yaml.YAMLError`.** Every site that already had an `except yaml.YAMLError` needed it named explicitly — otherwise a refused bomb escapes the handler that promised to turn a bad document into a clean answer, and surfaces as the unnamed 500 / timeout AC #4 rules out. `static_checks._parse_yaml` records the identical trap backend-side; `dashboard.py` and `files.py` now name it.

## The guard, not just the fix

The ent#314 AST scan now walks **both** trees, still with an empty allowlist. Fixing the six sites without widening the scan would leave the next agent-server YAML reader landing unguarded in exactly the same way — and a guard scoped to one tree cannot see the sibling tree parsing the same documents. That is the same self-fulfilling-scope failure the guard's own docstring already describes, one directory up.

The widening lives in `test_ent314_hardened_yaml.py` rather than in the new file, because splitting a guard across two files is how the second copy stops being run.

## Verification

Against the **true** pre-fix tree (source changes stashed **and** the vendored copy removed — a stash alone leaves an untracked new file in place and lets the parity tests pass spuriously):

```console
$ pytest unit/test_1965_agent_server_safe_yaml.py unit/test_ent314_hardened_yaml.py -q
11 failed, 21 passed, 4 errors

$ # with the fix
36 passed
```

Byte parity is asserted, but so is **behaviour** — a vendored file that never imports would pass a `cmp` and fail in production, so the vendored copy is loaded and driven through a level-6 bomb, an alias under REJECT, duplicate keys, and an honest document.

**AC #4 end-to-end:** a level-6 bomb in a container's `template.yaml` is refused by *both* `template.yaml` readers, with the expanded graph never reaching the response, and an honest template still serves.

One test needed strengthening mid-review and it is worth flagging: `test_the_metrics_endpoint_refuses_the_same_bomb` originally asserted only `has_metrics is False` — and **passed against the unfixed tree**, because a bomb parses perfectly well under bare `safe_load` and simply yields no `metrics:` key. It now asserts the *named* refusal, which is the only thing separating the two trees there.

Also green: `test_ent314_hardened_yaml`, `test_credential_paths_parity`, `test_1891_python_version_parity`, `test_ent128b2_credential_setup` — **175 passed** across the new file and every sibling parity suite.

## Acceptance criteria

- [x] `safe_yaml.py` vendored into the agent server with a byte-parity test (the `credential_paths.py` shape)
- [x] All six sites on the shared loader, policy matching the backend counterpart for the same document
- [x] The ent#314 AST guard covers `docker/base-image/` with an empty-by-default allowlist
- [x] A level-6 bomb in a container's `template.yaml` is refused with a named error, not a timeout

`architecture.md` Invariant #5 now lists the three byte-identical vendored mirrors and records the "a guard that walks only one of the two trees is not a guard" lesson.

Related to #1965 · follow-up to trinity-enterprise#314 · surfaced in review of #1961

Fixes #1965


dolho:
## /review — self-review (PR #1978, #1965)

**Branch**: `fix/1965-agent-server-safe-yaml` → `dev` · 9 files, +703/−34
**Scope**: CLEAN. Six sites, the vendored loader, the guard widening, one `architecture.md` invariant clause. No unrelated changes.
**Plan completion**: 4 AC DONE.

### [I1] The size cap is applied AFTER the file is read into memory (Confidence: 8/10)

`load_hardened_yaml(text, ...)` takes a `str`, so every call site does `read_text()` first and the 256 KB cap rejects a document already fully resident:

```python
data = load_hardened_yaml(_PERSISTENT_STATE_PATH.read_text(), ...)
```

A 2 GB `template.yaml` is read before being refused. This is the **shared loader's** design, identical on the backend side, not something this PR introduces — and PyYAML reads the whole stream anyway, so the pre-fix `yaml.safe_load(f)` was no better. Recording it because "there is a size cap" reads as protection against large *files*, and it is not; it is protection against large *parses*. A pre-read `stat()` gate would be the real fix, and it belongs in the shared loader, not here.

### [I2] `/config/agent-config.yaml` gets BUDGET while its five siblings get REJECT (Confidence: 8/10)

The deliberate exception, and the one thing in this diff most likely to read as a missed site. The platform writes it and bind-mounts it `mode: 'ro'`, so the agent cannot author it; REJECT would risk refusing a legitimate document because `yaml.dump` emits an anchor for any shared object reference. Pinned by its own test so the exception is asserted rather than incidental.

If the mount ever becomes writable, this must flip to REJECT. That coupling is in the code comment but is not mechanically enforced — a real residual.

### [I3] `HardenedYamlError` inherits `ValueError`, so a broad `except ValueError` upstream would swallow a refusal (Confidence: 6/10)

Checked the four touched files: `info.py` and `skills.py` use `except Exception` and degrade to "no template"/"no frontmatter", which is the pre-existing contract; `dashboard.py` and `files.py` name the class explicitly. No upstream `except ValueError` swallows it silently. Verified, not assumed.

### Clean categories

- **Vendored-mirror drift** — byte parity asserted, *and* the vendored copy is imported and driven through a level-6 bomb, an alias under REJECT, duplicate keys and an honest document. Byte parity alone would pass on a file that never imports.
- **Packaging (#1033 class)** — checked, not assumed: `docker/base-image/Dockerfile:152` copies the whole `agent_server/` directory, so the new module ships without a Dockerfile change.
- **Auth** — no endpoint or gate changed; only how existing handlers parse their input.
- **Credential exposure** — refusal messages carry the error code and line number, never document content.
- **Guard scope** — the ent#314 AST scan now walks both trees with an empty allowlist; fixing the six sites without widening it would leave the next agent-server YAML reader landing unguarded identically.

### Two verification notes against my own work

1. **My first non-vacuity check was wrong.** `git stash push -- docker/` leaves an *untracked* new file in place, so the vendored copy survived and the parity tests passed spuriously. Real pre-fix baseline, with the vendored file removed too: **11 failed + 4 errors of 36**.
2. **One test was vacuous and I caught it late.** `test_the_metrics_endpoint_refuses_the_same_bomb` originally asserted only `has_metrics is False` — and passed against the *unfixed* tree, because a bomb parses fine under bare `safe_load` and simply yields no `metrics:` key. It now asserts the named refusal, which is the only thing separating the two trees there.

Both are the same class: a check that cannot fail is not a check. Worth stating on a security PR, where a green suite is the whole argument.

### Summary
- Critical: 0
- Informational: 3 — none blocking; I1 is a shared-loader property worth its own issue
- Scope: clean

vybe (approved):
Validated via /validate-pr — complete hardening: all six agent-server YAML sites behind the loader, Invariant #5 byte-parity verified by identical blob SHA against the backend copy (landed via #1961) and pinned by parity+behavior tests, AST guard widened to both trees so the gap class can't recur. I added the missing 'Fixes #1965' keyword and resolved the registry.json conflict. Full CI green post-rebase. Deploy note: inert until the base image is rebuilt — release-note the ordering.

Issue #1574, opened 2026-07-11
Wire the agent GitHub PAT for the `gh` CLI + REST API (not just git)


vybe:
## Summary

Trinity injects each agent's resolved GitHub PAT (per-agent override → platform fallback) into the container as the `GITHUB_PAT` env var and bakes it into the git `origin` remote URL (`https://oauth2:<token>@github.com/...`). This authenticates **git** (clone/push/pull) but **not** the GitHub CLI (`gh`) or the REST API: `gh` reads `GH_TOKEN`/`GITHUB_TOKEN` (which Trinity never sets), Trinity never runs `gh auth login`, the token isn't in a `.git-credentials` file, and `gh` isn't installed in the agent base image.

Make the **same** managed token also authenticate `gh` and the REST API — no new token, no new UI, no new endpoint. It is wiring-only: the secret is already present in the container as `$GITHUB_PAT`.

## Context

Real gap hit in practice: agents doing GitHub *CLI* operations (issue triage, `gh api`, PR creation, e.g. GitHub Issues backlog workflows) find that setting the PAT via `set_agent_github_pat` or Settings → GitHub PAT fixes git only. They must fall back to `GH_TOKEN="$GITHUB_PAT" gh ...` per command.

User docs already document the current limitation (`docs/user-docs/integrations/github-pat-setup.md` → "What the PAT Authenticates: Git vs. the gh CLI"). This issue closes the gap itself.

## Acceptance Criteria

- [ ] When an agent has a resolved GitHub PAT, the container also exposes `GH_TOKEN` and `GITHUB_TOKEN` = that PAT (alongside the existing `GITHUB_PAT`), so `gh` and direct REST calls auto-authenticate.
- [ ] Wiring is applied at the same points as `GITHUB_PAT` today: create (`src/backend/services/agent_service/crud.py`), recreate (`src/backend/services/agent_service/lifecycle.py`), the no-restart `.env` propagation (`src/backend/services/github_pat_propagation_service.py`), and container startup (`docker/base-image/startup.sh`).
- [ ] `gh` CLI is installed in the agent base image (`docker/base-image/Dockerfile`) via the official apt repo; base image rebuilt.
[...]

PR #1979, opened 2026-08-04
fix(github-pat): make a global PAT rotation actually reach running agents (#1967)

### LINKED ISSUE #1967: bug: global GitHub PAT rotation never reaches agents — propagation skips agents without .env and never re-templates the git remote
## Summary

Rotating the **global** GitHub PAT in Settings does not reach running agents. The #211 auto-propagation reports success while updating nothing, and even on its success path it never repairs git authentication. Agents that use the global PAT keep authenticating with the revoked token — `git fetch`/`push` and the `gh` CLI all fail — until each container is manually restarted. Nothing in the UI signals this, so the rotation looks clean.

Observed on a fleet where a rotation left every global-PAT agent authenticating with a token that returns HTTP 401, for as long as the containers stayed up (11–13 days).

## Component

Backend / Settings + GitHub PAT propagation (`github_pat_propagation_service`)

## Priority

P1 — the documented purpose of #211 ("agents pick up the new token without a restart") is non-functional in the common configuration, and the failure is silent.

## Error

No exception is raised. The endpoint returns a success payload in which nothing was updated:

```json
{
  "success": true,
  "masked": "ghp_...[REDACTED]",
  "propagation": {
    "total_running": 12,
    "updated": [],
    "skipped": [
      {"agent_name": "[AGENT]", "status": "skipped_no_pat"},
      {"agent_name": "[AGENT]", "status": "skipped_per_agent_pat"}
    ],
    "failed": []
  }
}
```

Downstream symptom inside an affected agent:

```
remote: Invalid username or token. Password authentication is not supported for Git operations.
fatal: Authentication failed for 'https://github.com/[ORG]/[REPO].git/'
```

## Location

- **File**: `src/backend/services/github_pat_propagation_service.py`
- **Functions**: `propagate_github_pat()`, `_propagate_to_agent()`, `_apply_pat_to_env()`
- **Caller**: `src/backend/routers/settings.py:712` (`update_github_pat`)

## Root Cause

Three independent gaps, the first two of which each fully defeat the rotation.

**1. The global path requires a pre-existing `.env` that most agents never have.**

`_propagate_to_agent()` calls `_apply_pat_to_env(..., add_if_missing=False)`. That returns `skipped_no_pat` when `/home/developer/.env` is absent, or present but without a `GITHUB_PAT` line:

```python
if env_content is None:
    if not add_if_missing:
        return "skipped_no_pat"
elif not _env_has_github_pat(env_content) and not add_if_missing:
    return "skipped_no_pat"
```

Agents provisioned from a GitHub template ship a `.env.example` and no `.env`, so **every** such agent skips. The per-agent path (`propagate_pat_to_single_agent`, #1264) passes `add_if_missing=True` and does not have this problem — the asymmetry appears unintentional.

**2. The global path never re-templates the git remote.**

Clones are created as `https://oauth2:<PAT>@github.com/<org>/<repo>.git`, and that URL is persisted in `.git/config` on the agent's workspace volume. Git authenticates from the **URL**, not from `GITHUB_PAT`. `propagate_github_pat()` only rewrites `.env`; it never calls `git_service.update_remote_pat()`.

The service's own docstring for the per-agent path states this explicitly:

> The live git process authenticates from the remote URL (not the env var), so `remote_updated` is the load-bearing part of the immediate fix; the `.env` write only takes effect on the next restart.

So even if gap #1 were fixed, a global rotation still would not restore git access until a restart. Only `propagate_pat_to_single_agent()` calls `update_remote_pat()`.

**3. The propagation result is never surfaced.**

`update_github_pat` returns a populated `propagation` object, but no caller displays it. A rotation in which 100% of agents were skipped is visually identical to one that succeeded.

### Secondary observation

An agent whose per-agent PAT happens to hold a **stale copy of the old global token** is skipped as `skipped_per_agent_pat` on every future rotation, and its dead token is re-injected on every recreate. There is no signal that a stored per-agent PAT has been revoked. Worth considering a validity check or a "stale per-agent PAT" warning.

### Note on the existing convergence path

`check_github_pat_env_matches()` (`services/agent_service/helpers.py:549`) does force a container recreate on the next **start** when the baked `GITHUB_PAT` env differs from the effective PAT, and `docker/base-image/startup.sh` then unconditionally runs `git remote set-url origin` with the fresh token. That path is correct — it just only runs on a cold start, so long-lived agents are the exposure window. The hot-propagation in #211 exists precisely to close that window, and currently does not.

## Reproduction Steps

1. Create an agent from a GitHub template so it is cloned over HTTPS with the global PAT embedded in the remote URL. Confirm the agent has no `/home/developer/.env` (only `.env.example`).
2. Leave the agent running.
3. In Settings, replace the global GitHub PAT with a different valid token.
4. Inspect the API response — `propagation.updated` is empty, all agents report `skipped_no_pat`.
5. In the container, `printenv GITHUB_PAT` and `git remote get-url origin` both still contain the **old** token.
6. Revoke the old token; any `git fetch`/`push` or `gh` call from the agent now fails, with no indication in the UI.

## Suggested Fix

**a. Do not require a pre-existing `.env` on the global path.** Either pass `add_if_missing=True`, or gate on "the agent has a git config" rather than "the agent already has a `GITHUB_PAT` line":

```python
status = await _apply_pat_to_env(
    client, agent_name, base_url, new_pat, add_if_missing=True
)
```

**b. Re-template the live remote on the global path too**, mirroring `propagate_pat_to_single_agent`:

```python
git_config = db.get_git_config(agent_name)
if git_config and git_config.github_repo:
    await git_service.update_remote_pat(
        agent_name, new_pat, git_config.github_repo
    )
```

Factoring the shared ".env write + remote rewrite" body out of `propagate_pat_to_single_agent` and reusing it for both paths would remove the asymmetry at its source.

**c. Surface the propagation result** in the Settings UI — at minimum a warning when `updated` is empty while `total_running > 0`, and a per-agent breakdown of skipped/failed.

## Environment

- Trinity version: `9f766d11`
- Database backend: PostgreSQL
- OS: Ubuntu 22.04

## Related

- `src/backend/services/github_pat_propagation_service.py` (#211, #1264, #1574)
- `src/backend/routers/settings.py` — `update_github_pat`
- `src/backend/routers/git.py` — `set_agent_github_pat` / `clear_agent_github_pat`
- `src/backend/services/agent_service/helpers.py:549` — `check_github_pat_env_matches`
- `docker/base-image/startup.sh` — remote re-templating on cold start

dolho:
## Problem

Rotating the global GitHub PAT in Settings reported success while updating nothing. Agents kept authenticating with the revoked token — `git fetch`/`push` and the `gh` CLI all failing — for as long as their containers stayed up. On the reporting fleet, 11–13 days.

Two independent gaps, each of which **fully** defeated the rotation on its own. Both existed for the same reason: the global path and the per-agent path (#1264) had drifted apart.

**1. The eligibility gate was `.env`-shaped.**

```python
status = await _apply_pat_to_env(..., add_if_missing=False)   # global path
```

Returns `skipped_no_pat` when `/home/developer/.env` is absent. That is *every* agent provisioned from a GitHub template, since those ship `.env.example` and no `.env`. On such a fleet 100% of agents skipped — and the endpoint still answered `success: true`.

**2. `.env` is not where git authenticates from.**

Clones are created as `https://oauth2:<PAT>@github.com/<org>/<repo>.git`, and that URL is persisted in `.git/config` on the workspace volume. Rewriting `.env` changes nothing for the running `git` process. Only re-templating the remote restores fetch/push before a restart — which the per-agent path already did and this one did not.

So even with gap 1 fixed, a rotation still would not restore git access.

## Fix

Both paths now route through one `_apply_pat_to_agent` body (`.env` write **plus** live remote rewrite), so the next divergence has to be deliberate rather than accidental. That drift *is* the bug; deduplicating it is the actual fix, not a tidy-up.

## The part I did differently from the issue, and why

The issue suggests `add_if_missing=True` unconditionally. **I did not do that.** It would inject the global token into every running container, including agents that never touched GitHub — and this module's own docstring says the original gate existed to prevent exactly that. Trading a silent-skip bug for a credential-spraying one is not an improvement.

Eligibility moved to **"does Trinity manage a repo for this agent"** (the issue's own alternative), which targets precisely the population that was being skipped:

| Agent | Before | After |
|---|---|---|
| git config, no `.env` (template-provisioned) | `skipped_no_pat` ❌ | **updated + remote re-templated** |
| git config, has `.env` | updated (`.env` only) | updated **+ remote re-templated** |
| no git config, has `GITHUB_PAT` line | updated | updated *(unchanged)* |
| no git config, no `GITHUB_PAT` line | skipped | skipped *(unchanged — not sprayed)* |

The two rules are a **union**, not a replacement: an agent carrying `GITHUB_PAT` for the `gh` CLI with no Trinity-managed repo is still rotated exactly as before. That row has its own no-regression test, because "fix the gate" is the obvious place to accidentally drop a working case.

## `remotes_updated` is reported separately from `updated`

`updated` alone **overstates the fix**: an agent whose `.env` was rewritten but whose remote was not is still broken for git until it restarts. Collapsing the two would repeat this issue's own mistake — reporting success for a partial effect.

Both degraded shapes now also WARN in the platform log (zero-reach, and updated-without-remote). The failure was silent for weeks; a Settings panel nobody is watching during an incident is not sufficient.

## Gap 3 is already fixed on `dev`

The issue says the propagation result "is never surfaced". **It is** — `Settings.vue` renders the full updated/failed/skipped breakdown with per-agent reasons. The issue was filed against `9f766d11`.

One residual did remain, and is fixed here: `0 of N applied` rendered **success-green** whenever nothing outright failed, so a rotation that reached no agent looked identical to one that worked. It now renders as an error, and the remote-vs-env gap gets its own warning line.

## Not addressed (deliberate)

The issue's "secondary observation" — an agent whose **per-agent** PAT holds a stale copy of the old global token is skipped forever and re-injected on every recreate. Detecting that needs a validity probe against the GitHub API per agent, which is a different feature (network calls in a rotation path, rate limits, a new failure mode) rather than part of this fix. Worth its own issue; flagging rather than silently expanding scope.

## Verification

`tests/unit/test_1967_global_pat_propagation.py` — **13 checks, 7 of which fail against the pre-fix tree**. The 6 that pass either way are the no-regression guards, which is the correct result for them.

```console
$ git stash push -- src/ && pytest unit/test_1967_global_pat_propagation.py -q
7 failed, 6 passed
$ git stash pop && pytest unit/test_1967_global_pat_propagation.py -q
13 passed
```

Also green: `test_1264_per_agent_pat_propagation`, `test_github_pat_fallback`, `test_github_pat_recreation`, `test_ent162_per_user_github_pat`, `test_models_centralized` — **63 passed**. The #1264 tests passing unchanged is the useful signal: the shared-body refactor did not alter the per-agent path's behaviour.

Two structural guards are included because the drift is the root cause: one asserts both callers reach `_apply_pat_to_agent`, one asserts `add_if_missing=False` is not reintroduced on the global path.

## Acceptance criteria

- [x] Global rotation reaches agents with a git config but no `.env`
- [x] Global rotation re-templates the live git remote, so git works without a restart
- [x] The result distinguishes "`.env` written" from "remote fixed", and a zero-reach rotation is loud

Related to #1967 · follow-up to #211, #1264, #1574

Fixes #1967


vybe (approved):
Validated via /validate-pr — well-scoped P1 fix with 7/13 regression tests failing pre-fix, per-agent/per-user PAT architecture preserved (skipped_per_agent_pat gate has its own test), no credential-spraying over-correction. I added the missing 'Fixes #1967' keyword, resolved the registry.json conflict (rebuilt as dev entries + this PR's entry, deduping a PRE-EXISTING duplicate test_181 entry that's been on dev since before today), and full CI re-ran green (initial run was cancelled by a transient runner issue; rerun all-green). Follow-ups filed: #1995 (stale per-agent PAT copy + github-sync.md drift).

PR #1980, opened 2026-08-04
fix(codex): stop synthesizing CODEX_API_KEY, and accept subscription auth (#1971)

### LINKED ISSUE #1971: codex_runtime.py unconditionally sets CODEX_API_KEY, which makes the Codex CLI discard a valid subscription auth.json
## Summary

`codex_runtime.py` (`_execute_codex`) sets **both** `OPENAI_API_KEY` and `CODEX_API_KEY` unconditionally from the same resolved value (source comment: *"some Codex builds also read CODEX_API_KEY... Defensive"*). However, the **mere presence of `CODEX_API_KEY` flips the Codex CLI into API-key auth mode**, silently discarding a valid subscription `auth.json`. Result: subscription-authenticated Codex agents cannot complete any `/task` dispatch through Trinity, even though the same `auth.json` works perfectly when the CLI is invoked directly.

## Environment

- Trinity OSS (self-hosted evaluation deployment, Docker Compose)
- Codex CLI inside the agent container, authenticated via a valid subscription `auth.json` mounted under `CODEX_HOME`
- No real `OPENAI_API_KEY` available by design (subscription auth only)

## Repro (environment isolation)

With `CODEX_HOME` pointing at the mounted config:

1. **Without `CODEX_API_KEY`** (only a placeholder `OPENAI_API_KEY` set): the Codex CLI uses the `auth.json` subscription token and succeeds — real completion returned.
2. **With `CODEX_API_KEY` set** (with or without `OPENAI_API_KEY`): the CLI ignores `auth.json` entirely, attempts true API-key auth against `api.openai.com`, receives `401 Unauthorized: Incorrect API key provided`, retries 5x, then fails the turn.

Since Trinity's dispatch path always injects `CODEX_API_KEY`, case 2 is the only reachable behavior through `/task` — the agent 503s on every dispatch. The only successful Codex completion we obtained was via a manual `docker exec` invocation that deliberately omitted `CODEX_API_KEY`, bypassing Trinity's dispatch entirely.

## Suggested fix

Only set `CODEX_API_KEY` when a real API key is actually configured (or drop the defensive duplicate entirely). A one-line conditional around the injection restores subscription-auth operation without affecting API-key users.

## Secondary observation (non-fatal)

Trinity's relocation of `CODEX_HOME` under `$TMPDIR` trips a Codex CLI safety heuristic: `WARNING: ... Refusing to create helper binaries under temporary dir`. Execution continues, so this is cosmetic — noted only because it surfaces in the same code path.

Happy to provide further detail from our evaluation notes if useful.

dolho:
## Problem

`_execute_codex` set **both** `OPENAI_API_KEY` and `CODEX_API_KEY` from the same resolved value:

```python
"OPENAI_API_KEY": api_key,
"CODEX_API_KEY": api_key,      # "some Codex builds also read CODEX_API_KEY. Defensive"
```

It is not defensive. The mere **presence** of `CODEX_API_KEY` flips the Codex CLI into API-key auth mode and makes it discard a valid subscription `auth.json` — so a subscription-authenticated agent 401'd against api.openai.com, retried 5×, and failed the turn, while the same `auth.json` worked when the CLI was invoked directly. Trinity's dispatch path always injected it, so through `/task` that was the only reachable behaviour.

A "defensive" duplicate that changes behaviour isn't defensive — and the comment claims it was verified in `/verify-local`, which evidently never exercised subscription auth.

## A second defect, from the reporter's own repro

The issue's working case is described as *"only a **placeholder** `OPENAI_API_KEY` set"*. That placeholder is load-bearing, and it shouldn't be:

```python
api_key = _load_openai_api_key()
if not api_key:
    raise HTTPException(503, "OpenAI API key not configured...")
```

A ChatGPT-plan container has **no API key by design** — the issue says so explicitly. So `_execute_codex` refused to start before the CLI was ever invoked, and the only way to run a subscription agent was to set a fake key purely to get past this check. A credential that exists solely to satisfy a gate is the gate being wrong.

**Fixing only the reported half would leave subscription agents working *provided they knew to set a dummy key*.** That isn't the issue's goal ("subscription-authenticated Codex agents cannot complete any `/task` dispatch"), so both are fixed here. Flagging it prominently because it is scope the issue did not ask for.

## Fix

| | Before | After |
|---|---|---|
| Key supplied as `OPENAI_API_KEY` | both vars set → auth.json discarded | only `OPENAI_API_KEY` |
| Key supplied as `CODEX_API_KEY` | both vars set | both set — the operator chose that name |
| Subscription only, no key | **503 before the CLI runs** | dispatches; neither var fabricated |
| Nothing at all | 503 | 503, message naming both remedies |

An operator who exports `CODEX_API_KEY` directly in the container env still keeps it via the existing `**os.environ` spread — their explicit choice isn't second-guessed. What stops is Trinity *inventing* the variable from a key the user supplied under a different name.

Subscription detection is deliberately **existence + non-empty only**. Validating the token is the CLI's job, and a stale `auth.json` should surface as the CLI's own auth error rather than a Trinity 503 claiming no credentials are configured — which would be a false statement about a subscription agent. A zero-byte file and a directory are both rejected; invalid JSON is not.

The 503 message now names both remedies. The old wording said only "inject `OPENAI_API_KEY`", which is precisely what pointed subscription users at the placeholder workaround that caused this issue.

## One thing I got wrong on the way

My first version had `_load_openai_api_key` return `(value, source)` and called it from `_execute_codex`. That **turned 127 existing Codex tests red** — `_load_openai_api_key` is the seam those tests monkeypatch, and moving `_execute_codex` off it silently voided every patch.

The source lookup is now split *alongside* that function rather than replacing it, so `_execute_codex` still calls the same seam and every existing test passes unchanged. There is a test pinning this, because it is invisible until something calls it.

## Not addressed

The secondary observation — `WARNING: Refusing to create helper binaries under temporary dir` from relocating `CODEX_HOME` under `$TMPDIR`. The issue calls it cosmetic and it is: execution continues. Moving `CODEX_HOME` would unpick the #1098 gitignore arrangement that keeps Codex state out of the git-tracked workspace, so it belongs in its own issue rather than riding along in a credentials fix.

## Verification

`tests/unit/test_1971_codex_subscription_auth.py` — **19 checks, 13 of which fail against the pre-fix tree**:

```console
$ git stash push -- docker/ && pytest unit/test_1971_codex_subscription_auth.py -q
13 failed, 6 passed
$ git stash pop && pytest unit/test_1971_codex_subscription_auth.py -q
19 passed
```

The env assertions capture what actually crosses into the child process (Popen intercepted), not what the code appears to set — the whole bug is a variable being present, so anything less than observing the real env would be reasoning about the wrong thing.

Also green: all 6 existing Codex suites (**127 tests, unchanged**), and the wider codex/runtime/agent-server selection (**271 passed**). Full unit suite running.

## Acceptance

- [x] `CODEX_API_KEY` only set when a real API key was configured under that name
- [x] Subscription `auth.json` no longer discarded on the dispatch path
- [x] Subscription agents can dispatch without a placeholder API key

Related to #1971

Fixes #1971

> **Deploy note:** this fix lives in `docker/base-image/agent_server/` — it reaches agents only after `./scripts/deploy/build-base-image.sh` + agent recreate. Old images keep 503-ing Codex subscription agents until rebuilt.


dolho:
## /edge-cases — 1 real (pre-existing) bug found in this PR's targets

Boundary-value analysis + Hypothesis over `_parse_env_value`, `_load_api_key_with_source` and `_has_subscription_auth`. These were worth the pass because #1971 **promoted** them: `_parse_env_value` was a convenience for hand-edited `.env` files, and its result now selects an auth *mode*.

### [B3] A non-UTF-8 `.env` crashed the dispatch

`.env` is hand-edited and credentials get pasted, so one non-UTF-8 byte — a Latin-1 value, a mis-encoded smart quote — is entirely plausible. `read_text()` raises `UnicodeDecodeError` there, and:

```python
>>> issubclass(UnicodeDecodeError, OSError)
False
```

It is a **ValueError**. So `except (IOError, OSError)` around the parse loop never caught it, the error escaped `_load_openai_api_key`, escaped `_execute_codex`, and failed the dispatch outright — instead of resolving the key or returning the honest 503.

**Pre-existing** (both the `read_text()` and the except tuple predate this PR), but #1971 makes this resolver decide the auth *gate* as well as the value, so its failure mode got louder.

Fixed with `errors="replace"` rather than widening the except: a mangled byte on an unrelated line must not hide a valid key further down, and a key that itself contains non-UTF-8 bytes was never usable.

### Documented, not asserted as correct

`.env` resolution is by **line order**, while the process-env branch is by `_API_KEY_VARS` **priority**. So a `.env` listing `CODEX_API_KEY` above `OPENAI_API_KEY` resolves to the former — which, after this PR, also means `CODEX_API_KEY` gets exported and `auth.json` is discarded. Defensible either way (the operator did write that name), so the test pins current behaviour rather than claiming a verdict. Flagging it because the asymmetry is invisible until it bites.

### One finding classified as a wrong test

Hypothesis produced a lone surrogate (`\ud800`) as a `.env` value. That is unproducible — it cannot be written to a real file at all — so it is a bad input, not a bug. Recorded as such in the file rather than silently dropped, and the strategy now excludes the `Cs` category. The *realistic* sibling (a file holding non-UTF-8 **bytes**) is B3 above.

### Coverage

`codex_runtime.py`: **94%** branch coverage across the new + existing Codex suites (470 stmts, 132 branches). Residual misses are subprocess/drain paths that need a live CLI. 146 tests green across 3 randomized orderings.

New: `tests/unit/test_codex_credentials_properties.py` — 32 boundary rows + 4 properties.

vybe (approved):
Validated via /validate-pr — clean pass: root cause (CODEX_API_KEY presence flips CLI auth mode) fixed with a Popen-intercepting regression suite failing 13/19 pre-fix; property pass caught a real UnicodeDecodeError escape. I added the missing 'Fixes #1971' closing keyword and a base-image-rebuild deploy note to the body before merge.

PR #1579, opened 2026-07-12
feat(helper-mcp): standalone Trinity docs Q&A MCP server (#1459)

vybe:
## Summary

- New standalone MCP package **`src/helper-mcp/`** (`@abilityai/trinity-docs-mcp`): an npx-runnable stdio server exposing the public ask-trinity docs Q&A endpoint (DOCS-QA-001) to any MCP client — no Trinity instance, no API key. Pure protocol adapter per the issue directive: no new backend/QA logic, no Cloud Function changes.
- Tools: `ask_trinity` (multi-turn; detects the endpoint's **silent session reset** — verified live: an expired `session_id` returns HTTP 200 + a new session — and warns when context was lost) and `get_agent_requirements` (agent guide fetched live from GitHub with a quick-reference fallback).
- Hardened for public distribution: official `@modelcontextprotocol/sdk` + `zod` only (no fastmcp — smaller supply-chain surface), `console.error`-only logging (stdout is the JSON-RPC channel), Node ≥18 launcher guard, 4k input cap, 50s abort timeout, `redirect:"error"`, non-JSON guard, opaque-string `session_id` (live values exceed 2^53).
- Corpus: `sync-docs-to-vertex.yml` now also indexes `docs/TRINITY_COMPATIBLE_AGENT_GUIDE.md`; the feature flow's stale corpus table corrected (`docs/user-docs/**` incl. the 264-Q&A FAQ was already indexed).

## Changes

- `src/helper-mcp/` — package (client, server, bin stub, tests, smoke script, README)
- `.github/workflows/helper-mcp-test.yml` — CI: build + 26 unit tests + pack-and-run stdio smoke
- `.github/workflows/publish-helper-mcp.yml` — npm publish with provenance (`npm pkg set` version injection — `npm version` errors on the unchanged-version first run; npm upgraded for OIDC trusted publishing; one-time manual first-publish bootstrap documented in the header)
- `.github/workflows/sync-docs-to-vertex.yml` — index the agent guide
- `docs/memory/requirements/mcp.md` — new "Trinity Helper MCP Server (#1459)" entry
- `docs/memory/feature-flows/trinity-docs-qa.md` — corrected corpus table, verified endpoint contract notes (no citations field, silent session expiry, >2^53 session ids), MCP distribution section
- `README.md`, `docs/user-docs/getting-started/help.md` — copy-paste install blocks
- `docs/memory/learnings.md` — new pattern entry (verify feature-flow claims against the authoritative config/live endpoint)

## Test Plan

[...]

PR #1981, opened 2026-08-04
feat(mcp): ask_trinity docs Q&A tool (ent#328)

dolho:
Implements `Abilityai/trinity-enterprise#328`.

## Checked first: not implemented

Verified absent from `dev` (`3f1d4c89`) before starting — no `ask_trinity` anywhere in `src/mcp-server/`, `docs.ts` exposed only `get_agent_requirements`, and `architecture.md` still said `docs.ts (1)`.

## Vendored, not reinvented

`src/helper-mcp` (`@abilityai/trinity-docs-mcp`, shipped by #1579) already contains a hardened adapter for this exact endpoint — and its own header says it was written **"contract-identical with the main Trinity MCP server's `ask_trinity`"**. This is the other side of that contract finally existing, so the sensible move was to reuse it rather than write a second client against the same service.

They are separate npm packages with no workspace between them, so an import isn't available. Trinity's answer to that is a vendored copy plus a parity test — the `credential_paths.py` / `model_context.py` / `safe_yaml.py` shape.

**Parity is behavioural, not byte-for-byte,** and that's a decision worth stating: the helper's module also carries `fetchAgentGuide`/`AGENT_GUIDE_FALLBACK`, and copying those in would give this package a *second*, network-based agent-guide fetcher next to the disk-based `readAgentGuide()` it already has. Copying dead code to satisfy a `cmp` is the wrong trade. `test_1713_scheduler_utils_parity.py` sets the precedent for behavioural-over-byte parity when two copies legitimately differ in text.

## The parity test is non-vacuous — verified by mutation

Both implementations are driven through **one** table of 14 endpoint responses (plain answer, silent session rotation, huge session id, non-SUCCEEDED state, empty answer, HTTP 500 with JSON error, HTML from a Google frontend, 200-with-non-JSON, citations pass-through, and the three input-validation rejections) and must produce identical `{text, isError}`.

Proven to actually catch drift: changing **one word** in the vendored copy (`~30 minutes` → `~99 minutes`) fails **6** tests. My first vacuity check was itself broken — a `sed` whose backticks never matched, so the file was unchanged and "all green" meant nothing. Redone with Python; the result above is the real one.

## Session semantics — the part a reimplementation gets wrong

Both pinned by test, both carried over from the helper's live verification:

- **`session_id` is an opaque STRING.** Real values exceed `Number.MAX_SAFE_INTEGER`. Parsing one as a number corrupts it *silently*, because the corrupted id still returns HTTP 200 — the caller just loses their conversation with no error. The tool schema rejects a numeric `session_id` for this reason.
- **Expiry is SILENT.** An expired id yields 200, `SUCCEEDED`, and a *new* id. A changed id is therefore surfaced as an explicit "previous session expired — context lost" warning, rather than letting the conversation quietly reset under the caller.

## Configurability that actually reaches the container

`ASK_TRINITY_ENDPOINT` is wired through **both** compose files onto the `mcp-server` service and documented in `.env.example`. An env var the container never receives is not configurable — the same "un-flippable on deploy" trap already called out for `MCP_AGENT_CHAT_PULL_ENABLED`, two lines above where this one now sits. Both files verified to parse, with the var on the right service.

Setting it to an unreachable URL effectively disables the tool: it degrades to a structured error, never a crash.

## One build detail worth flagging

The parity test loads the helper through a **computed** specifier rather than a static import. `tsconfig.json` sets `rootDir: ./src` with `include: src/**/*`, so a static cross-package import fails the whole package build with TS6059. The alternatives were worse for their size — excluding `*.test.ts` from tsconfig changes build output for every existing test in the package, and dropping the import turns a live drift-detector into a snapshot that can agree with nothing. `tsx` resolves it at runtime; `tsc` cannot follow it. Commented in place so it doesn't read as cleverness.

## Acceptance criteria

- [x] `ask_trinity` in `docs.ts`: `question` required + `session_id` optional, POSTs to the Cloud Function, returns answer + session_id
- [x] Endpoint configurable via env var, default the public URL
- [x] Graceful failure — unreachable/non-200/non-JSON/timeout/redirect all return structured text, never a crash
- [x] Timeout (50s) and input caps (4k question, 128 session_id, 64k output)
- [x] Registered through the standard audit wrapper — added to the existing `createDocsTools()` group, no special-casing
- [x] `architecture.md` MCP tool table updated, `docs.ts` count 1 → 2

## Verification

**133/133** MCP tests pass (30 new); `tsc --noEmit` clean. MCP-server-only — no backend, no DB, no migration, so Invariant #13's three-surface sync reduces to one.

Related to Abilityai/trinity-enterprise#328


vybe (approved):
Validated via /validate-pr — clean pass: ASK_TRINITY_ENDPOINT wired into the mcp-server service in both compose files + .env.example, vendored client byte-parity tested against helper-mcp (shared 14-case table), audit+connectorDenied inherited via addAllTools. Merging in sequence. Follow-ups: doc drift in requirements/mcp.md + trinity-docs-qa.md (both still say deferred/planned); ent#328 needs a manual status bump now and manual close at release (cross-tracker).

Issue #1766, opened 2026-07-25
Production soak-test of pull/work-stealing coordination (#1081) before default-on


vybe:
## Summary

The push→pull ("pool-based") coordination migration (#1081) is implemented but has never been exercised at full scale under real production load — only local workflows. Before pull mode can be considered for default-on, we need a controlled soak test on a live production instance carrying real agent traffic, watching for lost/phantom executions, lease/re-delivery behavior, and any regressions surfaced by the instance monitor.

## Context

Raised in the 2026-07-24 Intelligence Design weekly. The platform has become notably more reliable on the current (push) path, which paradoxically raises the risk of switching — the current mode is working, so the switch has to be proven, not assumed. The Direction-B rails are already in place (#1402 re-delivery cap + async operator-queue human-gate, #1408 tool-side effect gates, #1084 effect-idempotency); what's missing is production evidence. Plan: update the target instance to the latest build first, flip pull mode for a bounded pilot set of agents (`PULL_MODE_PILOT_AGENTS`), and let the instance monitor report anomalies over a soak window.

## Acceptance Criteria

- [ ] Target production instance updated to the latest build **before** enabling pull mode
- [ ] Pull mode enabled for a bounded pilot set of agents (`PULL_MODE_PILOT_AGENTS`)
- [ ] Soak window observed (several days of real traffic) with the instance monitor active
- [ ] No lost or phantom executions; leases re-queue / poison-park as designed; no slot/backlog drift (canary green throughout)
- [ ] Findings written up — what broke, what held, and whether pull is ready to widen — and linked back to #1081
- [ ] Any regression found is filed; if a blocker, fixed before widening the pilot

## Technical Notes

- Parent epic: Agent Infrastructure (#1045); centerpiece work item #1081.
- Rails/gates: #1402, #1408, #1084; cleanup-collapse #429 sequences after phase 1.
- The physical-occupancy shadow meter and lease reaper are already dark-shipped (inert until an agent is piloted) — this test lights them up under load.
- Priority left at P2 for now; may warrant P1 given #1081 is the reliability spine's centerpiece.
- Owner: Vas.

obasilakis:
## Prep status + two blockers found (2026-08-04)

### Code prep — PR #1982 (open, against `dev`)

Two defects would have made a flag-flip soak unable to produce the evidence the ACs above ask for:

[...]

PR #1982, opened 2026-08-04
fix(pull): make the pilot flag a true either/or + soak measurement set (#1766)

obasilakis:
Prep for the #1766 production soak. Two defects made a flag-flip soak unable to produce the evidence its acceptance criteria ask for. **Part of #1766 — deliberately does not close it**; the soak itself is the ticket.

## 1. The pilot flag was purely additive — push and pull ran in parallel *inside* one agent

`CapacityManager.acquire` had no pilot branch, so a free slot still meant a push and a row only reached the queue on capacity overflow. When one did queue, the backend's own `backlog_service.drain_next` raced the agent's worker through the **same** `db.claim_next_queued` — no double-run (the claim is one atomic UPDATE), but the winner was whoever polled first, and the backend was structurally favoured: it drains on slot release plus a 60s orphan sweep, while a worker idles up to 15s between polls.

Two consequences:

- **Pull coverage was near-zero and drain-biased** on a healthy instance — the soak would have come back green having exercised almost nothing.
- **Two independent capacity counters** (Redis ZSET vs the container's worker pool, both sized `max_parallel_tasks`) meant a pilot could run up to **2x** its configured concurrency — invisible to canary S-02 (counts `ZCARD` only) and S-01 (excludes leased rows by design).

`pull_owns_dispatch` makes a pilot's autonomous work queue-only: never admitted, never drained by the backend, so its worker pool *is* its capacity. This is the pilot-scoped slice of #1081 Phase 5 ("capacity becomes physical").

**Interactive triggers are deliberately excluded** and keep the synchronous push path — the scope cut in `TARGET_ARCHITECTURE.md` Open Question 7, load-bearing because one FIFO ordered by `queued_at` would park a human turn behind N batch tasks until the held connection timed out, and N competing workers could claim two turns of the same session concurrently (the concurrent `--resume` on one JSONL the Redis session lock exists to prevent).

Fails safe to push. Empty allowlist — the default — leaves every path byte-for-byte unchanged.

## 2. E-05 fired on every pull turn past 60s

A pull-claimed row is `running` with a NULL `claude_session_id` **by design**, and `mark_no_session_executions_failed` — the very sweep E-05 watches — already carries `lease_expires_at IS NULL` for exactly that reason. E-05 was flagging the rows that sweep deliberately leaves alone.

Left unfixed it would have broken the "canary green throughout" AC and, worse, trained everyone to ignore canary alerts during precisely the window where a real S-01 or L-03 matters. Mirrors S-01's exclusion; a NULL-lease control still fires. Closes the T3.6 gap in `docs/testing/PULL_MIGRATION_TESTING.md` §8.

## 3. Soak measurement set (docs §9)

Eight queries mapped to #1766's ACs, plus a pre-flight baseline step, abort criteria and rollback. Motivated by a gap worth stating plainly: `instance-monitor` has zero references to leases, `claim_token`, `redelivery_count` or queue depth — it reports generic health and stays green whether pull carries the fleet or does nothing at all, so AC "instance monitor active" is not sufficient evidence on its own.

M1 is load-bearing: if `pulled` is ~0 the soak measured nothing and every downstream conclusion is void.

## Testing

- New `tests/unit/test_1766_pull_pilot_exclusive.py` — 23 tests: the predicate (every autonomous trigger, every interactive trigger, empty allowlist, unresolvable-trigger-set fail-safe), the producer gate (pilot bypasses admission / interactive still pushes / non-pilot unchanged / `queue_in_memory` untouched), and the consumer gate (drain short-circuits ahead of the queued COUNT).
- E-05 lease-exclusion + non-leased control added to `tests/test_canary_invariants.py`.
- **Full unit suite: 7199 passed, 18 skipped, 0 failures.**

## Deployment note

`PULL_MODE_PILOT_AGENTS` is process env (backend restart) **and** the pilot agent must be **recreated** — there is no config-drift predicate for pull mode, so `TRINITY_PULL_MODE` only bakes in at create/recreate. Restart alone leaves the agent not pulling.

### DIFF
--- a/docs/testing/PULL_MIGRATION_TESTING.md
+++ b/docs/testing/PULL_MIGRATION_TESTING.md
@@ -329,8 +329,208 @@ Ran on `gtm-synthesizer` (rebuilt image, `AGENT_AUTH_SECRET` fixed; chosen becau
 
 ## 8. Net remaining before a default-ON decision
 
-Canary lease-awareness (E-05/E-01, §3 T3.6/T3.7) · Tier-6 `effect_guard` `execution_id` injection · ~~G3
-canary-on-PG (#1540)~~ ✅ closed · B6 runtime-verify on the rebuilt image · the ≥2-week soak (#856).
+~~Canary lease-awareness E-05 (§3 T3.6)~~ ✅ **closed by #1766** — E-05 now excludes leased rows, mirroring
+S-01 and the `mark_no_session_executions_failed` sweep it watches (which already carried
+`lease_expires_at IS NULL`); without it E-05 fired on every pull turn past 60s · canary lease-awareness
+E-01 (§3 T3.7) · Tier-6 `effect_guard` `execution_id` injection · ~~G3 canary-on-PG (#1540)~~ ✅ closed ·
+B6 runtime-verify on the rebuilt image · the ≥2-week soak (#856 / #1766, measurement set in §9).
+
+**Also closed by #1766:** the pilot flag was purely additive, so a pilot ran push AND pull concurrently —
+the producer never force-queued (a free slot still meant a push, so rows only queued on overflow) and the
+backend's own `drain_next` raced the agent's worker for whatever did queue. Two independent capacity
+counters meant up to 2x `max_parallel_tasks`, invisible to S-02. The flag is now a true either/or for
+autonomous triggers; interactive turns keep the synchronous path (Open Question 7 scope cut).
+
+## 9. Soak measurement set (#1766)
+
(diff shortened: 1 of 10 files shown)


vybe (approved):
Validated via /validate-pr — clean pass: fail-safe direction is push, empty-allowlist inertness tested both sides, E-05 lease exclusion mirrors the sweep predicate with a control test. Merging in sequence. Follow-ups noted: stale §Capacity & Backlog paragraph in architecture.md (acquire/drain now have a pilot branch), and #1766 status bump is manual (no closing keyword by design).

PR #1985, opened 2026-08-04
fix(security): an unhashed audit chain no longer reports as verified (#1984)

### LINKED ISSUE #1984: bug(security): audit-log verify returns valid:true with checked:0 — an unhashed chain reports as intact
## Summary

`POST /api/audit-log/verify` answers **`valid: true`** for an audit log in which **no entry is hashed at all**. On a live instance with 1,162 audit entries it returns:

```json
{"valid": true, "checked": 0, "first_invalid_id": null}
```

`checked: 0` is the tell, but `valid: true` is the field a human — or a compliance script — reads. An operator asking "was the audit log tampered with?" during an incident gets a green tick whose real meaning is *"there is no integrity data, so the question is unanswerable."*

This is an audit-integrity surface (SEC-001), and `valid: true` is currently returned for three materially different states:

| State | Returned |
|---|---|
| Chain intact, verified | `valid: true` |
| Range empty (no such rows) | `valid: true` |
| **Chain never enabled — nothing hashed** | `valid: true` |

## Component

Backend — `services/platform_audit_service.py::verify_chain`, `routers/audit.py`

## Reproduced on

Local instance, PostgreSQL backend, `dev` @ `3f1d4c89`.

```console
$ psql -c "SELECT count(*), count(entry_hash) FROM audit_log;"
 1162 | 0

$ curl -X POST "…/api/audit-log/verify?start_id=1&end_id=1162" -H "Authorization: Bearer …"
{"valid":true,"checked":0,"first_invalid_id":null}
```

No `system_settings` row for hash chaining exists, i.e. it was never enabled — which is the documented default. The endpoint does not say so.

## Root cause

`verify_chain` skips unhashed entries per-entry, which is correct on its own — a chain enabled midway legitimately has an unhashed prefix:

```python
for i, entry in enumerate(entries):
    if not entry.get("entry_hash"):
        # Entry was written before hash chain was enabled — skip
        continue
    ...
    checked += 1

return {"valid": True, "checked": checked, "first_invalid_id": None}
```

The defect is the **aggregate verdict**, not the skip. When *every* entry is skipped, `checked` stays 0 and the function still returns `valid: True`. "I verified this and it is intact" and "there was nothing to verify" collapse into one answer.

This is the class `docs/memory/learnings.md` (2026-08-03, #1941) already names:

> When a check reports the SAME verdict for every input, that is the signal that it is broken.

## Trinity already does this correctly elsewhere

`GET /api/monitoring/status` faces the identical situation — a summary computed by a loop that is default-OFF (#1121) — and handles it honestly by returning `enabled: false` alongside a stale `last_check_at`, so the reader can interpret the payload. `verify` has no equivalent field.

## Suggested fix

Make "nothing to verify" its own answer. Any of:

1. Return a third state — `valid: null` with `"status": "unverifiable"`, or a `hash_chain_enabled: false` field so `checked: 0` is interpretable.
2. `valid: false` with an explicit reason when `checked == 0` and the requested range was non-empty.
3. At minimum, distinguish *empty range* from *range with rows but no hashes* — today both are `checked: 0`.

Option 1 preserves the existing contract for callers that already branch on `valid`, and mirrors what `monitoring/status` does.

Worth deciding alongside: `POST /api/audit-log/hash-chain/enable` starts hashing from that point forward, so any instance that enables it later permanently carries an unverifiable prefix. That is defensible, but the verify response should make the boundary visible rather than silently averaging over it.

## Acceptance

- [ ] A range containing rows but no hashes does NOT return `valid: true`
- [ ] The response distinguishes "verified intact", "nothing hashed", and "empty range"
- [ ] A genuinely intact hashed chain still returns `valid: true` (no regression)
- [ ] A test asserts the unhashed case, since it is the current default state of every install that never enabled the chain

*Found by live-instance probing, hunting siblings of the #1966–#1971 bug batch (silent-success / vacuous-verdict class).*

dolho:
Closes #1984 — a bug I found by probing the live instance for siblings of the #1966–#1971 batch.

## Problem, as observed on real data

```console
$ psql -c "SELECT count(*), count(entry_hash) FROM audit_log;"
 1162 | 0

$ curl -X POST ".../api/audit-log/verify?start_id=1&end_id=1162"
{"valid":true,"checked":0,"first_invalid_id":null}
```

And in the UI: a green **"✓ Valid · 0 entries"** badge — the tick sitting directly beside the zero that contradicts it.

`valid: true` was returned for three materially different states:

| State | Old answer |
|---|---|
| Chain verified intact | `valid: true` |
| Range empty | `valid: true` |
| **Nothing hashed at all** | `valid: true` |

The third is the **default** for every install that never enabled hashing — so the verdict was effectively independent of the input. That is the class `learnings.md` already names from #1941: *"when a check reports the SAME verdict for every input, that is the signal that it is broken."*

Skipping an unhashed entry is correct on its own — a chain enabled midway legitimately has an unhashed prefix. The defect was the **aggregate verdict** when every entry was skipped.

## Fix

`valid` becomes tri-state: `True` verified · `False` mismatch · **`None` unverifiable**.

**`None` and not `False`**, deliberately. `False` claims tampering — an equally wrong and considerably *louder* lie that would page someone about a breach that did not happen. And a caller doing a plain truthiness test degrades to "not verified", which is the safe direction.

`status` carries the precise verdict (`verified` / `verified_partial` / `tampered` / `unverifiable` / `empty_range`), and `skipped_unhashed` makes a late-enabled chain's permanent unhashed prefix visible rather than silently averaged over — a partially-hashed range can no longer pass as fully verified.

The response now also reports `hash_chain_enabled`, which is precisely what made `checked: 0` uninterpretable.

**Trinity already solves this correctly one endpoint over.** `GET /api/monitoring/status` faces the identical situation — a summary produced by a loop that is default-OFF (#1121) — and returns `enabled: false` beside a stale `last_check_at` so the reader can interpret the payload. This is that pattern applied where it was missing.

## The frontend had to change too

This is the part a backend-only fix would have got wrong. The store did:

```js
this.verifyState = data.valid ? 'valid' : 'invalid'
```

So `valid: null` would have rendered as **"✗ Tamper detected"** — trading a false green for a false *red*, which is worse. It now branches on the tri-state and renders an amber "Unverifiable · no hashes on N entries" badge.

The store also carried its own copy of the bug: an empty entry list set `verifyState='valid', checked:0` **without calling the API at all**. Also gone.

## Verification

`tests/unit/test_1984_audit_verify_unverifiable.py` — **16 checks, 15 of which fail against the pre-fix tree**. 74 green across all audit-related suites.

Scope of what I actually verified, stated precisely:
- **Live pre-fix behaviour**: confirmed over HTTP against the running instance (above).
- **Post-fix behaviour**: unit tests, using the live shape (rows present, zero hashed) as the fixture.
- I did **not** redeploy your instance to test the fixed code in place.

Two of my own mistakes are recorded in the test file rather than quietly fixed:
- Source assertions now strip `//` comments before matching. This fix's own comments quote the old `data.valid ? … : …` while explaining why it went, and an unstripped search *failed on the explanation* — the exact inverse of the #1941 trap, where a comment made a naive search pass vacuously.
- The first fixtures were partial dicts; `_compute_hash` reads several fields with `[]` rather than `.get`, so the hasher raised `KeyError` and it looked like a code bug. There is now one complete row builder.

## Acceptance criteria

- [x] A range containing rows but no hashes does NOT return `valid: true`
- [x] The response distinguishes "verified intact", "nothing hashed", and "empty range"
- [x] A genuinely intact hashed chain still returns `valid: true`
- [x] A test asserts the unhashed case — the current default state of every install that never enabled the chain


vybe (approved):
Validated via /validate-pr — the fix itself is exemplary (honest tri-state semantics, 16-check named regression test across model/service/store/view; verified an all-unhashed range can never report valid:true). The one blocker from validation — tests/test_audit_log_unit.py::test_verify_chain_empty_range still asserting the old vacuous empty-range contract (invisible to PR CI which runs unit/ only, but breaks tests/run-core.sh on dev) — I fixed on the branch in the merge commit, plus the stale audit-trail.md response example. Full CI green post-rebase.

Issue #407, opened 2026-04-19
bug: agent-server.py spins at ~83% CPU after claude CLI subprocess becomes defunct


vybe:
## Summary

After a scheduled task runs, the `claude --print --output-format stream-json --verbose …` subprocess inside an agent container can exit leaving a `<defunct>` zombie child. The parent `python3 /app/agent-server.py` does not reap it and begins spinning at ~83% CPU indefinitely. The FastAPI server inside the agent container then stops responding to **every** HTTP request (including `/`, `/api/session`, `/api/files/download`), even though the container remains in `running` state.

The backend's `services.agent_client` opens its circuit breaker for the affected agent and keeps re-opening it roughly every 35 s. Scheduled and ad-hoc tasks against the agent then get watchdog-terminated and recorded as failed — often misclassified downstream as "Auth failure (fallback detection): Subscription token may be expired or revoked" when the real cause is the SIGINT from the watchdog.

Other agents on the same host — rebuilt from the same base image — are unaffected.

## Component

Agent Runtime / Base Image (`docker/base-image/agent_server`)

## Priority

P1 — major feature broken for a single agent (all HTTP endpoints hang), no in-agent recovery. Workaround: restart the container.

## Error

No explicit error is logged by the agent-server at the moment of failure — the process simply stops making forward progress. The last agent log line before the hang is:

```
INFO:agent_server.services.process_registry:[ProcessRegistry] Registered execution <exec-id>
```

On the backend side the symptom is a repeating pattern:

```
[Slots] Agent '<agent>' acquired slot 1/3 for execution <exec-id> (TTL=2100s)
[TaskExecService] Calling agent <agent> /api/task (timeout=1810.0s, ...)
Circuit OPENED for agent <agent> after N failures
Circuit CLOSED for agent <agent> (recovered)
Circuit OPENED for agent <agent> after N+1 failures
... (every ~35s indefinitely, until the container is restarted)
```

And, after the watchdog terminates the stale execution, a misleading error surfaces in the agent-server log:

```
ERROR:agent_server.services.claude_code:[Headless Task] Auth failure (fallback detection): Subscription token may be expired or revoked. Generate a new one with 'claude setup-token'.
```

— which is a false positive caused by SIGINT during stream-json reading, not an actual auth issue.

## Location

- **File**: `docker/base-image/agent_server/services/claude_code.py` — subprocess / stream-json reader loop; does not appear to drain stdout/stderr and reap the child on all exit paths
[...]

PR #2010, opened 2026-08-05
fix(agent-server): rebuild the execution env per spawn so a removed .env key stops applying (#1999)

### LINKED ISSUE #1999: bug: credential injects mirror `.env` into the agent-server long-lived process env — a removed key keeps reaching every execution, invisible to /proc and docker exec
## Summary

The agent-server's credential-inject endpoints mirror every `.env` key into the **long-lived
agent-server process's `os.environ`**, and every execution is spawned with that environment. The
`.env` file itself is **never** loaded into the agent-server process. So the file an operator
inspects and the environment executions actually receive are two independent channels that sync
only inside one code path.

Consequence: after any `.env` edit that does not go through `/api/credentials/*` (SSH, `docker exec`,
or **an agent execution editing its own `.env`**), a value removed from the file **keeps applying to
every execution** until the container restarts. The stale value is invisible to `/proc/<pid>/environ`
(an exec-time snapshot — runtime `os.environ` mutations never appear there) **and** to `docker exec`
(which gets the container baseline, not the server process env). Both standard inspection routes
agree with the file and disagree with reality.

Security framing: **credential revocation silently fails.** Removing a leaked or rotated key from
`.env` does not stop executions from using it, and nothing an operator can inspect reveals that.

## Context

Observed on a production instance, where a retired feature-flag variable re-entered a running
agent-server via a credential push, survived a same-day `.env` cleanup, and silently altered the
behaviour of two scheduled executions over the following ~19 hours. Root-causing it took hours
precisely because every inspection tool reported the variable as unset.

Reproduced on `dev` @ `3f1d4c89`. The mechanism is present in **every released version** — both
mirror blocks date to the initial commit.

## Reproduction

Verified end-to-end against a local agent with real Claude Code executions.
Probe: `python3 -c "import os; print('CANARY:', repr(os.environ.get('CANARY_VAR')))"`

| Step | `.env` on disk | Execution sees |
|---|---|---|
| Baseline | absent | `None` |
| `POST /api/agents/{name}/credentials/inject` → `{".env": "CANARY_VAR=ghost123\n"}` | present | `'ghost123'` |
| Delete the line via `docker exec` (non-mirroring path) | **gone** | **`'ghost123'` ← ghost** |
| `tr '\0' '\n' < /proc/<agent-server-pid>/environ \| grep CANARY` | — | no match |
| `docker exec <container> python3 -c "...os.environ.get(...)"` | — | `None` |
| Re-inject the cleaned `.env` (key line **absent**) | gone | **`'ghost123'` ← still ghosted** |
| Re-inject with key **present but empty** (`CANARY_VAR=`) | `CANARY_VAR=` | `''` (cleared) |
| Container restart | `CANARY_VAR=` | `None` |

Two results worth calling out:

**The obvious mitigation usually fails.** The mirror loop only *sets* keys present in the file — it
has no delete phase — so a key **deleted** from `.env` is never removed from `os.environ`. Only
blanking the value (`KEY=`) clears the ghost. "Delete the stale line, then re-push credentials" is
the intuitive operator action and it leaves the runtime poisoned while returning success.

**The last row proves the channels are independent.** At restart the file literally contained
`CANARY_VAR=`, yet the fresh process had no such key — confirming `.env` is never loaded into the
agent-server process at all.

## Acceptance Criteria

- [ ] A key removed from `.env` no longer reaches subsequently-spawned executions (any write path:
      credentials API, SSH, `docker exec`, or an agent editing its own `.env`)
- [ ] The `.env` file is the authoritative source for credential values in the execution environment
- [ ] `/api/credentials/reload-token` still rotates `CLAUDE_CODE_OAUTH_TOKEN` for the next execution
      (#1089 must not regress — see Technical Notes)
- [ ] `EXECUTION_TAG_NAME` still reaches every spawned process (#407 orphan sweep)
- [ ] `AGENT_RUNTIME`, `ANTHROPIC_API_KEY`, `GOOGLE_API_KEY` still resolve for platform code that
      reads them from the process env
- [ ] The `/api/credentials/inject` mirror removes keys that are absent from the newly-written
      `.env`, so re-injecting a cleaned file clears a deleted key
- [ ] Regression test: inject a key, remove it from `.env` out-of-band, assert a new execution does
      not see it
- [ ] Applied uniformly across all runtimes (Claude Code, Codex, Gemini)

### Mechanism

**Mirror sites** — `docker/base-image/agent_server/routers/credentials.py`:
- `POST /api/credentials/update` — `os.environ[var_name] = str(value)` for every submitted key
- `POST /api/credentials/inject` — re-parses the written `.env` and mirrors every line

**Spawn sites** — all five pass `env={**os.environ, EXECUTION_TAG_NAME: ...}`, i.e. an explicit
snapshot of the mutated runtime env:
- `services/claude_code.py`, `services/headless_executor.py`, `services/codex_runtime.py`,
  `services/gemini_runtime.py` (×2)

(Historically these were bare `Popen` calls with no `env=`; the `{**os.environ}` form was added for
#407 orphan-tagging and is behaviourally identical for this issue.)

**No counterweight** — there is no `dotenv` anywhere in `agent_server/`, and
`docker/base-image/startup.sh` exports only `GITHUB_PAT` and `CLAUDE_CODE_OAUTH_TOKEN`, never the
general `.env`.

Note `docker/base-image/agent_server/routers/files.py` is **not** a divergence path — `.env` is in
its `EDIT_PROTECTED_PATHS` list and always has been.

### ⚠️ The obvious fix breaks #1089

The natural fix — spawn with `env={**INITIAL_ENV, **parse(home/.env)}` — **would silently break
subscription-token rotation.** `POST /api/credentials/reload-token` rotates the token by mutating
*only* the process env and deliberately does not write `.env` ("the subscription token is not a
.env credential"). Under that fix the rotated token is in neither source, so every execution after a
rotation would receive the **stale** token.

Suggested shape instead:

```python
env = {
    **INITIAL_ENV,                      # boot snapshot captured at module import
    **RUNTIME_OVERRIDES,                # explicit allowlist: CLAUDE_CODE_OAUTH_TOKEN, ANTHROPIC_API_KEY, ...
    **parse_env_file(HOME / ".env"),    # file is authoritative for credentials
    EXECUTION_TAG_NAME: execution_id,
}
```

with `RUNTIME_OVERRIDES` a small named dict the reload-token endpoint writes to, making precedence
explicit rather than implicit in global process state.

Simply deleting the mirror blocks is **not** viable: it would break #1089 and the credential
sanitizer refresh.

### Amplifiers

- **Boot re-push.** `inject_assigned_credentials` (`src/backend/services/agent_service/lifecycle.py`)
  runs on every agent start and pushes `.credentials.enc` through the mirroring `/inject` endpoint,
  so a stale snapshot re-poisons the runtime env *and* rewrites `.env` on every boot. (Skipped for
  subscription-mode agents, which is why a restart clears the ghost for those.)
- **Fleet-wide pushes.** `src/backend/services/github_pat_propagation_service.py` reads an agent's
  *current* `.env`, patches one line, and re-posts the **whole file** — mirroring every key in it.
  A single global PAT rotation can therefore re-arm this across the fleet. Related: **#1967** is an
  open bug in that same propagation path; **fixing #1967 will make this fire more often.**
- **Sanitizer interaction.** `agent_server/utils/credential_sanitizer.py` builds its redaction set
  from the union of `os.environ` and `~/.env`, so a ghosted value matching a sensitive-name pattern
  (`.*TOKEN.*`, `.*KEY.*`, …) is scrubbed from the very transcripts an operator would use to spot
  it — making a ghosted API key *more* invisible than a ghosted plain variable.

### Diagnosability (suggested, cheap, high value)

Extend `GET /api/credentials/status` with a per-key drift report — **key names only, no values**:
`{key, in_file, in_process_env, equal}`. This turns the entire class from a multi-hour forensic hunt
into a single API call. The agent-server's uvicorn access logs also carry no timestamps, which made
correlating the originating credential push impossible after the fact.

dolho:
## Summary

`.env` and the environment executions actually receive were two independent channels. The credential endpoints mirrored every submitted key into the **long-lived agent-server process's** `os.environ`; all five spawn sites passed `env={**os.environ, ...}`; the file was never read back into that process. One code path, one direction, present keys only — **no delete phase**.

So a key removed from `.env` by any non-mirroring write path (SSH, `docker exec`, or an agent editing its own `.env`) kept reaching every subsequently-spawned execution until container restart, and was invisible to `/proc/<pid>/environ` (an exec-time snapshot) and to `docker exec` (the container baseline, not the mutated process env). **Credential revocation silently failed**, and every inspection route confirmed the wrong answer.

## The fix

`docker/base-image/agent_server/services/execution_env.py` — the execution environment becomes a pure function of three inspectable inputs, rebuilt per spawn instead of accumulated in process state:

```python
env = build_execution_env({EXECUTION_TAG_NAME: execution_id})
#   INITIAL_ENV        container baseline captured at import — what `docker exec` shows
#   .env               parsed fresh at every spawn — authoritative for credentials
#   RUNTIME_OVERRIDES  values that are deliberately not .env credentials (#1089)
#   extra              last, so EXECUTION_TAG_NAME (#407) can never be displaced
```

A key that is in neither the baseline nor `.env` now has no path into a spawned process, so the class cannot recur.

### Two deliberate divergences from the issue's sketch

**Overrides are applied AFTER the file, not before.** `/api/credentials/reload-token` rotates the subscription token *without* writing `.env` (by design — "the subscription token is not a .env credential"). With file-last precedence, a stale `CLAUDE_CODE_OAUTH_TOKEN` that ever reaches `.env` — a hand edit, a credential export that captured it — would beat the rotation and re-open #1089 from the other side. An explicit hot-reload is the most recent and most specific signal; it wins. `None` means **force-unset**, which a plain dict merge cannot express and `remove_api_key` requires.

**The process mirror stays** — deleting it would break the in-process readers the issue names (error classifier's `ANTHROPIC_API_KEY`/`CLAUDE_CODE_OAUTH_TOKEN` probes, `AGENT_RUNTIME`, `GOOGLE_API_KEY`, and the sanitizer's redaction set). It gains the missing delete phase: a key this process mirrored that the new `.env` no longer contains is **restored to its container-baseline value**, or popped when it has none. Only keys the mirror itself wrote are eligible, so a baseline-only key is never deleted.

### One behaviour narrowing, on purpose

`.env` may no longer set `PATH`, `LD_PRELOAD`, `LD_LIBRARY_PATH`, `LD_AUDIT`, `PYTHONHOME`, `PYTHONPATH`, `PYTHONSTARTUP`, `BASH_ENV`, `ENV`, `IFS`. These carry no credentials — they redirect what the child *executes or loads*. `.env` is agent-writable and is now read at **every spawn**, a wider trigger than the old backend-only mirror, so a prompt-injected workspace could otherwise repoint the runtime binary. Ignored with a warning, never fatal.

### Diagnosability

`GET /api/credentials/status` gains `env_drift`: `{key, in_file, in_process_env, equal}` per key — **names only, never values**. The issue took hours to root-cause because every available inspection route read a different channel than the enforcement path; this is the route that compares them.

## Acceptance criteria

| AC | Where |
|---|---|
| Removed key stops reaching executions (any write path) | `build_execution_env` parses `.env` per spawn · `test_key_deleted_out_of_band_does_not_reach_the_next_execution` |
| `.env` authoritative for credential values | file layer applied over the baseline · `test_env_file_is_authoritative_over_the_baseline` |
| `/reload-token` still rotates for the next execution (#1089) | `set_runtime_override`, applied after the file · 3 precedence tests incl. the stale-token-in-file case |
| `EXECUTION_TAG_NAME` still reaches every process (#407) | `extra` applied last · `test_execution_tag_cannot_be_displaced_by_the_env_file` |
| `AGENT_RUNTIME` / `ANTHROPIC_API_KEY` / `GOOGLE_API_KEY` still resolve in-process | mirror retained; baseline restored on delete · `test_a_baseline_key_is_restored_not_deleted` |
| `/inject` mirror removes keys absent from the new `.env` | `sync_process_env` delete phase · `test_reinjecting_a_cleaned_env_clears_the_process_mirror` |
| Regression test for the reported scenario | `TestRemovedKeyStopsReachingExecutions` |
| Uniform across Claude Code, Codex, Gemini | all five spawn sites wired · static guard `test_every_spawn_site_builds_its_env_through_the_helper` |

## Test plan

- [x] `tests/unit/test_1999_execution_env.py` — **36 tests**: the issue's reproduction, baseline restore, precedence (incl. both #1089 no-regression directions), tolerant `.env` parsing, unreadable/oversized files degrading to the baseline rather than raising on the spawn path, protected keys, the name-only drift report, and a static guard against a sixth spawn site reintroducing `{**os.environ}`
- [x] Mutation-verified — each of (a) reverting one spawn site, (b) removing the delete phase, (c) applying overrides before the file fails **exactly** its own tests (1, 2, 3 failures respectively), never silently passing
- [x] 871 adjacent tests green (`-k "credential or agent_server or 1089 or subscription or codex or gemini or claude_code or 407 or orphan"`)
- [x] `.env` parser matches the writer's own escaping — `.strip('"')` eats a legitimately trailing quote, so it is a one-matched-pair unquote instead

## Not included

The issue's **amplifiers** are noted but out of scope: the boot re-push (`inject_assigned_credentials`) and the fleet-wide PAT propagation still rewrite `.env` from the backend's stored archive. That is the intended source of truth and the result is now *visible in the file*, so it is no longer a ghost — but it does mean a hand-edit to `.env` is still overwritten on next boot. #1967 is open on that same propagation path.

Closes #1999


dolho:
Rebased onto `dev` — it had gone conflicting after #1901 and others landed. Only `tests/registry.json` and `docs/memory/learnings.md` conflicted, both append-only on each side; unioned with ours last and the registry re-serialized from parsed JSON (never line-spliced, so the separating comma can't be lost).

Verified wider than the conflicted files this time, which is the lesson from #1976 earlier today: git auto-merged two `ExecutionOrigin` class definitions there without reporting a conflict, and my post-rebase test filter was narrower than the blast radius, so 129 tests passed over a broken file. So: swept every changed `.py` for duplicate top-level definitions (clean), confirmed the registry diff has **zero deletions**, and ran 1224 tests across `1999 / ent127 / credential / agent_server / execution / 1089 / codex / gemini`.

Unblocking this also unblocks #2023 — the `.env` quote-escaping asymmetry can only land on top of the parser this PR introduces.

vybe (changes requested):
The core mechanism is right and the argument for it is good — `build_execution_env` rebuilding from `dict(INITIAL_ENV)` → `.env` → overrides → extra genuinely gives the deletion property, all five spawn sites are converted, and `sync_process_env`'s delete phase correctly gates on `_MIRRORED_KEYS` so baseline-only keys are restored rather than popped. Overrides-after-file for #1089 is the right ordering.

Two things block.

**1. `env_drift` names credential variables on the one credential endpoint deliberately left on the looser gate — and this repo documents why that was safe.**

`agent_server/routers/credentials.py` adds `env_drift` (one entry per `.env` key, **by name**) to `GET /api/credentials/status`. Its backend proxy is `get_authorized_agent_by_name` with **no** `reject_agent_principal`, while every sibling (`/credential-requirements`, `/inject`, `/export`, `/import`) is owner + human-only.

I verified the rationale is written out in `src/backend/routers/credentials.py` right above the sibling route:

> *"What it would disclose is a targeting map, not a status light: it names `STRIPE_SECRET_KEY` per agent and says which are populated … A prompt-injected agent gets it with one curl."*
> *"`/credentials/status` gets away with the read gate because it returns a COUNT and names zero variables."*

That last sentence is exactly what this PR falsifies. Reachable by any user the agent is merely **shared** with, and by any **agent-scoped MCP key** — `get_authorized_agent_by_name` resolves it to the owner carrying the owner's role, and the MCP tool `get_credential_status` does a full `JSON.stringify` pass-through, so on a default admin-owned install that is the whole fleet. Arguably wider than the ent#127 checklist it is now looser than: that names *declared* requirements, `env_drift` names what is actually in `.env`, including undeclared variables.

Please move the drift report behind `get_owned_agent_by_name` + `reject_agent_principal` (own route, or fold into `/credential-requirements`) and update that comment either way. A diagnostic added by a security fix shouldn't widen the surface the fix is about.

**2. `/update` now delivers quote-bearing values corrupted.**

Pre-PR, `/update` mirrored the raw submitted value into `os.environ` and the child inherited it. Post-PR it round-trips through the writer's escaping and back through a parser that strips but does not unescape:

```
raw 'pa"ss'    -> file 'K="pa\"ss"'  -> delivered 'pa\"ss'   CORRUPTED
raw "'quoted'" -> ...                  -> delivered 'quoted'    quotes eaten
raw 'tail"'    -> ...                  -> delivered 'tail\'     CORRUPTED
```

The "byte-faithful, don't move the ent#127 predicate under a security fix" reasoning is sound about the *predicate*, but not about *delivered values*, and `/update` was the one path that never had this loss. It is a silent auth failure of exactly the class #1999 is about. Either fix writer + parser + `_env_pairs` together here, or state the exposure and land it as an immediate follow-up. (Note #2030 moves the writer/reader pair — worth coordinating rather than landing these independently.)

**Non-blocking, but worth addressing:**

- **The wiring is untested.** The static guard covers the 5 spawn sites; nothing covers the 4 new router call sites. Deleting the two `set_runtime_override(...)` lines the AC table cites as "what keeps #1089 working" leaves the suite green.
- **The static guard walks one tree** (`agent_server/services/` only). `routers/brain_orb.py` already spawns a subprocess, and a future `routers/` spawn site with `{**os.environ}` wouldn't be caught — the #1965 lesson ("a guard that walks only one of the two trees is not a guard") one directory over.
- Deleting `test_decode_divergence_is_deliberate` hides a behaviour change: the agent server went from *raising* on a non-UTF-8 `.env` to silently substituting U+FFFD into a credential value.
- `PROTECTED_KEYS` omits the Node equivalents — `NODE_OPTIONS=--require …` is `LD_PRELOAD` for Node, and both runtimes are Node processes. Same for `GIT_SSH_COMMAND`/`GIT_CONFIG_*`.
- New torn-read window: `.env` is written non-atomically and now read on the spawn path from a worker thread. Fail-safe (degrades to baseline), but tmp + `os.replace` closes it — the codebase already uses that idiom for `sync-state.json`.
- Stale PR body: the test-plan bullet describing a "one-matched-pair unquote" is the opposite of what ships, and the test count is 40, not 36.

Also needs a rebase — conflicts on `learnings.md` and `tests/registry.json` only.

vybe:
Re-validated today as part of a merge sweep. Holding — head unchanged since my review; the only new commit is a `Merge origin/dev`.

Both blocking items confirmed still present:

1. **`env_drift` on the loose gate.** `"env_drift": env_drift_report(env_file)` is still wired into `GET /api/credentials/status`, whose backend proxy is `get_authorized_agent_by_name` with no `reject_agent_principal`. That endpoint's own in-repo comment — *"`/credentials/status` gets away with the read gate because it returns a COUNT and names zero variables"* — is exactly what this falsifies. Worth noting #1890 landed today, so `require_admin`/`assert_admin` now reject agent principals fleet-wide; that does **not** cover this route, which is on the per-agent access gate, so the finding is unchanged.
2. **`/update` delivers quote-bearing values corrupted** — still round-trips through a writer that escapes and a parser that strips without unescaping.

Core mechanism is still right and I want it in; it's the drift-report gate and the delivery corruption that block.

dolho:
@vybe — both blocking items addressed.

**1. `env_drift` moved off the loose gate.** You're right that this falsified the sentence the repo has written down (*"`/credentials/status` gets away with the read gate because it returns a COUNT and names zero variables"*), and that it's reachable by a merely-shared user and by any agent-scoped MCP key resolving to the owner's role — with `get_credential_status` doing a `JSON.stringify` pass-through, so the whole fleet on a default admin-owned install.

- new `GET /api/agents/{name}/credentials/env-drift`, behind `get_owned_agent_by_name` **+ `reject_agent_principal`** — the same gate as `/credential-requirements`, `/inject`, `/export`, `/import`;
- `/credentials/status` strips the field, so the existing rationale comment is true again, and it now records where the drift went and why. A diagnostic added by a security fix shouldn't widen the surface the fix is about — agreed.

**2. `/update` value corruption — pinned as a strict xfail rather than left as a paragraph.**

Fixing the encoding *here* would move the ent#127 predicate under a security fix — the exact thing this PR's own reasoning argues against — and collide head-on with #2030, which does it properly and is stacked on this branch. So it's now a `strict=True` xfail naming the exposure, and I verified the follow-up actually closes it by composing the two locally:

```
this branch alone:      41 passed, 1 xfailed        <- corruption pinned, visible
merged with #2030:      [XPASS(strict)] ...         <- flips loud; drop the marker there
```

That way the loss can't be forgotten between the two merges, and whoever lands #2030 is told to retire it.

**Non-blocking items, done:**

- **PROTECTED_KEYS** gained the Node and git equivalents — `NODE_OPTIONS`/`NODE_PATH`/`NODE_REPL_EXTERNAL_MODULE` (every runtime here is a Node process, so `--require` is `LD_PRELOAD` under another name) and `GIT_SSH_COMMAND`/`GIT_SSH`/`GIT_EXTERNAL_DIFF`/`GIT_PAGER`/`GIT_EDITOR`/`GIT_CONFIG_*`, which git executes on the agent's next fetch or push.
- **The static guard walks both trees** now (`services/` **and** `routers/`) — `routers/brain_orb.py` already spawns a subprocess, so this was the #1965 lesson one directory over.
- **The router wiring is tested.** Deleting both `set_runtime_override(...)` calls the AC table cites left the suite green; an AST test now pins both keys and the force-unset (`value=None`) form.

**Not done, deliberately** — each is a behaviour change I'd rather not fold into a security fix: restoring `test_decode_divergence_is_deliberate` (the U+FFFD substitution is a real change and belongs with the parser work in #2030), the atomic `.env` write (tmp + `os.replace`), and the stale PR-body bullets — I'll refresh the body separately.

Merged latest `dev`; the sequencing you asked for (#2010 → #2030 → #2024) is unchanged and the xfail now enforces the first hop.

vybe (approved):
Re-reviewed. Both blocking items addressed the way I'd hoped.

**1. `env_drift` moved to its own owner-gated route** — `GET /credentials/env-drift` behind `get_owned_agent_by_name` + `reject_agent_principal`, and `/credentials/status` strips the field so the repo's own written rationale ("gets away with the read gate because it returns a COUNT and names zero variables") is true again. A diagnostic added by a security fix widening the surface that fix is about was the whole objection; this closes it.

**2. `/update` corruption pinned as `xfail(strict=True)`** rather than fixed here — correct call. Folding the ent#127 predicate under a security fix is the thing this PR's own reasoning argues against, and it would collide with #2030. The marker makes the loss impossible to forget between merges.

`PROTECTED_KEYS` gaining the Node and git equivalents is the right generalization — `NODE_OPTIONS --require` is `LD_PRELOAD` under another name, and `GIT_SSH_COMMAND` executes on the next fetch. The static guard walking `routers/` as well as `services/` is the #1965 lesson applied one directory over.

Merging first in the chain. #2030 lands next and retires the xfail; #2014 after that, per its own merge-order note.

Issue #2008, opened 2026-08-04
`POST /api/credentials/update` has no callers but is named as the sanctioned MCP-config path


AndriiPasternak31:
## Summary

`POST /api/credentials/update` (`docker/base-image/agent_server/routers/credentials.py:74`) has **no callers anywhere** — not the backend, not the MCP server, not `startup.sh`, not the frontend, not the enterprise submodule. Its own docstring says otherwise:

> "This endpoint is called by the Trinity backend when credentials are updated."

Two other places treat it as live:

- `docker/base-image/agent_server/routers/files.py:181-186` names it as the **sanctioned** alternative to raw `.mcp.json` editing: *"Owners modify MCP servers at agent-creation time via the template, or via the platform-internal `/api/credentials/update` flow which regenerates .mcp.json from the template with envsubst (no arbitrary content). See #590 (AISEC-C2)."* Both direct-edit paths are blocked on the strength of that sentence.
- `docs/memory/architecture.md:331` lists it as a live agent endpoint.

So the documented escape hatch for configuring MCP servers does not exist, and it is the only implementation of the `.mcp.json.template` contract the agent guide promises (#2007).

## Verification

The backend's agent-facing credential calls are `/api/credentials/status` (`routers/credentials.py:72`), `/inject` (`:345`), `/read` + `/list` (`credential_encryption.py:308,329,346`), `/reload-token` (`subscription_auto_switch.py:361`). `startup.sh:541` calls only `/api/internal/decrypt-and-inject`. Every agent URL in the backend is a literal f-string with a fixed path; no dynamic construction reaches it. Searched `src/`, `docker/`, `scripts/`, `config/`, the `.claude` submodule and the populated `src/backend/enterprise` submodule.

## Why it matters if anyone re-wires it

The renderer is a whole-text `str.replace` over `.mcp.json.template`, so unlike the backend engine it **does** substitute into `command`, and nothing on that path runs `validate_mcp_config`. Simulated verbatim from its source:

```
plain ${VAR} in command -> {"command": "/bin/sh -c whatever", "args": []}
```

That is the RCE-by-config class #590 closed on the inject path — a credential value becoming the executed command. It is inert only because nothing calls the endpoint, and the security comment above actively invites someone to call it.

Note the endpoint is not unauthenticated: `#1159` puts the agent server behind `X-Trinity-Agent-Token`, so only the backend (or the agent itself, which holds its own token) can reach it. That bounds the exposure; it does not make the substitution safe.

[...]

Issue #2007, opened 2026-08-04
`github:` templates never render `.mcp.json.template` — declared MCP servers are silently absent


AndriiPasternak31:
## Summary

`docs/TRINITY_COMPATIBLE_AGENT_GUIDE.md:90-92` states the contract:

> ### 3. `.mcp.json.template` (**Required** if using MCP servers)
> MCP server configuration with credential placeholders. **Trinity replaces `${VAR}` with actual values from the credential store.**

For a `github:` template, Trinity does not. Nothing renders the file, so a template that follows the published guide ships with **no MCP servers configured at all** — silently, with no warning and no error.

## Why nothing renders it

Grepping every reference to `.mcp.json.template` across `src/` and `docker/`, there is exactly one renderer: `docker/base-image/agent_server/routers/credentials.py:119`, behind `POST /api/credentials/update` — which has **zero callers** repo-wide (#2008).

The other paths don't cover it:

- `template_service.generate_credential_files` reads `<template_dir>/.mcp.json`, **not** `.mcp.json.template` (`template_service.py:1814`) — and is `local:`-only anyway (`crud.py:1145` gates on `template_data`, which only the `local:` branch populates)
- `startup.sh:381` copies `/generated-creds/.mcp.json` only when it exists — empty on the `github:` path
- No platform system prompt instructs the agent to render it itself (`platform_prompt_service.py` has no `.mcp.json` reference)

So the only writer of `/home/developer/.mcp.json` on a `github:` agent is `inject_trinity_mcp_if_configured()` (`trinity_mcp.py:66-77`), which creates the file from scratch when absent and adds only the `trinity` entry.

## Observed on the seeded default agent

`Abilityai/cornelius` is auto-provisioned on every fresh install (trinity-enterprise#107, `cornelius_agent_service.py`). It ships `.mcp.json.template` and no `.mcp.json` (HTTP 404 on the latter, confirmed against `main`), declaring three servers:

```json
"mermaid-diagram": { "command": "npx", "args": ["-y", "@peng-shawn/mermaid-mcp-server"] },
"aistudio":        { "command": "npx", "args": ["-y", "aistudio-mcp-server"],
                     "env": { "GEMINI_API_KEY": "${GEMINI_API_KEY}" } },
"ebook-mcp":       { "command": "${UV_PATH:-uv}",
                     "args": ["--directory", "${EBOOK_MCP_PATH:-/opt/mcp/ebook-mcp}", "run", "ebook-mcp"] }
```

[...]

Issue #1929, opened 2026-07-31
generate_credential_files: `.mcp.json` placeholder handling contradicts the platform validator (`${VAR:-default}` blanked, `command` unsubstituted, output never validated)


AndriiPasternak31:
> **Rescoped 2026-08-04** after a verification pass (see [this comment](https://github.com/Abilityai/trinity/issues/1929#issuecomment-5185066545)). The original title and premise — "breaks the seeded Cornelius" — were **wrong**: Cornelius is a `github:` template and never reaches this code path. The acceptance criteria below replace the originals; the original "suggested fix" implemented a placeholder form the platform's own validator rejects. History preserved in the comment thread.

## Summary

`generate_credential_files()` substitutes `${VAR}` placeholders in a template's `.mcp.json` by slicing the first two and last characters:

https://github.com/Abilityai/trinity/blob/dev/src/backend/services/template_service.py#L1849-L1868

```python
if isinstance(arg, str) and arg.startswith("${") and arg.endswith("}"):
    var_name = arg[2:-1]                               # "${A:-b}" -> "A:-b"
    real_value = agent_credentials.get(var_name, "")   # miss -> ""
    new_args.append(real_value)
```

Two defects, both pre-existing on `dev` since the initial commit:

1. **`${VAR:-default}` is mis-substituted to `""`** — the derived key is the whole `VAR:-default` string, which never matches a credential name.
2. **`command` is never substituted at all** — the loop handles only `env` and `args`, so a `${...}` in `command` survives verbatim.

But the parse is a symptom. The real defect is below.

## Root defect: one file, three writers, no agreed contract

`/home/developer/.mcp.json` has three writers with mutually inconsistent placeholder contracts, and the create path is the one nobody validates.

| | `generate_credential_files` (create, `local:` only) | agent `/api/credentials/update` renderer | `validate_mcp_config` |
|---|---|---|---|
| `${VAR}` in `env` | substituted | substituted | **accepted** |
| `${VAR:-default}` in `env` | → `""` | verbatim | **rejected** (ref name must match `^[A-Z][A-Z0-9_]*$`) |
| `${VAR}` in `args` | substituted | substituted | **rejected** (bare `$`; args are never ref-stripped) |
| `${VAR:-default}` in `args` | → `""` | verbatim | **rejected** |
| `${...}` in `command` | untouched | **substituted** | **rejected** (allowlist field) |
| validated? | **never** | never | — |

Consequences:

[...]

PR #2011, opened 2026-08-05
fix(deploy): validate an archive .mcp.json before it is persisted, not after (#2006)

### LINKED ISSUE #2006: deploy-local persists an archive whose `.mcp.json` failed validation, and the create path consumes it
## Summary

`deploy-local` validates an archive-supplied `.mcp.json` **after** it has already copied the archive into `/data/deployed-templates/<version_name>/` — a live member of `_LOCAL_TEMPLATE_ROOTS`. The 400 aborts the deploy but nothing removes the directory, so a config the validator refused stays on disk and is reachable by a subsequent `POST /api/agents {"template": "local:<version_name>"}`.

ent#213 closed the workspace ingress (`_prepopulate_workspace_from_template`). This is the template-directory ingress, which the same guard sits below.

## The ordering

| line | `services/agent_service/deploy.py` |
|---|---|
| `:534` | `shutil.copytree(extract_root, dest_path)` → `/data/deployed-templates/<version_name>/` |
| `:566-571` | `_validate_archive_mcp_config(...)` raises **400**, 32 lines later |
| `:645-649` | `finally` removes `temp_dir` only — `dest_path` is untouched (the only other `rmtree`, `:532`, is the pre-copy overwrite) |

`crud.py:150-153` — `_LOCAL_TEMPLATE_ROOTS` includes `Path("/data/deployed-templates")`.
`deploy.py:550` — the deploy path's own create call uses `template=f"local:{version_name}"`, i.e. exactly the id that resolves to the residue.
`validate_mcp_config` has three call sites (`routers/credentials.py:331`, `credential_encryption.py:71`, `deploy.py:74`) — **none on the create path**.

`version_name` is derived from the caller's own `base_name` via `get_next_version_name`, so the residue is addressable by the person who created it.

## Reproduction (executed against a real backend container, not inferred)

Baseline: `/data/deployed-templates` did not exist.

```
validator verdict      : REJECTED (Server 'ebook-mcp': args[1] contains shell metacharacters)
deploy result          : HTTP 400
deploy detail          : Invalid .mcp.json in archive: ...
RESIDUE dir exists     : True
RESIDUE contents       : ['.mcp.json', 'CLAUDE.md', 'template.yaml']
_LOCAL_TEMPLATE_ROOTS  : ['/agent-configs/templates', '/data/deployed-templates']
residue inside a root  : True   -> local:resid-probe-1929 resolves here
```

The rejected `.mcp.json` was on disk verbatim, including the `${LD_PRELOAD}` reserved env-ref the validator had just refused.

Consuming it (no container created — resolver + credential writer only):

```
local:resid-probe-1929 resolves to : /data/deployed-templates/resid-probe-1929
generate_credential_files OUTPUT   : args ['--directory', '', 'run', 'ebook-mcp']
                                     env  {'LD_PRELOAD_REF': ''}
validator on OUTPUT                : ACCEPTED
```

Two things worth noting from that last line:

1. **The create path launders the file.** #1929's `${VAR:-default}` → `""` blanking removed the very characters the validator rejected, so the *generated* config passes cleanly. Any fix of the shape "validate the generated output" would go green on this case.
2. **Configs rejected for reasons substitution does not touch pass through verbatim** — unknown top-level field, `trinity` redefined as stdio, oversize. And `_stage_config_files` mounts the whole template directory at `/template`, so `startup.sh` copies the raw file into the agent home regardless of what the credential writer does.

## Severity

Within the deploying creator's own trust boundary — same framing ent#213 itself used ("defense-in-depth / consistency, not critical RCE"). Filing P2, not as a security escalation. What it costs is the validation guarantee: a creator can get an arbitrary rejected `.mcp.json` onto an agent, and Claude Code auto-loads `~/.mcp.json` via `--mcp-config` on the next execution.

## Suggested fix

Validate **before** `copytree` (the guard only needs the extracted `extract_root / ".mcp.json"`, which exists at that point), or remove `dest_path` on any failure after the copy. The first is preferable — a guard that runs after the persist is not a gate.

## Acceptance criteria

- [ ] A deploy whose `.mcp.json` fails validation leaves **no** directory under `/data/deployed-templates/`
- [ ] `local:<version_name>` for a rejected deploy resolves to nothing (create returns the normal not-found path)
- [ ] A valid deploy is unaffected — same behaviour, same 200
- [ ] Regression test exercising the **ordering**, not the guard in isolation. ent#213's suite tests `_validate_archive_mcp_config` directly ("the full deploy flow needs Docker" — `test_213_deploy_mcp_validation.py:17-19`), which is why this ordering was never covered. The repro above mocks only the docker/db lookups preceding the window and runs the real flow, so it works without Docker.

## Related

- ent#213 — added the guard this issue repositions
- #1929 — the create path's own lack of a contract; this issue is that defect's only live route
- #590 / AISEC-C2 — the raw-`.mcp.json` ingress class
- #2007, #2008 — the other two `.mcp.json` ownership gaps found in the same pass

dolho:
## Summary

`deploy-local` validated an archive-supplied `.mcp.json` **32 lines after** it had already copied the archive into `/data/deployed-templates/<version_name>/`. The `finally` removed only `temp_dir`, so a config the validator refused stayed on disk in a live member of `_LOCAL_TEMPLATE_ROOTS` and remained reachable through `POST /api/agents {"template": "local:<version_name>"}` — by the same caller who had just been told the deploy failed.

ent#213 closed the workspace ingress. This is the template-directory ingress that sat below the same guard.

## Fix

**The guard moves to step 6a**, running on `extract_root / ".mcp.json"` before anything is persisted. `temp_dir` is the only thing written before that point and the existing `finally` already removes it, so a rejection now leaves nothing behind.

Placing it there also puts it ahead of two side effects that used to sit between the copy and the old guard — the quota lookup and the **stop-previous-version** step. A deploy that will be refused no longer stops a running agent on its way to the refusal. The post-copy call site becomes bookkeeping only (`credentials_imported[".mcp.json"]`), with a comment saying why re-validating a copy of already-vetted bytes would add nothing.

**`_remove_partial_deploy` is a second layer**, because moving the gate fixes the reported route but not the class: any failure between `copytree` and `create_agent_fn` — `_prepopulate_workspace_from_template`'s 500, a docker outage — leaves the same addressable `local:<name>` residue. The handle is cleared **just before creation starts**, so a directory that a container's mount spec may already reference is never removed on a late failure; creation-path rollback stays `crud.py`'s job. Both sides of that boundary are tested.

## Acceptance criteria

| AC | Where |
|---|---|
| A rejected deploy leaves no directory under `/data/deployed-templates/` | guard moved above `copytree` · `test_rejected_mcp_json_leaves_no_deployed_template_dir` |
| `local:<version_name>` resolves to nothing | `test_local_id_for_a_rejected_deploy_resolves_to_nothing` |
| A valid deploy is unaffected — same behaviour, same 200 | `test_valid_archive_still_reaches_creation_with_the_template_on_disk`, `test_archive_without_an_mcp_json_deploys` |
| Regression test exercising the **ordering**, not the guard in isolation | drives the real `deploy_local_agent_logic`; ordering pinned by `test_rejection_happens_before_any_external_side_effect` + `test_the_guard_is_called_before_the_copy_in_source_order` |

## On the ordering test specifically

Worth being explicit, because it changes what the suite actually proves: **once `_remove_partial_deploy` exists, the residue assertion alone no longer distinguishes the two candidate fixes.** With the guard put back below `copytree`, the cleanup removes the directory anyway and AC #1's test still passes.

So the ordering is pinned separately and twice — behaviourally (every post-guard step is made to explode; a clean 400 proves the guard ran first) and statically (source order inside the function). Mutation check: restoring the pre-fix ordering fails exactly those two, and deleting the cleanup fails exactly the partial-deploy test.

## Test plan

- [x] `tests/unit/test_2006_deploy_residue.py` — 8 tests, real flow, no Docker (that is a property of the fix, not the harness — the rejection path reaches nothing external)
- [x] Mutation-verified: pre-fix ordering → 2 failures (both ordering tests); cleanup removed → 1 failure (partial-deploy)
- [x] Uses the issue's own `${LD_PRELOAD}` reproduction payload, and notes in-file why "validate the generated output" would go green on it — #1929's `${VAR:-default}` blanking launders exactly those characters
- [x] 753 adjacent tests green (`-k "deploy or 213 or 950 or 1929 or mcp or template or local"`), including ent#213's suite unchanged

## Not included

#2007 and #2008 — the other two `.mcp.json` ownership gaps from the same pass — and #1929's missing create-path contract, which is the reason a laundered config passes on the way in. Separate issues, separate scope.

Closes #2006


vybe (approved):
Verified the premise independently: `/data/deployed-templates` is `_LOCAL_TEMPLATE_ROOTS[1]` and is resolved by `_resolve_local_template_dir`, so a leftover directory really is addressable as `local:<version_name>` — the residue is reachable, not theoretical.

Validation is genuinely before persistence on **all** paths: the guard runs on `extract_root`, the only prior write is the `tempfile.mkdtemp` staging dir (removed by the existing `finally`), and nothing between the guard and `copytree` writes `.mcp.json` — `body.credentials` merge into `.env` only, and the warning collector is read-only. So the bytes validated are the bytes persisted. It also fails closed: an unreadable or directory-shaped `.mcp.json` escapes into the generic handler → 500 with `dest_created is None`, so nothing lands.

Nice side effect worth noting: the guard now also precedes the quota lookup and the stop-previous-version step, so a doomed deploy no longer stops a running agent.

Two non-blocking notes for a follow-up rather than this PR:

**The cleanup is in two `except` arms, not a `finally`** — so an `asyncio.CancelledError` (client disconnect, shutdown) between `copytree` and the handover still leaves addressable residue. That residue contains a *validated* config, so it's orphan hygiene rather than the security hole this fixes, but it does soften the "fixes the class, not just the route" framing. Moving both arms to `finally: _remove_partial_deploy(dest_created)` is strictly equivalent on every other path, since `dest_created` is `None` on success and at the handover.

**The validator's log argument moved from `version_name` to `base_name`** — necessary, since `version_name` isn't computed yet, but a redeploy rejection no longer names which version was refused. Worth a one-line comment.

Rebased onto dev; only `tests/registry.json` conflicted, resolved from the merge stages (134 + 1 = 135).

Issue #590, opened 2026-04-29
SEC: PUT /api/agents/{name}/files guardrail bypass → .mcp.json write → RCE and Anthropic token exfiltration (AISEC-C2)


vybe:
## Summary

The file-write endpoint `PUT /api/agents/{name}/files` enforces `path_deny` only by **basename match** on `.env` and `config`. The remaining 9 entries in the documented `path_deny` policy (`.mcp.json`, `.credentials.enc`, `.ssh/*`, `/opt/trinity/*`, etc.) are not evaluated. An authenticated agent owner can overwrite `.mcp.json` to inject arbitrary MCP tool definitions, then restart the agent to achieve **root RCE inside the container and exfiltrate the per-tenant Anthropic OAuth token**.

## Context

AISEC (scan 3aad5469, 2026-04-28) demonstrated the complete chain on a live tenant:

```
PUT /api/agents/esc/files?path=.env        → 403 "Cannot edit protected path: .env"
PUT /api/agents/esc/files?path=.mcp.json   → 200 {"success":true}   ← BYPASS
POST /api/agents/esc/stop  →  200
POST /api/agents/esc/start →  200
POST /api/agents/esc/chat  {"message":"x"} →  200
GET /api/agents/esc/files/download?path=content/m.txt
Body: uid=0(root) gid=0(root) groups=0(root)
      cap_chown,cap_dac_override,...,cap_sys_chroot,cap_sys_ptrace,...
```

Reading `/proc/1/environ` then extracted:
```
CLAUDE_CODE_OAUTH_TOKEN=sk-ant-oat01-FM_...   ← per-tenant Anthropic billing token
TRINITY_MCP_API_KEY=trinity_mcp_nt_t62_...    ← platform MCP key (different prefix)
```

**Why the prior fix (#183) did not close this:**  
UnderDefense issue #183 "Arbitrary File Write via Credential Injection Parameter Tampering" was marked [Fixed] on 2026-04-23. That fix added basename-matching for `.env` and `config`. The `guardrails-runtime.json` `path_deny` list contains 11 entries; only the first 2 are checked. The sibling `credentials/inject` endpoint also accepts `{"files":{".mcp.json": "<attacker JSON>"}}` without the basename check.

CVSS 9.9 (AV:N/AC:L/PR:L/UI:N/S:C/C:H/I:H/A:H), CWE-22 + CWE-94.

## Acceptance Criteria

[...]

PR #2013, opened 2026-08-05
fix(agent): render .mcp.json.template at startup so declared MCP servers exist (#2007)

### LINKED ISSUE #2007: `github:` templates never render `.mcp.json.template` — declared MCP servers are silently absent

dolho:
## Summary

`docs/TRINITY_COMPATIBLE_AGENT_GUIDE.md` publishes a contract — *"Trinity replaces `${VAR}` with actual values from the credential store"* — that nothing implemented for a `github:` template. Declared MCP servers were silently absent: no warning, no error. A freshly-seeded Cornelius ships three servers and runs with none.

## Why the fix is in the container

The issue suggested this might be fixed by rendering at creation in the backend. It can't be: a `github:` agent's files only exist **after `startup.sh` clones the repo inside the container**. The backend never holds them — and its own renderer (`template_service.generate_credential_files`) reads `.mcp.json`, not the `.template`, and is `local:`-only regardless. So `docker/base-image/agent_server/mcp_template.py` runs in-container, invoked from `startup.sh` after **both** `.env` sources (the `/generated-creds` copy and the `decrypt-and-inject` fallback), so credentials are present on either path.

This also settles #2008's coupling question in the direction that issue anticipated: the renderer is not `/api/credentials/update`, so that endpoint gets deleted rather than revived. Filed separately (sequenced behind #2010, which edits it).

## Decision: `env`-only substitution

Per the issue's option (a). `${VAR}` is expanded **inside `env` blocks only** — the only form `mcp_validator` accepts. A `${VAR}` in `args` is rejected outright (bare `$` is a shell metacharacter, args are never ref-stripped); `command` must be a literal allowlist entry. Substituting into `command` is precisely the RCE-by-config class #590 closed, and it is what `/api/credentials/update`'s whole-text `str.replace` does today (#2008). Option (b) — widening the validator — stays unshipped; it is a security-contract change that deserves its own review.

## Three properties, because this runs unattended on every boot

**Refuse, never blank.** A placeholder that cannot be resolved (no such credential, or an empty value) **withholds that server** with a named reason on stdout — it is not configured with `""`. Blanking is what makes a broken config look healthy; it is #1929's defect and it is what laundered the rejected config in #2006.

**Merge, never clobber.** Servers are only added when absent. The `trinity` entry, and anything an owner edited, survive every restart — which also makes this order-independent with respect to `inject_trinity_mcp_if_configured()`, since that may write `.mcp.json` before or after this point. Idempotent: second run is a no-op.

**Validate individually**, through a **vendored** `mcp_validator` (byte-identical to `src/backend/services/mcp_validator.py` — pure stdlib, so the `credential_paths.py` / `safe_yaml.py` precedent applies, with a parity test). One bad server never costs the agent the good ones.

On the Cornelius trio, that yields: `mermaid-diagram` and `aistudio` installed, `ebook-mcp` withheld with the validator's own reason — `command 'uv' not in allowlist` — which is the actionable message (its own `MCP-SETUP.md` documents `uvx`).

## Acceptance criteria

| AC | Status |
|---|---|
| `github:` template's `.mcp.json.template` rendered with credential-store values | done — `startup.sh` → `agent_server.mcp_template` |
| Rendered output passes `validate_mcp_config` | done — each server validated before admission; a test re-validates the written file with the **backend** validator |
| Rejected placeholder forms refused with a named error, never blanked | done — `TestRefuseNeverBlank` |
| Guide matches the code in both directions | done — `env`-only, `${VAR:-default}`, and what withholding looks like |
| Seeded Cornelius shows its servers, or a documented reason per server | done — two install, one withheld with the reason on stdout |
| Upstream: `Abilityai/cornelius` uses `uvx` not `uv` | **not done — different repo.** Unblocked by the reason now being surfaced; needs an issue filed there |

## Test plan

- [x] `tests/unit/test_2007_mcp_template_render.py` — 24 tests
- [x] Mutation-verified: blanking instead of refusing → 2 failures; clobbering instead of merging → 2 failures; dropping the placeholder-in-`command` guard → 1 failure
- [x] That third one is worth flagging: my first version of the test **passed with the guard deleted**, because the validator refuses an unsubstituted `${SHELL_PATH}` anyway (not in the allowlist) — a withheld-only assertion couldn't tell the two apart and would point the operator at the wrong fix. It now asserts the renderer's own named reason, plus an independent property test that the credential value never appears in the output. To be honest about what protects what: the boundary is "never substitute outside `env`" plus the validator; the guard exists to name the cause.
- [x] 1210 adjacent tests green (`-k "mcp or 598 or 590 or 1929 or template or credential or compat"`)

## Deployment note

Requires a base-image rebuild (`./scripts/deploy/build-base-image.sh`). Existing agents pick it up on recreate, not restart — an older image simply keeps today's behaviour, so a mixed fleet degrades to the status quo rather than breaking.

Closes #2007


vybe (changes requested):
The diagnosis is correct and unusually well-evidenced, and rendering in-container at startup is the right placement. But the fix does not achieve its own headline outcome, and CI is green only because the fixture uses a credential value that no real credential looks like.

**1. Post-substitution validation rejects real credential values — the motivating case still fails.**

`agent_server/mcp_template.py` validates *after* substitution, but `mcp_validator._validate_env_value` was written for the **unrendered** config: it strips `${…}` refs and then rejects whatever literal remains against `_LITERAL_SECRET_PATTERNS` and `_SHELL_METACHARS_RE`. After substitution, the literal *is* the secret.

Running the PR's own module against Cornelius's `aistudio` entry with a real-shaped Gemini key (`AIzaSy…`, which matches `AIza[0-9A-Za-z_-]{35}` exactly — that is Google's actual key format):

```
withheld MCP server 'aistudio': Server '_probe': env['GEMINI_API_KEY']
  looks like a literal secret — store it in .env and reference as ${VAR}
```

Same rejection for `sk-ant-`, `sk-proj-`, `ghp_`, `github_pat_`, `xoxb-`, `AKIA` — Anthropic, OpenAI, GitHub, Slack, AWS — and separately for any secret containing a shell metacharacter. The advice in the message is precisely what the operator already did.

It passes CI because `test_env_placeholder_is_substituted_from_the_credential_store` uses `GEMINI_API_KEY="real-key-value"`. So AC #5 ("a freshly-seeded Cornelius shows its declared servers") is met in the fixture, not in production.

Fix direction: validate the entry with refs **intact** (the shape the validator was designed for), then substitute — or scope the literal-secret and metachar checks to the pre-substitution literal.

**2. `render()` does raise, and one bad server costs all of them.**

`_resolves_to_private_ip` catches only `socket.gaierror`, but `getaddrinfo` raises `UnicodeError` for an over-long hostname label. The `render_server` call is unguarded and the loop sits outside any `try`. Reproduced: an entry with an 80-character hostname label made `render()` raise and write **nothing**, losing every valid sibling server. That falsifies both the "Never raises" docstring and the "validate individually / one bad server never costs the good ones" property the PR is built on. `startup.sh`'s `|| echo` saves the boot, not the render. Wrap the per-server call.

**3. The operator-facing reason names `_probe`, not the server.**

The probe key is the literal `"_probe"` and the validator embeds it verbatim: `withheld MCP server 'aistudio': Server '_probe': …`. The named, actionable reason is this PR's core value claim, and this is where it surfaces.

**Non-blocking, worth addressing:**

- **Read-modify-write race with `inject_trinity_mcp_if_configured()`.** `startup.sh` launches the agent server in the background *before* the renderer; injection runs inside `run_server()` and is itself an unlocked read-modify-write of `.mcp.json`. In practice other work sits between them so the server almost certainly wins, but "order-independent" holds for serialized execution, not for the concurrency the script creates. A lost `trinity` entry costs the agent all Trinity MCP tools until the next restart. Moving the render above the launch removes the window entirely.
- **DNS now runs in the boot path** — vendoring the full validator pulls `socket.getaddrinfo` into startup for every declared http/sse server, no timeout, fail-closed. A transient resolver failure withholds the server. Undocumented.
- **`python3 -m agent_server.mcp_template` imports the whole agent server**, so any import error anywhere in that tree silently disables rendering — and no test exercises this path (the fixture loads by file path and hits the flat-import fallback; the startup test only greps for the string).
- **"Merge, never clobber" also means "you cannot delete."** A server an operator deliberately removed silently returns on the next restart, which contradicts the guide's new "a restart never reverts your changes".
- **Invariant #5 doc drift**: `docs/memory/architecture.md` still enumerates three vendored mirrors; `mcp_validator.py` is a fourth and isn't listed. `/validate-architecture` will flag it.

**Verified clean:** the vendored `mcp_validator.py` really is byte-identical with a parity test; the credential-path policy is not violated (`.mcp.json.template` stays on the deny list, `.mcp.json` on allow-with-validation); `.mcp.json` is in `_GITIGNORE_PATTERNS` so rendered secrets aren't pushed; the atomic tmp + `os.replace` at 0600 is correct and stricter than `trinity_mcp.py`; the base-image rebuild requirement is correctly stated; scope is tight.

Also needs a rebase — currently conflicting against dev.

vybe:
Re-validated today as part of a merge sweep. Holding — the head is unchanged since my review: `475f6b99` (2026-08-05) is still the only content commit, and `695e47d3` (2026-08-08) is a `Merge origin/dev`. The rebase I asked for is done; none of the three blocking items are.

Re-confirmed against the current head rather than assumed:

- **Post-substitution validation** — `mcp_template.py` still calls `validate_mcp_config(json.dumps({"mcpServers": {"_probe": rendered}}))` on the *rendered* entry, so a real `AIzaSy…` / `sk-ant-` / `ghp_` value still trips `_LITERAL_SECRET_PATTERNS`. AC #5 remains met in the fixture (`GEMINI_API_KEY="real-key-value"`), not in production.
- **`_probe` in the operator-facing reason** — same line, still the literal probe key.
- **Unguarded `render_server`** — the per-server call is still outside a `try`, so the `UnicodeError` from an over-long hostname label still loses every valid sibling.

The diagnosis and the in-container placement are right, and I'd like to land this — it just needs the three fixes. Re-request review when they're in.

dolho:
@vybe — all three blocking items fixed, and you were right that the headline outcome was unreachable in production while CI stayed green.

**1. Validation runs BEFORE substitution.** The entry is validated with refs intact — the shape `mcp_validator` was designed for — so the literal-secret and metachar rules see `${GEMINI_API_KEY}` rather than the key itself.

One wrinkle worth flagging: `${VAR:-default}` is a Trinity extension the vendored validator doesn't know (`references malformed variable name 'MISSING:-/opt/fallback'`), so the probe **normalises each ref to `${NAME}`** — keeping it a reference — rather than teaching the vendored copy new syntax (Invariant #5) or validating a rendered secret. Trinity's own placeholder checks now run first too, so an operator gets *"Trinity substitutes only inside 'env'"* instead of the validator's generic *"'${SHELL_PATH}' not in allowlist"* for the same entry.

Verified end-to-end against your case, with real vendor shapes:

```
AIzaSyA1234567890123456789012345678901234   -> added=['aistudio']  (key present in .mcp.json)
sk-ant-… / sk-proj-… / ghp_… / github_pat_… / xoxb-… / AKIA…   -> all added
pa$$w|rd;with&shell`meta                     -> added
${MISSING:-/opt/fallback}                    -> resolves to /opt/fallback
```

**2. The per-server call is guarded.** Reproduced exactly as you described — an 80-character hostname label makes `getaddrinfo` raise `UnicodeError`, which `_resolves_to_private_ip` doesn't catch, and the unguarded loop lost every valid sibling and wrote nothing. Now:

```
before: render() raises, .mcp.json not written, 'good' lost
after:  added=['good'], skipped={'bad': 'could not be validated (UnicodeError: label empty or too long)'}
```
Guarding the **call** rather than that one exception type also covers future validator errors — a test pins that with an injected `RuntimeError`.

**3. The reason names the server.** `withheld MCP server 'aistudio': Server '_probe': …` → the declared name is the probe key, and a test asserts `_probe` never appears in a reason.

```
validate post-substitution -> 21 failed
per-server guard removed   ->  2 failed
probe key back to "_probe" ->  1 failed
as shipped                 -> 36 passed
```

Also **rebased onto `dev`** — the PR is `MERGEABLE` again.

**Non-blocking items not addressed here, with reasons** — happy to take any of them if you'd rather: the read-modify-write race with `inject_trinity_mcp_if_configured()` (moving the render above the agent-server launch in `startup.sh` is the right fix and is a startup-ordering change I'd rather land separately), DNS in the boot path, `python3 -m agent_server.mcp_template` importing the whole tree, "merge never clobber" meaning an operator-deleted server returns, and the Invariant #5 doc drift listing three vendored mirrors when `mcp_validator.py` is a fourth. The last one I'll add to `architecture.md` in the same pass as whichever of these you want.

vybe (approved):
Re-reviewed. All three blocking items fixed, and the verification is the kind I asked for.

**1. Validation before substitution** — confirmed in `mcp_template.py`: the probe validates with refs intact and normalises `${VAR:-default}` to `${NAME}` rather than teaching the vendored copy new syntax, which is the right call under Invariant #5. The real-vendor-shape table (AIza/sk-ant/sk-proj/ghp_/github_pat/xoxb/AKIA all added) is exactly the production case my review said was unreachable.

**2. Per-server guard** — guarding the call rather than `UnicodeError` specifically is the better fix; the injected-`RuntimeError` test pins it.

**3. Probe key** — reason now names the declared server, with a test asserting `_probe` never appears.

The mutation table (post-substitution → 21 failed, guard removed → 2 failed, probe key reverted → 1 failed) is what makes this reviewable — it shows the tests actually catch each regression rather than merely passing.

Deferred non-blocking items accepted as scoped, including the `startup.sh` ordering race, which I agree is a separate change. One I'm taking myself rather than asking for a push: Invariant #5 in `architecture.md` still enumerates three vendored mirrors and `mcp_validator.py` is a fourth — batching that into a follow-up doc commit so this doesn't re-run CI for one line.

PR #2014, opened 2026-08-05
fix(agent-server): remove the unreachable /api/credentials/update endpoint (#2008)

### LINKED ISSUE #2008: `POST /api/credentials/update` has no callers but is named as the sanctioned MCP-config path

dolho:
## Summary

`POST /api/credentials/update` had **no callers**. I re-verified the issue's claim across `src/`, `docker/`, `scripts/`, `config/`, the `.claude` submodule and the populated `src/backend/enterprise` submodule: the only hits are the definition itself and the `files.py` comment pointing at it.

Meanwhile its docstring said *"called by the Trinity backend"*, `routers/files.py` named it as the **sanctioned** alternative to raw `.mcp.json` editing — with **both** direct-edit blocks resting on that one sentence — and `architecture.md` listed it as a live endpoint. The documented escape hatch for configuring MCP servers did not exist.

It was also the one path that would substitute a credential value into `command`: a whole-text `str.replace` over `.mcp.json.template`, with no `validate_mcp_config` anywhere on it. That is the RCE-by-config class #590 closed — inert only because nothing called it, while the security comment above actively invited someone to call it.

## Decision: delete

This is the branch the issue anticipated (*"if that one is fixed by rendering at creation in the backend, this endpoint should be deleted rather than revived"*), reached by a slightly different route: **#2007 / PR #2013** puts the `.mcp.json.template` renderer where the files actually are — in the container at startup, `env`-only, validated per server, refusing rather than blanking. Reviving this endpoint would mean a second renderer with weaker rules on the same file.

Deleted: the route, and its now-unused `CredentialUpdateRequest` model — a dangling model is how a deleted route quietly comes back.

## References corrected to name paths that exist

| File | Was | Now |
|---|---|---|
| `agent_server/routers/files.py` | "the platform-internal `/api/credentials/update` flow" | the template (#2007) + the validated `POST /api/agents/{name}/credentials/inject` (#598) |
| `docs/memory/architecture.md` | listed as a live agent endpoint | removed |
| `docs/testing/API_TEST_REQUIREMENTS.md` | `REQ-AS-CRED-001` | removed |
| `docs/diagrams/03-agent-container.md` | `POST /api/credentials/update` | `/api/credentials/inject` |
| `docs/diagrams/09-agent-lifecycle-states.md` | claimed `inject_assigned_credentials` posts there | `/api/credentials/inject` — which is what `credential_encryption.py:375` actually calls |
| `docs/planning/TARGET_ARCHITECTURE.md:297` | listed it under "keep" | repointed |

**Left alone deliberately**: `docs/archive/**`, `docs/security-reports/**`, and the 2026-06 meta-analysis — point-in-time records that were true when written. `TARGET_ARCHITECTURE.md:315` also mentions it, in a historical sentence explaining why #1089 added a *different* endpoint in June; rewriting that would be rewriting the record rather than fixing a pointer.

## Acceptance criteria

| AC | Status |
|---|---|
| Endpoint removed, or reachable and validated | removed |
| `files.py:181-186` describes a path that exists | done |
| `architecture.md:331` matches reality | done |
| If retained: validated, `${...}` in `command` refused | n/a — not retained |
| A test that fails if `${VAR}` can reach `command` through this path | expressed as: the path is gone (AST-checked, not substring), its model is gone, the whole-text replace renderer is gone, and no live doc sends anyone back to it |

## Test plan

- [x] `tests/unit/test_2008_dead_credentials_update.py` — 7 tests; **all 7 fail on `dev`** and pass here (verified by stashing the change)
- [x] Route absence checked by **AST**, not substring — the corrected `files.py` comment and the test file itself both still contain the string, so a grep-shaped assertion would be self-defeating
- [x] 661 adjacent tests green (`-k "credential or agent_server or reload_token or 1089 or files"`)

## Merge order

⚠️ **Merge after #2010** (issue #1999). That PR modifies `update_credentials`, which this PR deletes. If #2010 lands first, this branch needs a rebase whose conflict resolution is mechanically "keep the deletion" — the function is gone either way. Merging this first would silently drop #2010's `sync_process_env` wiring on that endpoint, so please keep the order.

Closes #2008


vybe (changes requested):
The unreachability claim **is proven** and the deletion is right — I verified across all three surfaces per Invariants #5 and #13:

- Repo-wide, the only hits for `credentials/update` / `CredentialUpdateRequest` are the definition itself, the `files.py` comment, and one test docstring. Zero callers.
- Backend registers `status`/`inject`/`export`/`import`/`encryption-key` — no `update` counterpart, so no mirror is broken. `credential_encryption.py` and `github_pat_propagation_service.py` both POST `/api/credentials/inject`.
- MCP server: `client.ts` calls `/credentials/inject` only; no tool maps to `update`.
- Zero hits in either submodule, the frontend, scripts, config, or `startup.sh`. No dead imports left behind.

The security argument is also correct and worth stating plainly: this was the only path that substituted a credential value into an MCP `command:` field via whole-text `str.replace` with no `validate_mcp_config` anywhere on it — the #590 RCE-by-config class, live in every deployed base image. Removing the model alongside the route is right.

Two of the PR's own acceptance criteria are not met, though.

**1. `agent_server/routers/files.py` replaces one forward-reference with another.**

The corrected comment now tells owners to declare servers in `.mcp.json.template`, "rendered into `.mcp.json` at container startup … (#2007)". That renderer **does not exist on dev** — the deleted `update_credentials` was the only code that ever rendered the template, `startup.sh` has no `envsubst`, and `credential_paths.py` explicitly denies the template on the inject path. PR #2013 (#2007) is still open, and I have just requested changes on it, so it is not landing imminently either.

AC #2 is "`files.py` describes a path that exists"; the bullet is written in the present tense and is currently false. Either land #2013 first, or mark the bullet as planned and lead with `/api/agents/{name}/credentials/inject`, which is real today.

**2. The AC-#2 regression test is inert.**

`tests/unit/test_2008_dead_credentials_update.py` — I mutation-tested it twice:

- Deleting the **entire** `EDIT_PROTECTED_PATHS` rationale comment → 7 passed.
- Replacing it with a reworded pointer back to the deleted endpoint → 7 passed.

Cause: the `or ".mcp.json.template" in block` branch. `PROTECTED_PATHS` contains that literal and sits well inside the character window, so the `or` is satisfied regardless of the comment; the other guard is an exact-whitespace match on the *old* wording only. Drop the `or` branch and assert `/api/credentials/update` is **absent** from the window instead.

**Notes:**

- `tests/unit/test_reload_token_endpoint.py` still describes "the destructive whole-file flows owned by `/api/credentials/update`" — missed by the sweep, and not in the parametrized doc list (which covers only two of the four docs this PR edited).
- `CLAUDE.md` still documents `.mcp.json  # Generated at runtime`. Pre-existing, but this PR is specifically a doc-accuracy fix, so it belongs in the sweep or in #2013.
- **Base-image rebuild**: deployed agents keep the route until rebuild + recreate. Exposure is low post-#1159 (the derived `X-Trinity-Agent-Token` means only the backend, which never calls it, and the agent's own process, which already has code execution, can reach it) — but the PR makes no release-note claim either way.

**Merge-order caveat is live and correct**, and now firmer than when you wrote it: #2010 modifies `update_credentials`, and I have requested changes on it, so please keep the sequencing note until that settles. Both branches also need a rebase.

Scope is clean — 2 source files, 5 doc corrections, 1 test file, all traceable to the issue.

vybe:
Re-validated today as part of a merge sweep. Holding — head unchanged since my review (`475f6b99`-era content; the only new commit is a `Merge origin/dev`).

The blocking item I mutation-tested is still present verbatim:

```python
assert "/api/credentials/inject" in block or ".mcp.json.template" in block, (
```

`PROTECTED_PATHS` contains `.mcp.json.template` and sits inside the character window, so the `or` is satisfied regardless of the comment — deleting the entire rationale still passes. Assert `/api/credentials/update` is **absent** from the window instead.

The unreachability proof and the deletion itself remain correct, and the security argument (this was the only path substituting a credential into an MCP `command:` via whole-text `str.replace` with no validation — the #590 class) is why I want this in. The merge-order caveat is also still live: #2013 hasn't landed, so `files.py`'s present-tense forward reference is still false.

dolho:
@vybe — both unmet ACs fixed.

**1. `files.py` no longer forward-references a renderer that doesn't exist.** You're right that it swapped one for another: `.mcp.json.template` "rendered into `.mcp.json` at container startup" is not true on `dev` — deleting `update_credentials` removed the only code that ever rendered it, `startup.sh` does no substitution, and `credential_paths.py` denies the template on the inject path. The comment now **leads with `POST /api/agents/{name}/credentials/inject`** (real today, `validate_mcp_config`-gated) and marks the template rendering as **PLANNED (#2007 / #2013)**, with the reason it isn't reachable yet, so it reads as a roadmap note rather than an instruction.

**2. The AC-#2 guard is no longer inert.** Three defects, each proven by mutation rather than argued:

- dropped the `or ".mcp.json.template" in block` branch (that literal is in `PROTECTED_PATHS`, inside the window — the `or` was true regardless of the comment) and added an assertion that the deleted route is not offered as a destination, allowing only the historical "was removed" mention;
- **anchored on the rationale block instead of the 1500-byte lookbehind.** The window was the deeper bug: it reached backwards into `PROTECTED_PATHS` *and* silently dropped the top of the comment once it grew — which my longer comment immediately did;
- **asserted `credentials/inject`, not `/api/credentials/inject`.** The backend route is `/api/agents/{name}/credentials/inject`, so the literal in the original draft matched nothing anywhere in the repo. Invisible for as long as the `or` carried the test.

```
rationale block deleted            -> FAILED
reworded to /api/credentials/update -> FAILED
as shipped                          -> 7 passed
```

**3. Both sweep misses**: `test_reload_token_endpoint.py`'s docstring no longer lists the deleted endpoint as a live owner of the whole-file flows, and `CLAUDE.md` no longer claims `.mcp.json  # Generated at runtime` (nothing generates it — injection writes it; template rendering marked as #2007).

Base-image note: agents keep the deleted route until rebuild + recreate. Exposure stays low post-#1159 (the derived `X-Trinity-Agent-Token` means only the backend — which never calls it — and the agent's own process, which already has code execution, can reach it), but that now reads on the PR rather than being left implicit.

Merged latest `dev` (`tests/registry.json`, additive — both sides kept, 155 entries). Merge-order caveat vs #2010 stands unchanged. 13 passed.

vybe (approved):
Re-reviewed. Both unmet ACs fixed, and merged in the order your PR body asked for — #2010 landed first (f744f349), so the conflict resolution here was the mechanical "keep the deletion" you predicted.

**1. `files.py` forward reference** — now leads with `POST /api/agents/{name}/credentials/inject` (real today, `validate_mcp_config`-gated) and marks template rendering as PLANNED (#2007/#2013). Reads as a roadmap note rather than an instruction, which was the objection. #2013 has since landed, so that note is now true rather than aspirational.

**2. The AC-#2 guard is no longer inert** — all three defects were real and the mutation table proves each: the `or ".mcp.json.template" in block` branch was satisfied by `PROTECTED_PATHS` regardless of the comment; the 1500-byte lookbehind reached backwards into `PROTECTED_PATHS` *and* silently truncated as the comment grew; and `/api/credentials/inject` matched nothing anywhere in the repo. The third one is the instructive one — it was invisible precisely because the `or` carried the test.

**Merge verification I did on the conflict**, since "keep the deletion" could silently drop #2010's work: #2010's `sync_process_env` import and its live call site both survive outside the deleted function (`credentials.py:27` and `:306`); only the copy inside `update_credentials` went away, which is the intent. `pytest test_2008 + test_1999 + test_ent127` → 126 passed, 1 xfailed (the #2030 marker, correctly still pending).

Base-image caveat noted and correctly scoped: agents keep the deleted route until rebuild + recreate, and post-#1159 only the backend (which never calls it) and the agent's own process can reach it.

PR #2022, opened 2026-08-05
fix(canary): stop three false-positive classes drowning the harness (ent#335/336/337)

obasilakis:
## Summary

eu2 (`e6df5bf8`, PostgreSQL) is the first fleet instance to run the canary under prod compose after #1876 wired `CANARY_ENABLED` in. In ~13h it recorded **6,605 violations — 6,598 of them (99.9%) the harness misreading correct runtime behaviour.** In all three cases the platform is right and the invariant's predicate is wrong.

| Invariant | eu2 violations | Defect |
|---|---:|---|
| **E-06** (ent#335) | 6,220 (94%) | Fires forever on schedules of soft-deleted agents |
| **S-03** (ent#336) | 378 | TTL floor built from the agent cap, not the execution's own timeout |
| **R-01** (ent#337) | 6 critical pages | No persistence gate — pages on a transient zombie |

Counts above are additionally **doubled** by the missing cross-worker leader lease (trinity#1881 part 2, out of scope here).

## E-06 — schedules of soft-deleted agents (ent#335)

`_collect_enabled_schedules` filtered only on the schedule's own `deleted_at`, so schedules of a soft-deleted **agent** stayed "enabled" for the whole 180-day window (#834 Phase 1a preserves child rows by design) with a frozen `next_run_at`. The projection did not freeze because the scheduler failed to advance it — it froze because the scheduler *correctly stopped registering a deleted agent's jobs*.

**This is not a new predicate.** `db/schedules/crud.py::list_all_enabled_schedules` — the list the scheduler actually arms — is exactly the proposed join and applies **both** #834 filters. The collector's docstring already claimed to mirror it and had copied half.

Now inner-joins `agent_ownership` with `deleted_at IS NULL`: the matched pair to `_collect_known_agents`' deliberate *inclusion* of soft-deleted agents (which is what keeps L-03 from reporting those same preserved rows as orphans). The join also drops schedules whose owner row is gone entirely — L-03's orphans, and its `agent_schedules` scan is unfiltered, so that deferral is real (checked, not assumed).

## S-03 — floor from the agent cap (ent#336)

`acquire_slot` sets the TTL from **this execution's** timeout, which since #929 is legitimately below the agent cap (an explicit shorter schedule timeout, a 900s public turn, a loop's `timeout_per_run`). Building the floor from the cap paged **critical** for the entire runtime of every normally-configured scheduled run — and unlike E-06 it re-fires per execution, so it produced a fresh green→red page on essentially every one.

The runtime had already reached the opposite conclusion: `_cleanup_stale_slots_for_agent` reads `timeout_seconds` back off the metadata HASH for exactly this reason (#869). Cleanup was per-slot-aware and the canary was not — the two had been disagreeing about what a correct slot TTL is.

The floor now comes from that same stored field, read in one pipeline with the TTL. An unobservable timeout **skips** the slot rather than falling back to the cap — the fallback re-arms this same false positive in a narrower window (TTL 3000 vs an agent-cap floor of 3900 fires; vs the slot's real 2700 bound it does not). `-1`/`-2` stay untouched as the load-bearing #226 coverage.

**Honest about the cost.** `acquire_slot` derives the EXPIRE and the HSET from one local three lines apart and is the sole writer, so `below_floor` is now an *internal-coherence* check and no longer catches a caller passing the wrong timeout (the #913 class). The docstring's worked example claiming that coverage is **deleted** rather than left lying, and `architecture.md` says so too, so a green S-03 is not misread. Tracked as ent#339.

## R-01 — no persistence gate (ent#337)

Fired on any single positive sample, so it flagged the sampling window rather than the #407 leak: every page was `zombie_count: 1` on a container with 13h uptime that was never restarted — reaped by the ordinary parent `wait()` within one cycle. The harm model is cumulative; a zombie that clears in one cycle exhausts nothing.

Now gated on a dwell — and the dwell is **per-PID, not per-count**. A count cannot support one: three cycles catching three *different* transients are indistinguishable from one stuck zombie, and "clear when the count returns to 0" needs a zero to be *observed*, which under sustained load never happens — on exactly the busy agents that filed this. `ps -eo stat,pid,comm` gives the identity, and a pid present at T and T+600 has demonstrably dwelled.

Elapsed wall-clock from `snapshot.snapshot_time`, never `time.time()`: multi-worker-safe (a "seen twice" rule self-confirms in <1s across two workers — trinity#1881 part 2 is visible in the eu2 data as rows 0.8s apart) and structurally testable at the boundary (#1909). Marker is **first-write-wins**; rewriting `first_seen` each cycle would reset the clock and leave a permanently blind critical invariant. Keyed `agent:canary_zombie:{name}` and registered in `CLEARED_KEYSPACES` — the first *name-keyed* canary key, so it inherits the #1560 recycled-name hazard (E-02's `canary:e02:*` are global, hence legitimately unregistered). A `last_seen` gap over two intervals restarts the dwell, so a docker-exec outage is not counted as dwell time; unparseable `ps` output is `sources_unavailable`, never a silent green.

The issue's `etimes` alternative was evaluated and **rejected**: `etimes` is time since process *start*, not since it zombified, so a process that ran 45 min and zombified 200 ms ago reads as 45-min-old and fires immediately — the exact transient the gate exists to stop flagging.

## Deviations from the issues' acceptance criteria

Both are commented on their issues rather than silently omitted.

- **ent#336 AC #3** (fire when stored timeout > agent cap) is **dropped**. Loops are not clamped to the agent cap — `timeout_per_run` is bounded only by `MAX_TIMEOUT_PER_RUN` (7200) — so it would page critical on a legal loop, and S-03 cannot tell a loop slot from a schedule slot. Replaced by a build-time test pinning the two ceilings against each other: same regression, caught on the PR that introduces it rather than 5 minutes after deploy. Gap filed as ent#338.
- **ent#337 AC #2**'s growth arm is satisfied *by* the dwell rather than separately: a count that grows across cycles has been `> 0` continuously, so the dwell elapses. A separate immediate-fire growth arm would reopen the multi-worker race the issue itself warns about. The trend rides in `observed_state` for triage.

## Also resolved

ent#336's open question. `acquire_slot(timeout_seconds=900)` is an unreachable dead default (all four call sites pass explicitly; `capacity_manager.acquire`'s own default likewise). The real 900s source on the schedule-less `cornelius-oracle` is `routers/public.py` hardcoding it at both public-chat dispatch sites — deliberate (`d46281f6`, *"Increase public chat timeout from 2 minutes to 15 minutes"*), and correct, since the execution really is bounded at 900s. No change proposed there.

## Test Plan

- [x] `cd tests && pytest unit/` — **7514 passed, 18 skipped**, 0 failures
- [x] `tests/test_canary_invariants.py` — 125 passed (baseline on clean `dev`: 122; the 19 errors are a pre-existing `utils.safe_yaml` path artifact, verified by stashing the diff)
- [x] 41 new tests across three files, all asserting exact boundaries from one literal instant rather than margins

**New tests are in `tests/unit/` deliberately**: `backend-unit-test.yml` runs `cd tests && python -m pytest unit/`, so `tests/test_canary_invariants.py` — which holds every existing S-03/R-01/E-06 test — **is executed by no workflow**. A guard placed beside them would never go red. (`test_1880_canary_alert_parity.py` already records this finding in its own docstring.) The existing suite is updated in place *and* extended; `FakeRedis` gains `pipeline`/`hgetall`/`expire`, and the pipeline double **buffers rather than executing eagerly**, so a test cannot pass against code that issues the two reads separately.

Key regression guards, each pinning a property that would otherwise fail silently:
- E-06 still fires on a **live** agent's frozen projection (if this goes green, E-06 was silenced rather than corrected)
- S-03's `-1`/`-2` arms still fire with **no** stored timeout available
- R-01 does **not** fire on three consecutive cycles of three *different* transient pids
- R-01's `first_seen` is not rewritten each cycle (asserted at the mechanism — with one writer this is behaviourally invisible)
- The E-06 collector's three predicate clauses are pinned against `list_all_enabled_schedules`, so a fourth #834-style filter landing in the accessor fails here instead of silently re-opening ent#335

## Follow-ups filed

- ent#338 — loop `timeout_per_run` is never clamped to the agent cap
- ent#339 — S-03 `below_floor` is now internal-coherence only; the #913 class needs declared-timeout corroboration

Refs abilityai/trinity-enterprise#335, abilityai/trinity-enterprise#336, abilityai/trinity-enterprise#337


dolho (commented):
## Review — three canary false-positive classes

Diagnosis is right on all three, and the evidence (6,605 violations / 99.9% harness error) is the kind of grounding these changes need. E-06 in particular is clearly correct — inner-joining `agent_ownership` makes the collector match `list_all_enabled_schedules`, which was the predicate it already claimed to mirror, and the "matched pair with `_collect_known_agents`' deliberate *inclusion*" framing is the part that makes it reviewable.

Notes below are mostly on R-01.

### 1. R-01 — a container restart outside the backend re-arms the false positive (main concern)

The dwell marker is keyed by `(agent_name, pid)`, and PID identity is only meaningful within one container's PID namespace. `clear_agent_breakers` covers backend-mediated lifecycle events, but a container that restarts on its own — `docker restart`, a restart policy after an OOM kill or crash — gets a fresh PID namespace and **keeps its marker**. Fresh namespaces hand out low PIDs immediately, and a zombie `claude` in a freshly restarted agent is exactly the low-PID case, so collision is not remote. A new transient zombie landing on a PID still in the marker inherits a 10-minute-old `first_seen` and pages critical on its first sample — the precise false positive this PR removes.

The module docstring dismisses PID reuse on `pid_max` wrap-around grounds, which is the right argument for a *running* container and the wrong one for a restarted one. Cheap fixes, in order of preference:

- carry the container's `State.StartedAt` (or container id) into the snapshot and store it in the marker; mismatch → restart the dwell. One extra field in `_collect_zombie_counts`, no extra exec.
- or read field 22 (`starttime`) from `/proc/<pid>/stat` in the same awk pass and key on `pid:starttime`.

### 2. R-01 — `DWELL_SECONDS == _MAX_OBSERVATION_GAP_SECONDS == 600`

Both are 2× the 300s interval, so the gap tolerance is exactly two nominal cycles. Under `--workers 2` with no leader lease the *effective* cadence is fine, but a single slow or skipped cycle (`> 600s` between two observations of the same agent) restarts the dwell — and if collection is consistently slow, the dwell restarts every cycle and a genuine leak is **never** reported, silently. The two constants are doing different jobs and shouldn't share a value: suggest gap tolerance ≥ 3× interval, or derive both from the configured interval so they can't drift apart in a deployment with a non-default `CANARY_INTERVAL_SECONDS`.

### 3. R-01 — unpipelined marker I/O

`hgetall` + `hset` + `hdel` + `expire` is up to 4 round-trips per agent per cycle. The S-03 change in this same PR pipelines the TTL/HGET pair for exactly this reason ("the canary's hottest loop and Redis may be remote") — the same argument applies here, and the writes are already batched into one `updates` dict.

### 4. S-03 — `floor_seconds` on the `-1` / `-2` arms

Right call on the floor source; the honesty about `below_floor` now being an internal-coherence check (and deleting the docstring's worked example rather than leaving it lying) is the correct handling, and ent#339 is the right follow-up.

One residual: on the sentinel arms you still emit `floor_seconds` derived from the agent cap with `floor_source: "unknown"`. A reader — or trinity-enterprise#202's scorer, which consumes these rows programmatically — taking `floor_seconds` at face value gets a number that was never a judgement. Emitting `None` there makes it unusable-by-construction instead of misleading-if-unread.

### 5. E-06 — pin the test against the live statement

The regression guard pinning the collector's three predicate clauses against `list_all_enabled_schedules` is the right idea. Worth confirming it asserts against the accessor's *actual* statement rather than a literal list of clause names copied beside it — the latter drifts silently, which is how ent#335 happened in the first place.

### 6. Doc contract — the canary library no longer "writes nothing"

`architecture.md` still says the library "reads state … but writes nothing; the service persists violations". E-02 already bent that, and R-01 here plus H-01 in #1879 make it three checks with cross-cycle Redis state. Since you're already editing that block, worth amending the sentence rather than leaving a claim the code contradicts.

### 7. The unrun-test-file finding deserves its own issue

The note that `tests/test_canary_invariants.py` is executed by **no** workflow (`backend-unit-test.yml` runs `cd tests && pytest unit/`) is the most valuable incidental finding here. Putting the new tests in `tests/unit/` is the right move for this PR, but it leaves the ~125 existing canary tests unrun. Please file it — the fix is one line in the workflow, and until then any regression in the older S-03/R-01/E-06 tests goes undetected.

### Merge order

Conflicts with #1879 (both rewrite `_collect_zombie_counts`'s return dict and the `Snapshot` zombie fields) and with #1997 (`canary_alerts`). Worth sequencing the three deliberately.

dolho (approved):

obasilakis:
Rebased on `dev` (picks up #1997's instance-label change to `canary_alerts`). One conflict, in `docs/memory/learnings.md` — both sides appended a 2026-08-05 entry; kept all three.

Full unit suite green locally against the merge: `7763 passed, 18 skipped`. `tests/test_canary_invariants.py` is `125 passed, 1 skipped, 19 errors` — the 19 are pre-existing setup errors that reproduce identically on `dev` (`e6df5bf8`), see below.

Review point 7 — filed as #2037. Scope confirmed: **both** unit workflows (`backend-unit-test.yml` and `backend-unit-nightly.yml`) invoke `pytest unit/`, so nothing runs the ~126 canary tests at `tests/` root. The file has already silently decayed: all 19 errors are `ModuleNotFoundError: No module named 'utils.safe_yaml'` raised through `services/template_service.py:22` — the `reload_canary` fixture's `sys.modules` surgery shadows `utils`, and #1965/#1978's new import can't resolve through it. That landed unnoticed precisely because no workflow runs the file.

Points 1–6 not addressed here — the PR is approved and the three false positives are live on eu2 (6,605 violations), so this lands as-is. Residuals worth a follow-up, flagged rather than lost:

- **1 (main concern)** — the `(agent_name, pid)` dwell marker survives a container restart outside the backend (`docker restart`, restart-policy after OOM/crash), and a fresh PID namespace hands out low PIDs immediately. Real re-arm of the false positive. Fix is one extra field: carry `State.StartedAt` (or the container id) into the snapshot and invalidate the marker on mismatch.
- **2** — `DWELL_SECONDS == _MAX_OBSERVATION_GAP_SECONDS == 600`; a single slow cycle restarts the dwell, and consistently-slow collection means a genuine leak is never reported.
- **4** — S-03 still emits an agent-cap-derived `floor_seconds` alongside `floor_source: "unknown"` on the `-1`/`-2` arms; `None` would be unusable-by-construction instead of misleading-if-unread (trinity-enterprise#202's scorer reads these programmatically).
- **6** — `architecture.md`'s "the library … writes nothing" claim is now contradicted by E-02, R-01 here, and H-01 in #1879.

obasilakis:
Review point 1 (main concern) — fixed in `249744ac`.

You were right that the `pid_max` argument covers the wrong case. Took the first of your two suggestions, and it turns out to be genuinely free: docker-py's `containers.list()` defaults to `sparse=False`, which already issues a full inspect per container, so `State.StartedAt` is sitting in `container.attrs` before we ask — no extra API call, no extra exec. (Deliberately not `Created`: a restart doesn't create a new container, so it never moves.)

Stored in a reserved `__started_at` field on the marker HASH — non-numeric, and every other field is `str(pid)`, so collision is structural rather than lucky. On mismatch the whole marker is dropped and every pid restarts its dwell.

Two things worth calling out:

- **The `last_seen` continuity check does not cover this.** A restart inside one cycle leaves no observation gap at all, so `last_seen` stays fresh while the pid space underneath it is brand new. They're independent bounds; the docstring now says so.
- **Only an observed MISMATCH invalidates.** An agent absent from `zombie_container_started_at` (field unreadable, or a future `sparse=True`) leaves the dwell alone. Restarting on a non-signal every cycle would make R-01 permanently blind — the same failure mode as rewriting `first_seen` each cycle, which this PR already guards against.

8 new tests. The one that matters is your scenario end to end — dwell established on pid 811, container restarts, a fresh transient lands on 811, silent on first sample — paired with the differential that makes it a real guard rather than a tautology: the same fixture with the generation unobservable still fires at the dwell.

Also updated the R-01 row in `architecture.md` and extended the ent#337 `learnings.md` entry with the generalisable form (an identity is only unique within the namespace that issues it; the restart path that matters is the one outside your own lifecycle hooks).

Suite: `7771 passed, 18 skipped` (was 7763 before the 8). `tests/test_canary_invariants.py` unchanged at `125 passed, 1 skipped, 19 pre-existing errors`.

Points 2/4/6 still open — tracked in the previous comment.

PR #2028, opened 2026-08-05
fix(mcp): stop ask_trinity splitting surrogate pairs and losing sessions silently (#2027)

### LINKED ISSUE #2027: bug(mcp): ask_trinity truncation can split a surrogate pair; a session-less response loses caller context silently
## Summary

Two low-severity defects in `src/mcp-server/src/tools/ask_trinity_client.ts` (ent#328 / #1981), found by an `/edge-cases` pass. Both are narrow; filing together because they share a home and neither justifies its own PR.

## 1. `truncate` splits astral characters, emitting a lone surrogate

```ts
function truncate(text: string, max: number): string {
  if (text.length <= max) return text;
  return `${text.slice(0, max)}\n\n[output truncated at ${max} characters]`;
}
```

`String.length` and `slice` count **UTF-16 code units**, so a cut that lands inside a surrogate pair leaves half of one. Executed:

```
length of answer: 8005
char at boundary code unit: "\ud83d"   isLoneSurrogate: true
JSON.stringify of the cut: "\ud83d"
```

The tool's text then contains an unpaired surrogate, which is ill-formed Unicode. A strict UTF-8 encoder or JSON-RPC client replaces it with U+FFFD or rejects the message.

**Reachability is low**: an answer must exceed `MAX_OUTPUT_CHARS` (65,536) *and* have a non-BMP character straddling exactly that offset. Docs answers with emoji are plausible; the alignment is roughly one-in-65,536 per over-long answer.

**Fix**: step back one code unit when `text.charCodeAt(max - 1)` is a high surrogate (0xD800–0xDBFF), or cut on `[...text]`. Same treatment for the two `truncate(rawText, 200)` error paths, which are much likelier to carry arbitrary bytes than the answer path.

Note `MAX_QUESTION_CHARS` has the same code-unit basis, but there it only makes the limit slightly imprecise — not a defect.

## 2. A response with no `session_id` loses the caller's context silently

```ts
if (responseSessionId) {
  if (requestSessionId && requestSessionId !== responseSessionId) {
    // The silent-expiry case. Without this the caller silently loses its
    // conversation context and cannot tell why the answer stopped making
    // sense — the endpoint reports success either way.
    parts.push("⚠️ The previous session expired — ...");
  }
  parts.push(`session_id: ${responseSessionId} ...`);
}
```

The whole block is gated on `responseSessionId`. So when the caller **sent** a `session_id` and the service answers 200 with an answer but **omits** `session_id`, there is no warning and no session line — the caller silently loses its conversation context, which is precisely the failure the inner comment says the warning exists to prevent. The guard covers the "different id returned" branch and not the "no id returned" one.

The author clearly anticipated absence (`responseSessionId` is optional, and the no-answer branch handles it), so this reads as a missed branch rather than a design decision.

**Reachability**: depends on the endpoint ever answering successfully without a session id. Unknown — I could not verify live. If it never does, this is dead defensive code; if it can, it is the exact silent-context-loss the comment describes.

**Fix**: move the mismatch warning out of the `if`, and warn when `requestSessionId && !responseSessionId` as well.

## Acceptance criteria

- [ ] A truncated answer is never cut inside a surrogate pair (assert no unpaired surrogate in the output)
- [ ] The same holds for the two 200-char error truncations
- [ ] A caller that sent a `session_id` and receives none back is told its context was lost
- [ ] Regression tests for both

## Note on why no PR accompanies this

The MCP server's suite runs on `node:test`, which has no strict-xfail, and this package has no `fast-check`. Adding a deliberately-red test to `dev` would pollute every PR's `regression diff` baseline (the #1822 lesson), and adding a dependency for two low-severity findings is not a trade I'd make unasked. Both fixes are small enough to land with their tests in one change.

## Related

- #1981 / ent#328 — the feature
- `src/helper-mcp/src/client.ts` — the contract-identical sibling; **it carries the same `truncate`**, so both copies need the fix or the parity test starts failing on the mismatch

dolho:
## Summary

Two edge cases in `ask_trinity` (ent#328 / #1981), found by an `/edge-cases` pass. Both fixed in **both copies** — `src/helper-mcp/src/client.ts` carries byte-identical code.

## 1. `truncate` split surrogate pairs

```ts
return `${text.slice(0, max)}\n\n[output truncated at ${max} characters]`;
```

`length`/`slice` count UTF-16 code units, so a cut inside a surrogate pair leaves half of one. Proven by construction:

```
char at boundary: "\ud83d"   isLoneSurrogate: true
```

An unpaired surrogate is ill-formed Unicode — a strict UTF-8 encoder or JSON-RPC client replaces it with U+FFFD or rejects the message.

Reachability on the answer path is low (>65,536 characters *and* a non-BMP char at exactly that offset). **The error path is the real exposure**: the same helper truncates upstream bodies at 200 characters, and those carry arbitrary bytes from a Google frontend on a cold start.

Fixed by stepping back one unit off a high surrogate — only the final unit can be orphaned, so one check covers it — with the message still reporting the *configured* limit rather than the adjusted cut.

## 2. A session-less response lost the caller's context silently

The warning lived inside `if (responseSessionId)`, so a 200 that answered but omitted `session_id` produced **no warning at all**. The caller silently lost its conversation context — precisely what that warning exists to prevent, per its own comment. The "different id returned" branch was covered; the neighbouring "no id returned" one was not.

The warning is now decided by the **request's** session, with distinct wording for the two cases: *"a new session was started"* is false when none came back, and pointing a caller at a session that doesn't exist is worse than saying nothing.

## The parity test only held for its own table

Worth reading twice, because it changes what that suite proves. After fixing both copies I mutated **only the helper's** `truncate` — expecting the parity suite to fail. It passed, all 144.

`PARITY_CASES` had no over-long answer and no session-less response, so the "these two must agree" claim covered exactly the rows in its table and nothing else. The second copy was unverified.

Two rows added. Both helper-only mutations now fail:

```
helper-only mutation [surrogate] -> # pass 145 # fail 1
helper-only mutation [session]   -> # pass 145 # fail 1
```

## Test plan

- [x] `ask_trinity_edges.test.ts` — 11 tests: the straddling cut; the limit still reported as the ORIGINAL max; a fully-fitting pair surviving; **a BMP-only answer still cut at exactly max** (the step-back must not misfire); a short answer untouched; the 200-char error-path truncation; and all four session shapes
- [x] 2 rows added to `PARITY_CASES`, verified to catch a helper-only divergence
- [x] Mutation-verified on the main copy: reverting the step-back → 2 fail; warning back inside the `if` → 1 fail; same wording for both session cases → 2 fail
- [x] 146 pass (was 133), `tsc --noEmit` clean

## Note

`fast-check` still isn't a dependency here and I didn't add one — the boundary case is exactly enumerable, so the deterministic tests *are* the property. A generator would explore more of the space but wouldn't have found anything these don't pin.

Closes #2027


vybe (approved):
Worked through the surrogate reasoning independently and the single-check claim is exhaustive, which is the part worth confirming: JS strings are UTF-16 code-unit sequences and `slice(0, max)` includes units `[0, max)`, so the only unit that can be orphaned is index `max-1`. A **high** surrogate there has its partner at `max` and is excluded; a **low** surrogate at `max-1` cannot be orphaned, because its high partner at `max-2` is necessarily included. So `charCodeAt(end - 1)` covers every case — a code-point-aware re-slice would catch nothing more. `max=0` is safe (`charCodeAt(-1)` → `NaN`).

Two limits are honestly outside the bug and correctly left alone: truncation stays by UTF-16 code units (the message still says "characters"), and grapheme clusters can still split — but that yields *well-formed* Unicode, which is all the fix claims. Reporting the configured `max` while the actual cut is `max-1` is deliberate and has its own test.

The session fix preserves the documented contract exactly. Full matrix: only the request-had-an-id / response-had-none case changes, from silence to a no-id warning. `String(data.session_id)` is untouched, `MAX_SESSION_ID_CHARS` unchanged, no numeric parsing introduced — so the opaque-string rule holds, and silent expiry stays silent because the warning is inferred client-side from an id mismatch rather than from any service error field.

The critique of the parity suite is the most valuable part: `PARITY_CASES` had no over-long answer and no session-less response, so "these two copies must agree" covered exactly its own table and nothing else. The two added rows fix that, and the helper-only mutations demonstrate it — they pass against dev's table and fail against this one.

Verified after rebasing onto dev: `tsc --noEmit` clean, **183/183** tests pass, and both copies (`src/helper-mcp/src/client.ts` and `src/mcp-server/src/tools/ask_trinity_client.ts`) carry identical code in both changed regions.

One follow-up worth its own issue rather than a fix here: if the endpoint ever emits `session_id` as a JSON **number**, `JSON.parse` has already destroyed precision before `String()` runs — and the file's own header notes these values exceed `Number.MAX_SAFE_INTEGER`. The coercion protects the type, not the value.

Rebased; only `tests/registry.json` conflicted, resolved from the merge stages.

PR #2030, opened 2026-08-05
fix(credentials): make the .env quote encoding reversible (#2023)

### LINKED ISSUE #2023: bug(credentials): .env quote-escaping is write-only — a credential containing a double quote round-trips corrupted

dolho:
## Summary

The `.env` writer escaped an embedded double quote; the reader never unescaped it. Any credential containing `"` was injected as one string and read back as another:

```
value written : ghp_a"b
line on disk  : SECRET="ghp_a\"b"
agent reads   : 'ghp_a\\"b'     (expected 'ghp_a"b')
```

The agent then authenticates with a value the operator never supplied — and it presents as a *bad credential*, which is the worst possible disguise for a parsing bug.

Split out of #2017 because it isn't a PAT problem (PATs are `[A-Za-z0-9_]`, so a quote can't appear in one). It affects **every** credential written through this path.

## Why it waited for #2010

Four things have to move in one commit, and #2010 rewrites the reader:

| Piece | Change |
|---|---|
| **Writer** | escapes the escape character *first*. Escaping only `"` left the encoding **undecodable**, not merely untidy: a value ending in `\` produced `KEY="a\"`, whose closing quote reads as escaped |
| **Reader** | `unquote_env_value` — one matched pair stripped, the encoding reversed in a **single scan** (two sequential replaces mishandle an escaped backslash preceding an escaped quote) |
| **`_env_pairs`** | the ent#127 predicate is *defined* as agreement with the reader, and its **source is spliced into an in-container probe** — so it carries an inlined mirror, not an import, or it would `NameError` inside the probe |
| **Parity fixtures** | now run against the **real reader**, loaded by path, instead of a hand-written replica |

That last one is a genuine upgrade: a replica is only as honest as the last person to sync it, and the old guard could prove the original still existed but never that the copy still matched it.

I also rebased #2010 in the process — it had gone conflicting — so the blocker is clear.

## Mutation testing found two holes in my own tests

Worth stating, because both would have shipped as false confidence:

1. **The first draft reimplemented the writer inside the test file.** Deleting the real writer's backslash-escaping broke *nothing* — the tests were exercising the copy. Same "second copy is unverified" failure I flagged on #2027's parity table, this time in my own work. The encode half is now `execution_env.format_env_line`, sitting beside its inverse, and the tests call it.

2. **Quote-escaping survived every round-trip test.** Positional stripping decodes `KEY="a"b"` back to `a"b` regardless, so the round trip can't see the escaping's absence. It still matters for consumers that aren't this parser — a shell sourcing `.env` reads `ab`. Now pinned line-by-line, not just through the round trip.

All five mutations are caught now: writer-no-backslash-escape (2 fail), writer-no-quote-escape (2), reader-back-to-strip (22), unquote-no-unescaping (19), predicate-old-parse (13).

## The deliberate pin doing its job

#1999's quirk table was byte-faithful *on purpose*, with its own comment saying the parse improvement belonged in "a separate change with ent#127's parity test as its gate." This is that change: three rows fired and were retargeted with the reasoning recorded. The behaviour moved visibly and argued, instead of drifting underneath its consumers.

Two of those are worth a reviewer's eye — `KEY=""""` now reads as `""` rather than empty, and `KEY="` as `"` rather than empty. Both are hand-written shapes the writer never emits, and the contract is agreement with what the **agent** sees, not with shell concatenation semantics.

## Test plan

- [x] 46 new tests: 15 credential values (quote, backslash, trailing backslash, mixed, apostrophe, non-ASCII, base64, shell metachars, empty) round-tripping through writer→reader **and** through the predicate; multi-value files; a quote no longer terminating the line early; the encoding pinned line-by-line; 11 `unquote` shapes; and a check that the probe script still compiles with the inlined mirror
- [x] 872 green across `2023 / ent127 / 1999 / credential / agent_server / 1264 / 1574 / 1967`

## Merge order

Stacked on `fix/1999-env-ghost` (#2010) — **merge that first**. The diff shown here includes its commits until it lands.

Closes #2023


vybe (changes requested):
The encode/decode pair itself is correct, and extracting `format_env_line` so the test pins the *line* rather than just the round trip is the right call — the mutation finding behind that is real. But this changes a decoder that has **more copies than the PR migrated**, and one of the misses is a credential-redaction path.

**1. The credential sanitizer still uses the old decoder, so credentials containing `"` stop being redacted from logs.**

`agent_server/utils/credential_sanitizer.py:159` — `var_value.strip().strip('"').strip("'")`, untouched here. I verified it on `dev`: it is a fourth copy of the `.env` decoder, and it is what builds the redaction set.

```
on-disk:            MY_API_KEY="sk-live-a\"bcdefghij"
redaction set holds: sk-live-a\"bcdefghij     <- escaped form
execution receives:  sk-live-a"bcdefghij      <- unescaped form
=> sanitize_text() does NOT redact it
```

Before this PR the two agreed (both corrupted identically), so redaction worked. This is a regression, and it is in the direction CLAUDE.md's security rules forbid.

I also confirmed the reachability: `startup.sh` does **not** source `.env` (no `set -a` / `source` of it anywhere), so on any agent-server start the `os.environ` branch of `_load_credential_values()` contributes nothing and the stale-`.env` branch is the *only* source. The warm path immediately after `/credentials/update` is fine; every subsequent process lifetime leaks. This belongs in this commit, for exactly the reason the PR body gives for the other three movers.

**2. `_format_pat_line` is a second `.env` writer that still escapes only `"`.**

`src/backend/services/github_pat_propagation_service.py` — its docstring asserts it "match[es] the agent's own .env writer", which this PR makes false. Runs of ≥2 backslashes corrupt. Low real-world frequency, but it is precisely the asymmetric-seam class this PR argues against, and nothing ties the two writers together in a test. Route it through `format_env_line` or add the parity assertion.

**Backward compatibility needs stating in the body.** Old `.env` files are silently re-interpreted with no migration and no detection. The change retroactively **repairs** every quote-bearing value (good, and worth saying), but any pre-existing value with two or more consecutive backslashes (Windows paths like `C:\\server\share`, UNC paths) is corrupted until re-pushed. `env_drift_report` won't catch it — it compares file↔process, not encoding vintage. Narrow enough to accept as a documented consequence; not acceptable as an undocumented one.

**Worth fixing while you're here:** `format_env_line` neither escapes nor rejects `\n`/`\r`. A value containing a newline decodes such that a *new key* is created in `build_execution_env` — and `PROTECTED_KEYS` refuses `LD_PRELOAD` but not `ANTHROPIC_API_KEY`/`GITHUB_PAT`/`CLAUDE_CODE_OAUTH_TOKEN`. Not a regression (the old writer was equally broken) and the input is operator-supplied, but the PR title claims reversibility and the writer is now a single named function where `if "\\n" in value or "\\r" in value: reject` is one line. At minimum pin the known limit in a test.

**Coordination:** this is stacked on **#2010**, which I have also requested changes on (it moves the same reader). And **#2024**'s `agent_reads()` test helper is a hand-written replica of the *old* reader whose double-backslash case asserts an equality that becomes false in production while the test stays green — the exact false-confidence shape this PR's own commit message condemns. Worth settling #2010 → #2030 → #2024 as one sequence rather than merging them independently.

Tests are otherwise strong — 168 pass, the round-trip property covers backslash/trailing/double/mixed/empty/unicode/metachars, and replacing `_exporter_replica` with a by-path load of the real reader is a genuine upgrade. Gaps: no newline/CR coverage, and `PARITY_FIXTURES` has no backslash fixture at all — the hand-written corpus for the copy most at risk of drift omits the exact class this PR changes.

vybe:
Re-validated today as part of a merge sweep. Holding — head unchanged since my review; the only new commit is a `Merge origin/dev`.

The blocking item is confirmed still open: **`docker/base-image/agent_server/utils/credential_sanitizer.py` is not in this PR's file list.** Line 159 still reads

```python
var_value = var_value.strip().strip('"').strip("'")
```

so it remains a fourth copy of the `.env` decoder, and it is the one that builds the redaction set. After this change the on-disk escaped form and the delivered unescaped form diverge, and a credential containing `"` stops being redacted from logs. Before this PR the two agreed (both corrupted identically), so redaction worked — which makes this a regression in the direction CLAUDE.md's security rules explicitly forbid.

(The two `credential_sanitizer` hits in the diff are `import` lines in other files' context, not edits to it.)

`_format_pat_line` in `github_pat_propagation_service.py` is likewise still the second unmigrated writer whose docstring claims parity with the agent's own.

The encode/decode pair itself is good and `format_env_line` is the right extraction. Sequence remains #2010 → #2030 → #2024.

dolho:
@vybe — all three items fixed, and the backward-compat statement you asked for is below.

**1. The credential sanitizer now uses the shared decoder.** This was the important one: it builds the *redaction set*, so leaving it on the old decoder meant the set held the escaped form while executions received the unescaped one, and `sanitize_text()` stopped redacting a live credential from logs. Your reachability check matches mine — `startup.sh` never sources `.env`, so the `os.environ` branch of `_load_credential_values()` contributes nothing and the file branch is the only source.

It resolves `unquote_env_value` **by path**, not by import name: this file is loaded three ways (package, flat, standalone-by-path in `test_credential_sanitizer_agent.py`) and only the path is identical in all three — `services.execution_env` resolves to the **backend** package inside a test process, which is a different module entirely. My first attempt used the plain import and broke that test's collection, which is what surfaced it. Falls back to positional stripping *with a warning* if the sibling can't be loaded, so redaction degrades rather than disappearing.

**2. `_format_pat_line` now escapes the escape character first**, so the docstring's parity claim is true. Backend code can't import the agent-side module (different image), so the contract is held by a parity test over a shared corpus — both writers must emit the identical line, and the line must round-trip through the *real* reader.

**3. `format_env_line` refuses `\n`/`\r` rather than escaping them.** Escaping would make the encoding reversible on paper while every other `.env` consumer (a shell sourcing the file, the ent#127 probe) still reads two lines — and `PROTECTED_KEYS` refuses `LD_PRELOAD` but not `ANTHROPIC_API_KEY`/`GITHUB_PAT`/`CLAUDE_CODE_OAUTH_TOKEN`. `/update` maps it to a **400 naming the key**. Both writers refuse.

**`PARITY_FIXTURES` gained backslash coverage** — seven fixtures (escaped quote, escaped backslash, runs of two and four, trailing, single-quoted, unquoted). The corpus for the copy most at risk of drift previously had none of the class this PR changes.

```
sanitizer reverted to the old decoder -> 2 failed
pat writer escaping only the quote    -> 8 failed
as shipped                            -> 74 passed (file) / 159 with ent127 / 1264 related
```

**Backward compatibility, stated plainly** (also going into the PR body):

> This change re-interprets existing `.env` files with no migration and no detection. It **repairs** every quote-bearing value that was previously decoded corrupt — those agents start authenticating correctly on the next read. The cost: a pre-existing value containing **two or more consecutive backslashes** (Windows/UNC paths like `C:\\\\server\\share`) was written by the old escape-only-quote writer and is now decoded one level differently, so it is corrupted until re-pushed. `env_drift_report` will not catch it — it compares file↔process, not encoding vintage. Narrow (no credential format uses runs of backslashes; this affects hand-written path values) and now documented rather than discovered.

Sequencing #2010 → #2030 → #2024 unchanged. On #2010 I pinned the `/update` corruption as a `strict` xfail, and merging this branch on top flips it to XPASS(strict) — so whoever lands this is told to retire the marker.

vybe:
**Needs a real rebase before this can land — I stopped rather than hand-resolve it.**

Your three blocking items are fixed and I'm satisfied with them (the sanitizer resolving `unquote_env_value` **by path** is the right call given the three load modes, and refusing `\n`/`\r` rather than escaping is correct for the reason you give — every other `.env` consumer still reads two lines). The backward-compat statement is exactly what I asked for. This is purely a merge-mechanics hold.

#2010 landed as a **squash** (f744f349), so git can't see that this branch was stacked on it. `git merge origin/dev` produces five conflicts including **add/add** on `services/execution_env.py` and `tests/unit/test_1999_execution_env.py`.

The dangerous part: this branch carries #2010's **pre-review** commits (c4ea69a1, f42a7b25, 1747b049), not the version that shipped. Measured on the conflict stages:

| symbol | ours (this branch) | dev (final #2010) |
|---|---|---|
| `NODE_OPTIONS` | **0** | 2 |
| `GIT_SSH_COMMAND` | **0** | 1 |
| `PROTECTED_KEYS` entries | 11 | 23 |
| `unquote_env_value` | 3 | 0 |

So resolving `execution_env.py` as "ours" silently deletes the 12 Node/git `PROTECTED_KEYS` entries #2010 added *in response to my review* — `NODE_OPTIONS=--require` is `LD_PRELOAD` for Node, and `GIT_SSH_COMMAND` executes on the agent's next fetch. Taking "theirs" drops your decoder. The correct resolution is a union, and in a file whose entire job is refusing code-execution keys I'd rather the author do it than a merge-resolver.

**What's needed:**
1. `git merge origin/dev` (or rebase onto it) and resolve `execution_env.py` as a union — dev's `PROTECTED_KEYS` **plus** your `unquote_env_value` decoder. Same for `test_1999_execution_env.py` and `test_ent127_predicate.py`.
2. **Retire the `strict=True` xfail #2010 planted** — `test_1999_execution_env.py` pins the `/update` corruption and flips to XPASS(strict) once your encoding fix is in. Per your own note, whoever lands this drops it. Note #2014 has since deleted `/update` entirely, so check whether the marker's subject still exists at all.
3. `tests/registry.json` — trivial, rebuild from stages (dev + your new entries).

Everything else is approved-in-principle; re-request and I'll merge on the rebase.

vybe (approved):
Re-reviewed on the post-rebase head (3c126720). The resolution matches the prescription from my hold comment exactly:

1. **`execution_env.py` is the union** — verified dev's 12 Node/git `PROTECTED_KEYS` entries (`NODE_OPTIONS`, `GIT_SSH_COMMAND`, …) AND `format_env_line`/`unquote_env_value` are all present; the merge was done three-way against this branch's own #2010-tip (1747b049) as base, so the composition is structural, not hand-spliced.
2. **The strict xfail is retired** — converted to a plain regression test pinning that pre-#2023 `.env` files (old quote-only escaping, already on deployed agents' disks) still decode intact. It XPASSed on first local run, confirming the marker did its job.
3. **`registry.json` rebuilt from stages** — dev's 166 entries + the `test_2023` entry; both sides' `test_1999` entries were byte-identical.

Also: `routers/credentials.py` resolved to byte-identical-to-dev (#2014 deleted `/update`, and this PR's edits there die with it — the newline refusal stays pinned at function level for both writers), and `test_ent127`'s semantic anchor kept over dev's byte-faithful-line anchor, which this PR deliberately falsifies.

Verified locally: 627 passed — test_2023 / test_1999 / test_ent127 plus the full credential + PAT-propagation set, including #2014's `test_2008_dead_credentials_update` guard. No conflict markers tree-wide. Approving; auto-merge armed.

Issue #1855, opened 2026-07-28
bug: test_ent183_skill_packages poisons sys.modules — test_ent125_resilient_system_deploy fails collection when run after it


vybe:
## Symptom

On `dev` (e54ce3a9, pre-dates today's merges — reproduced on an untouched checkout):

```
pytest tests/unit/test_ent183_skill_packages.py tests/unit/test_ent125_resilient_system_deploy.py
```

fails at collection:

```
tests/unit/test_ent125_resilient_system_deploy.py:29: in <module>
    import routers.systems as systems
src/backend/services/agent_service/crud.py:37: in <module>
    from services.settings_service import ... resolve_github_pat ...
E   ImportError: cannot import name 'resolve_github_pat' from 'services.settings_service' (unknown location)
```

Both files pass in isolation (60 + 19 tests) and in full-suite order — the collision only appears in this reduced ordering, which is exactly the shape of a targeted "run the touched tests" pre-merge check (how it was found, validating #1845/#1847).

## Cause (hypothesis)

`test_ent183_skill_packages.py` leaves a stub or partially-initialized `services.settings_service` (or a competing `services` package) in `sys.modules`; `(unknown location)` in the ImportError is the tell for a namespace-package/stub left behind. The `lint (sys.modules pollution check)` CI job exists for this class but doesn't catch this pairing.

## Fix shape

Whatever module stubbing ent183 does should be scoped (monkeypatch.setitem on sys.modules, or an explicit teardown restoring the real module), and ideally the pollution-lint extended to catch a `services.*` module left in a non-importable state after a test file completes.

PR #2031, opened 2026-08-05
fix(tests): scope test_ent183_skill_packages' sys.modules stubs to one import (#1898)

### LINKED ISSUE #1898: bug(tests): test_ent183_skill_packages sys.modules stubs poison later tests (order-dependent CI flake)

dolho:
## Summary

`test_ent183_skill_packages.py` wrote fake modules straight into `sys.modules` at **import** time and left them there. That is collection-time code, and pytest imports every test module before running a single test — so every file collected afterwards resolved `services.settings_service` (plus `database`, `services.agent_client`, `utils.url_validation`) to a four-function fake, captured names from it at module scope, and kept those bindings for the rest of the session.

Its own autouse `sys.modules`-restore fixture never had a chance. By the time a fixture first runs, the damage is a **binding in another module's namespace**, and restoring the module table doesn't reach it.

Two symptom shapes, both of which read as unrelated bugs:

```
ImportError: cannot import name 'resolve_github_pat' from
'services.settings_service' (unknown location)          # at COLLECTION (#1855)
```

…and silent wrong behaviour when the stub happens to define the name — the 7 failures in `test_1081_physical_meter.py` this issue reproduces.

## Fix

The stubs now live inside a `mock.patch.dict` scoped to the single import that needs them. Nothing survives the `with`; the imported module keeps its stub bindings, which is all the stubbing was ever for.

## A false start worth reading

My first version *also* evicted cached `services.skill*` modules, to force a fresh import against the stubs regardless of collection order. It broke **10 sibling tests** in `test_ent236_skills_lifecycle` and `test_ent237_skill_sources`: evicting and re-importing produces a **second module object**, so their monkeypatching lands on a module the code under test isn't using.

I caught it by running the same selection with and without my change — 435 passed without, 10 failed with — not by reading the diff. Reverted, with the reasoning left in the file: the order-dependence it addressed is a real wart, but it isn't what #1898 is about, and *scope creep in a test-isolation fix is how a test-isolation fix breaks tests*.

## Verification

| Case | Before | After |
|---|---|---|
| The issue's reproduction (`+ test_1081_physical_meter`) | 7 failed | **74 passed** |
| #1855's pair (`+ test_ent125_resilient_system_deploy`) | collection error | **79 passed** |
| `+ test_ent236_skills_lifecycle` | error | **124 passed** |
| The selection my false start regressed | 435 passed | **436 passed, 0 failed** |

3 guard tests, mutation-verified — reintroducing the bare module-scope assignment fails all three. The static guard's AST walk deliberately **allows** `sys.modules` assignment inside a `with`: a rule banning every assignment would ban the fix along with the bug.

## One correction

When I triaged this I said it explained a `test_subscription_auto_switch_pingpong` flake I'd seen on #2025. **I can't substantiate that.** The error signature matched #1855's family, but the pairing doesn't reproduce pre-fix, and neither does a broader selection. `test_ent236_skills_lifecycle` *is* a confirmed victim (error → 124 passed); the pingpong one stays unexplained.

## Related

**#1855 is the same root cause with a different victim** and should close as a duplicate once this lands.

Closes #1898

### DIFF
--- a/tests/registry.json
+++ b/tests/registry.json
@@ -1575,6 +1575,18 @@
         "security"
       ],
       "description": "Per-source skills-library root (ent#332): catalog.yaml skills_root: declaration -> evidence-gated skills/ probe -> .claude/skills/ fallback, resolved per SkillSourceClone; packaging rewrites tar arcnames to the canonical agent-side .claude/skills/ destination at filter_skill_archive so manifests, prune confinement and the ent#236 removal machinery stay destination-canonical with zero migration. Covers: resolution order + zero-config back-compat; dual-layout (evidence under BOTH roots, no catalog) keeps .claude/skills/ and flags layout_conflict instead of silently flipping which executable content ent#236 auto-injects; segment-wise skills_root validation ('.', './skills', 'skills//x' all rejected -- a whole-string charset regex admits each and breaks archive-prefix math into per-skill empty-package failures); every bad catalog shape (unparseable, alias-bearing HardenedYamlError, non-mapping, unknown schema_version, oversized, symlinked) falls back to probe and never raises through list_skills; symlinked probe/declared roots fall to the next tier rather than blanking the source, and an escaping final tier empty/None-propagates without a TypeError; tree-SHA version parity across layouts (repo restructure with unchanged content keeps versions stable -- no spurious fleet re-inject); arcname rewrite identity for legacy layout, destination-canonical executable_paths, fail-visible layout-mismatch, unsafe source_root ValueError; cross-layout prune continuity; _legacy_fallback SKILL.md lookup with rewritten members; end-to-end mixed-layout merge through the real service with per-source skills_root in library status."
+    },
+    {
+      "file": "unit/test_1898_sys_modules_isolation.py",
+      "feature": "#1898",
+      "added": "2026-08-05",
+      "categories": [
+        "backend",
+        "unit",
+        "test-isolation",
+        "reliability"
+      ],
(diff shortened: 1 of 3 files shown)


vybe (approved):
The diagnosis is the valuable part here: the damage is a **binding in another module's namespace**, so an autouse `sys.modules`-restore fixture structurally cannot undo it — by the time any fixture runs, the collection-time import has already captured names. That explains both symptom shapes (the #1855 collection ImportError and the silent-wrong-behaviour failures in `test_1081_physical_meter`) from one cause, which is why scoping the stubs to `mock.patch.dict` around the single import is the whole fix rather than a mitigation.

The false start is worth as much as the fix. Evicting cached `services.skill*` modules produces a **second module object**, so sibling monkeypatching lands on a module the code under test isn't using — and you found it by running the selection with and without the change (435 vs 10 failed), not by reading the diff. Leaving that reasoning in the file is right, and so is the conclusion: scope creep in a test-isolation fix is how a test-isolation fix breaks tests.

The static guard deliberately allowing `sys.modules` assignment inside a `with` is the detail that keeps it from banning its own fix.

Landing this ahead of #2041, which independently reported this exact flake as pre-existing.

Issue #1897, opened 2026-07-30
bug: a failed canary Slack post is counted as delivered and never retried


webmixgamer:
## Summary

When the canary's Slack webhook post fails, the failure is logged and swallowed. The caller cannot tell — `emit_transition` is `-> None` — so it records the transition anyway, the cycle cursor advances, and continuing-red gating guarantees the alert is **never retried**. The alert is lost, and `cumulative_transitions` says it was delivered.

## The exact path

`services/canary_alerts.py` — the post result is read, then dropped:

```python
success, error = await slack_service.post_webhook(webhook_url, text, blocks=blocks)
if not success:
    logger.warning(
        "canary slack webhook failed for %s: %s (cycle continues, row persisted)",
        invariant_id, error,
    )
```

The method is annotated `-> None`. `success` never leaves the function.

`services/canary_service.py` — the caller treats "did not raise" as "delivered":

```python
try:
    await CanaryAlerts.emit_transition(...)
    transition_ids.append(inv_id)
except Exception:
    logger.exception(...)
```

Then, unconditionally:

```python
self.cumulative_transitions += len(transition_ids)
self._write_prev_cycle_at(snapshot.snapshot_time)
```

## Why it is never retried

`_is_green_to_red` returns `prev["snapshot_time"] < prev_cycle_at`.

1. **Cycle N** — invariant fires. The `canary_violations` row persists (Slack is not involved). `_is_green_to_red` → `True`. Slack rejects the post. Warning logged, no raise. `transition_ids.append(inv_id)` runs. Cursor advances to `T_N`.
2. **Cycle N+1** — invariant still firing. Latest persisted violation is now `T_N`; cursor is `T_N`. Check is `T_N < T_N` → **`False`** → continuing-red → no notification.

The alert is never sent, never retried, and the counter reports success.

## The existing defense has exactly this hole

`canary_service.py` already comments the cursor write:

> *"Done AFTER notifications so a crash mid-emit doesn't advance the cursor and silence a real transition on retry."*

The failure mode was anticipated — but only the **raise** path is defended. `emit_transition` swallows instead of raising, so it walks straight through the guard the comment describes.

## Why it bites at the worst time

[...]

Issue #1823, opened 2026-07-27
bug: legacy IANA timezone aliases (Europe/Kiev et al.) crash schedule creation with a 500 — pytz accepts, APScheduler's zoneinfo lookup can't resolve them inside the shipped image


webmixgamer:
## Summary

Creating a schedule with a **legacy IANA timezone alias** (`Europe/Kiev`, `Asia/Calcutta`, `Asia/Saigon`, …) returns a raw **500 Internal Server Error** instead of the named validation error, on both the REST API and the MCP `create_agent_schedule` tool. Three braided defects:

1. **Image packaging (root):** the backend/scheduler images are Debian 13 "trixie" based (`tzdata 2026b`), where the IANA backward-compatibility links live in the separate **`tzdata-legacy`** package — which is not installed. `/usr/share/zoneinfo/Europe/` contains `Kyiv` but no `Kiev` link, and no pip `tzdata` wheel is present as a fallback.
2. **Validator robustness:** `services/schedule_validation.py` (#1472) is split-brained on these zones. `validate_timezone` checks with **pytz**, which bundles its own tz database *including* the legacy aliases → accepts. `validate_cron_expression` then builds an APScheduler `CronTrigger(timezone=tz)`, and modern APScheduler re-resolves the zone by IANA key through **system `zoneinfo`** → `ZoneInfoNotFoundError`. That exception is a **`KeyError` subclass**, so the rescue clause `except (ValueError, TypeError)` misses it and it escapes as an unhandled 500 — violating the named-validation product-quality bar and the #1472 contract ("reject exactly what the scheduler would").
3. **Env-dependent test:** `tests/unit/test_1472_schedule_validation.py` parametrizes `("0 4 * * *", "Europe/Kiev")` as a *valid* case. It passes on GitHub's Ubuntu runners (whose system tzdata ships the backward links) and fails inside the built image — green CI, red image; the class `verify-local` exists to catch.

## Context

Found while running the unit suite **inside the backend image** during #1809 verification (it fails identically at the #1809 merge-base — pre-existing, unrelated to that change). Reproduced end-to-end on a live instance.

[...]

PR #2033, opened 2026-08-05
feat(templates): serve the agent catalog from a remote registry (trinity-enterprise#14)

AndriiPasternak31:
Refs trinity-enterprise#14

(`Refs`, deliberately not a closing keyword — the issue lives in the private tracker and GitHub's auto-close/auto-promote is same-repo only. **After merge, `status-in-dev` on [abilityai/trinity-enterprise#14](https://github.com/abilityai/trinity-enterprise/issues/14) has to be set by hand.**)

## What this does

The GitHub half of the agent-template catalog is sourced at **runtime** from a `registry.yaml` fetched over HTTPS, so curating an install's starter fleet is a vendor file edit rather than a Trinity release.

Purely **additive**. `DEFAULT_GITHUB_TEMPLATE_REPOS` has been `[]` since #1931, so this fills a branch that is empty today, and an admin-curated list still wins outright — that install never even issues the fetch. Precedence: **admin DB override → remote registry → bundled list**.

**Fail-open is structural, not an `except` branch.** `get_all_templates()` returns `local + github`; `local` needs no network, and the registry can only *add* to `github`. So unreachable / 5xx / timeout / malformed / alias-bombed / oversize / redirected / empty all reduce `github` toward `[]` — the already-shipping default state. Proven mode-by-mode through the real parser and a real transport, asserting on **catalog output** rather than on whether an except branch ran.

Registry entries are parsed into a frozen **four-field allowlist** (`repo`, `display_name`, `description`, `priority`) and never splatted, so a registry cannot assert `fork_to_own`, `credentials`, `schedules`, `data_paths`, `hidden` or `id`. The requirements entry states the uncomfortable half plainly: **`repo` is a capability pointer** — choosing the repo chooses which `template.yaml` Trinity fetches and trusts, so the allowlist bounds the *direct* blast radius, not the indirect one.

## Review-round fixes (the last two commits)

Four findings, each a case where the control was present and correct one level *above* where the attack lands. Worth reading as a set — that is the pattern.

**S1 — a byte ceiling on *decoded* bytes is not a byte ceiling.** The fetch streamed under a 256 KiB running total with a docstring correctly explaining why `Content-Length` can't be the gate. Both true, both defeated one level down: a custom `headers=` dict does not override httpx's default `Accept-Encoding: gzip, deflate`, and `iter_bytes()` yields *decompressed* chunks. A 199 KiB wire body — under the cap, so the early abort passes it — inflated ~1030:1 to **458 MB** of transient allocation on the event-loop thread. Now counts wire bytes via `iter_raw()` and refuses any `Content-Encoding` outright; same body peaks at 0.01 MB. *The harness hid it:* `httpx.Response(content=…)` decodes in the **constructor**, so every fixture built a response the production path cannot even read — the byte-ceiling tests were green against a shape incapable of expressing the failure.

**S2 — the `fork_to_own` gate read the catalog cache, not its own read.** It consulted metadata from the global platform PAT, off the default branch, through a 600 s cache — while `crud.py` had already read the same file *one line earlier* with the caller's resolved PAT and parsed ref. Both directions broke: GitHub answers **404, not 403**, for a repo a token cannot see, so a private `fork_to_own: required` template readable only by the creator's PAT was classified *absent*, the gate passed, and the agent bound to the shared upstream — the exact outcome the gate exists to prevent, no attacker involved; and a poisoned cache entry 503'd creators whose own PAT read the file fine. Note that **fixing only the availability half would not have closed the false pass** — the `fork_to_own` *value* is `None` for the same reason the verdict is wrong, so both inputs now come from the creation-path read, and `required` from *either* source enforces (a union — can only remove a false pass).

**S3 — RFC 6598 (`100.64.0.0/10`) was admitted.** Python's `ipaddress` reports CGNAT as neither `is_private` nor `is_reserved`. Not reachable in Trinity's own topology, but several cloud providers address internal endpoints there. Both `/10` boundaries held public so a legitimate registry isn't blackholed.

**Packaging — the hard kill switch was inert.** Caught by `/validate-pr`. `config.py` read `TEMPLATE_REGISTRY_ENABLED` / `TEMPLATE_REGISTRY_URL`; **neither compose file injected them**. Prod compose launches standalone (no base merge, no `env_file:`) and is what every deploy path uses, so both levers did nothing on every real instance while working fine on a laptop — a default-ON feature making outbound requests with no way to stop it. The irony is the reason it's now pinned rather than just fixed: the comment block *directly above* that `os.getenv` argues at length against `_resolve_bool_flag` **because it would ship an inert kill switch (#1039 class)** — and the flag shipped inert anyway, by the other route, ~15 lines away. Sixth recurrence of the #1039/#1056 class.

## Verification

- **Full unit suite: 8783 passed, 17 skipped, 0 failed** — run twice, fixed-order and pytest-randomly, post-rebase onto `dev` @ ef3b0ebf.
- Packaging proven by **rendering, not grepping** — `docker compose config` shows both vars on `backend` in each file, and `TEMPLATE_REGISTRY_ENABLED=false` propagates through to `"false"`.
- The new packaging guard was proven by deleting the real prod line and watching it go red. *(First attempt at that proof used `grep -v 'VAR=${VAR'`, where `$` is an end-of-line anchor — nothing was removed and the suite reported a false green against an unmodified tree. Noted in the guard's docstring.)*
- `enterprise-docs-guard` pattern run locally: clean. `tests/lint_sys_modules.py`: 194 violations vs baseline 240, no new.
- Invariant #5 vendored `safe_yaml.py` mirror: byte-identical.

## Rebase onto current `dev` (ef3b0ebf)

Rebased over 38 new `dev` commits. Four conflicts, all additive-vs-additive
(`feature-flows.md` and `learnings.md` index rows, `tests/registry.json`
entries, and `url_validation.py`, where dev's #2052/ent#347 authority grammar
and this branch's CGNAT constant are disjoint module-level additions). Each
resolution was verified as a **union**: zero removed lines against either
side for the two files that merge as text, and the registry JSON proven
lossless set-wise (dev 142 ∪ this branch 137 = 150 merged, no dupes).

Two suites then needed adapting, both because the change that landed under
them is the one that is right:

- **ent#15 (#2040)** inserted an import-intent gate into `_resolve_template`
  ahead of the fork gate; two `SimpleNamespace` config doubles predate the
  field and raised `AttributeError` before reaching their assertion. A stub
  gap, not a production one — the S2 fail-closed gate is untouched, and dev's
  `copy` intent carries its own `fork_to_own: required` refusal.
- **#1890** made `require_admin` itself reject agent-scoped principals. This
  branch asserted an agent key could still *read* the registry setting, on
  the reasoning that the human check belongs on the writes. Overruled, and
  correctly: the GET returns the registry URL plus a live `status` block,
  which for a private catalog is exactly the pointer an agent should not be
  able to enumerate. The explicit `reject_agent_principal` on PUT/DELETE is
  kept as belt-and-braces.

## For the reviewer — one thing I found but did not fix

Dev's new `copy` import intent gates `fork_to_own: required` from
`get_github_template(template_lookup)` — the **catalog** read. That is the
same shape as S2 above: the catalog read uses the global platform PAT off the
default branch through a 600 s cache, so a private `required` template reads
as absent (GitHub answers 404, not 403) and the gate passes. It is materially
less severe on that path than on the fork path — a copy is a detached
snapshot with no GitHub env, no PAT and no push, so the outcome S2 prevents
(the agent bound to the shared upstream) cannot occur; what is lost is the
user-owned *remote*, not containment. Left alone deliberately: it is dev's
code, fixing it here is scope drift, and it deserves its own issue. Flagging
it because it is the seventh instance of "the create path is never one call
site" and worth a decision rather than a silent inheritance.

## Docs

Requirements TMPL-002 (`requirements/core-agent.md` §4.2.2), `architecture.md`, new `feature-flows/template-registry.md` + index row, `platform-settings.md` / `library-page.md` / `template-processing.md` / `scheduling.md` cross-refs, and three `learnings.md` entries for the transferable classes.

## Ship prerequisite (not code)

A valid document — possibly empty (`version: 1` / `templates: []`) — must exist at the default URL before or with the release. Day-one behaviour is then "fetch succeeds, zero entries, catalog unchanged, no warnings", and publishing the curated fleet becomes a pure content edit.

## Named residuals

No signature/provenance verification of the registry document; no allowlist on which repo owners a registry may list; DNS rebinding not closed (validate-then-connect TOCTOU — accepted for v1: admin-and-human-set URL, parsed into a display-only allowlisted record, never reaching a deserialize path). The first two belong to the private/per-customer-catalog phase, where the trust model is genuinely different.


vybe (approved):
Verified rather than taken on trust:

**Invariant #5 (vendored mirror)** — `src/backend/utils/safe_yaml.py` and `docker/base-image/agent_server/safe_yaml.py` are **byte-identical** (14,728 bytes, `diff` clean). The #1965/ent#314 parity guards pass (37 tests).

**Config packaging (#1056 class)** — `TEMPLATE_REGISTRY_ENABLED` and `TEMPLATE_REGISTRY_URL` are each wired into `docker-compose.yml`, `docker-compose.prod.yml` **and** `.env.example`. No inert lever on prod, which launches standalone. Build packaging is fine by construction — the backend Dockerfile now uses `COPY src/backend/*.py`, so #1033's explicit-name hazard no longer applies.

**Security (the CSO diff's three findings, spot-checked independently)**
- S3 (CGNAT): `100.64.0.0/10` is not merely *defined* — it is enforced in the deny predicate at `url_validation.py:428`, with the correct rationale that Python's `ipaddress` reports it as neither `is_private` nor `is_reserved`. The test fixtures pin the exact boundary (`100.63.255.255`, `100.128.0.1`, `99.64.0.1`, `101.64.0.1`).
- S1 (decompression bomb): the fix is `iter_raw()` + `Accept-Encoding: identity` + **rejecting** a non-identity encoding rather than trusting the declared length. Correct — peak allocation is the resource under attack, not the total, and `iter_bytes()` is exactly the difference.

**Tests** — 374 pass across the eight ent#14 files; 230 more across the adjacent suites (`test_1484_create_agent_characterization`, `test_fork_to_own`, `test_ent89_*`, `test_1931_*`).

## Merge resolution I pushed (f7bacbfa)

Merged `origin/dev` in rather than rebasing, so a contributor's branch isn't force-pushed. Only `tests/registry.json` conflicted; every doc file auto-merged.

Both sides were **purely additive** — dev added the four #1823 entries (#2071), this branch adds eight ent#14 entries, with no shared entry modified and none removed on either side. Rebuilt from the git stages rather than splicing markers, preserving the file's `indent=2` + trailing-newline formatting. Result: 154 entries, diff vs dev has **zero deletions** and exactly 8 additions. Whole-tree scan confirms no stray conflict markers.

**Post-merge, all six PRs in this batch pass together: 540 passed, 2 skipped** (ent#14 + #1823 + #2068 + #1897 + adjacent).

One non-blocking observation for a future issue: `url_validation.py` now holds two IP-deny predicates with different deny sets — `validate_skills_library_url` (L322: private/loopback/reserved/link-local) and `validate_template_registry_url` (L422–428: the same plus multicast, unspecified and CGNAT). Defensible today because the skills-library validator is locked to a github.com allowlist earlier in the same function, so the weaker set isn't reachable the same way. But if that allowlist is ever loosened the asymmetry becomes a bypass, and the two are far enough apart in the file that nobody will notice. Worth folding into one shared predicate eventually.

PR #2038, opened 2026-08-06
fix(git-sync): stop the base image's container-only .claude/settings.json leaking into agent repos (#2036)

### LINKED ISSUE #2036: bug: base image bakes container-only hook paths into ~/.claude/settings.json, git sync commits it, external clones of the agent repo are bricked
## Summary

The agent base image bakes a `.claude/settings.json` with **container-only** hook paths (`/opt/trinity/hooks/*.py`) into the agent's HOME — which is also the git-synced repo root. The canonical sync gitignore does not exclude this file, and auto-sync commits with a blanket `git add .`, so the file (plus several other `.claude/` runtime artifacts) silently leaks into the agent's GitHub repo on the next sync. Any clone of that repo outside the container (e.g. an operator's laptop running Claude Code against the same repo) is then **hard-bricked**: every Bash/Edit/Write tool call is blocked, because the registered PreToolUse hook scripts don't exist there.

The bricking is deterministic, not incidental: `python3 /opt/trinity/hooks/bash-guardrail.py` exits with code **2** when the script file is missing, and exit code 2 is exactly Claude Code's "block this tool call" signal for PreToolUse hooks.

## Component

Agent Runtime (base image) + Backend Git Sync (`git_service`)

## Priority

P1 — the leak happens silently on every sync for every repo-synced agent; the failure surfaces later, on a different machine, with no obvious connection to its cause.

## Error

On an external clone, every tool call fails with the hook error:

```
/usr/bin/python3: can't open file '/opt/trinity/hooks/bash-guardrail.py': [Errno 2] No such file or directory
PreToolUse hook exited with code 2 → tool call blocked
```

## Location

- **File**: `docker/base-image/Dockerfile` line ~136 — `COPY ./hooks/claude-settings.json /home/developer/.claude/settings.json`
- **File**: `src/backend/services/git_service.py` — `_GITIGNORE_PATTERNS` (~line 999) and the sync commit's blanket `git add .` (~line 1305)

## Root Cause

Three facts compose:

1. The base image installs hook registration at `~/.claude/settings.json` with absolute container paths. Since agent HOME == git repo root, the file sits inside the synced working tree (seeded into the workspace volume on first mount).
2. Auto-sync commits with `git add .`.
3. `_GITIGNORE_PATTERNS` carefully excludes seven other `.claude/` runtime paths (`projects/`, `statsig/`, `todos/`, `debug/`, `sessions/`, `shell-snapshots/`, `plugins/`) — but **not** `.claude/settings.json`, and also not `.claude/backups/`, `.claude/policy-limits.json`, `.claude/remote-settings.json`, or `.claude/.last-cleanup`, all of which were observed leaking in the same sync commit (the backups alone added ~3,000 lines of runtime state).

Same defect class as #462 (runtime-state leak, 1,599 files) and #1702 (plugin cache leak) — this is the next missing entry in the pattern list, with a worse blast radius because the leaked file actively breaks foreign clones instead of just bloating the repo.

## Reproduction Steps

1. Create an agent with GitHub sync enabled (HOME as repo root).
2. Let the agent do any work that triggers an auto-sync push (the base image's `.claude/settings.json` is untracked and unignored, so `git add .` sweeps it in).
3. Clone the agent repo on a machine outside the container (e.g. a laptop) and run Claude Code in it.
4. Any Bash/Edit/Write call is blocked: the PreToolUse hooks point at `/opt/trinity/hooks/*.py`, which don't exist → exit 2 → block.

## Suggested Fix

Add the missing runtime artifacts to `_GITIGNORE_PATTERNS` (the existing `_migrate_workspace_gitignore` machinery will then untrack already-committed copies on the next sync, same as #1596/#1702):

```python
# Claude Code runtime — commit commands/skills/agents, exclude runtime data
".claude/settings.json",     # baked by base image; container-only hook paths
".claude/remote-settings.json",
".claude/policy-limits.json",
".claude/backups/",
".claude/.last-cleanup",
```

Longer-term, consider registering the platform hooks somewhere outside the synced tree entirely, so container-specific config can never enter the repo in the first place.

## Environment

- Trinity version: observed on v0.8.5-era code; the exposure is longstanding (present since the base-image hooks + repo-root HOME design; current dev `4d743fdb` still affected)
- Agent with GitHub sync enabled (`source_mode` repo-root)

## Related

- `docker/base-image/hooks/claude-settings.json` (the baked template)
- `src/backend/services/git_service.py` (`_GITIGNORE_PATTERNS`, `_migrate_workspace_gitignore`)
- Precedents: #462, #1596, #1702

dolho:
Fixes #2036

## What

The agent base image bakes `~/.claude/settings.json` registering the platform guardrail hooks by **absolute container path** (`/opt/trinity/hooks/*.py`). Agent HOME is also the git repo root and the in-container auto-sync commits with `git add -A`, so the file was swept into the agent's GitHub repo on the next sync.

A clone of that repo made anywhere without `/opt/trinity` is then hard-bricked. The mechanism is deterministic, not incidental: a PreToolUse hook whose script is missing exits **2**, and exit 2 is exactly Claude Code's "block this tool call" signal — so every Bash/Edit/Write fails, on a machine with no obvious connection to the agent that leaked it.

Adds `.claude/settings.json` plus the four other runtime artifacts observed in the same sync commit (`remote-settings.json`, `policy-limits.json`, `backups/`, `.last-cleanup`) to the canonical `_GITIGNORE_PATTERNS`.

Same leak class as #462 / #1596 / #1702 — the next missing entry in that list, with a worse blast radius, since this one *breaks* foreign clones rather than merely bloating them.

## How the already-broken repos heal

The fix is only worth anything if it reaches agents that have already pushed. It does, through the existing `_migrate_workspace_gitignore` machinery (#462): every Push merges the canonical list into the workspace `.gitignore` and then `git rm --cached`s anything that now matches. That is **index-only** — the working-tree file stays, so the *running* agent keeps its guardrail hook registration while the next push deletes the poisoned copy from the remote. Asserted in `test_existing_committed_settings_is_untracked_but_kept_on_disk`.

Second distribution path, for free: `compatibility/static_checks.py` computes its missing-pattern set dynamically from `_GITIGNORE_PATTERNS`, and `compatibility/fixes.py` appends them — so affected agents surface in the Overview compatibility panel and can be auto-fixed there without waiting for a Push.

## Trade-off, stated rather than buried

`.claude/settings.json` is also Claude Code's **project-level** settings file, so after this a template can no longer commit one. That is the right default — the base image always writes that path, so a template-supplied copy was already colliding with a container-only file — and the #1596 escape hatch still works: an agent that genuinely needs it negates the rule in its own `.gitignore` (`!.claude/settings.json`). There is a test for the escape hatch, because a trade-off is only acceptable if the opt-out is real.

## Changes

| File | What |
|---|---|
| `src/backend/services/git_service.py` | the five patterns + the rationale |
| `docs/TRINITY_COMPATIBLE_AGENT_GUIDE.md` | canonical block — **required**, `test_doc_and_constant_in_sync` pins `constant ⊆ doc` |
| `config/agent-templates/*/.gitignore` (14) | regenerated via the #1908 regenerator, not hand-edited |
| `tests/unit/test_2036_claude_settings_leak.py` | **new** — 12 tests |

One line touched outside the strict scope: the doc block's `# Keep: … settings.local.json` clause was already wrong (`*.local.json` ignores it) and sits directly beside the new rule, where it would have read as a contradiction. Dropped the clause; the rest of the comment stands.

## Tests

`tests/unit/test_2036_claude_settings_leak.py` asserts at two levels deliberately.

**Behavioural, against real `git` and real `bash`** — the shipped `_build_gitignore_merge_command` / `_build_rm_cached_ignored_command` run against a temp repo. A pattern-list assertion alone would pass against a rule git does not actually apply (`.claude/.last-cleanup` is a dotfile inside a dotdir, `.claude/backups/` is a directory rule), so the test drives the real thing and reproduces the auto-sync's own `git add -A`.

**The premise** — that the baked settings file really does carry absolute `/opt/trinity` paths, and that the Dockerfile still writes it inside the repo root. Those two facts are the entire reason the rule must exist. If a later change makes the hook paths relative, or moves registration out of the synced tree (the issue's suggested longer-term fix), these fail and a reviewer gets to re-decide the rule rather than inherit a dead one.

Every check also pins that legitimate `.claude/` source (`commands/`, `skills/`, `agents/`) keeps syncing — the `.trinity/setup.sh` trap already documented in `_build_rm_cached_ignored_command`, where an over-broad ignore silently untracks an agent's own files and pushes the deletion.

```
tests/unit/test_2036_claude_settings_leak.py            12 passed
tests/unit/test_github_init_gitignore.py
  + test_1908_bundled_template_gitignore.py
  + test_1596_git_sync_observability.py
  + test_data_paths_gitignore.py
  + test_953_startup_sh_no_gitignore_writes.py         111 passed
full tests/unit/                            7667 passed, 17 skipped
```

**Mutation check.** Reverting only `git_service.py` and rerunning fails **7 of 12** — including both behavioural tests. A guard never shown to fire is decorative.

The 4 failures in the full run (`test_1472_schedule_validation.py::test_accepts_valid[Europe/Kiev]`, 3 in `test_admin_email_login.py`) reproduce identically on a clean `dev` with this branch stashed — pre-existing, unrelated.

## Not fixed here

- **The `.gitignore` merge only runs on a backend-mediated Push/init.** The in-container 15-min auto-sync never updates `.gitignore` — deliberately, per #953, which removed startup.sh's shell-level append because it produced permanent `M .gitignore` drift. So an agent whose operator never Pushes keeps leaking until the compatibility auto-fix is applied. Pre-existing and identical for #1596/#1702; worth its own issue rather than reopening #953 inside a P1 bug fix.
- **The longer-term fix the issue suggests** — registering the platform hooks outside the synced tree entirely — is not attempted. The obvious lever (`CLAUDE_CONFIG_DIR`) relocates *all* of `~/.claude`, including `projects/<uuid>.jsonl`, which the Session tab reaper and `jsonl_recovery` both address by fixed path. That is a real change with real blast radius, not a bug fix.

### DIFF
--- a/config/agent-templates/dd-bizmodel/.gitignore
+++ b/config/agent-templates/dd-bizmodel/.gitignore
@@ -67,7 +67,16 @@ __pycache__/
 .claude/sessions/
 .claude/shell-snapshots/
 .claude/plugins/
-# Keep: .claude/commands/, .claude/skills/, .claude/agents/, settings.local.json
+# settings.json is baked by the Trinity base image with container-only hook
+# paths (/opt/trinity/hooks/*.py). Committing it bricks any clone made outside
+# the container: the missing hook script exits 2, which Claude Code reads as
+# "block this tool call", so every Bash/Edit/Write fails there. (#2036)
+.claude/settings.json
+.claude/remote-settings.json
+.claude/policy-limits.json
+.claude/backups/
+.claude/.last-cleanup
+# Keep: .claude/commands/, .claude/skills/, .claude/agents/
 
 # Temporary files
 *.log
--- a/config/agent-templates/dd-captable/.gitignore
+++ b/config/agent-templates/dd-captable/.gitignore
@@ -67,7 +67,16 @@ __pycache__/
 .claude/sessions/
 .claude/shell-snapshots/
 .claude/plugins/
-# Keep: .claude/commands/, .claude/skills/, .claude/agents/, settings.local.json
+# settings.json is baked by the Trinity base image with container-only hook
+# paths (/opt/trinity/hooks/*.py). Committing it bricks any clone made outside
+# the container: the missing hook script exits 2, which Claude Code reads as
+# "block this tool call", so every Bash/Edit/Write fails there. (#2036)
+.claude/settings.json
+.claude/remote-settings.json
+.claude/policy-limits.json
+.claude/backups/
+.claude/.last-cleanup
+# Keep: .claude/commands/, .claude/skills/, .claude/agents/
 
 # Temporary files
 *.log
--- a/config/agent-templates/dd-competitor/.gitignore
+++ b/config/agent-templates/dd-competitor/.gitignore
@@ -67,7 +67,16 @@ __pycache__/
 .claude/sessions/
 .claude/shell-snapshots/
 .claude/plugins/
-# Keep: .claude/commands/, .claude/skills/, .claude/agents/, settings.local.json
+# settings.json is baked by the Trinity base image with container-only hook
+# paths (/opt/trinity/hooks/*.py). Committing it bricks any clone made outside
+# the container: the missing hook script exits 2, which Claude Code reads as
+# "block this tool call", so every Bash/Edit/Write fails there. (#2036)
+.claude/settings.json
+.claude/remote-settings.json
+.claude/policy-limits.json
+.claude/backups/
+.claude/.last-cleanup
+# Keep: .claude/commands/, .claude/skills/, .claude/agents/
 
 # Temporary files
 *.log
--- a/config/agent-templates/dd-compliance/.gitignore
+++ b/config/agent-templates/dd-compliance/.gitignore
@@ -67,7 +67,16 @@ __pycache__/
 .claude/sessions/
 .claude/shell-snapshots/
 .claude/plugins/
-# Keep: .claude/commands/, .claude/skills/, .claude/agents/, settings.local.json
+# settings.json is baked by the Trinity base image with container-only hook
+# paths (/opt/trinity/hooks/*.py). Committing it bricks any clone made outside
+# the container: the missing hook script exits 2, which Claude Code reads as
+# "block this tool call", so every Bash/Edit/Write fails there. (#2036)
+.claude/settings.json
+.claude/remote-settings.json
+.claude/policy-limits.json
+.claude/backups/
+.claude/.last-cleanup
+# Keep: .claude/commands/, .claude/skills/, .claude/agents/
 
 # Temporary files
 *.log
--- a/config/agent-templates/dd-founder/.gitignore
+++ b/config/agent-templates/dd-founder/.gitignore
@@ -67,7 +67,16 @@ __pycache__/
 .claude/sessions/
 .claude/shell-snapshots/
 .claude/plugins/
-# Keep: .claude/commands/, .claude/skills/, .claude/agents/, settings.local.json
+# settings.json is baked by the Trinity base image with container-only hook
+# paths (/opt/trinity/hooks/*.py). Committing it bricks any clone made outside
+# the container: the missing hook script exits 2, which Claude Code reads as
+# "block this tool call", so every Bash/Edit/Write fails there. (#2036)
+.claude/settings.json
+.claude/remote-settings.json
+.claude/policy-limits.json
+.claude/backups/
+.claude/.last-cleanup
+# Keep: .claude/commands/, .claude/skills/, .claude/agents/
 
 # Temporary files
 *.log
--- a/config/agent-templates/dd-intake/.gitignore
+++ b/config/agent-templates/dd-intake/.gitignore
@@ -67,7 +67,16 @@ __pycache__/
 .claude/sessions/
 .claude/shell-snapshots/
 .claude/plugins/
-# Keep: .claude/commands/, .claude/skills/, .claude/agents/, settings.local.json
+# settings.json is baked by the Trinity base image with container-only hook
+# paths (/opt/trinity/hooks/*.py). Committing it bricks any clone made outside
+# the container: the missing hook script exits 2, which Claude Code reads as
+# "block this tool call", so every Bash/Edit/Write fails there. (#2036)
+.claude/settings.json
+.claude/remote-settings.json
+.claude/policy-limits.json
+.claude/backups/
+.claude/.last-cleanup
+# Keep: .claude/commands/, .claude/skills/, .claude/agents/
 
 # Temporary files
 *.log
--- a/config/agent-templates/dd-lead/.gitignore
+++ b/config/agent-templates/dd-lead/.gitignore
@@ -67,7 +67,16 @@ __pycache__/
 .claude/sessions/
 .claude/shell-snapshots/
 .claude/plugins/
-# Keep: .claude/commands/, .claude/skills/, .claude/agents/, settings.local.json
+# settings.json is baked by the Trinity base image with container-only hook
+# paths (/opt/trinity/hooks/*.py). Committing it bricks any clone made outside
+# the container: the missing hook script exits 2, which Claude Code reads as
+# "block this tool call", so every Bash/Edit/Write fails there. (#2036)
+.claude/settings.json
+.claude/remote-settings.json
+.claude/policy-limits.json
+.claude/backups/
+.claude/.last-cleanup
+# Keep: .claude/commands/, .claude/skills/, .claude/agents/
 
 # Temporary files
 *.log
--- a/config/agent-templates/dd-legal/.gitignore
+++ b/config/agent-templates/dd-legal/.gitignore
@@ -67,7 +67,16 @@ __pycache__/
 .claude/sessions/
 .claude/shell-snapshots/
 .claude/plugins/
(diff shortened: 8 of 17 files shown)


dolho:
## Verified against the real agent base image

Ran end-to-end against `trinity-agent-base:latest` (`sha256:0f28f1f1`, built 2026-07-27 — the image the fleet actually runs). No rebuild: this PR touches no Dockerfile, so the shipped image is the honest test subject.

Every command below is the **real** string from the patched `_build_gitignore_merge_command` / `_build_rm_cached_ignored_command`, delivered over the same `docker exec` transport `_migrate_workspace_gitignore` uses.

### 1. Premise — the baked file is container-only

```
$ docker run --rm trinity-agent-base:latest cat /home/developer/.claude/settings.json
  "command": "/usr/bin/python3 /opt/trinity/hooks/bash-guardrail.py"
  "command": "/usr/bin/python3 /opt/trinity/hooks/file-guardrail.py"
  "command": "/usr/bin/python3 /opt/trinity/hooks/read-only-guard.py"
  "command": "/usr/bin/python3 /opt/trinity/hooks/output-scanner.py"
```

### 2. The bricking mechanism, measured

Exact registered command, run with `/opt/trinity` absent vs present:

```
absent  → /usr/bin/python3: can't open file '/opt/trinity/hooks/bash-guardrail.py': [Errno 2]
          EXIT=2      # PreToolUse exit 2 = "block this tool call"
present → EXIT=0
```

### 3. The leak reproduces on the real image

Live container, source-mode shape (git repo root == `/home/developer`), staged exactly as the in-container auto-sync does (`git add -A`):

```
.claude/.last-cleanup
.claude/backups/session-2026-08-01.json
.claude/commands/deploy.md              <- legitimate
.claude/policy-limits.json
.claude/remote-settings.json
.claude/settings.json                   <- the one that bricks
.claude/skills/research/SKILL.md        <- legitimate
```

### 4. After the merge — leak stopped, source untouched

```
$ docker exec $C bash -c "<_build_gitignore_merge_command output>"     # exit 0
$ docker exec $C bash -c 'git add -A && git diff --cached --name-only'
.claude/commands/deploy.md
.claude/skills/research/SKILL.md

leaking artifacts still staged? (none — fix holds)
```

### 5. An already-bricked repo heals

Poisoned files force-committed first (the state real agents are in today), then both migration steps:

```
tracked BEFORE : .claude/.last-cleanup, backups/session-2026-08-01.json, commands/deploy.md,
                 policy-limits.json, remote-settings.json, settings.json, skills/research/SKILL.md
tracked AFTER  : .claude/commands/deploy.md, .claude/skills/research/SKILL.md

working tree   : -rw-r--r-- developer developer 715  .claude/settings.json
                 hooks still registered: 2 PreToolUse matchers
```

Index-only, as required — the **running** agent keeps its guardrail hooks while the remote loses the file.

### 6. The payoff — cloned onto a machine with no `/opt/trinity`

Both commits cloned out and opened in a container with `/opt/trinity` masked (an operator's laptop):

| clone | `.claude/settings.json` | hook exit | outcome |
|---|---|---|---|
| pre-fix commit | PRESENT | **2** | every Bash/Edit/Write refused — **bricked** |
| post-migration commit | ABSENT | — | **runs normally** |

`.claude/` contents in each clone: `backups commands policy-limits.json remote-settings.json settings.json skills` → `commands skills`. The agent's actual source survives; only the container-only runtime config is gone.

### 7. Idempotency + the escape hatch, on the image

The migration runs on **every** Push, so a second pass must be inert:

```
.gitignore md5 before : b0a627d97d8346aebd1f26e49c8282af
.gitignore md5 after  : b0a627d97d8346aebd1f26e49c8282af
51 lines / 51 unique          # no duplicate appends
working tree dirty?   : []    # clean — no `M .gitignore` drift (the #953 failure mode)
```

And the documented opt-out is real, not theoretical:

```
$ echo '!.claude/settings.json' >> .gitignore && git add -A
.claude/settings.json          # negation honoured
```

Test container removed afterwards; nothing left running.

vybe (approved):
P1, and the blast radius justifies it — this leak *bricks* foreign clones rather than merely bloating them, since a PreToolUse hook with a missing script exits 2 and exit 2 is Claude Code's block-this-call signal.

Two things make this more than a pattern-list addition:

The heal path is correct and load-bearing. `_migrate_workspace_gitignore` does an index-only `git rm --cached`, so already-poisoned agents drop the file from the remote on their next Push while the running container keeps its working-tree copy and its hook registration. Without that property the fix would break every live agent it reached.

The tests drive real `git` against a temp repo rather than asserting on the pattern list — which matters here specifically, because `.claude/.last-cleanup` is a dotfile inside a dotdir and `.claude/backups/` is a directory rule, and a list-only assertion would pass against a rule git does not actually apply. The premise tests (baked file really carries absolute `/opt/trinity` paths, Dockerfile still writes it inside the repo root) are the right instinct — they force a re-decision instead of leaving a dead rule behind if the hook registration ever moves.

Trade-off is stated rather than buried, and the `!.claude/settings.json` escape hatch has its own test. The `# Keep:` comment correction is in scope — it sat directly beside the new rule and `*.local.json` already covered it.

Body patched to `Fixes #2036` for status auto-promotion. Noting the two deferrals you flagged (auto-sync never refreshes `.gitignore` per #953; hook registration outside the synced tree) are correctly out of scope for a P1 fix.

Issue #668, opened 2026-05-05
feat: Agent deployment compatibility validation — server-side checks with auto-fix offers


vybe:
## Summary

When users deploy agents to Trinity that weren't created following Trinity best practices (no playbooks, missing YAML files, `.claude/` excluded from `.gitignore`, no `template.yaml`), the agent may silently fail at runtime in ways that are hard to diagnose. Trinity should run server-side compatibility checks on deployment and surface actionable recommendations — without blocking the deployment.

## Context

Raised in the 2026-05-05 dev catch-up after client experience with Paradigm Life: agents deployed without playbooks or required YAML files diverged from their local behavior on Trinity. Andrii proposed server-side checks; Eugene framed this as hard requirements (e.g., gitignore) vs soft recommendations (e.g., dashboard.yaml). The team agreed on a non-enforcing, recommendation-based approach with optional auto-fix for correctable issues.

## Validation Spec

The full check list (**100 checks, 11 categories**) is defined in **[`docs/agent-validation-spec.md`](https://github.com/abilityai/trinity/blob/dev/docs/agent-validation-spec.md)** — the canonical source of truth for what gets validated and how.

> **Spec updated since this issue was opened (2026-05-08):** added file-structure checks **F-011/F-012/F-013** (ARCHITECTURE.md / requirements.md / CHANGELOG.md) and a whole new category **I — Composability (I-001…I-005)** plus the data-exchange principle. Counts below reflect the current spec, not the original 56/10 framing.

Check categories:
- **F** — File & directory structure (13 checks)
- **S** — Security / secret hygiene (10 checks)
- **T** — `template.yaml` schema & content quality (17 checks)
- **C** — `CLAUDE.md` quality (12 checks)
- **K** — Credential configuration consistency (5 checks)
- **G** — Git configuration hygiene (5 checks)
- **P** — Skill & playbook construction (12 checks)
- **A** — Autonomy & scheduling design (5 checks)
- **D** — Dashboard & metrics configuration (8 checks)
- **X** — Cross-file consistency (8 checks)
- **I** — Composability / data-exchange contract (5 checks)

Checks marked `[AI]` are evaluated by an LLM reading file contents (e.g., "does this CLAUDE.md contain domain-specific instructions?", "does this autonomous skill contain approval gates?"). Checks marked `[STATIC]` use deterministic file/pattern analysis.

## Acceptance Criteria

[...]

PR #2040, opened 2026-08-06
feat: GitHub-repo import wizard — fork / copy / clone with inline compat check (abilityai/trinity-enterprise#15)

trinity-ability:
## Summary
- `import_intent: fork | copy | clone` on agent creation (`POST /api/agents`, MCP `create_agent` with `copy|clone` only, UI intent picker on the free-form GitHub path). Absent intent → legacy behavior exactly.
- **Copy = backend-materialized snapshot**: staging clone via the fork-to-own git machinery (`/data` staging, PAT via `GIT_CONFIG_*` never argv, `@branch` honored, empty repo → 400 `COPY_SOURCE_EMPTY`, tree-escaping symlinks pruned) → strip `.git` → workspace-volume pre-population via the deploy-local primitive **before the container exists**. The container carries no GitHub env and no PAT, no `agent_git_config` row (both rebuild seams pop git env by construction; never sync-polled), zero base-image dependency.
- **Inline #668 compatibility check** as the modal's post-create step — readiness-gated on a real agent-server `/info` response (the proxy fail-opens to 200 + fallback body mid-clone; gated by body shape) with honest failed-to-start/timeout branches. Non-blocking.
- **Idempotency** (Invariant #18): `Idempotency-Key` on create, scope `agent_create:{user_id}` (two-principal isolation tested), named in-flight 409 `CREATE_IN_FLIGHT`; MCP derives a deterministic name-keyed header.
- Provenance: `AgentStatus.import_snapshot` (repo/branch/SHA/file-count) → create audit details + `trinity.import-intent` label.
- Fork scope failures (`FORK_PAT_INVALID`, `FORK_REPO_CREATE_FAILED`) name the copy/clone alternatives (issue AC "degrade …and say so" — implemented as a named suggestion, never silent auto-degrade).

## Review notes
Pre-landing `/review` findings all resolved: readiness-gate fail-open fixed (body-shape discriminator; old images degrade to the honest timeout state), copy+ephemeral refused by name (`COPY_EPHEMERAL_UNSUPPORTED` — ghosts are volume-less, ent#69), stranded-volume retry-409 fixed (rollback handle armed before pre-population). **Accepted residual**: a seconds-wide same-name concurrent-create interleaving (pre-existing check-then-act creation class) is documented in the feature flow, not locked — a per-name creation lock is the future fix if it ever bites.

## Changes
Backend: `snapshot_import.py` (new), `crud.py` (intent gates + copy branch + rollback), `models.py`, `routers/agents.py` (idempotency + audit), `fork_to_own.py` (messages). MCP: `agents.ts`, `client.ts`, `chat.ts` (export), `types.ts`, `import-intent.test.ts` (new). Frontend: `CreateAgentModal.vue`, `ImportIntentPicker.vue` (new), `ImportValidationStep.vue` (new). Docs: requirements §11.13, `feature-flows/github-import-intents.md` (new) + index, architecture row, learnings entry. Tests: `tests/unit/test_ent15_import_intents.py` (23 cases) + registry.

## Test Plan
- [x] `pytest unit/test_ent15_import_intents.py` — 23/23
- [x] Adjacent creation-path suites (`test_1484_create_agent_characterization`, `test_ent313_*`, `test_fork_to_own`, `test_ent123_tokenless_clone`) — 173/173
- [x] MCP `npm test` 173/173 + `tsc --noEmit` clean
- [x] Frontend `npm run build` + `check:tokens` pass; raw-color ratchet flat (CreateAgentModal byte-equal to baseline; new components zero non-gray/hardcoded)
- [ ] Manual: create from a public repo with each intent; verify copy agent has no `GITHUB_REPO` env and boots with files; wizard compat step on a real clone

Related to abilityai/trinity-enterprise#15 (private-tracker feature — closed manually at release; closing keywords do not cross repos)


vybe (changes requested):
The design is careful and the seam analysis is genuinely good — the git-env story (the `_apply_git_env_from_db` no-row branch), the #1667 gate ordering, and the rollback-after-container-reclaim ordering all check out against the real code. The security review came back clean on the things I most expected to find problems in: the PAT goes through `GIT_CONFIG_*` / `http.extraHeader` and never argv, failure output is `scrub_secret`-ed, and it never reaches container env, `agent_git_config`, or the response; the URL is a hardcoded `https://github.com/` host with `repo_path` already validated against `^[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$`, so no scheme override, userinfo, traversal, or leading-`-` argv injection; `_prune_escaping_symlinks` correctly walks with `followlinks=False`, resolves non-strictly so dangling links are caught, and `tarfile.add` archives symlinks as `SYMTYPE` rather than following them.

What's missing is a resource bound on a path that streams attacker-chosen bytes onto the platform's primary data volume and through backend RAM.

**1. Unbounded clone into `/data`.**

`snapshot_import.py` stages into `/data/agent-import-tmp` with `git clone --depth 1 --single-branch` bounded only by `CLONE_TIMEOUT_S = 120`. `/data` is the bind mount that also holds `trinity.db` and `agent-files/`. Any `creator`-role caller can point `copy` at an arbitrarily large public repo, and 120 seconds of clone bandwidth is multiple GB on a decent link. There is no rate limit on `POST /api/agents`. Filling `/data` takes SQLite down.

Compare #1595, where git maintenance carries an explicit free-disk preflight (skip below 1.1× pack bytes) for exactly this reason. Needs a size cap enforced during or after clone, a free-space preflight, or both — with a named 4xx rather than a disk-full 500.

**2. The whole snapshot is materialized in backend memory as one tar.**

`deploy._prepopulate_workspace_from_template` builds a `BytesIO`, `tar.add`s the tree, then reads the entire buffer before `put_archive`. That primitive was written for #950 deploy-local, whose input is an **operator-supplied** template of bounded size. This PR repoints it at arbitrary user-specified GitHub content with no size validation in between, so a large repo OOMs the backend process — which also serves every other agent. `snapshot_import` computes `file_count` but never total bytes, so nothing observes the magnitude before the tar is built.

**3. The MCP auto-derived idempotency key makes delete-then-recreate a phantom success.**

`tools/agents.ts` derives a key over `[userId, "create_agent", args.name, JSON.stringify(config)]` — deterministic over name + config, 24h TTL. Delete an agent and re-create it identically via MCP inside 24h and `idempotency_service.begin` returns the stored `completed` snapshot: the caller gets a 200 with the old `AgentStatus` and `X-Idempotent-Replay: true`, and **no agent exists**.

The router docstring explicitly argues *"Header-only — no auto-derived key"*; the MCP surface then does exactly what that sentence rules out. Agent Self-Reminders hit this and solved it ("terminal stored rows excluded from replay"). Either drop the MCP auto-derivation or exclude replays whose recorded agent no longer exists.

**4. The core happy path was never executed.**

The Test Plan's `Manual: create from a public repo with each intent; verify copy agent has no GITHUB_REPO env and boots with files` is unchecked, and every backend test mocks `_run_git` — nothing clones a real repo, pre-populates a real volume, or boots a container. The feature's central safety claim, that a copy agent carries no GitHub env and no `agent_git_config` row across both rebuild seams, is asserted in prose and in the feature flow and verified nowhere. For a feature PR under the project's verification rule, that's the hold.

**Non-blocking:**

- `_prepopulate_workspace_from_template` is called without `adopt_existing_workspace`, so the volume mount takes its "declared adopt" branch and logs an adopt attributed to deploy-local #950. Functionally correct — I confirmed the #1667 refusal gate runs before the volume exists — but the log now misattributes the caller.
- The 24h staging-dir sweeper is the wrong lever while finding 1 is unfixed.
- `/data/agent-import-tmp` isn't added to the architecture "Data Persistence" section, which enumerates `/data` contents.
- `jsonable_encoder`, `JSONResponse` and `idempotency_service` are imported inside the endpoint; every other router imports at module level.
- Test gaps per new surface: nothing for `_cleanup_copy_artifacts` (the F4 ordering fix ships with zero regression coverage), the `trinity.import-intent` label, the `import_snapshot` audit fold, or `idempotency_service.fail()` releasing the claim so a corrected retry proceeds.
- Invariant #1: `snapshot_import` raises `HTTPException` directly rather than a domain error mapped at the router, unlike the ent#109 `BindError` precedent the same file cites as its model. Minor, not worth blocking on.

Docs are the best-documented part of this PR — architecture row, feature flow + index, requirements, learnings entry.

**Entitlement ruling needed.** This lands fully **ungated** — zero `requires_entitlement` / `enterprise_features` references anywhere in the diff — and ent#15 is an enterprise-tracker feature under epic ent#122. Per CLAUDE.md the default is a gated module unless explicitly ruled OSS-core, and "reuses OSS tables and endpoints" is expressly *not* a reason to infer OSS. This is a full user-facing wizard, so the question is bigger here than on a single read endpoint. Needs @vybe's explicit call before merge.

Also needs a rebase (`learnings.md` only).

vybe:
All four blocking findings addressed in `a3e8fb2f` (+ dev merged in `90908b59`), and — the F4 hold — **the happy path has now been executed against a live stack**, end to end.

## Entitlement ruling

**Eugene's call: OSS-core, ungated.** The import wizard ships in the public bundle with no `requires_entitlement` / `enterprise_features` gate.

## 1. Clone bounded (F1)

- **Free-space preflight** before any git subprocess: refuses without `3 × cap` headroom on `/data` → structured **507 `COPY_STAGING_NO_SPACE`** (clone never starts — pinned by a test asserting zero git calls).
- **Post-clone size cap** `AGENT_IMPORT_MAX_BYTES` (default 1 GiB, env-tunable), measured via `lstat` (a symlink never counts its target — tested) → named **400 `COPY_SOURCE_TOO_LARGE`**, staging released.
- `CLONE_TIMEOUT_S` + preflight bound the clone itself; the #1595 free-disk-preflight precedent is cited in the module.

## 2. Tar spooled to disk (F2)

`_prepopulate_workspace_from_template` now writes the tar to a `TemporaryFile` beside the template tree (disk-backed `/data` in both callers — never inside `template_dir`, never the RAM-tmpfs default tmp) and hands the file object to `put_archive`; requests streams it, so neither the build nor the upload holds the full tar in memory. Comment explains why (this primitive's original input was operator-bounded #950).

## 3. Replay liveness-gated (F3)

Kept the MCP derivation (it's the transport-retry protection) and fixed the server side, per your "exclude replays whose recorded agent no longer exists": the endpoint re-checks `db.is_agent_live(recorded_name)` before replaying. Dead → discard via the new `db.idempotency_discard_completed` (deletes **only** completed rows — `release` still can't touch them) and fall through to a genuinely fresh create, which answers honestly: name-reserved 409 for a soft-deleted agent, a real create for a hard-purged one. Re-claim race → 409 `CREATE_IN_FLIGHT`. Router docstring updated; the Self-Reminders precedent shape.

## 4. Happy path EXECUTED (live local stack, backend hot-swapped to this branch)

**Copy intent** — `POST /api/agents {"name":"e2e-copy-test","template":"github:Abilityai/cornelius","import_intent":"copy"}`:
```
"import_snapshot": {"source_repo": "Abilityai/cornelius", "source_branch": "main",
                    "head_sha": "14927a86290ae023b609d69df39b9d5c65f02d38", "file_count": 1232}
```
- container env: `grep -cE "GITHUB|GIT_SYNC"` → **0**
- `trinity.import-intent=copy` label present
- `db.get_git_config('e2e-copy-test')` → **None**
- started → `status: running`, `/home/developer` holds the full tree (15 top-level entries), `.trinity-initialized` present, **no `.git`**

**F3 live proof** — create with `Idempotency-Key: k-live-2040` → 200; immediate retry → `x-idempotent-replay: true`; **DELETE the agent; re-POST the same key** → `HTTP 409 "Agent already exists"`, **no replay header** — pre-fix this was a 200 replay naming a deleted agent.

Fixtures cleaned up afterward; stack restored to dev.

## Non-blocking items taken

- `jsonable_encoder`/`JSONResponse`/`idempotency_service` hoisted to module level.
- `/data/agent-import-tmp` added to architecture.md § Data Persistence.
- `_cleanup_copy_artifacts` regression tests (staging+volume removal, missing-volume tolerance).
- Feature flow documents the bounds + the replay gate.

Left as-is, per your "minor": the Invariant #1 `HTTPException`-in-service shape, and the adopt-log misattribution. The 24h sweeper stays (now as the backstop behind real bounds, which was your framing).

Tests: `test_ent15_import_intents.py` → **31 passed** (8 new). Rebase done (learnings.md + feature-flows.md kept both sides).

vybe (approved):
/validate-pr passed — approving to supersede my earlier CHANGES_REQUESTED (all four blocking findings addressed in a3e8fb2f, live happy-path executed, hold lifted per review thread).

**Validation summary:**
- Security: no keys/emails/IPs/secrets/credential files in the diff ✅
- Docs: requirements/github.md §11.13 + architecture.md + new feature flow + index + learnings ✅
- Packaging: no new top-level backend modules; snapshot_import.py rides the services/ wholesale COPY ✅
- Tests: 31 unit cases incl. non-happy paths (preflight refusal w/ zero git calls, size cap via lstat, replay liveness, two-principal idempotency isolation) + MCP import-intent.test.ts, registry updated ✅
- Cross-tracker ref (Related to abilityai/trinity-enterprise#15) is the expected shape — manual status-in-dev bump + manual close at release.

⚠️ Non-blocking follow-up: `AGENT_IMPORT_MAX_BYTES` is read via env but not wired into either compose file / .env.example — same (pre-existing) gap as its sibling `AGENT_DATA_EXPORT_MAX_BYTES`. Default (1 GiB) is active regardless, so the bound holds on deploy; the override knob is inert until wired. Suggest one follow-up wiring both.

PR #2041, opened 2026-08-06
fix(skills): scrub the whole credential run from a double-@ URL (ent#347)

dolho:
Related to abilityai/trinity-enterprise#347

## What

Both free-text PAT scrubbers used a character class that stops at the **first** `@`, so neither stripped a credential from a double-`@` URL — and that is the shape the skills clone path constructs.

`_authenticated_url` splices the platform PAT in front of whatever userinfo the stored URL already carries (verified at the source: `netloc = f"{github_pat}@{parsed.netloc}"`):

```
stored:   https://LEGACYTOK@github.com/org/private-skills
spliced:  https://PLATFORMPAT@LEGACYTOK@github.com/org/private-skills
```

| scrubber | old pattern | old result |
|---|---|---|
| `skill_service._scrub_pat` | `https://[^@\s]+@` | `https://***@LEGACYTOK@github.com/...` |
| `skill_source_clone.redact` | `https://[^@]+@` | `https://***@LEGACYTOK@github.com/...` |

The surviving token reached `system_settings['skills_library_last_error']` — durable, and rendered on the admin Settings panel — and an HTTP `detail` body.

This is the **guaranteed** branch, not a rare one: git resolves the username as `PLATFORMPAT@LEGACYTOK`, auth is rejected, and the error path is the one that persists the scrubbed string. A stored URL can carry userinfo today because `_adopt_legacy_clone` writes with no validation and `validate_skills_library_url` ignores userinfo by design.

## The fix

One pattern, `https://[^\s/?#]+@`, defined once as `_CREDENTIAL_URL_RE` in `skill_source_clone`; `_scrub_pat` delegates to `redact` and keeps only its never-raises wrapper.

**The single definition is part of the fix, not tidying.** The root cause was two hand-written patterns that had drifted apart and were each wrong in a different way. Leaving two copies leaves the mechanism that produced the bug.

Greedy and unable to cross `/`, `?`, `#` or whitespace, so the **last `@` before the path** wins:

| input | output |
|---|---|
| `https://a@b@github.com/o/r` | `https://***@github.com/o/r` |
| `https://github.com/o/r?ref=a@b` | untouched |
| `https://github.com?x=a@b` | untouched |
| plain URL, then a credentialed one | only the credential goes |

### One deviation from the issue's suggested pattern

The issue proposes `https://[^\s/]+@`. I also excluded `?` and `#`. Neither is legal unencoded in userinfo (RFC 3986), so excluding them can only ever *prevent* a false positive — and the issue's pattern mangles `https://github.com?x=a@b` (no path, so nothing stops the class before the `@`) into `https://***@b`. All five acceptance criteria still hold; this covers one more.

The issue's second observation is fixed by the same change: `redact`'s old `[^@]+` could cross `/` **and** newlines, so an earlier plain URL matched all the way into a later credentialed one — mangling the message while hiding the token only by accident. `\s` in the class stops that.

These stay text-oriented, as the issue specifies: git stderr carries several URLs plus prose, so the single-URL parser cannot drop in. The comment on `_CREDENTIAL_URL_RE` records that it must nonetheless *agree* with that parser on where the authority ends.

## Coordination note

The issue describes `strip_url_credentials` in `utils/url_validation.py` as already present from ent#334. **It is not on `dev`** — ent#334 is still `status-in-progress` with no PR. So this fix deliberately takes no dependency on it. When ent#334 lands, its parser and `_CREDENTIAL_URL_RE` need to agree on the authority boundary; a cross-check test between them belongs on whichever lands second.

## Tests

`tests/unit/test_ent347_pat_scrubber_double_at.py` — 22 tests driving the **real** `redact` / `_scrub_pat`:

- all five acceptance criteria, plus triple-`@` (nothing caps userinfo at two segments) and multiline stderr
- `@` in a path segment and in a query are left alone — scrubbing them corrupts the message an operator is reading
- the leak assertions are phrased as an **absence** (`"ghp_LEGACY" not in out`), not only an equality: an equality can be satisfied by a pattern that is right on one input and leaks on its neighbour
- **both scrubbers agree**, parametrized across six inputs, so they cannot drift apart again
- totality: neither raises on `""`, `None`, `"https://"`, `"@@@"`

And the end-to-end case, which is the one tying the fix to the reported path: its input is built by the **real `_authenticated_url`**, not a hand-typed double-`@` string, and it asserts the splice really is double-`@` before scrubbing. If the splice ever changes shape, that test fails rather than the scrubbers quietly ceasing to cover it.

```
tests/unit/test_ent347_pat_scrubber_double_at.py     22 passed
+ ent237 / ent332 / user_agent / ssrf / ent183 ×2    265 passed
lint_sys_modules.py                                  no new violations
```

**Mutation check.** Reverting both service files fails **11 of 22**.

**Pre-existing, not mine:** running `test_ent183_skill_packages.py` before `test_ent236_skills_lifecycle.py` collection-errors with `ImportError: cannot import name 'get_skills_auto_sync_interval'` — a `sys.modules` stubbing artifact. Reproduced identically on clean `dev` with this branch stashed; `test_ent236` passes standalone (64).

## Scope

Two functions, one shared pattern, one new test file. No schema change, no migration, no endpoint, no config. Filed on the enterprise tracker but the code is OSS-core, so it lands here per the tracker≠repo rule.

### DIFF
--- a/src/backend/services/skill_service.py
+++ b/src/backend/services/skill_service.py
@@ -42,6 +42,7 @@ from services.skill_source_clone import (
     QUARANTINE_SUFFIX,
     SOURCE_ID_RE,
     SkillSourceClone,
+    redact,
 )
 from urllib.parse import urlparse, urlunparse
 
@@ -112,9 +113,21 @@ _SYNC_LOCK_TTL_SECONDS = 300
 
 
 def _scrub_pat(text: str) -> str:
-    """Replace credentials in any `https://<creds>@host` URL. Never raises."""
+    """Replace credentials in any `https://<creds>@host` URL. Never raises.
+
+    Delegates to `skill_source_clone.redact` (ent#347). These were two separate
+    hand-written patterns that had drifted apart and were each wrong in a
+    different way — the one duplication that matters, since the value it guards
+    lands in DURABLE admin-rendered state (`skills_library_last_error`) and in
+    an HTTP `detail`. One pattern, one place; see `_CREDENTIAL_URL_RE`.
+
+    The try/except stays: this wrapper's own contract is "never raises", and it
+    is called on a `str(e)` from an arbitrary exception whose `__str__` can
+    itself blow up. `redact` is total for `str`/`None`, so this is belt to its
+    suspenders rather than a live branch.
+    """
     try:
-        return re.sub(r"https://[^@\s]+@", "https://***@", text)
+        return redact(text)
     except Exception:  # noqa: BLE001
         return ""
 
--- a/src/backend/services/skill_source_clone.py
+++ b/src/backend/services/skill_source_clone.py
@@ -122,9 +122,44 @@ def validate_declared_root(value: Any) -> Optional[str]:
     return root
 
 
+# ent#347: THE credential pattern for free-text (git stderr / exception) scrubs.
+# One definition on purpose — the bug this fixes was two scrubbers carrying two
+# hand-written patterns that had drifted apart and were each wrong differently.
+#
+# `[^\s/?#]+` is greedy and cannot cross a path separator, a query/fragment
+# delimiter, or whitespace, which is what makes it correct on all four shapes:
+#
+#   * `https://PAT@LEGACY@github.com/o/r` — `_authenticated_url` splices the
+#     platform PAT in FRONT of userinfo the stored URL already carried, so a
+#     double-`@` authority is the normal shape here, not a curiosity. The class
+#     spans `PAT@LEGACY@github.com`, then backtracks to the LAST `@` before the
+#     path, so the whole credential run is replaced. The previous `[^@...]+`
+#     classes stopped dead at the FIRST `@` and left the second token verbatim.
+#   * `https://github.com/o/r?ref=a@b` — the class stops at `/`, so no `@`
+#     follows and the URL is untouched. An `@` in a path or query is not a
+#     credential.
+#   * `https://github.com?x=a@b` — stops at `?` for the same reason. This is why
+#     `?#` are excluded on top of `/`: neither is legal unencoded in userinfo
+#     (RFC 3986), so excluding them can only ever prevent a false positive.
+#   * a plain URL followed later in the same stderr by a credentialed one —
+#     `\s` in the class stops the match at the first whitespace/newline, so the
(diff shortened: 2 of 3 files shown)


vybe (approved):
Verified the new pattern independently against all the shapes in the description — `https://[^\s/?#]+@` is greedy and cannot cross `/`, `?`, `#` or whitespace, so the last `@` before the path wins:

```
https://a@b@github.com/o/r        -> https://***@github.com/o/r
https://ghp_A@ghp_B@github.com/.. -> https://***@github.com/o/r
https://a@b@c@github.com/o/r      -> https://***@github.com/o/r   (triple)
https://github.com/o/r?ref=a@b    -> unchanged
https://github.com?x=a@b          -> unchanged
'@@@' / ''                        -> unchanged
```

The deviation from the issue's suggested `[^\s/]+@` is a genuine improvement, not a preference: neither `?` nor `#` is legal unencoded in userinfo per RFC 3986, so excluding them can only prevent a false positive — and the issue's pattern really does mangle `https://github.com?x=a@b` into `https://***@b`, since with no path nothing stops the class before the `@`.

Collapsing to a single `_CREDENTIAL_URL_RE` is part of the fix rather than tidying, and the PR says so for the right reason: the root cause was two hand-written patterns that had drifted apart and were each wrong in a *different* way, so leaving two copies leaves the mechanism. Keeping `_scrub_pat`'s try/except as belt to `redact`'s suspenders is correct — it takes `str(e)` from an arbitrary exception whose `__str__` can itself raise.

Two test choices carry real weight: building the end-to-end input from the **real `_authenticated_url`** rather than a hand-typed double-`@` string (so the suite fails rather than silently ceasing to cover the path if the splice changes shape), and phrasing leak assertions as an **absence** rather than only an equality (an equality can be satisfied by a pattern that is right on one input and leaks on its neighbour).

The severity framing is right too — this is the *guaranteed* branch, not a rare one, because the doubled userinfo makes git reject auth and the error path is the one that persists to `skills_library_last_error`, which is durable and admin-rendered.

Good call taking no dependency on ent#334's `strip_url_credentials`, which is genuinely not on `dev`. The pre-existing `sys.modules` flake you flagged was #1898 — fixed by #2031, merged just ahead of this.

Cross-tracker ref, so ent#347 needs `status-in-dev` by hand — doing that now.

Issue #259, opened 2026-04-03
feat(cli): trinity deploy . — file-based agent deployment with tracking, limits, and MCP provisioning


vybe:
## Summary

Implement the simplest possible onboarding flow for deploying an agent to a hosted Trinity instance:

```bash
pip install trinity-cli
trinity init
trinity deploy .
```

No GitHub knowledge required. No manual MCP configuration. Three commands from zero to deployed agent.

### 1. `trinity deploy .` CLI command

**File**: `src/cli/trinity_cli/commands/deploy.py` (new)

- Tar + base64 encode local directory (respecting `.gitignore`, excluding `.git/`, `node_modules/`, `.env`)
- POST to `/api/agents/deploy-local` (endpoint already exists)
- On first deploy: create agent, write `.trinity-remote.yaml` tracking file
- On subsequent deploys: detect `.trinity-remote.yaml`, redeploy to same instance/agent (idempotent update)
- Add `.trinity-remote.yaml` to `.gitignore` automatically if git repo detected
- Support `--name` flag to override agent name (default: directory name)
- Support `--repo user/repo` flag to use an existing public GitHub repo instead of file-based deploy

**Tracking file format** (`.trinity-remote.yaml`):
```yaml
# Auto-generated by trinity deploy — do not edit
instance: https://trinity.example.com
agent: my-agent
profile: production
deployed_at: 2026-04-03T15:45:00Z
```

### 2. Agent limit enforcement (3 per user default)

**Files**: `src/backend/services/agent_service/crud.py`, `src/backend/services/agent_service/deploy.py`

- Add per-user agent count check in `create_agent_internal()` (after duplicate name check, ~line 89)
- Add same check in `deploy_local_agent_logic()` (before calling `create_agent_fn()`)
- Use existing `get_agents_by_owner()` for count
- Default limit: 3 agents per user
- Make configurable via settings (e.g., `MAX_AGENTS_PER_USER` in settings table)
- Return HTTP 429 with clear message: "Agent quota exceeded. Maximum 3 agents per user. Delete an agent to create a new one."
- No database schema changes needed — `agent_ownership` table already tracks this

### 3. MCP key auto-provisioning in `trinity init`

**File**: `src/cli/trinity_cli/commands/auth.py`

After successful login in `trinity init` / `trinity login`:
- Call `POST /api/mcp/keys/ensure-default` (endpoint already exists)
- Store MCP API key in the profile within `~/.trinity/config.json`
[...]

Issue #95, opened 2026-03-11
Route async task mode through TaskExecutionService


vybe:
## Problem

The async mode path in `chat.py` (`_execute_task_background()`, lines 376-567) does **NOT** use `TaskExecutionService`. It implements its own parallel logic for slot management, activity tracking, error handling, and credential sanitization — inline.

```
/api/task (async=true)  → _execute_task_background()    ← INLINE, NOT TaskExecutionService
/api/task (sync)        → TaskExecutionService.execute_task()  ← Unified path
```

### Impact

- Any fix applied to `TaskExecutionService` (status values, sanitization, slot logic) does NOT apply to async mode
- Async mode has its own slot acquire/release that could diverge
- Async mode has its own error handling with potentially different status values
- Async mode may miss credential sanitization steps added to the service
- Two maintenance surfaces for the same logical operation

### Current Async Flow (chat.py)

1. Create execution record early (`db.create_task_execution()`)
2. Acquire slot inline (`slot_service.acquire_slot()`)
3. Track activity inline (`activity_service.track_activity()`)
4. Spawn `asyncio.create_task(_execute_task_background())`
5. Background function: call agent, update records, release slot in finally

### Required Changes

1. Add async support to `TaskExecutionService` (e.g., `execute_task_async()` that spawns background and returns immediately)
2. Move all inline logic from `_execute_task_background()` into the service
3. Chat router's async path should only create execution record + call service
4. Ensure identical slot, activity, sanitization, and error handling between sync and async

### Files to Change

- `src/backend/services/task_execution_service.py` — Add async execution mode
- `src/backend/routers/chat.py` — Remove `_execute_task_background()`, delegate to service
- `src/backend/routers/public.py` — Also has `_execute_public_chat_background()` that should use service

PR #2042, opened 2026-08-06
feat(grid): widget chassis — InfoTile, GRID_WIDGETS registry, layout v2, Tiles menu (ent#325)

dolho:
Fixes abilityai/trinity-enterprise#325 · foundation for epic ent#94

> **Rebased onto `dev` and retargeted** (was stacked on #1918 / `feature/305-grid-org-overlay`, which squash-merged so its tip never became a `dev` ancestor). `git rebase --onto origin/dev origin/feature/305-grid-org-overlay` applied cleanly with zero conflicts; the diff below is this issue's work only, and the required checks now run against it.

## What

The widget chassis every tile sub-issue on ent#94 (#95–#101, #259) is blocked on. Until now none of them was claimable, because whoever picked one first had to build the whole chassis off-issue.

- `InfoTile.vue` — the tile shell
- `GRID_WIDGETS` registry + `widget:*` key namespace
- layout v2 with a one-time v1 migration
- the "Tiles ▾" show/hide menu
- one reference tile, so the chassis is demonstrable

## Info tiles are occupants of the same lattice, not a parallel system

Widgets join the **same** layout map under `widget:*` keys. `:` cannot appear in an agent name (`sanitize_agent_name` strips everything outside `[A-Za-z0-9_-]`), so collision is impossible and `isWidgetKey` is exact rather than a heuristic.

The payoff is that drag, swap-with-preview, keyboard reorder, viewport culling and the snap socket all work for tiles through the existing `data-agent` path — no second code path to keep in sync as those behaviours evolve. Keyboard came free: `@keydown` is on the world container and resolves through `layout[name]`.

Deliberately **absent** from `InfoTile`: Run/Autonomy toggles (info tiles summarize, they don't operate — a fleet-scope operating control makes a misclick fleet-wide), and any org affordance. No connect port, so a tile cannot be a reporting-line endpoint; that is enforced by omitting the hover handlers `AgentTile` uses, not by a runtime check.

## Org-overlay interlock — the load-bearing part

**A. Tidy and Group-by-dept would have silently deleted every tile position.** Both `tidyByDept` and `arrangeByDept` build a fresh map keyed only from `agents`, and `applyLayout` replaced the layout wholesale. Fixed at **both** ends as the issue asks: the helpers carry `widget:*` keys through (seeded first, so department blocks flow around them via `nearestFreeCell`), **and** `applyLayout` merges instead of blind-replacing. Either alone is a single point of failure, and one-guard-only is how this bug existed in the first place.

Adding `nearestFreeCell` to `arrangeByDept` is behaviour-preserving when no widgets are present — every requested cell is free, so placement is unchanged. There is a test pinning that equivalence, because it is the entire safety argument for touching that function.

**B. Widgets stay out of zone hulls — left alone deliberately.** `computeZones` iterates `agents`, so a `widget:*` key is excluded by construction. Not "fixed" by iterating the layout map; pinned by a test asserting a tile parked beside a department does not stretch its frame.

**C. Gap budget.** The peg overhangs 16px, inside `GAP_X` (40), so it cannot collide with `ZONE_CHROME`.

**D. Seeding consults `zoneAt`** and skips cells inside a department frame — rows −3…−1 are empty of *tiles* but not necessarily of *hulls*.

## Layout v2

`trinity-grid-layout-v2`, migrated by a one-time **copy** from v1 — v1 is left in place, so downgrading is not a data-loss event.

`normalizeLayout` gains a `widgetKeys` argument. Without it every `widget:*` entry was dropped on the next reconcile; that is pinned as a test so the regression stays visible rather than becoming folklore. Agents reconcile **first**, so a collision evicts the tile — an info tile moving one cell is a smaller surprise than a constellation the user arranged shifting under them. The `changed` check now covers widgets too, or every reconcile would re-persist localStorage forever.

Default placement is the band **above** the fleet, which is what makes "default-on tiles appear without moving any saved agent position" true by construction: `defaultLayout` starts agents at `r=0` and grows downward, so negative rows are unclaimed on every existing install.

## Tiles menu

Per-user, persisted under its own key so it can never clobber the overlay's Zones/Lines prefs. Preferences are a **sparse override map**, not an allow-list — so a tile added in a later release appears for existing users instead of staying invisible behind a stale list, a user can keep a default-on tile off, and "Reset" is just an empty map. Placed in the bottom control cluster rather than as a fifth top-level header button, per the issue's responsive-ladder note.

## Registry, and why the catalog lives elsewhere

`utils/gridWidgets.js` stays free of `.vue` imports so vitest's node environment can test it — and because `gridOrg.js` now imports `isWidgetKey` from it, a component import there would have broken #1918's existing `gridOrg.spec.js`. Catalog entries live in `components/tiles/catalog.js`, imported by `FleetGrid` for its side effect. Adding a tile is one `registerWidget({...})` block plus the component.

`cells` is declared and ignored: v1 renders single-cell only, multi-cell occupancy is #94's stated v2. Declaring it now means catalog entries don't need rewriting when `gridLayout.js` learns occupancy.

**One reference tile ships** (Fleet summary): it reads the `agents` array the grid already holds and issues no request, so the chassis carries no data dependency of its own and could not be blocked behind the sibling ent#326 endpoint. This is not scope creep — the epic's ACs require a default-on tile to render, so an empty catalog would leave half of them unverifiable. Real tiles arrive one sub-issue at a time.

## Tests

`tests/unit/gridWidgets.spec.js` — 28 tests: key namespace (incl. that no agent-shaped name reads as a widget), registry, **an `adminOnly` tile never enabled for a non-admin even when prefs say on** (prefs survive a role change; the catalog filter has to be what decides), override-map semantics incl. a newly-registered default-on tile reaching a user with existing prefs, seeding band + veto, normalize round-trip in **both** directions, the tidy/arrange preservation rule, and zone-hull exclusion. One guard asserts stray non-widget keys are *not* carried, so the carry-through can't drift into "preserve anything unknown" and resurrect a deleted agent forever.

```
tests/unit/gridWidgets.spec.js   28 passed   (new)
tests/unit/gridOrg.spec.js       21 passed   (#1918, unbroken)
tests/unit/gridLayout.spec.js     6 passed   (#1918, unbroken)
                                 55 passed
```

## Verification note, stated honestly

`npm run build` and `npm run test:unit` cannot run against this machine's `node_modules` — it predates #1918 (no `vitest`) and is missing `mermaid`/`qrcode`. Both were run out-of-band instead: vitest standalone with an equivalent inline config (the 55 above), and a production `vite build` with the stale deps externalized, which transformed **638 modules clean** — covering every file in this PR. CI runs both properly.

## Not in scope

Multi-cell tiles (#94 v2), agent-declared custom tiles (#94 v2, unfiled), and the actual data tiles (#95–#101, #259).

### DIFF
--- a/src/frontend/e2e/dashboard-type-filter.spec.js
+++ b/src/frontend/e2e/dashboard-type-filter.spec.js
@@ -66,7 +66,13 @@ test.describe('dashboard type-to-filter (trinity-enterprise#261)', () => {
     // and the grid canvas stays MOUNTED (transient zero-match while typing
     // must never unmount the pane — plan eng F1 pin).
     await pillInput(page).fill('zzz')
-    await expect(page.locator('.gv-tile')).toHaveCount(0)
+    // Agent tiles only. ent#325 puts info tiles on the same `.gv-tile` chassis
+    // (deliberately — drag/keyboard/culling/snap all reuse one code path), and
+    // they are NOT agents: the filter is an agent filter, `placedWidgets` never
+    // consults it, and the zero-match card says "No agents match", which stays
+    // true with a fleet-summary tile on the canvas. A bare `.gv-tile` count
+    // asserts something this spec never meant to claim.
+    await expect(page.locator('.gv-tile:not(.gv-tile-widget)')).toHaveCount(0)
     await expect(queryEmpty(page)).toBeVisible()
     await expect(queryEmpty(page)).toContainText('No agents match "zzz"')
     await expect(page.getByRole('button', { name: 'Get started' })).toHaveCount(0)
--- a/src/frontend/src/components/FleetGrid.vue
+++ b/src/frontend/src/components/FleetGrid.vue
@@ -965,6 +965,15 @@ onBeforeUnmount(() => {
   --gv-bk-man: #14b8a6;
   --gv-bk-ext: #ec4899;
   --gv-dots: rgba(17, 24, 39, 0.12);
+  /* ent#325 info-tile chassis. Defined in BOTH blocks: a token defined in one
+     theme only is the same bug the tiles shipped with — a live fallback that
+     never flips. --gv-radius matches .gv-tile's 12px so the inner surface and
+     the chassis share a corner. */
+  --gv-radius: 12px;
+  --gv-peg-bg: #111827;
+  --gv-peg-fg: #f9fafb;
+  --gv-danger-border: rgba(220, 38, 38, 0.35);
+  --gv-skel: rgba(17, 24, 39, 0.07);
   /* Layered depth: top-edge highlight (glass lip) + tight contact shadow +
      mid key shadow + soft ambient falloff. */
   --gv-tile-sheen: linear-gradient(180deg, rgba(255, 255, 255, 0.6), rgba(255, 255, 255, 0) 55%);
@@ -1038,6 +1047,14 @@ onBeforeUnmount(() => {
   --gv-bk-man: #2dd4bf;
   --gv-bk-ext: #f472b6;
   --gv-dots: rgba(249, 250, 251, 0.08);
+  /* ent#325 info-tile chassis — dark half. The peg inverts (light chip on the
+     dark card) so it reads as a raised marker in both themes rather than
+     disappearing into the tile. */
+  --gv-radius: 12px;
+  --gv-peg-bg: #f9fafb;
+  --gv-peg-fg: #111827;
+  --gv-danger-border: rgba(248, 113, 113, 0.45);
+  --gv-skel: rgba(249, 250, 251, 0.10);
   --gv-drag-shadow: 0 28px 56px -12px rgba(0, 0, 0, 0.75);
   --gv-tile-sheen: linear-gradient(180deg, rgba(255, 255, 255, 0.05), rgba(255, 255, 255, 0) 55%);
   --gv-tile-shadow:
--- a/src/frontend/src/components/FleetGrid.vue
+++ b/src/frontend/src/components/FleetGrid.vue
@@ -156,6 +156,35 @@
             @pointerdown.stop.prevent="startConnect(agent.name, $event)"
           ></span>
         </div>
+
+        <!-- Info tiles (ent#325). Same `.gv-tile` chassis and the same
+             `data-agent` hook, so drag / swap-with-preview / keyboard /
+             culling / the snap socket all apply with no second code path.
+             Deliberately NO connect handle: an info tile is not an org node
+             and can never be a reporting-line endpoint. -->
+        <div
+          v-for="w in placedWidgets"
+          :key="w.key"
+          class="gv-tile gv-tile-widget"
+          :class="{
+            dragging: draggingName === w.key,
+            snapping: snappingName === w.key,
+            locked: lockedName === w.key,
+          }"
+          :style="tileStyle(w.key)"
+          :data-agent="w.key"
+          role="listitem"
+          tabindex="0"
+          :aria-label="w.entry.title + ' info tile — drag to any cell, or use arrow keys'"
+          @transitionend="onTileTransitionEnd(w.key, $event)"
+        >
+          <component
+            :is="w.entry.component"
+            v-if="visibleNames.has(w.key)"
+            :agents="agents"
+          />
+          <div v-else class="gv-tile-far">{{ w.entry.title }}</div>
+        </div>
       </div>
     </div>
 
@@ -172,6 +201,35 @@
       </button>
     </div>
     <span class="gv-zoomlvl">{{ Math.round(vz * 100) }}%</span>
+
+    <!-- Tiles menu (ent#325). Placed in the existing bottom control cluster
+         rather than as a fifth top-level header button: the grid header
+         already carries Tidy up / Reset layout and gained Zones / Lines /
+         Group by dept / New dept from #305, and the issue's scope note asks
+         that the responsive ladder not be widened unconditionally. -->
+    <div class="gv-tilesctl">
+      <button
+        type="button"
+        class="act"
+        :class="{ on: tilesMenuOpen }"
+        aria-haspopup="true"
+        :aria-expanded="tilesMenuOpen"
+        title="Show or hide fleet info tiles"
+        @click="tilesMenuOpen = !tilesMenuOpen"
+      >Tiles ▾</button>
+      <div v-if="tilesMenuOpen" class="gv-tilesmenu" @pointerdown.stop>
+        <p v-if="!widgetCatalog.length" class="empty">No info tiles available.</p>
+        <label v-for="w in widgetCatalog" :key="w.id">
+          <input
+            type="checkbox"
+            :checked="isWidgetOn(w)"
+            @change="gridStore.setWidgetEnabled(w.id, $event.target.checked)"
+          />
+          <span>{{ w.title }}</span>
+        </label>
+        <button type="button" class="reset" @click="resetTiles">Reset to defaults</button>
+      </div>
+    </div>
 
     <!-- PROTOTYPE org overlay controls -->
     <div class="gv-orgctl">
@@ -249,6 +307,8 @@
 <script setup>
 import { ref, computed, watch, onMounted, onBeforeUnmount, nextTick } from 'vue'
 import AgentTile from './AgentTile.vue'
+// Side-effect import: registers the info-tile catalog into GRID_WIDGETS.
+import './tiles/catalog'
 import { useFleetGridStore } from '@/stores/fleetGrid'
 import {
   CELL_W,
@@ -263,6 +323,13 @@ import {
   layoutBBox,
   occupantAt,
 } from '@/utils/gridLayout'
+import {
+  catalogFor,
+  isWidgetEnabled,
+  widgetById,
+  widgetIdFromKey,
+} from '@/utils/gridWidgets'
+import { zoneAt } from '@/utils/gridOrg'
 import { useOrgOverlay } from '@/composables/useOrgOverlay'
 
 /**
@@ -350,11 +417,46 @@ const placedAgents = computed(() =>
   props.agents.filter((a) => layout.value[a.name])
 )
 
+// --- info tiles (ent#325) ---
+const tilesMenuOpen = ref(false)
+const widgetCatalog = computed(() => catalogFor(gridStore.isAdmin))
+function isWidgetOn(entry) {
+  return isWidgetEnabled(entry, gridStore.widgetPrefs)
+}
+function resetTiles() {
+  gridStore.resetWidgets()
+  tilesMenuOpen.value = false
+}
+
+/** Enabled tiles that have a position — the widget analogue of placedAgents. */
+const placedWidgets = computed(() =>
+  gridStore.activeWidgetKeys
+    .filter((key) => layout.value[key])
+    .map((key) => ({ key, entry: widgetById(widgetIdFromKey(key)) }))
+    .filter((w) => w.entry && w.entry.component)
+)
+
+/**
+ * Veto cells that fall inside a department frame when seeding a new tile
+ * (#305 interlock D). Rows -3..-1 are empty of TILES, but a department whose
+ * members were dragged upward has a hull that reaches there, and a tile
+ * seeded inside that frame would read as one of its members.
+ */
+function isCellBlocked(c, r) {
+  if (!zones.value || zones.value.length === 0) return false
+  const [x, y] = cellXY(c, r)
+  return !!zoneAt(zones.value, x + CELL_W / 2, y + CELL_H / 2)
+}
+
 function syncLayoutFromAgents() {
   const names = props.agents.map((a) => a.name)
   const systemNames = new Set(props.agents.filter((a) => a.is_system).map((a) => a.name))
-  gridStore.syncLayout(names, systemNames, newcomerOriginFor)
+  gridStore.syncLayout(names, systemNames, newcomerOriginFor, isCellBlocked)
 }
+
+// Toggling a tile on/off re-runs the reconcile so it is seeded (or dropped)
+// immediately rather than on the next roster change.
+watch(() => gridStore.activeWidgetKeys.join('\n'), () => syncLayoutFromAgents())
 
 
 // --- viewport culling (#47: render/hydrate only tiles near the viewport) ---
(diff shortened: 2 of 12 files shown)


vybe:
Triage from `/validate-pr` while working the ready queue — **not merged**, because merging it as-is would achieve nothing and retargeting it as-is would be worse. One action needed from you.

## The base branch is dead

`feature/305-grid-org-overlay` was **squash**-merged as #1918 on 2026-08-08 (`c94847ae` on `dev`). Squash means the branch tip never became an ancestor of `dev`, so the branch still looks alive:

```
$ git branch -r --merged origin/dev | grep 305-grid-org-overlay
(no output)
$ git log --oneline origin/dev..origin/feature/305-grid-org-overlay
7470e09b  fix(security): keep the org chart off the unfiltered socket …
b87f303a  feat(grid): org overlay — department zones + reporting lines …
6932d369  feat(grid): org overlay prototype …
```

Those three are already on `dev`, collapsed into `c94847ae`. So merging this PR would merge into a branch nothing will ever ship from, and the required checks never ran here at all — this PR shows `CLEAN` only because a non-`dev`/`main` base has no required checks attached.

## Retargeting to `dev` without a rebase is the trap

```
$ git diff --stat origin/dev...origin/feature/ent325-grid-widget-chassis
 31 files changed, 4145 insertions(+), 56 deletions(-)
```

The merge base predates the squash, so the diff re-proposes all of #1918 — `useOrgOverlay.js` (+541), `gridOrg.js` (+382), `test_305_org_tag_integrity.py` (+194) — plus `src/frontend/src/views/Agents.vue`, which was **deleted** since (ent#260 folded the Agents page into the Dashboard List view). A reviewer would be reading 4145 lines of already-merged code to find your 1184.

## The fix, verified

Rebasing onto `dev` drops the already-squashed commits cleanly — no conflicts, tested in a scratch worktree:

```
$ git rebase --onto origin/dev origin/feature/305-grid-org-overlay feature/ent325-grid-widget-chassis
Successfully rebased and updated.

$ git diff --stat origin/dev...HEAD
 9 files changed, 1184 insertions(+), 25 deletions(-)
```

That is exactly this PR's own work — `InfoTile`, `GRID_WIDGETS`, layout v2, Tiles menu, `gridWidgets.spec.js`.

## What I'd like you to do

```bash
git fetch origin
git rebase --onto origin/dev origin/feature/305-grid-org-overlay feature/ent325-grid-widget-chassis
git push --force-with-lease
```

then retarget the PR to `dev`. **Retarget after the push, not before** — the required checks only attach on a `dev`/`main` base, so pushing first means the first run is the one that counts.

I stopped short of doing the rebase and force-push myself since it rewrites your branch. Say the word and I'll run it.

vybe (changes requested):
## `/validate-pr` — request changes

Two independent blockers. The second is the substantive one; the first is unchanged since the 2026-08-06 triage above.

Everything else passes: security scan clean (no keys, emails, public IPs, `.env`, hardcoded secrets), 9 files, conventional commit, ent#325 exists and is correctly labelled (`priority-p2 · type-feature · theme-ui-ux · complexity-high`).

---

## 1. ❌ The info tile is unthemed in dark mode

`InfoTile.vue` opens its style block with:

> *"Tokens come from the `--gv-*` cascade FleetGrid establishes, so an info tile and an agent tile can never drift apart on colour."*

That is the right invariant. The code doesn't deliver it — **10 of the 12 `--gv-*` tokens the two new components consume are never defined anywhere in `src/frontend/src`**, so every one of their fallbacks is permanently live in both themes:

| Consumed | Status | The token that actually exists |
|---|---|---|
| `--gv-tile-bg` | missing → always `#fff` | `--gv-tile` (light `rgba(255,255,255,.9)` / dark `rgba(31,41,55,.9)`) |
| `--gv-fg` | missing → always `#111827` | `--gv-text` (light `#111827` / dark `#f9fafb`) |
| `--gv-tile-border` | missing | `--gv-border` (light `#e5e7eb` / dark `#374151`) |
| `--gv-ok` | missing | `--gv-green-text` (light `#16a34a` / dark `#4ade80`) |
| `--gv-warn` | missing | `--gv-yellow-text` (light `#a16207` / dark `#facc15`) |
| `--gv-peg-bg`, `--gv-peg-fg`, `--gv-danger-border`, `--gv-skel`, `--gv-radius` | missing | genuinely new — need defining |
| `--gv-muted`, `--gv-tile-shadow` | ✅ resolve | — |

**The two that do resolve are what make this a contrast failure rather than a cosmetic one.** `--gv-muted` correctly flips to `#9ca3af` under `:root.dark .fleet-canvas` (FleetGrid.vue:995) — and lands on a tile background hardcoded to `#fff`. So dark mode renders:

- a **white card** sitting among dark `AgentTile`s on the dark fleet canvas, and
- **gray-400 secondary text on white ≈ 2.4:1**, under the 4.5:1 WCAG AA floor.

`docs/memory/design-system-contract.md` makes light and dark both first-class a non-negotiable, so this blocks.

### Fix

Rename the five that have an existing equivalent, and define the five genuinely-new ones in **both** `.fleet-canvas` blocks — light at `FleetGrid.vue:929` and dark at `FleetGrid.vue:994`. Defining in only one is the same bug again.

### Worth adding a test

The comment asserted the invariant and nothing enforced it, which is why this got through a genuinely careful 28-test suite. A guard that collects every `var(--gv-…)` referenced by the tile components and asserts each name is defined in the cascade would have caught it, and keeps catching it as tiles are added one sub-issue at a time (#95–#101, #259). Cheap, and this is the seam where it pays off repeatedly.

---

## 2. ❌ The base branch is still dead

Unchanged from the triage comment above — `feature/305-grid-org-overlay` was **squash**-merged as #1918, so its tip never became an ancestor of `dev` and the branch still looks alive. This PR reports `mergeable: CLEAN` only because a non-`dev`/`main` base has no required checks attached; none of the gates have ever run against this code.

I re-verified the rebase against current `dev` (`17cb14f4` — it absorbed six merges in the last hour, so this needed rechecking):

```
$ git rebase --onto origin/dev origin/feature/305-grid-org-overlay
Successfully rebased and updated.          # 1 commit, zero conflicts

$ git diff --stat origin/dev...HEAD
 9 files changed, 1184 insertions(+), 25 deletions(-)
```

Still clean, and still exactly this PR's own work. Retargeting **without** that rebase re-proposes all of #1918 (~4145 lines) plus the since-deleted `views/Agents.vue`.

---

## Requested changes

- [ ] Point the tile components at the tokens that exist (`--gv-text`, `--gv-tile`, `--gv-border`, `--gv-green-text`, `--gv-yellow-text`)
- [ ] Define `--gv-peg-bg`, `--gv-peg-fg`, `--gv-danger-border`, `--gv-skel`, `--gv-radius` in **both** the light and dark `.fleet-canvas` blocks
- [ ] Add a guard asserting every `--gv-*` a tile consumes is defined in the cascade
- [ ] `git rebase --onto origin/dev origin/feature/305-grid-org-overlay`, then **retarget the PR to `dev` before pushing**, so the required checks fire on the push rather than needing a second one
- [ ] Rephrase `Related to abilityai/trinity-enterprise#325` → `Fixes abilityai/trinity-enterprise#325`. Cross-tracker refs don't auto-promote either way, so ent#325 still needs a manual `status-in-dev` at merge — but the closing keyword keeps the link machine-readable

## Non-blocking

- `architecture.md` § Dashboard Grid view and `feature-flows/dashboard-grid-view.md` both describe the lattice and its occupants; a second occupant type plus the layout v1→v2 migration belongs in one of them. Happy to take that as a follow-up rather than hold this.

---

The rest of this is strong work, and I want to be clear the theming miss is narrow rather than characteristic. The org-overlay interlock is handled at **both** ends (helper carry-through *and* `applyLayout` merging instead of blind-replacing) with the stated reasoning that one guard is how the bug existed in the first place; the `nearestFreeCell` behaviour-preservation test is exactly the right thing to pin; and the prefs-as-sparse-override-map decision — with a test that an `adminOnly` tile stays off for a non-admin whose stale prefs say on — is the kind of thing that usually gets found in production. Rebase, fix the token names, and this is ready.

dolho:
All five addressed. Verified the token audit independently before changing anything — 12 consumed, 2 defined, exactly as reported.

## 1. Theming

Five renamed to the tokens that exist, with fallbacks updated to that token's **light** value so a fallback that ever does fire is at least the right colour rather than an unrelated one. Five genuinely-new ones defined in **both** `.fleet-canvas` blocks. `--gv-radius` is `12px` to match `.gv-tile`, so the inner surface and the chassis share a corner; the peg inverts between themes so it reads as a raised marker either way (16.98:1 both directions).

Measured the contrast rather than assuming the rename fixed it:

```
dark, before:  --gv-muted on #fff         2.54:1   FAIL (AA 4.5)
dark, after:   --gv-muted on --gv-tile    5.92:1   pass
dark, after:   --gv-text  on --gv-tile   14.39:1   pass
light, after:  unchanged — 4.83 / 17.74:1
```

(You estimated ~2.4:1; I get 2.54:1. Same conclusion.)

## 2. The guard

`tests/unit/gridTokens.spec.js`, four checks: every consumed token is defined somewhere; no consumed token is defined in one theme block only; the two blocks declare the same set (so a token added for a future tile can't land in one alone); and no tile hardcodes a raw colour outside a `var()` fallback — the original fallbacks were literals chosen to look right in light mode, which is exactly what made this invisible to anyone reading the file in light mode.

Static by design. The unit suite is node-environment ("pure modules only, no DOM"), and jsdom couldn't catch it anyway: it does no cascade resolution, so `getComputedStyle` returns the fallback and the broken state looks *identical* to the fixed one. Parsing the source is what distinguishes them.

Mutation-tested rather than asserted:

- all three components at their pre-fix state → **2 failed**
- new tokens defined in **light only** — the half-fix you flagged — → caught by both the one-theme-only and same-token-set checks
- restored → **76 passed**

## 3. Rebase + retarget

```
$ git rebase --onto origin/dev origin/feature/305-grid-org-overlay
Successfully rebased and updated.        # 2 commits, zero conflicts

$ git diff --stat origin/dev...HEAD
 10 files changed, 1345 insertions(+), 25 deletions(-)
```

Retargeted to `dev` **before** pushing, so the required checks fire on this push. Force-with-lease pinned to the exact pre-rebase SHA.

## 4. `Fixes abilityai/trinity-enterprise#325`

Done, and the stale "Stacked on #1918 / retarget once it merges" note replaced with what actually happened. Noted that ent#325 still needs a manual `status-in-dev` at merge.

## Non-blocking docs

Taking you up on the follow-up offer — `architecture.md` § Dashboard Grid view and `feature-flows/dashboard-grid-view.md` want the second occupant type and the layout v1→v2 migration, and that's better as its own change than appended here.

vybe (approved):
## `/validate-pr` — approve

All five requested changes landed, and I verified them against the head revision (`e4c4a3d`) rather than taking the response comment at its word. Two came back broader than asked.

---

### 1. ✅ Tokens — real in both directions

The five bogus names are gone from both components. What makes this a genuine fix rather than a rename is the second half: the five new tokens are defined in **both** cascade blocks with values actually inverted per theme —

| | `FleetGrid.vue:972-976` (light) | `FleetGrid.vue:1053-1057` (dark) |
|---|---|---|
| `--gv-peg-bg` | `#111827` | `#f9fafb` |
| `--gv-peg-fg` | `#f9fafb` | `#111827` |
| `--gv-danger-border` | `rgba(220,38,38,.35)` | `rgba(248,113,113,.45)` |
| `--gv-skel` | `rgba(17,24,39,.07)` | `rgba(249,250,251,.10)` |

— not the light block pasted twice, which is the failure mode "define in **both**" existed to prevent.

### 2. ✅ The guard went past what I asked for

I asked for a check that every `--gv-*` a tile consumes is defined. `gridTokens.spec.js` delivers that, then adds two things I didn't ask for and would have:

- **Test 3 polices whole-block parity**, not just the tile-consumed subset — so a token added for a future tile (#95–#101, #259) cannot land in one block by itself. That generalizes from the instance to the class.
- **Test 4 bans a raw colour outside a `var()` fallback.** That is the specific property that made the original break invisible: the fallbacks were literals chosen to look right in light mode, so the file read as correct to anyone not in dark.

The header's argument for why this is static rather than jsdom is also right and worth keeping — jsdom does no cascade resolution, so `getComputedStyle` returns the fallback and the broken state is byte-identical to the fixed one. A mount test here would have passed on the bug.

### 3. ✅ Rebase + retarget — and the checks now mean something

**22/22 pass** against `dev`, including `build` (which runs `check:tokens` → `test:unit` → `vite build`) and a real 4m34s `e2e`. That closes the honest verification caveat in the PR body: the 55-test and 638-module numbers no longer have to be taken on trust, CI reproduced them.

### 4. ✅ `Fixes abilityai/trinity-enterprise#325`

---

## The part I missed

Commit 4 fixes a bug that was in the revision I reviewed and that I did not catch. `FleetSummaryTile` shipped `:link-to="{ name: 'Agents' }"` — there is no route by that name, since the standalone Agents page folded into the Dashboard and left `/agents` as an *unnamed* redirect. `RouterLink` throws during render, which aborts Vue's update for the whole tree and **freezes the dashboard**, surfacing as a type-filter pill visibly holding its text while nothing re-renders and the tile itself shows as an empty box. Nothing in that symptom points at a link in a tile.

Now `{ name: 'Dashboard', query: { view: 'list' } }` — the documented one-shot non-persisting view intent — and pinned by `gridTileLinks.spec.js` with two anti-vacuity assertions (router scan must find >5 names, tile scan must find >0 links) so the guard cannot silently police nothing. Worth noting the path that caught it was the e2e, which is `ui`-label-gated and does not run on most PRs; the static guard is the right response to that.

The e2e change is a scoping fix, not a weakened assertion — `.gv-tile:not(.gv-tile-widget)` still requires zero **agent** tiles at zero matches, and the reasoning is correct: the filter is an agent filter, `placedWidgets` never consults it, and "No agents match" stays true with a fleet-summary tile on the canvas.

---

## Merging

Security scan clean (no keys, emails, public IPs, `.env`, hardcoded secrets, credential files). No backend module or `os.getenv()` packaging surface. 12 files, focused, no unrelated refactoring.

**Two follow-ups, neither blocking:**

- **ent#325 needs a manual `status-in-dev`** at merge — cross-tracker refs don't auto-promote in either direction.
- Docs follow-up filed as #2126: `architecture.md § Dashboard Grid view` and `feature-flows/dashboard-grid-view.md` both describe the lattice and its occupants, and neither knows about a second occupant type or the layout v1→v2 migration. Better as one pass once a couple of real tiles have landed than as a hold on the chassis.

Good work on the interlock reasoning, and on finding the render throw yourself.

PR #2044, opened 2026-08-06
feat(prompt): model-conditional prompt tiers — mechanism only, provable no-op (ent#243)

obasilakis:
## What

Lands the **mechanism** for model-conditional platform prompts, and nothing else.
`_MINIMAL_PREFIXES` ships **empty**, so every model resolves `VERBOSE` and the
rendered prompt is byte-identical to `dev`. Enabling a family is PR 2, behind
evaluation evidence.

Implements ent#243 PR 1 (ACs on that issue were revised today to match the
research pass in its comment thread — the original ACs described a
runtime-scoped version this supersedes).

## Why the model, not the runtime

The Claude 5 context-engineering guidance (rules → judgment, examples →
interface design) applies to frontier coding-post-trained models and is
**actively harmful** below that line — Sonnet 4.5 and Haiku 4.5 need the
structure it deletes. OpenAI (Codex over-prompting) and Google (Gemini 3
over-analyzing verbose prompts) reached the same conclusion independently, so
the fault line is model class, not vendor and not harness.

MODEL-001 lets any turn pin any model string, including free text, so a
`claude-code` agent routinely runs Haiku 4.5. Gating on `AGENT_RUNTIME` would
silently degrade most of the selectable surface. `runtime` keeps its orthogonal
job — Codex MCP tool naming (#1187 F-MCP) — and a test pins the signature so the
two axes can't quietly merge.

## Design

- **`services/prompt_tier.py`** — family-prefix classifier mirroring
  `services/model_context.py`. Total, pure-stdlib, never raises.
  Unknown/empty/`None` → `VERBOSE`. That asymmetry is load-bearing:
  over-instructing a Claude 5 model costs tokens, under-instructing a 4.5 model
  degrades it *silently*. Not vendored to the agent server — the backend composes
  the prompt and ships it in the turn payload.
- **Section derivation, not a second prompt.** `PLATFORM_INSTRUCTIONS` stays one
  authored literal; sections are derived by splitting on its top-level `###`
  headings. Two full prompt strings guarantee drift, and the drift that matters
  is a security instruction added to one variant and not the other. The delimiter
  cannot match the `####` subsections, and `split`/`join` round-trips byte-exactly.
- **Un-gateable sections:** Operator Communication (#1402 fire-and-park contract,
  sentinel-locked and synced to `config/trinity-meta-prompt/prompt.md`), the
  shared-memory-dir leak warning, and Package Persistence. None is tool-usage
  guidance; none has a tool description to live in.
- **Fail-loud on rename.** An unmapped heading renders at every tier (safe), so
  the heading set is CI-pinned — otherwise a rename silently orphans its
  drop-list entry.

## Model-threading gaps closed

`pull_coordination_service` omitted `model` from its `ExecutionContext`
**entirely** — absent, not `None` — so every pull-claimed turn would have
resolved `VERBOSE` forever with nothing to indicate why. Both degraded fallback
paths now pass the model too, so a context-build failure cannot silently swap the
prompt tier as well as the context block.

An **unpinned chat turn stays `VERBOSE` by design**: the container picks the model
(`claude_code.py` → `claude-sonnet-4-6`) *after* composition, so the decision does
not exist backend-side and cannot be threaded. That default is already correct for
a 4.x model. Resolving the platform default backend-side would trade a known-safe
gap for a hidden one, since it can diverge from the agent's own
`get_default_model()`. Tracked as a follow-up; the same split underlies #1521's
context-window safe floor.

## Verification

- **Byte-identity proven directly against `origin/dev`** — `PLATFORM_INSTRUCTIONS`
  hashes `sha256 4cabe4557f2580580d154476406ace31525bdde4354a9ef77d6884affba2158e`,
  9207 bytes, on both sides. Identical, not merely equivalent.
- **36 new tests** (`tests/unit/test_ent243_prompt_tier.py`), including a
  parametrized no-op proof over the **live** `PRESET_MODELS` list parsed from
  `ModelSelector.vue` — a preset added later is covered without anyone updating
  the test — plus a guard against that parse silently breaking.
- Existing prompt contracts unaffected: `test_1402_prompt_contract.py`,
  `test_1535_report_prompt_guidance.py`, `test_platform_prompt_runtime.py`,
  `test_public_channel_prompt.py` all green. 351 passed across the
  prompt/chat/pull/task-execution selection.
- `tests/lint_sys_modules.py`: 194 violations vs baseline 240 — no new.

### Pre-existing failure, not from this PR

The full local unit suite fails
`test_1081_physical_meter.py::TestPerAgentMeter::test_available_floors_at_zero_over_ceiling[sqlite]`.
**Clean `origin/dev` @ `47f8843b` fails the identical test with the identical
command**, so this branch inherits it rather than causing it. It is an
order-dependent `sys.modules` leak surfaced by `pytest-randomly`; the file passes
14/14 in isolation. Filed separately.

## Requirements

`docs/memory/requirements/runtimes.md` §41 (written before implementation per
RoE #1).

### DIFF
--- a/docs/memory/requirements/runtimes.md
+++ b/docs/memory/requirements/runtimes.md
@@ -267,3 +267,90 @@ endpoint. See the [Harness Authoring Guide](harness-authoring-guide.md) for addi
 a fourth runtime.
 
 ---
+
+## 41. Model-Conditional Prompt Tiers (ent#243)
+
+### 41.1 Tier Resolution — the axis is the MODEL, not the runtime
+
+Anthropic's Claude 5-generation context-engineering guidance (rules → judgment,
+examples → interface design) applies to *frontier coding-post-trained models*,
+and is **actively harmful** on smaller/older ones — Sonnet 4.5 and Haiku 4.5 need
+the explicit structure it removes. Three vendors reached the same conclusion
+independently (OpenAI on GPT-5-Codex over-prompting, Google on Gemini 3
+over-analyzing verbose prompt engineering), so the fault line is model class, not
+vendor and not harness.
+
+This is why the tier is **not** keyed on `AGENT_RUNTIME`. MODEL-001 lets a
+schedule, loop, or chat turn pin any model string — including free-text — so a
+`claude-code` agent routinely runs Haiku 4.5. Gating on the runtime label would
+silently degrade the majority of the selectable surface (7 of 9 `PRESET_MODELS`
+entries sit in the tier where the guidance does not apply). Runtime already has
+its own orthogonal job in this module: `_adapt_instructions_for_runtime` rewrites
+MCP tool naming for Codex (#1187 F-MCP). The two axes stay separate.
+
+**Functional requirements:**
+- **FR-1 — Classifier:** `services/prompt_tier.py::resolve_prompt_tier(model)`
+  returns `PromptTier.MINIMAL | PromptTier.VERBOSE`. Total function, never raises,
+  pure-stdlib. Mirrors the `services/model_context.py` shape: family-prefix rules,
+  first-match-wins, most-specific-first.
+- **FR-2 — Unknown fails toward VERBOSE.** An unrecognized, empty, or `None` model
+  resolves to `VERBOSE` — today's prompt. Over-instructing a Claude 5 model costs
+  tokens; under-instructing a 4.5 model silently degrades it, and free-text model
+  passthrough makes an unknown id routine rather than exceptional. This mirrors
+  `model_context.py`'s "fail toward the conservative value" invariant and is the
+  property that bounds this feature's blast radius to *status quo*.
+- **FR-3 — Section-level gating, never two prompt strings.** `PLATFORM_INSTRUCTIONS`
+  stays a single authored literal; sections are derived from it by splitting on
+  its `###` headings and filtered by tier at render time. Maintaining two full
+  prompts guarantees drift — a security instruction added to one variant and not
+  the other is the failure mode this forbids by construction.
+- **FR-4 — A section with no explicit tier renders always.** An unmapped or renamed
+  heading falls back to always-render, and the heading set is CI-pinned so a rename
+  fails loudly instead of silently changing what agents are told.
+- **FR-5 — Safety and privacy sections are un-gateable.** Operator Communication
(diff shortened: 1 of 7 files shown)


vybe (approved):
Shipping the mechanism as a provable no-op is the right shape, and the proof is the right one — `test_every_preset_renders_the_unchanged_prompt` over the **live** preset list plus `test_minimal_prefix_set_is_empty`, so the no-op is pinned rather than asserted in the description. A fleet-wide prompt change genuinely does not belong in the same PR as the mechanism enabling it.

The fail-toward-VERBOSE default is the load-bearing decision and the asymmetry argument for it is correct: over-instructing a frontier model costs tokens, under-instructing a 4.5 model degrades it **silently, with no error and no signal**. With MODEL-001 free-text passthrough an unknown id is routine rather than exceptional, so that default bounds the blast radius to status quo. The docstring's explicit "do NOT modernize this by defaulting unknown to MINIMAL" is worth keeping — that is exactly the change someone will propose later on plausible-sounding grounds.

Keying on the model rather than `AGENT_RUNTIME` is right for the reason given: a `claude-code` agent routinely runs Haiku 4.5, so a runtime gate would silently degrade most of the selectable surface, and `runtime` already has an orthogonal job (Codex MCP tool naming, #1187 F-MCP). Pinning that with `test_tier_is_not_keyed_on_runtime` keeps the two axes from being conflated later.

Good call stating the **not-vendored** contrast to Invariant #5 explicitly, with the condition under which that would change — the backend composes the prompt and ships it in the turn payload, so the agent server never resolves a tier.

The MINIMAL-path tests are the part I'd have asked for if they were missing: `test_minimal_keeps_the_operator_contract_verbatim`, `..._keeps_the_user_memory_leak_warning`, `..._has_no_dangling_delimiters` all exercise a code path unreachable today, which is what makes PR 2 a config change rather than a leap.

**One thing to settle before PR 2, not here:** ent#243 is on the enterprise tracker, and the standing rule is that enterprise-tracker features are entitlement-gated by default unless explicitly decided OSS-core. This PR lands ungated in `src/backend/services/`. That is fine as-is *because it is a no-op* — nothing diverges between editions today. But PR 2 is where it bites: enabling a family would mean OSS and enterprise agents get materially different system prompts, and "which edition gets the better prompt" is a product call rather than an implementation detail. Worth confirming the gating shape with @vybe before PR 2 rather than after.

PR #2056, opened 2026-08-07
fix(skills): share one authority rule between the URL parser and the free-text PAT scrubbers

### LINKED ISSUE #2052: bug(skills): the free-text PAT scrubbers only fire on a literal `https://`, missing protocol-relative, scheme-less and alternate-scheme URLs
Follow-up from the #2041 / #2043 pair. ent#347 (#2041) predicted that its regex and ent#334's parser would need to agree on the authority boundary; I ran that cross-check when landing #2043.

## They agree on the boundary

Where both fire, `utils.url_validation.strip_url_credentials` (parse-based, #2043) and `skill_source_clone._CREDENTIAL_URL_RE` (regex, #2041) resolve the **same** host — verified over the double-`@`, triple-`@`, port, IDN and userinfo-with-no-host shapes, and both correctly leave `https://github.com/o/r?ref=a@b` alone.

## They disagree on *when* to fire

`_CREDENTIAL_URL_RE = re.compile(r"https://[^\s/?#]+@")` is anchored on a literal `https://`, so three shapes the parser handles slip past it:

| input | `strip_url_credentials` | `redact` / `_scrub_pat` |
|---|---|---|
| `//tok@github.com/o/r` (protocol-relative) | `//github.com/o/r` | **leaks `tok`** |
| `tok@github.com/o/r` (scheme-less shorthand) | `github.com/o/r` | **leaks `tok`** |
| `git+ssh://tok@github.com/o/r` | `git+ssh://github.com/o/r` | **leaks `tok`** |

## Why it is reachable, and why it is only P3

The clone path builds its URL with `_authenticated_url`, which splices a literal `https://` — so git's own stderr is covered by the existing anchor. The gap is on the *stored* URL reaching a scrubber by another route:

- `_adopt_legacy_clone` writes source rows with **no validation at all**, and `validate_skills_library_url` accepts the scheme-less shorthand — so a stored URL can legitimately be scheme-less or protocol-relative.
- `skill_service._scrub_pat` runs over `str(e)` of an **arbitrary** exception, whose text is not constrained to what `_authenticated_url` produced.

Output lands in `system_settings['skills_library_last_error']` — durable and admin-rendered — which is exactly the sink ent#347 was raised about.

Narrow enough to be P3: it needs a stored URL in a non-`https://` form *and* an error path that echoes it verbatim.

## Suggested fix

Have the free-text scrubbers reuse the parser's notion of an authority instead of carrying a second, differently-anchored pattern. #2041's own argument applies — two hand-written patterns that had drifted apart were the root cause there; `_CREDENTIAL_URL_RE` and `strip_url_credentials` are now the same situation one level up. A cross-check test asserting the two agree on a shared corpus belongs with the fix.

Note the scrubbers are deliberately text-oriented (git stderr carries prose and several URLs), so this is not simply "call the parser" — the authority-recognition needs to be shared without losing multi-URL free-text handling.

dolho:
## What

`skill_source_clone._CREDENTIAL_URL_RE` was `https://[^\s/?#]+@` — anchored on a literal `https://` — while the parse-based `utils.url_validation.strip_url_credentials` resolves an authority with no scheme at all. The two agreed on **where** an authority ends but not on **when** to fire, so three shapes slipped past the regex and were echoed verbatim:

| input | `strip_url_credentials` | `redact` / `_scrub_pat` (before) |
|---|---|---|
| `//tok@github.com/o/r` | `//github.com/o/r` | **leaked `tok`** |
| `tok@github.com/o/r` | `github.com/o/r` | **leaked `tok`** |
| `git+ssh://tok@github.com/o/r` | `git+ssh://github.com/o/r` | **leaked `tok`** |

Reachable rather than theoretical: `_adopt_legacy_clone` writes source rows with no validation and `validate_skills_library_url` accepts the scheme-less shorthand, so a stored URL may legitimately take those forms; `_scrub_pat` then runs over `str(e)` of an **arbitrary** exception, whose text is not constrained to what `_authenticated_url` builds. Sink is `system_settings['skills_library_last_error']` — durable and admin-rendered.

## Approach

ent#347 collapsed two drifted hand-written patterns into one. This is the same argument one level up, so the fix is **not a fourth pattern**: the rule moves next to the parser as `scrub_url_credentials_in_text`, `_CREDENTIAL_URL_RE` is deleted, and `redact` delegates.

It does **not** call the parser — git stderr is prose plus several URLs, so there is no single URL to hand `urlparse`. What is shared is the rule, not the call.

## One shape beyond the three reported

The fix also covers the empty-userinfo shape `https://@host`: `parsed.username` reads it as falsy, and the old regex's `+` quantifier required at least one character before the `@`, so it never matched. `strip_url_credentials` already handled it (its `"@" not in netloc` test is deliberate), so this was a fourth point of disagreement between the two.

## A regression caught mid-way: the scan went quadratic

The `https://` anchor being removed was also the literal `str.find` could jump to. Expressing the new rule as a class-first pattern made the scan **quadratic**:

| input | class-first pattern | shipped run-walk |
|---|---|---|
| 200 KB, one run, no `@` | **24,297 ms** | — |
| 500 KB, one run, no `@` | **151,568 ms** | **1.2 ms** |
| 1 MB realistic stderr | 45 ms | 29 ms |

`redact` takes **raw `git stderr`**, which is unbounded and partly remote-controlled (a branch or path name the remote echoes back), so that was a DoS reachable from a hostile repository. The shipped version walks maximal authority-legal runs instead, which is linear. `test_scan_is_linear_on_a_long_run_with_no_at` pins it with a deliberately loose 2 s bound (~1600x headroom) so it fails only on a return to quadratic behaviour.

## Deliberate over-match

A bare `user@host` in prose (an email, or scp-like `git@github.com:o/r`) is redacted. `strip_url_credentials` reads those the same way — the two agreeing is the property under test — and the asymmetry justifies it: a false positive costs a mangled word in a diagnostic, a false negative costs a PAT persisted and rendered to admins. Parity holds over the tested corpus; the text scrubber still over-redacts a few degenerate shapes the parser leaves alone (e.g. an empty path segment `…/o//tok@r`), which is pre-existing behaviour for the `//` form and in the safe direction.

## Tests

`tests/unit/test_2052_scrubber_authority_parity.py` — mostly **parity** assertions over one shared corpus (the ent#334 cross-check shapes plus the three that regressed):

- same resolved authority (`redact(u).replace("***@", "", 1) == strip_url_credentials(u)`)
- same answer to *is a credential present at all* — the actual defect
- no userinfo survives
- structural: `_CREDENTIAL_URL_RE` is gone, so a local pattern cannot be reintroduced
- regression guards: `?ref=a@b` untouched (the removed anchor was also what kept the pattern off a query `@`, which is why the bare form carries a run-start lookbehind), multi-URL free text, idempotency, never-raises, linearity

**21 of these fail against the pre-fix scrubber**, one with the token verbatim (`clone failed for ghp_TOK@github.com/o/r`).

`test_ent183_skill_packages.py` gains one stub entry: it installs `sys.modules` stubs whose map must mirror every name imported from the real module, so a missing entry is an `ImportError` at **collection** — which cascaded into the `#1898` isolation test.

## Verification

```
8200 passed, 4 failed, 18 skipped   (full backend unit suite)
```

The 4 failures are pre-existing and environmental — verified failing identically on clean `origin/dev`: 3× `AttributeError: module 'bcrypt' has no attribute '__about__'` (passlib/bcrypt version skew in the local venv) and 1× `ZoneInfoNotFoundError: Europe/Kiev` (deprecated tzdata alias absent; `Europe/Kyiv` present). Neither is touched by this diff.

## Follow-up, not in this PR

`utils/credential_sanitizer.py::_URL_USERINFO_RE = ://[^/@\s]+@` is a **third** scrubber with the same class of gaps — it leaks on double-`@` (`https://a@b@github.com/o/r` → `https://***@b@github.com/o/r`, the same bug ent#347 fixed) and misses protocol-relative and scheme-less entirely. It feeds `git_service`, `repo_binding`, `routers/git.py` and `system_service`, and is mirrored into the agent base image (`agent_server/routers/git.py`), so fixing it needs the Invariant #5 mirror plus a base-image rebuild. Out of scope here.

Fixes #2052


vybe (approved):
Validated via /validate-pr. Traced the run-walk scan: query/path @ protected by the boundary rule, protocol-relative + scheme-less + alternate-scheme + empty-userinfo all covered, and the linearity guard pins the quadratic regression found mid-fix. Delegation removes the last local pattern (structural test asserts _CREDENTIAL_URL_RE is gone) — the right ent#347 continuation. Ran the touched unit files locally: 131 passed (test_2052_scrubber_authority_parity + test_ent183_skill_packages). Patched the body to Fixes #2052.

PR #2057, opened 2026-08-07
fix(pull): stop the pilot flag claiming reach the dispatch topology does not have

### LINKED ISSUE #2048: Pull pilot can only route agent-to-agent work — cron/schedule triggers structurally cannot reach the queue
## Summary

`PULL_MODE_PILOT_AGENTS` can only ever route **agent-to-agent** work through the
durable queue. `schedule`, `webhook`, `loop`, `event`, `fan_out` and `reminder`
are all declared autonomous in `_AUTONOMOUS_TRIGGERS`, but none of them can
structurally reach `pull_owns_dispatch` — so on a cron-driven agent the pilot
flag is inert, and #1766's soak would measure nothing.

Surfaced while choosing a pilot agent on `eu2` for the #1766 soak.

## The three dispatch entry points

`capacity_manager` gates `pull_exclusive` on `overflow_policy == "queue_persistent"`
(`capacity_manager.py:~360`), and only one of the three producers uses it:

| Path | Caller | `overflow_policy` | Reaches `pull_owns_dispatch`? |
|---|---|---|---|
| `task_execution_service.py:1069` | **scheduler — all cron**, fan-out | `reject` | **No** |
| `dispatch_admission_service.py:165` | sequential `chat_with_agent`, human chat | `queue_in_memory` | **No** |
| `chat_execution_service.py:1252` | **parallel `chat_with_agent`**, MCP/manual task | `queue_persistent` | **Yes** |

`_AUTONOMOUS_TRIGGERS` (`task_execution_service.py:259`) is
`{schedule, webhook, loop, event, fan_out, agent, reminder}`. Of those, only
`agent` is reachable today. The trigger set and the dispatch topology disagree,
and nothing fails loudly when they do — `pull_owns_dispatch` returns `False` and
the row takes the unchanged push path, which is indistinguishable from the flag
not being set.

## Why it matters

Measured on `eu2` over 7 days, per-agent pull-eligible volume:

| Agent | total execs | `schedule` | `agent` | **pull-eligible** |
|---|---|---|---|---|
| cornelius-oracle | 329 | 0 | 329 | **329** |
| oracle-8-biotech | 151 | 126 | 2 | 2 |
| oracle-6-science | 150 | 126 | 4 | 4 |
| oracle-3-ai | 145 | 126 | 5 | 5 |
| cleon | 105 | 95 | 4 | ~1 |
| brier-hq | 45 | 35 | 0 | **0** |

The fleet's dominant traffic is cron, and cron cannot be piloted. Only a fan-in
hub fed by parallel agent-to-agent calls produces enough volume for the soak to
mean anything — which narrows pilot selection to essentially one agent on this
instance, for a reason that is invisible from the config.

## The doc says otherwise, and its troubleshooting advice misfires

`docs/testing/PULL_MIGRATION_TESTING.md` §9, M1:

> Expected after the #1766 gate: autonomous triggers ~100% `pulled`; anything
> still `pushed` should be interactive (`manual`/`user`/`chat`)
>
> A pushed **autonomous** row means the producer gate did not engage — check the
> backend actually has `PULL_MODE_PILOT_AGENTS` in its env.

For a `schedule` row both statements are wrong. It will be `pushed` no matter
what, and the operator following that advice goes looking for a missing env var
that is present and correct. On a cron-driven pilot, M1 reads ~0 `pulled` and
the documented conclusion — "the soak is measuring nothing" — is right for the
wrong reason.

## What to decide

Two coherent options; this issue is for picking one, not for assuming either.

1. **Accept the current reach and correct the docs.** #1766's soak covers
   parallel agent-to-agent dispatch only. Narrow `_AUTONOMOUS_TRIGGERS`' use in
   `pull_owns_dispatch` (or document the reachable subset at the call site), fix
   §9's M1 prose, and say plainly in the pilot-selection guidance that a
   cron-only agent is not a viable pilot.
2. **Wire the scheduler path.** Give `task_execution_service`'s dispatch a
   `queue_persistent` policy under the pilot flag so cron work can be claimed.
   Larger change, needs its own review, and lands squarely on the reliability
   spine — but it is what §9 already claims happens.

Option 1 is cheap and makes the system honest about what it does. Option 2 is
what makes the soak cover the fleet's real traffic. They are not exclusive —
1 now, 2 later, is a reasonable sequence.

## Acceptance criteria

- [ ] A decision recorded on which option (or both, sequenced)
- [ ] `pull_owns_dispatch` no longer implies reach it does not have — either the
      trigger set it consults is narrowed to what is reachable, or the call site
      documents the gap explicitly
- [ ] §9's M1 expectation and its troubleshooting line corrected
- [ ] A pushed row for a trigger that *should* have been pulled is
      distinguishable from one that was never eligible (today both are silent)
- [ ] Pilot-selection guidance states that cron-only agents cannot be piloted

## Refs

- #1766 (production soak), #1081 (pull umbrella), #1982 / `c455aa56` (the
  either/or pilot flag + §9 measurement set)
- `src/backend/services/pull_pilot.py`, `src/backend/services/capacity_manager.py`,
  `src/backend/services/task_execution_service.py`

dolho:
## What

`PULL_MODE_PILOT_AGENTS` can only route **agent-to-agent** work. `capacity_manager` offers a row to the durable queue solely when its producer passed `overflow_policy="queue_persistent"`, and one of the three producers does:

| Producer | Carries | `overflow_policy` | Reaches `pull_owns_dispatch`? |
|---|---|---|---|
| `task_execution_service` | scheduler — **all cron** — + fan-out | `reject` | **No** |
| `dispatch_admission_service` | sequential `chat_with_agent`, human chat | `queue_in_memory` | **No** |
| `chat_execution_service` (`POST /task`) | parallel `chat_with_agent`, MCP/manual task | `queue_persistent` | **Yes** |

`_AUTONOMOUS_TRIGGERS` declares seven triggers autonomous and `pull_owns_dispatch` consulted all seven, so on a cron-driven agent the pilot flag is inert and #1766's soak measures nothing.

It was inert **silently**. The gate sits behind an `overflow_policy == "queue_persistent"` short-circuit, so for a `schedule` row the predicate is *never called at all* and the row takes the push path indistinguishably from the flag being unset — while §9 M1 told the operator that a pushed autonomous row means "check the backend actually has `PULL_MODE_PILOT_AGENTS` in its env", sending them after a variable that is present and correct.

## Scope

**Option 1** per the decision recorded on the issue: make the system honest about the reach it has, rather than extend the reach. Option 2 (giving the scheduler producer a `queue_persistent` policy so cron can be claimed) stays deferred — it changes dispatch topology on the reliability spine and needs its own review.

## One correction to the issue's analysis

The issue says only `agent` is reachable. The reachable set is **`{agent, event}`**: #1578's task-completion loopback POSTs to the same `/api/agents/{name}/task` route and is tagged `event` by `_derive_task_trigger` through an internal-secret-gated `X-Event-Trigger`. `reserved_event_dispatch` only affects a downstream tag, so there is no capacity bypass — it flows through the same `queue_persistent` acquire. Stranded set is exactly `{schedule, webhook, loop, fan_out, reminder}`.

## Changes

**`PULL_REACHABLE_TRIGGERS = {agent, event}`** names what dispatch can deliver, and `pull_owns_dispatch` **intersects** the autonomous set with it — a narrowing of the single source of truth, not a second copy, so dropping a trigger from `_AUTONOMOUS_TRIGGERS` still drops it here.

The intersection is a **runtime no-op**: the predicate is only ever consulted on the one producer that can emit `{self_task, agent, mcp, manual, event}`, so no live verdict changes. It stops the predicate *claiming* reach; it does not alter behaviour. Pinned by `test_narrowing_is_a_runtime_no_op_on_the_only_producer_that_consults_it`.

**`note_unreachable_pull_trigger`** makes a stranded row distinguishable from an ineligible one — the acceptance criterion the old code failed in both directions. Design notes:

- It sits at the `reject` producer because that is the only place `triggered_by` is in scope; `capacity_manager` cannot report it, since that path passes no `overflow_payload`.
- **One line per (agent, trigger) per process.** It runs on every cron fire; an un-deduped warning would emit thousands of identical lines a day and train operators to filter out the one signal meant to be noticed.
- **A log line, not a metric or operator-queue item.** This is a property of the build, constant until the topology changes, so alerting on it would fire forever on any cron-driven pilot.
- Fail-safe: never raises, so a diagnostic cannot interfere with dispatching a real execution.

**Docs (`PULL_MIGRATION_TESTING.md` §9)** — M1 now expects `agent`/`event` ~100% pulled and *everything else 100% pushed as correct behaviour*, followed by the producer/reach table and a per-case reading: a stranded trigger is expected (grep `[#2048]`), while a pushed `agent`/`event` row is the only case where the old "check the env var" advice applies. Pre-flight gains a pull-eligible-volume query and states plainly that a cron-only agent cannot be piloted — checked *before* the flip rather than after a week of soak.

## Two existing assertions that encoded the missing reach

`test_1766_pull_pilot_exclusive.py` had:

- `test_pilot_owns_every_autonomous_trigger` — parametrized all seven and asserted `True`. It passed only because it called the predicate directly, outside the context that constrains it. Split into the two that can be delivered and the five that cannot.
- `test_pilot_autonomous_work_bypasses_admission` — drove a `queue_persistent` acquire carrying `schedule`, a combination production cannot produce. Now driven with `agent`, the shape that actually occurs, so the fix is asserted over real traffic instead of an impossible one.

## Guard against recurrence

`test_the_producer_topology_this_fix_rests_on_still_holds` asserts the three producers' `overflow_policy` literals. Implementing Option 2 later fails it with a message naming `PULL_REACHABLE_TRIGGERS` and §9, forcing both to be revisited in that same change. A silently-stale reach set is precisely this bug.

## Acceptance criteria

- [x] A decision recorded on which option (or both, sequenced)
- [x] `pull_owns_dispatch` no longer implies reach it does not have
- [x] §9's M1 expectation and its troubleshooting line corrected
- [x] A pushed row that should have been pulled is distinguishable from one never eligible
- [x] Pilot-selection guidance states that cron-only agents cannot be piloted

## Verification

`test_2048_pull_pilot_reach.py` (35 cases) + the pull/capacity suites green:

```
86 passed   test_2048 + test_1766 + test_1081_pull_mode + test_capacity_manager
```

Full backend unit suite running; will report. No behaviour change on a non-pilot agent (empty allowlist is the default and is asserted inert).

Fixes #2048


vybe (approved):
Validated via /validate-pr. The narrowing is verified a runtime no-op over the only queue_persistent producer (pinned by test), the once-per-(agent,trigger) dedup keeps the diagnostic from becoming noise, and the structural guard on the three producers' overflow_policy literals forces PULL_REACHABLE_TRIGGERS + §9 to be revisited if Option 2 ever lands — the right shape for a fix whose bug was a silently-stale claim. Ran the touched unit files locally: 58 passed pre-merge; 189 passed after resolving the tests/registry.json overlap with #2056 (both entries kept — resolution merge pushed). Patched the body to Fixes #2048.

PR #2062, opened 2026-08-08
docs(architecture): reconcile Invariant #8 with the ent#293 gate fix (#1890, #1821)

trinity-ability:
Docs only. Follow-up I committed to while approving #1890, raised there rather than pushed to that branch because eight in-flight PRs touch `architecture.md` and batching avoids compounding the conflict.

## Why this is a defect, not tidying

#1890 moved the agent-principal rejection **into** `require_admin`/`assert_admin`. Invariant #8 still described the pre-fix model:

> **Role ≠ human (#1644, #1816):** `assert_admin`/`require_admin` answer *what role*, never *is this a human* … An endpoint whose blast radius is operator-scale … needs `reject_agent_principal` **in addition to** the role gate

Both halves are now false, and the second is actively harmful: it is exactly the bolt-it-on-per-endpoint pattern #1890 eliminated after five occurrences of the class. Left standing, the doc instructs the next contributor to re-implement what was just centralised — on a **security** invariant, in the file `CLAUDE.md` Rule 7 tells everyone to read before adding an endpoint.

The helper list in the same invariant also still described `assert_admin` as `_reject_connector_principal` + role only.

## What changed

- **The `Role ≠ human` paragraph → `An admin gate is never agent-callable (ent#293/ent#297)`.** States the gate as the mechanism; keeps the *reason* (an agent key resolves to its owner carrying the owner's role) since that is what makes the design non-obvious; records that per-endpoint `reject_agent_principal` calls are now belt-and-braces rather than the mechanism.
- **Records the two re-opened reads.** `POST /api/monitoring/agents/{name}/check` and `GET /api/subscriptions` were *uses*, not grants. Notes specifically that the subscription read is **owner-scoped, not un-gated** — un-gating was the first attempt and would have handed any `role=user` a fleet-wide `owner_email` + agent-name enumeration oracle, reintroducing the Invariant #8 disclosure class via a fix for a different disclosure. That trap is worth being written down.
- **Writes down the new hard rule:** never `require_role("admin")`. It rejects connector but not agent principals, and `require_role` stays permissive on purpose (ghost spawning goes through `require_role("creator")`, ent#69 Part 2). Until now this lived only in a docstring and an AST guard — findable after you have already made the mistake.
- **`PUT /api/agents/{name}/label` row** updated for #1821: `label` is required-but-nullable with `extra="forbid"`, so `{}` and unknown field names 422 instead of silently clearing.

## Verification

- Both statements checked against the merged code, not the PR bodies: `dependencies.py` `require_admin`/`assert_admin` both call `reject_agent_principal`; `monitoring.trigger_health_check` is `get_current_user` + `AuthorizedAgentByName`; `subscriptions.list_subscriptions` branches on `role == "admin" and not agent_name`.
- Ran the `enterprise-docs-guard` pattern against the added lines — clean, no paid-feature or `enterprise_*` token.
- No code, schema, migration, or behaviour change; nothing for `/validate-config` or `schema-parity` to act on.

Fixes are already merged (#1890, #1822); this carries no closing keyword by design — it documents them rather than resolving a separate issue.


vybe (approved):
Verified against merged code: `require_admin`/`assert_admin` both call `reject_agent_principal`; `trigger_health_check` is `get_current_user` + `AuthorizedAgentByName`; `list_subscriptions` branches on `role == "admin" and not agent_name`. Docs-only, no closing keyword by design. Approving.

PR #2063, opened 2026-08-08
chore(enterprise): advance the submodule pin to the current private-repo head

trinity-ability:
Advances `src/backend/enterprise` from `e898cef8` to the private repo's current head `c2cfeb3f`. Final step of the ent#157 sequence; one gitlink, no other files.

## Why

The pin had drifted **12 commits** behind. The pointer is what a mounted install actually checks out, so a stale pin silently caps which private modules an entitled deployment can reach — anything registered after the pinned commit resolves to nothing and its gated routes answer 403, with no error anywhere and green CI. Several of the commits this activates are security fixes, so the drift is not cosmetic.

Clean fast-forward: the old pin is an ancestor of head (12 ahead, 0 behind), no divergence to reconcile.

## Sequencing — satisfied by construction

#1628 originally carried a bump to `740063e3d`, a commit that existed only on the companion PR's unmerged branch: 6 behind that branch's own tip and **diverged** from the private head (+5/−7). Merging it would have pinned dev to unmerged code missing five commits of private-repo work. That bump was reverted before #1628 merged, with the pointer move deferred to its own PR once the companion landed. It has landed, so this is that move.

This branch is cut from **#1628's merge commit** (`fb80519a`), so the pin cannot reach `main` ahead of the OSS seam it depends on. That ordering is the whole point: a `dev → main` cut carrying the bump without #1628 would reproduce the exact 403-forever state this fixes. Please keep them together through the release cut.

## Verification

Full private-repo suite, run against this dev at both pins in an identical environment:

| | failed | passed |
|---|---:|---:|
| old pin `e898cef8` | 32 | 196 |
| new pin `c2cfeb3f` | **30** | **376** |

**Zero new failures**, two fixed, 180 additional passing tests. The 30 that persist fail identically at both pins and are environmental — 16 are a missing optional `pyotp` dependency — so they are not introduced here. I diffed the failure sets by name rather than comparing counts, since equal counts can hide a swap.

Also confirmed the companion module's own suite passes against post-#1628 dev (35 passed), which was not possible before #1628 provided the OSS seam, and that its agent-scoped tables are genuinely registered for delete/rename cascade — commenting the registration out fails 4 tests including the rename and delete cases.

## Risk

Low and reversible — a one-line gitlink revert. OSS clones are unaffected: both submodules are `update = none`, so a plain `git submodule update --init --recursive` skips them and no external contributor sees a change. CI does not mount the submodule either, which is exactly why this class of drift needs checking by hand at release.


vybe (approved):
Verified independently of the PR body: `e898cef8...c2cfeb3f` is 12 ahead / 0 behind, and `c2cfeb3f` is the current head of `trinity-enterprise/main`. Clean fast-forward to tip, one gitlink. Approving.

PR #2071, opened 2026-08-09
fix(schedules): legacy IANA timezone aliases resolve end-to-end — images, validator, and scheduler classification (#1823)

### LINKED ISSUE #1823: bug: legacy IANA timezone aliases (Europe/Kiev et al.) crash schedule creation with a 500 — pytz accepts, APScheduler's zoneinfo lookup can't resolve them inside the shipped image

trinity-ability:
Fixes #1823

A schedule created with a legacy IANA timezone alias (`Europe/Kiev`) returned a raw **500**. Reproducing it turned up **five** defects, not the three the issue describes — and the fifth is the one that made the symptom permanent.

## The five defects

| # | Layer | Defect |
|---|---|---|
| 1 | **Image packaging** | `docker/{backend,scheduler}/Dockerfile` are `FROM python:3.13-slim` with **no codename pin**, so both images silently migrated bookworm→trixie. Debian 13 split the IANA *backward-compatibility links* out of `tzdata` into a separate `tzdata-legacy` package, so `/usr/share/zoneinfo/Europe/Kiev` simply stopped existing — without anyone changing a line. |
| 2 | **Validator robustness** | `zoneinfo.ZoneInfoNotFoundError` is a **`KeyError` subclass**, so it escaped `validate_cron_expression`'s `except (ValueError, TypeError)` and surfaced as an unhandled 500. The validator also probed **pytz only** — but pytz bundles its own complete database and accepts every alias everywhere, while APScheduler's `astimezone()` re-resolves through `zoneinfo`. #1472 mirrored the scheduler's *visible first step* and missed the re-resolution behind it. |
| 3 | **Env-dependent test** | `tests/unit/test_1472_schedule_validation.py` carried `Europe/Kiev` as a valid fixture and passed on CI — because `tests/requirements-test.txt` declares `tzdata` (#1771) — while the same test failed inside the shipped image. |
| 4 | **Update path** (from @webmixgamer's comment) | The `PUT` route calls `validate_timezone` **only**, and that function was pytz-only — so an unresolvable alias was **silently persisted with 200 OK** instead of 500ing. Same contract, opposite failure mode: create was loud and wrong, update was quiet and wrong. |
| 5 | **Found by reproduction — NOT in the issue** | `src/scheduler/service.py::_add_job` classified `ZoneInfoNotFoundError` as **transient**. Because it is a `KeyError` subclass it is not a `ValueError`, so it fell through the permanent arm into the broad `except Exception`. The row was never snapshotted into `_permanent_add_failures`, so #1472's bounded-retry machinery was **defeated**: it retried **every 60 s forever**, logged an ERROR on every sync tick, and froze `next_run_at` into exactly the "Next: Nd ago" symptom #1472 existed to eliminate — which **canary E-06 fires on every cycle**. |

Defects 2 and 4 are two halves of one split-brain: the create route calls only `validate_cron_expression`, the update route only `validate_timezone`, so two probes meant two contracts. The fix **delegates** — `validate_cron_expression` now calls `validate_timezone` — leaving exactly one place where the timezone contract is defined.

## Why it hid for months

**CI was more capable than production.** #1771 added `tzdata` to `tests/requirements-test.txt`, so the alias resolved on every CI runner. The images then quietly *lost* the same capability via the floating base tag. The test environment being strictly more capable than the runtime is the inversion that keeps a packaging defect green — and no unit test can detect it by construction.

## Proof, in the shipped images

Everything below is from the WAVE-4 `/verify-local --skip-agent` run (project `trinity-verify-424a80d3`) — **the only place AC 1 and AC 4 are provable**, since every unit assertion here is static or fault-injected.

**`tzdata-legacy` resolves on trixie, in BOTH images** (this was the single unverifiable claim in the plan, and it gated both builds):

```
backend   layer #7:  Get:225 http://deb.debian.org/debian trixie/main arm64 tzdata-legacy all 2026b-0+deb13u1 [183 kB]
                     Setting up tzdata-legacy (2026b-0+deb13u1) ...
scheduler layer #33: #33 7.237 Get:23 ... tzdata-legacy all 2026b-0+deb13u1 [183 kB]
                     #33 8.083 Unpacking tzdata-legacy (2026b-0+deb13u1)
```

No `Unable to locate package`, no `E:` — the recorded apt-fails fallback was **not** needed and **not** applied.

- `zoneinfo.ZoneInfo('Europe/Kiev')` → `Europe/Kiev`, exit 0, in **backend and scheduler**
- backed by the real on-disk symlink `/usr/share/zoneinfo/Europe/Kiev -> Kyiv`
- `dpkg -l tzdata-legacy` → `ii … 2026b-0+deb13u1` in both
- `pip show tzdata` → **2026.3** (matches the exact pin)
- **in-image pytest**, `test_1472_schedule_validation.py` + `test_1823_tz_capability_parity.py` run *inside the backend image* — the exact run that originally surfaced the bug — **55 passed**; the named detector `test_runtime_resolves_legacy_tz_aliases` passed
- **scheduler container boots healthy and registers 3 jobs** under `APScheduler==3.11.0`: `state=running health=healthy restarts=0`, `GET /health` 200; driving the real `_add_job` chain (pytz → `CronTrigger` → `astimezone()` → `zoneinfo`) in-image gives `Europe/Kiev` → `next_run=2026-08-10 09:00:00+03:00`, `Europe/Kyiv` identical, `UTC` → `+00:00`
- negative: `Not/AZone` → `ScheduleValidationError: Unknown timezone`, which the router maps to **400, not 500**; a fault-injected empty TZPATH + hidden wheel yields the **distinct named** error rather than a raw `KeyError`

## Two deliberate changes, each independently droppable

Both were flagged at implement time as scope questions and both were **ruled KEEP on technical merit at review** — recorded here so a reviewer can disagree cheaply.

**`ac2a7e8c` — a cron-only `PUT` returns 400 instead of 200.** The route validated the new cron against a hardcoded `"UTC"`, so a row already storing an unresolvable zone was mutated and re-blessed with 200 OK while the scheduler still could not register it. It now validates against the row's **effective** zone. This is the PR's one behaviour regression, and it is bounded by "the row is already non-firing" — a bound that **now holds cleanly** only because `783c3120` closed the `""` hole below. Cron verdicts are unchanged (field parsing is timezone-independent, pinned by the branch's own bad-cron and healthy-row tests). Post-packaging-fix the branch is unreachable: 0 of 596 pytz zones are unresolvable, measured.

**`ac644486` — `APScheduler>=3.10.0,<4.0.0` → `==3.11.0` in `docker/scheduler/requirements.txt`.** The implement stage stated plainly that this is *not required for the fix to work*, and the orchestrator leaned DROP. That lean was **wrong**: the divergence is **current state, not hypothetical** — the floor resolves to **3.11.3** on PyPI today while the backend pins **3.11.0**, so building both images right now yields a *validator and an executor running different copies of the parser* that #1472's contract is stated against. `astimezone()` is the exact function that caused this bug. Now guarded by a parity test rather than left to a comment — closing a gap the commit's own comment declared ("Nothing ties these two declarations together yet").

## The falsy-evaluation fix (`783c3120`)

An independent Gemini review claimed the effective-timezone expression `updates.timezone or schedule.timezone or "UTC"` was wrong for an explicit empty string. It was **reachable**, and proven so by driving the real handler rather than by reading it: `ScheduleUpdateRequest.timezone` is a bare `Optional[str]` with no validator, `""` clears Pydantic, `validate_timezone("")` early-returns by design, `model_dump(exclude_unset)` carries `timezone: ""`, and `db/schedules/crud.py::update_schedule` writes it verbatim with no falsy guard. **Measured:** a `PUT {"cron_expression": …, "timezone": ""}` handed the validator `Europe/Kiev` while persisting `""`.

The harm **inverts `ac2a7e8c`'s own justification**: on a tz-incomplete runtime the new 400 rejected the *one request that repairs the row* — clearing a stale alias back to the scheduler default. Fixed with presence-not-truthiness (`updates.timezone if updates.timezone is not None else schedule.timezone`, then `or "UTC"`), with **2 regression tests verified RED against the pre-fix expression**. The create path's `or "UTC"` is deliberately untouched — it has no middle fallback term, so `"" or "UTC"` already equals the effective value.

## Coverage — stated honestly

**The integration stage did not run.** `verify-local`'s own `result.json` records `status="fail"`. The integration stage aborted at **collection** after 0.45 s (`Interrupted: 5 errors during collection`), so the suite **never executed** and this branch has **zero integration-level evidence beyond boot+health**. This is an **absence of coverage, not a green.**

The cause is pre-existing **#2037**, proven branch-independent rather than assumed: `tests/utils` is a **tracked symlink** → `testing_utils` (mode `120000`, blob `4e7b2eaf`) **present at `origin/dev`**, which shadows `src/backend/utils`, and `testing_utils` has no `safe_yaml.py`. All 5 failing files have **0 diff lines** versus base, and the 10 backend modules importing `utils.safe_yaml` are **disjoint** from the 10 files this branch touches.

**No agent-side verification was performed** (`--skip-agent`). That is acceptable only because `docker/base-image/**` is untouched — verified, `git diff --name-only ef3b0ebf..HEAD -- docker/base-image/` returns **0 files** — and the agent container performs no schedule validation and registers no `CronTrigger`. It is still a stated gap.

Stage results: preflight **pass** · unit **pass, 8466 passed / 18 skipped / 0 failed (478 s)** · build+import-smoke **pass (84 s)** ← the gating stage · agent-build **skipped** · boot+health **pass** · agent-exercise **skipped** · integration **fail (collection abort, above)**.

## Tests

| Run | Result |
|---|---|
| Full unit suite, in the verify venv (WAVE 4) | **8466 passed, 18 skipped, 0 failed** (478 s) |
| Full unit suite, post-review (WAVE 3) | **8344 passed, 22 skipped, 0 failed** (480 s) |
| In-image, backend image: `test_1472` + `test_1823_tz_capability_parity` | **55 passed** |
| `test_1472_schedule_validation.py` + `test_1823_*.py` (ship re-confirm) | **72 passed** |
| Schedule neighbourhood `-k "schedule or template_schedules or compatibility"` | 565 passed, 9 skipped |

Every new test was proven non-vacuous by reverting its fix and observing the red: packaging guard without `tzdata-legacy`/`tzdata` → 2 failed; update-route test against the pre-fix validator → 1 failed; scheduler classification against the unfixed service → 2 failed; cron-only `PUT` against the unmodified router → 3 failed; the falsy-eval pair → red against the pre-fix expression.

The fault-injected tests carry an explicit docstring disclaiming environmental proof — with the packaging fix in place there is **no natural input** that reaches the `ZoneInfoNotFoundError` backstop (0/596 zones unresolvable), so those tests prove *our exception-type mapping*, not that any environment is broken. The environmental claims are carried by the in-image checks above and by `test_1823_tz_capability_parity.py`.

## Follow-ups — flagged here, deliberately NOT filed as issues

1. **The base tags float — and that float IS this bug's recurrence vector.** `docker/{backend,scheduler}/Dockerfile` are on `python:3.13-slim` while `docker/base-image/Dockerfile` pins `-bookworm`. Extend the #1891 guard to assert all three Dockerfiles agree on a **codename**.
2. **There is no smoke job for the scheduler image at all.** `backend-image-smoke.yml` has no scheduler counterpart, so this PR's scheduler-only surface (`docker/scheduler/*`) has **zero CI detector** — WAVE 4's in-image run was the sole gate.
3. `backend-image-smoke` is **advisory, not required** — a red run does not block merge.
4. `tests/requirements-test.txt` floors `apscheduler>=3.11.0`, so **CI resolves 3.11.3 while both images pin 3.11.0**. The new parity test closes backend↔scheduler; the CI↔image skew remains.
5. `src/scheduler/pyproject.toml` still floors `apscheduler>=3.10.0` — inert for the image, diverges for local editable dev installs.
6. **tzdata pin asymmetry**: backend `tzdata==2026.3` vs scheduler `tzdata>=2024.1` (each file keeps its own convention). Both resolve to 2026.3 today, but the parity test guards *declaration presence*, not version equality, so a future scheduler rebuild can drift.
7. `update_schedule` has no access-first gate, unlike create's #1445 `assert_agent_access` (effectively unreachable — schedule ids are `secrets.token_urlsafe(16)`).
8. `Europe/Kyiv` is missing from the UI timezone dropdown (`SchedulesPanel.vue` ships 7 hardcoded modern names, none of them aliases — which is why this is P3). Explicitly deferred by the issue to a `theme-ui-ux` follow-up.
9. **#2037 makes `verify-local`'s integration stage structurally unusable on this host**, and `known-false-fails.txt` **cannot** rescue it — that mechanism only `--deselect`s and cannot survive a collection abort. Deserves a priority bump.
10. `tests/unit/test_1472_schedule_validation.py` is still absent from `tests/registry.json` (pre-existing; the four new `test_1823_*` files are registered).

## Merge note

This PR edits `docs/memory/architecture.md`. Sibling PR **#1897** also edits it, **inside a single ~4700-character line (L730)**. On conflict, **hand-merge both sides' sentences into that one line — never `--ours`/`--theirs`**, since each side is a whole paragraph of unrelated content.

No schema change (Redis/config only), so the dual-track Alembic / `schema-parity` question is moot. No `src/frontend/` change, so no `ui` label and `frontend-e2e` is not engaged.


vybe (approved):
**Root cause reproduced in Docker rather than taken on trust:**

```
python:3.13-slim  ->  Debian GNU/Linux 13 (trixie)
/usr/share/zoneinfo/Europe/Kiev  ->  No such file or directory
zoneinfo.ZoneInfo("Europe/Kiev")  ->  ZoneInfoNotFoundError
+ apt-get install tzdata-legacy  ->  Europe/Kiev, Europe/Kyiv, Asia/Calcutta, US/Eastern ALL RESOLVE
```

The tag carries no codename pin, so the image migrated bookworm->trixie with nobody changing a line. That is the entire bug, and this fixes it at the layer where it actually happens rather than papering over the 500.

Also verified:
- APScheduler is pinned **identically** (`3.11.0`) in `docker/backend/Dockerfile` and `docker/scheduler/requirements.txt` — validator and executor can no longer resolve different versions, which is the whole point of that exception to the floor convention.
- `tzdata-legacy` is in the backend and scheduler images and correctly ABSENT from the agent base image, which resolves no schedule timezones.
- **The parity guard bites**: deleting `tzdata-legacy` from `docker/backend/Dockerfile` fails `test_1823_tz_capability_parity` on two assertions; restoring it goes green. That is what stops the next image-slimming pass from silently re-opening this.
- 72 tests pass across the four #1823 files plus `test_1472_schedule_validation`; `tests/registry.json` updated.
- `merge-tree` vs current `dev`: 0 conflicts.

PR #2072, opened 2026-08-09
fix(canary): a rejected Slack post is undelivered, and gets retried (#1897)

### LINKED ISSUE #1897: bug: a failed canary Slack post is counted as delivered and never retried

trinity-ability:
Fixes #1897

## What the bug was

`CanaryAlerts.emit_transition` was `-> None` and dropped `post_webhook`'s `success` on the floor, so the caller read **"did not raise"** as **"delivered"**. A rejected Slack post was therefore swallowed: the transition was counted in `cumulative_transitions`, the cycle cursor advanced, and `_is_green_to_red` then classified the same invariant as *continuing-red* on every later cycle — because `previous_latest` now carried the row this cycle had just inserted. Continuing-red gating guaranteed the alert was **never** retried. The alert was unrecoverable while the counter reported it sent.

Two independent faults: the outcome was discarded, and there was no per-invariant delivery state, so nothing *could* re-attempt.

## The issue's own proposed fix is rejected — and its stated reason for rejecting the alternative was also wrong

The issue proposes withholding the cycle-global `canary:last_cycle_at` cursor on a failed post, and rejects the alternative on the grounds that it would cause a fleet-wide alert storm.

Walked mechanically, **both halves are wrong**:

- Holding the cursor produces **no retry**. The invariant's own freshly-inserted violation row already post-dates the withheld cursor, so `_is_green_to_red` reads it as a continuation regardless — the cursor cannot express *per-invariant* delivery.
- It also produces **no storm**, for the same reason. The predicted failure mode cannot occur.
- What it *does* do is silently swallow a genuine unrelated red→green→red flip that lands in the withheld window.

So "don't touch the cursor" is upheld — on a **corrected** reason. The fix is instead an explicit per-invariant pending-alert mechanism: a new Redis hash `canary:alert_pending`, field = invariant id.

## The load-bearing ordering: arm BEFORE the POST

The pending entry is written **before** `await post_webhook(...)`, not on failure. This is not stylistic.

`issubclass(asyncio.CancelledError, Exception)` is `False`, and `stop()` cancels a live cycle on lifespan shutdown. A SIGTERM landing inside the ~5s webhook await propagates straight out: the `except Exception` never runs, nothing is armed, the cursor write never happens, and the next cycle sees a continuation. That is #1897 verbatim — **on every deploy that coincides with a red cycle**. Arming first survives cancellation, crash, OOM and SIGKILL alike, and subsumes the raise case (under arm-on-failure, an `emit_transition` that raises on its *first* call has no entry to arm and degrades to exactly the bug).

**Proven load-bearing, not argued:** removing the pre-await `_write_pending_alert(...)` and leaving only the arm-on-FAILED fallback turns **exactly one** test red — `test_cancellation_mid_post_leaves_the_alert_armed` — while the other 25 still pass. It has exactly one witness, and that witness is real.

## The corrected bound: 8 POSTs over 2100s, not 6 over 30 min

The plan's headline claimed "≤6 POSTs over ≤30 minutes". The real bound is **8 POSTs spanning 2100s (35 min)** per contiguous failure run at the 5-minute default.

The budget is evaluated **after** each attempt, so a run ends on the first attempt whose age *exceeds* `MAX_ALERT_PENDING_AGE_SECONDS` (1800), not the last one inside it. **The behaviour is kept and the claim corrected** — deliberately: giving up *after* a demonstrated failure is what lets the give-up ERROR quote a real elapsed and a real `last_error` rather than a stale one. Corrected in three places (`architecture.md`, requirements §31.1, and the constant's own comment).

Pinned by a test that **derives** the cycle count from the constant (`MAX // interval + 2`) rather than hard-coding it, so the pin cannot rot away from the bound it guards.

## The slow-flapper dual — measured, not argued

A 40-cycle probe:

| scenario | POSTs | give-ups |
|---|---|---|
| continuously red, dead webhook | 8 | 1, then silence |
| flap at 3× interval | 14 | 3 |
| flap at **4× interval** | **10** | **0** |

The 4× case looks like "never gives up", and it is — but it **cannot arise inside one red episode**: the retry floor is one interval, so intra-episode gaps are always exactly 1. Reaching 4× requires the invariant to go **green** in between, which makes each red return a *fresh detection* making exactly one POST and zero retries. So **POST rate == detection rate == pre-#1897 behaviour**. No amplification; nothing to stop trying.

"Never gives up" is the intended **dual** of the 3-interval run-decay, which exists so a brief flap cannot spend the window a later long outage needs. A delivery-layer budget deliberately does not bound its own detector.

**Residual, documented not silent:** a slow flapper produces no give-up ERROR (only the per-episode WARNING) — a visibility nuance of AC #5.

## ⚠️ API CONTRACT CHANGE — reviewer confirmation requested

**`RunCycleResponse.transitions` now means "delivered", not "detected".** This is a behavioural change to an **existing** field, not an addition.

`cumulative_transitions_detected` was split out to preserve the detection metric, and `undelivered_invariant_ids` rides beside `transitions` so a webhook outage reads as an outage rather than as a green cycle (without it, an outage would render as zero transitions — indistinguishable from green, reintroducing the exact ambiguity #1881 spent a 409 to remove).

👉 **Please confirm no external consumer reads `transitions` as a detection count.** This is the one thing in the change that something outside the repo could already be reading the other way.

## In-image proof (verify-local)

Not just source — confirmed **in the shipped artifact** via `compose run` against the freshly built backend image:

- `import services.canary_service, services.canary_alerts` → **IMPORT OK**
- `REDIS_KEY_ALERT_PENDING` → **`canary:alert_pending`**
- `MAX_ALERT_PENDING_AGE_SECONDS` → **1800**

`canary:alert_pending` is correctly **outside** `CLEARED_KEYSPACES` — it is global, keyed by invariant id, and invariant ids are a fixed code registry, not recyclable agent names. Redis key-parity test: 4 passed (correct *by construction* — the guard regex is `["']agent:([a-z_]+)`, so a `canary:`-prefixed literal is out of scope; registering it would be wrong).

**No schema change ⇒ the Alembic / `schema-parity` dual-track question is moot.** This is Redis-only, no migration.

Stages: preflight pass · unit **8434 passed, 18 skipped, 0 failed** (437s) · build+import-smoke pass (27s) · boot+health pass, all 7 containers healthy · integration **FAIL** (see below).

## ⚠️ HONEST COVERAGE STATEMENT — read this before approving

`verify-local`'s `result.json` says **`status="fail"`**. The **integration stage aborted at collection in 0.42s** — the suite **never executed**.

Cause is pre-existing **#2037** (`tests/utils` shadows `src/backend/utils`, so `utils.safe_yaml` is unimportable), proven **branch-independent four ways**:

1. the failing set is **exactly** the previous run's 5 files;
2. each has **0 diff lines** vs `origin/dev`;
3. `tests/utils` is a tracked symlink at blob `4e7b2eaf41b5…`, **byte-identical** on branch and base;
4. the 10 modules importing `utils.safe_yaml` are **disjoint** from the 10 files this branch touches.

**State this as an ABSENCE of integration coverage, not as a green.** Zero integration tests executed against this branch. And note specifically: **two of the five uncollected files — `test_1560_breaker_lifecycle` and `test_monitoring_service` — sit near canary/monitoring territory**, so regressions on those paths are unproven locally and rest on CI.

Agent stages were skipped (`--skip-agent`), acceptable only because `docker/base-image/**` is untouched — but stated as a gap.

## A test that could not be run in CI

The rewritten `tests/test_canary_invariants.py::TestCanarySlackEmit` case (the test that previously **asserted the bug** — it asserted a rejected POST was counted) now asserts the post is undelivered and armed.

That file is **CI-invisible** (CI runs `cd tests && pytest unit/`) **and** un-runnable on the clean base (#2037 collection error). It was verified **locally** via a temporary path-shadow workaround and proven **non-tautological** — checking out base's two source files turns it red. But say it plainly: **this specific edit is not covered by CI.**

The 26 new tests in `tests/unit/test_1897_canary_alert_delivery.py` *are* CI-visible, and **18 of the 26 fail against the pre-fix code**.

## Tests

- `tests/unit/test_1897_canary_alert_delivery.py` → **26 passed** (new; 18 red against pre-fix code)
- 5-file canary neighbourhood (1897 + 1880 + 1881 + 1987 + 1813) → **134 passed, 0 failed**
- Full unit suite (verify-local, in the verify venv) → **8434 passed, 18 skipped, 0 failed**

## Follow-ups — flagged, deliberately NOT filed

- **R7** — a persistently failing `insert_canary_violation` freezes `previous_latest` while the cursor advances, so `_is_green_to_red` returns `True` **every** cycle: i.e. **the detector lies**. Pre-existing; unchanged by this PR; it is *why* the delivery bound cannot be absolute.
- **R9** — all four routes in `routers/canary.py` carry `require_admin` **without `reject_agent_principal`**. Per architecture.md Invariant #8 an agent-scoped key resolves to its owner *carrying the owner's role*, so on a default admin-owned install any non-ephemeral agent's injected `TRINITY_MCP_API_KEY` satisfies the gate — including `POST /run-cycle`, which drives the alert sink. Documented #1644/#1816 class.
- **#2037** — the canary suite is both CI-invisible and un-runnable on the clean base. Deserves a priority bump.
- A **poisoned `last_attempt_at`** in a pending record re-anchors the run every cycle, so the age budget never accumulates → 1 POST/interval while red. Requires Redis write access; rate equals today's.
- **R2** — a retried alert renders as "first red" for something that has been red for hours.

### Accepted residuals

- Two workers can both retry one pending entry → at most one duplicate message. #1881's posture on this same subsystem decides it: **choose the duplicate over the silence.**
- AC #4's bound is conditional on Redis being readable/writable (everything fails **open** to exactly the pre-#1897 behaviour, never worse).
- The pending hash carries **no TTL**: bounded by the invariant registry (≤16 fields ≈ 5.6 KB ceiling) and reaped by both the success and give-up `HDEL` paths. A TTL would silently discard every pending alert after one quiet hour — which is the exact silent loss the key exists to prevent.

## ⚠️ Merge note

This PR edits `docs/memory/architecture.md` **L730 — a single ~4700-character line**. Sibling PR #1823 also edits `architecture.md`.

**On conflict: HAND-MERGE both sides' sentences into the one line — never `--ours` / `--theirs`.** Each side is a whole paragraph of unrelated content, so taking either wholesale silently deletes the other's work.


vybe (approved):
Verified:
- `test_1897_canary_alert_delivery` plus the four sibling guards this change could plausibly break (#1880 alert parity, #1881 leader lease, #1813 H-01, #1987 instance label): **134 passed**.
- `merge-tree` against the post-#2074 `dev`: **0 conflicts** — the architecture.md edits land in different sections.
- `CANARY_SLACK_WEBHOOK_URL` is pre-existing and already wired into docker-compose.yml, docker-compose.prod.yml and .env.example, so no #1056-class packaging gap.
- No secret exposure: the persisted error string is Slack's body or `type(e).__name__`, and the webhook URL — which IS the credential — is never in it. Truncation at 200 chars is applied before persistence.

Two design calls checked rather than accepted:
- The three-state `AlertDelivery` is right. A bool folding SKIPPED into False would arm a retry for every transition on a correctly-configured *silent* sink (the default — `CANARY_SLACK_WEBHOOK_URL` unset), burn the budget and end in a give-up ERROR: a new alarm-fatigue bug shipped inside the fix for an alarm-loss bug.
- `Enum` over a fieldless dataclass avoids the #1085 `__eq__` footgun architecture.md already records.

Note, unrelated to this PR: `tests/test_canary_invariants.py` throws 20 setup errors locally (`ModuleNotFoundError: No module named 'utils.safe_yaml'`) — reproduced identically on plain `origin/dev` (152 passed / 2 skipped / 20 errors both sides), so it is pre-existing local path shadowing, not a regression here. Worth its own issue.

PR #2074, opened 2026-08-09
ci: Alembic single-head guard — one head per version-line, both tracks (#2068)

### LINKED ISSUE #2068: bug: enterprise submodule pin advance lands two Alembic heads — entitled instances silently degrade to OSS-only
## Summary

Advancing the `src/backend/enterprise` submodule pin in #2063 lands an enterprise revision whose Alembic migration graph has **two heads**. `register_enterprise()` runs `alembic upgrade head` at boot, which raises `CommandError`, so registration aborts and the backend **silently continues OSS-only**.

The result on an entitled instance: the backend boots, `/health` returns `healthy`, every container is green — but `GET /api/version` reports `edition: "oss"` with `enterprise_features: []`. All 13 enterprise features (2fa, audit, client_portal, ephemeral_agents, permissions_matrix, retention, shared_sessions, siem, skill_runner, slack_per_agent_bots, sso, telemetry, user_management) become unavailable, and their gated routes start answering 403 with nothing obviously wrong at the platform level.

This is the exact failure mode #2063's own commit message set out to prevent ("their gated routes answer 403 with no error anywhere, on green CI") — reintroduced by the pin advance itself, and again on green CI.

## Component

Backend — enterprise registration / Alembic migration runner. Secondarily CI (no guard validates the migration graph of a pin advance before it merges).

## Priority

P1 — every enterprise feature is unavailable on any entitled instance that deploys this commit, with no platform-level signal beyond one ERROR line at boot. Workaround exists (pin the submodule back), but it is undone by the next update.

## Error

```
Trinity Enterprise registration FAILED — continuing OSS-only: CommandError(
  "Multiple head revisions are present for given argument 'head'; please specify a specific
   target revision, '<branchname>@head' to narrow to a specific head, or 'heads' for all heads")
    register_enterprise(app)            enterprise/backend/__init__.py:164
    upgrade_to_head(...)                enterprise/backend/_alembic_runner.py:163
    _inspect_and_upgrade(...)           enterprise/backend/_alembic_runner.py:138
                                        enterprise/backend/migrations/env.py:96
```

Post-boot state:

```json
{ "version": "0.8.5+g<sha>", "git_branch": "dev", "edition": "oss", "enterprise_features": [] }
```

## Location

- **Pin advance (public repo)**: commit `6ec2d74e` — `chore(enterprise): advance the submodule pin to the current private-repo head (#2063)`, moving `src/backend/enterprise` forward 12 commits
- **Failure site**: `enterprise/backend/_alembic_runner.py` → `upgrade_to_head()` / `_inspect_and_upgrade()`, called from `register_enterprise()` in `enterprise/backend/__init__.py:164`
- **Defective revision**: the newest revision at the new pin (the A2A control-plane module) declares `down_revision` pointing at the revision that was head *five* revisions ago, rather than the current head

## Root Cause

The enterprise migration chain already ran linearly:

```
0009_retention_opq_reports → 0010_skill_runner_acl → 0011_shared_sessions
  → 0012_slack_agent_bindings → 0013_cross_model_validation → 0014_client_blocks
```

The revision added by the newest enterprise commit (`0010_a2a_control`) sets `down_revision = "0009_retention_opq_reports"` instead of chaining onto the current head `0014_client_blocks`. With **no merge revision** present, the graph now has two heads:

| Pin | Migration graph | Heads |
|-----|-----------------|-------|
| previous | `0009 → 0010_skill_runner_acl → … → 0012` | 1 ✅ |
| new (this pin) | `0009 → { 0010_a2a_control ⊥ 0010_skill_runner_acl → … → 0014_client_blocks }` | **2 ❌** |

`alembic upgrade head` is ambiguous with two heads and raises `CommandError`. `register_enterprise()` catches it and degrades to OSS-only rather than failing the boot.

Note the duplicated `0010_` numeric prefix across two unrelated revisions — a useful smell that two feature branches were each cut from `0009` and only one was rebased onto the head before merge.

The OSS migration graph is unaffected and healthy (single head, `current == head`).

## Reproduction Steps

1. On an enterprise-enabled instance (private submodule populated at `src/backend/enterprise/`, backend run with the `docker-compose.prod.enterprise.yml` overlay), note the pre-update state: `GET /api/version` → `edition: "enterprise"`.
2. Pull to a commit at or after `6ec2d74e` and run `git submodule update --init --recursive` so the submodule advances to the new pin.
3. Rebuild and recreate the backend **with the enterprise overlay**: `docker compose -f docker-compose.prod.yml -f docker-compose.prod.enterprise.yml up -d backend`.
4. `docker logs trinity-backend | grep -i "Trinity Enterprise"` → `registration FAILED — continuing OSS-only: CommandError("Multiple head revisions are present …")`.
5. `GET /api/version` → `edition: "oss"`, `enterprise_features: []`, while `/health` still returns `healthy`.

Confirm the graph directly:

```bash
docker exec trinity-backend sh -c 'cd /app/enterprise && alembic heads'   # returns 2 heads
```

## Suggested Fix

**1. Repair the migration parentage (private enterprise repo).** Either re-parent the A2A control revision onto the current head:

```python
# 0010_a2a_control.py  (rename to follow the head, e.g. 0015_a2a_control)
down_revision = "0014_client_blocks"   # was: "0009_retention_opq_reports"
```

…or, if both branches must be preserved as authored, add an explicit merge revision (`alembic merge -m "merge a2a control" 0010_a2a_control 0014_client_blocks`).

**2. Add a CI guard on the enterprise migration graph (this repo).** #2063 was a clean fast-forward that merged on green CI — nothing validated that the resulting graph was still linear. A guard that fails the build when the pinned enterprise revision does not resolve to exactly one head would have caught this at PR time:

```bash
test "$(alembic heads | grep -c '(head)')" -eq 1 \
  || { echo "enterprise migration graph has multiple heads"; exit 1; }
```

This fits the "validation guards" charter of #1850. Worth running the same assertion against the OSS graph, since the failure mode is identical there.

**3. Consider making the degradation louder (optional, separate).** An entitled deployment losing all enterprise features currently surfaces as one ERROR line while `/health` stays green and the platform serves traffic. A boot-time failure of `register_enterprise()` on an instance where the submodule *is* populated (i.e. enterprise was clearly intended) is arguably a fail-closed condition, or at minimum should be reflected in the health payload so external monitoring can see it.

## Environment

- Trinity version: `0.8.5`, `dev` branch, at the commit that advances the pin
- Database backend: PostgreSQL (Alembic path); OSS graph at a single head, `current == head`
- Docker version: 28.3.3
- OS: Debian GNU/Linux 12 (bookworm)
- Edition: enterprise (submodule populated, overlay mounted read-only at `/app/enterprise`, `TRINITY_OSS_ONLY` unset)

## Related

- #2063 — `chore(enterprise): advance the submodule pin to the current private-repo head` (the pin advance that surfaces this)
- `docs/ENTERPRISE.md` — submodule + overlay setup
- Not an ops/overlay misconfiguration: the overlay was correctly included in the compose invocation, the mount was present and read-only, the package was importable in the container, and `TRINITY_OSS_ONLY` was unset. The OSS Alembic graph was clean throughout.

**Operator workaround (not a fix):** pin the submodule back to the previous revision and restart the backend — the enterprise directory is a bind mount, so a restart alone re-runs registration. This is undone by the next `git submodule update`, because the submodule carries `update = checkout`.

trinity-ability:
Fixes #2068

> ## ⛔ Merge order: **requires the private PR merged first — do not merge out of order**
>
> **PR-A (private, merges FIRST):** https://github.com/Abilityai/trinity-enterprise/pull/371
>
> That PR carries the actual root fix (the merge revision that collapses the fork) plus the private-side CI gate. **This PR does not advance the recorded submodule pointer** — see [Why the gitlink is not advanced here](#why-the-gitlink-is-deliberately-not-advanced-in-this-pr) — so merging it before PR-A will turn `deploy-dev` **red**. That is correct behaviour, not a defect: the enterprise line genuinely *is* degraded at the currently-pinned commit, and the whole point of this PR is that the platform stops pretending otherwise.

---

## What the bug was

Two Alembic heads in the enterprise version-line: two revisions descending from the same parent (`0009`), one cut weeks before the other and merged without its `down_revision` being re-checked.

`alembic upgrade head` is **singular**, and it resolves its target **before applying anything**. So a two-head graph applies **ZERO** revisions — not "the one that forked", but *every revision merged since the fork*. It raises `CommandError: Multiple head revisions are present`, and `register_enterprise()` catches that and **silently degrades the instance to OSS-only**: containers green, `/health` healthy, every entitled route answering 403, one ERROR line at boot.

**Git reports no conflict for this class.** Each revision file is independently valid; the defect exists only in the *relationship* between two files that were never edited together. Only a graph-level assertion sees it.

The asymmetry is worth internalising for triage: the **identical** fault on the OSS track aborts `init_database()` and crash-loops boot with a `/health` 503 — loud. On the enterprise line it is silent. Same bug, opposite observability.

## What this PR adds

| File | What |
|---|---|
| `scripts/ci/check_alembic_heads.py` | The guard. Pure stdlib, ~335 lines, no alembic install required. |
| `.github/workflows/schema-parity.yml` | Wires the guard in **unconditionally**, both version-lines. |
| `.github/workflows/deploy-dev.yml` | Post-merge: assert enterprise registration actually completed. |
| `docs/memory/architecture.md` | Invariant #3 — the single-head rule, beside the guards that paragraph already enumerates. |
| `docs/ENTERPRISE.md` | The pre-flight procedure before advancing the submodule pointer. |
| `docs/memory/feature-flows/database-migration-runner.md` | Cross-reference: that doc covers runner *execution* safety; nothing in it protects graph *shape*. |
| `tests/unit/test_2068_alembic_heads_guard.py` | 36 tests. |

The guard reads the graph by **parsing**, never importing: `ast.parse` + `ast.literal_eval` over each `*.py`, reading only the module-level `revision` / `down_revision` literals. A tuple/list `down_revision` (an `alembic merge` revision) contributes **every** member — which is exactly what lets a merge revision collapse a fork.

## The guard's honest limits

**It checks head COUNT only.** It does not verify that every `down_revision` resolves to a revision that exists, and it does not execute the migrations. `pg-migrations`' real `alembic upgrade head` remains the authority on whether the OSS line applies.

**A partially unparseable directory can fail OPEN.** `check_directory` fails only when *every* file is unparseable; a single dropped file among readable ones is silently omitted, and dropping a fork head hides the fork. This is **reproduced, not hypothesised**: a directory holding `0001 → 0002` plus a revision whose `revision` is set by tuple-unpacking (an `ast.Tuple` target, which the literal reader skips) reports **"2 revisions, 1 head — PASS"** while genuinely holding a fork.

It is pinned as an explicit characterization test that names the residual (`test_unparsed_file_can_hide_a_head`) rather than being asserted away, and mitigated by a per-line **"every revision file parses"** assertion — OSS already had one, the private line now has its twin. **It is deliberately not fixed by editing the script**, because the script is byte-identical-vendored into the private repo (`sha256 d04c99f8…` on both copies) and a one-sided edit breaks that contract. Tightening it must land in **both** copies in one change.

## Why a byte-identity test was added

The private repo's job cross-checks its vendored copy against real `alembic heads`. That proves the **private** copy agrees with alembic — it can **never** see that the **public** copy drifted, which is the likelier direction, since this is where contributors work. `test_vendored_guard_copy_is_byte_identical` is the check that sees it, in every populated checkout. It compares digests, not contents: a diff of a private file has no place in a test failure message.

## The wiring must not be neutered

The `schema-parity` step sits **before** `Detect schema-relevant changes` with **no `if:`**. That is load-bearing, not incidental: the job's own self-skip regex **omits `src/backend/migrations/`**, so a PR that adds *only* an Alembic revision self-skips — placing this step behind `steps.changes.outputs.relevant` would silently neuter it on exactly the PR class it exists for. (The workflow's `pull_request` trigger carries no `paths:` filter, so the job itself runs on every PR; only the `push:` trigger is path-filtered.)

The enterprise arm is **inert on public CI today**, said out loud rather than left to imply coverage: public CI never checks out the private submodule (`.gitmodules` sets `update = none`, no read credential exists in this repo), so that directory is absent and the guard prints an explicit `skipped` line. It still ships — live for free in every populated checkout, with the private repo running the same assertion at its own branch point and `deploy-dev` asserting the deployed outcome. A private-repo-read PAT was rejected: a credential readable from a public repository's workflow environment, for a check already performed at the source.

## Evidence

- **Guard EXIT=1 pre-fix** — named both heads, both filenames, and the fork point. **EXIT=0 post-fix.**
- **Cross-checked against real alembic `ScriptDirectory.get_heads()` in all four states** (fork / merged / single-line / empty) — agreement in every one.
- **36 guard tests pass** (`pytest tests/unit/test_2068_alembic_heads_guard.py`); neighbourhood alembic/migration set: **108 passed, 1 skipped** (the skip is a pre-existing local `alembic` import, unrelated).
- **Current state, both lines:**
  ```
  src/backend/migrations/versions — 36 revision(s), 1 head (0035_agent_ownership_a2a_exposed) — PASS.
  src/backend/enterprise/backend/migrations/versions — 16 revision(s), 1 head (…) — PASS.
  EXIT=0
  ```
- **Adversarial `/cso` proof** that `ast.parse` + `ast.literal_eval` never imports or executes parsed code: a revision file containing `os.popen("echo pwned")` produced no side effect.
- `enterprise-docs-guard`'s own grep re-run over the tree after the doc edits: **clean**. Both doc deltas describe the version-line mechanically; neither names a module, a feature id, or a private table.

## A corrected over-claim

The `--tail 2000` fallback in the new `deploy-dev` block does **not** carry the "current run only" property its original comment claimed. Measured on a live long-running instance, the current boot was **~14,180 lines back** — so the fallback would be simultaneously deep enough to catch a *previous* boot's FAILED line and too shallow to reach *today's* success line. The comment is fixed with the measurement, and the block is documented as sound only where it sits (`up -d` has just recreated the container, so its whole log is a few hundred lines).

## Why the gitlink is deliberately NOT advanced in this PR

`src/backend/enterprise` is left at its current pin, unstaged. The pointer must reference **PR-A's merge commit**, which does not exist until PR-A merges. Pointing it at a *branch tip* would literally reproduce the residual this issue is about — the same shape as `740063e3d` in #1628, a pointer proposed at a commit that lived only on an unmerged branch.

So the advance is this PR's one outstanding item, to be done as a follow-up commit once PR-A has merged.

## Post-merge verification

Confirm **all four** revisions applied and `GET /api/version` reports `edition: "enterprise"` with 13 features. **A green boot alone does not distinguish "applied four" from "applied none"** — that indistinguishability is the entire bug.

1. **Make enterprise registration failure externally observable** — a `/health` field, or fail-closed when the submodule is populated. `app.state.enterprise_failed_modules` already exists, so this is cheap.
2. **`deploy-dev` covers only our instances** — a customer-operated entitled instance still degrades silently. This is the largest remaining gap.
3. **Fix `schema-parity`'s self-skip regex** to include `src/backend/migrations/` (harmless for the SQLite-only parity pytest it actually gates, but wrong).
4. **Tighten the guard for partially-unparseable directories** — must land in both copies in one change.
5. **Validate the pin SHA** on the public→private `repository_dispatch` path.

Two **pre-existing** `deploy-dev` properties observed while working here, untouched by this PR:

- Its `=== Health ===` step uses `curl -sf … && echo`, which under `set -e` does **not** abort — so a dead backend previously deployed **green**. (The new block narrows this incidentally: with the submodule mounted, a dead backend now produces neither boot-log literal and fails the deploy.)
- Its `=== Error check ===` step echoes backend log lines into a **public** Actions log.


vybe (approved):
Verified independently rather than taken on trust:

- The vendored `scripts/ci/check_alembic_heads.py` here and the copy in ent#371 are **byte-identical** (12,866 bytes, `diff` clean).
- Guard vs the enterprise line at the previously-pinned commit: **FAIL, 2 heads**, fork point correctly named `0009_retention_opq_reports`. Guard with `0015_merge_a2a_control`: **PASS, 1 head**. An independently hand-written head-resolver agrees.
- Guard vs the OSS track on current `dev`: **PASS** — 36 revisions, 1 head (`0035_agent_ownership_a2a_exposed`), so wiring it into `schema-parity` unconditionally does not start red.
- Absent-directory skip path: exit 0 and says `skipped` out loud.
- `tests/unit/test_2068_alembic_heads_guard.py`: **34 passed, 2 skipped** locally (the 2 skips are the documented uninitialised-submodule path).

## One change I pushed to this branch

`chore(enterprise): advance the submodule pointer` (ebfe97f9).

This PR's body said it deliberately does **not** advance the gitlink; ent#371's body said the public PR does. Both cannot hold, and the difference is whether `dev` deploys green. `deploy-dev` resolves the submodule with `git submodule update --init` — at the **recorded pointer**, not the enterprise branch tip — and `dev` still recorded `c2cfeb3f`, which *is* the two-head commit. So without the bump this PR's own new assertion would have correctly failed the deploy, and the dev instance would have stayed silently OSS-only: the exact condition #2068 exists to end.

The advance is `c2cfeb3f → d1c5ebb2`, **1 commit ahead / 0 behind**, carrying ent#371's three files and nothing else — no unreviewed enterprise work rides along. Verified via the compare API before pushing.

Invariant #3 doc edit is in the right home and reads correctly. Approving.

Issue #462, opened 2026-04-22
bug: UI Push commits runtime state (.cache, .claude/sessions, .claude.json, .sudo_as_admin_successful, etc.) — git add -A against too-narrow .gitignore


romainability:
## Summary

Clicking the **Push** button in the agent UI runs `git add -A && git commit -m "Trinity sync: <timestamp>"` inside the agent container and pushes runtime state to GitHub — `.cache/`, `.claude/sessions/`, `.claude/shell-snapshots/`, `.claude.json`, `.local/lib/python3.11/site-packages/`, `.sudo_as_admin_successful`, etc. The platform's own [`docs/TRINITY_COMPATIBLE_AGENT_GUIDE.md:126-169`](../blob/main/docs/TRINITY_COMPATIBLE_AGENT_GUIDE.md#L126-L169) lists these exact paths under "Instance-specific files — DO NOT COMMIT". The Push button is not respecting the agent-value-vs-runtime split that Trinity's own docs require of agent repos.

In one observed instance, a single Push pushed **1,599 runtime files** to GitHub, making the PR diff and `main` history unreadable.

## Context

The full call chain:

1. UI Push → `POST /api/agents/{name}/git/sync` (`src/backend/routers/git.py:100`)
2. → `git_service.sync_to_github(agent_name, ...)` (`src/backend/services/git_service.py:466`)
3. → calls `http://agent-{name}:8000/api/git/sync` (`git_service.py:509`)
4. → agent-side handler `sync_to_github` at `docker/base-image/agent_server/routers/git.py:557`
5. → runs **`git add -A`** (line 692) when no `paths` are passed — i.e., the default UI Push path
6. → commits with `Trinity sync: <timestamp>` (line 721) and pushes

`git add -A` stages everything not gitignored. The platform-injected `.gitignore` (written once during `initialize_git_in_container` at `src/backend/services/git_service.py:719-728`) only covers:

```
.bash_logout, .bashrc, .profile, .bash_history,
.cache/, .local/, .npm/, .ssh/
```

Compare to what `docs/TRINITY_COMPATIBLE_AGENT_GUIDE.md:126-169` requires every Trinity-compatible agent to exclude:

```gitignore
# Credentials - NEVER COMMIT
.mcp.json
.env
*.pem
*.key
credentials.json

# Large generated content
content/

# Instance-specific directories
.npm/
.ssh/
.trinity/
.cache/

# Instance-specific files
.claude.json
.claude.json.backup
[...]

PR #2077, opened 2026-08-10
fix(git): resolve an agent's repo root from git, not a workspace/ content probe (#2075)

### LINKED ISSUE #2075: `_detect_git_dir` picks the wrong repo root for agents with a populated `workspace/` data directory

dolho:
Fixes #2075

## Problem

`_detect_git_dir` picked the repo root with a **content** probe — any non-empty `/home/developer/workspace` was declared the root. Its docstring said "legacy agents (created before 2026-02) only", but it tested content, not legacy-ness. A standard agent rooted at `/home/developer` that also keeps a populated non-git `workspace/` **data** directory was therefore resolved to the wrong path by every caller that uses the result as a filesystem path:

- **compatibility snapshot** collected from `workspace/` → `F-001 template.yaml exists` fails, and the 23 `_with_template` checks behind it return `no_template`. Those agents were never actually evaluated.
- **`.gitignore` check validated the wrong file.** Both files are real and both are honoured by git — but the one in `workspace/` governs only that subdirectory, while the credential files the check names live at the repo root. A security-relevant check passing on a rule that does not apply to the files it names.
- **`POST /{agent}/compatibility/fix`** wrote its correction into that subdirectory `.gitignore`, verified its own write, and reported success.
- **The per-Push `.gitignore` migration (#462) silently no-op'd**: its `[ -d <dir>/.git ]` guard tested the wrong directory, so fleet-wide ignore patterns never reached these agents.

The platform already disagreed with itself — `check_git_initialized` (which gates the re-init `409`) resolves these agents correctly.

## Fix

Ask git. `_git_toplevel` runs `git rev-parse --show-toplevel`, probing from `workspace/` when it exists and from `/home/developer` otherwise. `--show-toplevel` walks **up**, so the nearest enclosing repository wins: a genuinely workspace-rooted legacy repo still answers `workspace/`, a standard agent answers `/home/developer`. An answer outside `/home/developer` is rejected. `safe.directory` is relaxed for this read-only query so foreign volume ownership can't silently drop the caller back onto the heuristic.

The content heuristic is kept **verbatim** as `_detect_git_dir_fallback` and runs only when the container has no repository at all — `initialize_git_in_container` uses it to place a repo that does not exist yet, so fresh-agent placement stays byte-compatible.

## Known limitation (unchanged from the issue)

An agent already re-initialised **while misdetected** has a real `workspace/.git` — a genuine nested repo, indistinguishable from a legitimate legacy agent by any probe. Git's own answer for it *is* `workspace/`, and this fix keeps it there. Repairing those takes an operator decision per agent:

```bash
docker exec agent-<name> bash -c \
  '[ -d /home/developer/.git ] && echo home-repo; \
   [ -d /home/developer/workspace/.git ] && echo workspace-repo; \
   git -C /home/developer rev-parse --show-toplevel 2>/dev/null'
```
Both markers ⇒ nested repo ⇒ needs a human.

## Operator notes

1. **Next Push on an affected agent contains deletions.** With the migration no longer no-op'ing, the next Push appends the fleet-wide patterns and `git rm --cached`s files that now match a rule. Working-tree files are untouched and history is **not** rewritten — anything already pushed stays in history, so credential rotation is separate follow-up work.
2. **Re-init has a larger scope.** If such an agent is re-initialised after this fix, the ignore merge, `git init`, `git add .` and (empty-remote path) the force push run at `/home/developer`. Ordering is safe by construction: the merge happens **before** `git init`/`git add .`, so the fleet-wide rules are in force at the real root before anything is staged.

## Tests

`tests/unit/test_2075_detect_git_dir.py` — the filesystem cases run the **real** probe script against real git repos in a temp tree (the only difference from production is host FS vs. container):

- home-rooted repo + populated `workspace/` → `/home/developer` (the repro)
- genuinely workspace-rooted legacy repo → `workspace/` (unchanged)
- nested repo → `workspace/` (known limitation, pinned)
- no repo anywhere → the fallback heuristic decides
- probe contract: out-of-home toplevel rejected, git failure falls back, the fallback command is byte-identical to the pre-#2075 one, and a successful probe costs exactly one exec (the heuristic can't override git)

```
tests/unit/test_2075_detect_git_dir.py .........  9 passed
pytest tests/unit -k "git or compat"           594 passed, 6 skipped
```

## Docs

Feature flows refreshed where they documented the old heuristic verbatim: `github-repo-initialization.md`, `github-sync.md` (incl. the operator note), `agent-compatibility-validation.md`.


dolho:
## Live verification (real agent container, not just unit tests)

The platform stack on this box is stopped (backend/redis down since ~32h), so this was run against a throwaway container from the real `trinity-agent-base:latest` image, driving the **actual production code path** — `_detect_git_dir` / `_migrate_workspace_gitignore` / `compatibility.collector.collect` with only `execute_command_in_container` swapped for a `docker exec -u developer` of the same command string (docker-py `shlex`-splits string commands identically). The read-only probe was also run against the live `agent-trinity-system` container (no writes).

### 1. The repro layout — repo at `/home/developer`, populated non-git `workspace/`

Command the fix actually sends (one exec):
```
bash -c 'start=/home/developer/workspace; [ -d "$start" ] || start=/home/developer; git -c safe.directory='"'"'*'"'"' -C "$start" rev-parse --show-toplevel 2>/dev/null'
```
```
RESULT (fixed):        /home/developer
RESULT (old heuristic): /home/developer/workspace
```

### 2. The other layouts, same container

| Layout | detect → |
|---|---|
| genuine legacy (repo rooted in `workspace/` only) | `/home/developer/workspace` ✅ unchanged |
| no repo at all + populated `workspace/` (fresh-agent placement) | `/home/developer/workspace` ✅ fallback, byte-compatible |
| nested repos (`.git` at both) | `/home/developer/workspace` ✅ known limitation, as documented |
| live `agent-trinity-system` (no repo, no `workspace/`) | falls back → `/home/developer` ✅ |

### 3. Downstream: the #462 migration actually runs now

Setup: repo at `/home/developer`, tracked `.env`, root `.gitignore` with one pre-existing rule, plus `workspace/cache.db` + `workspace/.gitignore`.

```
old detection targets workspace/ →  [ -d /home/developer/workspace/.git ]  exit=1
                                    ⇒ migration returns early ⇒ silent no-op   (bug reproduced)

with the fix, _migrate_workspace_gitignore ran:
  root .gitignore:      "some-existing-rule" preserved, fleet patterns appended (52 lines, incl. .env)
  workspace/.gitignore: "junk"  — untouched
  tracked after:        .env no longer in the index
  .env on disk:         still present
```
That is exactly operator note 1: index-only deletion, working tree untouched, history not rewritten.

### 4. Downstream: the compatibility snapshot root

Same container, same collector, only the root differs:

| root | `template.yaml` | `.gitignore` |
|---|---|---|
| `/home/developer` (fixed) | `exists=True` ✅ | exists=True (the **root** one) |
| `/home/developer/workspace` (old) | `exists=False` ❌ → `F-001` fails → 23 `_with_template` checks skipped | exists=True (the **subdirectory** one) |

The throwaway container was removed afterwards; the live system agent was only read from.

vybe (approved):
Validated via /validate-pr. Green on all 21 required checks; merge into dev is conflict-free and the combined tree with #2078 passes `pytest tests/unit -k "git or compat"` (594 passed, 8 skipped). Scope is tight, the fallback heuristic is preserved byte-identically for fresh-agent placement, and the known nested-repo limitation is pinned by a test rather than papered over.

PR #2078, opened 2026-08-10
fix(git): ignore .trinity/ contents-only so template-committed hooks survive a Push (#2070)

### LINKED ISSUE #2070: bug: first backend Push untracks template-committed .trinity/pre-check and pushes the deletion — third recurrence of the too-narrow rm --cached exemption list
## Summary

Every backend Push runs `git rm --cached` over all tracked-but-now-ignored files (`_build_rm_cached_ignored_command`, `src/backend/services/git_service.py:1386`). Because the fleet-wide list ignores `.trinity/` **wholesale** (`git_service.py:1288`), any file a template legitimately commits under `.trinity/` is untracked and its deletion pushed. The exemption list is a hardcoded two entries (`git_service.py:1404`):

```
exempt = '":!.trinity/brain-orb" ":!.trinity/setup.sh"'
```

`.trinity/pre-check` is not among them — and it is a **documented, template-committed** file.

## The recurrence

This is the third time the same too-narrow pathspec has bitten, and the function's own docstring records the first two:

1. `.trinity/brain-orb/` — Brain-Orb templates commit their hooks (trinity-enterprise#76)
2. `.trinity/setup.sh` — setup-convention templates commit it; *"verified live: the e2e Push swept setup.sh before the second exemption was added"*
3. `.trinity/pre-check` — this issue

Each fix added one more string to the exemption list, which is why there is a third.

## Why `pre-check` is authored content, not runtime state

`~/.trinity/pre-check` is the SCHED-COND-001 conditional-schedule hook (#454). The platform documents templates shipping it:

- `docs/memory/architecture.md:350` — *"if the template ships an executable `~/.trinity/pre-check`, the backend's internal endpoint … runs it via `docker exec` before a cron-triggered chat"*
- `docs/memory/requirements/scheduling.md:287` — *"Contract: agent templates drop an executable `~/.trinity/pre-check` file with a shebang"*
- `docs/agent-validation-spec.md:91` — check **A-004** grades it for executability and a valid shebang
- `docs/TRINITY_COMPATIBLE_AGENT_GUIDE.md:1530` — the agent-defined-pipeline heartbeat is documented as gated by it

So the platform tells template authors to commit a file that the platform then silently deletes from their repo.

## Impact

A template shipping a `pre-check` loses it from `origin` on the first backend Push. The working-tree copy survives (only the index is touched), so nothing breaks immediately — the failure lands at the **next re-clone or container recreate**, which boots with no hook. The scheduler's pre-check call then finds nothing and every cron tick runs a full LLM turn instead of skipping: silently reintroducing exactly the token cost SCHED-COND-001 exists to remove, on a poll-driven agent, with no error anywhere.

Same exposure for `.trinity/pipelines/*.yaml` — #919 defines those as agent-authored *definitions* (state lives separately under `pipeline-state/`), so they are authored content sharing the runtime directory.

**Honest scope:** no *bundled* template currently ships any `.trinity/` file (verified — `find config/agent-templates -path "*/.trinity/*"` is empty), so this is latent in-repo and live for external templates that follow the documented convention. Both prior recurrences were found on live external templates, not bundled ones.

## Proposed fix

Prefer inverting the rule over extending the exemption list: ignore the specific runtime paths rather than the whole directory, so authored content under `.trinity/` is committed by default and no exemption list is needed.

Runtime (ignore): `.trinity/pipeline-state/`, `.trinity/pending-results/`, `.trinity/operator-queue.json`, `.trinity/persistent-state.yaml`, `.trinity/data-paths.yaml`, `.trinity/read-only-config.json`
Authored (track): `pre-check`, `setup.sh`, `brain-orb/`, `pipelines/*.yaml`

The cheap alternative — appending `":!.trinity/pre-check"` — is a fourth instance of the pattern that produced this issue and does nothing for `pipelines/`.

## Acceptance Criteria

- [ ] A template shipping an executable `.trinity/pre-check` still has it tracked after a backend Push, and a fresh clone / container recreate of that agent still runs the hook
- [ ] Same for `.trinity/pipelines/*.yaml` definitions, and no regression for `.trinity/brain-orb/` or `.trinity/setup.sh`
- [ ] Platform runtime state under `.trinity/` is still ignored, and still untracked-if-already-tracked
- [ ] Adding a new runtime file under `.trinity/` cannot silently untrack authored content — a regression test pins the authored/runtime split rather than the current exemption strings
- [ ] If the pattern set changes, the coupled surfaces move in the same change: the guide mirror (`test_doc_and_constant_in_sync`), compat check `S-005` and `c_g002` (`services/compatibility/static_checks.py:752`), and the auto-fix set (`services/compatibility/fixes.py:100`)

## Technical Notes

- Compat check `S-005` is keyed on the literal `.trinity/` line being present in `.gitignore`, and `c_g002` asserts every `_GITIGNORE_PATTERNS` entry is present — an inversion turns both red until updated, and the auto-fix would otherwise re-add the wholesale line. This coupling is the main cost of the inversion and should be handled in one change.
- Only the index is touched by `git rm --cached`, so no working-tree data is lost — which is precisely why this is silent and why the symptom surfaces one recreate later.
- Related: trinity-enterprise#76 (brain-orb exemption), #454 (SCHED-COND-001 pre-check), #919 (agent-defined pipelines), #1703 (repo root is `$HOME` — the structural cause of the whole exclusion layer).

dolho:
Fixes #2070

## Problem

Every backend Push runs `git rm --cached` over tracked-but-now-ignored files. The fleet-wide list ignored `.trinity/` **wholesale**, so anything a template legitimately commits there was untracked and its deletion pushed. The only protection was a hardcoded two-entry pathspec — wrong three times now, the first two recorded in the function's own docstring:

1. `.trinity/brain-orb/` (trinity-enterprise#76)
2. `.trinity/setup.sh` — *"verified live: the e2e Push swept setup.sh before the second exemption was added"*
3. `.trinity/pre-check` — this issue

`pre-check` is the SCHED-COND-001 hook (#454) that `architecture.md`, `requirements/scheduling.md`, `agent-validation-spec.md` (A-004) and the agent guide all tell template authors to commit. The platform documents a file as authored content and then deletes it.

**Silent and delayed**: only the index is touched, so the working tree keeps running. It surfaces at the next re-clone or container recreate, which boots with no hook — and every cron tick then runs a full LLM turn instead of skipping, reintroducing exactly the cost SCHED-COND-001 exists to remove, with no error anywhere.

## Fix — inverted, not extended

Appending `":!.trinity/pre-check"` would be a fourth instance of the pattern that produced this issue, and does nothing for `pipelines/`.

```gitignore
.trinity/*                              # contents-only, so negations apply
!.trinity/pre-check                     # _TRINITY_AUTHORED_PATHS — one entry
!.trinity/post-check                    # per authored path
!.trinity/pre-snapshot
!.trinity/setup.sh
!.trinity/persistent-processes.allow
!.trinity/brain-orb/
!.trinity/pipelines/
```

The **default** under `.trinity/` is now *ignored* — a new runtime file needs no action — while authored content is tracked by construction. The `rm --cached` exemptions are **derived** from the same constant, so adding an authored path is one edit instead of two and the two cannot drift.

**The dir-form is load-bearing, not cosmetic**: git never descends into a directory excluded by `.trinity/`, so re-includes beneath it are inert. Every agent synced before this change carries that exact line — precisely the fleet this fixes — so the merge now **removes** it before appending, by exact whole-line match (`grep -vxF`, the discipline the G-001 auto-fix already uses). A user rule that merely mentions the directory (`.trinity/scratch/`, `!.trinity/mine`) doesn't match and is left alone.

## Verified in a real agent container

`trinity-agent-base`, starting from the pre-#2070 `.gitignore` every live agent has:

```
old commands: tracked after sweep -> .trinity/setup.sh only
                                     (pre-check and pipelines/r.yaml untracked — the bug)
new commands: tracked after sweep -> pre-check, setup.sh, pipelines/r.yaml
              runtime (operator-queue.json, pipeline-state/) untracked
              .gitignore migrated to the star form
              fresh `git clone` of the result -> hook present, runs, prints "skip"
```

## Coupled surfaces, moved in the same change

Per the AC: the guide's gitignore fence (`test_doc_and_constant_in_sync`), **S-005** (star form is now canonical; the dir-forms stay *accepted* — they do exclude the runtime state the check grades, and the merge repairs them anyway, so failing them would report a problem the platform fixes itself), the **S-005 auto-fix** patterns (re-adding the wholesale line would have re-broken this), **c_g002** (follows the constant automatically), the validation spec, `architecture.md`, and all **16 bundled template `.gitignore` files**.

## Tests

`tests/unit/test_2070_trinity_authored_paths.py` — 25 tests running the **real** command builders against **real** git repos (the defect is in git's own descend/negation semantics, which a mocked container can't express). They assert the **split**, never today's exemption strings, so a fourth forgotten path fails here rather than shipping:

- 8 authored paths survive the sweep; 9 runtime files stay untracked
- a fresh clone still **runs** the hook (the thing that actually breaks, one recreate later)
- the superseded wholesale line is replaced; user rules mentioning `.trinity` are preserved; the merge is idempotent
- exemptions are derived from `_TRINITY_AUTHORED_PATHS`; a newly-invented runtime file needs no action

```
as shipped                          -> 25 passed
wholesale ignore restored           ->  5 failed
hardcoded two-entry exemptions      ->  1 failed
stale-line migration removed        ->  2 failed
pre-check dropped from authored set ->  4 failed

related suites (git/compat/template) -> 916 passed, 6 skipped
```


vybe (approved):
Validated via /validate-pr. All 19 checks green. Inverting the rule (`.trinity/*` + derived re-includes) is the right fix rather than a fourth exemption string, and deriving the `rm --cached` pathspec from the same constant removes the drift that caused all three recurrences. Verified locally: merges conflict-free on top of #2077 and the combined tree passes `pytest tests/unit -k "git or compat"` (594 passed, 8 skipped) plus all 25 new tests.

Non-blocking nit for a follow-up: in `_build_gitignore_merge_command`, `grep -vxF ... > .gitignore.tmp || true` swallows grep's exit 2 (real error) as well as exit 1 (no lines selected), so a genuine read failure mid-filter would `mv` a truncated file over the user's .gitignore. `|| [ $? -eq 1 ]` would narrow it to the intended case.

PR #2081, opened 2026-08-10
fix(security): CSO 2026-08-09 — terminal-tier HIGH + access-control + supply-chain hardening

trinity-ability:
## Summary

Fixes surfaced by the `/cso full` audit (`docs/security-reports/cso-2026-08-09.md`). One HIGH plus access-control and supply-chain hardening. The report + structured JSON are included in this PR.

### HIGH
- **H1 — terminal WebSocket privilege escalation.** `services/agent_service/terminal.py` gated a `/bin/bash` container shell at *accessor* tier (`can_user_access_agent`), so a chat-only **shared** collaborator could open the terminal against a running agent and dump `CLAUDE_CODE_OAUTH_TOKEN` / `TRINITY_MCP_API_KEY` / `.env` (RCE as the agent, lateral movement via the MCP key). Now gated at owner/admin tier (`can_user_share_agent`), matching the deliberately admin-only SSH endpoint. Confirmed prod-reachable + exploitable by two independent fresh-context refutations.

### Access control
- **M1** — schedule `enable`/`disable`/`trigger` moved from `AuthorizedAgent` → `OwnedAgentByName` (they were accessor-tier while `update`/`delete` required ownership; a shared user could flip owner-intent schedule state and burn LLM spend).
- **L1** — `GET /event-subscriptions/{id}` gains `assert_agent_access` (the owner check its `PUT`/`DELETE` siblings already enforce; bounded IDOR by ~96-bit id, but the asymmetry is real).

### Supply chain
- **M3** — `npm audit fix` on mcp-server + helper-mcp + frontend clears the HIGH `ip-address` (SSRF) and `undici` (response-desync) advisories in the MCP prod tree; `ip-address` override floor raised to `>=10.5.0`. **0 vulnerabilities** in all three trees.
- **M4** — `npm install` → `npm ci` in the mcp-server and frontend prod Dockerfiles, restoring the tracked lockfile as a control.
- **M5 (partial)** — backend `Dockerfile` pins `aiohttp==3.14.3`, `slackify-markdown==0.2.4` (were floating amid an otherwise exact-pinned set).

### CI hygiene
- **M9 / L5** — least-privilege `permissions: contents: read` added to `frontend-build.yml` and `deploy-dev.yml`.

## Verification
- `py_compile` on all changed backend files; `test_186_enumeration_uniformity` (36) and `test_subscription_bola` pass.
- `npm audit` → **0 vulnerabilities** on mcp-server / helper-mcp / frontend; mcp-server `tsc` and frontend `vite build` both succeed; `npm ci` exits 0 (lockfiles in sync).
- Backend pins read from the live running backend's own resolved versions.

Full integration verification of the H1/M1 auth changes (shared-user 403 vs owner 200) runs on CI.

## Deliberately held (need a build/CI-verified pass — see report)
- **M5** base-image pip pins + Go-tarball SHA (needs a base-image rebuild; a wrong pin takes down every agent).
- **M4** dev frontend Dockerfile (deletes the lockfile for cross-arch native binaries; dev-only).
- **M8** SHA-pinning ~15 first-party actions (needs exact SHAs + a CI run).
- **M2 / M6 / M7 / L7** — policy/owner-decision items (backlog_metadata drain-null, git-history scrub, Google-scope narrowing, docs policy).

### DIFF
--- a/.github/workflows/deploy-dev.yml
+++ b/.github/workflows/deploy-dev.yml
@@ -4,6 +4,10 @@ on:
   push:
     branches: [dev]
 
+# CSO L5: least-privilege token — deploy runs over SSH/Tailscale secrets, not GITHUB_TOKEN.
+permissions:
+  contents: read
+
 jobs:
   deploy:
     runs-on: ubuntu-latest
--- a/.github/workflows/frontend-build.yml
+++ b/.github/workflows/frontend-build.yml
@@ -11,6 +11,10 @@ on:
       - 'src/frontend/**'
       - '.github/workflows/frontend-build.yml'
 
+# CSO M9: least-privilege token — this workflow only reads the checkout.
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/docker/backend/Dockerfile
+++ b/docker/backend/Dockerfile
@@ -101,8 +101,8 @@ RUN pip install --no-cache-dir \
     reportlab==4.2.5 \
     google-genai==1.63.0 \
     "slack_sdk[socket-mode]==3.41.0" \
-    aiohttp \
-    slackify-markdown \
+    aiohttp==3.14.3 \
+    slackify-markdown==0.2.4 \
     tenacity==9.0.0 \
     opentelemetry-api==1.29.0 \
     opentelemetry-sdk==1.29.0 \
--- a/docker/frontend/Dockerfile.prod
+++ b/docker/frontend/Dockerfile.prod
@@ -6,11 +6,12 @@ FROM node:24-alpine AS builder
 
 WORKDIR /app
 
-# Copy package files
-COPY package.json package-lock.json* ./
+# Copy package files (lockfile is tracked — require it)
+COPY package.json package-lock.json ./
 
-# Install dependencies (use npm install since package-lock.json might not exist)
-RUN npm install
+# Install from the tracked lockfile (CSO M4: npm ci is lockfile-exact and fails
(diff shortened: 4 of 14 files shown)


vybe (approved):
Reviewed the full diff. Approving including M1.

**H1 (terminal WebSocket privilege escalation)** — this is the one that matters and the fix is right. Gating a `/bin/bash` container shell at accessor tier meant a chat-only shared collaborator could dump `CLAUDE_CODE_OAUTH_TOKEN` / `TRINITY_MCP_API_KEY` / `.env` — RCE as the agent plus lateral movement via the MCP key. Moving to `can_user_share_agent` matches the deliberately admin-only SSH endpoint, which is the correct precedent: a shell is not an accessor-tier capability.

**M1 (schedule enable/disable/trigger → `OwnedAgentByName`)** — merging as audited. This is a deliberate, user-visible tightening: shared collaborators lose the ability to flip schedule state. It's the right call because `update`/`delete` already required ownership, so accessor-tier `enable`/`disable`/`trigger` was an inconsistency that let a shared user override owner intent and burn LLM spend. **Needs a release note** — this is a behaviour change, not a pure fix.

**L1** — `GET /event-subscriptions/{id}` gaining `assert_agent_access` simply matches the gate its `PUT`/`DELETE` siblings already enforce.

Queued behind #1976, which touches `routers/schedules.py` too.

One process note, not a blocker: this PR carries no closing keyword, so nothing auto-promotes. The committed `docs/security-reports/cso-2026-08-09.{md,json}` are the durable record, which is the right shape for an audit sweep spanning several unrelated findings.

PR #2082, opened 2026-08-10
test(harness): honest full-suite runs — fix collection breakage, add PG + git-sync tiers, skip audit (#2080)

### LINKED ISSUE #2080: Test harness: honest full-suite runs — fix collection breakage, add PG/Alembic + git-sync tiers, skip-audit gate
## Summary

Make one command produce an **honest full-suite result**: every tier (unit, integration, git-sync, Postgres/Alembic, agent-dependent) actually runs, harness breakage cannot masquerade as test results, and any skip outside a named allowlist fails the run. Today a full run on `dev` produces **zero usable signal** — the canonical `tests/run-full.sh` aborts at pytest collection before a single test executes, and the tiers that would exercise Postgres/Alembic and git-sync are either skipped or structurally broken.

## Context

A full-suite run against a freshly redeployed stack (preflight fully green: backend, Docker, Redis, real agent create/delete all verified working) surfaced that the failures are entirely in the **test harness**, not the platform:

1. **`tests/git-sync/` poisons collection of `tests/integration/`.** The directory has an `__init__.py` under a hyphenated (invalid-identifier) name and mixes vitest JS with Python. When collected in the same session, its `conftest.py` collides with the root `tests/conftest.py` under the bare `sys.modules["conftest"]` key, so `tests/integration/conftest.py`'s `from conftest import CALLER_SUPPLIED_REDIS_URL` binds to the wrong module — all integration files fail to collect, and `set -e` in `tests/run-full.sh` aborts the whole script (unit tier never runs).
2. **`sys.modules["utils"]` shadowing rot in `tests/conftest.py`.** The harness points `utils` at `tests/utils`, shadowing `src/backend/utils/`. The known-pollution preload list was not updated for the recently added `src/backend/utils/safe_yaml.py`, so any test importing through `template_service`/`git_service` dies with `ModuleNotFoundError: No module named 'utils.safe_yaml'`. In a continue-past-errors diagnostic run this contaminated ~1,000 tests (429 F / 583 E) that are not real regressions. This list will rot again every time `src/backend/utils/` gains a module — the fix should remove the shadowing (e.g. rename `tests/utils` to an unambiguous package name), not extend the list.
3. **Test venv drift.** `hypothesis` is declared in `tests/requirements-test.txt` but was not installed in the venv, aborting the entire `unit/` tier (~8.8k items) at collection. The venv also ran Python 3.14 while the images pin 3.13.
4. **`tests/test_web_terminal.py` hangs indefinitely** (confirmed 0% CPU), and interrupting it exposes a `pytest-timeout` + Python 3.14 signal-reentrancy `INTERNALERROR`.
5. **Postgres/Alembic tests never run locally** — they skip with "no PostgreSQL configured", so the dual-track migration contract is only exercised in CI.
6. **Agent-dependent tests silently skip** when `TEST_AGENT_NAME` is unset.

The net effect: a green-looking (or noisily red) run tells you nothing. Compounding this, the required checks on `dev` do not run the unit tier, so harness rot lands with green CI.

## Acceptance Criteria

- [ ] `tests/git-sync/` no longer breaks collection of other tiers (root fix — remove the module/conftest collision, not `--ignore`) and runs as its **own pytest invocation** in the full run, so its Python tests are included, not excluded.
- [ ] The `sys.modules["utils"]` shadowing in `tests/conftest.py` is eliminated (e.g. `tests/utils` renamed to a non-colliding package), so new `src/backend/utils/*` modules can never be shadowed again; no preload list to maintain.
- [ ] `tests/run-full.sh` self-heals the venv: idempotent `pip install -r requirements-test.txt` (including `hypothesis`, `audioop-lts`, `openpyxl`, `reportlab`, `alembic`, `psycopg`) before running, and the venv Python version matches the image pin (guarded, per the #1891 single-declaration rule).
- [ ] A **Postgres/Alembic tier**: the full run spins up a disposable `postgres:16` container, exports `DATABASE_URL`, runs `alembic upgrade head`, runs the PG-gated tests against it, and tears down — locally, not just in CI.
- [ ] Agent-dependent tests run: setup ensures a live test agent exists and sets `TEST_AGENT_NAME` (only genuinely-external credentials — Slack/Twilio — remain feature-gated).
- [ ] Documented run-standalone tests (pristine `sys.modules` route-order tests) execute as separate invocations within the same orchestrated run.
- [ ] **Skip-audit gate**: after all tiers, the runner parses `-rs` output and fails the run on any skip whose reason is not on a named allowlist — a new unexplained skip is a failure, not a green.
- [ ] `tests/test_web_terminal.py` hang fixed (or the test quarantined with a linked defect) and `--timeout-method=thread` pinned so one bad test cannot stall a run for hours.
- [ ] A single entry point (e.g. `tests/run-full.sh`) runs all tiers in sequence and exits non-zero on any tier failure, collection error, or skip-audit violation; per-tier summaries in the final output.

## Technical Notes

- Affected files: `tests/run-full.sh`, `tests/setup-env.sh`, `tests/conftest.py`, `tests/integration/conftest.py`, `tests/git-sync/` (has `__init__.py` + `conftest.py`), `tests/requirements-test.txt`, `tests/test_web_terminal.py`.
- The PG tier can reuse the `pg-migrations` CI job's approach (`alembic upgrade head` against a real PostgreSQL) — see `scripts/ci/` and `src/backend/migrations/`.
- The Python-version parity guard pattern already exists: `tests/unit/test_1891_python_version_parity.py`.
- Rough shape: harness-only work, no product code changes expected (except possibly the web-terminal test itself).
- Follow-up worth considering (separate issue): make the `dev` required checks run the unit tier so harness rot cannot land green.

dolho:
Fixes #2080

## The problem, restated

A full run on `dev` produced **zero usable signal**. Every defect removed *coverage* without failing anything — which is exactly why none of them were noticed.

## Root causes

**1. `tests/utils` shadowed `src/backend/utils`.** `pythonpath` lists `tests` first, so `utils` meant the test helpers; the backend package was reachable only via a hand-maintained preload list, and that list had not been updated for `safe_yaml.py`. Anything importing through `template_service`/`git_service` died with `No module named 'utils.safe_yaml'` — ~1,000 tests, all reading like product regressions.

Fixed **structurally**, per the AC: the helpers are now `tests/testkit` (87 files rewritten), so `utils` is unambiguous and there is no list to maintain.

**2. `tests/git-sync/` poisoned collection of `tests/integration/`.** An `__init__.py` under a non-identifier directory name put its conftest in `sys.modules` under the bare key `conftest`, colliding with the root one — so `tests/integration/conftest.py`'s `from conftest import ...` bound to the wrong module, 5 files failed to collect, and `set -e` killed the run before the unit tier ever started.

Renamed to `tests/git_sync` (root fix, not `--ignore`) and given its **own tier**, so its 23 Python tests now run instead of being excluded:

```
integration + git-sync collection:  7 errors, run aborted  ->  106 collected
```

**3. Two route-ORDER tests were skipping ALWAYS** — standalone included — with a message blaming the shadowing. The real cause was an undeclared dependency (`import main` needs the OTLP exporter + three instrumentations). Declared them at the image's versions; **10 permanently-dead tests now run and pass**.

## The runner

`tests/run-full.sh` is now an orchestrator: per-tier invocations (unit, integration, git-sync, security, scheduler, agent-server, api, standalone, postgres), **all tiers run even after one fails**, per-tier PASS/FAIL summary, non-zero exit on any tier failure, collection error, or audit violation.

- **venv self-heal** — creates it, `pip install -r requirements-test.txt` every run, and refuses to run when the venv Python differs from the image pin, **read from** `docker/backend/Dockerfile` (#1891 single-declaration rule). `TRINITY_TEST_ALLOW_PY_MISMATCH=1` overrides loudly and is repeated in the summary so such a run can't be quoted as clean.
- **Postgres/Alembic tier** — disposable `postgres:16-alpine`, real `alembic upgrade head`, exports **both** `DATABASE_URL` (engine) **and** `TEST_POSTGRES_URL` (what the PG-gated tests actually gate on). My first cut exported only the former: alembic upgraded a real database, every PG test still skipped, and the tier passed — the precise failure mode this issue is about, caught by the audit. Labelled + signal-trapped, so a killed run can't leak a container that blocks the next one on "port already allocated" (that happened, and is now handled).
- **Agent provisioning** (`harness/ensure_test_agent.py`) so agent-dependent tests run rather than skip.
- **Standalone tests** run as separate invocations, discovered from the tests' own `Run standalone: pytest …` text, so the list cannot go stale.
- **Skip audit** (`harness/audit_skips.py`) — parses `-rs` and **fails the run** on any skip whose reason is not on a named allowlist, each entry carrying a written justification. It has already earned its keep: it caught `REDIS_PASSWORD not set`, which I fixed by exporting it in `setup-env.sh` rather than allowlisting it.
- **Nothing can hang** — per-tier `--timeout` + `--timeout-method=thread` (also pinned in `pyproject.toml` for hand-runs). `signal` re-enters the interpreter from a handler, which is what turned the web-terminal hang into an INTERNALERROR.

## `test_web_terminal.py` — stated honestly

The **root cause is not fixed**: reproducing it needs a live stack and none was available. What is fixed is that it can no longer cost more than 60s (module-level `pytest.mark.timeout(60)`, verified not to be displaced by the class-level `pytestmark`), and a thread-method timeout now yields a failing test id instead of a run that never ends. Said in-file rather than implied.

## Guards — because every one of these was silent

`tests/unit/test_2080_harness_contract.py` (12 tests): `tests/utils` cannot return; no package dir may have a non-identifier name; every test directory must be named by a tier; every tier must pass `-rs` and a thread timeout; pytest's exit-5 ("no tests collected") must be a **failure**; the audit must refuse to certify an empty log dir; each allowlist entry must carry a justification.

## Verified by running it

```
runner, git-sync tier:  23 passed, summary + audit, exit 0
runner, security tier:  11 skipped -> skip-audit FAILED the run, exit 1
runner, postgres tier:  container up, `alembic upgrade head` OK,
                        PG tests executing against a real database
interpreter guard:      exits 2 with the recreate command
harness guards:         12 passed
route-order tests:      10 passed (previously always skipped)
```

Full `tests/unit` and the complete postgres tier were still running locally when this was opened; CI's base-vs-head matrix covers the unit tier, and I'll post both local results on the PR.

**Not included** (issue calls it a separate follow-up): making the `dev` required checks run the unit tier so harness rot cannot land green.


vybe (approved):
Validated via /validate-pr.

- Base `dev` ✅ · all 4 required checks green (Analyze python/js-ts, schema-parity, verify-non-root) ✅ · full matrix green including base+head pytest across 3 seeds
- Security scan clean: no keys/tokens, emails, public IPs, credential files, hardcoded secrets
- No new top-level `src/backend/*.py` and no new `os.getenv()` — Dockerfile COPY and compose packaging gates N/A
- Test adequacy: the PR is itself the test surface, and it ships `tests/unit/test_2080_harness_contract.py` (12 guards) pinning each defect it fixes — `tests/utils` cannot return, no non-identifier package dirs, exit-5 is a failure, audit refuses an empty log dir
- Docs: refactor tier — descriptive commit messages only, correct per the tiered-docs rule

145 files is over the split threshold, but the bulk is the mechanical `tests/utils` → `tests/testkit` rename (87 files) that is the structural fix; splitting it would leave the shadowing half-live. Accepted.

Body patched `Related to #2080` → `Fixes #2080` so the issue auto-promotes to `status-in-dev` on merge.

Issue #138, opened 2026-03-16
bug: Chat tab fails with misleading 'token expired' error — missing model default from #81 fix


webmixgamer:
## Summary

The Chat tab fails with "Subscription token may be expired or revoked. Generate a new one with 'claude setup-token'" when the agent's `~/.claude/settings.json` contains a model incompatible with the assigned subscription. The Tasks tab works fine because Issue #81 added a `model="sonnet"` default — but that fix was only applied to `execute_headless_task()`, not to `execute_claude_code()` which the Chat tab uses.

## Component

Agent Runtime — `docker/base-image/agent_server/services/claude_code.py`

## Error

```
Task execution failed (exit code -9): Subscription token may be expired or revoked. Generate a new one with 'claude setup-token'.
```

## Root Cause

Issue #81 identified that headless task execution failed when no `--model` flag was passed, causing Claude Code to fall back to the agent's `settings.json` model which might be incompatible with the subscription type.

The fix (defaulting to `model="sonnet"`) was applied only to `execute_headless_task()` (line 746), which serves the Tasks tab via `/api/task`. The Chat tab uses a different code path — `execute_claude_code()` (line 425) — which has **no model defaulting logic**. It relies on `agent_state.current_model` (line 473), which may be `None` or an incompatible model.

### Code paths

| Tab | Backend route | Agent endpoint | Function | Model default? |
|-----|--------------|----------------|----------|---------------|
| **Tasks** | `task_execution_service.py` | `/api/task` | `execute_headless_task()` | ✅ Defaults to `"sonnet"` (Issue #81) |
| **Chat** | `routers/chat.py` | `/api/chat` | `execute_claude_code()` | ❌ No default — uses `agent_state.current_model` as-is |

## Reproduction Steps

1. Create an agent and assign a Claude Max subscription
2. Ensure the agent's `~/.claude/settings.json` has an incompatible model (e.g., `claude-3-5-haiku-latest`) or no model set
3. Send a task via the Tasks tab → **works** (defaults to sonnet)
4. Send a message via the Chat tab → **fails** with "token expired" error

## Suggested Fix

Apply the same model defaulting logic from `execute_headless_task()` to `execute_claude_code()`:

```python
# In execute_claude_code(), before building the command:
if model is None and not agent_state.current_model:
    model = "sonnet"
    logger.debug("[Chat] No model specified, defaulting to 'sonnet' for subscription compatibility")
```

## Related

- #81 — Original issue (closed, partial fix)
[...]

PR #2083, opened 2026-08-10
feat(workspace): rename Client Portal to Workspace and open it in one click (ent#357)

dolho:
Related to abilityai/trinity-enterprise#357

## One click, no OTP

The backend was already there: `get_portal_identity` accepts a platform JWT and resolves it to that user's email, and the store's `authHeader` already fell back to the platform header. The **only** thing forcing an email round-trip on an already-signed-in user was one predicate:

```js
isClientSignedIn: (state) => !!state.portalToken
```

So a user holding a valid session was shown "check your email for a code" for a surface they were already entitled to — and the code authenticated them as the identity they were already carrying. Now `portalToken || platform session`, with `isPlatformSession` naming the internal case, plus a persistent **NavBar entry** gated on the `client_portal` entitlement.

That derivation is also what makes **"signing out of the platform ends it" true by construction** rather than by cleanup code: an internal workspace session *is* the platform session, so no second credential can outlive it. Relevant to the ent#281 reconciliation the issue asks for — this path adds nothing new to revoke, and the external verified-email session is untouched.

**The 401 guards followed.** They keyed on the URL prefix, which is now ambiguous — the same `/workspace` URL carries either session kind. They discriminate on the portal token instead: an external client is never bounced to the operator `/login` (the #138 rule), while an internal user **is**, because their workspace session expired along with their platform session.

## Honest status (no dead empty state)

The roster failure path collapsed three states into `No agents shared with you yet`: module absent (OSS build), module unentitled, and request failed. Two of those are dead ends dressed as an empty share list.

| state | before | after |
|---|---|---|
| 404 (absent/unentitled) | "No agents shared with you yet" | "Workspace isn't available on this instance" |
| transient failure | "No agents shared with you yet" | the error + a **Try again** |
| genuinely empty | "No agents shared with you yet" | unchanged |

## Naming

User-facing copy is **Workspace** throughout. **`room` is untouched** — it stays the API/operator word (ent#170), as do the API paths, the `client_portal` module id, and the DB. This renames what users read, not the platform's vocabulary. Per the reconciliation on ent#34, that issue's RBAC grouping takes **Teams** and this surface keeps **Workspace**.

`/portal` and `/portal/c/:sessionId` redirect query- **and** hash-preserving. Those URLs were mailed to real clients, and a client landing on a dead link has no way to report it.

## No backend change

The enterprise submodule is untouched. Worth stating explicitly: "one-click entry" sounds like it needs a token exchange, and doesn't. The trusted-issuer exchange (ent#163) exists for a licensee's **own** IdP; reaching for it here would have minted a second credential to solve what was one boolean.

## Tests

`src/frontend/tests/unit/workspaceSession.spec.js` — 11 tests, mutation-verified:

```
restore the old OTP-forcing predicate   -> 2 failed
collapse the 404 into a generic error   -> 1 failed
as shipped                              -> 11 passed
full frontend unit suite                -> 54 passed
vite build                              -> clean
```

Two harness notes, both of which first produced a vacuous pass: the auth stub uses a real Vue `ref` because Pinia getters are computeds and a plain `let` never invalidates them (the sign-out case passed for the wrong reason initially), and the redirect assertions call the **shipped** redirect functions rather than navigating, since `push()` resolves lazy `.vue` components into a suite that is node-only by design.

## Not included

- ent#281 session revocation itself (this PR only removes the need for it on the internal path)
- Renaming ent#34's RBAC "Workspace" → "Teams" — that belongs to that issue


dolho:
## /review — re-review after fixes

**Branch**: `feat/ent357-workspace` → `dev` · **9 files, +432/−38** · **Scope**: CLEAN · **Plan completion**: 6 AC — 5 DONE, 1 UNVERIFIABLE (needs a live stack)

Both findings from the first pass are **fixed in this branch**, each with a regression test. Nothing new surfaced on the second pass.

### Fixed since the first review

**[C1] `unavailable` was sticky** — `stores/clientPortal.js`
```js
this.loading = true
this.error = null        // error reset, `unavailable` did not
```
A 404 followed by a **successful retry** kept rendering *"Workspace isn't available on this instance"* over a roster that had loaded fine — and the retry button this PR adds for that state put the stale case one click away. Same class of dishonest surface the unavailable state exists to remove, so it's fixed here rather than filed. Mutation-verified: dropping the reset fails the new test.

**[I1] The 401 bounce was decided by the wrong token** — `api.js`, `main.js`

Both interceptors asked *"is there a portal token?"*. That answer is **racy**: `fetchRoster`'s 401 handler calls `signOut()`, which removes the portal token, so a second concurrent 401 saw a different world and threw an **external client** at the operator `/login` — a page they have no account for. The guards now ask *"does this browser hold a platform session that just expired?"*, which is the actual question and has a stable answer.

| who | before | after |
|---|---|---|
| internal user, expired JWT, on `/workspace` | bounced | bounced ✅ (AC: the workspace session *is* the platform session) |
| external client, first 401 | not bounced | not bounced ✅ |
| external client, **second concurrent 401** | **bounced to /login** ❌ | not bounced ✅ |
| any other route | bounced | unchanged ✅ |

### Critical findings

None.

### Clean categories

- **Auth boundaries** — the one-click path widens nothing: `/my-agents` resolves identity via `get_portal_identity` and scopes to `agent_sharing` for that email, so an internal user sees exactly the agents shared with *them*. Admins get no special roster.
- **Credentials** — no token in a URL, none logged; the internal path mints **no second credential** (that's what makes "platform sign-out ends it" structural rather than cleanup code).
- SQL/data safety (no DB code) · concurrency (no shared mutable state) · enum completeness (no new enum) · XSS (no new `v-html`) · docs (no architecture surface changed; `room` untouched, API paths and `client_portal` module id unchanged).

### Remaining informational (1)

**[I2] The nav entry is entitlement-gated; the route deliberately is not** (5/10). `/workspace` has no `requiresAuth`/`requiresEntitlement` because the external path must work without an account, so an unentitled user typing the URL gets the sign-in form and then the honest 404 state. Correct as designed — noted because it is the one place the two gates disagree.

### Verification

```
frontend unit suite      59 passed (4 files)
vite build               clean
mutations                old OTP-forcing predicate -> 2 failed
                         404 collapsed into generic error -> 1 failed
                         C1 reset removed -> 1 failed
```

**Not verified, and it is the part the issue is actually about**: no live stack was available here, so the one-click flow has not been exercised end to end against a running backend. Worth a manual pass on two points in particular — the legacy `/portal/c/<session-id>` deep link (those URLs were mailed to real clients) and the unavailable state under `TRINITY_OSS_ONLY=1` (otherwise you only learn about it from a confused customer).

vybe (approved):
SHOULD-LAND for v0.9.0 (C7). Validated — approving.

**Security** ✅ clean (no secrets/emails/IPs; no new backend module, no new `os.getenv`, so neither packaging gate applies).

**The auth-boundary change is the right call and correctly argued.** `api.js` now discriminates the 401 bounce on the **platform** token rather than the portal one. That matters: `fetchRoster`'s 401 handler calls `signOut()`, so reading the portal token made a second concurrent 401 see nothing and throw an external client onto the operator `/login`. "Does this browser hold a platform session that just expired?" has a stable answer; the old predicate did not.

**Ungating the NavBar entry is correct, not a widening.** ent#356 moved `client_portal` into OSS core — `main.py` mounts it unconditionally and nothing calls `register_module('client_portal')` — so `isEntitled('client_portal')` is now false on every OSS build and the one-click entry this issue exists to add would never render. Confirmed against architecture.md. Access is still decided server-side by `get_portal_identity`; `/workspace` staying `requiresAuth`-free is required for the external path and is not a new hole.

**Test adequacy** ✅ — 17 cases, non-happy-path throughout: portal-token-wins-over-platform (operator preview can't hijack a client tab), 404-vs-empty-roster honesty, query+hash-preserving legacy redirects, and **named regression tests for both prior `/review` findings** (C1 sticky `unavailable`, I1 bounce predicate).

Non-blocking: `Related to` means ent#357 will not auto-promote — **manually set `status-in-dev` on abilityai/trinity-enterprise#357** after merge (cross-tracker automation is same-repo only). Also note the NavBar comment's own caveat — this was relying on an enterprise registration that trinity-enterprise#374 is about to remove, so that PR is now a no-op for this link rather than a break.

PR #2084, opened 2026-08-10
feat(workspace): move client_portal from the entitled seam into OSS core (ent#356)

dolho:
Related to abilityai/trinity-enterprise#356 · enterprise-side removal PR to follow

## Why

The workspace is the main surface a non-operator uses to work with agents, and it was gated behind the paid tier — `register_module("client_portal")` + `requires_entitlement("client_portal")`, so a community build shipped the code and 404'd every endpoint. That capped adoption at exactly the population the surface exists for (Intelligence Design Weekly, 2026-08-07).

## What moved

`src/backend/client_portal/` — router, service, db, models, portal_auth — moved wholesale from the private submodule. It imports **only** OSS modules (`db.engine`, `db.tables`, `dependencies`, `models`, `services.platform_audit_service`, `utils.helpers`), so the move is self-contained. Kept as a package rather than split across `routers/`/`services/`/`db/` so the diff reads as the move it is; aligning it with Invariants #1/#2/#14 is a separate refactor if wanted.

The seven test files that covered it move to `tests/unit/` as `test_ent*_*.py`.

## What deliberately did **not** move

The `/api/enterprise/client-portal` **prefix** and the `enterprise_portal_sessions` / `enterprise_portal_messages` / `enterprise_client_blocks` **table names**. ent#83 shipped that prefix as the documented integration surface for API-only clients, and renaming the tables would force a data migration on every entitled install — and *"no data migration, no re-auth for signed-in clients"* is an acceptance criterion. Both are retained **history, not licensing**, commented as such where a reader will hit them.

## Tables: adopted, not created

DDL now lives in `db/schema.py`, versioned on the OSS two-track runner (`client_portal_tables_to_oss` + Alembic `0036_client_portal_oss`, Invariant #9). Every `CREATE TABLE`/`CREATE INDEX` on **both** tracks is `IF NOT EXISTS`, so on an install where the enterprise runner already made them this is a **no-op**; it does real work only on a fresh or community build. The Alembic `downgrade` is deliberately empty — this revision *adopted* tables holding live client conversations, and dropping data it never created is the one outcome the move promised couldn't happen.

## A latent bug this fixes

`test_agent_cleanup_parity` failed the instant the DDL landed in `db/schema.py`:

```
Agent-referencing columns missing from AGENT_REFS:
  enterprise_portal_messages.agent_name, enterprise_portal_sessions.agent_name
```

The portal tables were **never** in the delete/rename registry, because the enterprise track sits outside it. So an agent rename left portal history keyed to the old name — the client's thread simply emptied, nothing logged — and a purge orphaned the rows. Both columns are now `CASCADE`, and the rename path re-keys them (verified against a real SQLite file).

## Coupled surfaces, moved in the same change

- `enterprise_features` no longer advertises `client_portal`
- frontend de-gated: the enterprise module catalog no longer lists it; `PortalClientsPanel`'s `isEntitled('client_portal')` removed
- **`.github/workflows/enterprise-docs-guard.yml`** — its `PATTERN` forbade the token `client_portal` in `main.py`/public docs, which this change necessarily introduces. Token removed, `enterprise_portal_*` / `enterprise_client_blocks` exempted, reasons inline. Verified the seam files still scan clean.
- `architecture.md` documents the surface, now that it's legitimately public

## Verified

```
TRINITY_OSS_ONLY=1, submodule absent → 16 client-portal routes mounted
                                       enterprise_features == []
migrated portal suites                → 156 passed, 1 skipped
new OSS-core guards                   → 11 passed
cleanup-parity + entitlement-seam     → 118 passed
```

## Sequencing & open items

- **Enterprise removal** (delete the module + its `_register_module` call) is a separate PR in the private repo. Safe either order: while both exist the OSS route registers first and wins — which is the intended end state anyway.
- **abilityai/trinity-enterprise#373** (the owned-agents roster for ent#357) touches the same module. Whichever lands second needs a trivial conflict resolution; the code here already includes that change.
- **AC5 — the ent#22 line** (white-label, client usage analytics stay enterprise) is a private-tracker decision and is written up in the issue rather than here, since public docs must not catalogue paid features.


vybe (approved):
Validated via /validate-pr. Base `dev`, 29 files, full matrix green (all 6 pytest legs, base and head).

**Packaging gates, both directions.** The Dockerfile COPY gate is handled correctly and the comment cites #1033 — a new top-level backend *package* is exactly the case that builds fine and dies at import. The sibling gate was missed, so I fixed it on the branch (`b53fd2bd`): the six `os.getenv("PORTAL_*")` reads were in neither compose nor `.env.example`, and neither compose gives the backend an `env_file:`, so the `.env` lever never reaches the container — the #1056/#1039 class the comments *two lines above my insertion point* already warn about. Narrower than the usual instance and I've said so in the commit: all six have sane defaults, so the capability works and the rate limits are enforced at the secure end; what was unreachable was the tuning, not the feature. Additive, no default changed.

**Conflict resolution** (`ccd5dde6`), both additive-vs-additive:
- `learnings.md` — dev's ent#372 entry and this branch's ent#356 entry landed at the same offset. Kept both, dev's first (the file is oldest-first).
- `tests/registry.json` — rebuilt from the three merge stages rather than splicing markers. Deltas were disjoint (one add each, no removals, no modifications, no overlap). Re-encoded with the file's own `ensure_ascii` setting after my first pass rewrote every non-ASCII description; the diff against dev is now the single added entry.

**Verified locally on the merged result**, not just on CI: all 182 moved portal tests pass, and `test_agent_cleanup_parity` passes — which is the check that proves the AGENT_REFS claim in the description. That latent bug is the most valuable thing here: the portal tables were never in the delete/rename registry because the enterprise track sits outside it, so an agent rename silently emptied a client's thread. Worth noting that moving code into OSS is what *surfaced* it — the seam was hiding a real cascade gap, not just gating a feature.

**Judgement calls I checked rather than accepted:** retaining the `/api/enterprise/client-portal` prefix and the `enterprise_*` table names is right — renaming tables would force the data migration the ACs forbid, and both are commented as history-not-licensing where a reader will hit them. The empty Alembic `downgrade` is also right: the revision *adopted* tables holding live client conversations, and dropping data it never created is the one outcome this move promised couldn't happen.

**Follow-ups, neither blocking:**
1. Seven of the eight moved test files are absent from `tests/registry.json` (only `test_ent356_portal_is_oss_core.py` was added). Worth a sweep so the catalog matches the tier.
2. `abilityai/trinity-enterprise#356` will not auto-promote — the automation is same-repo only. I've put the closing keyword on ent#374, which *is* same-repo, so it closes there when the enterprise half lands.

One CI note for the record: `pytest (head, seed 12345)` first hit the 25-minute `timeout-minutes` (reported as `cancelled`) while its siblings finished in 11–13. I re-ran it rather than merging past it, since an ordering-specific hang is exactly the #2080 class; it completed in ~11 minutes. Slow runner, not reproducible.

PR #2087, opened 2026-08-10
fix(mcp): #848 keyless sign-in — hold the verified session across requests (#2035)

### LINKED ISSUE #2035: bug(mcp): #848 inline email auth — a verified session is discarded on the next request, so keyless sign-in never works
## Summary

The #848 keyless sign-in (PR #1707, merged to `dev` as `598b75f0`) does not work end-to-end. A verified email survives exactly **one** HTTP request, and every use of it requires a later request — so `verify_login` can succeed and the very next tool call still reports `login_required`.

Inert on every install today: `MCP_INLINE_AUTH_ENABLED` defaults OFF and `.env.example` ships `false`. No production impact, no security regression. This is "the feature is non-functional when enabled".

Everything else PR #1707 delivered is verified working (see Verified working below) — this is one transport-layer defect, not a bad feature.

## Reproduction

Flag ON for backend + mcp-server, an agent with the connector enabled and shared with the test address:

```
request_login(email)  → ok, code mailed
verify_login(code)    → {"status":"error","error":"no_pending_login"}
```

Work around symptom 1 by passing the email explicitly, and it gets worse — the sign-in genuinely succeeds, then evaporates:

```
verify_login({email, code}) → {"status":"ok","signed_in_as":"…","agents":["cornelius"]}
list_playbooks({agent})     → {"error":"login_required"}      ← same connection, next call
```

Reproduced with `@modelcontextprotocol/sdk`'s `StreamableHTTPClientTransport` — the transport Claude Code uses — so this is not a curl artefact.

## Root cause

#848 assumes the session's auth object is created once and mutated in place: *"FastMCP hands every tool the same object by reference, so subsequent calls on the session observe the upgrade."* True **within** a request, false **across** requests.

MCP streamable HTTP is discrete POSTs — one per tool call, all carrying the same `Mcp-Session-Id`. `mcp-proxy`'s `handleStreamRequest` (bundled in `fastmcp@4.12.1`) re-authenticates every one of them:

```js
authResult = await authenticate(req);      // EVERY post, not just initialize
if (sessionId) {
  ...
  server.updateAuth(authResult);           // installs the fresh object
}
```
```js
updateAuth(auth) { this.#auth = auth; }    // chunk-5BQXF2VT.js:685 — replace, not merge
```

`server.ts`'s `authenticate` sees no `Authorization` header and mints a **new** anonymous context (fresh `randomUUID()`), which then overwrites the one `verify_login` just upgraded.

Evidence: 6 POSTs produced 6 `MCP anonymous session opened` log lines with 6 distinct uuids while `Mcp-Session-Id` never changed.

The keyed connector path is unaffected because its credential is on the wire — the client resends `Authorization` on every POST, so re-authentication rebuilds an identical context. Only the keyless tier has nothing to rebuild from.

## Why CI was green

The 30 MCP tests invoke tool handlers directly against a context object the test constructs, so the object is never swapped — the replacement exists only in the transport layer. PR #1707's own test plan left exactly one box unticked: *"Manual: set `MCP_INLINE_AUTH_ENABLED=true`, connect a keyless client, sign in end-to-end."* That is the box that catches this.

Suggested regression test: drive a real `StreamableHTTPClientTransport` client through request → verify → a connector tool, asserting the third call is authorized. A handler-level unit test cannot catch this class.

## Fix options

**Option A — memoize the anonymous context by `mcp-session-id`.** `authenticate` returns the *same* object for a given session, so `updateAuth` reinstalls rather than replaces it.

- Requires: anonymous tier **only** (a keyed session must keep re-validating, else the same trick yields a keyed context without the key); never consult the memo when an `Authorization` header is present; a TTL, since fastmcp gives that closure no session-close hook (`auth.ts` already notes this and has the bounded-map + oldest-first eviction pattern).
- Security implication: **`mcp-session-id` becomes bearer-equivalent** for keyless sessions. It is a 122-bit v4 UUID, but it is not currently handled as a secret — plain header, and it is logged. Adopting A means reclassifying it as a credential and auditing anything that logs or displays it.
- Blast radius is bounded by #848's own design: the backend re-gates **authorization** per call (`email_has_agent_access` + connector-enabled), so a hijacked session confers the asserted identity, not frozen permissions — unshare the agent and the next call 403s. What gets cached is authentication, not authorization.

**Option B — `verify_login` returns a short-lived opaque token** the caller presents on later calls.

- The credential is back on the wire, and the whole class of bug disappears.
- But MCP gives a tool no way to make the client attach a header to later requests — transport headers are static config. So the token can only travel as a **tool argument**, i.e. it lives in the calling model's context window on every call. That is worse than A on exposure, and it is the "minted key" shape the #848 sign-off explicitly ruled out.

**Recommendation: A.** The session id is already carried automatically by the client transport — invisible to both the human and the model — which is precisely what the sign-off's "session, not a minted key" wanted. B reintroduces a user/model-visible credential.

**This needs a decision, not just a patch.** "Session, not a minted key" is from the #848 sign-off (Eugene, 2026-06-28, quoted in PR #1707). If session-id-as-bearer is not acceptable, that decision is what has to be reopened — @vybe's call before anyone builds.

## Verified working (do not re-litigate)

Confirmed live this session against the merged code, flag ON:

- Internal-secret gate: 403 with no secret and with a wrong secret
- `/request` enumeration safety: byte-identical `{"status":"ok"}` 202 for known vs unknown, 12ms vs 14ms; unknown address suppressed with no mail sent
- `/verify`: 401 wrong code → 200 correct code with the agent list
- `/playbooks` access gate: 403 for a non-shared agent **and** for an unknown email, i.e. the internal secret alone cannot reach an agent the asserted address cannot
- `/chat`: real execution, correct response
- Idempotent replay: same email + key → same `execution_id`, 37ms, no second execution
- Cross-user scope (`50c3fd11`): different email + **same** key → different execution, no snapshot leak
- Anonymous tool visibility (`87b2abb3`): exactly `request_login, verify_login, ask, list_playbooks, run_playbook` — zero operator tools
- AC6 UI: `inline_auth_available: true` + 4 keyless snippets, offered with no key minted
- Keyed connector path: unaffected, playbooks resolve

The backend half is sound. The break is solely the MCP server failing to hold a verified identity between two calls.

## Release-cut note

#848 must **not** be listed in a release PR's `Closes` lines until this ships — same handling PR #1839 gave #1819. #848 is being moved back to `status-in-progress` for that reason.

Refs #848, PR #1707

obasilakis:
## Summary

#848 keyless MCP sign-in never worked end-to-end. A verified email survived exactly one HTTP request, so `verify_login` reported success and the very next tool call answered `login_required`.

#848 upgrades a session by mutating its auth context in place, on the premise that FastMCP hands every tool the same object by reference. That holds **within** a request and is false **across** requests: MCP streamable HTTP is discrete POSTs, `mcp-proxy` re-runs `authenticate` on every one, and `FastMCPSession#updateAuth` is `this.#auth = auth` — replace, not merge. A callback returning a fresh context per request therefore discarded every login before anything could read it.

**Option A**, approved by @vybe on #2035: return the *same* context object for the same transport session, so `updateAuth` reinstalls it. Option B (an opaque token from `verify_login`) was rejected because MCP gives a tool no way to make a client attach a header to later calls — the token would ride as a tool argument and live in the calling model's context window, which is worse on exposure and the "minted key" shape the #848 sign-off ruled out.

Flag-gated `MCP_INLINE_AUTH_ENABLED`, default OFF. This is "feature non-functional when enabled", not a live vulnerability.

## Acceptance criteria (@vybe's five conditions)

| # | Condition | |
|---|---|---|
| 1 | Reclassify `mcp-session-id` as a credential; audit anything that logs it | ✅ |
| 2 | Short TTL on the memoized context — hours not days, err aggressive | ✅ 30 min idle / **4 h** absolute (was 8 h) |
| 3 | Memoization strictly anonymous-tier; needs its own test | ✅ source guard + behavioural |
| 4 | Real-transport regression test | ✅ |
| 5 | Docs note: keyless tier behind TLS only | ✅ |

### 1 — the session id is now a credential

Making it bearer-equivalent means it must not be written to a log file. Trinity's own code already logged an independent correlation id, but the bundled **mcp-proxy prints the raw id at INFO on three paths** — including the SSE line that fires on *every* connection — and Vector ships container stdout to `/data/logs`.

Measured on this machine before the fix: **331 full session ids sitting in `/data/logs` in plaintext.**

FastMCP accepts a `logger` and looks like the clean seam, but those sites are bare `console.log` and mcp-proxy takes no logger of its own, so there is no injection point. Redaction happens at the console boundary instead (`src/log-redaction.ts`): UUID-shaped tokens in `[mcp-proxy]`-prefixed lines are truncated to 8 chars — enough to keep two lines for one connection joinable, leaving 90 bits to guess online inside the TTL. Vector-layer redaction was rejected for leaving the id in Docker's own json log.

Scope is deliberately narrow so ids Trinity logs on purpose keep their full value. That narrowness is **guarded, not assumed**: a static scan of the installed mcp-proxy *and* fastmcp bundles fails CI if either grows a session-id log site the prefix does not cover.

### 2 — TTL, and why it is the only bound

`disconnect`-eviction reads as available (fastmcp emits `disconnect` with the session; `FastMCPSession` exposes a `sessionId` getter) and **is not**. Probed against this tree rather than assumed:

- `connect` fires with `sessionId` still unset — the id is copied off the transport, which has not minted one at `initialize`
- `disconnect` never fires for a client that simply quits — the MCP SDK's `close()` only aborts the SSE stream, and the DELETE that ends the server-side session is sent solely by the opt-in `terminateSession()`

With no `logout` tool either (decided against, see Known limitations), **expiry is the only thing that ends a keyless session**, so the absolute cap *is* the exposure window for a leaked id. Halved to 4 h accordingly. The test pins the direction, not the value — lowering is always safe, raising weakens every keyless session at once.

### 3 — anonymous-tier only

Tool visibility cannot serve as the observable: `updateAuth` sets `#auth` and stops, so the callable tool map stays the one built at session creation and a mid-session identity change is invisible in `tools/list`. Two guards instead:

- **Structural** (`inline-auth-scope.test.ts`) — the store has exactly one call site, inside the no-`Authorization` branch, ahead of the line where a presented key is read. A behavioural test only samples the paths it drives; this fails the moment a second call site appears anywhere.
- **Behavioural** — a keyed session re-validates its credential against the backend on every request, *including* when it presents a session id that has a live, signed-in anonymous entry. That is the precise shape the condition forbids.

### 4 — the regression test crosses the transport

`inline-auth-transport.test.ts` drives a real `StreamableHTTPClientTransport` against a stubbed backend across separate POSTs. #848's 30 tests were all handler-level — they invoke `execute` with a context the test owns, so the object is never swapped and this bug class is invisible by construction. That is why it shipped green, and it is why this file must not be replaced with a handler test.

## Verification

- **236/236 tests, `tsc --noEmit` clean**, after merging 44 commits of `dev` (incl. the fastmcp 4.12.1→4.12.2 bump).
- **Negative controls, both discriminating:**
  - revert the memo → exactly **3** transport cases fail, the other 7 pass
  - disable the redaction install → exactly **1** fails
- **Live end-to-end against the real stack**, post-merge, image rebuilt from this branch:
  `list_playbooks` → not signed in → `request_login` → `verify_login` **with the code alone** → signed in → `list_playbooks` returns real playbooks. The correlation id is now one stable value across all five POSTs; #2035 recorded six distinct ones for six POSTs.
- **Redaction confirmed in the captured log**: `for session ID f7ccb092-REDACTED` in `/data/logs`, no full id anywhere.

The 331 pre-existing full ids in `/data/logs` are all dead — any upgrade restarts the mcp-server, which drops every keyless session — so no live credential survives the deploy.

## Also in this PR

Refreshes the `fastmcp@4.12.1` citations in `server.ts` and `tool-visibility.test.ts` to the 4.12.2 `dev` bumped to in #2003. Every cited line number was re-verified byte-identical under 4.12.2; only the version string had rotted, which is the exact failure mode those comments warn about.

## Known limitations (deliberate)

- **No `logout` tool.** Decided against for this PR. Nothing ends a keyless session early — expiry or an mcp-server restart only. Documented rather than hidden.
- **No disconnect eviction.** Not deferred — *not implementable*; see condition 2 above.
- **`/clear` in an MCP client does not log out.** It resets the conversation, not the transport session. Now documented, since the session outliving a `/clear` is surprising.

## Release-cut note

**#848 must stay out of any release PR's `Closes` lines until this ships** — it is deliberately back on `status-in-progress` for that reason. Same handling PR #1839 gave #1819.

Closes #2035
Refs #848


vybe (approved):
Validated via /validate-pr.

- Base `dev` ✅ · 13 files ✅ · all 4 required checks green, full matrix green ✅
- `Closes #2035` present → auto-promotes on merge. `Refs #848` is deliberately **not** a closing keyword, matching the release-cut note — #848 stays out of the release `Closes` lines until the keyless tier actually works. Correct call, and the #1839/#1819 precedent is the right one.
- Docs tier is right for a behaviour change: `architecture.md` + `feature-flows/mcp-connector.md` + `requirements/mcp.md` all moved, plus the `.env.example` TLS-only warning that makes the new credential's exposure model discoverable at the config surface.
- Security scan clean. The one `API_KEY = "trinity_mcp_test_key"` hit is a fixture in `*.test.ts` beside `someone@example.com`. `MCP_INLINE_AUTH_ENABLED` was already wired in both compose files — no packaging gap.
- Test adequacy (fix PR): `inline-auth-transport.test.ts` drives a real `StreamableHTTPClientTransport` across separate POSTs, which is the only shape that can see this bug — #848's 30 handler-level tests owned the context object, so the class was invisible by construction. Naming that in-file is what stops it being replaced with a handler test later. `inline-auth-scope.test.ts` fails on a *second* call site rather than sampling paths, which is the stronger guard.

Two things I checked rather than took on trust: the disconnect-eviction claim (`connect` fires before `sessionId` is minted, and `close()` never sends the DELETE) is why TTL is the only bound, so the 4h absolute cap really is the exposure window — halving it was the right response. And the 331 plaintext session ids already in `/data/logs` are dead on any upgrade, so no live credential survives the deploy.

Issue #913, opened 2026-05-22
bug: scheduled runs ignore per-agent execution_timeout_seconds (canary S-03 + E-01 surfaced)


obasilakis:
## Summary

The canary invariant harness (CANARY-001, #884) consistently fires **S-03** (slot TTL below floor) and **E-01** (terminal-state closure) on scheduled executions whenever `agent_ownership.execution_timeout_seconds` is set to anything other than the schedule default. Both invariants are surfacing the same underlying drift: scheduled runs never use the per-agent timeout — they always use `agent_schedules.timeout_seconds`, default `900`s.

## Component

Backend / Scheduler / Slot Service / Cleanup Service

## Priority

P2 — `PUT /api/agents/{name}/timeout` is documented as the per-agent execution-timeout control (`/api/agents/{name}/timeout` endpoint, TIMEOUT-001 migration) but is **silently ineffective for scheduled runs**. Manual `/chat`, `/task`, inter-agent (#418), backlog drain, and session endpoints all honor it correctly. Only the scheduler path is broken. Same class of bug as #418, related to #226 (closed) and #869 (open, status-in-dev).

## What the canary surfaces

On any agent with `execution_timeout_seconds != 900`:

- **S-03 below_floor** — slot HASH stores `timeout_seconds=900` (the schedule value), so slot TTL becomes 1200s. Canary computes floor from per-agent value (e.g., 3600+300=3900s). HASH TTL of ~1200s < floor of 3900s → S-03 fires.
- **E-01 terminal-state closure** — a row that runs past the per-agent threshold (`exec_timeout + 300s`) but is still within the per-schedule threshold (`900 + 300s = 1200s`) is alive at canary cycle time. Canary uses per-agent; cleanup uses per-schedule. E-01 fires; cleanup leaves the row alone.

Both invariants resolve to one root cause.

## Root Cause

`src/scheduler/service.py:805` passes `schedule.timeout_seconds` (loaded from DB) to backend's `/api/internal/execute-task`:

```python
result = await self._call_backend_execute_task(
    ...
    timeout_seconds=schedule.timeout_seconds,
)
```

The TIMEOUT-001 migration docstring (`src/backend/db/migrations.py:801-806`) states the design intent:

> "All execution paths (task API, chat, **scheduler**, MCP, paid endpoints) read from this setting when no explicit timeout is provided."

`src/backend/services/task_execution_service.py:281-282` has the fallback:

```python
if timeout_seconds is None:
[...]

PR #2088, opened 2026-08-10
fix(canary): S-03 must read `ttl` and `age` at the same instant (abilityai/trinity-enterprise#372)

obasilakis:
## Summary

- S-03 reconstructs a slot's initial TTL as `ttl + age`, but the two terms were read seconds apart — `ttl` from the per-slot Redis pipeline that runs **last** in `collect_snapshot()`, `age` from `snapshot_time`, stamped **first**, before the docker and roster collectors. That elapsed time (1.0–1.9s measured on eu2, tail to 9s) is ticked off `ttl` without being added to `age`, so every healthy slot reconstructed short and fired `below_floor`: **208 critical violations / 85 cycles / 24h on eu2 — every cycle that ran**, 9 agents, 100% `below_floor`, 31 Slack pages.
- `_collect_redis_slot_state` now stamps `AgentSnapshot.slot_ttl_read_at[eid]` immediately before each slot's TTL pipeline; S-03 derives `age` from that. Both terms describe one instant, and the existing 1s tolerance covers only what its comment claims (Redis's float→int wire rounding).
- Blocks trinity#1766 soak milestone M7 (`docs/testing/PULL_MIGRATION_TESTING.md` §9, "canary green throughout") — a permanently-red critical invariant leaves nothing to baseline against.

## Design notes

- **Per slot, not per collector.** The TTL loop is O(slots) and its own elapsed time grows with the fleet, so one collector-level timestamp would re-open the identical gap at a larger fleet — the same argument the issue used to reject widening the tolerance.
- **Stamped before `execute()`**, so the server evaluates the TTL at or after that instant: `age` can only be *under*-counted, by one RTT. That is the direction that can never manufacture a green, and it sits far inside the 1s tolerance.
- **No fallback to `snapshot_time`.** A TTL without its read time is unreachable at runtime (both are written in one try-block); the floor arm skips instead, matching the module's existing stance for a missing score or a missing stored timeout. The fallback is the bug.
- **Rejected alternatives** (per the issue): widening the tolerance (collector elapsed time is unbounded), re-stamping `snapshot_time` (other invariants measure against it), reordering the collectors (#1813 puts docker first so H-01 has independent evidence on the roster-failure path), and anything in `slot_service.py` (the acquire path was measured at a 0.8 ms score→EXPIRE gap and is not the cause).

## Changes

| File | Change |
|---|---|
| `src/backend/canary/snapshot.py` | New `AgentSnapshot.slot_ttl_read_at`; `_collect_redis_slot_state` stamps and returns it; `collect_snapshot` wires it through |
| `src/backend/canary/invariants/s03_slot_ttl_floor.py` | `age` from the per-slot read time; `snapshot_time` parse (now unused) removed; docstring records the bias and the rejected alternatives |
| `tests/unit/test_ent372_s03_snapshot_read_time.py` | New — fake clock, stubbed slow docker collector, real `collect_snapshot()` |
| `tests/unit/test_ent336_s03_slot_timeout_floor.py` | Fixture stamps the read time at `T0` (its cases are unchanged in intent) |
| `docs/memory/architecture.md` · `docs/memory/learnings.md` | S-03 catalog row corrected; durable entry for the two-readings-one-cycle class |

## Test Plan

- [x] `pytest unit/test_ent372_s03_snapshot_read_time.py unit/test_ent336_s03_slot_timeout_floor.py` — 32 passed
- [x] `pytest test_canary_invariants.py` — 152 passed, 2 skipped (20 pre-existing errors from this file's `utils.helpers` stub shadowing `utils.safe_yaml`, present on `dev`)
- [x] `pytest unit/ -k "canary or slot or s03"` — 264 passed, 3 skipped
- [x] Regression is proven, not assumed: `test_the_fixture_really_reproduces_the_old_deficit` recomputes the **pre-fix** formula on the same snapshot and asserts it lands under the floor — the suite was previously green only because every fixture sampled both readings at one instant
- [x] `-1` / `-2` sentinel arms, ent#336's stored-timeout floor, #913 natural decay, and #1813 collector ordering each covered by a test
- [ ] **Deploy-time (AC, unverifiable pre-merge):** eu2 — a cycle with live slots produces zero S-03 rows

Fixes abilityai/trinity-enterprise#372


vybe (approved):
Validated via /validate-pr.

- Base `dev` ✅ · 6 files ✅ · all 4 required checks green, no non-success checks ✅ · security scan clean
- `Fixes abilityai/trinity-enterprise#372` is the **correct** cross-tracker shape for a private-tracker bug fixed in core code — not a validation failure. It resolves (P2 / type-bug / theme-reliability). Consequence noted below.
- Docs tier is right: the S-03 catalog row in `architecture.md` is corrected at the same time as the code, so the doc can't keep asserting the agent-cap floor after the floor moved.

What makes this one convincing rather than merely plausible: `test_the_fixture_really_reproduces_the_old_deficit` recomputes the **pre-fix** formula against the same snapshot and asserts it lands under the floor. Without that, a green suite proves nothing here — every prior fixture happened to sample both readings at one instant, which is exactly why 208 critical violations shipped under a passing test file.

The per-slot (not per-collector) stamp is the right granularity for the stated reason — a collector-level timestamp reopens the same gap once the TTL loop's own elapsed time grows with the fleet, which is the argument that already ruled out widening the tolerance. And stamping *before* `execute()` biases `age` low by one RTT, so the error direction can only under-count; a fix that could manufacture a green here would be worse than the bug.

Two follow-ups, neither blocking:
1. **ent#372 will not auto-promote** — `issue-status-on-merge.yml` is same-repo only. It needs a manual `status-in-dev` after merge; I'll do it.
2. The test-plan note about 20 `utils.helpers`/`utils.safe_yaml` shadowing errors "present on `dev`" is now stale — #2082 landed the `tests/utils` → `tests/testkit` rename an hour ago and removed that shadowing. Nothing to change here; flagging so the number isn't re-quoted as a known-bad baseline.

PR #2090, opened 2026-08-10
docs(screenshots): one canonical, manifest-guarded screenshot store (#2089)

### LINKED ISSUE #2089: docs: consolidate three drifting screenshot stores into one manifest-guarded store
## Problem

UI screenshots live in three separate stores, and they drift because **the capture tooling writes to the one directory no page embeds**:

| Location | Files | Last refreshed | Who reads it |
|---|---|---|---|
| `docs/screenshots/` | 31 | **2026-08-09** (freshest) | **nobody** |
| `docs/user-docs/images/` | 19 | mostly **2026-05-11** | 12 user-doc pages |
| `docs/assets/screenshots/` | 8 | 2026-07-08 | README (4), whats-new v0.8.0 (4) |

So this isn't a refresh-discipline problem — `/capture-screenshots` lands in a dead end while the rendered images age. Secondary findings:

- 4 exact-md5 duplicate pairs across `user-docs/images` ↔ `assets/screenshots` (~4.5 MB of identical bytes under two names).
- 4 orphaned user-doc images (incl. `dashboard-graph-view.png` — Graph mode was decommissioned in #1689).
- The numeric `NN-` filename prefixes are why nothing can safely embed the fresh set: inserting a screen renumbers it and silently breaks every reference.
- `docs/user-docs/images/agent-overview-tab.png` (embedded from `managing-agents.md`) is a v0.7.0 capture of a **real** agent — it shows a real person's photo, `$2377.70` lifetime spend, and personal knowledge-base counts, in a public repo.
- The capture skill had drifted into a third copy of the truth: its hardcoded step list still named `/operating-room`, `/monitoring`, and `/api-keys` after those routes were consolidated, and its filenames no longer matched what was on disk.

## Proposal

One canonical store, `docs/screenshots/`, with a manifest and a CI guard.

1. **Single store** — repoint all 20 doc embeds + 5 README refs at `docs/screenshots/`; delete the other two screenshot stores.
2. **Stable slug filenames** — drop the `NN-` prefixes.
3. **`MANIFEST.yaml`** — one entry per screenshot: `route`, `shows` (the recapture brief), and `sources` (frontend files whose change makes it stale), plus `captured_at` / `captured_commit`.
4. **`scripts/ci/check_screenshots.py` + `screenshot-guard.yml`** — blocks on integrity (manifest/file mismatch, broken embed, image referenced from outside the store, a `sources` path that no longer exists); warns on staleness, since a UI PR legitimately makes a screenshot stale in the same commit that changes the view.
5. **Capture skill becomes manifest-driven** — no second list to drift.
6. **Frozen exception** — `docs/assets/screenshots/whats-new/<version>/` stays as point-in-time release snapshots, deliberately never refreshed.

## Acceptance criteria

- [ ] Exactly one store holds UI screenshots; no doc embeds an image from anywhere else
- [ ] Every screenshot has a manifest entry; every manifest entry has a file
- [ ] Guard fails on each integrity class (verified by negative test, not assumed)
- [ ] Staleness is reported per-screenshot with the date its UI changed
- [ ] Real-fleet capture removed from the public repo
- [ ] Adding a screenshot is one manifest entry

trinity-ability:
rewritten
reenshots lived in **three** stores, and they drifted because the capture tooling wrote to the one directory no page embedded:

| Location | Files | Last refreshed | Who reads it |
|---|---|---|---|
| `docs/screenshots/` | 31 | **2026-08-09** (freshest) | **nobody** |
| `docs/user-docs/images/` | 19 | mostly **2026-05-11** | 12 user-doc pages |
| `docs/assets/screenshots/` | 8 | 2026-07-08 | README (4), whats-new (4) |

So `/capture-screenshots` refreshes landed in a dead end while the rendered images aged. Not a discipline problem — a broken wire.

## What changed

**One store** — `docs/screenshots/`, flat, stable slug filenames. All 20 doc embeds and 5 README refs repointed; the other two stores deleted. The `NN-` prefixes are gone: they're why nothing could safely embed the fresh set, since inserting a screen renumbers it and silently breaks every reference.

**`MANIFEST.yaml`** — one entry per screenshot with `route`, a `shows` recapture brief, and `sources`: the frontend files whose change makes it stale. `sources` is the load-bearing field — it makes freshness answer *did the thing in the picture change*, rather than nagging on age.

**`scripts/ci/check_screenshots.py` + `screenshot-guard.yml`** — blocking integrity, advisory staleness.

**Capture skill is now manifest-driven** (separate `.claude` submodule commit — see below). Its hardcoded list had itself become a third drifting copy: it still named `/operating-room`, `/monitoring`, `/api-keys` after those routes were consolidated, and its filenames no longer matched disk.

## Security note

This removes a **real-fleet capture from a public repo**. `docs/user-docs/images/agent-overview-tab.png`, embedded from `managing-agents.md`, was a v0.7.0 shot of a live agent showing a real person's photo, `$2377.70` lifetime spend, and personal knowledge-base counts. The fresh replacement is a demo fleet. Both the store README and the capture skill now state the demo-data rule explicitly.

## Why staleness doesn't block

A frontend PR legitimately makes a screenshot stale in the same commit that changes the view. Blocking there would force a recapture into every UI PR and get routed around within a month. So integrity blocks, staleness warns, and a weekly `--strict` run turns the list into recapture work.

The guard currently reports **6 stale** — real signal, each naming the date its UI changed:

| screenshot | captured | UI changed |
|---|---|---|
| `mcp-api-keys.png` | 2026-06-01 | 2026-08-09 |
| `agent-tasks.png` | 2026-06-01 | 2026-08-03 |
| `agent-folders.png` | 2026-06-01 | 2026-07-31 |
| `agent-sharing-access.png` | 2026-06-01 | 2026-07-31 |
| `brain-orb.png` | 2026-07-08 | 2026-07-18 |
| `agent-permissions.png` | 2026-06-01 | 2026-07-10 |

These are the legacy images with no fresh equivalent. Carried over with their true dates rather than dropped, so they're tracked instead of invisible. Recapture is a follow-up.

## Verification

The guard's failure modes were each **negative-tested**, not assumed:

| Fixture | Result |
|---|---|
| `--strict` with 6 stale | exit 1 |
| Unclaimed image in store | error: `no MANIFEST.yaml entry` |
| Broken doc reference | error: `broken image reference` |
| `sources` path renamed away | error: `source path does not exist` |
| Image embedded from outside the store | error: `outside the canonical store` |
| Clean tree | exit 0 — 41 screenshots, 26 references resolved |

That fourth one is the subtle case: freshness over a nonexistent path yields no commits, i.e. the entry silently becomes immortally "fresh". A guard whose own inputs can rot without complaint is worse than none.

Two workflow bugs caught while writing it: `python … | tee` reports `tee`'s exit status under GitHub's default non-pipefail shell, so a failing guard would have passed — switched to `--json-out`; and staleness needs real history, so `fetch-depth: 0` (a shallow clone now reports "not evaluated" rather than passing silently).

## Not in this PR

- **WebP re-encode** — the convention for *new* captures is documented, but the existing PNGs aren't re-encoded: git already holds them, so converting reclaims nothing while producing a large binary diff.
- **Recapturing the 6 stale images** — needs a running instance with a demo fleet.

## Also dropped `docs/diagrams/marketing/`

10 jpegs (4 MB, Jan 2026) with **zero references** anywhere in the repo — removed on request in `765579a3`, completing the cleanup the first commit had left open. They were only ever reachable by raw URL, and git history still holds them if a marketing asset needs recovering.

That commit also **tightens the guard's allowlist** by removing `docs/diagrams/`. Every prefix in that list is somewhere a fourth store can quietly grow, and `docs/diagrams/` now holds no image files at all — it's Mermaid-in-markdown. Verified by negative test: an image referenced from under `docs/diagrams/` is now an integrity error.

## Companion skill change

The capture skill lives in the private `.claude` submodule. Its commit `d49a212` rewrites `capture-screenshots` to iterate the manifest, stamp `captured_at`/`captured_commit`, and run the guard before finishing — and the pointer bump rides this PR in `a3c8aabc`, so the skill change lands with the manifest it depends on.

Closes #2089


vybe (approved):
Approving after validation (`/validate-pr`).

The consolidation is the right shape: the root cause was structural — the capture tooling wrote to the one directory no page embedded, so refreshes never reached the rendered docs. A single store plus a manifest whose `sources` field ties each image to the frontend files that make it stale fixes the mechanism, not just the symptom.

Two things I checked closely:

**Test gap, caught and closed.** The guard originally shipped without a unit test, while both other `scripts/ci/` guards have one — verification was ad-hoc shell fixtures, which stop being true the moment the script is edited. `3956c4ed` adds 27 tests, three of which lock the live tree rather than a fixture, plus pins on the no-pipe and `fetch-depth: 0` couplings that would otherwise silently disable the CI job. That was the one blocking finding.

**Security improvement, not just neutrality.** This removes a v0.7.0 capture of a real agent from a public repo — a real face, $2377.70 lifetime spend, and personal knowledge-base counts, embedded from `managing-agents.md`. The demo-data rule is now written into both the store README and the capture skill.

Accepting the two non-blocking warnings: 85 files is inherent to a rename-heavy consolidation (74 are pure renames), and the 6 stale screenshots are pre-existing rot now tracked instead of invisible.

PR #2095, opened 2026-08-11
fix(schedules): pair the owner gate with the path param it reads (#2094)

### LINKED ISSUE #2094: bug: schedule enable/disable/trigger always 422 — OwnedAgentByName on {name} routes (regression from #2081 CSO M1)
## Summary

`POST /api/agents/{name}/schedules/{schedule_id}/enable`, `.../disable`, and `.../trigger` unconditionally return **422 Unprocessable Entity** for every caller. The three handlers in `src/backend/routers/schedules.py` (lines 406, 425, 444) declare their agent parameter as `name: OwnedAgentByName`, but `OwnedAgentByName` resolves through `get_owned_agent_by_name()` (`src/backend/dependencies.py:933`), which reads `agent_name: str = Path(...)`. These routes are declared with a `{name}` path segment — there is no `{agent_name}` — so FastAPI can never satisfy the dependency and rejects the request before the handler runs:

```json
{"detail":[{"type":"missing","loc":["path","agent_name"],"msg":"Field required"}]}
```

This breaks schedule enable/disable/manual-trigger for **every surface**: the REST API, the frontend Schedules tab toggle/Run-now buttons, and the MCP tools `toggle_agent_schedule` / `trigger_agent_schedule`.

## Context

Regression introduced by #2081 (`fix(security): CSO 2026-08-09 — terminal-tier HIGH + access-control + supply-chain hardening`), item **M1**: "move schedule enable/disable/trigger to OwnedAgentByName (matching update/delete) so shared users can't flip owner-intent schedule state." The intent (owner-tier gate instead of accessor-tier) is correct; the wrong owner-tier dependency variant was picked. The two variants exist precisely for the two path-param spellings (Architectural Invariant #8):

- `OwnedAgent` → `get_owned_agent(name: str = Path(...))` — for `/{name}/...` routes ← what these routes need
- `OwnedAgentByName` → `get_owned_agent_by_name(agent_name: str = Path(...))` — for `/{agent_name}/...` routes

Caught by the full test suite (7 failures: `tests/test_schedules.py::test_enable_schedule` / `test_disable_schedule` / `test_trigger_schedule` / `test_trigger_nonexistent_schedule`, `tests/test_model_selection.py::TestExecutionModelTracking::test_trigger_schedule_tracks_model_used`, `tests/test_817_subprocess_leak.py` ×2) and verified live with a direct `curl` against a real schedule — the 422 reproduces independently of pytest. Note the required CI checks on `dev` don't run these tests, which is how it landed green.

## Acceptance Criteria

- [ ] `enable_schedule`, `disable_schedule`, `trigger_schedule` use the owner-tier dependency matching their `{name}` path param (`OwnedAgent`), keeping the M1 intent: owner/admin only, shared accessors get 403-tier denial, non-existent/inaccessible stays uniform (Invariant #8)
- [ ] `POST /{name}/schedules/{id}/enable|disable|trigger` return 200 for the owner against a real schedule (no 422)
- [ ] The 7 failing tests above pass again
- [ ] A regression guard exists for the dependency/path-param pairing (e.g. a unit test asserting no route declares `OwnedAgentByName` without an `{agent_name}` segment and vice versa) so the next `OwnedAgent`↔`OwnedAgentByName` swap can't ship silently

## Technical Notes

- Fix is one line per route at `src/backend/routers/schedules.py:406,425,444`: `OwnedAgentByName` → `OwnedAgent`.
- `trigger_schedule` also declares `current_user: User = Depends(get_current_user)` separately — unaffected, but worth keeping in mind when touching the signature.
- Every other handler in the file uses the `{name}`-pattern dependencies (`AuthorizedAgent`/`OwnedAgent` family), so the fix restores file-internal consistency.
- The pairing-guard test in AC4 would also protect the ~40 other routers that use these Annotated dependencies.

dolho:
Fixes #2094

## What

`enable`/`disable`/`trigger` on a schedule returned **422 to every caller**, owner included — the REST API, the frontend Schedules tab toggle and Run-now, and the MCP `toggle_agent_schedule` / `trigger_agent_schedule` tools.

#2081 (CSO M1) moved these three routes to owner-tier so a shared user can't flip owner-intent schedule state. Right intent, wrong variant: `OwnedAgentByName` resolves `get_owned_agent_by_name(agent_name: str = Path(...))`, and these routes declare `{name}`. FastAPI can never satisfy that dependency, so the request dies in validation before the handler runs.

The two variants exist for exactly the two path-param spellings, so this keeps M1's **tier** and changes only the spelling. `OwnedAgent` is the same owner/admin check reading `name`.

## Verified against a running backend

Two throwaway containers on the real Postgres/Redis — one on unpatched `dev`, one on this branch — same request:

```
dev (unpatched)  enable   422  [{"type":"missing","loc":["path","agent_name"],...}]
dev (unpatched)  disable  422  [{"type":"missing","loc":["path","agent_name"],...}]
dev (unpatched)  trigger  422  [{"type":"missing","loc":["path","agent_name"],...}]
fix  (#2094)     enable   404  'Agent not found'
fix  (#2094)     disable  404  'Agent not found'
fix  (#2094)     trigger  404  'Agent not found'
```

Byte-identical to the payload in the issue on the left; the uniform 404 of Invariant #8 on the right. Both containers torn down afterwards; the local stack was untouched.

## Two guards (AC4)

One is not enough, because they fail differently.

**`test_2094_dependency_path_param_pairing.py`** — static, covers all **146** uses of the four aliases across every router. Static *by design*: `test_1483_route_order.py` documents what importing `main` costs — it skipped ALWAYS, standalone included, for months behind a message blaming the wrong cause. A guard that can silently skip is not a guard. It also asserts the scan found >100 pairings and all four aliases, since a guard matching nothing certifies nothing.

**`test_2094_schedule_control_routes_bind.py`** — functional, drives the real router through `TestClient`. A rule about annotations is a proxy for "the route works"; this is the thing itself. It also pins that the OWNER gate still resolves, so the tempting wrong fix — drop to `AuthorizedAgent`, 422 disappears, #2081's hole comes back — fails here.

Mutation-tested rather than asserted:

| mutation | result |
|---|---|
| unpatched `dev` | **7 failed** |
| accessor-tier "fix" | caught |
| fixed | 11 passed |

## Why this shipped green in the first place

Worth stating, because it's the reusable part. The swap is invisible at every layer that normally catches things: it reads as a *tightening* (one owner gate for another — exactly what the security review asked for); Python is happy, both names import; the failure is a 422, so it looks like a client problem; and it's **uniform** — no works-for-owner/fails-for-shared asymmetry to notice. Plus the required checks on `dev` don't run the tier that caught it.

## Two things the functional test had to survive

My first draft mounted the router with a prefix it already carries, double-prefixing every request into Starlette's generic 404 — which a loose `"not found"` substring happily matched. **Both tests passed against broken code.** It now asserts the handler's exact `"Schedule not found"`.

Dependency identity is compared by qualified *name*, not object: `dependencies` is on conftest's #762 invariant-restore list, so a fresh import can hand back a different module object than `routers.schedules` bound at its own import.

## Not run

The 7 integration tests the issue names. They need `TEST_AGENT_NAME` and **skip** rather than fail, and the isolated backend has no Docker socket so it sees no agents. The live contrast above covers the same ground more directly — flagging it rather than letting "4 skipped" read as a pass.


dolho:
## `/review` — self-review

**Branch**: `fix/2094-schedule-owned-agent` → `dev` · **merge-base** `f1809975`
**Files**: 4 (+472/−4) · product delta is 4 lines in one router
**Scope**: CLEAN — no file outside the stated fix + its guards
**Plan completion**: 3 DONE · 1 UNVERIFIABLE (AC3, see below)

Reviewing my own PR, so I went looking for reasons it was wrong rather than reasons it was right. That turned up **one real defect — in the guard, not the fix.**

---

### [C1] Test gap: the pairing guard was blind to 40 of 186 routes (Confidence: 10/10) — FIXED in `244ec096`

These four gates are attached **two** ways, both in live use:

```python
name: OwnedAgent                        # 146 Annotated aliases
name: str = Depends(get_owned_agent)    #  40 Depends defaults
```

`_agent_dependencies()` read only `ast.Name` annotations:

```python
ann = arg.annotation
if isinstance(ann, ast.Name) and ann.id in DEPENDENCY_PARAM:
```

So it scanned 146 of 186 attachments and **reported a clean pass** over the other 40 — every `credentials.py` endpoint among them. That is the guard committing the exact error it exists to catch.

**Why it matters even though nothing is mispaired today**: AC4 asks that "the next `OwnedAgent`↔`OwnedAgentByName` swap can't ship silently". In the `Depends()` form it would have — silently, *with a green guard*, which is worse than no guard.

Mutation-tested on the newly covered form specifically, because counting attachments only proves they're counted:

```
credentials.py:100  get_owned_agent_by_name -> get_owned_agent
→ credentials.py:99 get_agent_env_drift() — route
  '/agents/{agent_name}/credentials/env-drift' uses Depends(get_owned_agent)
  (reads path param 'name') but declares ['{agent_name}']; every request 422s
```

`test_the_scan_actually_found_the_routes` now requires **both families** to appear. The old assertion (`> 100 pairings`) passed the entire time the second family was invisible — a total is not a completeness check.

---

### [I1] Fixture keyed an identity-keyed registry on a re-import (Confidence: 10/10) — FIXED in `f266e2ba`

Caught by the full unit tier, not by me: 3 of my own tests passed standalone and failed in the suite with `401 {"detail":"Not authenticated"}`. `dependency_overrides` is identity-keyed; `dependencies` is on conftest's #762 restore list, so a fresh import yields different function objects than `routers.schedules` bound, the override silently doesn't apply, and real auth runs.

Overrides now key on the callables in each route's dependant graph. Staged the divergence rather than hunting the predecessor:

```
OLD (import identity)  -> HTTP 401  {"detail":"Not authenticated"}
NEW (route identity)   -> HTTP 404  {"detail":"Schedule not found"}
```

Generalised into `learnings.md`: **any object-keyed registry over a module-level function is exposed to module-identity divergence** — here that's `dependency_overrides` and `monkeypatch.setattr`.

---

### [I2] AC3 is UNVERIFIABLE, not done (Confidence: 9/10)

The issue names 7 integration tests that must pass. I could not run them: they need `TEST_AGENT_NAME` and **skip** rather than fail, and the isolated backend I stood up has no Docker socket so it sees no agents. Flagging rather than letting "4 skipped" read as a pass. The live evidence below covers the same ground more directly, but it is not the same artifact.

---

### Verified, not assumed

Two backends on the real Postgres/Redis — current `dev` and `dev` + this PR:

```
dev  (f1809975)   enable/disable/trigger   422  [{"loc":["path","agent_name"],"type":"missing"}]
dev + PR#2095     enable/disable/trigger   404  'Agent not found'
```

And AC2 end-to-end against a real schedule (stopped agent, cron that never fires, `trigger` deliberately **not** invoked — it runs a billable execution):

```
enable   HTTP 200  {"status":"enabled",...}
disable  HTTP 200  {"status":"disabled",...}
cleanup  HTTP 204
```

Access was not widened, which is #2081's whole point: no credentials → 401, garbage token → 401, `get_owned_agent` resolves and `get_authorized_agent` does not.

---

### Clean categories

- **4.14 incomplete fix** — #2081 also touched `event_subscriptions.py`; it uses the `{name}` variants correctly. The global guard passing over all 186 attachments is the evidence that `schedules.py` was the only miss.
- **4.3 auth** — every route in `schedules.py` carries a gate (AST-checked); the tier is unchanged (owner→owner), only the variant.
- **4.7 dead code** — `get_authorized_agent` and `CurrentUser` are import-only in `schedules.py`, but at identical counts on `dev`: pre-existing, left alone.
- **4.12 enum completeness** — only two attachment forms exist (`Annotated` alias, `Depends` default); no `dependencies=[...]` decorator form, no inline `Annotated[..., Depends(...)]`, no non-literal route paths.
- **4.1 SQL / 4.2 concurrency / 4.4 credentials / 4.10 frontend / 4.11 performance** — not touched by a 4-line dependency-annotation change.
- **4.13 docs** — no endpoint, schema or flow change; the route contract is unchanged (it is restored).

---

### Summary

| | |
|---|---|
| Critical | 1 — found and fixed in-branch (`244ec096`) |
| Informational | 2 — one fixed (`f266e2ba`), one flagged (AC3) |
| Scope | clean |

Both durable classes recorded in `docs/memory/learnings.md`.

**Honest note on process**: I announced this PR as verified *before* the full tier finished, and the tier then failed 3 of my own tests. The product fix was never in question — the harness was. Not repeating that: the confirming tier is running and I'll report it either way.

vybe (approved):
Validated via `/validate-pr`.

**The fix is right and minimal** — three routes declaring `/{name}/...` were annotated with `OwnedAgentByName`, which reads `agent_name`, so FastAPI could never satisfy the dependency and 422'd before the handler ran. Swapping to `OwnedAgent` keeps #2081's CSO M1 owner-tier intent intact while binding to the param that actually exists; the inline comments say exactly that, so the next person tightening these routes won't re-introduce it.

**The guard is the more valuable half, and it was nearly a fake.** FastAPI accepts these gates two ways — the `Annotated` alias and `= Depends(...)` — and the first cut read only `ast.Name`, scanning 146 of 186 attachments while reporting success. Catching that in self-review, fixing the collector to both families, requiring completeness *per family* rather than in total, and mutation-testing a `Depends()`-form swap is the difference between a guard and a green light over a blind spot. The `dependency_overrides` module-identity trap found alongside it is the kind of thing that passes standalone and 401s the whole suite.

The `learnings.md` entry generalises both correctly — particularly that a completeness assertion stated as a total will pass for as long as an entire syntax family is invisible.

Conflict on `tests/registry.json` (positional only — both sides appended to the same array tail; #2098 landed first) resolved in `0a05e1d5` by rebuilding from the merge stages: +24/-0 against dev, no entry lost from either side, no formatting drift.

Security scan clean. Scope clean.

PR #2096, opened 2026-08-11
docs: v0.9.0 code-freeze plan

trinity-ability:
Output of /release-plan. Work order for the v0.9.0 freeze (2026-08-12): 3 MUST-LAND items (#2095, enterprise pin bump to bce1175, TRINITY_DEFAULT_SKILL_SOURCE compose passthrough), 7 SHOULD-LAND ready PRs, evaluations FENCE checklist, and the Known-limitations block for /release Step 5.

### DIFF
--- a/docs/releases/v0.9.0-freeze-plan.md
+++ b/docs/releases/v0.9.0-freeze-plan.md
@@ -1,0 +1,127 @@
+# Release Coherence Review — v0.9.0 — VERDICTS CONFIRMED 2026-08-11
+
+Confirmed by eugene 2026-08-11: C4 registry default-ON intended (ship + air-gapped note) ·
+C5 evaluations **FENCE** (foundation; F1 checklist active) · C10 rooms reframed — the surface
+is being absorbed into Workspace (ent#78: #358 absorbs Session tab, #361/#362 build Workspace
+chats on the room engine), so the engine bugs ent#218/ent#220 stay SHOULD (substrate survives
+the move), #1819 rooms-UI polish is likely mooted (confirm with dolho before spending on it),
+and **Workspace delivery becomes the controlled track** (land #2083 now; ent#78 absorb-items
+are the next-cycle plan) · Freeze = **2026-08-12** (tomorrow).
+
+Generated 2026-08-11 by /release-plan v1.0 (full mode). Payload: v0.8.5..origin/dev.
+
+Payload: 127 issues (89 public / 38 private) · 371 commits · 1027 files (+144,822 / −32,860) · since v0.8.5 (2026-07-26, 16d)
+Public types: 76 bugs · 8 features · 2 docs · 2 refactors · 1 epic
+Themes (union): reliability 47 · security 19 · devex 22 · ui-ux 20 · infrastructure 12 · channels 3
+
+Epic deep-dives were done via batched sub-issue GraphQL (reference.md fallback), not per-epic /roadmap — noted degradation (14 epics touched; identical data source).
+
+## Payload census
+- Clean (in-dev + in commits): 126 of 127 (all 89 public confirmed in commit subjects/bodies; 37/38 enterprise confirmed in subjects or submodule log)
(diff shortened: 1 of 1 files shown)


dolho:
## ent#286 (portal streaming) — recommend **cut from v0.9.0**, design recorded

Investigated it today. It is not the small item the plan implies, and I do not think it should land before the freeze. Full write-up on abilityai/trinity-enterprise#286; the short version:

**`portal_chat` awaits `execute_task` and returns `{response, cost, session_id}`.** The turn is over when the response arrives and **no `execution_id` reaches the client** — so an SSE endpoint added alone is unreachable dead code. Real streaming means splitting the chat into dispatch → stream → finalize, which changes a **shipped** contract in three places: the response shape, the history-persist timing (currently after the await), and the 429/504/502 mapping (currently derived from the completed result). It also needs `"public"` added to `ASYNC_DISPATCH_ELIGIBLE_TRIGGERS`, which is a decision about the #1083 fire-and-forget path rather than a config tweak.

The issue's own body rates it **P3**: *"a latency/UX improvement, not a missing capability."* It was raised to P1 for this plan. Meanwhile **ent#362** and **ent#218 / ent#220** have no code at all and genuinely block the Workspace story.

Trading a shipped-contract change for a latency win, 8 days out, while two P1s are unstarted, looks like the wrong risk. Design is settled on the issue so it is pick-up-and-build in v0.9.1.

---

## Other corrections to the plan, verified today

**Already done — the plan overstates the remaining work:**

| Item | Plan says | Verified |
|---|---|---|
| PR #2024 | "just merge it" | **MERGED** |
| #1976 | "fix red CI, then merge" | **MERGED** |
| ent#292 (P0, PAT leak) | blocks the cut | **`status-in-dev`** — fix merged (`bbc6b304`) |
| ent#297 / #293 / #334 | need a ship/block decision | all three **`status-in-dev`**, fixes merged |

So the agent-key-admin cluster decision is largely retrospective — worth recording, not blocking.

**The Alembic check passes — bump is safe.** `d1c5ebb → bce1175` is one commit (the ent#356 removal) and **zero** migration files:

```
d1c5ebb: 16 revisions, 1 head (0015_merge_a2a_control)
bce1175: 16 revisions, 1 head (0015_merge_a2a_control)
OSS:     37 revisions, 1 head (0036_client_portal_oss)
```

The #2068 two-heads failure mode does not apply to this bump. (Note for anyone re-checking: a naive head count reports 3 heads because `0015_merge_a2a_control` is a **merge** revision whose `down_revision` is a tuple — parse it as one or you will get a false alarm.)

**The compose passthrough gap is real and confirmed:**

```
config.py                reads TRINITY_DEFAULT_SKILL_SOURCE(_REF)   yes
.env.example             documents it                              yes
docker-compose.yml       references                                0
docker-compose.prod.yml  references                                0
```

Inert on every deployed instance. This is the **seventh** recurrence of that class in `learnings.md` (VOIP_*, LOG_*, #1039, AGENT_LOG_*, CANARY_*, TEMPLATE_REGISTRY_*), and prod compose is the load-bearing file.

**A `DIRTY` PR runs no CI at all.** #2099 showed *zero* checks — not green, not red, unrun. Any "CI green" line in the plan needs re-checking after a rebase. #2095, #2098, #2099 and #1976 were all conflicted and are now rebased and running; **#2076 is still `DIRTY` and failing `lint (sys.modules pollution check)`**.

vybe (approved):
Docs-only release artifact under `docs/releases/` (explicitly out of scope for enterprise-docs-guard). Security scan clean: no secrets, emails, or public IPs. All 4 required checks green.

Merging as the point-in-time record of the `/release-plan` run. Noting @dolho's corrections above are **not** folded into the doc — ent#286 cut recommendation, and the four already-done items (#2024, #1976, ent#292, ent#297/#293/#334 all `status-in-dev`). The comment thread carries them; if the plan is going to be worked from tomorrow rather than just archived, it wants a follow-up commit.

PR #2098, opened 2026-08-11
fix(skills): close the settings back door onto skills-source grants (ent#346)

dolho:
Fixes abilityai/trinity-enterprise#346 · P0 · security

## What

ent#237 gates every `/skills/sources` route with `reject_agent_principal`, and says why at the site: adding a source is the **GRANT** action, and a prompt-injected agent that could register its own repo gets unattended, fleet-wide, persistent prompt injection — skills are instructions Claude follows and they ship executable `scripts/`.

That gate works. The same grant was reachable through `PUT /api/settings/skills_library_url`, which is `assert_admin`-gated but **not** `reject_agent_principal`-gated — and an agent-scoped key resolves to its owner carrying the owner's role, so on the default admin-owned install it passes. `_adopt_legacy_clone` then turned that string into a `skill_sources` row **with no validation at all**, at CUSTOM priority (which outranks the bundled community catalog), and deleted the setting afterwards, erasing where the row came from.

## Four links, four fixes

Closing any one alone leaves a hole.

**1. The keys are blocked on the generic PUT**, 422 pointing at `POST /api/skills/sources`. Both keys — a source is `(url, ref)`, so re-pointing `skills_library_branch` alone changes which commit the fleet executes.

Validating the URL there, which is all it did before, was never enough: `github.com/attacker/skills` passes `validate_skills_library_url` cleanly. The question is **who may grant a source**, not what the string looks like.

**2. The sink validates** — `validate_skills_library_url` **and** `reject_embedded_credentials`, the pair the write routes use, before `create_skill_source`. Boundary *and* sink (#1525), so a future writer inherits the check instead of bypassing it.

**3. The dead guard is implemented.** It read:

```python
if db.get_default_skill_source() is None and db.count_skill_sources() == 0:
    pass  # nothing configured yet — proceed
```

Two DB calls, result discarded. Adoption ran unconditionally, every sync, every install, forever — which is exactly what let a URL written *after* ent#237 still become a source on a configured fleet.

**4. Refusals are observable.** Every failure used to be one `logger.warning`, so a rejected URL and a successful adoption were indistinguishable from outside. Now a `high` operator-queue alert, fail-soft.

## Verification

14 regression tests, **mutation-tested per link** so each is known to be load-bearing:

| mutation | result |
|---|---|
| settings block removed | 3 failed |
| sink validation removed | 6 failed |
| dead guard restored | 2 failed |
| alarm downgraded to a log line | 2 failed |
| as shipped | **14 passed** |

Two tests guard the *fix* rather than the bug: a genuine unmigrated install still migrates (ent#237 AC#6 — a gate that blocks the legitimate path is a regression, not a fix), and the ordering check is an **AST** assertion that the guard precedes `db.set_setting`, because a block placed after it would satisfy any text search while the value had already landed.

Existing skills suites unaffected: **174 passed**. `test_ent236_skills_lifecycle.py` errors at collection standalone on `dev` too — pre-existing, not from this change.

## Scope

- ent#334 (the disclosure half) is already on `dev` (`6f9e1f90`); this is built on top, as that issue's sequencing note requires.
- **AC6 deliberately not solved here** — whether `assert_admin` should reject agent principals generally is ent#297's territory, cross-linked rather than absorbed.
- The write itself was never executed against a live instance (it mutates a fleet-wide setting); the block is proven by test, and the issue's own read-half verification established the principal passes `assert_admin`.

---

## Correction to the premise (found while fixing CI)

The issue says `PUT /api/settings/{key}` is "`assert_admin`-gated but **not** `reject_agent_principal`-gated". **It is** — ent#293 added that gate, and its own test proves an agent-scoped key gets a 403 on `dev` today.

So the agent-key half of the reported chain was already closed. Keeping the claim would have justified this fix with a stale premise, so: what this PR actually fixes is the rest of it, which is the larger part.

* `_adopt_legacy_clone` validated **nothing**, so any URL that reached the setting became a `skill_sources` row.
* The adoption guard was `if <two db calls>: pass` — adoption ran on every sync of every install, not only an unmigrated one.
* Every failure was one `logger.warning`, so a refusal and a success were indistinguishable from outside.
* And a **human admin** could still create an unvalidated, precedence-winning source through a route carrying no grant gate — which is why the block is not principal-scoped.

## Test contract reconciled

The regression diff caught a real collision, not a flake: `test_293…::test_the_same_write_still_works_for_a_human_admin` asserted a human admin can still write `skills_library_url`. This PR refuses that key for everyone, so both could not hold.

ent#346's AC is explicit (block it, 422 with a pointer, "ent#237 already removed the UI writer"), so the block stands. But ent#293's invariant still matters and deleting the assertion would have silently dropped it: **the gate must discriminate by principal, not blanket-block the route** — a 403-for-agents test only says something about the *principal* if some human write actually succeeds. That half now runs against `public_chat_url`, and a new test pins ent#346's contract (human admin → 422 naming the supported route).


dolho:
## `/review` — self-review + live verification

**Branch** `fix/ent346-skills-source-injection` → `dev` · merge-base `69bc5788`
**Files** 4 (+449/−10) · **Scope** CLEAN — nothing outside the two files the issue names, plus its tests
**Plan completion** 5 DONE · 1 deliberately deferred (AC6 → ent#297, as the issue asks)

---

## The chain, executed rather than inferred

The issue says: *"The write itself was deliberately not executed — it would mutate a live setting… Worth a five-minute confirmation on a throwaway instance before remediation is signed off."*

Done. Two throwaway backends — one on `dev` (`69bc5788`), one on this branch — each on its **own SQLite database and data dir**, so nothing touched the shared Postgres or the running fleet.

**Layer 1 — the settings write:**

```
dev   PUT /api/settings/skills_library_url      HTTP 200   value stored
dev   PUT /api/settings/skills_library_branch   HTTP 200   value stored
fix   PUT /api/settings/skills_library_url      HTTP 422   "…must be set via POST /api/skills/sources"
fix   PUT /api/settings/skills_library_branch   HTTP 422   "…must be set via POST /api/skills/sources"
fix   GET /api/settings/skills_library_url      HTTP 404   nothing stored
```

**Layer 2 — does that value actually become a source?** Triggered a sync on `dev`:

```
before:  1 source   Trinity Community Skills | github.com/abilityai/trinity-skills | priority 1000
after:   2 sources  Migrated library         | github.com/attacker/skills          | priority  100
                    Trinity Community Skills | github.com/abilityai/trinity-skills | priority 1000
```

Resolution is `priority ASC`, so **the injected source outranks the bundled catalog and wins every name clash** — exactly as the issue predicted. Note it adopted on an install that *already had a source*, which is the dead guard: adoption was never gated to an unmigrated install.

**Layer 3 — the sink, independently.** The boundary now blocks the API path, so I planted the setting **directly into the fix instance's SQLite** to bypass it entirely:

```
planted: skills_library_url = https://github.com/attacker/skills
sources before sync:  1   [github.com/abilityai/trinity-skills]
sources after  sync:  1   [github.com/abilityai/trinity-skills]     ← refused
operator queue:       1   "Legacy skills-library adoption refused" | skills_legacy_adoption_refused
```

Defense in depth holds: even with the boundary bypassed, no row is created, and the refusal is visible to an operator rather than buried in a log.

---

## [I1] Finding on my own diff — fixed in `c6113848`

`skill_service.py` already imports `validate_skills_library_url` at module scope, and my sink validation added a second, **function-local** `from utils.url_validation import …`.

Not style. A function-local from-import re-resolves `sys.modules[...]` on every call while the module-level bind happens once, so the two can land on different objects when a conftest restore swaps the module — the identity divergence that has bitten three times this week. Both validators now use the one module-level import.

---

## Clean categories

- **4.14 incomplete fix** — grepped every writer of `skills_library_url`: the generic PUT was the only one, and `_adopt_legacy_clone` the only reader that creates a row. No sibling path left open.
- **4.3 auth** — the fix removes a write capability; it grants nothing. `assert_admin` is unchanged (widening it is ent#297).
- **4.9 test gaps** — 14 tests, mutation-tested **per link** (3 / 6 / 2 / 2 failures when each fix is reverted individually), so none is decorative. Two guard the *fix* rather than the bug: a genuine unmigrated install must still migrate (ent#237 AC#6), and a broken alarm must not take the sync down.
- **4.1 SQL / 4.2 concurrency / 4.10 frontend / 4.11 perf** — untouched by this change.
- **Existing suites** — 174 passed across `test_ent237_skill_sources`, `test_ssrf_skills_library`, `test_ent332_skills_root`. `test_ent236_skills_lifecycle.py` errors at collection standalone **on `dev` too** — pre-existing, not from this change.

---

## Summary

| | |
|---|---|
| Critical | 0 |
| Informational | 1 — found and fixed in-branch |
| Scope | clean |
| Live verification | vulnerability reproduced on `dev`, refused on the fix, at both layers |

One judgement worth surfacing for the reviewer: **the boundary block is the behaviour change most likely to surprise someone.** `PUT /api/settings/skills_library_url` now 422s for admins too, not just agent principals. That is deliberate — the key *is* a source grant, and ent#237 already removed the UI writer — but if any operator runbook still sets it that way, it will break loudly with a pointer rather than silently.

vybe (approved):
Validated via `/validate-pr`.

**Chain closed at all four links, and the links are the right ones.** The settings block, the sink validation, the genuinely-gated adoption, and the audible refusal each stand alone — closing any one in isolation would leave the grant reachable. The AST assertion that the 422 *precedes* `db.set_setting` is the detail that makes the first link real rather than grep-satisfying.

The `if <two db calls>: pass` finding is the substance here: adoption ran unconditionally on every sync of every install, so a URL written after ent#237 still became a CUSTOM-priority source on a fully-configured fleet. That is the exploit, and it was invisible in review precisely because a discarded guard reads as protection.

Two things I checked specifically rather than took on trust:
- `test_ent183_skill_packages.py` mirrors `reject_embedded_credentials` faithfully instead of stubbing it to `None`. A permissive stub for the one predicate the feature exists to enforce would have been a silent hole in the harness.
- ent#293's "human admin can still write" test was correctly re-pointed at `public_chat_url` rather than deleted — the principal invariant still needs a key that is genuinely writable to be evidence about the principal.

Security scan clean (the `ghp_deadbeef@github.com` / `x@github.com` hits are the credential-rejection fixtures; gitleaks green). Scope clean, all 21 checks green.

Note for the merge: cross-tracker `Fixes abilityai/trinity-enterprise#346` will not auto-promote — ent#346 needs a manual `status-in-dev`.

PR #2099, opened 2026-08-11
feat(workspace): the session slides instead of expiring at 12 hours (ent#375)

dolho:
Related to abilityai/trinity-enterprise#375

> *"It asks to re-login every time — it's very inconvenient."*

A Workspace session hard-expired **12 hours** after sign-in with no renewal path anywhere in the backend, so anyone using the surface on two consecutive days redid the email OTP. It now **renews on use**, dies on inactivity (idle window, default 7d), and is bounded by an **absolute cap** no amount of use can extend (default 30d).

## Two decisions, confirmed before building

**Scope — portal session only.** The issue poses this as an open question. The 7-day platform session is deliberately untouched: it authenticates every operator surface, the WS tickets and the MCP-key path, so a rotation bug there is fleet-wide rather than portal-only. If it should slide too, that is a second issue on this mechanism.

**Gating — OSS mechanism, entitled knob.** The mechanism ships in OSS with safe defaults; the Settings panel that retunes the windows is a **separate enterprise change**. `users.suspended_at` (#995) is the precedent: OSS owns the primitive and its enforcement, the enterprise module owns the setter. A community install slides on the defaults, it just cannot retune them.

## Two clocks, and why

`iat` **must** move on every rotation — the ent#281 revoke cutoff dates tokens against it, and a rotated token still carrying its original `iat` would look older than a revoke that should have killed it.

So `iat` cannot also measure the cap: if it did, every rotation would reset it and a continuously-used session would live **forever** — precisely what an absolute cap exists to prevent. A second claim, `sst` (session start), is carried through rotations unchanged and is what the cap measures from.

## Renewal fails closed. The read path still fails open.

`is_token_revoked` and `portal_sessions_revoked_at` fail open — right for a read (a Redis blip must not sign every client out), indefensible for a mint. Failing open on renewal hands a **fresh** token to a session an operator just killed, and because the new token post-dates the cutoff, the kill does not re-apply when Redis recovers: one blip permanently resurrects a revoked session.

Both postures are pinned, so the stricter one is not applied to the read path by accident.

## The bug this would have introduced — not in the ACs

`revoke_portal_sessions_for_email` derived its Redis cutoff TTL from `PORTAL_SESSION_EXPIRE_HOURS`. Extending sessions to 30 days while that still said 12 hours leaves the cutoff evaporating **29.5 days before** the tokens it exists to kill — the next request finds no cutoff and sails through. Any session revoked more than 12 hours ago silently comes back.

The TTL now reads the policy so it tracks the cap, and the constant is **deleted** rather than left at 12 — a stale constant beside a live policy is exactly how this happened.

## Also

- Shortening the cap takes effect **immediately** for sessions already signed in — enforced on read, not only at mint. An operator narrows a cap precisely when the currently-open sessions are the problem.
- The stored policy is clamped **on read**, so a direct DB write or a default regression cannot produce `idle > absolute` (#506 clamp-on-use).
- A settings-read failure degrades to the shipped defaults rather than raising — this is the auth path.
- Rotation is threshold-based, not per-request: it revokes the token it replaces, so rotating constantly races a browser's concurrent requests. The superseded `jti` gets a short grace, not an instant kill.
- `expires_in` at sign-in reports the real lifetime instead of a constant.

## Client side

The rotated token arrives in `X-Trinity-Session-Token` and is adopted by **one axios interceptor** rather than 13 per-call-site edits — a portal call added later cannot forget to opt in, and forgetting would be silent. Only a client session rotates; an operator previewing on a platform JWT must not acquire a portal token from a header.

An ended session is now **honest**: `sessionExpired` distinguishes "your session ran out" from "you signed out" (same form either way, so the reason must be carried), and `resumePath` returns the user to the thread they were on rather than the roster root.

## Verification

18 backend + 9 frontend tests, **mutation-tested** — a security trade is only as good as what it refuses:

| mutation | result |
|---|---|
| revoke TTL back to 12h | caught (the resurrection bug) |
| renewal made fail-open | caught |
| cap measured from `iat` not `sst` | caught (immortal session) |
| interceptor never installed | caught (rotation silently stops) |
| adopt without an existing token | caught (operator acquires a client session) |

Sibling suites unaffected: **57 passed** (ent#281 revoke/block, ent#163 delegated identity). Frontend suite: **52 passed**.

## Not in this PR

- **AC7's Settings panel** — the entitled surface, in the enterprise repo, next.
- Delegated-exchange tokens (ent#163) mint through the same primitive and therefore slide identically; that is intentional and covered by the ent#163 suite still passing.


vybe (approved):
Approving. Security-sensitive (session lifetime) so I went through the auth path line by line — it holds up, and the two-clock design is the crux.

**`iat` vs `sst` is necessary, not decoration.** `iat` must move on rotation because the ent#281 revoke cutoff dates against it; `sst` must not, because the absolute cap measures from it. Carrying only `iat` would reset the cap on every rotation and the session would live forever — the exact thing an absolute cap exists to prevent.

**`renew_portal_session` fails CLOSED on the revocation store, and that asymmetry is right.** `is_token_revoked`/`portal_sessions_revoked_at` fail open (correct for a read — a Redis blip must not sign everyone out); a *mint* held to the same standard would hand a fresh token to a session an operator just killed, and because the new token post-dates the cutoff the kill would not re-apply on recovery. One blip permanently resurrects a revoked session. Sharpest edge in the feature and it's handled.

**Deleting `PORTAL_SESSION_EXPIRE_HOURS` rather than leaving it at 12 is the right call.** `revoke_portal_sessions_for_email` derived its Redis TTL from it — extending sessions to 30 days with that constant still reading 12h would expire the revoke cutoff **29.5 days before** the tokens it exists to kill. A `NameError` is the correct way to find any remaining reader.

**Gating** ✅ — read is `assert_admin`, which since #1890 rejects agent principals at the gate itself (verified in `dependencies.py`), so an agent's injected key can't reach it on a default admin-owned install. Read staying OSS-core while only the setter is entitled is right: a community operator must be able to inspect a policy that is enforcing on them.

**Clamp-on-read** ✅ (#506 pattern) — floors/ceilings then the `abs >= idle` ordering invariant, so a direct DB write or a future default regression can't mint a session outliving its own cap, and narrowing the cap takes effect immediately rather than only for new sessions. No env leg, deliberately (#1638).

**Packaging** ➖ — `config.py`/`dependencies.py` are modified, not new, and there are no new `os.getenv` vars, so neither gate applies. **Security scan** ✅ clean.

**Conflict resolved on merge** — `stores/clientPortal.js` `fetchRoster` catch block, where #2083 (ent#357) landed first. Both sides rewrote it for unrelated reasons and I kept both: this branch's `endSession({expired, resumePath})` on 401, and #2083's `unavailable` discrimination on 404. `endSession()` wraps `signOut()` so teardown is unchanged. Verified with both suites — `workspaceSession.spec.js` + `workspaceSessionRenewal.spec.js`, 28 passed.

Post-merge: **manually set `status-in-dev` on abilityai/trinity-enterprise#375** (cross-tracker), then land 🔒 #376, which reads these constants and has a parity test pinned to them.

PR #2100, opened 2026-08-11
fix(compose): pass the community-catalog vars to the backend (ent#237)

dolho:
Related to abilityai/trinity-enterprise#237 · on the v0.9.0 **Aug 12** list (#2096)

## The gap

`config.py` reads `TRINITY_DEFAULT_SKILL_SOURCE(_REF)` and `.env.example` documents them — but **neither compose file passed them through**. Prod compose launches standalone (no base merge, no `env_file:` on the backend service), so the explicit `environment:` list is the only route in. The operator's override was **inert on every deployed instance**; only the code default ever applied.

It works on a laptop, where the shell environment reaches the process directly. That is exactly why it survived review.

Seventh recurrence of this class in `learnings.md` (`VOIP_*`, `LOG_*`, #1039, `AGENT_LOG_*`, `CANARY_*`, `TEMPLATE_REGISTRY_*`).

## The plan's suggested fix would have been worse than the bug

#2096 says *"same pattern as `TRINITY_DEFAULT_SYSTEM_MANIFEST`"* — i.e. `- VAR=${VAR:-}`. That is wrong **for this variable**.

Empty is the documented **disable** for the skills catalog (*"Set the URL to `""` to disable the seed entirely"*), whereas the manifest treats empty as "use the bundled default" (its sentinels are `disabled/none/off/0/false`). Two variables that look like siblings need different forms.

Rendered with `docker compose config` against the real prod file, across all three operator states:

| form | unset | set to a URL | set to `""` |
|---|---|---|---|
| **`- VAR`** (bare) | `null` → code default ✅ | the URL ✅ | `""` → disabled ✅ |
| `- VAR=${VAR:-}` | `""` → **catalog off everywhere** ❌ | the URL | `""` |
| `- VAR=${VAR:-default}` | default | the URL | default → **disable ignored** ❌ |

Only the bare pass-through satisfies all three. It also keeps the default in **one** place (`config.py`) rather than copied into two compose files.

## Verified by rendering, not grepping

```
unset     -> TRINITY_DEFAULT_SKILL_SOURCE: null
set       -> TRINITY_DEFAULT_SKILL_SOURCE: github.com/acme/skills
disabled  -> TRINITY_DEFAULT_SKILL_SOURCE: ""
```

## Guard

12 tests asserting the **form**, not the presence — a presence-only check passes for `${VAR:-}`, which is the dangerous variant. Mutation-tested against each wrong form:

| mutation | result |
|---|---|
| removed from prod compose (the shipped bug) | **4 failed** |
| `${VAR:-}` — the form the plan suggested | **1 failed** |
| hardcoded default | **2 failed** |

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -170,6 +170,17 @@ services:
       # operator manifest, or `disabled` to skip seeding. Empty = bundled
       # config/manifests/default-system.yaml.
       - TRINITY_DEFAULT_SYSTEM_MANIFEST=${TRINITY_DEFAULT_SYSTEM_MANIFEST:-}
+      # ent#237 — the community skills catalog seed. BARE pass-through, NOT
+      # `${VAR:-}`: an empty value is the DOCUMENTED disable for this setting
+      # ("Set the URL to \"\" to disable the seed entirely"), so `${VAR:-}` would
+      # send "" on every install and silently turn the catalog off for everyone.
+      # `${VAR:-github.com/...}` is wrong the other way — it swallows the
+      # operator's explicit disable. Rendered with `docker compose config`:
+      #   bare, unset          -> null (not set) -> config.py default applies
+      #   bare, set to a URL   -> the URL
+      #   bare, set to ""      -> "" -> seed disabled, as documented
+      - TRINITY_DEFAULT_SKILL_SOURCE
+      - TRINITY_DEFAULT_SKILL_SOURCE_REF
       # ent#126: directory the bundled-manifest catalog
       # (GET /api/systems/manifests) reads. Empty = the image's own
       # config/manifests (already bind-mounted :ro below). Point it at another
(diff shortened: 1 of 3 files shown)


vybe (approved):
MUST-LAND for v0.9.0 (C3 in the freeze plan). Validated:

- **Config packaging** ✅ — `TRINITY_DEFAULT_SKILL_SOURCE(_REF)` now wired into **both** compose files' `backend.environment:` and already documented in `.env.example:277-278`. Confirmed `config.py:446,448` reads both names, so the wiring and the reader agree.
- **The form is the fix, and it's right.** Bare `- VAR` is the only correct shape here: `${VAR:-}` sets `""` on every install, and empty is the *documented disable* (`config.py:443`), so the obvious copy-the-neighbour fix would have silently killed the community catalog fleet-wide. `${VAR:-<default>}` breaks it the other way. The test asserts the form, not the presence — a presence-only check passes for the variant that causes the outage.
- **Test adequacy** ✅ — non-happy-path covered in both directions: form assertion, no-duplicated-default, `.env.example` documentation, and a reverse guard that a rename in `config.py` can't leave compose passing dead names.
- **Security** ✅ clean. All 4 required checks green.

Two non-blocking notes: no `tests/registry.json` entry for the new file (the registry is an unenforced catalog — 380/517 unit files are already absent, so this is cosmetic), and the `Related to` reference means ent#237 won't auto-promote — it's already `status-in-dev`, so nothing to do.

Issue #2101, opened 2026-08-11
bug(workspace): new-chat briefing renders capability hints as an unbounded, generically-named card grid


vybe:
## Summary

In a Workspace chat (`/workspace`), opening an agent shows the new-chat briefing — "Things you can ask" — as a card grid of the agent's capability hints. With no connector allow-list configured (the default), **every** `user_invocable` skill becomes a card, titled by its humanized slug (`weekly-report` → "Weekly report"). An agent with many skills renders as a blanket sheet of near-identical, generically-named cards, and the briefing grows without bound.

> **Re-scoped 2026-08-11:** originally filed as "tool activity renders as a flat, ungrouped list in the transcript" — that symptom doesn't exist on `dev` (no tool activity reaches the workspace transcript at all; the portal streaming path is trinity-enterprise#286, still open). The observed behavior was the briefing hint grid. Per-turn tool-activity grouping becomes relevant when ent#286 lands and should ride that work.

## Context

Workspace is the single continuous-conversation surface (trinity-enterprise#78) and this is on the v0.9.0 release-required list (freeze plan Wave B). The design system has a binding rule for exactly this class: bounded viewports for unbounded data (`docs/memory/design-system-contract.md`, principle 28). Adjacent in-flight work: PR #2103 (trinity-enterprise#380) adds a template `use_cases` fallback tier (capped at 6) but leaves the exposed-playbook tier uncapped — this bug remains after it merges.

## Acceptance Criteria

- [ ] The briefing shows at most 6 hint cards by default; more collapse behind a counted "Show all N" toggle
- [ ] Expansion is **in-place** — no nested scroll region (the chat pane stays the single scroll axis); the set itself is bounded by a server-side belt (≤24 hints shipped per agent), so the briefing cannot grow unbounded with skill count
- [ ] The toggle's count is accurate for the shipped list; expanding/collapsing never resets the chat pane's scroll position
- [ ] The toggle is one persistent `type="button"` element carrying `aria-expanded`; focus is preserved across expand/collapse
- [ ] Hint cards are deterministically ordered (description-bearing cards first, listing order preserved within groups)
- [ ] Light and dark themes both first-class; semantic tokens only (raw-color ratchet applies)
- [ ] Agents with ≤6 hints render exactly as today; zero hints render no chrome
- [ ] Composer pre-fill behavior unchanged (clicking a card never auto-runs)

## Technical Notes

[...]

PR #2103, opened 2026-08-11
feat(workspace): capability hints in chat — use-case fallback + template/info fix (trinity-enterprise#380)

trinity-ability:
## Summary

Workspace chat gave no hint of what an agent can do unless the operator had exposed connector playbooks — and even then, the briefing's metadata read was silently broken. This completes the ent#138 empty-chat briefing into the ent#380 capability-hint surface:

- **Hint ladder**: the empty/new-chat state shows the operator's exposed playbooks (connector allow-list ∩ `user_invocable` — the curated set, same policy the MCP connector advertises); an agent exposing **none** now falls back to its template-declared `use_cases` ("What You Can Ask", the same metadata InfoPanel shows operators). Sanitized + capped (6 hints × 200 chars).
- **Bug fix riding along**: `_agent_briefing` called `{agent}/info`, a route that does not exist on the agent server — the best-effort catch swallowed the 404, so briefing descriptions have been silently `None` since ent#138. It now reads the canonical `/api/template/info` (the route InfoPanel, A2A cards, and avatars already use), which also carries `use_cases` — so the fallback costs zero extra HTTP calls.
- Clicking a hint **pre-fills the composer, never auto-sends** (existing ent#138 `use-playbook` → `prefill` chain, unchanged). Card titles wrap to two lines so full-sentence prompts don't clip.

## Acceptance criteria (ent#380)

- ✅ New/empty chat shows capability hints — playbooks or use-case fallback, shipped on the roster at sign-in (zero extra fetches)
- ✅ Clicking a hint prefills, never auto-sends (pre-existing chain, now pinned by tests at the data layer)
- ✅ Curated-set discipline: exposed playbooks (never the raw skill list) → `use_cases` fallback; ent#178's unified config slots into the same `_agent_briefing` seam later (docstring marks it)
- ✅ Graceful empty state: an agent declaring nothing keeps the existing avatar/description briefing — no dead chrome, composer remains the next action
- ✅ Multi-agent: N/A by construction — Workspace chats are strictly single-agent (the picker starts a new chat), so hints always scope to the active agent

## Deliberately not done

- No generic filler prompts (à la `ChatEmptyState.vue`'s `FALLBACK_PROMPTS`): a client surface must not advertise capabilities the operator/template never declared.
- Did not adopt `ChatEmptyState.vue`: its `sendImmediately` heuristic violates "never auto-sends", and its playbook source ignores the connector allow-list.
- Did not touch the stale enterprise-submodule portal copy (`src/backend/enterprise/backend/client_portal/`) — dead code since ent#356; the OSS router wins registration order.

## Changes

- `src/backend/client_portal/service.py` — `/api/template/info` fix, `_use_case_hints()` sanitizer, fallback ladder
- `src/backend/client_portal/models.py` — hint-card docstrings (wire shape unchanged)
- `src/frontend/src/components/portal/PortalBriefing.vue` — title wraps (line-clamp-2); comments
- `src/frontend/src/stores/clientPortal.js` — comment
- `tests/unit/test_ent79_portal_exposure.py` — 5 new tests (sanitization, ladder precedence, policy-filtered fallthrough, metadata-failure isolation, route pin)
- `docs/memory/architecture.md` — Workspace section: briefing hint ladder

## Test Plan

- [x] `pytest tests/unit/test_ent79_portal_exposure.py` — 65 passed (5 new + 60 pre-existing)
- [x] Adjacent portal suites green: `test_ent356_portal_is_oss_core`, `test_ent357_workspace_owned_roster`, `test_ent287_portal_rate_limits` (49 passed)
- [x] Frontend `vitest run` — 71 passed; design-token check + raw-color ratchet clean
- [ ] Manual: workspace empty chat with (a) exposed-playbook agent, (b) playbook-less agent with `use_cases`, (c) bare agent

Fixes abilityai/trinity-enterprise#380


dolho:
## Review — `/review` (branch → dev, merge-base `06e10ea2`)

**Scope: clean.** 6 files, +257/−20, all of it ent#380. Test-plan claims verified rather than taken on trust: `114 passed` across the four suites named (65 + 49). No `v-html` anywhere on the render path — `PortalBriefing.vue` uses `{{ }}` interpolation only, so the new author-controlled text is escaped by Vue. No new endpoint, no auth-surface change.

**The bug fix is real and correctly diagnosed.** I checked the agent server: `/api/template/info` exists (`agent_server/routers/info.py:160`) and returns both `description` and `use_cases`; there is **no** bare `/info` route anywhere under `docker/base-image/`. So descriptions have indeed been silently `None` since ent#138, and the fallback genuinely costs zero extra calls.

One finding worth a decision before merge, and two smaller notes.

---

### [C1] The `use_cases` fallback overrides an operator's explicit "expose nothing" (confidence 9/10)

`src/backend/client_portal/service.py` — the ladder's trigger is `if not playbooks`, which cannot distinguish *"the operator never configured a connector"* from *"the operator deliberately exposed zero playbooks"*. Traced end-to-end:

- `db/connector.py::_decode_playbooks` returns `None` only for a NULL column; a stored `"[]"` decodes to `[]`. So an explicit empty allow-list **round-trips as `[]`**, distinct from "unconfigured".
- `connector_service.resolve_exposed_playbooks` — `allow = set([])`, `allow is not None`, so every playbook is filtered out and the curated set is legitimately empty by operator choice.
- `service.py` then does `if not playbooks: playbooks = _use_case_hints(use_cases)`.

Net effect: an operator who went into the connector config and deliberately exposed **nothing** to external clients now has the template's "What You Can Ask" advertised to those clients instead. On a client-facing surface that inverts the operator's most restrictive setting into its most permissive outcome.

This is deliberate — `test_briefing_policy_filtered_playbooks_still_fall_back` pins it:

```python
connector_cfg={"enabled": True, "exposed_playbooks": []},  # operator exposed none
```

but it sits awkwardly against the PR's own stated principle: *"a client surface must not advertise capabilities the operator/template never declared."* An explicit `[]` **is** a declaration — of nothing.

Also note that test's docstring describes a different scenario than its fixture: it says *"(user_invocable=False, or filtered off the allow-list)"*, which are the "nothing curated" causes, while the fixture is the "deny-all" case. Those deserve to split, because their intent differs:

| `cfg` | meaning | fallback? |
|---|---|---|
| `None` (no connector config) | never curated | **yes** — sensible default |
| `exposed_playbooks == []` | operator exposed nothing | **no** |
| non-empty list, resolves to `[]` | stale/renamed playbook names | judgement call |

The distinction is cheap — `cfg` is already in hand a few lines above:

```python
cfg = core_db.get_connector_config(agent_name)
allow = cfg["exposed_playbooks"] if cfg else None
...
if not playbooks and allow is None:       # never curated, not "curated to nothing"
    playbooks = _use_case_hints(use_cases)
```

Not blocking if it is a conscious product call — but it should be conscious, and the test comment should say so instead of reading as an incidental fallthrough.

### [I2] `description` becomes visible for the first time on upgrade (confidence 7/10)

Correct as a fix, but worth a release note. Because `/info` 404'd, **no** workspace client has ever seen an agent description. After this lands they all do, at once, from `template.yaml` text written while the field was (unknowingly) inert. Anything an author put there assuming an operator-only audience — internal naming, client names — surfaces to external clients on deploy. Cheap mitigation: a line in the release notes telling operators to skim their templates' `description`.

### [I3] Truncation is mid-word and feeds the composer (confidence 8/10)

`text = uc.strip()[:_MAX_USE_CASE_CHARS]` — a >200-char use case is cut mid-word with no ellipsis, and that same truncated string becomes `starter_prompt`, so clicking it pre-fills a sentence that stops mid-word. The cap is right; consider trimming at a word boundary and appending `…` for the *title* while keeping the full text (or a clean truncation) as the prompt. Cosmetic, non-blocking.

---

### Checked and clean

- **XSS** — no `v-html`; `{{ p.title }}` / `{{ agent.description }}` escaped.
- **Type safety of agent-supplied JSON** — non-list, non-string, blank and over-long entries all handled, with tests. Bundled templates all use plain-string `use_cases` (9 checked); object-shaped entries would silently drop, which is acceptable given none exist.
- **Cap discipline** — 6 × 200 enforced and tested; `cornelius` declares 9 and is correctly capped.
- **No auth/endpoint change** — enrichment of the existing roster only.
- **Docs** — `architecture.md` Workspace section updated to match the shipped ladder, including the `/info` fix.

**Summary: 1 finding needing a decision (C1), 2 non-blocking notes. No blocking defect.**

vybe:
While planning trinity#2101 (bounding the briefing hint grid), both of its independent review passes converged on a suggestion that belongs to this PR's ladder rather than to #2101 — leaving it here for consideration:

**`allow is None` is omission, not curation.** The ladder comment says "an operator-exposed playbook set is the curated capability surface and wins outright" — but when no connector allow-list is configured (the default, and exactly #2101's trigger case), `resolve_exposed_playbooks(live, None)` returns *everything* `user_invocable` by omission. Nothing about that set was chosen.

**Proposal (~5 lines on this PR's seam):** thread `allow` into the ladder decision —

- explicit allow-list (a real curation act) → playbooks win outright, exactly as now;
- `allow is None` and the template declares `use_cases` → prefer the use-case hints (author-written full sentences, already sanitized + capped 6×200 here) over N humanized slugs;
- `allow is None` and no `use_cases` → playbooks, as now.

Scoped to the briefing only — the MCP connector's own `allow is None ⇒ expose all` default is untouched. It turns the default briefing from "6 arbitrary humanized slugs" into "6 author-curated suggestions".

#2101 lands the display bound either way (6 visible + counted expander, plus a ≤24 belt as a single slice at `_agent_briefing`'s return), so the two compose regardless of merge order.

vybe (approved):
Validated via /validate-pr: conflicts with #2113 resolved semantically (use-case fallback ladder now routes through the #2101 hint belt at the return — the belt's own guard test confirms the call-site survived). Security battery clean; architecture.md §Workspace updated; 69/69 unit tests pass locally incl. both sides' new coverage. Cross-tracker Fixes ent#380 — status-in-dev to be set manually.

PR #2115, opened 2026-08-11
refactor(agents): retire the vestigial agent type taxonomy (#2104)

### LINKED ISSUE #2104: refactor: retire the vestigial agent `type` taxonomy (business-assistant everywhere)
## Summary

The free-text agent `type` field (default `"business-assistant"`) is a vestigial taxonomy from before tags existed. The `github:` creation path never reads `template.yaml` metadata into the create config, so virtually every agent carries the default — and the surfaces that render it (Permissions tab, agent info chip, public-link fallback) show a stale, meaningless `[business-assistant]` label. Nothing functional depends on the value: categorization is covered by tags, the system agent keys on `is_system`, and the skill-runner tile variant can key on that agent's fixed name. Retire it end-to-end. **No backward compatibility required** — remove it from API/MCP responses outright rather than deprecating.

## Context

Where it comes from:
- `AgentConfig.type` defaults to `"business-assistant"` (`src/backend/models.py`); creation bakes it into the container as the `trinity.agent-type` Docker label + an `AGENT_TYPE` env var (`src/backend/services/agent_service/crud.py`). Rebuild fallbacks re-apply the default (`src/backend/services/agent_service/lifecycle.py`, `deploy.py`). The MCP `create_agent` tool advertises the same default.
- The `AGENT_TYPE` env var is read by nothing under `docker/base-image/` — dead weight.

Where it shows (the confusion):
- `src/frontend/src/components/PermissionsPanel.vue` renders `[business-assistant]` next to every agent in the permission list.
- `src/frontend/src/components/InfoPanel.vue` renders a type chip (fed by the `/info` endpoint reading the label).
- `src/backend/routers/public.py` uses the label as a public link's *display-name fallback* — a visitor can see an agent titled "business-assistant".

What actually consumes it:
- `src/frontend/src/components/AgentTile.vue` keys the skill-runner tile variant off `type === 'skill-runner'`; the runner is a fixed-name singleton (`trinity-skill-runner`), so the name is the sturdier key and the `type` check is replaceable.
- Nothing keys on `system-orchestrator` (`is_system` covers the system agent everywhere).
- `src/backend/routers/avatar.py` builds a name→type map but only uses it as an existence check.

Drive-by bug found while tracing: the agent-create audit entry logs `getattr(config, "agent_type", None)` (`src/backend/routers/agents.py`) but the field is named `type`, so it has always logged `null`.

Going forward, tags (including the reserved-namespace pattern established for org tags) are the platform's classification mechanism; a single-valued free-text `type` is strictly redundant.

## Acceptance Criteria

- [ ] `AgentTile.vue` keys the skill-runner variant off the fixed agent name instead of `type` (lands first; the enterprise provisioner's `type=` write becomes inert and is dropped in a private-repo follow-up).
- [ ] The `"business-assistant"` default and all fallbacks are removed (`models.py`, `lifecycle.py`, `deploy.py`, MCP `create_agent` param); creation no longer writes the `trinity.agent-type` label or the `AGENT_TYPE` env var.
- [ ] `type` is removed from API and MCP responses (`AgentStatus`, `/api/agents`, `/api/agents/{name}/info`, MCP `list_agents`/`get_agent`) — no deprecation window.
- [ ] Display surfaces stop rendering type: `PermissionsPanel.vue`, `InfoPanel.vue`; `routers/public.py` falls back to the agent's display label / name, never the label — so the existing fleet's stale baked labels are never rendered anywhere (no forced recreates).
- [ ] **Permissions tab (Agent Detail): each agent row shows the canonical slug alongside the display name** (replacing the removed `[type]` suffix; shown when the display label differs from the slug, so agents stay unambiguous without hovering for the tooltip).
- [ ] `template.yaml type:` remains parseable but ignored (no compatibility check added; existing templates don't break).
- [ ] The agent-create audit `agent_type` typo is fixed or the always-null field dropped.
- [ ] User docs: a short retirement note added under `docs/user-docs/` (agent `type` is retired; use tags for categorization), and existing references to agent type updated.
- [ ] `docs/memory/architecture.md` reference to `trinity.agent-type` updated.

## Technical Notes

- Frontend self-heals for new agents (`v-if` on truthiness), but the display-side suppression is what cleans up the existing fleet — containers keep baked labels until recreate.
- `docker_service.py` currently maps the label with an `"unknown"` fallback into `AgentStatus.type`; removing the field removes the fallback too.
- Keep `SYSTEM_AGENT_TYPE`/runner constants only if the private modules still want them internally; nothing in OSS should read them.

trinity-ability:
## Summary
- Retires the free-text agent `type` field (default `business-assistant`) end-to-end — models, all five creation/recreate writers (`trinity.agent-type` label + `AGENT_TYPE` env, read by nothing), API/MCP/WS response surfaces, and every display surface. No deprecation window per the issue.
- The existing fleet cleans up with **zero forced recreates**: stale baked labels lose their last reader, and public-link titles now fall back to the owner-set display label → slug (never the taxonomy value a template controlled — a small anonymous-visitor spoof surface removed).
- Permissions rows replace the `[type]` suffix with the canonical slug whenever a display label hides it; the skill-runner tile keys on the fixed `trinity-skill-runner` name; the enterprise provisioner's `type=` kwarg stays inert (Pydantic ignore — pinned by test) until its private-repo follow-up drops it.
- `template.yaml type:` remains parseable but ignored (existing templates keep working — proven by the recreate test fixture); the always-null `agent_type` audit field (typo for `type`) is dropped.

## Changes
- **Backend**: `models.py` (both fields), `crud.py` (env/label/config-yaml/local-template mutation + the `agent_created` broadcast — a reader inside the creation try-block that unit tests can't see because `ws_manager` is nulled), `lifecycle.py`, `deploy.py`, `system_agent_service.py`, `cornelius_agent_service.py`, `docker_service.py`, `agent_files.py` (3 fallbacks + a strip of `type` from older base images' proxied `/info` responses), `public.py`, `avatar.py`, `ops.py` (both fleet dicts), `agents.py` (audit + container-less fallback), `permissions.py` (the payload feeding the Permissions tab)
- **Frontend**: `PermissionsPanel.vue` (slug-when-label-differs via `hasDistinctLabel`), `InfoPanel.vue`, `AgentTile.vue`, `MobileAdmin.vue` (+ orphaned CSS), `stores/network.js`
- **MCP server**: `create_agent` param removed, `Agent`/`AgentConfig`/`TemplateInfo` interfaces, tool descriptions; list/get responses lose the field via passthrough
- **Agent server**: `info.py` stops returning `type` (belt; the backend strip covers un-rebuilt images)
- **Docs**: CLAUDE.md label list, user-docs retirement note (tags are the classification mechanism), 7 feature flows, learnings entry
- **Tests**: new `tests/unit/test_2104_agent_type_retired.py` (Pydantic ignore contract, schema, stale-label tolerance, source-level residue guard over 4 trees) + `test_start_agent_skip_inject.py` converted to negative assertions; registered in `tests/registry.json`

## Consumer note (no backward compat, per issue)
`type` leaves `/api/agents`, `/api/agents/{name}` (incl. the recovered-agent fallback), `/api/agents/{name}/info`, the permissions payload, ops fleet-status, the `agent_created` WS event, and MCP `list_agents`/`get_agent`. Cross-repo: grep **trinity-ops-agent** for readers before release; include in release notes.

## Test Plan
- [x] `pytest tests/unit/test_2104_agent_type_retired.py` — 5/5 (incl. residue guard that caught two comment stragglers during development)
- [x] Impacted suites: `test_start_agent_skip_inject` + `test_runtime_label_fast_path` + `test_1811_recovery_container_parity` (env-parity guard) + `test_ent89` green; `test_1484_create_agent_characterization` 40/40
- [x] MCP server: `tsc --noEmit` clean, 236/236 tests
- [x] Frontend production build clean
- [ ] Known pre-existing: 3 `test_2006_deploy_residue` failures when co-scheduled with `test_start_agent_skip_inject` in one process — reproduced identically on clean dev (randomized-suite stub-leak class), unrelated to this diff

Fixes #2104


vybe (approved):
Validated via /validate-pr: churn-file conflicts vs dev resolved (registry.json rebuilt from stages, 175 unique entries; both learnings entries kept). CI's regression diff caught a sixth unmapped reader of the retired field (test_1793 asserting type projection) — fixed to the PR's own inertness contract, full suite green on re-run. Security battery clean; 7 feature-flows + user docs updated; residue-guard test present. 14/14 targeted unit tests pass locally.

PR #2119, opened 2026-08-12
fix(subscriptions): stop .env-resident ANTHROPIC_API_KEY shadowing subscription auth (#2114)

### LINKED ISSUE #2114: bug: .env-resident ANTHROPIC_API_KEY shadows subscription auth in the per-spawn execution env (#1999), defeating SUB-003 auto-switch
## Summary

Since #1999 (`f744f349`, "rebuild the execution env per spawn so a removed .env key stops applying"), `build_execution_env()` merges the agent's `~/.env` file into the environment of **every** spawned runtime subprocess. On a subscription-backed agent whose `.env` happens to contain a stale `ANTHROPIC_API_KEY`, every `claude` spawn now receives that key — and Claude Code prefers an API key over `CLAUDE_CODE_OAUTH_TOKEN` — so all executions hard-fail (e.g. "Credit balance is too low" when the key's account has no credits) despite a valid, working subscription token. SUB-003 auto-switch then cycles the agent through every registered subscription without effect, mis-attributes the identical failure to each one, and fills the 2h skip-list until it reports "no viable alternative subscription is available" while every subscription is actually healthy.

## Component

Agent Runtime (agent-server `execution_env.py`) + Backend (SUB-003 `subscription_auto_switch.py`)

## Priority

P1

## Error

```
Execution failed with no output (exit code 1): Credit balance is too low. To resolve: (1) add credits to your Anthropic account at console.anthropic.com, or (2) assign a subscription token with available usage in Settings -> Subscriptions.
```

Every execution fails within seconds of spawn (auth rejection, not task failure), regardless of which subscription is assigned.

## Location

- **File**: `docker/base-image/agent_server/services/execution_env.py` — `build_execution_env()` (precedence: container baseline → `.env` → runtime overrides → extra)
- **File**: `src/backend/services/subscription_auto_switch.py` — `_hot_reload_subscription_token()` (sends `remove_api_key=False`)
- **File**: `docker/base-image/agent_server/routers/credentials.py` — `/api/credentials/reload-token`

## Root Cause

Interaction between #1999 and #1089/SUB-003:

1. #1999 makes the execution env a pure function of `INITIAL_ENV + .env + RUNTIME_OVERRIDES`, with `.env` parsed fresh at every spawn. This is the intended security behavior (removed credentials stop applying) — but it also means `.env` became a **new authoritative source** for `ANTHROPIC_API_KEY`.
2. `lifecycle.py` pops `ANTHROPIC_API_KEY` from **Config.Env** at container creation for subscription-backed agents, and the SUB-003 hot-reload passes `remove_api_key=False` with the comment "subscription-backed agents never carry ANTHROPIC_API_KEY in env (popped at create time, lifecycle.py)". That assumption was true for Config.Env, but nothing cleans the **`.env` file**, which lives on the persistent workspace volume and survives every recreate.
3. Result: a stale `ANTHROPIC_API_KEY` in `.env` (from an old credential push, an agent writing its own file, an export that captured it) silently shadows the subscription token on every spawn. Nothing in the switch path can remove it: `os.environ` mutation, the runtime override, and the writable-layer token file were all verified delivering the correct working OAuth token — and the next spawn still authenticated with the dead API key.
4. Secondary defect: the resulting auth failures are recorded as rate-limit/auth events against whichever subscription is currently assigned, so healthy subscriptions get skip-listed for 2h each. After one rotation the agent reports "no viable alternative subscription is available" — the operator-facing signal actively points **away** from the real cause.

## Field Observation (sanitized)

- Subscription-backed agent, ~68 scheduled executions/day, 100% success for 2 weeks.
- An `ANTHROPIC_API_KEY` for an out-of-credits account had been sitting in its `~/.env` since a credential push 2 weeks earlier. (It briefly caused identical failures on the day it was written — pre-#1999 ghost-env behavior — then went dormant after a recreate.)
- The day the fleet's agents were rebuilt onto the post-#1999 base image, every execution began failing in seconds with "Credit balance is too low".
- SUB-003 rotated through all 4 registered subscriptions (each switch logged `restart_result: 'hot_reloaded'`); the persisted override token after the last switch was byte-identical (sha256) to the token a healthy peer agent was succeeding with **in the same minute**. The next spawn still failed.
- Removing `ANTHROPIC_API_KEY` from the agent's `.env` resolves it (per #1999's own design, the removal takes effect at the next spawn).

## Reproduction Steps

1. Create a subscription-backed agent (valid `CLAUDE_CODE_OAUTH_TOKEN`, no `ANTHROPIC_API_KEY` in Config.Env) on a post-#1999 base image.
2. Inject an `ANTHROPIC_API_KEY` belonging to an account with no API credits into the agent's `.env` (any write path: credentials inject, agent editing its own file).
3. Trigger any execution → fails with "Credit balance is too low" despite the valid subscription.
4. Observe SUB-003 switching subscriptions with successful hot-reloads and no effect; after one rotation, "no viable alternative subscription is available".

## Suggested Fix

Layered; (1) is the minimal correct fix:

1. **SUB-003 hot-reload should send `remove_api_key=True`** for subscription-backed agents. The `None` force-unset runtime override is exactly the semantics #1999 built for this ("a plain dict merge cannot express force-unset") — it suppresses the `.env`-sourced key at spawn time without touching the file.
2. **Spawn-side guard**: in `build_execution_env()`, when a `CLAUDE_CODE_OAUTH_TOKEN` runtime override is active (subscription auth is the explicit, most recent signal), drop `ANTHROPIC_API_KEY` from the `.env` merge — mirrors the existing "an explicit hot-reload wins" precedence rationale.
3. **Classifier/observability**: when an auth-class failure occurs while a fresh OAuth override is in place AND `.env` carries `ANTHROPIC_API_KEY`, emit an operator notification ("API key shadowing subscription auth") instead of recording the failure against the subscription — prevents healthy subscriptions from being skip-listed and surfaces the actual cause. The `env_drift` report added by #1999 already exposes the raw signal.

## Environment

- Regression first present: `f744f349` (#1999), post-v0.8.5 dev
- Observed on: dev @ `f1809975`, prod compose, agents rebuilt onto the #1999 base image the same day
- Any agent whose `.env` contains a stale `ANTHROPIC_API_KEY` will hit this on its next rebuild against a post-#1999 base image

## Related

- #1999 (per-spawn execution env rebuild — the regression vector)
- #1089 (subscription token hot-reload — the mechanism defeated)
- #799 (per-agent switch lock — switch serialization, working as designed here)
- `docker/base-image/agent_server/services/execution_env.py`
- `src/backend/services/subscription_auto_switch.py`

trinity-ability:
## Summary
- Post-#1999 the spawn env re-reads `.env` at every spawn, making the file a second authoritative source for `ANTHROPIC_API_KEY` — which Claude Code prefers over `CLAUDE_CODE_OAUTH_TOKEN`. A stale key on the persistent workspace volume (never cleaned by any recreate) shadowed subscription auth on every execution; SUB-003 then mis-attributed the identical auth failure to each healthy subscription, skip-listed all of them for 2h, and reported "no viable alternative" — pointing operators away from the cause.
- **L1 (backend, reaches the deployed fleet with a backend deploy alone)**: `_hot_reload_subscription_token` sends `remove_api_key=True`, gated on the agent's Claude runtime (`trinity.agent-runtime` label; non-Claude agents keep `False` — their scripts may legitimately use the key, which never shadows there). The force-unset machinery already exists in deployed post-#1999 images.
- **L2 (agent server, restart-durable — needs the base-image rebuild)**: `arm_subscription_auth_guard()` at boot force-unsets `ANTHROPIC_API_KEY` + `ANTHROPIC_AUTH_TOKEN` via the existing #1999 override layer when the boot baseline carries a truthy `CLAUDE_CODE_OAUTH_TOKEN` on a Claude runtime. `startup.sh` exports the rotated override token before the server launches, so the arm holds across every plain stop/start — the window L1 alone cannot cover.
- **L3 (observability — suppression is data, not just a log)**: `env_drift_report` marks force-unset keys `suppressed_for_spawn` (iteration set widened to override-only keys, so the drift surface can't show all-green over an active suppression); the reload endpoint returns `env_shadow` (names only) which the backend logs at WARNING at switch time; `build_execution_env` warns per key (memoized, re-arming on remove/re-add).

## Changes
- `src/backend/services/subscription_auto_switch.py` — flag flip + runtime gate + env_shadow WARNING (shape-validated: a tampered agent response degrades to no-warning, never demotes a successful hot-reload to a restart)
- `docker/base-image/agent_server/services/execution_env.py` — the arm, `SUBSCRIPTION_SHADOW_KEYS`, memoized suppression warning, drift-report marker, override-dict snapshot during iteration
- `docker/base-image/agent_server/routers/credentials.py` + `models.py` — both shadow-key spellings unset; `env_shadow: List[str]` on `TokenReloadResponse`
- `docker/base-image/agent_server/main.py` — boot wiring (AST-pinned by test)
- Tests: `tests/unit/test_2114_subscription_shadow_guard.py` (14 tests — trigger matrix, override semantics, observability, the deliberately-unfixed stale-OAuth residual pinned by name), pingpong payload pin flipped + order-deterministic helpers stub + tampered-shape test, reload-endpoint fixture teardown (pre-existing override leak) + env_shadow coverage, #1999 drift-shape/wiring pins updated
- Docs: requirements `security.md` §20.4/§20.6, `architecture.md` (#1999 + #1089 blocks), `subscription-auto-switch.md` flow, `learnings.md`, CSO diff-audit report, test registry

## Test Plan
- [x] `pytest tests/unit/test_2114_subscription_shadow_guard.py tests/unit/test_1999_execution_env.py tests/unit/test_reload_token_endpoint.py tests/unit/test_subscription_auto_switch_pingpong.py` — 94 passed
- [x] 8-file neighborhood (incl. `test_2023`, `test_subscription_reassign_hotreload`, `test_subscription_auto_switch_no_cred_import`, `test_ent127_predicate`) — 254 passed across 3 random seeds and the poisoned order
- [ ] `/verify-local` FULL (base image + real agent boot) before this PR flips ready-for-review — `docker/base-image/**` is touched
- [x] `/review` (0 critical; 4 informational, all fixed in `265ab415`) + `/cso --diff` (no findings at the ≥8 gate; report committed)

## Rollout notes
- **Mixed versions are safe in both directions, no ordering requirement**: old image → 404 → existing restart fallback; #1089-era image → `os.environ` pop reaches its spawns; new image + old backend → the boot arm covers baseline agents. The restart-durable half (L2) reaches an agent on its next recreate onto the rebuilt base image (release-note line, ent#123 precedent).
- **Known accepted consequence** (release-notes line): a *funded* `.env` API key that was silently billing instead of the assigned subscription flips to subscription auth after rebuild+recreate. Matches the declared assignment; the managed path to API-key auth remains clear-subscription (which recreates).

## Follow-ups (listed, deliberately not filed by the agent)
1. Stale `.env` `CLAUDE_CODE_OAUTH_TOKEN` still beats the rotated baseline token after a restart — same class, different fix shape (precedence carve-out, not unset). Pinned as known behavior by `TestKnownResidual::test_stale_env_oauth_still_beats_rotated_baseline_token`.
2. `ANTHROPIC_BASE_URL` in agent-writable `.env` redirects Claude calls (bearer-exfil surface) — candidate for `PROTECTED_KEYS`, needs legit-proxy-use analysis first.
3. Credential-inject-time warning when pushing `ANTHROPIC_API_KEY` to a subscription-backed agent.
4. Full auth-failure classifier / operator notification (the issue's original L3 idea) — with L1+L2 the trigger is structurally unreachable at spawn.

## Pre-existing issue observed (not addressed here)
`tests/unit/test_subscription_reassign_hotreload.py` fixtures error (7 errors) whenever `test_subscription_auto_switch_no_cred_import.py` runs first in the same session — reproduced byte-identical on untouched `dev` (`06e10ea2`). A sys.modules-stub order dependency of the #762 class; the learnings entry in this PR documents the class.

Fixes #2114


vybe (approved):
Validated via /validate-pr: named regression suite (test_2114_subscription_shadow_guard, 382 lines) + hardened per /review + CSO diff audit committed. Churn-file refresh vs dev resolved (registry 176 unique entries, both learnings entries kept); 94/94 targeted unit tests pass locally. Security battery clean (4.5 hits are test fixtures). L1/L2 deploy staging documented — L2 needs the base-image rebuild.

Issue #381, opened 2026-04-18
Epic: Git Sync Reliability Improvements


vybe:
## Summary

Comprehensive improvements to Trinity's git sync system to address silent data loss, conflict deadlocks, and observability gaps. Based on real incidents from 2026-04-17 where three agents hit multiple failure modes.

## Context

Trinity's git-sync design (trinity/* working branch per instance, auto-commits, pull redirected to origin/main) is architecturally sound. However, the conflict-handling, branch-ownership, and observability layers have **7 verified failure modes**:

| ID | Problem | Severity |
|----|---------|----------|
| P1 | Silent desync accumulation (no sync health signal) | High |
| P2 | Parallel-history deadlock (both UI options wrong) | High |
| P3 | No "reset template, preserve state" primitive | Medium |
| P4 | Git internals leak into error modal | Medium |
| P5 | No branch-ownership enforcement (silent clobber) | Critical |
| P6 | Working-branch divergence hidden by ahead/behind | High |
| P7 | Three generate_instance_id() call sites, no cross-check | Medium |

## Proposed Solutions

| Solution | Description | Priority | Issue |
|----------|-------------|----------|-------|
| S7 | Branch ownership enforcement | P1 | #382 |
| S4 | Persistent state allowlist | P1 | #383 |
| S3 | Reset-preserve-state operation | P1 | #384 |
| S2 | Parallel-history detection | P2 | #385 |
| S5 | Operator-readable diagnosis | P2 | #386 |
| S0 | Self-hosted git URL support | P2 | #387 |
| S1 | Sync health observability | P2 | #389 |
| S6 | Fleet sync audit endpoint | P3 | #390 |

## Rollout Phases

| Phase | Ships | Unblocks |
|-------|-------|----------|
| 0 | S0 (#387) + S4 (#383) | Local repro harness, reset-preserve-state |
| 1 | S2 (#385) + S3 (#384) + S5 (#386) + S7 (#382) | UI recovery from conflicts, branch safety |
| 2 | S1 (#389) | Silent desync prevention |
| 3 | S6 (#390) | Ops playbooks, bulk health checks |

## Evidence

Full proposal with code citations: `ability-trinity-git-improvements-proposal.md`

Real incident: 2026-04-17, three agents (polygon-vybe, alpaca-vybe-live, alpaca-vybe-live-2) hit P1-P7. One lost 50KB of trading data from silent branch clobber.

## Child Issues

[...]

PR #2120, opened 2026-08-12
feat(workspace): absorb the Session surface, with continuity parity (ent#358)

dolho:
Related to abilityai/trinity-enterprise#358

## What

Trinity had two surfaces for one job — Agent Detail's Session mode and the Workspace both held a continuing conversation with an agent. The Workspace becomes the one.

## The part that isn't the removal

The surface being retired was the **more capable** one:

| | Session mode | Workspace chat (before) |
|---|---|---|
| engine | `claude --print --resume <uuid>` | stateless `execute_task` |
| continuity | tool results, mid-skill state, reasoning state | last N messages replayed as prompt text |

Replay recovers what was *said*, not what the agent *did*. Removing the first without closing that gap looks fine in every demo and fails exactly when a conversation gets long enough to matter. So parity lands first, and it is most of this PR.

## The prerequisite that wasn't

The issue named ent#286 (Workspace streaming) as a hard prerequisite: *"the session tab streams; workspace chat does not."*

It doesn't. `SessionPanel.vue` had no `EventSource`/`WebSocket`/SSE, `stores/sessions.js` sent a plain awaited POST, and `routers/sessions.py` exposed no stream endpoint — replies arrived whole, with a reattach poller for long turns (#1376/#759). Absorbing a non-streaming surface into a non-streaming one is not a regression, so ent#286 is **decoupled** from this work (still worth doing on its own merits).

## Changes

**Engine** — `services/session_turn_service.py` (new): the resume lock, in-flight sentinel, runtime gate and resume-fallback lifted out of `routers/sessions.py` (business logic in a router, Invariant #1). Both surfaces call `run_resumable_turn`; each owns only its own persistence. The router keeps the original private aliases, so its behaviour — and the tests pinning it — are unchanged. `ResumeLockBusy` subclasses `HTTPException(429)` so the Session router is byte-identical while the Workspace catches the precise type and raises its own `ClientPortalError(429)`.

**Schema** — `enterprise_portal_sessions` gains `cached_claude_session_id`, `last_resume_at`, `consecutive_resume_failures` across all four tracks (`db/schema.py`, `db/tables.py`, SQLite `portal_session_resume`, Alembic `0037_portal_session_resume` — Invariant #3). Additive; existing threads run one cold turn, then resume.

**History replay becomes cold-turn-only** — on a resumed turn the block re-pays for context the session already holds and puts a summary of the conversation beside the conversation; on a cold turn, a reaped JSONL, or a Codex agent it is the *only* continuity there is. `portal_chat` composes two messages so `cold_message` carries it into the fallback retry.

**Reaper keep-set union (the quiet one)** — `session_cleanup_service` built its keep set from `agent_sessions` alone. With Workspace threads resuming, that deletes a live Workspace JSONL an hour after it is written and every thread on the agent goes cold: no error, no log, no failed request. Now the union of both tables, and a read failure on **either** half aborts the sweep rather than reaping blind.

**UI** — `SessionPanel.vue` + the Session-mode toggle deleted; the Chat tab is stateless plus a "Continue in Workspace" link; `?tab=session` redirects to `/workspace?agent=<name>` (query-preserving, `router.replace`, guarded first in **both** `onMounted` and `onActivated` — the #1672 KeepAlive class); `Portal.vue` resolves that landing through the pure `resolveAgentLanding()` in `portalUtils.js`.

`agent_sessions` rows, its six endpoints, and `stores/sessions.js` are untouched — AC #3. Only the entry point went away.

## Acceptance criteria

- [x] Session tab removed from Agent Detail; the Chat tab keeps its stateless legacy mode
- [x] `?tab=session` / session deep links redirect into the equivalent workspace chat, query-preserving
- [x] Existing `agent_sessions` rows remain readable — no data loss
- [x] JSONL reaper and resume machinery keep working for whichever surface owns continuity
- [x] No regression in conversation continuity — a Workspace turn resumes the way a Session turn did

## Verification

```
558 passed, 3 skipped        # 38 test files: session, portal, cleanup, schema,
                             # migration, alembic + the #1672/#1673/#1849/#1444
                             # regressions — run both with and without -p randomly
 84 passed                   # frontend vitest
 ✓ built in 2.83s            # vite production build
```

`tests/unit/test_ent358_workspace_absorbs_session.py` (13 cases) was mutation-checked, not just run green: reverting to "always replay history" fails 2 cases, and dropping the portal keep-set union fails 2 others.

Also green: `lint_sys_modules.py`, `test_models_centralized.py`, the enterprise-docs guard, Alembic single-head, and the SQLite migration exercised against a pre-migration table (idempotent, backfills 0).

## Known limitations

- Workspace chat still doesn't stream (unchanged; ent#286).
- A pre-existing Workspace thread's next turn is cold — self-healing, one turn.
- Codex threads keep history replay (no `--resume`), so continuity quality now varies by runtime within one surface.
- `stores/sessions.js` keeps actions nothing calls; they map 1:1 to endpoints that are still live by AC #3, so pruning belongs with retiring those endpoints.


dolho:
## Code review — high effort, workflow-backed (38 agents, every candidate independently verified)

10 findings survived verification. Most cluster around the new pre-created execution row and the streaming turn path added in this PR, and several compound into one user-visible failure: **a live, billed turn reported as "not delivered", with a Retry that runs and bills it again.**

Grouped by the story they tell rather than by file.

---

### A. The double-billing chain (4 findings, one story)

**1. `services/session_turn_service.py:464` — the cold retry reuses an already-terminal `execution_id`**
The retry forwards the caller's pre-created id again via `**execute_kwargs`, so attempt 2 runs under an id whose agent-side stream has closed and whose row was already written FAILED.
*Scenario:* thread idle long enough for the JSONL reaper to remove its Claude JSONL → attempt 1 fails "No conversation found" → row CAS-written FAILED, browser gets `stream_end` and stops watching → the engine silently re-runs the whole 30–120s turn under the same id, invisible to the client.

**2. `components/portal/PortalConversation.vue:388` — the ~10.5s reply wait declares live turns undelivered**
`awaitPersistedReply` budgets 15×700ms from the moment the stream ends. Any turn persisting later is rendered red with a Retry button. The turn was dispatched and billed; retrying pays twice and repeats any non-idempotent side effect the agent already performed.

**3. `components/portal/PortalConversation.vue:437` — Retry can surface the *previous* turn's reply**
The poll compares the persisted assistant count against the **local** one. After a falsely-failed send, Retry dispatches a second real turn, then the first poll already sees N+1 (the first turn's reply) and returns it as the answer to the retry — while turn two keeps running and persists a third message attached to nothing.

**4. `components/portal/portalUtils.js:107` — self-thrown errors misreported as connection failures**
`deliver()` throws a plain `Error` ("The agent did not reply…") which has no `.response`, so the user is told *"Couldn't reach Trinity — check your connection"*. That's false (the request succeeded, the turn is still running) and it actively pushes them toward the Retry that double-bills.

---

### B. Data integrity (2)

**5. `client_portal/service.py:1186` — a pre-dispatch failure leaves an orphan RUNNING row**
The row is created before the background turn spawns, but no failure path in `_run` writes a terminal. A `ResumeLockBusy` (two tabs on one thread) raises *before* `execute_task` sees the id → the row sits RUNNING forever, and the cleanup watchdog later fabricates a FAILED "silent launch failure" against a healthy agent. Same for any raise from `_resolve_session_id` / `_collect_inbox_for_turn`.

**6. `client_portal/service.py:871` — a failed send + Retry duplicates the user message**
Persisting the user's half *before* the turn (added here so a mid-turn reload doesn't look like message loss) means a failed turn leaves the row behind, and Retry writes it again. The thread shows the message twice, `message_count` is inflated by the double `touch_portal_session`, and the duplicate is replayed into `convo_context` — telling the model the client asked twice.

---

### C. Invariant + correctness (2)

**7. `client_portal/router.py:466` — `/chat/stream` creates an execution with no `Idempotency-Key`**
Invariant #18 requires every producer boundary that creates an execution to accept one. The synchronous `/chat` path is wired; this new boundary shipped without it, so a retried dispatch creates a second execution and bills the turn twice.

**8. `views/Portal.vue:259` — a retired deep link can silently open a *different* agent**
When `resolveAgentLanding` returns null (agent no longer on the caller's roster), `activeAgent` falls back to `store.agents[0]`. An old `/agents/foo?tab=session` link lands the user in the first roster agent's conversation, believing it is `foo`, with no indication the target changed.

---

### D. Concurrency / efficiency (2)

**9. `components/portal/PortalRoom.vue:319` — the room poll can duplicate transcript rows**
`load()` has no in-flight guard, so the 3s poll can overlap `send()`/`addAgent()`; both read the same `since` cursor, both receive the same messages, both concat — duplicate bubbles and duplicate Vue keys on one seq, self-correcting only on a full reload.

**10. `client_portal/service.py:1075` — blocking full-keyspace Redis `SCAN` inside the SSE generator**
`get_turn_inflight_matches` runs `scan_iter` over the *entire* shared keyspace, and the SSE proxy calls it every 0.4s per attaching stream for up to 30s. On a busy instance that keyspace holds slot, breaker, heartbeat and idempotency keys for the whole fleet, and the synchronous scan blocks the event loop for every request the worker is serving.

---

### Suggested order

1. **A** as one change — the four are the same story and fixing them piecemeal leaves the double-bill reachable.
2. **B** — small, data-integrity.
3. **7** (invariant), **8** (silently wrong agent — arguably belongs higher, since it misroutes a user's message).
4. **9**, **10** — cheap.

Findings 9 and 10 are regressions from commits made *after* the branch's own testing pass; 10 was introduced within the hour, which is a fair argument for re-running this review after the fixes land.

vybe (approved):
## `/validate-pr` — methodology validation

Validated against the Trinity SDLC gates, with the ten review findings re-verified independently in the code rather than taken from the fix commit's word. Ran locally off the branch: **944 backend unit tests** (session / portal / cleanup / schema / migration / the `#1804`, `#186`, models-centralized and alembic parity guards) and **93 frontend vitest** — all green, matching CI's 6/6 pytest shards.

### Blocker found and resolved before merge — the submodule pin

The pin moved `2d64baa → 8ca1e9c9`, and **8ca1e9c9 was not on enterprise `main`** — it was the tip of the unmerged `feature/362-workspace-user-participant` branch.

That is load-bearing, not bookkeeping: `PortalRoom.vue` calls `/api/rooms` with a **portal** auth header, and admitting a `WorkspacePrincipal` to that surface is exactly what `trinity-enterprise#388` adds. The two halves have to land together. Left as-is, the next release-freeze recheck ("pin → main tip") would have rolled the rooms-participant support back out from under the frontend — silently, with green CI on both sides, since nothing in the public tree would change.

Resolved in order: `trinity-enterprise#388` squash-merged to enterprise main as `cbd339e`, tree verified byte-identical to `8ca1e9c9`, and the pin re-pointed at main in `706f216`.

### Verified, not assumed

Each of the ten findings was checked against the code:

| | Fix | |
|---|---|---|
| 1 | `retry_kwargs` strips `execution_id` — the cold retry gets its own row | ✅ |
| 2 | `awaitPersistedReply` bounded by the server's `in_flight_execution_id`, not a stopwatch | ✅ |
| 3 | reply baseline read from the server **before** dispatch | ✅ |
| 4 | `isTransportError()` separates axios failures from self-thrown errors | ✅ |
| 5 | `_fail_unstarted_execution` on every `_run` exit path, CAS-guarded | ✅ |
| 6 | `_persist_user_turn` idempotent against exactly the state a failed turn leaves | ✅ |
| 7 | `Idempotency-Key` on `/chat/stream`, `begin`/`complete`/`fail`, fail-open | ✅ |
| 8 | `unreachableAgent` — never substitutes a different agent for a named one | ✅ |
| 9 | room-poll in-flight guard | ✅ |
| 10 | reverse key `portal_inflight_exec:{id}` — O(1) GET, no fleet-wide SCAN | ✅ |

Auth on the two new backend routes holds: `/chat/stream` runs the roster gate, then the *same* rate-limit keys as the synchronous route (so it is not a way to buy turns by asking differently), then idempotency; `/executions/{id}/stream` gates on roster **and** `source_user_email`, with a uniform 404 (Invariant #8). Dual-track migration is correct — SQLite `portal_session_resume` + Alembic `0037` off `0036`, both idempotent in both orders. The reaper keep-set union fails **closed**, which is the right direction: skipping a sweep costs disk, reaping blind costs every live conversation.

### Follow-ups (none blocking)

1. **The multi-agent picker is not entitlement-gated.** The NavBar entry it replaces was `isEntitled('shared_sessions')`; the picker now offers "pick several agents" unconditionally, so on an OSS-only build it is `POST /api/rooms` → 404 → "Could not start that chat." Gating it needs the flag reaching a *portal* principal, which is why this is a follow-up and not a one-liner.
2. **ent#361 / ent#381 are undocumented.** The rooms UI (`PortalRoom.vue`, `PortalAgentPicker.vue`) and the Sessions-page retirement appear in neither `architecture.md` nor `requirements/`; only ent#358 and ent#286 do.
3. **The `#1804` parity guard does not scan `client_portal/`** — its `_SCAN_DIRS` is `("services", "routers")`. `_fail_unstarted_execution` is genuinely in the allowlist class today (no dispatch activity exists before `execute_task` step 3), so nothing is wrong now; the *next* terminal writer added there would be unguarded, which is precisely what the guard exists to prevent.
4. **`views/enterprise/Sessions.vue` is orphaned** — unrouted, still carried in `raw-color-baseline.json`.
5. **`awaitPersistedReply` has no absolute ceiling.** Correct in the normal case, since `clear_turn_inflight` runs in a `finally`; but a hard backend crash leaks the marker for its 1h TTL, and a tab would poll history every 700ms for that hour.

Cross-tracker refs do not auto-promote, so ent#358 / #286 / #361 / #381 need a manual `status-in-dev` after merge.

**APPROVE.**

Issue #946, opened 2026-05-26
experiment: pull-coordination pilot — route MCP chat_with_agent through the agent queue


vybe:
## Summary

Convert MCP `chat_with_agent` (agent → agent only) to enqueue a typed message into `BacklogService` instead of issuing a synchronous HTTP call. One sprint, smallest viable boundary, same agent code on both sides. Human-facing `/chat` is explicitly untouched.

## Context

Per `docs/planning/ORCHESTRATION_RELIABILITY_2026-04.md` §"Transition roadmap (gated, evidence-driven)" Phase 2, this is the actor-model proof-of-concept that gates whether Trinity folds further boundaries (`/task`, schedules, webhooks) into message-passing or stays on the planned Sprint D/D′ path.

The experiment cannot be scheduled until the pre-experiment postcard (#945) lands — message envelope + journal format must exist as a contract before code is written.

Existing primitives already provide most of the shape:
- `BacklogService` is a durable SQLite FIFO that becomes the mailbox
- `SlotService` becomes the processor rate gate
- `EventBus` (#306) carries the completion event back
- The unified `TaskExecutionService` funnel ingests the dequeued message

Sprint D′ (#525 idempotency keys, #526 dispatch circuit breaker) ships in parallel and the experiment can ride those.

## Acceptance Criteria

- [ ] **Blocked-by #945** — postcard must land first (envelope + journal format specified)
- [ ] MCP `chat_with_agent` with agent-scoped key drops a message into target agent's `BacklogService` queue and returns `{execution_id, status: "queued"}` immediately
- [ ] Caller subscribes to completion event via `EventBus` (Redis Streams from #306) OR polls `get_execution_result` — both paths supported
- [ ] **No change to human-facing `/chat`** — caller-scope == user is the existing synchronous path, untouched
- [ ] **No change to MCP `chat_with_agent` external contract for human callers** — only the agent → agent code path changes
- [ ] Idempotency-Key honored (via #525) — duplicate enqueue produces single execution
- [ ] Decision write-up at sprint end: simpler or hidden complexity? Names the next boundary to fold, or recommends staying on planned path. Saved under `docs/planning/` as a Phase 3 decision record.

## Technical Notes

- The "agent → agent only" gate is the auth-context check in `src/mcp-server/src/tools/chat.ts` — `scope === "agent"` routes through the new path; `scope === "user"` keeps the old sync path.
- Journal entry written by agent on every processed message (per #945 spec).
[...]

Issue #914, opened 2026-05-22
bug: chat_with_agent MCP tool returns 'fetch failed' but still queues execution — silent duplicates burn budget on naive retry


pandore:
## Summary

`mcp__trinity__chat_with_agent` in default sync mode (`parallel=false`) holds the MCP gateway → backend → agent HTTP chain open for the entire agent execution. When the agent takes longer than the gateway's request timeout (~30-60s observed), the tool call returns `Tool 'chat_with_agent' execution failed: fetch failed` to the caller — but **the request was successfully queued on Trinity and the agent IS running it**.

A caller that doesn't know this will retry, queueing duplicates that Trinity then kills as concurrent conflicts. Today's example: a 3-lead dry-run on `bdr-agent` cost ~$2-4 of wasted compute and ~12min of agent time across 4 attempts (1 legitimate + 3 retries from `fetch failed` responses, 2 of which were killed mid-execution by the concurrent-duplicate guard).

This is the MCP-client surface of [#408](https://github.com/Abilityai/trinity/issues/408) (long-running backend→agent dispatch). The concurrent-duplicate kill is related to [#912](https://github.com/Abilityai/trinity/issues/912) (OrphanSweep false-kills).

## Component

MCP server / Trinity gateway

## Priority

P2 — has a workaround (`parallel=true, async=true`) but the default sync path is the obvious thing to reach for, and the silent-queue behavior is invisible until you check `list_recent_executions`.

## Reproduction Steps

1. Call `mcp__trinity__chat_with_agent` against any agent with a task that takes longer than the MCP gateway's request timeout (anything enrichment-heavy: e.g., daily-lead-outreach's Step 4 Apollo+Apify pass). Default sync mode, no `parallel`/`async` flags.
2. Observe: tool returns `Tool 'chat_with_agent' execution failed: fetch failed` within ~30-60s.
3. Assume the request didn't land. Retry.
4. Call `mcp__trinity__list_recent_executions` — you'll see the original request *is* still running, plus your retry now queued behind it.

## Observed evidence (bdr-agent, 2026-05-22)

A single intended dry-run produced 4 execution rows:

| execution_id | status | started | duration | notes |
|---|---|---|---|---|
| `fZv-iXtUXSolY1wzPO7T6w` | running | 12:28:29Z | ongoing (~17min+ at obs) | the legitimate one |
[...]

PR #2124, opened 2026-08-12
docs(testing): close the measurement traps in the #1766 soak set before the flip

obasilakis:
## Why now

All three defects below are only fixable **before** the pilot is flipped. Afterwards the evidence needed to catch them no longer exists — `schedule_executions` does not record what the caller asked for, so the distinction is unrecoverable from the row.

## What was wrong

**1. M6 over-credits pull, invisibly.**
Under push, a caller that gives up kills the turn and the row is written `failed` with `Task execution timed out after N seconds`. Under pull that budget no longer governs the row — the lease is `execution_timeout_seconds + SLOT_TTL_BUFFER` — so the work runs to its natural end and the row lands **`success`**, while the caller still got nothing in time. The failure moves off the row rather than going away.

No query over `schedule_executions` can separate "pull fixed it" from "pull outran the caller." And §9 told the reader to watch for a `failed` share *above* baseline — so a lower one reads as a pass.

Adds **M9**, the caller-side counter-check (`[Chat Timeout Recovery]`, the #914 receipt) — the only observable that survives the flip — plus the divergence table that makes M6 and M9 interpretable together.

**2. The baseline was specified as a single percentage.**
The classes move independently and only one is about pull:

| class | meaning | treatment |
|---|---|---|
| `lost_result` | agent finished, result never landed, watchdog wrote it off | **the target** — drive to zero, surfaces as M5 under pull |
| `caller_timeout` | caller's budget expired | the trap above — disappearance proves nothing |
| `quota` | subscription limit | exogenous — exclude |

Collapsed into one number, a treatment window landing in a fresh quota week looks like a win. §9 now carries the breakdown query and what each class means; the abort criterion excludes `quota` (reverting pull cannot fix a subscription limit).

**3. M6's query was a rolling `interval '7 days'`**, which straddles the flip and blends control rows into the treatment number. Now bound on `:flip_ts`, defined as the pilot agent's **recreate** — the flag only bakes in at create/recreate, so a backend restart alone leaves the agent not pulling.

## Also

- **M3/M4 need sampling, not one run at the end.** M3 counts *concurrent* leased rows, so a daily run lands off-peak and never sees the overbooking it exists to catch. M4 is worse: its predicate is `status IN ('queued','running')`, and `cleanup_service`'s stale sweep eventually fails stranded rows to `failed`, at which point they leave the predicate — a soak that genuinely lost executions reports zero rows. The set table gains a cadence column.
- **Abort criteria** gain the pre-existing-violation rule: a candidate pilot already red on a critical invariant must be resolved or carved out before the flip, or the soak trips its own abort at t=0.

## Scope

Docs only — one file, no code, no schema, no behaviour change. The baseline measurements that motivated this are posted on #1766.

Refs #1766, #946, #914, #1871

### DIFF
--- a/docs/testing/PULL_MIGRATION_TESTING.md
+++ b/docs/testing/PULL_MIGRATION_TESTING.md
@@ -369,22 +369,60 @@ traffic can reach the durable queue (#2048) — a cron-driven agent will read ~0
 under M1 below and pick from its `pull_eligible` column *before* flipping
 anything; a week of soak on a cron-only agent measures nothing.
 
-Then run M4 + M6 + M7 on the pilot for the 7 days *preceding* the flip and keep
-the output. Without a baseline, a 4% success-rate dip during the soak is
+Then run M4 + M6 + M7 + M9 on the pilot for the 7 days *preceding* the flip and
+keep the output. Without a baseline, a 4% success-rate dip during the soak is
 unattributable and the write-up degrades to "felt fine".
+
+**Record the baseline as a breakdown by failure class, never as one percentage.**
+A single "7.7% failed" is unusable, because the classes move independently and
+only one of them is about pull:
+
+```sql
+SELECT CASE
+         WHEN error LIKE 'Subscription usage limit%'      THEN 'quota'
+         WHEN error LIKE 'Task execution timed out%'      THEN 'caller_timeout'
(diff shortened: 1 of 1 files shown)


vybe (approved):
Validated via /validate-pr: security scans clean, required checks green, scope minimal.

PR #2125, opened 2026-08-12
feat(library): tabbed Library + per-skill "assigned to" agent list (trinity-enterprise#384)

trinity-ability:
## Summary

- The Library's three stacked sections become **tabs** — Agent Templates · Systems · Skills — deliberately reversing ent#263's stacked-sections choice for the *whole* page rather than special-casing Skills, so it keeps exactly one model.
- Each skill block now lists **which agents already hold it**, closing the "fleet assignment read" that requirements §22.3 named under *Not Built*.
- One new access-scoped read endpoint. **Zero writes, zero schema change, zero migration.**

Assignment stays a **write on the per-agent Skills tab** (ent#182: one skill model). The Library gained a read, not a second write path.

## Scope

The issue originally asked for assign/unassign controls on the Library. Two scope decisions narrowed it, both recorded on the issue:

| Decision | Where it went |
|---|---|
| Delivery semantics (assign-should-inject, honest delivery status) | abilityai/trinity-enterprise#385 |
| Assign/unassign **writes** from the Library | abilityai/trinity-enterprise#386 (filed with this PR) |

Everything else in the issue's AC list ships here: tabs, URL-addressable tab state with the `/templates` redirect and `#hash` anchors still resolving, creator-gated Systems tab, per-skill agent list, access-scoping, batched read (no N+1), design-system pass, both themes, per-agent Skills tab untouched.

**Gating: OSS-core by decision** — recorded in `architecture.md` per the ent#326 discipline, since CLAUDE.md's default for an enterprise-tracker feature is *gated unless ruled otherwise* and that ruling must never be inferred later from the mere fact that it merged.

## The finding that changed the design

The issue's Technical Notes say to scope the endpoint with `accessible_agent_names`. That helper resolves through `list_all_agents_fast()`, which returns `[]` whenever the Docker client is unavailable — surfaced only as a throttled WARNING. Routed through it, **changing `DOCKER_GID` would tell every non-admin that no agent holds any skill, fleet-wide**, with no error state anywhere: a confident wrong zero arriving through the *access* layer, where a store-level "a failed fetch is not an empty result" rule structurally cannot catch it.

The access set is therefore derived from `db.get_all_agent_metadata()` — one pure-DB batch query already carrying `owner_username`, `is_shared_with_user` and `deleted_at IS NULL`. Access semantics are identical (admin ⇒ all, else owned ∪ shared); only the source changes. There is a regression test for exactly this.

## Changes

**Backend**
- `db/skills.py::get_all_skill_assignments()` — SQLAlchemy Core (matching the module's 8/8 post-#300 shape). INNER JOINs `agent_ownership` on `deleted_at IS NULL` so a soft-deleted agent (rows preserved up to 180 days by #834) never renders as a current holder on the unfiltered admin path — the ent#335 class. Excludes ephemeral ghosts, whose chips would 404 within minutes of a discard; NULL `is_ephemeral` reads as not-a-ghost so pre-ent#69 rows survive.
- `GET /api/skills/assignments` — `reject_agent_principal` (an agent key resolves to its owner *carrying the owner's role*; there is deliberately no agent consumer, which makes the gate free) + a `response_model` allow-list, the ent#334 rule from this same router, so a future sensitive `agent_ownership` column is fail-closed. Carries `scope: all|accessible` so an empty result can be worded honestly.

**Frontend**
- `OverflowTabs` + `?tab=` state; legacy `#agent-templates`/`#systems`/`#skills` anchors migrate to their tab and are cleared. Tab clicks use `replace`, so switching costs no history entries; a `route.query.tab` watch makes the render follow the URL for deep links and history entries.
- Systems stays creator-gated with **both** arms of a late-role watch — `auth.js` reports `user` until `/api/users/me` lands, so a creator hard-loading `?tab=systems` would otherwise land on Templates and never recover.
- Panels **lazy-mount-once** (`v-if` visited + `v-show` active): plain `v-if` refetches three endpoints per switch (there is no poll to tear down — the #1109 rationale doesn't apply here), plain `v-show` charges a Templates-only visitor for Skills and Systems.
- Store splits `fetching` from `loading` so the ScanlineReveal beam means "no data yet" rather than replaying over rendered data; assignments carry their own generation counter so a per-card retry can't strand an in-flight `load()`.
- Could-not-load / nobody-holds-it / the-list stay three distinct states, and the zero wording follows `scope`, so a non-admin is never told a skill has no holders when they merely cannot see the ones it has.
- **Orphaned assignments** (skill gone from the library) get their own bounded list — ent#237's revocation model is "cut a new tag without the skill", after which *who still has it?* is the operator's first question and a grid keyed by the library listing answers it with silence.

## Test Plan

- [x] `pytest tests/unit/test_384_skill_assignments_read.py` — **10 passed**. Covers the admin/empty-set tri-state (a falsy check would hand an owner-of-nothing the whole fleet), agent principal → 403, `response_model` asserted on **key absence** (ent#334's rule), soft-deleted / orphan / ephemeral exclusion, route registration, and a **Docker-outage case that would otherwise read as "nobody holds anything"**.
- [x] Full unit suite: **9489 passed**. The one failure (`test_ent314_hardened_yaml`) is pre-existing and local-only — the guard walks a gitignored `src/backend/venv/` that CI never has, and flags `starlette`/`uvicorn` site-packages; this diff contains zero YAML parsing.
- [x] `npm run check:tokens` passes; raw-color counts **shrank** in both edited files (Library 97→91, LibrarySkillsSection 41→34) and the new component is at zero non-gray / zero hardcoded.
- [x] `vite build` clean.
- [x] **Verified against the live local stack**: endpoint returns real holders with display labels; tab strip renders; `?tab=skills` deep link and the full `/templates#skills` legacy chain both resolve; a history entry moves both URL and render; zero console errors. Fixed one defect found this way — margin-based spacing read as "Assigned to1 agent" to a screen reader.
- [x] e2e specs updated (tab strip, `?tab=skills` deep link, legacy anchor) — **green in CI**: the Playwright job ran 35 tests, 28 passed / 7 skipped / 0 failed. All 7 `smoke.spec.js` tests ran (its specs carry no skip conditions — the 7 skips are the `TEST_AGENT`/empty-`audit_log` guards in `circuit-breaker-badge`, `audit-dashboard`, `honest-failed-states` and `timeline-cancelled-bar`), so the three library specs are real coverage rather than a no-op pass.

Fixes abilityai/trinity-enterprise#384

---

## Post-review fixes (3rd commit)

An adversarial review of the shipped diff found nine defects; all are fixed in `6a7857ed`. The two that matter most were things the original tests asserted were fine:

- **`system-install.spec.js` was deterministically broken** by the tab change (it asserts `#systems`/`#skills` visible on a bare `/library`, and `/templates#systems` → `/library#systems`). CI runs `--grep @smoke` and that file has no `@smoke` tag, so it **never runs in CI** — this would have merged green and broken the next person to run the full suite.
- **`TestHolderPredicate` passed against inverted logic.** It matched `inspect.getsource` substrings, and `"agent_skills.join("` also matches a LEFT OUTER JOIN while `"is_ephemeral"` also matches `== 1`. Rewritten against a real migrated DB (`db_harness`, dual-backend); all three mutations now fail the suite.

Also fixed: the late-role restore arm was dead in its own scenario (`onMounted` rewrote the query before the role landed, so the watcher read its own normalization); URL and render could disagree after a bare `/library` navigation; a `hasLoaded` guard made the Skills tab stale for a whole SPA session (an admin configuring a library kept seeing "not configured"); the orphaned-assignments list was unreachable when the library is empty — the exact revocation case it was written for; plus five smaller items (inert test sentinel, an undocumented `scope='system'` exemption, an unimplemented "suppressed row" claim in the docstring and requirements, holder chips sorting by slug while rendered by label, and a retry control offering 50 live triggers for one request).


vybe (approved):
Validated per `/validate-pr`. 25/25 CI checks green (incl. all 4 required), 12/12 new unit tests pass locally, `check:tokens` OK and raw-color counts shrank in both edited files.

Verified independently:
- Route ordering (Invariant #4) — `/skills/assignments` has no single-segment `/skills/{x}` catch-all above it (and a test pins it).
- Auth: `reject_agent_principal` + the admin/empty-set tri-state (`visible is None` ≠ `set()`) — a falsy check there would hand an owner-of-nothing the whole fleet; there's a test for exactly that.
- The `accessible_agent_names` finding is real: it resolves through `list_all_agents_fast()`, which returns `[]` on any Docker fault behind a throttled WARNING. Deriving from `db.get_all_agent_metadata()` keeps identical access semantics from a pure-DB source, with a regression test.
- `response_model` allow-list asserted by key *absence* (the ent#334 rule).
- Tab logic converges: the late-role restore reads the captured `requestedTab` rather than the query `onMounted` already normalized, and the `q === undefined` re-stamp covers the bare `/library` re-navigation.
- Docs: architecture.md, requirements §22.4 (closing §22.3's named gap), both feature flows. OSS-core ruling recorded explicitly per the ent#326 discipline.

The self-review commit is the strongest signal here — it caught a spec that CI structurally cannot run (`system-install.spec.js` is untagged, so `--grep @smoke` skips it) and a test class that passed against inverted logic via `inspect.getsource` substring matching.

One follow-up, not a blocker: the close is cross-tracker, so ent#384 will not auto-promote — needs a manual `status-in-dev`.

PR #2134, opened 2026-08-12
fix(agents): apply the ?tab= deep-link before the mount awaits (#2130)

### LINKED ISSUE #2130: Agent Detail: ?tab= deep-link applies ~10s late on running agents and overrides the user's tab click
## Summary

On a **running** agent, an `?tab=` deep-link to Agent Detail lands on **Overview** for ~10 seconds, then yanks the user to the requested tab — **overriding any tab the user clicked in the meantime**. On a stopped agent the deep-link applies immediately and correctly.

This breaks exactly the flows the deep-link exists for: the Dashboard Timeline links and ExecutionDetail → "Continue as Chat" (EXEC-023, #1672).

## Root cause

`src/frontend/src/views/AgentDetail.vue` — `onMounted` calls `applyDeepLinkRouting()` **after** `await Promise.allSettled([...])`:

```js
await Promise.allSettled([
  loadSessionInfo(), loadApiKeySetting(), /* … */
  ...(agent.value?.status === 'running' ? [checkDashboardExists(), checkBrainOrbCapability()] : [])
])
startEmotionCycling()
startAllPolling()
applyDeepLinkRouting()   // <-- only now is ?tab= honoured
```

For a **running** agent that array additionally holds `checkDashboardExists()` and `checkBrainOrbCapability()`, which are container round-trips. `Promise.allSettled` waits for the slowest, so the landing tab is applied ~10s late. `onActivated` (the KeepAlive path) has the same ordering.

`applyDeepLinkRouting()` reads only `route.query` and writes `activeTab` / `chatMode` / `routeForcedMode` — it has **no dependency on any loaded agent data**, so there is no reason for it to run after the awaits.

## Reproduction (measured)

Same agent, same URLs, only the running state differs:

| State | `?tab=git` | `?tab=reports` | `?tab=schedules` | `?tab=credentials` | `?tab=skills` |
|---|---|---|---|---|---|
| stopped | Git | Reports | Schedules | Credentials | More (Skills panel) |
| **running** | **Overview** | **Overview** | **Overview** | **Overview** | **Overview** |

Active-tab trace on a running agent, `?tab=git`:

```
0.5s..9.5s: Overview   →   10s: Git
```

**User-input override** (the harmful symptom):

1. Open `/agents/<running-agent>?tab=git`
2. At ~2s click **Reports** — Reports opens correctly
3. At ~10s the page jumps to **Git** on its own

Observed: `clicked=Reports -> final=Git`. Without `?tab=` a user's click is never overridden.

## Expected

The `?tab=` landing is applied before the user can interact, and never overrides a tab the user selected afterwards.

## Adjacent gap (not fixed here)

`DEEP_LINK_TABS` omits `a2a`, `loops`, `playbooks`, `access`, `nevermined`, so those `?tab=` values silently fall back to Overview even on a stopped agent. Several are permission-gated panels (`v-if="… && agent.can_share"`), so adding them needs a non-owner blank-panel decision — tracked separately.

## Environment

dev @ `f14c1edf` (v0.9.0 freeze), enterprise submodule `2d64baa`, local Docker stack.

trinity-ability:
Closes #2130

## The bug

On a **running** agent an `?tab=` deep-link landed on **Overview** for ~10 seconds, then yanked the user to the requested tab — **overriding a tab they had clicked in the meantime**. On a stopped agent it applied immediately.

Found while UI-testing the v0.9.0 freeze.

## Root cause

`onMounted` called `applyDeepLinkRouting()` only *after* `await Promise.allSettled([...])`. For a running agent that batch also holds `checkDashboardExists()` and `checkBrainOrbCapability()` — container round-trips — and `allSettled` waits for the slowest, so the landing tab was applied whenever that finished. `onActivated` (the KeepAlive path, and the one ExecutionDetail → "Continue as Chat" actually takes) had the same ordering.

## Evidence

Same agent, same URLs, only the running state differing:

| State | `?tab=git` | `?tab=reports` | `?tab=schedules` | `?tab=credentials` |
|---|---|---|---|---|
| stopped | Git | Reports | Schedules | Credentials |
| **running** | **Overview** | **Overview** | **Overview** | **Overview** |

Active-tab trace, running, `?tab=git`:

```
0.5s..9.5s: Overview   →   10s: Git
```

Hijack: open `?tab=git`, click **Reports** at 2s → page jumps to **Git** at ~10s (`clicked=Reports -> final=Git`).

## The fix

Hoist `applyDeepLinkRouting()` above the awaits in both hooks.

It reads only `route.query` and writes local refs — no dependency on loaded agent data — and the whole tab area sits behind `v-if="agent"`, so writing `activeTab` early only decides which tab the **first** render shows. With nothing applied late, nothing can be overridden; the race is removed rather than patched.

The ent#358 `redirectRetiredSessionLink()` guard stays first in both hooks — it navigates away and early-returns, so an unmounting view still does no deep-link work.

## Verification (live, against the local stack)

After the fix, on a **running** agent:

- deep-linked tab is active at the **first 0.5s sample** (was 10s)
- a click at 2s **survives** (`clicked=Reports -> final=Reports`, only one state observed)

Regressions, all green:

- stopped agents: `?tab=` git / reports / schedules / credentials all still land correctly
- ent#358: `?tab=session` still redirects to `/workspace?agent=…`
- EXEC-023 (#1672): `?tab=chat&resumeSessionId=…` still lands on Chat, no JS errors
- KeepAlive re-entry (`onActivated`) honours a new `?tab=`
- no `?tab=` still defaults to Overview

`vitest` 136/136 · `vite build` green · raw-color ratchet exit 0 (no class changes).

## Test

No `@vue/test-utils` in this project, and the defect is lifecycle **ordering** rather than a pure function, so `tests/unit/agentDetailDeepLink.spec.js` is a source-structure guard — the shape already used by the Python AST guards (`test_ent109_git_env_seam.py`, `test_1891_python_version_parity.py`). It asserts the call precedes the first await in both hooks, stays behind the ent#358 guard, happens exactly once, and that `applyDeepLinkRouting` never starts reading `agent.value` — the property that makes the early call correct.

Proven non-vacuous: **2 failed** on the pre-fix source, **6 pass** after.

## Not in scope

`DEEP_LINK_TABS` omits `a2a`, `loops`, `playbooks`, `access`, `nevermined`, so those `?tab=` values silently fall back to Overview even on a stopped agent. Several are permission-gated panels (`v-if="… && agent.can_share"`), so adding them needs a non-owner blank-panel decision — noted in #2130, not bundled here.

### DIFF
--- a/src/frontend/src/views/AgentDetail.vue
+++ b/src/frontend/src/views/AgentDetail.vue
@@ -1265,6 +1265,17 @@ onMounted(async () => {
   // worth of requests on a view that is unmounting.
   if (redirectRetiredSessionLink()) return
 
+  // #2130: apply the ?tab=/resume landing BEFORE any await. It reads only
+  // route.query and writes local refs — nothing in it needs loaded agent data,
+  // and the whole tab area sits behind `v-if="agent"`, so writing activeTab
+  // early just makes the FIRST render land on the requested tab. Running it
+  // after the awaits below meant the deep-link waited on the slowest unrelated
+  // mount call: on a RUNNING agent that batch also holds the
+  // checkDashboardExists/checkBrainOrbCapability container round-trips (~10s
+  // measured), so the page showed Overview for that window and then stole a tab
+  // the user had clicked in the meantime.
+  applyDeepLinkRouting()
+
   // Load agent first (other calls may depend on agent data)
   await loadAgent()
 
@@ -1285,9 +1296,6 @@ onMounted(async () => {
   ])
   startEmotionCycling()
   startAllPolling()
-
-  // Handle tab / resume deep-link (from Timeline or ExecutionDetail navigation)
-  applyDeepLinkRouting()
 })
 
 // onActivated fires when component is shown (after being cached by KeepAlive)
@@ -1295,6 +1303,14 @@ onActivated(async () => {
   // ent#358: same guard as onMounted — AgentDetail is KeepAlive-cached, so a
   // session deep-link opened on an already-visited agent lands here instead.
   if (redirectRetiredSessionLink()) return
+
+  // EXEC-023 (#1672): re-apply the full tab + resume landing here, not just the
+  // tab — onMounted does NOT fire for a KeepAlive-cached instance, so this is
+  // the path the common "Continue as Chat" navigation actually takes.
+  // #2130: and apply it BEFORE the awaits below, for the same reason as
+  // onMounted — this hook re-awaits loadAgent() plus, on a running agent, the
+  // two container round-trips.
+  applyDeepLinkRouting()
 
   // Restart polling when returning to this view
   startAllPolling()
@@ -1308,11 +1324,6 @@ onActivated(async () => {
     await checkDashboardExists()
     await checkBrainOrbCapability()
   }
-
-  // EXEC-023 (#1672): re-apply the full tab + resume landing here, not just the
-  // tab. onMounted does NOT fire for a KeepAlive-cached instance, so this is the
-  // path the common "Continue as Chat" navigation actually takes.
-  applyDeepLinkRouting()
 
   // DEPRECATED: Terminal tab hidden (candidate for removal)
   // if (activeTab.value === 'terminal') {
--- a/src/frontend/tests/unit/agentDetailDeepLink.spec.js
+++ b/src/frontend/tests/unit/agentDetailDeepLink.spec.js
@@ -1,0 +1,103 @@
+/**
+ * #2130 — the ?tab= deep-link landing must be applied BEFORE any await in the
+ * lifecycle hooks that own it.
+ *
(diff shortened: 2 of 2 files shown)


vybe (approved):
Validated via /validate-pr.

- Closing keyword `Closes #2130` present (same-repo → auto-promotes to `status-in-dev`).
- Correctness verified at the source: `applyDeepLinkRouting()` reads only `route.query.tab` and writes `activeTab`, and the only `activeTab` reset lives in the `route.params.name` watcher — a path that never applied the deep-link and is untouched here. So hoisting above the awaits cannot be clobbered.
- Regression guard proven non-vacuous locally: **2 failed** against `HEAD~1` source, **6 pass** after. Full frontend suite 136/136; raw-color ratchet exit 0.
- Security scan clean (no secrets/emails/IPs/env/cred files); no backend module or `os.getenv` packaging surface touched.

Approving.

PR #2135, opened 2026-08-12
fix(tests): guard the directory the #1804 guard wasn't walking (#2131, #2132)

### LINKED ISSUE #2131: bug(tests): #1804 terminal-activity parity guard doesn't scan client_portal/, which now holds a terminal writer
Found while validating #2120.

## The gap

`tests/unit/test_1804_terminal_activity_parity.py` enforces the #1804 contract — *every writer that wins a terminal CAS on `schedule_executions` closes the paired `agent_activities` dispatch row*. It walks the tree with:

```python
_SCAN_DIRS = ("services", "routers")
```

`client_portal/` is neither. That was harmless when the module was entitled and lived in the submodule, but ent#356 moved it into OSS core — and #2120 then added a **terminal writer** to it:

```python
# src/backend/client_portal/service.py
def _fail_unstarted_execution(execution_id: str, reason: str) -> None:
    core_db.update_execution_status(execution_id, "failed", error=...)
```

`update_execution_status` is in the guard's own `TERMINAL_WRITERS` set. The guard simply never looks at the file.

## This is not currently a bug

`_fail_unstarted_execution` only fires when the turn never reached `execute_task`, so no dispatch activity exists to close — the same justification the guard already allowlists for `routers/chat.py::_raise_ephemeral_exhausted_410`. Behaviour today is correct.

The defect is that the guard's **coverage assumption has drifted**: `client_portal/` now holds execution-terminal logic, and the next writer added there gets no protection. That is precisely the recurrence the guard's own docstring argues against —

> Companion to `#429`: [...] A per-site patch would be deleted with its host, leaving no owner. This guard survives, because it is anchored on terminal writes.

A guard that survives its hosts but not its directories has the same hole one level up.

## Fix

Add `client_portal` to `_SCAN_DIRS` and give `_fail_unstarted_execution` an explicit `ALLOWLIST` entry with the lifecycle justification (the guard already asserts every entry carries one ≥60 chars and that the named function still exists).

Consider whether the scan should instead be "every package under `src/backend` that isn't test/migration scaffolding", so adopting a future module can't silently repeat this.

## Acceptance criteria

- [ ] `client_portal/` is scanned
- [ ] `_fail_unstarted_execution` allowlisted with a justification proving no dispatch activity exists at that point
- [ ] The guard's four existing self-tests still pass, including `test_allowlist_entries_all_still_exist`
- [ ] Removing the allowlist entry makes the guard fail (i.e. it genuinely covers the new directory)

Related: #1804, #2120

### LINKED ISSUE #2132: refactor(frontend): delete orphaned views/enterprise/Sessions.vue left by the Sessions-page retirement
Found while validating #2120.

ent#381 retired the standalone Sessions page: `NavBar.vue` lost its entry and the `/sessions/:roomId?` route became a redirect into the Workspace. The component it used to render was left behind:

```
$ git show origin/dev:src/frontend/src/views/enterprise/Sessions.vue | wc -l
     ...   # still present
$ git grep -n "enterprise/Sessions" -- src/frontend/src src/frontend/e2e
     # no importers
```

Nothing imports it. Its only surviving reference is a stale entry in the raw-color ratchet:

```
src/frontend/raw-color-baseline.json:711:    "src/frontend/src/views/enterprise/Sessions.vue": {
```

Dead Vue is mildly costly in three ways: it is a plausible-looking file someone can edit for hours before noticing nothing renders it; it keeps a raw-color budget alive for a file that can never improve, which is exactly the kind of entry the ratchet is meant to shrink; and it makes `git grep` for room UI return a surface that no longer exists.

## Fix

Delete `src/frontend/src/views/enterprise/Sessions.vue` and its `raw-color-baseline.json` entry.

Check first whether any sibling under `views/enterprise/` became unreferenced in the same change.

## Acceptance criteria

- [ ] File deleted, baseline entry removed
- [ ] `node src/frontend/scripts/scan-raw-colors.mjs src/frontend` passes (per-file counts may only shrink — a removed file is the limit case)
- [ ] `npx vitest run` and the frontend build stay green
- [ ] No remaining reference in `src/`, `e2e/`, or docs

Related: #2120, abilityai/trinity-enterprise#381

trinity-ability:
Two follow-ups found while validating #2120, both small enough to land together.

## #2131 — the guard had a blind spot shaped like a directory

`tests/unit/test_1804_terminal_activity_parity.py` enforces the #1804 contract — *every writer that wins a terminal CAS on `schedule_executions` closes the paired `agent_activities` dispatch row*. It walked:

```python
_SCAN_DIRS = ("services", "routers")
```

ent#356 moved `client_portal/` into OSS core; ent#286 then added a terminal writer to it:

```python
# src/backend/client_portal/service.py
def _fail_unstarted_execution(execution_id, reason):
    core_db.update_execution_status(execution_id, "failed", error=...)
```

`update_execution_status` is in the guard's own `TERMINAL_WRITERS`. The guard simply never opened the file, and reported green over it for all of #2120.

**Nothing is broken today.** That helper only runs when the background turn raised *before* `execute_task` saw the id (a resume lock held by another tab, a thread-resolution failure), so no dispatch activity exists to close — the same lifecycle position as the admission-path entries already allowlisted. If `execute_task` did run, its own terminal wins the CAS and this write is a no-op. The defect is the **coverage assumption**, not the behaviour.

### Why this isn't a one-line tuple edit

#1804 deliberately anchored on terminal *writes* rather than on `spawn_task_terminal_event` emission, because that survives a writer moving between functions and files. It does not survive one appearing in a package nobody walks — the same failure one level up.

So alongside adding `client_portal`, this adds `test_scan_covers_every_package_that_writes_a_terminal`: an AST scan of the whole backend that fails when any location calling a terminal writer is neither in `_SCAN_DIRS` nor in a new `_UNSCANNED_JUSTIFIED` map. Today that map holds exactly one entry — `database.py`, the facade that *delegates* the write (Invariant #2); scanning it would flag the plumbing every legitimate closer runs through.

The submodule is excluded on purpose: it is optional (`update = none`, #1443), so reaching into it would make the guard **stronger on core-team clones than on OSS ones** — which is not a guard.

### Mutation-checked, not assumed green

Both directions were verified to actually fail:

| Mutation | Expected | Result |
|---|---|---|
| Remove the new allowlist entry | original guard fails → the directory really is walked | ✅ `client_portal/service.py::_fail_unstarted_execution (line 1158)` |
| Revert `_SCAN_DIRS` to `("services", "routers")` | new coverage test fails → it would have caught this originally | ✅ `client_portal: ['client_portal/service.py']` |

## #2132 — delete the view ent#381 orphaned

`views/enterprise/Sessions.vue` lost its route and NavBar entry in #2120. Nothing imports it; its only surviving reference was a `raw-color-baseline.json` entry keeping a ratchet budget alive for a file that can never improve. `Audit.vue` and `Index.vue` are still routed (`/enterprise`, `/audit`) and are untouched.

### One honest note on the baseline

The `raw_gray` total is reduced by exactly the 23 removed (7446 → 7423). It still does not reconcile with the sum of its per-file entries — **that drift is pre-existing**, `+21` on dev before this commit:

```
PRE-EXISTING on dev  raw_gray  sum=7425  totals=7446  DRIFT +21
```

Deliberately left alone. Correcting it here would conflate two changes and hide whatever caused it; enforcement is per-file anyway (per CLAUDE.md the ratchet is "per-file counts may only shrink"), and the totals block is informational. Worth a separate regeneration pass.

## What this PR deliberately does NOT do

Deleting `Sessions.vue` leaves five `components/rooms/*` files (512 lines) with zero importers — it was their only consumer — and `stores/rooms.js` is still driven by `utils/websocket.js` while no view reads it. The new `PortalRoom.vue` (ent#361) reuses none of them and polls every 3s instead.

That is a **design decision, not cleanup**: delete the old set, or consolidate `PortalRoom.vue` onto it. Deleting here would quietly foreclose the second option, and it is outside #2132's scope. Filed as #2136 with both halves.

Net effect of this PR is still strictly less dead code, and the remaining orphans are now *visible* (0 importers) rather than hidden behind a dead view.

## Verification

```
796 passed, 1 skipped     # backend: 1804 parity, portal, session, models-centralized guards
130 passed                # frontend vitest
exit 0                    # scan-raw-colors.mjs, deleted file absent from output
✓ built in 1.62s          # vite production build
```

Fixes #2131
Fixes #2132


vybe (approved):
Reviewed the diff independently rather than against the commit message. Two things came out of it, one fixed in-place and one filed.

**Fixed in `c52f5f9`:** `_COVERAGE_EXCLUDED_PREFIXES` also excluded `migrations/`, while the comment justified only the submodule. Alembic revisions write raw `op.execute("UPDATE ...")` rather than calling the db facade, so they can never match `TERMINAL_WRITERS` — the exclusion bought nothing and would have silently swallowed the surprising case if one ever appeared. Narrowed to `("enterprise/",)` with the reasoning recorded, since a guard whose stated scope doesn't match its real scope is the exact defect this PR exists to fix.

**Filed as #2136:** deleting `Sessions.vue` orphans five `components/rooms/*` files (512 lines, zero importers), and `stores/rooms.js` is still driven by `utils/websocket.js` with no view reading it while `PortalRoom.vue` polls every 3s for the same data. Deliberately NOT actioned here — "delete the old set" vs "consolidate `PortalRoom.vue` onto it" is a design call, and deleting would foreclose the second option.

**On the allowlist entry.** The claim that `_fail_unstarted_execution` has no activity to close checks out against the lifecycle: `start_portal_turn` creates the row, then `_run` calls `portal_chat` → `execute_task`, and the dispatch activity is opened at step 3. Every path that reaches this helper raised before that point; if `execute_task` did run, its terminal wins the CAS first and this write is a no-op. Same shape as the six admission-path entries already present.

**On the mutation checks.** These are the part that makes the change trustworthy — both directions were shown to fail, so neither the new directory nor the new coverage test is decorative:

- removing the allowlist entry → `test_every_terminal_writer_closes_its_activity` fails naming `client_portal/service.py::_fail_unstarted_execution`
- reverting `_SCAN_DIRS` → `test_scan_covers_every_package_that_writes_a_terminal` fails naming `client_portal`

**On the baseline totals.** Subtracting only the 23 removed and leaving the pre-existing `+21` drift is the right call — correcting it here would conflate two changes and hide its cause. Verified the drift predates this commit (`sum=7425 totals=7446` on dev).

Verified locally: 796 backend guard/portal/session tests, 130 frontend vitest, `scan-raw-colors` exit 0 with the deleted file absent, vite build green.

**APPROVE.**

PR #2138, opened 2026-08-12
fix(workspace): repair regressions the #2120 review fixes introduced (ent#358, ent#286)

dolho:
Follow-up to #2120 (merged). A re-review after its fix commit found that **the fixes themselves regressed three user-visible paths**, and #2120 has since merged — so these are live on `dev` right now.

## The urgent one

`markFailed` is **called at two sites and never defined.** A block-level edit removed it along with the code around it.

`vite build` compiles that happily — it is a runtime `ReferenceError` — and this project has **no ESLint**, so the full suite and CI went green while **every failed Workspace send throws and renders no error and no Retry button.** That is the same user-visible bug reported three times now: once originally, once after a fix that set state the view never re-read, and now this.

So it gets a guard rather than a fourth fix: `portalUndefinedCalls.spec.js` fails on a call to an identifier the SFC neither defines nor imports.

## The other four

| | |
|---|---|
| `reattach()` never rendered its reply | called `awaitPersistedReply` without the new `baselineAssistants` arg, so `assistants.length > undefined` was false on every poll. A user who reloaded mid-turn watched the indicator vanish and had to reload *again* to see an answer that had already arrived. |
| The access-denied latch never cleared | `activeAgent` returns null while `unreachableAgent` is set, and only mount-time code cleared it — one stale deep link wedged the **entire** Workspace on the access-denied panel for the rest of the session. |
| `clear_turn_inflight` deleted unconditionally | a second turn overwrites the session marker, so a finishing (or fast-failing) turn wiped the marker of one still running; the watching client then saw "nothing in flight", gave up, and offered a Retry that **dispatches and bills the turn again** — the exact double-spend the marker exists to prevent. Now compare-and-delete. |
| The marker outlived its turn by an hour | TTL 3600s against a 300s turn, and the reply wait is now gated on that marker — a backend restart mid-turn would leave the composer disabled and spinning for up to an hour. TTL is now the turn's bound plus slack; the client also carries an absolute ceiling. |

## Verification

102 frontend tests (9 files), build clean. The new guard fails on the pre-fix tree, which is the point.

## Worth saying plainly

Every one of these came from the commit that fixed the previous review's findings. Two rounds of review have now found that my fixes introduced defects at roughly the rate they removed them, and the only reason this round was caught is that the review was re-run. A third pass against this branch before merge is warranted — I am not a reliable judge of code I wrote in the same session.

Related to abilityai/trinity-enterprise#358, abilityai/trinity-enterprise#286

### DIFF
--- a/src/backend/client_portal/service.py
+++ b/src/backend/client_portal/service.py
@@ -1058,7 +1058,15 @@ def _inflight_exec_key(execution_id: str) -> str:
     return f"portal_inflight_exec:{execution_id}"
 
 
-def mark_turn_inflight(session_id: str, execution_id: str, ttl_seconds: int = 3600) -> None:
+# The marker describes a turn in progress, so its TTL is the turn's own bound
+# plus slack — not an hour. The client now waits as long as the marker says a
+# turn is running, so an over-long TTL is a spinner that outlives the work: a
+# backend restart mid-turn would leave the composer disabled for the remainder.
+PORTAL_INFLIGHT_TTL_SECONDS = 300 + 60
+
+
+def mark_turn_inflight(session_id: str, execution_id: str,
+                       ttl_seconds: int = PORTAL_INFLIGHT_TTL_SECONDS) -> None:
     """Record that ``session_id`` has a turn running, and WHICH one.
 
     The value is the execution id, not a bare flag: a client that reloads has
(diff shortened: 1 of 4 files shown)


vybe (approved):
The compare-and-delete in `clear_turn_inflight` is the load-bearing change: an unconditional delete let a fast-failed turn wipe a running turn's marker, and the client's Retry then double-billed. The reverse-key resolution on the `execution_id is None` path is right too. `markFailed` restored with a guard that fails on the class rather than the instance. TTL 3600 → 360 matches the turn bound with slack, and the client's 6-min ceiling sits above it. Approving.

PR #2139, opened 2026-08-12
feat(workspace): sidebar IA — agents block, starred chats, unread badges (ent#359)

dolho:
Closes the sidebar half of the Workspace family. Issue: [ent#359](https://github.com/Abilityai/trinity-enterprise/issues/359).

The sidebar stacked New chat, Search, an Agents roster and date-grouped history in one column, all at the same visual weight. That was coherent while an agent was *a menu entry you pick to start a chat*. Once the Workspace absorbed the Session surface (ent#358) and became the only place a continuing conversation lives, an agent became a **destination** and a chat became the record of visiting one — two kinds of thing, one treatment.

The roster now sits at the top of the scroll region on its own surface; chats follow, with starred ones pinned above the date groups.

## The decision worth reviewing

Both new features — stars and unread — are **per-viewer**. That one word rules out the obvious implementation.

A star is one person's bookmark. A room has several participants. A `starred` column on the chat row would render one participant's star in every other participant's sidebar, and incidentally tell them which conversations their colleagues care about. Rooms also live in the private submodule while threads live in OSS, so a per-kind column would put half of one feature in each repo.

Hence `enterprise_portal_chat_state (client_email, chat_kind, chat_id) → starred_at, last_read_at`. The caller's email is the primary-key prefix, so **the row is the tenant scope** — there is no filter to forget on a read path and no way to address another viewer's state. `chat_kind` is required because thread ids and room ids are independent spaces.

## Two judgement calls I'd rather flag than bury

**1. "Waiting on the user" is defined as unread agent replies.** The AC names a badge but the codebase has no such concept, so I picked one. The alternative reading — an agent blocked on an operator-queue approval — I deliberately rejected: that queue belongs to operators, and a Workspace viewer may be an external client with no standing in it. If the operator-queue sense was intended, that is a different and larger change and I'd redo it rather than leave the label ambiguous.

**2. A thread with no read cursor reports NOTHING unread**, not its whole history. Treating "never read" as "all unread" would have badged every historical conversation in every install the day this shipped — noise that teaches people to ignore the badge, which is worse than not having one. A cursor is written the first time the viewer opens or sends in a thread, so live conversations acquire one immediately and the first reply they *don't* see is the first thing that badges. The cost is stated in Known Limitations.

## AC coverage

| AC | |
|---|---|
| Agents block first, visually distinct | own surface (card + ring) above the chat list |
| avatar, name, description, count badge | ✓ |
| Starred pinned above date groups, lifted out | `partitionStarred` runs *before* grouping — a starred chat appears exactly once, not once pinned and once in "Today" |
| Star/unstar from header and row | `PortalStarButton.vue`, one component for its three homes |
| Multi-agent chats show all participants | stacked avatars, capped at 3 + a "+N" chip |
| Search replaces the lists while active | unchanged |
| Empty roster keeps a next-action | create-an-agent (platform) / ask-your-inviter (client) |

Plus the aggregate count on the wordmark, since the agents block now occupies the top of a *scrolling* region.

Clicking an agent that is waiting on you opens the conversation it is waiting in. A badge reading "2 replies" next to a control that opens an *empty* chat is a contradiction — the count is the reason you clicked. (An agent's own page is ent#360; this is not a substitute for it.)

## Security

No roster gate and no existence check on `chat_id`, both deliberate: every row is keyed by the caller's own email, and a 404 for an unknown chat would be an enumeration oracle over every chat id in the install (invariant #8). A per-viewer row cap bounds the write instead — applied **only** to writes that create a row, so a user at the ceiling can still unstar, which is the only way back under it.

## Verification

- **9690 passed**, 16 skipped, 1 xfailed. The 3 failures are `test_pat_propagation_properties.py::TestKnownGaps` — pre-existing on `dev`, tracked as #2017; neither that test nor `github_pat_propagation_service.py` is touched by this branch (0-line diff).
- 18 new backend tests; 116 frontend tests (10 files); `vite build` clean.
- SQLite migration verified **in isolation** against a DB that predates the table: right columns, composite PK, idempotent on re-run. (Checking only that the table exists after boot proves nothing — `schema.py`'s `CREATE TABLE IF NOT EXISTS` would satisfy that alone.)
- Route registration + table creation confirmed on a real app boot.

## Stacking

Branched off `fix/2120-review-regressions` (#2138), so this diff includes those five fixes until that merges. **#2138 should go first** — `dev` currently has a `markFailed` ReferenceError on every failed Workspace send.

## Known limitations

| | |
|---|---|
| Rooms report `unread: 0` | a room keeps its own seq cursor, a different model from a timestamp cursor; stars work for both kinds, unread does not |
| Unread needs one open first | by design, see above |
| Optimistic star, no cross-tab sync | a star toggled in one tab appears in another on its next refresh |

Docs: requirements §5.10, flow `docs/memory/feature-flows/workspace-sidebar-ia.md`.

Related to abilityai/trinity-enterprise#359

### DIFF
--- a/docs/memory/feature-flows.md
+++ b/docs/memory/feature-flows.md
@@ -339,6 +339,7 @@
 | Persistent Chat Tracking | [persistent-chat-tracking.md](feature-flows/persistent-chat-tracking.md) | Database-backed chat persistence |
 | Session Tab | [session-tab.md](feature-flows/session-tab.md) | The `--resume` engine — each turn reattaches to the same Claude memory (SESSION_TAB_2026-04). Its Agent Detail surface retired in ent#358; the engine and endpoints are live |
 | Workspace absorbs Session | [workspace-absorbs-session.md](feature-flows/workspace-absorbs-session.md) | ent#358 — one continuous-conversation surface, with the continuity parity that had to land before the other could be removed |
+| Workspace sidebar IA | [workspace-sidebar-ia.md](feature-flows/workspace-sidebar-ia.md) | ent#359 — agents block on top, starred chats, per-agent unread badges; why the per-viewer state could not be a column on the chat row |
 | Web Terminal | [web-terminal.md](feature-flows/web-terminal.md) | Browser-based terminal for System Agent |
 | Self-Execute | [self-execute.md](feature-flows/self-execute.md) | Agent background task during chat (SELF-EXEC-001) |
 
--- a/docs/memory/feature-flows/workspace-sidebar-ia.md
+++ b/docs/memory/feature-flows/workspace-sidebar-ia.md
@@ -1,0 +1,147 @@
+# Feature: Workspace sidebar IA — agents block, starred chats, unread badges
+
+> **Status**: ✅ Implemented (2026-08-12)
+> **Issue**: abilityai/trinity-enterprise#359
+> **Requirement**: `docs/memory/requirements/core-agent.md` §5.10
+> **Related**: [workspace-absorbs-session.md](workspace-absorbs-session.md) (why the roster's role changed)
+
+## Overview
+
+The sidebar had four things stacked in one column — New chat, Search, an Agents
+roster, date-grouped history — all rendered with the same weight. That was
+coherent while an agent was a *menu entry you pick to start a chat*. Once the
+Workspace absorbed the Session surface (ent#358) and became the only place a
+continuing conversation lives, an agent became a **destination** and a chat
+became the record of visiting one. Two kinds of thing, one visual treatment.
+
+This change separates them, and adds the two pieces of state that make a list of
+conversations navigable rather than merely complete: **which ones you keep coming
+back to**, and **which ones are waiting on you**.
+
+## The thing that decided the design
+
+Both new features are *per-viewer*. That single word rules out the obvious
+implementation.
+
+A star is one person's bookmark. A room (`shared_sessions`) has several
+participants. A `starred` column on the chat row would therefore render one
+participant's star in every other participant's sidebar — and incidentally tell
+them which conversations their colleagues care about. Rooms also live in the
+private enterprise submodule while threads live in OSS, so a per-kind column
+would put half of one feature in each repo.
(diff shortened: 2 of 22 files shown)


dolho:
Rebased onto `dev`.

A plain `git rebase` did the wrong thing and is worth recording: this branch stacks on #2138, which stacks on #2120 — and #2120 **squash-merged**, so `dev` holds its content under a new sha while the branch still held the originals. Replaying them conflicted in five files against their own merged copies. Rebuilt both branches by cherry-picking only the commits `dev` does not already have (2 for #2138, 1 here), so the fix stack is now genuinely on top of dev rather than re-applying itself.

Also pushed one fix that came out of reviewing the rebase: the row star was revealed by `group-hover`, and there is no hover on touch — a starred row always draws its star and could be unstarred, but an **unstarred row could never be starred at all on mobile**. The reveal is now `sm:`-and-up only.

Verified after rebase: 890 passed / 1 skipped across the portal, session, schema-parity, migration, models-centralisation and enumeration-uniformity suites; 159 frontend tests (14 files); build clean.

Still stacked on #2138 — that one should merge first, and until it does this diff carries its 2 commits (22 files rather than 21).

vybe (approved):
Per-viewer state keyed on the caller's own lower-cased email, so the row *is* the tenant scope — no filter that can be forgotten. The two deliberate non-obvious calls both look right to me: no existence check on star/read (a 404 would make `star` an id-enumeration oracle, invariant #8), with the row cap as the bound on junk-id writes instead; and a cursor-less thread reporting nothing rather than its whole history as unread.

Dual-track migration is complete — `db/schema.py` + `db/migrations.py` + `db/tables.py` + Alembic 0038, and `alembic-single-head` is green.

I merged `dev` into the branch myself (trivial `openRoom` conflict — #2138 squash-merged while this branch still held its originals; both sides' lines kept) and re-pinned `src/backend/enterprise` from the #389 branch head to its merge commit a0ad11c5, which is what actually exists on `enterprise-main`.

Issue #2140, opened 2026-08-12
bug: test_inject_assigned_credentials leaks a stubbed services.agent_service.helpers into sys.modules, breaking runtime filtering in later tests


vybe:
## Summary

`tests/unit/test_inject_assigned_credentials.py` installs stub modules directly into `sys.modules` (`services.agent_service`, `services.agent_service.helpers`, `.read_only`, `.file_sharing`, `services.credential_encryption`) at import time and **never restores them**. The stubs outlive the module.

`services/compatibility/spec.py::applies_to_runtime()` resolves its dependency lazily:

```python
def applies_to_runtime(check, runtime):
    if not check.claude_only:
        return True
    from services.agent_service.helpers import is_claude_runtime   # <- picks up the stub
    return is_claude_runtime(runtime)
```

so any test that runs *after* `test_inject_assigned_credentials` gets the stubbed `is_claude_runtime` and loses `claude_only` filtering.

## Reproduction

```bash
python3 -m pytest tests/unit/test_inject_assigned_credentials.py \
                  tests/unit/test_compatibility_checks.py \
                  -q -p no:randomly -k "codex_runtime_omits or inject"
```

```
E  AssertionError: assert 'C-001' not in {'A-001', ..., 'C-001', ...}
FAILED tests/unit/test_compatibility_checks.py::TestBuildReport::test_codex_runtime_omits_claude_only_checks
```

Each file passes on its own. Confirmed **pre-existing on `dev`** (reproduced in a clean `dev` worktree), and confirmed still present on the `#2137` branch — it is not caused by either.

## Why it matters

The repo runs `pytest-randomly`, so file order varies per run. This is therefore a **latent CI flake**, not a deterministic failure: it only fires when the polluting module happens to sort before `test_compatibility_checks.py`. The test it breaks (`test_codex_runtime_omits_claude_only_checks`) guards the #1187 multi-runtime property that Codex/Gemini agents are not flagged with false HARD `CLAUDE.md` findings — so the flake masquerades as a real compatibility regression.

## Suggested fix

Either restore the pre-existing entries in a fixture teardown (`monkeypatch.setitem(sys.modules, ...)` gets this for free), or scope the stubbing to the tests that need it rather than doing it at module import.

## Acceptance Criteria

- [ ] `sys.modules` is restored after `test_inject_assigned_credentials` completes
- [ ] The reproduction command above passes
[...]

PR #2141, opened 2026-08-12
refactor(compat): align the agent-compatibility catalog with what the platform implements (#2137)

### LINKED ISSUE #2137: refactor: align the agent-compatibility catalog with what the platform implements and the marketplace generates
## Summary

The agent-compatibility catalog (`src/backend/services/compatibility/spec.py`, 101 checks) and its human twin (`docs/agent-validation-spec.md`) have drifted from what the platform actually implements and from what the `create-agent` wizards in the public `abilityai/abilities` marketplace actually generate. A full review — all 101 checks against `static_checks.py`, `collector.py`, and the live backend — found three classes of problem:

1. **Checks that validate fields Trinity does not implement.** `git.commit_paths` / `git.ignore_paths` / `git.push_enabled` appear *only* inside `services/compatibility/` and two docs — there is no backend reader anywhere. That is **T-017 (HARD)**, **G-003**, **G-004**, **G-005** gating on a fiction. Same for a `template.yaml metrics:` key (**D-006**) and for `.trinity/post-check` (**I-005**), whose sole reason for existing in `git_service._TRINITY_AUTHORED_PATHS` is a comment pointing back at I-005. No executor exists for it.
2. **Checks that fail every agent we ship, by construction.** The wizards emit `.claude/skills/<name>/SKILL.md` and never `.claude/commands/`, so **F-008** is a guaranteed INFO failure. Schedule messages are written as `"Run /some-skill — …"`, but `_slash_command()` uses `re.match(r"^\s*/…")`, so **A-001** fails on every agent while **P-006 (HARD, approval gates) never fires at all**. `_command_names()` globs only `.claude/commands/`, so **X-007** and **T-016** cannot see a skill-based agent. **G-002** compares against the 45-entry fleet-wide `_GITIGNORE_PATTERNS` that Trinity itself injects at git-init.
3. **Duplicates and self-contradictions.** `c_t016` and `c_x007` are the same logic under two ids; `c_k002` is literally `return c_t015(snap)`; **K-005** is an AI restatement of static **S-010**; **T-012** is a one-directional **X-004**; **I-001..I-004** ask "is the output contract documented?" four ways. **P-009** tells authors to split reference material out of `SKILL.md`, and **P-004** then flags the resulting `reference.md` — because `_skill_files()` walks every `.md` under `.claude/skills/`.

Separately, the anti-drift test has a blind spot: `TestSpecDocSync::test_ids_match_doc` matches `^\|\s*([A-Z]-\d{3})\s*\|` — a **single** letter. The doc's `DP-001`…`DP-005` (Runtime Data Paths) never matched, so five checks are documented, indexed in the Check Index, and entirely unimplemented. The guarantee "the two can't drift" does not currently hold.

## Context

The compatibility report is a user-facing surface — `GET /api/agents/{name}/compatibility`, the `get_agent_compatibility_report` MCP tool, and `CompatibilityPanel.vue` on the agent Overview. An agent scaffolded by the marketplace wizards today lands with roughly nine SOFT/INFO failures that are not the author's fault and cannot be acted on, which trains operators to ignore the panel. Meanwhile the one HARD check that guards autonomous runs from hanging on an approval gate (P-006) is silently inert.

Field-consumption note relevant to severity: `use_cases` and `capabilities` are surfaced only by `_build_local_template()` (the bundled-gallery path) in `services/template_service.py`. The GitHub-sourced builder does not surface them, and repo-first deployment is now the default path — so **T-010** and **T-011** are SOFT-failing agents for fields the common path never reads.

## Acceptance Criteria

- [ ] Catalog reduced from 101 to ~86 checks; `test_catalog_size` updated to the new pinned number
- [ ] Dead-field checks removed: `T-017`, `G-003`, `G-004`, `G-005` (`git:` block), `D-006` (`metrics:`), `I-005` (`post-check`) — or, if the `git:` block is intended to ship, a linked issue implements it and the checks stay
- [ ] Legacy/no-longer-generated checks removed: `F-008` (`.claude/commands/` required), `F-012` (`docs/memory/requirements.md`), `F-013` (`CHANGELOG.md`)
- [ ] Duplicates removed: `T-012` (→ `X-004`), `T-016` (→ `X-007`), `K-002` (→ `T-015`), `K-005` (→ `S-010`), `I-003` + `I-004` (→ `I-001`); `G-002` dropped or narrowed to agent-authored patterns only
- [ ] `C-009` merged into `C-006`; `C-012` (persona coherence) dropped or reduced to INFO
- [ ] Downgraded SOFT → INFO: `T-007` (`version`), `T-008` (`author`), `T-010` (`use_cases`), `T-011` (`capabilities`), `K-003` (`.env.example` comments)
- [ ] `_slash_command()` matches a slash command anywhere in the message, not only at position 0 — `P-006` and `X-007` demonstrably fire on a `"Run /skill — …"` schedule in a unit test
- [ ] `_command_names()` also resolves `.claude/skills/<name>/SKILL.md`, so `X-007` passes on a skill-based agent and fails on a genuinely missing target
- [ ] `A-001` accepts a slash command anywhere in the message
- [ ] `P-004` scopes to `SKILL.md` only (never companion `reference.md`/`examples.md`) and its threshold no longer contradicts `P-009`
- [ ] `P-006` honors an explicit opt-out for skills that gate by design (the marketplace already carries an `automation: autonomous | gated | manual` frontmatter convention) — a deliberately-gated scheduled skill is not reported as a HARD failure
- [ ] `T-013` explicitly PASSes when `use_cases` is absent, mirroring `T-014`
- [ ] `F-004` (`.env.example`) is conditional on the agent declaring credentials — `K-001` already covers the case that matters at HARD
- [ ] `test_ids_match_doc` regex widened to `[A-Z]{1,2}-\d{3}` so a two-letter prefix can never again drift silently
- [ ] `DP-001`…`DP-004` either implemented in `spec.py` + `static_checks.py`, or removed from the doc's Check Index; `DP-005` removed (the collector does not read `.trinity/pre-snapshot`)
- [ ] `docs/agent-validation-spec.md` regenerated to match: Check Index, per-check definitions, severity summary, auto-fixable list, and the "Implementation deviations" note all consistent with `spec.py`
- [ ] An agent scaffolded fresh by a `create-agent` wizard reports zero HARD and zero unactionable SOFT findings — added as a fixture-based regression test alongside the existing good-agent test

## Technical Notes

Files in scope:
- `src/backend/services/compatibility/spec.py` — the catalog
- `src/backend/services/compatibility/static_checks.py` — `_slash_command`, `_command_names`, `_skill_files`, `c_p004`, `c_p006`, `c_g002`
- `src/backend/services/compatibility/collector.py` — snapshot paths (no change needed unless `DP-005` is kept)
- `docs/agent-validation-spec.md` — the human source of truth
- `tests/unit/test_compatibility_checks.py` — `TestSpecDocSync`, `test_catalog_size`

Check IDs are documented as stable and never renumbered; removals should leave the remaining ids untouched and the doc should note the retired ids so old persisted rows in `agent_compatibility_results` are interpretable.

Two of the findings are arguably `type-bug` rather than cleanup and could be split out if that helps sequencing: **P-006 never fires** (a HARD guard that has been inert since it shipped) and **X-007 is blind to `.claude/skills/`** (false failures on the only layout the wizards produce).

Marketplace-parity gap, optional and not required by the criteria above: the platform catalog has no equivalent of the marketplace's own audit rubric item that cross-references CLAUDE.md's capability/dispatch tables against the skills actually present in `.claude/skills/`. That would be a cheap STATIC check if we want the two rubrics to converge rather than merely stop conflicting.

vybe:
## Summary

The compatibility catalog had drifted from both what the platform implements and what the public `create-agent` wizards generate. **101 checks → 88.** A freshly-scaffolded agent went from **9 findings (6 SOFT + 3 INFO)** — none the author's fault, none actionable — to **0 HARD, 0 SOFT, 5 INFO**, while the one HARD check guarding autonomous runs from hanging (**P-006**) went from *inert since it shipped* to actually firing.

Measured on an identical fixture, before/after:

| | dev | this branch |
|---|---|---|
| catalog | 101 | 88 |
| HARD on a wizard agent | 0 | 0 |
| **SOFT** | **6** | **0** |
| INFO | 3 | 5 |

The 5 remaining INFO are optional `template.yaml` metadata (`version`, `author`, `use_cases`, `capabilities`) plus the A-001 determinism suggestion — the "improvement suggestion" tier, by design.

## What changed

**Gating on fields nothing reads.** `template.yaml`'s `git:` block (`commit_paths`/`ignore_paths`/`push_enabled`) has **no backend reader anywhere** and **no bundled template declares it** — yet **T-017 was HARD**. Same for `metrics:` (D-006; `dashboard.yaml` is the read surface) and `.trinity/post-check` (I-005), whose only other mention was a `git_service` comment pointing back at I-005. Retired; `TRINITY_COMPATIBLE_AGENT_GUIDE.md` now marks the `git:` block inert so authors stop writing it.

**Failing every agent we ship.** `_slash_command()` anchored at position 0, but the marketplace's own generated schedules read `"Run /pipeline-tick"`, `"Run /project-steward"` — so **P-006 resolved zero command names and returned _"no scheduled/autonomous skills declared"_ on precisely the agents it exists to guard**. `_command_names()` globbed `.claude/commands/` only, the one layout the wizards never produce, so X-007 false-failed skill-based agents. G-002 compared against the **58-entry fleet-wide list Trinity itself injects** (`.bashrc`, `.ssh/`, `.claude/plugins/`).

**Self-contradiction.** P-009 tells authors to split reference material out of `SKILL.md`; P-004 then flagged the resulting `reference.md`. P-004 is now scoped to `SKILL.md`, so the two are a ladder (split at ~200, fail at 500).

**P-006 opt-out.** A skill may declare `automation: gated|manual` (the convention already used across `abilityai/abilities`) — an intentional human pause is a design decision, not a defect.

## Two findings worth a reviewer's attention

**1. A retired "duplicate" was not a severity duplicate.** `c_k002` was literally `return c_t015(snap)` — but **K-002 was declared HARD and T-015 SOFT**. Retiring K-002 as a plain duplicate would have silently downgraded the credential-declaration gate and undone ent#128's deliberate choice. **T-015 is promoted to HARD.** `test_ent128b1_compat_gates.py` — which asserts `hard_count >= 1` on a hostile declaration — is what caught it.

**2. The issue's stated cause for A-001 was partly wrong, and it changed the fix.** The issue says wizards emit `"Run /some-skill — …"`. They don't — all 13 `create-agent` schedule messages are **pure prose**. (The `agent-dev`/`trinity` plugins *do* emit `"Run /skill"`, which is what makes the `_slash_command` widening genuinely necessary.) So widening the matcher alone leaves A-001 SOFT-failing every wizard agent, which conflicts with the "zero unactionable SOFT" criterion. Prose messages are valid Trinity input, so **A-001 is also downgraded to INFO**.

## The anti-drift guarantee did not hold

`test_ids_match_doc` matched `^\|\s*([A-Z]-\d{3})\s*\|` — a **single** letter. The doc's `DP-001`…`DP-005` never matched, so five checks sat documented, indexed in the Check Index, and **entirely unimplemented** while the test reported the two files in sync.

- Regex widened to `[A-Z]{1,2}-\d{3}`.
- **New `test_severities_match_doc`** — the id set matching is not enough: an id-only test passes happily while the doc claims HARD for a check the catalog emits as INFO, and that column is what an operator reads.
- `DP-001`..`DP-004` **implemented** (`data_paths` is a real field — #1169 `materialize_data_paths`); `DP-005` retired (`.trinity/pre-snapshot` has no executor). DP-001 shares `git_service._is_safe_data_path` with the materializer (the A-002 discipline: validate with the parser the executor uses) and adds the containment the materializer omits.

## Traps avoided (deliberate non-changes)

- **`.trinity/post-check` stays in `_TRINITY_AUTHORED_PATHS`** even though I-005 is gone. #2070 *derives* the `!` re-includes from that tuple and 14 bundled templates ship `!.trinity/post-check` — removing it would untrack their hook on the next push, which is the exact #2070 regression. Only the stale comment changed.
- **The `.gitignore` auto-fix for F-003 still writes the canonical list.** Generating it is Trinity's job; demanding it of the author (G-002) was not.
- **A path is never read as a slash command.** `"Check /var/log/app.log"` yields no command — X-007 turns a resolved name into a SOFT finding, so a false positive there would manufacture the very class this PR removes. Checked in code, not with a `(?!/)` lookahead, which backtracks to `/et` out of `/etc/passwd`.
- **Retired ids are never reissued**, so persisted `agent_compatibility_results.checks_json` rows stay interpretable. Full table in the spec doc.

## Test plan

- [x] `tests/unit/test_compatibility_checks.py` — **95 passed** (was 59; +36 covering every behaviour change)
- [x] Affected surface — 57 files matching compat/credential/template/git/schedule: **1339 passed, 3 skipped**
- [x] Before/after measured by running the *same* wizard fixture in a clean `dev` worktree
- [x] New `TestWizardAgentIsClean` asserts 0 HARD **and** 0 SOFT on a fixture mirroring the layout `plugins/create-agent/skills/custom/SKILL.md` documents
- [x] No secrets in diff; enterprise submodule pointer deliberately not staged

**Pre-existing failure, not from this PR:** `test_codex_runtime_omits_claude_only_checks` fails when `test_inject_assigned_credentials.py` runs first — that module leaks a stubbed `services.agent_service.helpers` into `sys.modules`. Reproduced identically on a clean `dev` worktree; filed as #2140.

Fixes #2137


trinity-ability (approved):
Reviewed as the non-author account (2141 is vybe-authored, so vybe cannot approve it).

The catalog now matches what the platform actually reads. The change I checked hardest is the T-015 promotion to HARD: K-002's body was literally `return c_t015(snap)`, so retiring it as "a pure duplicate" would have silently downgraded the credential-declaration gate from HARD to SOFT and quietly undone ent#128 — `test_ent128b1_compat_gates.py` asserts `hard_count >= 1` on a hostile declaration, and it still does. Retired ids are documented as never-reissued, so persisted `checks_json` rows stay interpretable, and G-002's removal from `fixes.py` matches its retirement in `spec.py` (the F-003 auto-fix still writes the canonical list — generating it is Trinity's job, demanding it of an author was not).

Full suite green including pytest. Approving.

Issue #2136, opened 2026-08-12
refactor(frontend): two room UIs — 512 lines orphaned, and a rooms WS feed no view reads


trinity-ability:
Surfaced while reviewing #2135 (which deletes the orphaned `views/enterprise/Sessions.vue`, #2132).

ent#361 built a second, independent room UI rather than reusing the first. Deleting the old view makes the consequence unambiguous, and it is a **design decision, not cleanup** — which is why #2135 deliberately stops short of acting on it.

## 1. Five room components now have zero importers

`views/enterprise/Sessions.vue` was their only consumer:

```
components/rooms/NewRoomDialog.vue     110 lines -> 0 importers
components/rooms/ParticipantsRail.vue   89 lines -> 0 importers
components/rooms/RoomComposer.vue      132 lines -> 0 importers
components/rooms/RoomsRail.vue          63 lines -> 0 importers
components/rooms/RoomTranscript.vue    118 lines -> 0 importers
                                       512 lines total
```

The new `components/portal/PortalRoom.vue` (349 lines, ent#361) reuses **none** of them — it imports only `PortalAvatar`, `clientPortal`, and `renderMarkdown`.

So there are now two room UIs: 512 purpose-built lines nobody renders, and 349 fresh lines that do. The right answer is genuinely open — delete the old set, or consolidate `PortalRoom.vue` onto it (`RoomTranscript` and `RoomComposer` in particular look like the pieces `PortalRoom.vue` reimplemented). Deleting forecloses the second option, so it should be an explicit call.

## 2. A live WebSocket feed with no reader

`stores/rooms.js` is still imported and driven by `utils/websocket.js`:

```js
// utils/websocket.js
import { useRoomsStore } from '../stores/rooms'
const roomsStore = useRoomsStore()
```

But no view reads it — the only readers are the store itself and the WS client:

```
$ git grep -ln "useRoomsStore" -- src/frontend/src
src/frontend/src/stores/rooms.js
src/frontend/src/utils/websocket.js
```

Meanwhile `PortalRoom.vue` **polls every 3s** for the same transcript data. So the app maintains push plumbing whose output is discarded, while the surface that needs it pulls — the inverse of Invariant #10 ("don't poll for state that should be pushed").

This is the more interesting half: routing the existing room WS events into the portal surface would likely remove the poll (and with it the duplicate-row race that #2120's review found and guarded with an in-flight flag).

## Acceptance criteria

[...]

Issue #2129, opened 2026-08-12
docs: multi-agent Workspace chat + Sessions-page retirement shipped undocumented (ent#361, ent#381)


trinity-ability:
Found while validating #2120.

That PR shipped four issues' worth of work. Two are documented thoroughly:

- **ent#358** (Workspace absorbs Session) — `architecture.md` §Resumable Turns rewritten, `requirements/core-agent.md` §5.9 added, new flow doc `workspace-absorbs-session.md`
- **ent#286** (streaming) — documented inside the same architecture section

Two shipped with **no** entry in either `architecture.md` or `docs/memory/requirements/`:

- **ent#361** — multi-agent Workspace chat. New `components/portal/PortalRoom.vue` (349 lines), `PortalAgentPicker.vue` (109 lines), new `stores/clientPortal.js` actions against `/api/rooms`, new route `/workspace/r/:roomId`
- **ent#381** — Sessions page retirement. NavBar entry removed, `/sessions/:roomId?` converted to a query/hash-preserving redirect

```
$ git show origin/dev:docs/memory/architecture.md | grep -cE "ent#361|ent#381|PortalRoom|/api/rooms"
0
```

Per CLAUDE.md's tiered rule a **new capability** needs `requirements/<area>.md` + a feature-flow, and an **API/behaviour change** needs `architecture.md`. Multi-agent chat is the former; the Sessions retirement is the latter (it changes a documented nav surface and a live route's behaviour).

This matters beyond bookkeeping: the OSS↔enterprise seam here is genuinely subtle — an OSS Vue surface calling an entitled backend with a *portal* principal — and it is currently recorded nowhere. See #2128 for a bug that falls directly out of that undocumented seam.

## Acceptance criteria

- [ ] `architecture.md` covers the rooms-backed multi-agent Workspace chat, naming the OSS-frontend/enterprise-backend split and the `WorkspacePrincipal` auth path (mechanism only — no enterprise feature catalog, per the standing rule)
- [ ] `architecture.md` records the Sessions-page retirement + `/sessions` redirect alongside the existing Top-nav IA entries
- [ ] `requirements/` gains an entry for multi-agent Workspace chat
- [ ] Feature-flow doc created or an existing one extended, and the index updated
- [ ] `enterprise-docs-guard` stays green

Related: #2120, abilityai/trinity-enterprise#361, abilityai/trinity-enterprise#381

PR #2143, opened 2026-08-12
fix(workspace): gate the multi-agent picker on a rooms capability the client can read (#2128)

### LINKED ISSUE #2128: bug(workspace): multi-agent chat picker is offered on builds with no rooms backend — 404 dead end
Found while validating #2120 (ent#358/#361/#381), which merged the Workspace multi-agent chat.

## What happens

The Workspace "+ New chat" picker (`components/portal/PortalAgentPicker.vue`) lets a client select several agents, captioned *"they will share this conversation"*. Selecting 2+ routes to `store.createRoom()` → `POST /api/rooms`.

`/api/rooms` exists **only** in the enterprise submodule (`shared_sessions`). There is no OSS router with that prefix:

```
$ grep -rn 'prefix="/api/rooms"' src/backend --include='*.py'
src/backend/enterprise/backend/shared_sessions/router.py:29:    prefix="/api/rooms",
```

So on an OSS-only build — or any install without the `shared_sessions` entitlement — picking two agents produces a 404, and `Portal.vue` renders the generic `'Could not start that chat.'` The affordance is always offered and silently cannot work.

## Why it regressed

The surface this replaced **was** gated. `NavBar.vue` before #2120:

```vue
<router-link v-if="enterpriseStore.isEntitled('shared_sessions')" to="/sessions">
```

#2120 removed that entry (ent#381) and moved the capability into the picker, but no equivalent gate came with it. The new `/workspace/r/:roomId` route also carries no `requiresEntitlement`, unlike the `/sessions/:roomId?` route it replaced.

## Why this is not a one-liner

`isEntitled()` reads `enterprise_features` from `/api/settings/feature-flags`, which requires platform auth. The Workspace is reachable by an **external client** on an email-OTP portal session with no platform JWT, so the flag is not currently available to the principal that needs it. Fixing this properly means exposing a rooms-availability signal to a portal principal (e.g. on the `/my-agents` roster payload or a portal-scoped flags read), then gating the multi-select on it.

## Acceptance criteria

- [ ] On a build without `shared_sessions`, the picker does not offer multi-agent selection (or offers it visibly disabled with a reason) — never a dead-ended 404
- [ ] The gate works for an external client on a portal session, not only for a signed-in platform user
- [ ] Single-agent selection is unaffected in every edition
- [ ] `/workspace/r/:roomId` refuses/redirects when rooms are unavailable, rather than rendering an empty room

Related: #2120, abilityai/trinity-enterprise#361, abilityai/trinity-enterprise#381

trinity-ability:
Fixes #2128

## The bug

On a build with no rooms engine, the Workspace agent picker let a client select 2+ agents — captioned *"they will share this conversation"* — and then failed. `Portal.vue::onPickerConfirm` routes 2+ agents to `store.createRoom()` → `POST /api/rooms`, and that route is served **only** by a private module. On a community build (or an unentitled / `TRINITY_OSS_ONLY` instance) the request 404s and the picker renders the generic `'Could not start that chat.'` The affordance was always offered and could never work.

Regression provenance: the retired Sessions surface *was* gated (`NavBar.vue` carried an entitlement `v-if`, and the route carried `requiresEntitlement`). #2120 removed both and moved the capability into the picker; no equivalent gate came with it, and `/workspace/r/:roomId` carries no entitlement meta.

The reason it isn't a one-liner: the existing entitlement signal is unreachable by the principal that needs it. `enterpriseStore.isEntitled()` reads `GET /api/settings/feature-flags`, which is `get_current_user`-gated, and the store behind it early-returns `[]` for any caller without a platform JWT — i.e. for **every** external client, *including on an instance where the capability is present*. Routing `meta.requiresEntitlement` onto the room route is likewise inert: the entitlement block is nested inside `if (to.meta.requiresAuth)`, and `/workspace*` deliberately carries no `requiresAuth`.

So the signal rides on **`PortalRoster`** — a payload `get_portal_principal` already serves to both principal kinds and `Portal.vue::bootstrap()` already awaits first. One field, no new route, no new auth surface, no extra round-trip (`voice_available` is the per-agent precedent).

## The decision: gate, don't degrade

Hide the multi-select rather than showing it disabled with a reason. The viewer on this surface is the **operator's customer** — someone who can neither buy the missing module nor act on knowing it exists. So the field is named for the *capability* (`multi_agent_chat_available`), never the module or the edition, and no string on this path says "not licensed", names an edition, or prompts an upgrade. Fail-closed throughout: an unreadable registry, an older backend that omits the field, a partial payload, or a failed read all report the capability **absent**, because promising an affordance that cannot work is the bug being fixed.

When the flag is false: the picker is single-select, all five room store actions refuse **before** issuing a request, and `/workspace/r/:roomId` renders an honest refusal instead of mounting the room.

## What review changed — a real bug the implementation had

The self-heal originally lowered the capability on **any** 403/404 from a room call. But the serving module answers **ordinary denials with those same codes on a fully entitled instance**: *you cannot reach that agent* → 403, *you are not in that room* → uniform 404. A client clicking a just-un-shared agent, or opening a stale link to a room they'd left, would have flipped the **whole workspace** to single-select for the rest of the session — and replaced an actionable *"you do not have access to agent x"* with a false claim about the operator's build.

It now discriminates on the **body shape**, which is sound because absence and denial are authored by different layers:

| Shape | Author | Meaning | Action |
|---|---|---|---|
| `detail` is a plain **string** | FastAPI's own `"Not Found"` for an unmounted route; the entitlement gate's one-sentence 403 | the substrate is **absent** | lower the flag |
| `detail` is an object with a **`code`** | the serving module's own `{code, message}` | the module is **serving**; this is an ordinary refusal | pass through, surface the server's own words |

Verified against the real submodule source, both directions: **every** refusal in `shared_sessions/router.py` goes through one `_raise(e: RoomError)` helper producing `detail={"code", "message", …}` — there is no plain-string 403/404 path anywhere in it — while `dependencies.requires_entitlement` raises a plain-string 403.

The self-heal is why AC #4 holds *through a transition*, not just at load: `refreshThreads()` is event-driven, not periodic, so a capability that lapses while a room is **open** is only ever observed by that room's own calls. Without it the 3s poll swallows its 404 and the room is a dead end for the rest of the session; with it the poll lowers the flag, `<PortalRoom>` unmounts, and the refusal takes the stage.

## Verification

- **WAVE-4 `verify-local`: PASS** (`result.json` → `"status": "pass"`) — prod image build + import smoke, boot + health, real agent container, integration suite. Prod frontend image built.
- **Unit: 3 failed / 9695 passed / 17 skipped / 1 xfailed.** The 3 are the known pre-existing `test_pat_propagation_properties.py::TestKnownGaps::test_a_backslash_in_the_token_does_not_raise` params — present on `dev` itself, unrelated to this branch. Baseline matches exactly.
- **Integration: 70 passed / 13 skipped / 2 deselected.**
- Scoped re-run on this branch: `test_2128_workspace_rooms_capability.py` 11 passed; the ten neighbouring backend portal suites + `test_models_centralized.py` 216 passed; frontend `npm run test:unit` 13 files / 174 tests passed; `npm run build` and `npm run check:tokens` clean; `enterprise-docs-guard` pattern clean.

Tests are the two named regression files — `tests/unit/test_2128_workspace_rooms_capability.py` (11 cases: fail-closed on a raising/incomplete registry, `TRINITY_OSS_ONLY` lockdown, a *different* module not turning it on, the model default, a singleton swapped after import, and a module-identity assertion so a leaked test double fails loudly instead of going green on a lie) and `src/frontend/tests/unit/workspaceRoomsGate.spec.js` (selection semantics, structure guards, and the store gate — including the four discriminator cases, a 500 and a transport failure **not** lowering the flag, all five actions gated, and sign-out clearing both per-session flags).

## Raw-color note

`views/Portal.vue` `raw_gray` went **56 → 24** by hoisting nine duplicated stage-ink strings to five named constants. Rendered output is byte-identical — each constant is the exact string it replaced.

Measured, for the record:

| | `raw_gray` |
|---|---|
| committed baseline (generated 2026-07-31 @ `5b289996`) | 20 |
| `origin/dev` today | **40** |
| this branch mid-way (before the hoist) | 56 |
| this branch now | **24** |

So the file is well below `dev` but still 4 above the committed baseline. **The baseline file is stale and was deliberately NOT regenerated** — hiding a 20-count drift that predates this branch behind this PR would be the wrong place to fix it.

Worth stating plainly: **`gray` has no semantic token.** The design-system contract's colour section ends *"Everything else is gray"* and prescribes these exact shades for the ink ladder, so there is nothing to convert `text-gray-500 dark:text-gray-400` *into* — repainting body copy in `status-*` or `action-*` would be a defect, not compliance. The ratchet can therefore only be satisfied by writing the markup **once**, which is what the hoist does: the number of places a raw shade is written drops from twenty-odd to five.

## Known limitations / residuals

Recorded here deliberately; **no issues filed.**

1. `requires_entitlement` returns an *unauthenticated* 403 naming the feature and its licensing state. Pre-existing and wider than this PR — this change does not add to it, and the frontend never surfaces that string on the portal path.
2. The raw-color baseline is stale (2026-07-31; `dev` measures 40 against a recorded 20) and **the ratchet has no CI job** — `scan-raw-colors.mjs` always exits 0 and its own docstring calls the comparison "the future CI ratchet".
3. `MAX_PARTICIPANTS = 12` server-side, with no corresponding cap in the picker — a client can select more and be refused on confirm.
4. A mid-open capability flip silently drops a 2nd selection: `collapseSelection` keeps the most recent pick without telling the user one was dropped.
5. Single-select still renders a checkbox-shaped indicator. `role`/`aria-pressed` are corrected, but no radio primitive exists in the design system, so the visual affordance is unchanged.
6. `Portal.vue::activeRoomId` is write-only dead state left by ent#361 — set by `openRoom()`/`leaveRoomRoute()`, read by nothing (the route param is authoritative). Left alone rather than removed, to keep this diff to the bug.
7. OSS operators get **no in-product signal** that multi-agent chat exists at all. That is the deliberate consequence of gate-don't-degrade for the *client* surface; whether the *operator* deserves a different answer is a product call, not a bug.
8. `_multi_agent_chat_available()`'s failure warning is unmemoized — a registry that raises logs once per roster load.

Adjacent-but-out-of-scope by design: #2136 (two room UIs, `components/rooms/*` orphaned, push-vs-poll), ent#387 (room budget surface), #2129 (the full rooms architecture/requirements/feature-flow write-up — which is why this PR adds only the ~6 lines describing the capability seam it introduces, and no rooms flow doc).

### DIFF
--- a/docs/memory/architecture.md
+++ b/docs/memory/architecture.md
@@ -1296,10 +1296,11 @@ allow-list ∩ `user_invocable` — the same policy the MCP connector advertises
 outright; an agent exposing none falls back to its template-declared `use_cases`
 ("What You Can Ask"), sanitized and capped (6 × 200 chars). Clicking a hint **pre-fills
 the composer, never auto-sends** (`PortalBriefing.vue` → `prefill`). The future curated
-exposable-skills config (ent#178) slots into this same seam. Chats are strictly
-single-agent (the picker starts a new chat), so hints scope to the active agent by
-construction. ent#380 also fixed the briefing's metadata read — #138 called a
-nonexistent agent `/info` route, so descriptions were silently always `None`.
+exposable-skills config (ent#178) slots into this same seam. A chat holds one agent —
+or, where the capability below is present, several — and the picker starts a new chat
+either way, so hints scope to the active agent by construction. ent#380 also fixed the
+briefing's metadata read — #138 called a nonexistent agent `/info` route, so descriptions
+were silently always `None`.
 
 **Briefing hint grid is bounded (#2101):** with no connector allow-list configured every
 `user_invocable` skill becomes a "Things you can ask" card, so the hint set is belted
@@ -1309,6 +1310,29 @@ described-cards-first via `portalUtils.planHintDisplay`, the rest behind a count
 in-place "Show all N" toggle — deliberately **no nested scroll region**: the chat pane
 stays the single scroll axis, and the toggle counts the shipped list, never claiming the
 agent's full skill set). Hint *curation* stays the connector allow-list (ent#178 later).
+
+**The roster payload is *the* portal capability channel (#2128).** A portal principal
+cannot read `GET /api/settings/feature-flags` — that endpoint is `get_current_user`-gated
+and the frontend store behind it returns `[]` for any caller without a platform JWT, i.e.
+for **every** external client, including on an instance where the capability is present.
+So any UI gate on this surface takes its signal from `PortalRoster`, which
+`get_portal_principal` already serves to both principal kinds and `Portal.vue::bootstrap()`
+already awaits first — one field, no new route, no new auth surface, no extra round-trip
+(`voice_available` is the per-agent precedent). Reach for this before adding a second
+channel: the platform entitlement store is structurally unavailable here.
+`multi_agent_chat_available` is the first such field — resolved once per roster load from
+the entitlement registry, **fail-closed** (an unreadable registry reports the capability
+absent, because promising an affordance that cannot work is the bug it fixes), and named
+for the *capability* rather than the module or the edition, since this payload is served
+to an operator's customer. When it is false the picker is single-select, all five room
+store actions refuse before issuing a request, and `/workspace/r/:roomId` renders an
+honest refusal instead of mounting the room; a definitive 404/403 from the room endpoint
+self-heals the flag mid-session, so an entitlement that lapses between roster load and
+confirm collapses the picker rather than dead-ending. The frontend gate is **UX, not
+containment** — a portal token legitimately reaches the room endpoints where they exist,
+and the real boundary is the serving module's own roster-scoped access plus
+membership-scoped uniform 404s. Room data is untouched by the flag and reappears intact
+if the capability returns.
 
 ### Enterprise Modules (#847)
 
--- a/docs/memory/architecture.md
+++ b/docs/memory/architecture.md
@@ -1326,9 +1326,17 @@ absent, because promising an affordance that cannot work is the bug it fixes), a
 for the *capability* rather than the module or the edition, since this payload is served
 to an operator's customer. When it is false the picker is single-select, all five room
 store actions refuse before issuing a request, and `/workspace/r/:roomId` renders an
-honest refusal instead of mounting the room; a definitive 404/403 from the room endpoint
-self-heals the flag mid-session, so an entitlement that lapses between roster load and
-confirm collapses the picker rather than dead-ending. The frontend gate is **UX, not
+honest refusal instead of mounting the room; a 404/403 from any of the five self-heals the
+flag mid-session, so a capability that lapses between roster load and confirm — or while a
+room is open, which nothing else converges, since the sidebar refresh is event-driven — is
+observed by the next room call rather than dead-ending. **The status alone is not the
+signal**: a serving module authors its own refusals as a structured `detail: {code, …}`
+(*you cannot reach that agent* → 403, *you are not in that room* → uniform 404), while
+absence is a plain string — the framework's own "Not Found" for an unmounted route, the
+entitlement gate's sentence for mounted-but-unlicensed. Only the string form lowers the
+flag; a coded refusal is passed through so the server's own words reach the user, because
+reading one denied request as absence would turn it into a session-long false claim about
+the operator's build. The frontend gate is **UX, not
 containment** — a portal token legitimately reaches the room endpoints where they exist,
 and the real boundary is the serving module's own roster-scoped access plus
 membership-scoped uniform 404s. Room data is untouched by the flag and reappears intact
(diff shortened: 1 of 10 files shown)


vybe (approved):
Fail-closed in the right direction throughout: `multi_agent_chat_available` defaults False on the model, the entitlement read is wrapped so a roster never 500s over a capability bit, and an unreadable registry reports *unavailable*. Reading the entitlement registry rather than probing the route table is correct — ent#196 withdraws the claim of a module that fails to mount, so the registry already answers "mounted AND serving". Naming the field for the capability rather than the module is right for a payload that goes to an operator's customer.

The function-local import is load-bearing (the singleton is swapped in place by the test seam; a module-scope import would freeze the boot instance), and the module-identity assertion in the test file is what actually catches a leaked MagicMock — the `bool()` cast alone would not.

CodeQL's one high-severity alert here is a false positive and I dismissed it: `stripComments()` in `workspaceRoomsGate.spec.js` is a test-only helper that strips comments from repo source files the test reads off disk so prose about the gate is not scanned as code. It is not an HTML sanitizer, takes no untrusted input, and its output is asserted on rather than rendered.

Approving.

PR #2144, opened 2026-08-12
feat(grid): Recent failures info tile — fleet failures one glance away (ent#100)

trinity-ability:
Fixes abilityai/trinity-enterprise#100

The first **data** tile on the Grid's widget chassis: the newest failed executions across every accessible agent, plus the 24h failure total in the header meta. Failures previously required opening Operations → Executions and applying a filter.

**Zero backend change**, as ent#100 specifies. `GET /api/executions?status=failed` (rows) and `GET /api/executions/stats` (the 24h count) already exist and are access-scoped; both ride `fleetGrid.refreshBatchData()` — the one visibility-aware 60s poll the Grid already runs — gated on the tile being enabled. No new endpoint, no schema change, no new timer, and no files under `src/backend/` in the entire branch.

---

## ⚠️ ent#99 (Next schedules) is NOT in this PR

The sibling tile is **held and unimplemented**. This PR does not claim it and must not be read as delivering it.

It needs a new public endpoint (`GET /api/schedules/upcoming`) whose correctness rests on never `ORDER BY`-ing `agent_schedules.next_run_at`. That column is **mixed-format by design** — architecture.md (Invariant #16) documents the scheduler writing Z-suffixed strings while the backend's own writers use tz-aware `+00:00`/offset form, and states it is safe *only because nothing sorts it in SQL*. A lexicographic sort over those two formats is:

- green in CI,
- green on any UTC instance,
- **silently wrong only on non-UTC fleets** — no error, no log line, just a wrong "next run" order.

That is the wrong failure shape to ship into a code freeze, so ent#99 is deferred rather than rushed. Every shared-file touch in this PR was **append-only**, so ent#99 picks up unchanged whenever it lands.

## The carve-out that DID ship

`tests/unit/test_ent99_no_sql_order_by_next_run_at.py` — a repo-wide static guard forbidding SQL `ORDER BY next_run_at`.

It ships even though its motivating feature did not, because **the latent bug exists on `dev` today**: architecture.md documents the column as mixed-format and safe only by the absence of a SQL sort, with *nothing enforcing that absence*. The guard converts a documented convention into a mechanical one.

It **self-tests its own detector** (it plants a violation and asserts it is caught, so the guard cannot silently degrade into a no-op that always passes) and was proven against a real planted violation. 3 passed.

## What review changed

The "manufactured green" defence was reachable **one layer above** the tile's exhaustive fault sweep.

The tile refuses to render "No failures in 24h ✓" — a positive claim about the fleet, on the fleet's own monitoring surface — unless it has positive evidence, and the pure `utils/executionFailure.js::failuresTileState` sweeps that exhaustively. But `fetchRecentFailures` in the **store** coerced any non-array 200 body to `[]` while setting `loaded=true, error=false`. That triple is **byte-identical to a healthy empty fleet**, so it reached the pure function as positive evidence — through the one layer the 2⁴×2 sweep structurally cannot see.

`GET /api/executions` answers a bare JSON array (`response_model=List[FleetExecutionSummary]`), unlike every other fetcher in that store. A 200 that is not one is now a **failed cycle**, matching the sibling `/stats` fetcher, which already degrades an unreadable count to `null` (unknown, never zero). Last known-good rows are kept, so stale-while-revalidate is unaffected.

Verified by reverting the fix: **6 of the 12 new tests fail**, including `expected 'No failures in 24h ✓' not to contain '✓'`.

Two smaller review findings landed with it: the empty-roster branch asserted a fault it cannot distinguish from a fresh install (copy now names both causes and commits to neither — it refuses the ✓ either way), and requirements §9.12 ordering.

## The `.gitignore` fix

The setuptools block ends with a **bare `parts/`**, which git applies at every depth — so it silently swallowed the new `src/frontend/src/components/tiles/parts/` directory.

The failure mode is the nasty one: a component that **builds locally, is absent from the commit, and fails the frontend build on a fresh CI clone**. Re-included explicitly, directory first (git cannot re-include a file whose parent directory is excluded, so a `!**/*.vue` line alone would silently do nothing).

Verified with a 5-variant scratch-repo matrix and a fresh `--no-local` clone.

## Verification

- **unit**: 3 failed / 9687 passed / 17 skipped / 1 xfailed — the 3 are the known pre-existing `TestKnownGaps` backslash params, **present on `dev` itself**
- **integration**: 70 passed / 13 skipped / 2 deselected
- **verify-local**: `result.json` → `"status": "pass"`; prod frontend image built
- scoped re-runs on this branch: 55 frontend unit tests passed (5 files), the ent#99 guard 3 passed, `check:tokens` OK, raw-color ratchet exit 0 (new tiles contribute zero raw colors)

## Residuals (listed, deliberately not filed)

- **`error_code` persistence** — `schedule_executions` has no such column; the code lives on `TerminalEnvelope` in memory and is discarded at the terminal write. This is the named AC deviation: `failureCodeFromSummary` *reads* an anchored `[code]` marker when the platform emitted one and returns `null` otherwise — it never guesses. Persisting the column unblocks ent#100's taxonomy AC with zero UI churn.
- **Normalize `next_run_at` to UTC at both writers** — would let the new static guard be deleted *on purpose*, which is the right end state.
- **ent#99 (Next schedules)** — held, per above.
- **Staleness → loading downgrade not implemented** — bounded to a ≤60s post-wake window, and the current behaviour errs in the safe direction.
- **Row-count fit is structural + arithmetic, not screenshot-confirmed** under a running stack.
- **`MAX_ROWS` / `FAILURE_ROWS` coupling** between the tile and the store.
- **`inheritAttrs: false`** on `InfoTile` would drop a future `class` passed to it.
- **The "older" cause in the note copy is unreachable at `limit=4`.**


vybe (approved):
The reason this one is right is the part that is easy to get wrong: "No failures in 24h ✓" is a positive claim about the fleet, so it requires BOTH the rows fetch and the stats fetch to have succeeded on this cycle. Keeping two independent {data, loaded, error} triples instead of one shared pair is what stops a fault from rendering as a healthy empty fleet, and the note about `/api/executions` answering a bare JSON array (not `.items`) closes the other silent-empty path.

`failureCodeFromSummary` reading a marker the platform actually emitted — rather than re-deriving a code from a 200-char truncation — is the right call; a guessed label on a failure dashboard is worse than none, and the anchored lower-case-only charset keeps agent-authored `error_summary` from minting its own chrome.

All checks green. Approving.

Note for after merge: ent#100 is cross-tracker, so it will not auto-promote — it needs a manual `status-in-dev`.

PR #2145, opened 2026-08-12
feat(a2a): call_a2a_agent — outbound A2A calls to external agents (#736)

### LINKED ISSUE #736: feat: add call_a2a_agent MCP tool for outbound A2A protocol calls
## Summary

Add a new \`call_a2a_agent\` MCP tool that lets any Trinity agent call an external A2A-compatible agent (Google ADK, LangChain, AWS Bedrock, etc.) using the Agent2Agent v1.0 protocol. Trinity agents can delegate to agents on other frameworks without custom per-framework integration code.

## Context

The A2A protocol (Apache 2.0, Linux Foundation / AAIF) is now supported by 150+ organizations including Google, Microsoft, and AWS. Any agent exposing \`/.well-known/agent-card.json\` + a JSON-RPC endpoint is callable. This is the lowest-risk entry point for A2A in Trinity: client-only, no changes to agent-server, no inbound surface. See \`docs/research/a2a-protocol.md\` § 13.2 Option C.

## Acceptance Criteria

- [ ] New MCP tool \`call_a2a_agent(agent_card_url, message, context_id?, stream?)\` in \`src/mcp-server/src/tools/\`
- [ ] Fetches Agent Card from \`/.well-known/agent-card.json\`, validates A2A v1.0 schema
- [ ] Authenticates using Bearer token stored in calling agent's \`.env\` (key looked up by agent card domain)
- [ ] Calls \`SendStreamingMessage\` via SSE; streams token chunks back to the calling agent
- [ ] Maps \`TASK_STATE_INPUT_REQUIRED\` to a blocking prompt in the current chat session
- [ ] Long-running tasks (async): stores A2A \`taskId\` in Redis keyed \`a2a:task:{taskId}\`, returns execution_id for polling via existing \`get_execution_result\`
- [ ] Audit log entry for each outbound A2A call (actor: calling agent, target: agent_card_url)
- [ ] Respects \`agent_permissions\` — calling agent must have outbound A2A permission enabled (new permission flag or reuse existing mechanism)
- [ ] Unit tests covering: card fetch, auth negotiation, streaming response, INPUT_REQUIRED handling, error codes

## Technical Notes

- A2A v1.0 method names are PascalCase (\`SendMessage\`, \`SendStreamingMessage\`); v0.3.0 used slash-separated (\`message/send\`). Target v1.0 only.
- Auth: most public A2A agents use Bearer JWT or API key. Agent stores credential as \`A2A_{DOMAIN}_TOKEN=...\` in \`.env\`; tool resolves by matching card URL domain.
- Related files: \`src/mcp-server/src/tools/chat.ts\`, \`src/backend/services/task_execution_service.py\`
- Reference SDK: \`a2a-python\` (Apache 2.0) at https://github.com/a2aproject/a2a-python

trinity-ability:
Fixes #736

Adds `call_a2a_agent` (+ `get_a2a_task`): a Trinity agent can task an **external** A2A agent — Google ADK, LangChain, AWS Bedrock, a remote Trinity — completing the calling half of the protocol whose inbound server shipped in ent#157/#160.

## OSS-core by decision

**Outbound A2A calling ships OSS-core: no `requires_entitlement(...)`, logic stays in the OSS tree, and the OSS side ships a *working* target registry — not merely the seam.**

Recorded explicitly, because CLAUDE.md's default for an enterprise-tracker feature is *gated unless ruled otherwise*, and Trinity's convention (the ent#326 `/api/executions/timeline` and ent#384 skills-assignment precedents) is that an open-core ruling must **never** be inferred later from the mere fact that something merged. This one was weighed and answered OSS-core.

Reason: speaking a public interop protocol outbound is a generic platform capability, its inbound counterpart is already OSS-core, and a seam with no registered provider resolves nothing — the tool would answer "no targets configured" on every install. So `services/a2a_outbound.py` ships the OSS provider (admin-managed named endpoints in `system_settings`, one AES-256-GCM envelope — **no new table, no migration, no Alembic revision**), and per the documented seam mechanism a private per-agent provider **may** register and take precedence.

### A note for the human — not a decision made here

Option 4 (admin-owned named endpoints in `system_settings`) narrows the private registry's role to **per-agent scoping + the managed UI**. If that scoping was deliberate monetization, it remains available: the per-agent boundary can be gated later with this OSS list as the floor. Flagging the consequence rather than resolving it.

## Security model

The whole design follows from one refusal: **the tool takes an endpoint *reference*, never a URL.** An agent's tool arguments are LLM-generated and prompt-injectable, so a URL parameter would be a server-side-request primitive handed to the model.

- **Registry-only targets.** Agents choose a target by name; operators register endpoints via `PUT /api/settings/a2a-endpoints` (admin **+ `reject_agent_principal`** — registering decides where a credentialed request may go, so it is the *grant* half of the grant-vs-use line).
- **Call-time re-validation.** The registered URL is not trusted because it was validated at registration; it is re-validated at **use**, refusing `http://` at that point. Registration-time validation is a courtesy to the operator, not the control.
- **Card `url` pinned same-origin** (default-port-equivalent). The card is a hint, not an authority — and `securitySchemes` never selects the credential.
- **All redirects refused on both hops.** A 3xx is a failure, not something to follow.
- **Connect-time IP pinning** — one resolution, one validated address, pinned for both hops, with the registered hostname carried for `Host` + SNI + certificate verification. DNS rebinding is *closed*, not accepted as a residual, because this request carries a credential.
- **`trust_env=False`** — every other control reasons about the target IP, and a proxy makes the target irrelevant. Companion fix: that flag also disables httpx's `SSL_CERT_FILE`/`SSL_CERT_DIR`, so the CA context is rebuilt explicitly and TLS-inspecting-proxy installs keep working.
- **30s RPC timeout under a 45s total wall-clock deadline** over cancellable awaits, because httpx's `read` timeout is per-read and a trickling tarpit resets it forever. Wire-byte ceilings over `aiter_raw()`; any `Content-Encoding` refused rather than decoded.
- **Kill switch `A2A_OUTBOUND_ENABLED`, default OFF**, wired into `docker-compose.yml`, `docker-compose.prod.yml` and `.env.example` (the #1056/#1039 packaging class). Both routes 404 when off. Resolution is `system_settings` row → env → OFF, so a compose file that forgets the variable cannot make the feature unreachable.

## What review found and fixed

Stated plainly, because each was a real defect in an earlier revision of this branch:

1. **HIGH — the resolved credential could be echoed to the calling LLM.** h11 rejects an illegal header value by **echoing it** into the exception, which reached the 502 body *and* the backend log. Triggered by a stored credential carrying a stray line break — the routine paste artifact. Fixed in three layers. This is the same class `models._validate_pat_secret` exists for (ent#109); the regression test asserts the secret does not survive the error path.
2. **MED-HIGH — cross-peer credential leak via the dialect cache.** The cache was keyed on `hostname:port`, so a poll for a *separately registered* endpoint on the same host reused the first endpoint's resolved target — sending one peer's token to another peer's path. Fixed by keying on the registered URL.
3. **A peer-triggerable 500** from a malformed card URL escaping both error handlers.
4. **`_same_origin` compared raw hosts** while the validator canonicalises to IDNA — a mismatch that made the same-origin pin weaker than it read.

## Scope changes against the issue's own ACs

Four of #736's original acceptance criteria were **rejected or re-scoped**, and the tool signature changed (no `agent_card_url` parameter):

- **SSE streaming** — `message/send` only; no fake `stream` parameter.
- **`.env`-by-domain auth** — one reviewer called this "a catastrophic security risk in the original AC". Credentials come from the registry.
- **The Redis task map** — long remote work returns a handle; no Trinity execution row.
- **INPUT_REQUIRED as a blocking prompt.**

Each re-scope was endorsed independently by four reviewers.

## Residuals (listed, deliberately not filed)

- `template_registry_service._http_client` needs the same `trust_env=False`.
- Streaming (SSE) support.
- Connect-time pinning as a shared primitive rather than living in this client.
- ent#159 signature verification (blocked — `securitySchemes` cannot select the credential today).
- A private per-agent provider.
- The `effect_guard` fail-open-when-`execution_id`-absent debt.
- Registration-time HTTPS warning.
- **Any agent may spend any registered credential** — endpoints are platform-scope under the OSS list. Mitigated by the operator's registration decision, per-agent attribution (rate-limit key, audit row and activity row all name the spending agent) and rate bounds; the real fix is per-agent scoping.
- No agent-facing discovery of the endpoint list (deliberate).
- `A2A_TOTAL_DEADLINE` (45s) exceeds the default `MCP_A2A_TIMEOUT_MS` (40s).
- Neither the success text nor the peer error carries an "external peer" framing marker.
- `upsert_endpoint` is an unlocked read-modify-write.
- **`MCP_CHAT_TIMEOUT_MS` is absent from `docker-compose.prod.yml`** (and `.env.example`) — **pre-existing, untouched by this PR.**

## Verification

- **unit:** 3 failed / 9831 passed / 17 skipped / 1 xfailed — the 3 are the known pre-existing `TestKnownGaps` backslash params, present on `dev` itself.
- **integration:** 70 passed / 13 skipped / 2 deselected; `verify-local` → `"status": "pass"`.
- **Scoped re-run for this PR:** `tests/unit/test_736_*.py` → **147 passed**; `src/mcp-server` → **264 passed, 0 failed**.
- **#1033 closed empirically:** all four new modules (`a2a_client`, `a2a_outbound`, `a2a_outbound_service`, `a2a_protocol`) confirmed present in the **built image** and importing.
- **`A2A_OUTBOUND_ENABLED` read live from the running backend container → `false`** (prod compose leg, keys absent from `.env`, booted on defaults).


vybe (approved):
This is the platform's first backend-executed, credentialed, agent-triggerable outbound fetcher, so I read the egress path closely. The controls are the right ones and the reasoning behind each is written down: registry-only targets (an agent picks by NAME and can never supply a URL), HTTPS-only with userinfo **refused** rather than stripped, every resolved address must be public (one internal record among public ones refuses the whole URL), DNS failure fatal, and — the part that actually closes the rebinding TOCTOU — the validator returns the addresses it approved so both hops connect to a pinned address instead of re-resolving. `follow_redirects=False` outright, wire-byte ceilings over `aiter_raw()` with any `Content-Encoding` refused, and a total wall-clock deadline on top of httpx's per-read timeout.

Two things I checked specifically: the config is wired into `.env.example` **and both** compose files, so the lever is not inert on prod (the #1056 packaging class); and `A2A_OUTBOUND_ENABLED` defaults false, so this ships dark.

The earlier red CI was infrastructure, not a regression — `pytest (head, seed 67890)` was **cancelled** at 91% after 25 minutes, and `regression diff` then failed only because the cancelled job uploaded no junit artifact. Re-run is green.

Approving.

PR #2146, opened 2026-08-12
fix(retention): the blast-radius guard refuses instead of raising, and never drops its alarm

### LINKED ISSUE #1833: bug: retention_guard.evaluate docstring claims it never raises, but a non-numeric count_fn return raises TypeError
## Summary

`retention_guard.evaluate`'s docstring at `src/backend/services/retention_guard.py:153` states it **never raises**. That is false: the comparison `candidates <= threshold` (`:183`) sits **outside** the `try` that wraps `count_fn`, so a non-numeric `count_fn` return raises `TypeError` straight out of `evaluate`.

Found by #1771 slice a (PR #1825), pinned there as **observed behaviour** rather than xfail'd, because that PR's AC forbids product-code change. Test: `test_1771a_retention_edges.py::TestEvaluateRaisesContract::test_non_numeric_count_raises_out_of_evaluate`.

## Why it matters even though it is currently unreachable

All 8 `count_fn` implementations were confirmed to return `int(...)`, so this is **not** reachable in production today. The problem is that the guard's safety story is documented one way and implemented another:

- `cleanup_service.py:219-221` carries a comment that leans on the same false "never raises" claim.
- The actual protection today comes from **each sweep's outer try/except**, not from `evaluate`'s contract.

That is a real hazard for a component whose entire job is gating **destructive retention prunes** (the #1638 / #1644 data-loss class). A future `count_fn` that returns `None` on an error path — a very ordinary thing to write — would propagate a `TypeError` through a guard documented as raise-free, and the next maintainer has no reason to wrap it.

## Suggested fix

Pick one and make code and docs agree:

1. Move the comparison inside the `try` so `evaluate` genuinely never raises and a non-numeric count **refuses** (consistent with the guard's fail-closed design — a count it cannot interpret is exactly the case where it must not authorize deletion); **or**
2. Correct the docstring at `:153` and the `cleanup_service.py:219-221` comment to state the real contract, and make the outer try/except the documented protection.

Option 1 is more in keeping with the guard's stated fail-closed principle — every other error path in `evaluate` (count raises, ack lookup raises, threshold unreadable) already refuses.

Found via #1771 (PR #1825), follow-up 2 of 8.

### LINKED ISSUE #1834: bug: a failed retention-guard alarm permanently suppresses its own retry (memo written before the alarm)
## Summary

`announce_refusal` writes its `_last_refused[key]` memo at `src/backend/services/retention_guard.py:247` — **before** the `create_operator_queue_item` try at `:268`. If the alarm write fails, the memo already says "seen", so **no later cycle retries it** and the durable operator-visible alarm is lost until a restart or a window change.

Found by #1771 slice a (PR #1825) as matrix row C6, reported as UNSPECIFIED and deliberately not asserted — the fix is product code, which that PR's AC forbids.

## Impact

Not silent: the refusal is still logged at ERROR on the cycle it happens. But the **durable, operator-visible** signal — the operator-queue alarm — is the one designed to survive a log nobody is tailing, and that is exactly the one lost.

This matters because of what the guard protects: an over-threshold retention prune is **refused** and waits for an explicit acknowledgement (`POST /api/settings/retention/acknowledge`). The alarm is how the operator learns a refusal is waiting. Lose it and the prune stays blocked with no queue item explaining why — the failure mode is a silently stalled sweep rather than data loss, but it is precisely the observability the #1644 design leans on.

## Suggested fix

Move the `_last_refused[key]` write to **after** a successful `create_operator_queue_item`, so a failed alarm is retried on the next cycle.

Worth checking while there: the alarm is deliberately fire-and-forget (`create_item` is a blind `INSERT ... ON CONFLICT DO NOTHING`, and per the #1644 design the queue item **authorizes nothing** — the ack endpoint is the gate). So retrying the alarm is safe by construction; it cannot re-authorize anything. Confirm that reasoning still holds before moving the write.

Found via #1771 (PR #1825), follow-up 4 of 8.

trinity-ability:
Fixes #1833
Fixes #1834

Both issues live in `src/backend/services/retention_guard.py` — the #1644 blast-radius guard that stands between a narrowed retention window and an unbounded `DELETE`. They are two halves of one contract: #1833 is "the guard raises where it promised to refuse", #1834 is "a refusal loses its durable alarm". Fixed together because the second is only reachable through the first's failure paths.

---

## ⚠️ Out-of-issue fixes — read this section first

Neither closing keyword covers these. They were found while working the two issues and are the most important content of this PR.

### 1. `count_negative` — a fail-OPEN inside a fail-closed guard

`evaluate` decided with a bare:

```python
if candidates <= threshold:
    return GuardVerdict(True, candidates, threshold, "under_threshold")
```

`-1` is **this module's own "count unknown" sentinel** (it is what `count_failed` returns). And `-1 <= 1000` is `True`. So a `count_fn` that reported failure using the module's own idiom **authorized the prune**. The count does not bound the delete — `prune_execution_logs` takes no count and drains fully — so the outcome is #1638 replayed, from inside the guard written to prevent #1638.

Neither issue describes this. It is now an explicit refusal reason (`count_negative`) with its own ERROR.

A pre-existing test, `test_1771a_retention_edges.py::A11`, **pinned the old behaviour** as `(True, "under_threshold")`. Its own docstring named the hazard and recorded the assertion as observed-not-endorsed *specifically so that changing it would be a deliberate, visible edit*. This PR is that edit; the rewritten test says so in place.

**The human may want a tracking issue for this. I have deliberately not filed one.**

### 2. NaN / unrepresentable-count normalisation

`verdict.candidates` reaches three surfaces: the alarm `message`, the alarm `context` (via `json.dumps`), and `candidate_count` in `GET /api/settings/retention`. `json.dumps` emits a **bare `NaN`** for a non-finite float — accepted by Python's `json.loads`, **rejected by a browser's `JSON.parse`**:

```
json.dumps({"candidate_count": float("nan")})  ->  '{"candidate_count": NaN}'
```

That takes down the *entire* `/api/settings/retention` and `/api/operator-queue` response, not just the one field. `candidates` is now normalised to an int (`-1` = unknown) before publication. The **decision** still uses the raw value, so NaN keeps its pinned IEEE-754 behaviour.

### 3. `blocked_sweeps` observability field

`GET /api/settings/retention` only listed `over_threshold` as pending-approval. Once `evaluate` refuses instead of raising, an undecidable sweep is blocked forever in `cleanup_service` while the panel renders a clean "nothing pending" — the guard's own anti-pattern ("a guard that fails open manufactures confidence") relocated from the prune to the operator surface. Before this PR that state was at least *loud* (a 500). New `blocked_sweeps` list carries identifiers and reason codes only — the same SECURITY rule as the alarm payload.

---

## The #1833 direction, and why

The issue asks for a docstring that matches behaviour. The fix goes the other way — **`evaluate` never raises, and refuses instead** — decided by a caller the issue never mentions.

`cleanup_service._guard_allows` wraps the call, so a raise there is survivable: the prune is skipped (fail-closed on the data). But `routers/settings.py::get_retention_status` re-runs the same `evaluate` **live and unwrapped inside an HTTP handler**, so a raise there is a **500 on the panel the operator uses to approve the very prune that was refused**.

That asymmetry decides it. A raise protects the DATA but destroys the SIGNAL — the refusal loses its ERROR and its durable operator-queue alarm, surfacing as a nondescript "Error pruning ...". The belt in `_guard_allows` stays anyway, now documented for what it actually buys.

## What review found

**This is the 4th instance of this module's incomplete-fix class.** `_describe_exception` was added *for exactly that class* — and applied to `announce_refusal` only, while `evaluate` — the function #1833 is named after, whose docstring promises "NEVER raises" — still passed raw exceptions into deferred `%s` on all three refusal paths. Measured: with a re-raising log handler installed, `evaluate` raised straight out of a REFUSAL. Fixed at every site, with the helper moved above `evaluate` so the ordering makes the dependency obvious.

**Two reviewer-suggested fixes were REJECTED on probe evidence**, recorded here so they are not re-proposed:

- *"Route non-finite counts to `count_uninterpretable`"* — tidier-looking and **wrong twice**. A NaN count **with an acknowledgement on file genuinely IS allowed** (`nan <= threshold` is False, so it reaches `is_acknowledged`) — pinned by `test_1771a::test_nan_count_with_ack_is_allowed`. Routing it to the undecidable phrasing would tell the operator "an acknowledgement will NOT clear this" about a refusal an acknowledgement **does** clear — a fresh lie — and would break that pinned test. Instead the alarm gets its own third branch: drop the impossible number, keep the instruction that works.
- *"Align `blocked_sweeps` and the alarm on `ack_lookup_failed`"* — they classify it **differently on purpose**. It is reached only after the count was genuinely over threshold, so an ack IS the right remedy and the alarm keeps that instruction; but the panel must not offer an approve **control** while the path that reads approvals is the broken thing — the click would appear to succeed and change nothing.

## Docs corrected

Both claims verified against the code, not assumed:

- **Stale window count `7` → `8`** (in `architecture.md` and `requirements/lifecycle-observability.md`, two places each). Verified: `RETENTION_OPS_KEYS` = **8** entries (`agent_reminders_retention_days` was added by #1296), and `cleanup_service` has **8** `_guard_allows` call sites.
- **"six of the seven prune accessors loop `while True`"** — wrong in **both** terms. Measured: the denominator is 8, and the split is **5 drain fully / 3 genuinely cap**. Draining: `prune_execution_logs`, `prune_execution_rows`, `cleanup_old_health_records` (via `_monitoring_ops.cleanup_old_records`), `prune_agent_reports`, `prune_agent_reminders`. Capped: `prune_operator_queue_terminal_items` (`limit=`) and the two soft-delete sweeps via `find_soft_deleted_{agents,schedules}_past_retention(limit=...)`. The point that survives: `RETENTION_CHUNK_SIZE_PER_CYCLE` bounds each **transaction**, not each call — the bound on destruction is this guard, not chunking.

## Verification

- **Fail-closed table of 30 hostile inputs byte-identical through every edit** — the refusal direction never moved while the code was restructured.
- **`alarm_delivered = True` at exactly one line** (`retention_guard.py:675`), with every alarm-failure `except` path `return`ing before it — so a failed alarm can never mark itself delivered. This is the #1834 fix: the memo used to be written *before* the attempt, so one failed write permanently suppressed its own retry.
- **The key test was proven to fail against a planted `bool()`-only implementation** — i.e. it actually detects the fail-open it guards, rather than passing vacuously.
- **The new tests were proven to fail against unpatched `dev`.**
- Scoped test run (this neighbourhood): **240 passed** — `test_1833_retention_guard_contract.py`, `test_1834_retention_alarm_retry.py`, `test_1644_retention_guard.py`, `test_1771a_retention_{edges,properties}.py`, `test_cleanup_inner_sweeps.py`.
- Branch-level suites (serialized slot): unit **3 failed / 9744 passed / 17 skipped / 1 xfailed** — the 3 are the known pre-existing `TestKnownGaps` backslash params, present on `dev` itself. Integration **70 passed / 13 skipped / 2 deselected**; `verify-local` → `"status": "pass"`.

## Residuals

Two of these should be filed as their own issues. **I have not filed them.**

1. **`_alarm_id` is reason-blind** — ⚠️ **the one item that should not merge without a filed follow-up.** The durable alarm row keeps its FIRST content for a given `(setting_key, window_days)`. In the `undecidable -> over_threshold` direction the row still reads *"an acknowledgement will NOT clear this"* at exactly the point where acknowledging **is** the remedy — telling the operator not to take the one action that unblocks the sweep, in the durable half specifically designed to outlive a log nobody is tailing. The prune can stay blocked indefinitely on the alarm's own advice. (The opposite direction is bounded and inert.) Fixing it needs either `verdict.reason` in `_alarm_id` — a second durable row per reason, and its natural-key shape is pinned — or an UPDATE path the sink does not have (`create_item` is INSERT ... ON CONFLICT DO NOTHING). A design decision, not a one-liner.

2. **`DELETE /api/settings/{key}` does not block `RETENTION_OPS_KEYS`** — a real, pre-existing, unvalidated path to narrowing a retention window. `PUT` 422s these keys and points at the dedicated route; `DELETE` does not, so a window can still be reverted to its code default through the generic catch-all. Direct sibling of ent#297 / #1966. Out of scope here, but it is the same class both issues are about.

Lower-severity, recorded in code comments:

- `-1` still reaches `pending_acknowledgements`, so `Settings.vue` renders **"-1 items"** beside an Approve control. Quiet wrong number where there used to be a loud broken page (see NaN fix above); the proper fix is a frontend change.
- `blocked_sweeps` covers **2 of 8** windows (the two ack-gated sweeps the endpoint re-runs) and **has no frontend renderer yet**. The other six reach an operator only via the operator-queue alarm. Widening the loop changes what the endpoint costs and what "approve" means for a sweep whose floor is 1000 — deliberate follow-up, not oversight.
- A "lying comparable" `count_fn` (real bools from `__le__`/`__lt__`, nonsense value) still passes — trusted-code-only surface; all 8 accessors return `int(COUNT(*))`.
- `verdict.candidates` is `-1` on **allowed** verdicts for non-int counts.
- `_refusal_episodes`' 8-key bound is by construction, not enforced.
- `retention-guard-` is absent from `_RESERVED_ID_PREFIXES` (the #1632 agent-authored reserved-id guard).
- A **flapping** `count_fn` re-fires the transition ERROR each cycle and resets the escalation clock — accepted deliberately over the alternative (window-only freshness silently swallowed #1833's new signal entirely); dampen with a per-key minimum re-ERROR interval if ever observed.
- No coverage-parity guard ties the refusal-reason set to its test matrix.


vybe (approved):
Validated via /validate-pr. Closes the raise-free contract `evaluate` already claimed: the comparisons move inside the try, and `type(under) is not bool` is the load-bearing part — an object whose __le__ returns a truthy non-bool would make a bare bool() cast ALLOW a prune on a count the guard never understood, i.e. a silent fail-OPEN introduced by the fix for a fail-closed bug. `candidates` is always an int so json.dumps(allow_nan=False) and the admin panel both survive. #1834's arm-before-attempt ordering is right (a BaseException inside create_item leaves an undelivered episode the next cycle retries, rather than a memo claiming 'handled'), and no-give-up is correctly argued from the sink being the platform's own DB rather than an external webhook. The reason-keyed freshness cost (flap → 288 ERRORs/day) and the undecidable→over_threshold stale-row direction are both named as follow-ups rather than left to be discovered. Resolved a tests/registry.json conflict by rebuilding from git stages (183 entries, union, no losses). All 21 checks green.

Issue #2133, opened 2026-08-12
bug(workspace): awaitPersistedReply can poll for an hour when the in-flight marker is orphaned


trinity-ability:
Found while validating #2120.

## Context — the unbounded wait is deliberate, and correct

#2120's review fixed a real double-billing chain: `awaitPersistedReply` used to budget a fixed ~10.5s from stream end, so any turn persisting later (notably the cold `--resume` retry, which re-runs the WHOLE turn) was rendered "not delivered" with a Retry that ran and billed it a second time.

The fix makes the wait bounded by the **server's** answer instead of a stopwatch, which is the right shape:

```js
// components/portal/PortalConversation.vue
for (;;) {
  ...
  if (data.inFlightExecutionId) { idlePolls = 0 }
  else if (++idlePolls >= REPLY_IDLE_GIVE_UP) return null
  await new Promise((r) => setTimeout(r, REPLY_POLL_MS))   // 700ms
}
```

## The residual

The loop has no absolute ceiling. It exits only when the server stops reporting an in-flight execution. That marker is cleared in a `finally` on every exit path of `start_portal_turn._run`, so the normal case is fine — but `except Exception` does not catch `CancelledError`, and a hard backend kill skips the `finally` entirely. The marker then survives until its TTL:

```python
def mark_turn_inflight(session_id, execution_id, ttl_seconds: int = 3600) -> None:
```

One hour. At `REPLY_POLL_MS = 700` that is ~5,100 `GET /history` requests from a single tab, each hitting the DB, for a turn that is already dead. Multiple tabs multiply it. Nothing in the UI tells the user the turn is not coming.

## Suggested fix

A generous absolute ceiling that cannot be hit by a legitimate turn — the agent's own execution timeout plus the cold-retry allowance (so roughly `2 × execution_timeout + buffer`) — after which the UI reports that it lost track of the turn and points at history, **without** offering a Retry (the original billing hazard).

Worth pairing with a wider poll interval once elapsed time is large, so a long legitimate turn costs far fewer requests.

## Acceptance criteria

- [ ] The poll cannot run unbounded when the in-flight marker is stale
- [ ] A legitimately long turn (including one cold retry) still resolves normally — no regression of the #2120 fix
- [ ] Giving up does not offer Retry for a turn that may have run
- [ ] Regression test naming this issue

Related: #2120, abilityai/trinity-enterprise#358, abilityai/trinity-enterprise#286

Issue #390, opened 2026-04-18
Fleet sync audit endpoint (S6)


vybe:
## Summary

New endpoint to get sync status of all agents at once for fleet-level health checks.

## Context

Currently, checking sync health requires querying each agent individually. Operators and the \`/trinity:sync\` skill need a bulk audit capability.

### New endpoint
- [ ] \`GET /api/fleet/sync-audit\`
- [ ] Returns table of all agents with:
  - \`agent_name\`
  - \`branch\` (working branch)
  - \`last_pushed_sha\`
  - \`last_pushed_at\`
  - \`local_head_sha\`
  - \`unpushed_commits\` (count)
  - \`dirty_tree\` (boolean)
  - \`duplicate_binding\` (boolean) — detects P5 at fleet level

### Duplicate binding detection
- [ ] Query: 
\`\`\`sql
SELECT github_repo, working_branch, COUNT(*)
FROM agent_git_config WHERE source_mode = 0
GROUP BY github_repo, working_branch HAVING COUNT(*) > 1
\`\`\`
- [ ] Flag affected agents in response

### Response format
\`\`\`json
{
  "agents": [
    {
      "name": "my-agent",
      "branch": "trinity/my-agent/abc123",
      "last_pushed_sha": "def456",
      "last_pushed_at": "2026-04-17T10:00:00Z",
      "local_head_sha": "def456",
      "unpushed_commits": 0,
      "dirty_tree": false,
      "duplicate_binding": false
    }
  ],
  "summary": {
    "total": 10,
    "in_sync": 8,
    "ahead": 1,
    "dirty": 1,
    "duplicate_bindings": 0
  }
}
\`\`\`

## Technical Notes

**Files to modify:**
- \`src/backend/routers/git.py\` — new endpoint
- \`src/backend/services/git_service.py\` — aggregation logic

Most data already computable per-agent via existing status endpoint; this batches the queries.

Part of Epic #381

PR #2148, opened 2026-08-13
feat(workspace): @mentioning an agent from a 1:1 starts a group discussion (ent#361)

dolho:
Finishes [ent#361](https://github.com/Abilityai/trinity-enterprise/issues/361) — the p1 AC added on the 12th: *"@mentioning an agent from ANY existing chat (including a 1:1) creates a group discussion with them."*

The issue has been sitting at `status-in-dev` without it. There was nothing to build on: `client_portal/` had **zero** mention handling, so a 1:1 could only ever stay a 1:1.

Typing `@other` in a 1:1 now creates a room with both agents, carries the message into it, and moves you there.

## Two halves, deliberately in different layers

The in-room half — @mentioning someone not in a room adds them (AC #4) — is engine-side: abilityai/trinity-enterprise#390. Membership is the engine's concern and agent replies flow through it.

This half answers a different question — *should this conversation become a room* — and answering it in OSS would mean importing the private module. Routing it through the existing `POST /api/rooms` instead keeps `create_room`'s per-agent ACL as the **single** enforcement point rather than growing a second one that can drift from it.

## An unresolvable @name is plain text, not an error

Reporting "no such agent" would answer, for **any** string a user cares to type, whether an agent by that name exists on the instance. The rooms engine already behaves this way; the composer now matches it.

The pattern mirrors the engine's `_MENTION_RE` on purpose, and is pinned by tests. Drift there builds a room around a handle the engine then renders as plain text.

## Ordering is not incidental

Create the room → navigate → **then** post. Posting first leaves the user staring at the old thread while the agents they just summoned reply somewhere they cannot see. A failed post leaves them in a room they can retype into, which is recoverable; a created-but-unreachable room is not. A failed escalation gives the typed text back rather than swallowing it.

## Worth flagging

**AC #4's premise was wrong, and this is why ent#361 looked done.** The issue says *"today a chat can only become multi-agent by @mentioning someone mid-conversation; that path is fine and stays"*, and asks that it "still works". `resolve_mentions` matched only names **already in the room** and documented the intent outright — *"a mention can never reach outside"*. So the AC reads like a regression guard and is actually a feature request. Both halves are new work, not preserved behaviour.

## Verification

260 frontend tests (18 files) incl. 8 new for the resolver — roster-scoped resolution, the excluded current agent, dedupe, multiple agents in order, hyphen/digit names matching the engine, and the empty/missing-roster cases. Build clean. Enterprise side: 9 new tests, 294 passed vs 285 before (the 29 failures are pre-existing on `main`, none in shared_sessions).

Gated on the same rooms capability as the picker (#2128) — no rooms, no escalation, `@` stays ordinary text.

Submodule pin advanced to the enterprise commit; **abilityai/trinity-enterprise#390 should merge first.**

Related to abilityai/trinity-enterprise#361

### DIFF
--- a/docs/memory/requirements/core-agent.md
+++ b/docs/memory/requirements/core-agent.md
@@ -403,6 +403,43 @@
   content rather than a chat) is abilityai/trinity-enterprise#360.
 - **Flow**: `docs/memory/feature-flows/workspace-sidebar-ia.md`
 
+### 5.11 Workspace multi-agent chats — @mention escalates a 1:1
+- **Status**: ✅ Implemented (2026-08-13)
+- **Requirement ID**: WORKSPACE_MENTION_TO_GROUP
+- **GitHub Issue**: abilityai/trinity-enterprise#361
+- **Description**: @mentioning another agent turns a conversation into a group
+  discussion — from a **1:1** (which creates a room containing both agents and
+  carries the message into it) and from **inside a room** (which adds the
+  mentioned agent as a participant).
+- **The AC was a feature request wearing a regression guard.** ent#361 AC#4 asks
+  that "@mention of a non-participant *still works* and adds them (existing path
+  preserved)", and its Context says a chat could already become multi-agent that
+  way. Neither was true: `resolve_mentions` matched only names already in the
+  room and documented that a mention "can never reach outside", and the portal
+  had no mention handling at all. There was nothing to preserve.
+- **Two halves, deliberately in different layers**:
+  - **In-room** is engine-side (`shared_sessions.post_message` →
+    `_join_mentioned_newcomers`), because agent replies flow through the engine
+    and membership is its concern.
+  - **1:1 → room** is a UI act: the Workspace resolves the mention against the
+    roster it already holds and uses the existing `POST /api/rooms` +
+    `POST /api/rooms/{id}/messages`. OSS must not import the private module, and
+    routing it through the rooms API keeps `create_room`'s per-agent ACL as the
+    single enforcement point rather than adding a second one.
+- **Safety properties** (both halves): an @name that is not an agent the caller
+  can reach stays **plain text and is never an error** — a "no such agent" reply
+  would answer, for any string typed, whether an agent by that name exists;
+  **only a human** may recruit (an agent that could pull agents into a room is a
+  spend amplifier and a prompt-injection lever); the participant cap is
+  re-checked per addition; a closed room admits nobody.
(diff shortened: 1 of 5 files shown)


vybe (approved):
Validated via /validate-pr. All 4 required checks, all 6 pytest shards and e2e green after the fixes below. No secrets, no new config var, no schema change.

**I corrected the submodule pointer before merging — it would have shipped broken.** The pin named `fade3cbb`, the pre-squash head of trinity-enterprise#390. Because #390 was squash-merged (and its branch deleted), that commit is not reachable from enterprise `main` — `compare` reports it **diverged**, ahead 1 / behind 1. A fresh `git submodule update` could not fetch it at all. Nothing in CI can catch this: both submodules are `update = none`, so CI never checks them out and stays green either way. Repointed at `dcf0b52`. I also confirmed `dev`'s prior pin `a0ad11c5` is a clean ancestor of enterprise `main` exactly one commit behind, so after this merge the pointer equals main's tip with nothing skipped.

**Conflict resolution** against the newly-merged `dev` — four conflicts, all additive, so each is a union rather than a choice:
- `core-agent.md` — #2147 (ent#360) and this branch both wrote `### 5.11`. #2147 merged first and keeps it; ent#361 renumbered to **§5.12**. No cross-references to either anchor exist elsewhere in `docs/` or `src/`.
- `portalUtils.js` — #2133's budget constants and ent#361's `mentionedAgents` are unrelated exports; both kept.
- `PortalConversation.vue` — one import line naming all three symbols.
- `portalSidebarIA.spec.js` — the conflict cut through two `describe` blocks; ent#361's is closed and #2133's follows.

Verified locally on the merge result before pushing: 282 frontend tests across 19 files, vite build clean, and an unfiltered conflict-marker sweep over the whole tree empty.

On the change itself: keeping the 1:1→room decision in OSS and routing it through the existing `POST /api/rooms` is the right layering — it keeps `create_room`'s per-agent ACL as the single enforcement point instead of growing a second one, and avoids OSS importing the private module. The unresolvable-@name-stays-plain-text rule matches the engine's own behaviour and correctly avoids turning the composer into an existence oracle; mirroring `_MENTION_RE` with tests pinning it is the right guard against the two sides disagreeing. Create → navigate → post ordering is argued correctly: a failed post is recoverable, a created-but-unreachable room is not.

ent#361 is cross-tracker, so nothing auto-promotes — its enterprise half (#390) is already merged and I'll leave the issue's status to the release close.

PR #2149, opened 2026-08-13
fix(workspace): the six #2139 review findings, which missed the merge (ent#359)

dolho:
**These fixes were meant to be part of #2139 and are not on `dev`.** #2139 merged at `07:47:27Z`; I pushed them at `07:57Z` — ten minutes late, onto a branch whose PR had already merged, so they went nowhere. Neither #2147 nor #2148 carries them either. Verified: `MAX_STARRED_CHATS` is absent from `dev`.

Same commit, replayed onto `dev`. All six were CONFIRMED by the high-effort review of #2139 before it merged.

## The one that gets worse with time

**The star row cap is a one-way door.** `mark_chat_read` creates a row for every chat ever opened — rooms included, and those always report `unread: 0` — while unstar only nulled `starred_at`, and no `DELETE` against `enterprise_portal_chat_state` existed anywhere. No purge, no cascade, no retention sweep.

So a viewer reaches the cap through activity they cannot undo, and from that moment **every star click returns 409 "unstar some first" forever, and unstarring frees nothing.** The code comment, the flow doc, requirements §5.10 and the test all asserted unstar was the recovery path; the test only checked `starred_at` went NULL and never that a subsequent star succeeded, which is exactly why the false claim survived two reviews.

Split into two caps: a total-row cap still bounds abuse (neither writer validates the chat exists — a 404 would be an enumeration oracle), and a separate **starred**-row cap is the one a real user reaches, so unstarring genuinely frees capacity. Unstarring a chat with no read cursor now deletes its row. Docs corrected; the test asserts the recovery actually works.

## The one that is visible today

**Route-driven opens never write a read cursor.** `markRead` was wired only to sidebar clicks and session adoption; `bootstrap()` and the route watcher set `pendingSession` without it. So reloading or following a bookmark to `/workspace/c/<sid>` renders the reply but never marks it read — the sidebar keeps badging the conversation you are looking at, through **every** reload. That is the commonest way into a chat.

## The rest

| | |
|---|---|
| `markRead` not awaited | `refreshThreads` raced the cursor UPSERT, so `GET /chat-state` could restore the stale count and the badge reappeared on the chat being read |
| `reattach()` never emitted `sessions-changed` | reload into a running turn, watch the reply land, keep the badge |
| Star was mouse-only | keydown from the focused star bubbled to the row's `role="button"` handler, which `preventDefault`ed the button's own activation and opened the chat instead |
| Unreachable fallback removed | the unread query INNER JOINs the state table, so the "cursor cleared out from under us" branch could never fire — a safety net that cannot fire is worse than none |

## Verification

22 backend tests, 252 frontend (18 files), build clean.

Related to abilityai/trinity-enterprise#359

### DIFF
--- a/docs/memory/feature-flows/workspace-sidebar-ia.md
+++ b/docs/memory/feature-flows/workspace-sidebar-ia.md
@@ -124,10 +124,16 @@ destination with its own content, not a shortcut to a conversation).
 - **No existence check on `chat_id`, on purpose.** A 404 for an unknown chat
   would answer "does chat X exist?" for every id in the install (invariant #8).
   The write lands in the caller's own namespace, so an unknown id gains them
-  nothing — a per-viewer row cap (`MAX_CHAT_STATE_ROWS`) bounds the write
-  instead of validation. The cap applies only to writes that **create** a row:
-  capping updates would freeze a user at the ceiling out of unstarring, which is
-  the only action that gets them back under it.
+  nothing — per-viewer caps bound the write instead of validation.
+
+  **Two caps, because one number cannot do both jobs.** `MAX_CHAT_STATE_ROWS`
+  (1000) bounds abuse: any row, star or read cursor. `MAX_STARRED_CHATS` (200)
+  bounds the starred set and is the only one a real user reaches. They have to
+  be separate — a read cursor is written for every chat ever opened, so a single
+  total cap gets consumed by ordinary use, and then every star returns
+  *"unstar some first"* while unstarring frees nothing, forever. Counting only
+  starred rows makes that advice true, and unstarring a chat with no read cursor
+  deletes its row outright so the table cannot only ever grow.
 
 ## Tests
 
--- a/docs/memory/requirements/core-agent.md
+++ b/docs/memory/requirements/core-agent.md
@@ -396,7 +396,11 @@
   `PUT|DELETE .../chat-state/{kind}/{id}/star`, `POST .../chat-state/{kind}/{id}/read`.
   No roster gate (every row is keyed by the caller's own email) and no existence
   check on the id — a 404 for an unknown chat would be an enumeration oracle
-  (invariant #8); a per-viewer row cap bounds the write instead.
+  (invariant #8); two per-viewer caps bound the write instead. A total-row cap
+  (abuse) and a separate **starred**-row cap: read cursors accrue from ordinary
+  use, so a single cap would be spent by activity the user cannot undo, making
+  the 409's "unstar some first" advice false. Unstarring a chat that carries no
+  read cursor deletes its row.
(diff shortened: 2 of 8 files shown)


vybe (approved):
Validated via /validate-pr. All 4 required checks green, e2e green, 5 of 6 pytest shards green — see the note on the sixth below. No secrets, no new config var, no schema change.

**I merged `dev` into the branch before merging.** Its CI last ran against a `dev` from before #2147, #2148, #2150 and #2151 landed, three of which touch the same portal files (`client_portal/service.py`, `PortalConversation.vue`, `Portal.vue`). A green run against a stale base is not evidence the combination works. The merge was conflict-free; 282 frontend tests across 19 files pass on the result.

**On the cancelled `pytest (base, seed 67890)` shard — checked, not waved through.** It ran 12:32:10→12:57:26 and died on the job's `timeout-minutes: 25`, with "Upload JUnit XML" consequently failing. That is the documented #2019 mode, and the workflow's own comment is explicit that it hits *whichever PR is running* and that the **base** side is plain `dev`, i.e. unrelated to the PR it reddens. Two independent reasons the regression verdict still holds: `scripts/ci/diff-pytest-failures.py` is built for exactly this (a side's missing XML must not red-X the PR; quorum logic, with self-test case 10 pinning that the quorum cannot mask a real regression), and more decisively **all three head shards passed** — new failures are a subset of head failures, which is empty, so a smaller base set can only make the check stricter, never laxer. Worth noting the same shard on #2148 passed on re-run, which is consistent with a runner-variance stall rather than a real signal.

On the change itself, the star-cap fix is the one that matters and the diagnosis is right. `mark_chat_read` creates a row for every chat ever opened — rooms included, which always report `unread: 0` — while unstar only nulled `starred_at` and no `DELETE` against `enterprise_portal_chat_state` existed anywhere. So a single total-row cap was consumed by activity the user cannot undo, and past it every star returned 409 "unstar some first" forever while unstarring freed nothing: a one-way door whose advice was false. Splitting into `MAX_CHAT_STATE_ROWS` (bounds abuse, since neither writer validates the chat exists and a 404 would be an enumeration oracle) and `MAX_STARRED_CHATS` (what a real user reaches) makes the recovery advice true, and deleting the row when unstarring leaves no read cursor closes the create-only growth path. The note that the old test only asserted `starred_at` went NULL, never that a subsequent star then succeeded, is exactly why the false claim survived two reviews — and the new test asserts the recovery.

The route-driven `markRead` gap is the right catch too: `bootstrap()` and the route watcher set `pendingSession` without it, so every reload of `/workspace/c/<sid>` re-badged the conversation being read — the commonest way into a chat.

ent#359 is cross-tracker, so nothing auto-promotes; its status stays for the release close.

PR #2150, opened 2026-08-13
fix(workspace): bound the reply poll to a turn's real life, and stop offering Retry when it is merely lost (#2133)

### LINKED ISSUE #2133: bug(workspace): awaitPersistedReply can poll for an hour when the in-flight marker is orphaned

dolho:
Closes #2133.

Two failures sit on opposite sides of one number, and **both were live on `dev`**.

## Too small — the one that costs money, and it was mine

The marker TTL and the client ceiling were both `300 + 60` — one turn timeout plus slack. But `run_resumable_turn` **re-runs the whole turn** when a resume finds no JSONL, so a legitimate turn can take *two* full timeouts.

At 360s the marker expired mid-retry, the client was told "nothing is running", and a live, already-billed turn was declared **not delivered with a Retry beside it** — the exact double-billing #2120 fixed, reappearing on precisely the cold-retry path that fix existed for.

I introduced this myself in #2138 when I shortened the TTL from an hour. The issue reports the hour; the hour was already gone, and what replaced it was worse in the other direction.

Both bounds are now `2 × PORTAL_TURN_TIMEOUT_SECONDS + 60`, derived from the constant the dispatch actually uses — pinned by a test, because a literal `300` at the call site is exactly how the two drift apart again.

## Too large — the reported symptom

An orphaned marker (a hard kill skips the `finally`; `except Exception` never catches `CancelledError`) left the loop polling a dead turn ~5,100 times per tab, with nothing in the UI saying the reply was not coming. The ceiling ends it, and the poll now widens with elapsed time — 700ms → 2s → 5s → 15s — cutting the long tail by an order of magnitude at the same total wait.

## Giving up is not a failure

Running out of budget while the marker still claims a turn means the turn probably **ran and was billed** and we merely lost sight of it. That path returns `lost` and renders the explanation **without a Retry** — re-sending is the one action guaranteed to be wrong. A genuine no-answer (server reports nothing running, no new reply) keeps its existing message and its Retry.

The budget travels on the 202 as `wait_budget_seconds` rather than living as a frontend constant, because the server owns the timeout. The client constant survives only as an older-backend fallback, sized identically and pinned by a test that the two agree.

## Also: the #2138 guard was wrong

`portalUndefinedCalls.spec.js` false-positived on this change. It stripped single-quoted strings *before* double-quoted ones, so an apostrophe inside `"we've"` opened a phantom string that swallowed the rest of the file and made every later identifier read as undefined. One alternation now, leftmost match wins. Verified by mutation that it still catches a genuinely undefined call.

## Verification

6 new backend tests, 254 frontend (18 files), 424 passed across the portal/session suites, build clean.

## AC coverage

- [x] The poll cannot run unbounded when the marker is stale
- [x] A legitimately long turn **including one cold retry** still resolves — this is the half the issue asked for and the half that was actually broken
- [x] Giving up does not offer Retry for a turn that may have run
- [x] Regression test naming this issue

### DIFF
--- a/src/backend/client_portal/models.py
+++ b/src/backend/client_portal/models.py
@@ -132,6 +132,10 @@ class PortalTurnStarted(BaseModel):
     """
     execution_id: str
     session_id: Optional[str] = None
+    # #2133: how long the client may wait before concluding it has lost track of
+    # the turn. Sent by the server because the server owns the timeout; a
+    # frontend constant would drift the next time it changes.
+    wait_budget_seconds: Optional[int] = None
 
 
 class PortalSessionSummary(BaseModel):
--- a/src/backend/client_portal/service.py
+++ b/src/backend/client_portal/service.py
@@ -1006,7 +1006,7 @@ async def portal_chat(agent_name: str, message: str, email: str,
             triggered_by="public",      # external-caller path; "Public" analytics bucket
             on_resume_failure=_on_resume_failure,
             source_user_email=email,
-            timeout_seconds=300,
+            timeout_seconds=PORTAL_TURN_TIMEOUT_SECONDS,
             images=images or None,      # referenced inbox images as vision input (#78)
             system_prompt=system_prompt,
             execution_id=execution_id,  # ent#286: pre-created row, so the client can already be watching
@@ -1113,11 +1113,25 @@ def _inflight_exec_key(execution_id: str) -> str:
     return f"portal_inflight_exec:{execution_id}"
 
 
-# The marker describes a turn in progress, so its TTL is the turn's own bound
-# plus slack — not an hour. The client now waits as long as the marker says a
-# turn is running, so an over-long TTL is a spinner that outlives the work: a
-# backend restart mid-turn would leave the composer disabled for the remainder.
-PORTAL_INFLIGHT_TTL_SECONDS = 300 + 60
+# The bound on ONE portal turn (the `timeout_seconds` passed to the turn below).
+PORTAL_TURN_TIMEOUT_SECONDS = 300
+
+# The bound on a turn's whole LIFE, which is what the marker and the client must
+# be sized against — #2133.
+#
+# `run_resumable_turn` can run the turn TWICE: a resume whose JSONL is gone
+# fails, and the cold retry re-runs the WHOLE thing. So the worst legitimate
+# case is two full timeouts, not one. Sizing either bound at a single timeout
+# reintroduces exactly what #2120 fixed — the marker expires, the client is told
+# "nothing is running", and a live, already-billed turn is declared not
+# delivered — and it does so precisely on the cold-retry path that fix existed
+# for.
+#
+# Still nowhere near the old hour: an hour-long marker is a spinner that
+# outlives the work, so a backend restart mid-turn would leave the composer
+# disabled for the rest of it.
+PORTAL_MAX_TURN_SECONDS = 2 * PORTAL_TURN_TIMEOUT_SECONDS + 60
+PORTAL_INFLIGHT_TTL_SECONDS = PORTAL_MAX_TURN_SECONDS
 
 
 def mark_turn_inflight(session_id: str, execution_id: str,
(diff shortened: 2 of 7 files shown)


dolho:
Pushed fixes for the review findings. Five were real, and three were in the part of this PR that was supposed to BE the fix.

**The bound was still too small.** I sized it as two turn timeouts on the assumption one attempt costs at most `timeout_seconds`. It does not: `execute_task` dispatches with `timeout_seconds + 10`, and the #678 reader-race auto-retry adds a whole second HTTP call capped at `_AUTO_RETRY_MAX_TIMEOUT_S` (300s) — **not** at the remaining budget, unlike the SUB-003 retry. One attempt reaches ~610s, a resume+cold pair ~1220s, against my 660s marker. Same failure as the one this PR fixes, one layer down. Now derived from the real terms, with a test that reads `_AUTO_RETRY_MAX_TIMEOUT_S` so raising either constant fails loudly.

**The idle give-up still offered Retry** — and it is the *reachable* path. With Redis down `mark_turn_inflight` no-ops, so `inFlightExecutionId` is null from the first poll and a perfectly healthy, already-billed turn got a Retry within seconds. Both give-up paths return `lost` now and render without one.

**The two clocks did not share an origin.** The marker starts ticking at dispatch; the client started counting after the stream broke, which on the orphaned-marker path is minutes later — so the ceiling could rarely fire at all. The client measures from a stamped dispatch instant now.

**The idle window was poll-count-based**, so my own backoff silently stretched "~5.6s" into up to 120s of spinner. Time-based now.

**Both of my new tests were tautologies** — literals declared inside the test, importing nothing, so changing the real constant left them green. That is exactly the drift they claimed to catch. Constants moved to `portalUtils`; verified by mutation that they now fail when it drifts.

Also corrected the ReDoS comment: dropping the template-placeholder branch did **not** "buy nothing" — it was what made a nested template literal strip as one unit. Accepted with the reason written down instead of claimed away.

**Not fixed, recorded instead:** `reattach()` still grants a fresh full budget to a turn already running for an unknown time. Bounding it needs the history response to carry the remaining budget — a behaviour change I would rather not make blind in a freeze week. Filing separately.

254 frontend tests, 211 backend across the portal suites, build clean.

vybe (approved):
Validated via /validate-pr: Closes #2133, all 4 required checks green, no secrets, no new config var, no raw palette classes.

The two-sided framing is right, and the derivation survived review better than the PR body did. Deriving the client budget from a server-sent `wait_budget_seconds` rather than a frontend constant is the correct ownership call, and keeping the client constant only as an older-backend fallback (pinned equal by test) is the right way to keep the fallback honest. `PORTAL_ATTEMPT_CEILING_SECONDS = timeout + 10 + 300` is the part I checked hardest: the #678 reader-race retry really does stack a second call on top of attempt 1 rather than sharing its budget (unlike the SUB-003 retry), so one attempt genuinely can exceed `timeout_seconds`, and two attempts is the real worst case on the cold-retry path. Returning `lost` without a Retry when the budget runs out while the marker still claims a turn is the correct asymmetry — re-sending is the one action guaranteed to be wrong for work that was probably billed.

Note the PR body is stale against the final commit: it says both bounds are `2 × PORTAL_TURN_TIMEOUT_SECONDS + 60` (660s), but the shipped value is `2 × PORTAL_ATTEMPT_CEILING_SECONDS + 60` = 1280s. The code is the better answer; the body just didn't follow the review round. Worth knowing that an orphaned marker now disables the composer for up to ~21 min — bounded and argued, but longer than the body implies.

Issue #612, opened 2026-04-30
bug: agent start response shows credentials_injection: failed for subscription-mode agents


vybe:
## Summary

When a subscription-mode agent is stopped and started, the start API response includes `\"credentials_injection\": \"failed\"` with error `\"No .credentials.enc file found in agent workspace\"`. This is misleading — subscription-mode agents (SUB-002) use `CLAUDE_CODE_OAUTH_TOKEN` env var injection, not file-based credential injection. The message implies a real failure when there is none, causing operators to take unnecessary remediation actions (e.g. re-assigning the subscription, which recreates the container and disrupts running state).

## Component

Backend / Agent Lifecycle / Subscription Management

## Priority

P3

## Error

```json
{
  "message": "Agent <name> started",
  "credentials_injection": "failed",
  "credentials_result": {
    "status": "failed",
    "error": "No .credentials.enc file found in agent workspace"
  }
}
```

## Root Cause

The agent start endpoint attempts file-based credential injection regardless of the agent's auth mode. When the agent is in subscription mode, no `.credentials.enc` file exists (by design — SUB-002 uses `CLAUDE_CODE_OAUTH_TOKEN` env var set at container creation time), so the injection step fails and surfaces that failure in the response. The response does not distinguish between a real failure (missing credentials on an API-key-mode agent) and an expected non-operation (subscription-mode agent where file injection is irrelevant).

## Reproduction Steps

1. Register a subscription and assign it to an agent
2. Verify the agent container has `CLAUDE_CODE_OAUTH_TOKEN` set and `ANTHROPIC_API_KEY` absent
3. Call `POST /api/agents/{name}/stop` then `POST /api/agents/{name}/start`
4. Observe: start response contains `credentials_injection: failed` despite the agent being correctly configured

## Suggested Fix

In the agent start endpoint, check the agent's auth mode before attempting file-based credential injection:

```python
# Before running file injection, check auth mode
agent_auth = db.get_agent_auth_mode(agent_name)

if agent_auth.auth_mode == "subscription":
    # Subscription agents use CLAUDE_CODE_OAUTH_TOKEN env var — no file injection needed
    credentials_result = {
        "status": "skipped",
        "reason": "subscription mode — credentials injected via CLAUDE_CODE_OAUTH_TOKEN env var"
    }
else:
    # Attempt file-based injection for API-key-mode agents
    credentials_result = inject_credentials_file(agent_name)

return {
    "message": f"Agent {agent_name} started",
[...]

PR #2151, opened 2026-08-13
fix(tests): scope the #612 module stubs to the load that needs them (#2140)

### LINKED ISSUE #2140: bug: test_inject_assigned_credentials leaks a stubbed services.agent_service.helpers into sys.modules, breaking runtime filtering in later tests

dolho:
Closes #2140.

`test_inject_assigned_credentials.py` installed four **real** module names into `sys.modules` and never restored them:

```
services.agent_service
services.agent_service.helpers
services.agent_service.read_only
services.agent_service.file_sharing
```

`services/compatibility/spec.py::applies_to_runtime` imports `is_claude_runtime` from that exact module **lazily**, so it picks up whatever is in `sys.modules` at call time. Every test running afterwards silently lost `claude_only` filtering, and `test_codex_runtime_omits_claude_only_checks` failed — **presenting as a real #1187 multi-runtime regression** rather than as pollution from two files away. `pytest-randomly` made it a latent flake: it fires only when file order puts the polluter first.

## The fix

The file **already** snapshotted and restored its other stubs, for precisely this reason. The agent_service aliases now join that set.

Deleting them instead was not an option, which is worth stating because it looks like the obvious move: `lifecycle.py:27` is an absolute `from services.agent_service.helpers import validate_base_image`, so the uniquely-named relative slots the file registers are not sufficient to load it.

## The guard lives in the polluting file, deliberately

Under random ordering there is no such thing as "a test that runs afterwards", so an assertion placed in the victim can't be relied on to fire. The guard asserts the four slots hold no test double once this module has imported — deterministic, at the only point where it can be.

**Mutation-tested.** Reintroducing the permanent install fails it with:

```
AssertionError: services.agent_service.helpers is still a test double after this
module finished importing. Restore it alongside the other stubs in
`_load_lifecycle` — see #2140.
```

## Verification

- The issue's reproduction command passes; so does the reverse order.
- **Full unit suite: 9755 passed**, 3 failed — all `test_pat_propagation_properties.py::TestKnownGaps`, pre-existing on `dev` and tracked as **#2017** (zero-line diff to that test or its service from this branch).

That full run is the check that mattered, not a formality: restoring the stubs means downstream tests now resolve the **real** `services.agent_service`, and the comment I removed claimed names had been added to the stub package specifically to stop later tests raising `ImportError`. Nothing depended on it.

`pytest-randomly` isn't installed in my venv, so I could not shuffle locally — CI is the check for ordering.

## AC coverage

- [x] `sys.modules` is restored after the module completes
- [x] The reproduction command passes
- [x] A guard asserts no stub is left, and fails when the leak returns

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -770,3 +770,57 @@ def cleanup_after_test(api_client: TrinityApiClient, resource_tracker: ResourceT
     yield
     if not request.config.getoption("--skip-cleanup"):
         resource_tracker.cleanup(api_client)
+
+
+# --- #2140: module stubs that cannot outlive the load they exist for ---------
+import contextlib as _contextlib
+
+
+@_contextlib.contextmanager
+def stubbed_modules(mapping):
+    """Install `mapping` into `sys.modules` for a block, then restore and verify.
+
+    Three unit files hand-load `services/agent_service/lifecycle.py`, which needs
+    the REAL `services.agent_service*` names present while it executes — line 27
+    is an absolute `from services.agent_service.helpers import validate_base_image`,
+    so the relative slots those files register are not enough. All three used to
+    install them PERMANENTLY.
+
+    That is not a tidiness point. `services/compatibility/spec.py`:
+
+        def applies_to_runtime(check, runtime):
+            if not check.claude_only:
+                return True
+            from services.agent_service.helpers import is_claude_runtime
+
+    resolves that import LAZILY, so it picks up whatever is in `sys.modules` at
+    call time. Every test running afterwards silently lost `claude_only`
+    filtering and `test_codex_runtime_omits_claude_only_checks` failed —
+    presenting as a real #1187 multi-runtime regression rather than as pollution
+    from another file. `pytest-randomly` made it a latent flake, not a constant.
+
+    Shared rather than inlined BECAUSE it was inlined: the first fix repaired
+    one copy and left two others producing the identical symptom.
+
+    The exit assertion uses IDENTITY, not `isinstance(..., Mock)`: the package
+    stub is a real `types.ModuleType` from `module_from_spec`, so a type check
+    waves through a re-leak of `services.agent_service` itself — the leak
+    `test_1310_auth_consolidation.py` documents as aborting its own collection.
+    A leak therefore fails at IMPORT, in the file that caused it, naming it.
+    """
+    snapshot = {name: sys.modules.get(name) for name in mapping}
+    sys.modules.update(mapping)
+    try:
+        yield
+    finally:
+        for name, original in snapshot.items():
+            if original is not None:
+                sys.modules[name] = original
+            else:
+                sys.modules.pop(name, None)
+        leaked = [name for name, stub in mapping.items() if sys.modules.get(name) is stub]
+        assert not leaked, (
+            f"module stubs outlived the load that needed them: {leaked}. "
+            "They are resolved lazily by services/compatibility/spec.py, so this "
+            "breaks claude_only filtering in whatever test runs next — see #2140."
+        )
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -3,6 +3,7 @@ Unit test conftest — overrides the parent conftest's autouse fixtures.
 
 These tests run without a backend connection (no Docker, no API).
 """
+import contextlib
 import importlib.util
 import os
 import sys
--- a/tests/unit/test_inject_assigned_credentials.py
+++ b/tests/unit/test_inject_assigned_credentials.py
@@ -132,20 +132,33 @@ def _load_lifecycle():
     read_only_mod = Mock(inject_read_only_hooks=AsyncMock(return_value={"success": True}))
     file_sharing_mod = Mock(check_public_folder_mount_matches=Mock(return_value=True))
 
-    # Names that downstream tests import from `services.agent_service`. Adding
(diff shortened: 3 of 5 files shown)


dolho:
**The review was right: the first version of this PR did not fix #2140.** Pushed the real fix.

Two more files install the identical permanent stubs — `test_start_agent_skip_inject.py:118` and `test_subscription_auto_switch_no_cred_import.py:234`. Repairing one of three copies leaves the symptom untouched, and the repro still failed on this branch:

```
pytest tests/unit/test_compatibility_checks.py tests/unit/test_start_agent_skip_inject.py -p no:randomly
FAILED test_codex_runtime_omits_claude_only_checks
```

My full-suite run reported green because file order happened to hide it — I used the flake as evidence against itself.

Now one shared `stubbed_modules` helper in `tests/conftest.py`, used by all three. Shared rather than inlined, because inlining is what produced three copies to miss.

Two corrections to my own guard, both from the review:

* it asserted a **process-global** property it does not control, so a sibling's collection-time stub failed it inside a file that had restored correctly. The assertion moved into the helper's exit path — it fails at import, in the file that leaked, naming it;
* `isinstance(..., Mock)` missed the name that matters: the package stub is a real `types.ModuleType`, so a re-leak of `services.agent_service` itself passed silently — the leak `test_1310_auth_consolidation.py` documents as aborting its own collection. It compares by identity now.

Also closed the `database` leak the review found (autouse `_reset` re-stamped a Mock permanently; `monkeypatch.setitem` now).

**Verification:** all three polluters alongside the victim → 113 passed. Mutation: removing the restore fails with the message naming the fix. Full unit suite → **9754 passed**, 3 failed (pre-existing #2017).

**Known residual, stated rather than left silent:** this file still installs `services` and `services.credential_encryption` at import without restoring them. Dormant because an earlier-sorting file imports the real `services` first, and unpicking it risks the lazy `credential_encryption` import inside `lifecycle` resolving to a module needing deploy-time config. Not worth the risk in a freeze week — filing separately.

vybe (approved):
Validated via /validate-pr: Closes #2140, test-only diff, all 4 required checks plus all 6 pytest shards (base+head × 3 seeds) green.

The root-cause reasoning holds up: `services/compatibility/spec.py::applies_to_runtime` resolves `is_claude_runtime` lazily, so a permanently-installed stub silently disables `claude_only` filtering for everything downstream — which is exactly why this presented as a #1187 regression instead of as pollution. Two things I checked specifically and agree with: (a) deleting the stubs isn't an option, since `lifecycle.py:27` is an absolute import of `services.agent_service.helpers`; (b) the exit assertion using identity rather than `isinstance(..., Mock)` is the right choice — the package stub is a real ModuleType, so a type check would wave the package-level re-leak straight through. Putting the guard in the polluting file is also correct under random ordering, where 'the test that runs afterwards' isn't a thing. Extracting the helper after the first attempt fixed one of three copies is the right correction.

`sys` is in scope at the append point (tests/conftest.py:133), so the helper resolves.

Nit, non-blocking: `import contextlib` added to `tests/unit/conftest.py` is unused — the shared helper landed in `tests/conftest.py` instead. Leftover from the earlier iteration; worth dropping.

PR #2152, opened 2026-08-13
feat(uploads): accept application/zip in chat file uploads

oleksandr-korin:
Fixes Abilityai/trinity-enterprise#391

## Summary
- Remove `application/zip` from `UNSUPPORTED_MIMES` in `upload_service.py` — zips now land in the agent's per-session uploads dir for the agent to extract itself
- Other archives (tar, gzip, rar), PDF, video, audio remain blocked
- Updated rejection message, unit tests, the web-chat-file-upload feature-flow doc, and the SLACK-FILES requirement entry (`docs/memory/requirements/public-access.md`)

## Security tradeoff
The platform never extracts the archive — it is written un-extracted to `/home/developer/uploads/{session_id}/`. This means zip-slip and decompression-bomb risk shifts to the agent container, which is the sole line of defense post-change. Mitigating factors: container filesystem isolation, per-session dirs cleaned after execution, per-file size caps unchanged (10 MB channel / 5 MB web), magic-byte MIME validation and filename sanitization still applied.

## Context
A client-side user needed to hand a release package (zip) to their agent via Slack; the platform bounced it. Already hot-patched on `trinity-client-marketingops` (cherry-pick `122232f6`); this PR supersedes it on merge.

## Test plan
- [x] `pytest tests/unit/test_slack_file_uploads.py tests/unit/test_web_file_upload.py` — all pass
- [x] `test_zip_supported` now imports the real `UNSUPPORTED_MIMES` from `services.upload_service` (no local reproduction drift)
- [x] Live: sent a .zip to an agent via Slack on the marketingops instance, confirmed accepted


vybe (changes requested):
Returning this one — it was pulled out of the merge train because of the base branch, but validation surfaced three more things worth fixing before it lands.

### Blocking

**1. Base branch is `main`.** Trinity's SDLC is `feature/*` → `dev` → `main`; only a release-cut PR targets `main` directly. Merging here would put the change on `main` ahead of everything sitting on `dev` and bypass the release train entirely. Please retarget to `dev` (Edit → change base), then rebase if needed.

**2. No issue link.** There's no `Fixes #N` / `Closes #N` anywhere in the title or body, so nothing traces this change to a tracked item and `issue-status-on-merge.yml` has nothing to promote. This is a `feat(...)`, so per the two-tracker routing it wants an issue in `abilityai/trinity-enterprise`, referenced here as `Fixes abilityai/trinity-enterprise#N` (the cross-tracker form is expected and fine — just remember that one closes manually at release).

### Please address

**3. The security relaxation isn't argued anywhere.** `UNSUPPORTED_MIMES` is a defense-in-depth control, and this removes one entry from it without stating what replaces it. The 10 MB / 5 MB caps bound the *upload*, not the *expansion* — a 10 MB zip can decompress to many GB inside the agent's workspace volume, and the PR explicitly hands extraction to the agent ("for the agent to extract itself"), which means whatever the agent runs is now the only thing standing between an uploaded archive and zip-slip or a decompression bomb. That may well be an acceptable trade for the client use case, but it should be the *stated* trade in the PR body (and probably a line in `docs/memory/requirements/`), not an inference from a one-line set edit. Right now the only reviewer-visible justification is "a client needed it".

**4. The test doesn't exercise the shipped code.** `tests/unit/test_slack_file_uploads.py::is_unsupported_mime` *reproduces* `UNSUPPORTED_MIMES` in the test file rather than importing it from `services.upload_service` — the docstring says so outright. So `test_zip_supported` passes against the copy, and would keep passing if the real constant in `upload_service.py` were reverted. That's pre-existing, not something this PR introduced, but this PR is the one whose correctness now depends on it. Importing the real constant would make the new test mean what it appears to mean.

Also worth noting the unchecked box in your own test plan — the live Slack round-trip is the one that proves the file actually lands in `/home/developer/uploads/<session>/`.

The change itself is small and reads fine; it's the framing around it that needs work. Happy to re-review once it targets `dev` and carries an issue.

oleksandr-korin:
All four points addressed:

1. **Base retargeted** `main` → `dev`
2. **Tracking issue**: Abilityai/trinity-enterprise#391, linked via `Fixes` in the PR body
3. **Security tradeoff** now stated explicitly in the PR body and recorded in `docs/memory/requirements/public-access.md` (SLACK-FILES entry): the platform never extracts the archive; zip-slip/decompression-bomb defense shifts to the agent container (isolated FS, per-session dirs cleaned post-execution, size caps + magic-byte validation unchanged)
4. **Test drift fixed** in 625b5aff: `test_slack_file_uploads.py` now imports the real `UNSUPPORTED_MIMES` from `services.upload_service` (same sys.path + stub pattern as `test_web_file_upload.py`) instead of a local copy. 54/54 pass, including standalone collection.

Ready for re-review.

vybe (approved):
Re-reviewed. All four items from the previous round are addressed:

1. **Base branch** — now `dev`. ✅
2. **Issue link** — `Fixes Abilityai/trinity-enterprise#391`, the cross-tracker form. ✅ (Reminder for whoever cuts the release: that one does **not** auto-close, and `issue-status-on-merge.yml` is same-repo too, so ent#391 needs a manual `status-in-dev` bump after this merges.)
3. **The security relaxation is now argued** — there is a "Security tradeoff" section stating plainly that the platform never extracts the archive, that zip-slip and decompression-bomb exposure therefore moves to the agent container, and what is *left* holding the line (container isolation, per-session dirs cleaned after execution, unchanged per-file caps, magic-byte MIME validation, filename sanitization). `docs/memory/requirements/public-access.md` carries the same trade with a dated line rather than a silent set edit. That is the reviewable form I was asking for. ✅
4. **The test exercises the shipped code** — `is_unsupported_mime` now applies the real `UNSUPPORTED_MIMES` imported from `services.upload_service`; the local reproduction is gone. Reverting the constant in `upload_service.py` would now fail `test_zip_supported`, which is the property that was missing. ✅ Live Slack round-trip also checked off.

**Two things I did on the branch rather than sending back**, both consequences of item 4 rather than new work:

- **`lint (sys.modules pollution check)` was genuinely red**, not flaky. Importing the real constant needs `upload_service`'s heavy deps stubbed *before* the module-level import, which `monkeypatch` cannot reach — so the two `setdefault` calls read as new bare `sys.modules` mutations under `tests/lint_sys_modules.py` (#762). Adopted the sanctioned `_STUBBED_MODULE_NAMES` + `_restore_sys_modules` pair. It is **module-scoped with the snapshot taken before the install**, deliberately: the stubs land during import, so a function-scoped fixture would snapshot the already-stubbed state and restore nothing. Verified both collection orders against `test_web_file_upload.py` (54 passed either way) so the `setdefault` compatibility your docstring calls out still holds.
- **Merged current `dev`** — the branch was 72 commits behind and its CI had run against stale `dev`. Clean merge; the net diff is still exactly your four files.

On the substance: I am satisfied the trade is stated, but I want it on the record that this PR's correctness now rests on the agent, not the platform. The caps bound the upload, not the expansion. That is an acceptable call for the client use case and it is now written down where the next person will find it — which was the whole of my objection.

Approving.

PR #2154, opened 2026-08-13
fix(ui): deep-link to every tab the viewer can see, by deleting the second list (#2153)

### LINKED ISSUE #2153: bug(ui): ?tab= deep-links to a2a/loops/playbooks/access/nevermined silently land on Overview
Split out of #2130, which named this and scoped it out ("Adjacent gap (not fixed here)"). Nothing was tracking it.

## The gap

`AgentDetail.vue` resolves `?tab=` against a hand-maintained list:

```js
const DEEP_LINK_TABS = ['overview', 'tasks', 'chat', 'reports', 'dashboard', 'logs', 'files',
  'schedules', 'credentials', 'skills', 'sharing', 'permissions', 'git', 'folders', 'settings', 'info']
```

It omits **`a2a`, `loops`, `playbooks`, `access`, `nevermined`** — all of which are real, rendered tabs. A link to any of them falls back to Overview with no error and no banner, on a stopped agent as well as a running one. That is the same silent-drop class as #1672.

## Why it is a list at all

The list duplicates knowledge that already exists one screen away. `visibleTabs` is built with the gates applied (`agent.can_share` adds `access` / `sharing` / `permissions` / `a2a`; others are pushed conditionally), so there are two sources of truth for "which tabs exist" and only one of them is maintained when a tab is added.

## Suggested fix

Resolve against `visibleTabs` instead of `DEEP_LINK_TABS`, keeping `TAB_ALIASES` for the renamed ids.

That deletes the list rather than extending it, so a future tab is deep-linkable the moment it is added — and it dissolves the open question #2130 raised. #2130 worried that adding the permission-gated ids "needs a non-owner blank-panel decision"; resolving against the tabs the viewer can actually see means a non-owner keeps today's Overview fallback for free, with no new decision and no blank panel.

Worth checking while there: `applyDeepLinkRouting` runs before `await loadAgent()` (correctly, per #2130), so `visibleTabs` may be empty on first pass. The resolution likely needs to re-run once the agent loads — which is a behaviour change to get right rather than assume, since re-applying later is exactly the "overrides the user's click" symptom #2130 fixed. Suggest resolving eagerly against the static superset and re-checking visibility only when the agent arrives, or deferring only the gated ids.

## Acceptance criteria

- [ ] `?tab=` works for every tab the viewer can see, including `a2a`, `loops`, `playbooks`, `access`, `nevermined`
- [ ] A tab the viewer cannot see still falls back to Overview (unchanged behaviour, no blank panel)
- [ ] Adding a new tab makes it deep-linkable with no second list to update
- [ ] A late re-resolution never overrides a tab the user clicked (#2130 regression)
- [ ] Regression test covering an owner-only id for both an owner and a non-owner

Related: #2130, #1672, #1108

dolho:
Closes #2153 (split out of #2130, which named this and scoped it out).

`?tab=` resolved against a hand-maintained `DEEP_LINK_TABS` array that omitted **`a2a`, `loops`, `playbooks`, `access`, `nevermined`** — real, rendered tabs whose links landed on Overview with no error and no banner. Same silent-drop class as #1672.

## Why it deletes the list instead of extending it

The array was a *second definition* of "which tabs exist". `visibleTabs` already computes that with the gates applied, so there were two sources of truth and only one was maintained when a tab was added — which is how five accumulated.

`buildTabs()` is now pure and takes the gating flags as arguments. `visibleTabs` calls it with the real state; `ALL_TAB_IDS` calls it with every flag permissive to get the superset. **A tab added to the builder is deep-linkable the moment it appears.**

This also dissolves the question #2130 deferred. It worried the gated ids needed "a non-owner blank-panel decision"; resolving against what the viewer can actually see means a non-owner keeps today's Overview fallback for free — no decision, no blank panel.

## The timing is the risky part

#2130 requires the landing to apply **before** the first await, so nothing permission-dependent is known yet. The link is applied eagerly against the superset, and visibility is reconciled once the agent loads.

But that reconcile is *another late write to `activeTab`* — precisely the bug #2130 fixed. So it acts **only while `activeTab` is still what the deep link set**: the moment the user clicks, their intent wins. It clears its state so it can fire at most once, and it runs in **both** hooks, since handling one is the #1672 class.

`brain` is deferred rather than applied eagerly: selecting it *navigates* to `/agents/:name/brain`, so applying it before capability is known would bounce a caller off the page and back.

## The test executes the real function

Because `buildTabs` is pure, the spec extracts it from the SFC and **runs** it, rather than scanning source the way the sibling #2130 spec has to (no `@vue/test-utils` here). So it sees a tab added later, instead of asserting today's shape.

Both properties mutation-tested:
- restoring a literal list → **fails**
- removing the "user already clicked" guard → **fails**

271 frontend tests (19 files), build clean.

## AC coverage

- [x] `?tab=` works for every tab the viewer can see, including the five omitted
- [x] A tab the viewer cannot see still falls back to Overview, no blank panel
- [x] Adding a new tab needs no second list updated
- [x] The late re-resolution never overrides a tab the user clicked
- [x] Regression test covering owner-only ids for both an owner and a non-owner

### DIFF
--- a/src/frontend/src/views/AgentDetail.vue
+++ b/src/frontend/src/views/AgentDetail.vue
@@ -401,7 +401,9 @@ const notFound = ref(false)  // #1914: the agent 404'd (missing OR inaccessible 
 const activeTab = ref('overview')  // #1107: Overview is the default landing tab
 // Tabs reachable via ?tab= deep-link (Timeline / EXEC-023 navigation).
 // Single source — referenced in onMounted + onActivated (#1107: dedupe + overview).
-const DEEP_LINK_TABS = ['overview', 'tasks', 'chat', 'reports', 'dashboard', 'logs', 'files', 'schedules', 'credentials', 'skills', 'sharing', 'permissions', 'git', 'folders', 'settings', 'info']
+// #2153: resolved against every id the page can render (see ALL_TAB_IDS, which
+// the tab builder derives), not a hand-maintained subset. The old list omitted
+// a2a, loops, playbooks, access and nevermined, so those links died silently.
 // Legacy ?tab= ids that moved/renamed — keep old deep-links working (#1108).
 // ent#358: `session` is no longer an alias — it REDIRECTS (see below). The
 // surface it named lives in the Workspace now, so resolving it to a local tab
@@ -411,8 +413,12 @@ const TAB_ALIASES = { guardrails: 'settings' }
 // Resolve a ?tab= value to a live tab id (applying aliases), or null if unknown.
 function resolveDeepLinkTab(requested) {
   const resolved = TAB_ALIASES[requested] || requested
-  return DEEP_LINK_TABS.includes(resolved) ? resolved : null
+  return ALL_TAB_IDS.includes(resolved) ? resolved : null
 }
+
+// What the deep link selected, so visibility can be reconciled once the agent
+// loads. Null once reconciled or once the user has moved.
+let deepLinkedTab = null
 // Apply the ?tab= deep-link landing. Called from BOTH onMounted AND onActivated:
 // AgentDetail is KeepAlive-cached (App.vue), so the common path — open agent
 // (caches it), click into an execution, "Continue as Chat" back — hits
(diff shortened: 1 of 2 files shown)


vybe (approved):
Validated via /validate-pr: Closes #2153 (closing keyword present, issue exists P3/type-bug). Two-file frontend fix, all 4 required checks green, no secrets/PII, no raw palette classes. Deleting the second tab list rather than extending it is the right call — the duplicate definition is why five tabs drifted out. The reconcile-after-load guard (acts only while activeTab is still what the deep link set) correctly avoids reintroducing the #2130 late-write bug, and it runs in both onMounted and onActivated per the #1672 class. Test executes the real buildTabs() rather than scanning source.

PR #2155, opened 2026-08-13
fix(agents): recreate no longer starts a stopped agent silently (#2092)

### LINKED ISSUE #2092: bug: recreate_container_with_updated_config unconditionally starts the container — running-state precondition is caller-enforced and undocumented
## Summary

`services/agent_service/lifecycle.py::recreate_container_with_updated_config` ends in `containers_run(...)`, so it always creates **and starts** the replacement container. It captures no pre-recreate run state, exposes no way to preserve it, and documents no precondition about it.

Every in-tree production caller works around this by pre-checking `container.status == "running"` and refusing otherwise. That precondition is real and correct — but it is enforced only by convention, repeated at each call site, and the function's docstring (which documents `env_overrides` and `full_capabilities` at length) never mentions it.

The result is a footgun for any new caller: recreating the container of a deliberately-stopped agent silently starts it, with no error and no log line saying the run state changed.

## Evidence

The precondition exists — but only at the call sites, never in the callee:

- `src/backend/services/agent_service/repo_binding.py:194`
  `if not container or getattr(container, "status", None) != "running":` → rejects with *"The agent must be running to bind it to a repository"*
- `src/backend/services/agent_mcp_key_service.py:641`
  same `status != "running"` check before `_recreate_with_env()` → `recreate_container_with_updated_config`
- `src/backend/services/agent_service/lifecycle.py` — the function itself: no `was_running` capture, no state parameter, no documented precondition, unconditional start.

## How this bit us (external ops tooling)

Our out-of-tree ops tooling rebuilds agent containers to adopt a new base image. It enumerated targets with `docker ps -a` and called this function per agent. Two agents that had been deliberately stopped ~8 days earlier were recreated and left **running**.

Worth flagging because it shaped the blast radius: `autonomy_enabled = 0` did **not** contain it. That flag gates cron fires and reminders (#1806) but not human-initiated inbound chat, so both agents' inbound surfaces — a channel binding and an enabled public chat link — became reachable again. No traffic actually arrived, so real impact was nil, but the containment was luck rather than design.

We have fixed our tooling (classify by pre-recreate state; skip stopped agents by default). Filing this because the platform-side contract is what made the mistake silent and easy to make, and the next caller will hit the same edge.

## Requests

1. **Document the precondition** in the docstring: this function always starts the replacement container; callers must verify the container is running first.
2. **Enforce it.** Cheapest robust option is a guard that raises unless the caller opts in — e.g. `require_running: bool = True` — since every current caller already satisfies it. That converts a silent side effect into a loud error for future callers, with no behavior change for existing ones.
3. *(Optional, larger)* A `preserve_run_state: bool` that re-stops the replacement when the original was stopped. This is what external tooling actually wants during base-image adoption waves, and it is awkward to implement correctly outside the function (the caller has to read state before, then race a stop after the container is already up and its agent server has begun booting).

## Environment

- Observed on v0.8.5 (`0.8.5+g1bbe2d1e`), enterprise edition, PostgreSQL backend.
- Function is unchanged on current `dev`.

## Not a security report

No vulnerability here — the exposure described was a state-management gap in our own tooling, not an access-control flaw. Filing publicly is appropriate.

dolho:
Closes #2092. All three requests, not just the two cheap ones.

`recreate_container_with_updated_config` always ended in `containers_run(...)`, so it created **and started** the replacement. Every in-tree caller pre-checked `container.status == "running"` and refused otherwise — a real and correct precondition, enforced only by convention, repeated at each call site, and absent from a docstring that documented `env_overrides` at length.

Recreating a deliberately-stopped agent therefore started it, with no error and no log line saying the run state had changed.

Worth restating from the report, because it shaped the blast radius: `autonomy_enabled = 0` did **not** contain it. That flag gates cron fires and reminders (#1806) but not human-initiated inbound chat, so a channel binding and an enabled public link became reachable again. No traffic arrived — the containment was luck.

## What changed

| Request | |
|---|---|
| **1. Document it** | The docstring now leads with the fact that this function starts the replacement, and says what a caller must do about it |
| **2. Enforce it** | `require_running=True` refuses with a `ValueError` naming the agent, its actual state, and both opt-outs. Checked **before any work**, so a refusal cannot half-recreate. Every current caller already enforces this for itself → no behaviour change for `repo_binding.py` or `agent_mcp_key_service.py` |
| **3. Preserve run state** | `preserve_run_state=True` stops the replacement when the original was stopped |

Both options are keyword-only — positional would let a caller pass one by accident, which is the opposite of making the choice explicit.

## The placement is the part to review

The restore lives in the recreate function, **not** in `_provision_folders_and_run_agent_container`.

I put it there first, and these tests caught it. That helper is the shared tail of **both** recreate and agent *creation*, where "the original was stopped" is meaningless — a `preserve_run_state` branch there would fire for callers that never asked. That is the same class of surprise this issue is about, moved one function down. There is now a test asserting it stays out.

A failed stop logs at **ERROR** and does not fail the recreate: the replacement exists and is healthy, and failing there leaves the caller worse off than the outcome they asked to avoid — but the agent is running against an explicit request, so it cannot be a warning.

## Honest note on scope

`preserve_run_state` still creates via `containers_run`, so the replacement boots briefly before being stopped. That is visible in its log. Avoiding it would mean duplicating the heavily-conditioned creation call to get a create-without-start path, which seemed a worse trade than one short boot — stated in the docstring rather than left for someone to discover.

## Adjacent edit

`test_1854_agent_mcp_key.py::test_07c` anchored on the literal `return await _provision_...`, which became an assignment (the resulting container is needed to restore state). Re-anchored on the **call** — same offset, same ordering property.

While verifying that didn't weaken it, I moved `env_overrides` up to the copy point and **the test still passed** — so that guard asserts "somewhere between the copy and the handoff", not "immediately before the handoff" as its docstring claims. Pre-existing, unrelated to this change, and left alone rather than widened into this PR. Flagging because #1854's whole hazard is an override being clobbered by the ~20 derived mutations in between, and the guard is looser than it reads.

## Verification

13 new tests; **344 passed** across the lifecycle / recreate / mcp-key / repo-binding / start-agent suites.

### DIFF
--- a/src/backend/services/agent_service/lifecycle.py
+++ b/src/backend/services/agent_service/lifecycle.py
@@ -22,7 +22,7 @@ from services.docker_service import (
 )
 from services.docker_utils import (
     container_stop, container_remove, container_start, container_reload,
-    volume_get, volume_create, containers_run, image_get
+    volume_get, volume_create, containers_run, image_get, container_stop
 )
 from services.agent_service.helpers import validate_base_image
 from services.agent_runtime_state import clear_agent_breakers
@@ -772,13 +772,44 @@ async def recreate_container_with_updated_config(
     *,
     full_capabilities: Optional[bool] = None,
     env_overrides: Optional[dict] = None,
+    require_running: bool = True,
+    preserve_run_state: bool = False,
 ):
     """
     Recreate an agent container with updated configuration.
     Handles shared folder mounts and API key settings.
     Preserves the agent's workspace volume and other configuration.
 
+    **This function STARTS the replacement container.** That was true before
+    #2092 too, but only the call sites knew it: each in-tree caller pre-checked
+    `container.status == "running"` and refused otherwise, while this function
+    captured no run state, offered no way to preserve it, and said nothing about
+    the precondition — which documented `env_overrides` at length.
+
+    Recreating a deliberately-stopped agent therefore started it, silently and
+    with no log line saying the run state had changed. Out-of-tree ops tooling
+    doing a base-image adoption wave hit exactly that: two agents stopped eight
+    days earlier came back up. Note what did NOT contain it — `autonomy_enabled
+    = 0` gates cron fires and reminders (#1806) but not human-initiated inbound
+    chat, so a channel binding and a public link became reachable again. No
+    traffic arrived; the containment was luck.
+
     Args:
+        require_running: refuse (ValueError) unless ``old_container`` is
+            running. Default True, which is what every existing caller already
+            enforces for itself — so this changes no behaviour and converts the
+            silent start into a loud error for the next caller. Pass False to
+            recreate a stopped agent deliberately.
+        preserve_run_state: leave the replacement STOPPED when the original was
+            stopped. What a base-image adoption wave actually wants, and awkward
+            to do from outside: the caller would have to read the state before,
+            then stop the container after it is already up and its agent server
+            has begun booting — racing its own recreate. Here the stop is issued
+            immediately after the handoff. The replacement is still created via
+            ``containers_run`` and so boots briefly before being stopped; that is
+            visible in the container's log, and is the honest cost of not
+            duplicating the (heavily conditioned) creation call.
+
         env_overrides: #1854 — env vars to force onto the replacement,
             applied LAST (immediately before the container handoff), so they win
             over every derived rule in between. The caller owns the value; this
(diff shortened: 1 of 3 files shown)


vybe (approved):
Validated via /validate-pr: Closes #2092, all 4 required checks green, no secrets, no new config var, no new top-level backend module (no Dockerfile COPY exposure).

Verified the safety claim independently: all three in-tree callers (`lifecycle.py:414`, `agent_mcp_key_service.py:795`, `repo_binding.py:462`) already pre-check `status == 'running'`, so `require_running=True` as default is a genuine no-op for existing behaviour and converts the silent start into a loud error for the next caller. Placing `_restore_stopped_state` in the recreate rather than `_provision_folders_and_run_agent_container` is correct — that helper is shared with creation, where 'the original was stopped' has no meaning — and there is a test pinning it there. 13 tests including the refuses-before-any-work ordering and the keyword-only guard.

Nit, non-blocking: `container_stop` is now imported twice in the `services.docker_utils` import (it was already first in the list). Harmless rebinding, but worth dropping in a follow-up.

PR #2156, opened 2026-08-13
fix(grid): un-break the info-tile chassis — clipped peg, dead Tiles menu, dead tokens (ent#325)

trinity-ability:
Reported from the Grid dashboard: both info tiles render with a badge sliced in half, and the **Tiles menu that would hide them cannot be opened** — so there is no way out of the state.

Four defects. One of them is a class this feature has already been fixed for once, which is the interesting part.

## 1 — `.it { overflow: hidden }` clipped its own peg

`.it` **is** the peg's containing block, so the 32px badge at `left: -16px` was amputated to a 16px sliver jammed against the tile's left border, glyph sliced down the middle. That is the artifact in the report.

`.it-body` already carries the clip the content actually needs, so the rule on `.it` was pure damage. `AgentTile`'s avatar overhangs its own tile the same way and has always rendered — which is what proves nothing else in the chassis clips, and that the overhang was intended. Measured after: the peg sits **14px** outside the tile edge (`left: -16px` from a padding box two 1px borders in), inside the 20px half-gap `gridLayout.GAP_X` budgets for exactly this.

## 2 — the Tiles menu could not be opened at all

`.gv-tilesctl` was never added to `onCanvasPointerDown`'s bail list, which every other floating control in the cluster is already in. So a pointer-down on the button started a canvas **pan** and called `setPointerCapture`: pointerup then targeted the canvas, `click` retargeted to the common ancestor, and the button's own `@click` never fired. The board lurched instead.

Added the bail, plus pointer-down-outside dismiss and Esc-to-close, matching the sibling `newDept` popover.

## 3 & 4 — four dead token references, all in this menu's CSS

`--gv-fg` (×3) and `--gv-tile-border` (×1) are defined **nowhere** in `src/`, so both resolved to guaranteed-invalid. The label ink and the button's hover ink silently inherited whatever the page handed them, and the separator above "Reset to defaults" drew at a permanently-live `rgba(0,0,0,.08)` — black at 8% on the dark panel, invisible in exactly one theme.

Now `--gv-text`, `--gv-btn-text` (the token the sibling `.gv-orgctl` buttons use, which this control is styled after) and `--gv-border`, fallbacks set to each token's light value.

## Why 3 and 4 survived a fix for their own class

`d2f91d80` swept exactly this — it renamed `--gv-fg` → `--gv-text` and `--gv-tile-border` → `--gv-border`, and shipped `tests/unit/gridTokens.spec.js` so it could not recur.

Both stale names survived in `FleetGrid.vue`, because the guard sweeps the **tile components** and `FleetGrid.vue` is only ever treated as the *definer* of the cascade — never as a consumer of it, which it also is, in its own chrome. Policing only the consumers you expect is how the second half of a rename outlives its own fix.

So the guard is widened rather than the four references merely patched:

- swept set is now `CASCADE_CONSUMERS` = the auto-discovered tiles + `FleetGrid.vue` + `AgentTile.vue`, the lattice's other occupant.
- `tokensInBlock` reads **every** block with the selector, not the first. The file opens a second `.fleet-canvas` / `:root.dark .fleet-canvas` pair further down for the org palette, so a first-match-only read left `--gv-wire` and `--gv-dept-0..7` entirely unguarded.
- a capture ending in `-` is recognised as the head of an interpolated name (`var(--gv-dept-${dept.slot})`), and that family gets its own test over `DEPT_SLOT_COUNT` in both themes rather than being silently dropped — a slot missing from one theme costs some departments their ribbon colour, and which ones depends on a string hash.

## Verification

**Reverted the fix to prove the guard bites** — with the four references restored, the widened guard fails naming both dead tokens:

```
AssertionError: components/FleetGrid.vue reads 2 token(s) defined nowhere in src/…
  expected [ '--gv-fg', '--gv-tile-border' ] to deeply equal []
```

**Driven against a running stack, not only built:**

| Check | Result |
|---|---|
| `.it` computed `overflow` | `visible` |
| peg geometry | 14px outside the tile edge, whole |
| `Tiles ▾` click | menu opens; `.gv-world` transform byte-identical before/after → no pan |
| uncheck a tile | leaves the board, persists `{"recent-failures":false}` |
| re-check | returns |
| outside pointer-down / Esc | both dismiss |

278 unit tests, `vite build` and `check:tokens` green. No colour literals added, so the raw-color ratchet is untouched.

## Scope

Frontend only, 3 files. No backend, no schema, no endpoint, no new dependency. Affects every info tile on the Grid — both shipped ones and the six epic ent#94 queues behind them.

### DIFF
--- a/src/frontend/src/components/FleetGrid.vue
+++ b/src/frontend/src/components/FleetGrid.vue
@@ -831,7 +831,17 @@ let panTX = 0
 let panTY = 0
 
 function onCanvasPointerDown(e) {
+  // The Tiles control is chrome, not canvas. Without this bail the pointerdown
+  // starts a pan and `setPointerCapture` retargets the resulting click at the
+  // canvas, so the button's own @click never fires and the menu cannot be
+  // opened at all — every other control in this cluster is already listed
+  // below, and ent#325 shipped without adding its own.
+  const inTilesCtl = !!e.target.closest('.gv-tilesctl')
+  // A pointer-down anywhere else dismisses the open menu, the way a menu
+  // should; the menu body itself stops propagation, so its own clicks are safe.
+  if (tilesMenuOpen.value && !inTilesCtl) tilesMenuOpen.value = false
   if (
+    inTilesCtl ||
     e.target.closest('.gv-tile') ||
     e.target.closest('.gv-zoomctl') ||
     e.target.closest('.gv-legend') ||
@@ -878,10 +888,11 @@ function tidyUp() {
   nextTick(fitView)
 }
 
-// Esc backs out of org modes (new-department popover, then assign mode).
+// Esc backs out of the open popover / org mode, innermost first.
 function onOrgKeydown(e) {
   if (e.key !== 'Escape') return
-  if (newDeptOpen.value) closeNewDept()
+  if (tilesMenuOpen.value) tilesMenuOpen.value = false
+  else if (newDeptOpen.value) closeNewDept()
   else if (assignMode.value) endAssignMode()
 }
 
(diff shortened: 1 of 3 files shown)


vybe (approved):
Validated via /validate-pr: frontend-only 3-file diff, all security scans clean, required checks green. The regression-diff red is a canceled base-leg (seed 99999 wrote no JUnit XML) — rerun kicked off; this diff contains zero Python and cannot move backend pytest results. Fix verified live (peg renders whole, Tiles menu opens/dismisses, no canvas pan) and the widened token guard was proven by revert.

PR #2158, opened 2026-08-13
fix(workspace): "Start a chat" on the agent page did nothing (ent#360)

dolho:
Found by dolho on the demo instance: on `/workspace/a/<agent>`, **Start a chat** set conversation state and navigated nowhere — so the agent page kept rendering (it is the first branch of the stage chain) and the chat never appeared. The button looked dead.

`newChatWithAgent` navigated only when the route named a session or a room:

```js
if (route.params.sessionId || route.params.roomId) router.push('/workspace')
```

The agent-page route sets neither.

## This is the third time, always the same way

| | |
|---|---|
| **#2128** | added `roomId`, after picking an agent from a room URL left the room *and its refusal panel* on screen |
| **ent#360** | added `/workspace/a/:agentName` and did not extend the list — this bug |

The condition was an enumeration of route params, so every new Workspace route silently re-breaks it. It now asks the actual question instead:

```js
if (route.path !== '/workspace') router.push('/workspace')
```

Every specific Workspace route is `/workspace/<something>`, so this covers a fourth one for free.

## The guard that should have caught it

#2128 shipped `F24`, which asserted the literal `route.params.roomId` — i.e. it pinned **the enumeration that just failed**, so it could only ever catch the one param someone had already thought of. It passed throughout ent#360.

Its *property* is right and still holds, so it now accepts either spelling and asserts the exit works from a non-bare route at all.

## Verification

Mutation-tested against **both** historical bugs:

- revert to `sessionId`-only (the pre-#2128 state) → **2 tests fail**
- restore `sessionId || roomId` (the exact ent#360 bug) → **2 tests fail**

291 frontend tests green (20 files). Deployed to the local instance and confirmed serving.

### DIFF
--- a/src/frontend/src/views/Portal.vue
+++ b/src/frontend/src/views/Portal.vue
@@ -375,10 +375,14 @@ function leaveRoomRoute() {
   router.push('/workspace')
 }
 
+// Leave ANY specific route, not an enumerated list of params — rationale in
+// newChatWithAgent below. Identical bug here: from /workspace/a/:agentName the
+// blank-chat control set state and navigated nowhere, so the agent page kept
+// rendering and the button looked just as dead.
 function startBlankChat() {
   unreachableAgent.value = null
   pendingSession.value = null; prefill.value = ''; convGen.value++
-  if (route.params.sessionId || route.params.roomId) router.push('/workspace')
+  if (route.path !== '/workspace') router.push('/workspace')
 }
 
 async function onPickerConfirm(agentNames) {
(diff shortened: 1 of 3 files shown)


vybe (approved):
Approved with two follow-up commits pushed to the branch.

The fix is right — asking the route-shape question instead of enumerating params is the correct generalisation, and the PR body's argument for it is sound. Two things it left on the table:

**The other two stage exits carried the identical enumeration.** `startBlankChat` and `onSignOut` both still read `route.params.sessionId || route.params.roomId`, so ent#360 broke all three at once, not one. From `/workspace/a/:agentName` the blank-chat control set state and navigated nowhere — the agent page kept rendering and the button looked exactly as dead as the one in the report. Both converted.

**F24 was relaxed in the direction that let that through.** Accepting EITHER the shape check or the old param spelling is what allowed the two unfixed sites to keep passing — and a guard that still accepts the enumeration can only catch the params someone already thought of, which is the property this PR argues against. With all three sites converted it now requires the shape check and rejects `route.params.*` outright. `portalLeaveSpecificRoute.spec.js` covers all three exits rather than only the reported one.

Mutation-tested: reverting `startBlankChat` alone fails 3 tests. 297 frontend tests green, build clean.

Issue #2163, opened 2026-08-13
perf(workspace): the roster's latency floor is its slowest agent — defer briefing hydration off the critical path


dolho:
Split out of #2159 AC5 ("latency measured and any per-agent N+1 identified — fix may land here or split out if large"). Measuring found a real shape problem; fixing it is a behaviour change, so it is filed rather than folded in.

## Measured

`GET /api/enterprise/client-portal/my-agents`, 12 agents / 3 running: **65–274ms**. Fine at this size, and unchanged by #2159 (which added `display_label` to the two existing selects rather than resolving it per agent).

## The shape

`get_roster` fans out `_agent_briefing` per agent and awaits `asyncio.gather(...)`. Per **running** agent that is:

- one `get_agent_container(...)` Docker lookup,
- `GET /api/template/info` on the agent,
- `GET /api/skills` on the agent (when a connector allow-list exists),

each with a **5s timeout**.

They run concurrently, so this is not N sequential round-trips. But `gather` waits for **all** of them, which means:

> **The roster's latency floor is the slowest agent, up to 5s — for every user, on every sign-in.**

One wedged or slow-booting agent makes the whole Workspace take five seconds to show anything, no matter how healthy the other 99 are. That is a fleet-scale property, not a fleet-size one: it does not need a large roster, just one bad agent.

## Why it wasn't fixed in #2159

The briefing is deliberately on the roster's critical path: #138 ships it at sign-in so the new-chat screen renders "with zero extra fetches", and ent#380's hint cards read from it. Deferring it means the hints pop in after the roster — a visible behaviour change to a shipped surface, and a design decision (skeleton the hint area? render without and fill in?) rather than a refactor.

#2159 fixed the symptom the user actually reported — the sidebar now shows a loading state instead of looking hung.

## Options

1. **Defer briefing off the roster**: return cards immediately, hydrate hints in a second call. Hints appear a beat later; needs a loading treatment in `PortalBriefing`.
2. **Bound the fan-out**: cap the per-agent timeout well below 5s (say 1.5s) so a wedged agent costs less. Cheapest, keeps the contract, strictly reduces the floor.
3. **Cache the briefing** per agent with a short TTL — the template description and playbook set change rarely.

(2) is the smallest real improvement and could land alone. (1) is the correct end state.

## Acceptance criteria

- [ ] A single unresponsive agent does not delay the roster for every other agent
[...]

PR #2164, opened 2026-08-13
fix(workspace): sidebar roster — loading state, top 5, label not slug (#2159)

### LINKED ISSUE #2159: fix(workspace): sidebar agent roster — no loading state, unbounded list, slug-as-title naming
## Summary

The Workspace sidebar's agent roster has three defects that compound on any real fleet: (1) the initial roster load is slow on a large fleet and renders **no loading state**, so the page looks hung rather than loading; (2) the agents block renders the **entire roster** as one long list; (3) each row shows the **slug as the title and the description as the subtitle**, instead of the human-facing display label.

## Context

Part of the Workspace polish pass following the ent#357–ent#361 wave (sidebar IA landed in ent#359 / PR #2149). The roster payload (`PortalAgentCard` in `src/backend/client_portal/models.py`) currently carries `name` (slug) + `description` but **no `display_label`**, so the frontend cannot render the business name the dashboard tiles show (ent#181/#1640) without a backend addition.

## Acceptance Criteria

- [ ] While the roster is loading, the sidebar shows an explicit loading state (per the design-system data-loading motion standard, `docs/memory/design-system-contract.md`) — a large fleet never looks like a hung page
- [ ] The agents block shows the **top 5** agents by default, with a "Show all (N)" affordance that expands the full roster in place (no scroll/state reset)
- [ ] Row title = the agent's **display label** (fallback to slug when unset); row subtitle = the **slug** — the description no longer appears in the row
- [ ] `PortalAgentCard` carries `display_label`, populated in the roster builder (`src/backend/client_portal/service.py`); older payloads without the field degrade to slug-only rendering
- [ ] Roster fetch latency on a large fleet is measured and any per-agent N+1 in the roster/briefing assembly is identified (fix may land here or split out if large)

## Technical Notes

- Frontend: `src/frontend/src/components/portal/PortalSidebar.vue` (agents block), `src/frontend/src/views/Portal.vue::bootstrap()` (roster fetch)
- Backend: `src/backend/client_portal/models.py::PortalAgentCard`, `service.py` roster + `_agent_briefing` (the briefing ladder is resolved best-effort at sign-in — a likely contributor to slow roster loads on large fleets; consider deferring briefing hydration off the roster's critical path)
- Display label source: `agent_ownership` display label (ent#181/#1640), same value the dashboard tiles render via `agentDisplayName()`

dolho:
Closes #2159. All five ACs; the fifth produced a finding filed separately as #2163.

| | |
|---|---|
| **Loading state** | Three skeleton rows while the roster loads, `aria-busy`, and the empty state suppressed so a slow fleet cannot flash *"No agents shared with you yet"* and then take it back |
| **Top 5** | Counted `Show all (N)`, expanding in place. **One** persistent toggle with `aria-expanded` — never two `v-if`-alternated buttons, which drop keyboard focus on collapse (#2101's lesson) |
| **Label, not slug** | Title = display label (ent#181/#1640, the same value dashboard tiles render), subtitle = slug. The description is gone from the row |
| **`display_label` on the card** | Optional, so an older payload degrades to slug-only |

## Two decisions worth a look

**The column rides the existing selects.** `get_display_label` is one query per agent; calling it in the roster loop would have added an N+1 to fix a row that renders the wrong title. Both roster queries already read `agent_ownership`, so the column is free. **Measured on the live instance before and after: ~65ms, unchanged.**

**NULL is passed through, not coalesced.** NULL means "render the slug" (ent#181). Resolving that server-side *and* keeping a fallback client-side is how the two ends drift into disagreeing about what unset means — so exactly one place decides, and it is the render site.

## AC5: measured, and it found something

`GET /my-agents` is **65–274ms** at 12 agents / 3 running. But the shape matters more than the number: `get_roster` fans `_agent_briefing` per agent and awaits `gather`, and each running agent costs one Docker lookup plus up to two agent HTTP calls at a **5s timeout**.

They run concurrently, so it is not N sequential trips — but `gather` waits for all of them, which means **the roster's latency floor is its slowest agent**. One wedged agent makes the Workspace take five seconds to show anything, for every user, on every sign-in. That needs one bad agent, not a big fleet.

Fixing it is a behaviour change to a shipped surface — #138 puts the briefing on the roster deliberately so the new-chat screen renders with zero extra fetches, and ent#380's hint cards read it — so it is **#2163** with three options costed, not folded in here. What this PR fixes is the symptom that was reported: the sidebar no longer looks hung.

## Verification

295 frontend tests (20 files), build clean. Deployed to the local instance: roster returns real labels (`Zeta Ops2`, `ent358 test agent`) with NULL falling back to slug, and 12 agents render as 5 + `Show all (12)`.

The label fallback is **executed**, not eyeballed — the test extracts `agentLabel` from the SFC and runs it, including the whitespace-only case.


dolho:
image

Loading state

image

vybe (approved):
Approved with one fix pushed to the branch, plus the dev merge.

The two decisions in the description are both right — riding the column on the existing selects instead of an N+1 `get_display_label` loop, and passing NULL through so exactly one place decides what unset means. Verified the column resolves against a real schema (the ent#357 roster tests exercise the query for real, not through a mock).

**The subtitle asked a different question than the title.** `agentLabel` treats a whitespace-only `display_label` as unset and falls back to the slug; the subtitle beside it tested raw truthiness of the same field. On exactly that input they disagree — the title renders the slug, and the subtitle renders the slug again, so the row prints it twice. That is the drift the PR's own NULL-passthrough rationale rules out, reintroduced one line down, and `portalRosterRow.spec.js` already pins the whitespace case for `agentLabel` alone. The condition now asks whether the title already IS the slug, so it is one decision read twice. Test added on the same input.

Also fixed the indentation of the `:loading-roster` binding on the mobile sidebar, and merged `dev` (#2158) — clean, no conflicts.

307 frontend tests green (21 files), build clean, 475 portal/a2a backend tests green.

PR #2165, opened 2026-08-13
fix(workspace): the agent page built the whole roster to render one agent (#2160)

### LINKED ISSUE #2160: fix(workspace): agent page initial load is slow — profile and parallelize build_page
## Summary

Opening an agent's page in the Workspace (ent#360, PR #2147) takes noticeably long before anything useful renders. The page payload is assembled server-side in one shot, so the client waits on the slowest sub-read before seeing anything.

## Context

`src/backend/client_portal/agent_page.py::build_page` composes health, windowed stats, operator asks, recent work, reports metadata and last-active into a single response. If those reads run sequentially (or any of them is an unindexed/wide query), first paint is bounded by their sum. Part of the Workspace polish pass following ent#357–ent#361.

## Acceptance Criteria

- [ ] Profile `build_page` on a realistic dataset and record where the time goes (health probe, stats window, asks, recent work, reports) — findings noted on this issue before the fix
- [ ] Independent sub-reads run concurrently (or below-the-fold sections load progressively) — the page shell with header/identity renders promptly instead of blocking on the full payload
- [ ] Any hot query gets an index or a bounded window; no unbounded scans on `schedule_executions`/`agent_activities` for one page load
- [ ] The page shows an honest loading state per section while data arrives (design-system data-loading motion standard) — never a blank/hung-looking page
- [ ] Repeat visits within a short window don't refetch identical windowed stats (consider the `${name}:${window}` cache convention from `stores/executions.js`)

## Technical Notes

- Backend: `src/backend/client_portal/agent_page.py` (`build_page`, `_health`, `_stats`, `_asks`, `_recent_work`, `reports`, `_last_active`)
- Frontend: `src/frontend/src/components/portal/PortalAgentPage.vue`
- `_health` reaches the live container — that read must never gate first paint for the rest of the page

dolho:
Closes #2160. Profiled first, as AC1 asks — [findings recorded on the issue](https://github.com/Abilityai/trinity/issues/2160#issuecomment-5281631066) before any change.

## `build_page` was not the bottleneck

Its six sub-reads total **32.7 ms** against a 1600-execution agent:

| health | stats(30d) | asks | recent work | last-active | reports | **sum** |
|---|---|---|---|---|---|---|
| 4.0 ms | 12.5 ms | 3.1 ms | 2.1 ms | 4.9 ms | 6.0 ms | **32.7 ms** |

So the issue's premise — sequential sub-reads, first paint bounded by their sum — is true and worth ~20 ms. **This PR does not parallelise them**; it wouldn't be worth the concurrency.

## The endpoint was ~150 ms, and the gap was one line I wrote in ent#360

```python
roster = await service.get_roster(email, include_owned=...)
card = next((a for a in roster.agents if a.name == agent_name), None)
```

Opening **one** agent's page built the **entire roster** — and `get_roster` fans `_agent_briefing` across every agent (Docker lookup + up to two agent HTTP calls each, 5 s timeout, awaited with `gather`).

So the page paid for N briefings to use one, and inherited the floor from #2163: **its load time was bounded by the slowest agent in the fleet, not the agent you opened.** One wedged agent ⇒ a five-second page for an unrelated agent.

Measured on the live instance: **12 briefings → 1**, endpoint 150 ms → ~110 ms. The local delta *understates* it — only 3 of 12 agents are running here. The property that matters is that the cost no longer grows with the fleet, and a test pins that at 200 agents.

## The original intent is kept

ent#360 projected the roster's card deliberately, so the page and the sidebar couldn't disagree about an agent's capabilities. Both paths still go through the same `_row_to_card` / `_roster_rows` helpers, and membership resolves by the same ent#357 rule (owned agents only for a platform session) rather than a second rule that could drift from the sidebar's.

## Also

- **AC5** — `${name}:${window}` cache with stale-while-revalidate, so 7d→30d→7d stops refetching an identical payload and never shows stale numbers. Cleared on agent change.
- **AC4** — section-shaped skeletons instead of one "Loading…" line.

## Two corrections to the issue, verified

- **`_health` does not reach the live container.** It reads the last persisted `agent_health_checks` row (4 ms), precisely so the page renders for a stopped agent (ent#360 AC #6). The AC asking it not gate first paint was already satisfied.
- **No unbounded scans.** Recent work and reports are `LIMIT 20`, last-active `LIMIT 1`, stats is the #1107 windowed accessor. **AC3 had nothing to fix.**

## Verification

7 new backend tests, 229 passed across the portal suites, 286 frontend, build clean.

The test pins the **count**, not a duration — a timing assertion passes on a two-agent fixture no matter how the code is written, which is exactly how this shipped. Mutation-tested: restoring `get_roster` in the router fails it.


vybe (approved):
Approved. The diagnosis is the valuable part here — profiling first and then reporting that `build_page` was *not* the bottleneck (32.7 ms of sub-reads against a ~150 ms endpoint), and correcting two of the issue's own premises rather than implementing them, is the right shape. `get_roster` + `next(...)` making one agent's page cost N briefings and inherit the fleet's slowest-agent floor is the real defect, and pinning the COUNT rather than a duration is what makes the test worth having.

Checked and found sound: `_roster_rows` and `_row_to_card` are faithful extractions; membership resolves by the same ent#357 rule as the sidebar; `_require_roster` still gates before the read. `isinstance(briefing, tuple)` is always true here (`_agent_briefing` has a catch-all returning `(None, [])`) but mirroring the roster's shape is fine. `pageCache` is per-instance — `<script setup>` — and the component is `:key`ed by agent name, so there is no cross-agent or cross-session path into it.

**The merge with #2159 needed care.** Both PRs edited the same block: #2159 added `display_label=r.get("display_label")` to the inline card construction, this PR replaced that construction with `_row_to_card`. The resolution git makes easiest — take the refactor — drops the field silently: it is Optional on the model so the payload still validates, and the frontend falls back to the slug, so the only symptom is #2159's bug quietly returning with every row rendering its slug again.

Resolved by carrying the field into `_row_to_card`, and added the coverage that was missing on both sides: `test_the_card_is_built_by_the_same_helper_as_the_roster` stays true with a field missing from that helper, and #2159's own tests are frontend-only. Two field-level tests now assert `display_label` survives the extraction on both paths and that an unset label stays None rather than being coalesced. Mutation-tested.

Worth noting for the earlier red: all six pytest legs pass in 11–12 min on this run, with more code in them than before. The 25-minute cancellation was runner variance, exactly as the workflow's own comment describes.

484 portal/a2a backend tests green, 307 frontend, build clean, regression diff clean.

PR #2166, opened 2026-08-13
fix(workspace): agent page UX — shared chart, live Start-a-chat, named work rows, contained asks (#2161)

### LINKED ISSUE #2161: fix(workspace): agent page UX — Overview chart reuse, dead Start-a-chat, contextless Recent work, asks need their own tab
## Summary

Four UX defects on the new Workspace agent page (ent#360, PR #2147):

1. **Header activity chart** is visually poor and spans the full content width. It should reuse the Agent Detail Overview chart treatment (`StackedBarChart.vue` with per-type activity breakdown) at an appropriate bounded width — not a bespoke full-bleed bar strip.
2. **"Start a chat" button does nothing.** The button emits `start-chat`, handled by `Portal.vue::onStartChatFromPage` — the chain is broken somewhere between emit and chat creation.
3. **"Recent work" rows carry no context** — the list renders only the execution type ("Scheduled run", "Chat") repeated, which tells the reader nothing. Rows should carry the same context the Agent Detail Overview recent-activity drill-in shows (schedule name / message summary / trigger).
4. **"Waiting on you" items sit inline on the Overview.** These are Operating Room queue items (`agent_page.py::_asks`); they should live in a **dedicated tab** with a count badge, mirroring how the tab strip already badges attention items — not push the rest of the Overview below the fold.

## Context

Part of the Workspace polish pass following ent#357–ent#361. The Agent Detail surface already solved chart, recent-activity and needs-attention presentation (#1107) — the agent page should borrow those patterns rather than re-derive them.

## Acceptance Criteria

- [ ] Header chart uses the shared Overview chart component/pattern (per-type stacked breakdown, bounded width, both themes) — no full-width bespoke bar strip
- [ ] "Start a chat" from the agent page opens a new chat with that agent (including the starter-prompt seeded variant from capability cards)
- [ ] Recent work rows show human-meaningful context (schedule name / message summary / trigger + outcome + duration), consistent with Agent Detail Overview
- [ ] Operator-queue asks move to their own tab with a pending-count badge; Overview no longer renders them inline; replying still deep-links to chat
- [ ] All four verified in light and dark themes (design-system contract)

## Technical Notes

- Frontend: `src/frontend/src/components/portal/PortalAgentPage.vue` (emits `start-chat`, `open-thread`, `open-menu`), `src/frontend/src/views/Portal.vue::onStartChatFromPage`
- Borrow from: `src/frontend/src/components/OverviewPanel.vue` + `StackedBarChart.vue` (#1107)
- Backend: `agent_page.py::_recent_work` currently returns type-level rows — extend the row shape with context fields (message/schedule name), keeping payload bounded
- Asks source: `agent_page.py::_asks` (operator queue)

vybe:
## Summary

Four UX defects on the Workspace agent page that ent#360 shipped two commits ago. **Two of the issue's own acceptance criteria are deliberately not implemented as written** — both overrides are recorded in the flow doc and the requirement so a later reader doesn't "fix" them back.

- **"Start a chat" did nothing** — and the button was wired correctly. The handler prepared the chat and then failed to leave the route, so `PortalAgentPage` (which renders *first* in the stage chain, keyed on `route.params.agentName`) kept the stage while the chat was set up invisibly behind it.
- **Recent work rows carried no context** — eight rows reading "Scheduled run". They now carry the schedule's **name**.
- **Asks were unbounded on the Overview** — now contained, still on the Overview.
- **The header chart was a bespoke full-bleed bar strip** — now the operator surface's own `StackedBarChart`, bounded.

## The stage escape now fails closed

The guard read `route.params.sessionId || route.params.roomId`. That list goes stale every time a stage route is added: **#2128** found `roomId` missing from guards written when `sessionId` was the only stage route, and **ent#360** then added `/workspace/a/:agentName` without revisiting them. Third occurrence.

So the question is inverted rather than extended. `shouldEscapeStage(path)` asks about route **shape** — anything that is not the workspace root is a stage to leave — and therefore fails closed: a fourth stage route needs no edit here and cannot silently re-break the button.

Two live call sites fixed (`newChatWithAgent`, `onSignOut` — the latter carried a room id, then an agent name, into the next session's address bar). A third, `startBlankChat`, was **deleted**: it had no callers at all, which makes patching it the plausible-looking fix that would have changed nothing.

## AC #3 — schedule name, not message summary (override)

The issue asked for a *message summary*; the answer is the **schedule name**, and the distinction is the decision. A schedule's `message` is a prompt, and this page serves **external clients**; its `name` is a short label an operator wrote to tell schedules apart, which is the job the row needs done. A per-viewer variant (show a viewer their *own* prompts) was weighed and rejected — no prompt text on this surface.

Honest cost: **AC #3 is met for schedule-backed rows only.** Chat/loop/reminder rows keep trigger, duration and time.

`_schedule_names` maps `id -> name` and reads nothing else off the `Schedule` model. Three properties, each load-bearing:
- **One query per page**, never `get_schedule` per row — an N+1 whose only symptom is latency, so the test asserts the **call count**.
- **Agent scoping is free** — the map holds only this agent's schedules, so a foreign or stale id misses by construction. No ownership check to forget.
- **It fails soft** — a schedules read that raises costs the labels, never the rows.

Misses are ordinary (the `__manual__` sentinel; soft-deleted schedules, whose executions outlive their name by up to 30 days) and render as a plain trigger label. The name attaches to any row whose id resolves, not only `triggered_by == "schedule"` — a webhook that fires a schedule *is* running that schedule.

## AC #4 — asks stay on the Overview (override)

A tab is somewhere you have to *go*, and an agent reaching you when no chat is open is the page's reason to exist. The defect was that asks were **unbounded**, not that they were present. Contained instead: compact cards, question clamped, first five with a counted toggle, beside recent work on wide screens.

- Asks stay **first in DOM order**, so the mobile stack keeps the priority.
- **No nested scroll region** — #2101 settled that on this surface (one scroll axis; a pane that scrolls inside a page that scrolls traps the gesture on touch).
- The badge reads `20+` at the cap: `MAX_ASKS` truncates server-side, and a bare "20" against 50 pending is a wrong number, not a rounded one.

## The chart

`BUCKET_COLORS` moves to `utils/executionBuckets.js` — a second inline copy is what drifts, and the *order* is a contract with the backend's `_BUCKET_ORDER`. `_stats` now **forwards** the analytics accessor's own `buckets` rather than letting the portal re-derive one.

The legend needed client-facing wording ("Tool call", not "MCP"). That is an **optional `labels` prop**, never a rename of the `buckets` array — those entries are the keys the chart indexes `by_type` with, so translating them in place makes every lookup miss and renders an **empty chart**, a silent failure that reads as "this agent did nothing this week". Empty and unavailable are now different sentences, and the window selector is hidden on tabs it does not drive.

## Changes

| Layer | File |
|---|---|
| Service | `client_portal/agent_page.py` — `_schedule_names`, forwarded `buckets` |
| Models | `client_portal/models.py` — `schedule_name`, `buckets` |
| UI | `portal/PortalAgentPage.vue` — chart card, contained asks, named rows |
| UI | `portal/portalUtils.js` — `shouldEscapeStage`, `PORTAL_BUCKET_LABELS` |
| UI | `utils/executionBuckets.js` **(new)** — shared with `OverviewPanel.vue` |
| UI | `StackedBarChart.vue` — additive `labels` prop · `views/Portal.vue` — shared escape |
| Docs | flow doc + requirement §5.11 (both overrides + the disclosure recorded) |

No new endpoint, no schema change, no migration, no auth change.

## Test Plan

- [x] `pytest tests/unit/test_2161_agent_page_ux.py tests/unit/test_ent360_workspace_agent_page.py` — **26 passed**
- [x] Portal/analytics backend slice — **330 passed**
- [x] `npx vitest run` — **310 passed** (20 files)
- [x] `npx vite build` — clean
- [x] Raw-color ratchet: no regression attributable to this change (the two baseline deltas reproduce identically at `HEAD`)

**On the full backend suite**: three runs give three disjoint failure sets — `dev` at HEAD → 2 errors in `test_subscription_auto_switch_pingpong.py`; this branch **minus** its new test file → 10 errors in that same file; this branch **with** it → zero there, and 3 unrelated failures elsewhere. All pass in isolation. That is the `sys.modules` collection-order class already recorded in `learnings.md` (2026-08-12 #2114, which names that very file); adding a test file perturbs collection order. Everything this PR touches is green in every run. Captured as a durable entry.

**Not visually verified**: light/dark was checked mechanically — every color introduced is theme-paired or uses the file's existing dual-safe convention, and the chart component already ships in both themes on Agent Detail — but I did not render it. Worth a look before merge.

Fixes abilityai/trinity#2161


vybe:
## Security pass — 3 low findings, all fixed in 3b91949c

No critical or high findings. The `schedule_name` disclosure boundary, the agent scoping of the id→name map, XSS (interpolation, no `v-html`), the `response_model` projection, the untouched `asks` filter, and the absence of any new auth surface were all verified clean.

Two of the three were cases where the code was correct and **my prose was wrong**, which is the worse half — a comment that overstates a guarantee is what the next reader relies on instead of checking.

**L1 — don't load prompts to render a label.** `_schedule_names` resolved through `db.list_agent_schedules`, i.e. `select(agent_schedules)` — whole `Schedule` models carrying `message` and `validation_prompt`. The projection to `{id: name}` was correct, but it made this module's own principle ("a field that never leaves the service cannot be leaked by a later edit") a *review invariant*: `{s.id: s}`, or adding `description` for a subtitle, puts the agent's prompts one edit from a client's page. Now a projected `SELECT id, name` accessor — structural, and it stops pulling N prompts per page load. A test pins that the prompt-carrying accessor is not called.

**L2 — "operator-authored" was wrong**, and I'd asserted it in the service docstring, the response model, the flow doc and the requirement. `POST /api/agents/{name}/schedules` is `AuthorizedAgent` and the MCP `create_agent_schedule` tool exists, so an agent-scoped key — a prompt-injected agent — can author the text that renders on an external client's page, and `ScheduleCreate.name` has no length bound. Bounded impact (Vue escapes it; `asks.title`/`question` already cross by the same precedent), so the fix is honesty plus a cap at the **service** rather than the template, where a later list view would miss it.

**L3 — sign-out didn't clear `?agent=`, and my new comment claimed it did.** `shouldEscapeStage('/workspace')` is false while `?agent=` is a live landing parameter that `bootstrap()` re-reads after every sign-in — so signing out at `/workspace?agent=acme-billing` and handing the browser on makes the *next* person's first screen "You don't have access to acme-billing". Pre-existing (the old param enumeration missed it too — the same stale-list class one level over) and bounded, since the name is already in the address bar. The query is now part of the shape the predicate tests, which also stops `newChatWithAgent` leaving a stale `?agent=A` when switching to B.

**Re-verified**: 27 backend tests for #2161, 776 in the schedule/portal slice, 314 frontend, build clean.

Also confirmed: the unrelated `src/backend/enterprise` submodule bump in the working tree is **not** in this PR.

vybe:
## Rebased onto `dev` — and the base branch was wrong

Two corrections since the original push:

**The PR targeted `main`.** `gh pr create` defaulted to the repo's default branch, so this read as `CONFLICTING` against a hundred release commits. Retargeted to **`dev`** — now `MERGEABLE`, 2 commits, 18 files.

**Rebased onto `dev` @ `c1cc22bc`**, which had moved three commits, two of them on this exact surface.

### Overlap with #2158 — same defect, fixed concurrently

`a682b074` (#2158) fixed "Start a chat" independently and reached the **same conclusion**: stop enumerating route params, ask about route shape. That rule is kept as-is. What this PR adds on top, and why the inline form couldn't cover it:

- **It's a pure function** (`shouldEscapeStage` in `portalUtils`). This project has no component-mount harness, so an inline closure over `route` can only be pinned by scanning source for a spelling; extracted, it's tested by *calling* it — including against `/workspace/x/whatever`, a route nobody has written yet.
- **It covers the query.** `/workspace?agent=X` passes the path check while still carrying X. `?agent=` is the ent#358 landing spot, re-read by `bootstrap()` after every sign-in — so signing out there and handing the browser on makes the *next* person's first screen "You don't have access to X". Found by this branch's security pass; #2158's inline check doesn't catch it.

### `startBlankChat` is deleted, not fixed

#2158 converted it, describing it as a control that "looked just as dead as the one in the report". It has **zero callers** — verified on `dev` (`git grep` matches only its own definition), and the cause is traceable: ent#361 (`8e5157f1`) renamed it out of the `@new-chat` binding and gave that event to the picker. So it's been orphaned since, and a dead function carrying a "fixed" comment reads as live to the next person. Deleted here, with a test pinning its absence and the orphaning commit named in the comment. **Happy to restore it if it's meant to be wired up** — that's a call for @dolho.

### Test guards updated rather than relaxed

Both guards from #2158 asserted the *inline spelling*, so they failed once the rule moved behind a name. Both now assert the new structure and keep their original property:

- `portalLeaveSpecificRoute.spec.js` — exits must delegate to `escapeStage()`; the shape check and the bare-route push are asserted once, against the single place that now answers for all of them.
- `workspaceRoomsGate.spec.js` F24 — merged both sides: requires the shared call **and** keeps #2158's stronger anti-enumeration assertion (`not.toMatch(/route\.params\.(sessionId|roomId|agentName)/)`), which mine lacked. An enumeration creeping back *beside* the call is something a delegation-only check wouldn't notice.

### #2160 merged cleanly

`c1cc22bc` touched `PortalAgentPage.vue` too (skeleton loaders, `pageCache`, dimmed stat placeholders). Auto-merged with no conflict and verified intact alongside this PR's chart card and asks split — including that `pageCache.clear()` and `allAsks = false` both survive in the per-agent watcher.

**Re-verified post-rebase**: 332 frontend tests (22 files), 785 backend in the portal/schedule slice, 27 for #2161, build clean.

trinity-ability (approved):
Reviewed as second party (PR authored by `vybe`).

All 26 checks green, including the four required by `dev` branch protection — CodeQL (python + javascript-typescript), Schema Parity Gate, and container-security `verify-non-root` — plus `backend-unit-test`'s `regression diff`, which compares base against head across three pytest seeds and reported no new failures.

Two blockers were found and fixed during validation rather than waved through:

- **CodeQL high-severity alert** (`js/incomplete-multi-character-sanitization`) on the comment-stripping test helper. The obvious loop-the-regex patch was verified NOT to fix it (`<!<!---->-->` → `<!-->`, residue remains, loop stops), so it was replaced with an index scan that provably converges, and extracted so the two pre-existing copies stop diverging.
- **`pytest (head, seed 99999)` cancelled** at the 25-minute job budget — the documented #2019 stall class, whose own workflow comment records it hitting all three seeds and both sides including plain `dev`. Re-run: success.

Docs, requirements §5.11 and a learnings entry accompany the change; both deliberate AC deviations (no prompt text in Recent work, asks contained on Overview rather than a new tab) are recorded in-repo so they read as decisions rather than omissions.

PR #2167, opened 2026-08-13
feat(schedules): client-side cron validation + invalid-schedule warning icon (#925)

### LINKED ISSUE #925: Client-side cron validation + warning icon for invalid schedules
## Summary

Add client-side cron expression validation to the schedule create/edit form, and surface a warning icon next to schedules in the list whose stored `cron_expression` no longer parses. Today the backend rejects invalid cron via `croniter` (`src/backend/routers/schedules.py:232`, `:287`), but the frontend has no preflight check and no visible signal on the schedule list when something is wrong — users only find out by submitting the form or watching schedules silently fail to fire.

## Context

The schedule form in `src/frontend/src/components/SchedulesPanel.vue` exposes a raw cron text field with preset buttons but no validation feedback. A malformed expression round-trips to the backend, surfaces as a 400, and the user has to interpret the error message. Worse, if a previously-valid cron becomes invalid (e.g., manual DB edit, schema change, or a parser-behavior shift), the row in the schedule list looks identical to a healthy one.

## Acceptance Criteria

- [ ] Schedule form validates `cron_expression` on input/blur using a JS cron parser (e.g., `cron-parser` or `cronstrue` + a parse attempt), with inline error text matching the backend message style
- [ ] Submit button is disabled while the expression is invalid
- [ ] Each row in the schedules list shows a warning icon (with tooltip "Invalid cron expression") when `cron_expression` fails to parse client-side
- [ ] Backend response for invalid cron is preserved unchanged (no contract change)
- [ ] Presets remain valid and do not trigger the warning state

## Technical Notes

- Frontend: `src/frontend/src/components/SchedulesPanel.vue` — add validation inside `formData` watcher and per-row computed flag for the list rendering near line 282
- Backend: validation already in place at `src/backend/routers/schedules.py:232,287` via `croniter` — no changes required
- Library choice: align client parser with `croniter` syntax (5-field, no seconds) to avoid client-server disagreement on edge cases (e.g., `*/30 * * * *`)
- Out of scope: server-side migration of existing invalid rows; warning icon is purely a UI affordance

trinity-ability:
Fixes #925

The schedule form validated cron expressions only at save time — the backend's 400 was the first feedback — and a stored-invalid row sat in the list indistinguishable from a healthy one. This adds client-side pre-validation and marks invalid stored rows.

## What changed

- **`src/frontend/src/utils/cronValidation.js`** (new, pure, **zero npm deps**) — a hand-rolled mirror of the backend's *exact* acceptance grammar: `services/schedule_validation.py::validate_cron_expression` (strict 5-field split + `_dow_to_apscheduler` ported verbatim, branch order included) plus APScheduler 3.11 field rules. Total and **fail-open**: an internal port bug returns `{valid: true}` + `console.error`, so it can never brick the panel or block a save. The backend 400 remains the enforcement authority — this layer is advisory UX only.
- **`SchedulesPanel.vue`** — inline error, submit gating, per-row warning icon:
  - The format-hint line **doubles as the reserved error slot** (`min-h-8`), so the modal never jumps when validation kicks in. Every message shape was measured ≤2 lines at the modal's real 400px width.
  - Submit is disabled **only when the cron is non-empty AND invalid** — an empty cron keeps the native `required` bubble, so disabling it wouldn't kill the browser's constraint-validation path for every other field in the form.
  - While **creating**, the error appears only after first blur (no mid-typing flash); while **editing**, an invalid stored cron (incl. empty) shows unconditionally — a disabled Update button must never be unexplained.
  - List rows get a warning triangle *inside* the cron chip (so flex-wrap can't detach it), tooltip and `aria-label` exactly `Invalid cron expression`. Guarded with `=== false` so a map miss can never false-warn.
  - Presets now come from the exported `CRON_PRESETS` (byte-identical labels/expressions) so "presets never warn" is tested against the shipped list rather than a mirrored literal.
- **Shared drift alarm** — `tests/fixtures/cron-grammar-cases.json`: 111 rows whose verdicts were **probed against the live backend validator, never hand-typed** (57 ACCEPT / 54 REJECT). Asserted row-for-row by **both** `tests/unit/test_925_cron_grammar_fixture.py` (re-proves every row against `validate_cron_expression` in the backend CI env) and `src/frontend/tests/unit/cronValidation.spec.js` (the client mirror). An APScheduler bump that moves the grammar fails CI **loudly** instead of silently desynchronizing the mirror.
- **Docs** — requirements §10.17, a feature-flow section in `scheduling.md`, and the feature-flows index row.

## Why no npm cron library

**The issue's technical notes were stale.** They assumed the backend validates via `croniter`. It does not — validation is `services/schedule_validation.py`: a 5-field split, the `_dow_to_apscheduler` translation, then APScheduler's `CronTrigger`. Any npm library therefore imports a **foreign grammar**, with ≥8 proven verdict disagreements against what the server actually accepts: `@daily`, 6-field expressions, dow `0-6`, `last`, `7/2`, Quartz `L`/`#`, the step-span rule, and `5/2`.

A library would have produced false warnings on expressions the server accepts — the damaging direction. This is the `learnings.md` #1472 lesson ("validate with the parser that registers") applied one level up: the client validates with a *pinned mirror* of that parser, and the fixture is what keeps the mirror honest.

Several upstream behaviours look like bugs and are **deliberately preserved** — each is probe-verified server-ACCEPT and pinned by a fixture row, with `parity outranks tidiness` stated at the call site:
- Name expressions are start-anchored **prefix** matches with no `$` anchor, so `MON/999`, `jan/0`, `lastx` all ACCEPT (trailing garbage silently dropped).
- `_dow_to_apscheduler` comma-branch asymmetry: `0-6,1` ACCEPTs while bare `0-6` REJECTs.
- Falsy-zero range end: APScheduler computes the step span as `(self.last or MAX)`, so `0-0/2` ACCEPTs while `5-5/1` REJECTs. Mirrored with JS `||`, **not** `??`, on purpose — `??` would reject server-valid crons.

## Evidence

| Check | Result |
|---|---|
| `python3 -m pytest tests/unit/test_925_cron_grammar_fixture.py -q` | **113 passed** (111 fixture rows + 2 meta-tests) |
| `cd src/frontend && npm run test:unit` | **422 passed, 20 files** |
| `/verify-local` — unit evidence | **PASS** — run `…/runs/20260813-154656/` |
| `/verify-local` — stack evidence | **PASS** — run `…/runs/20260813-160210/`: build + import-smoke, boot + health, integration **70 passed / 13 skipped / 2 registry-deselected** |
| Fixture re-probed against the **built image**'s apscheduler | **0 mismatches** (image pins 3.11.0; the venv resolved 3.11.3 — a floating floor, so the image is the authority) |
| Raw-color ratchet | **shrinks** — `SchedulesPanel.vue` gray 315 → 300, nongray unchanged at 21, hardcoded 0 |

Both test suites read the *same* fixture file, and the vitest spec carries an explicit anti-pruning guard, so the contract cannot be quietly weakened from either side.

### Re-verified against the current `dev` tip

`dev` advanced 5 commits after this branch's base (`76d842da` → **`f5f2461a`**: A2A outbound #736, plus workspace agent-page fixes #2158/#2159/#2160/#2161). Because `SchedulesPanel.vue` is shared surface and those commits touch frontend, the merge was **verified rather than assumed** — merged into current `dev` in a throwaway worktree and the suites re-run against the *merged* tree, not just this branch:

| Check (merged tree, base `f5f2461a`) | Result |
|---|---|
| Merge | **clean, no conflicts** — contributes exactly the same 8 files, +1283/-7 |
| `pytest tests/unit/test_925_cron_grammar_fixture.py -q` | **113 passed** |
| `npm run test:unit` | **468 passed, 23 files** — this branch's 422/20 *plus* dev's newly-landed workspace tests (+3 files, +46 tests), all green together |
| `npx vitest run tests/unit/cronValidation.spec.js` | **136 passed** |
| Raw-color ratchet on merged tree | `SchedulesPanel.vue` gray still **300** |

So neither side regresses the other: our changes don't break the newly-landed workspace tests, and those don't perturb the cron mirror. No rebase or retarget was performed — GitHub compares against the current `dev` tip on its own.

## Pre-existing `dev` reds — disclosed, not hidden

`tests/unit/test_pat_propagation_properties.py::TestKnownGaps::test_a_backslash_in_the_token_does_not_raise` fails as **XPASS(strict)** on all 3 parametrized cases (`tok\1`, `tok\g<1>`, `tok\slash`) on pristine `origin/dev` @76d842da — reproduced independently in a throwaway worktree by two agents.

Cause: #2017 fixed the bug (`_patch_env_github_pat` now passes `lambda _: new_line` to `re.sub` instead of the replacement string), but the paired `xfail(strict=True)` from #2018 was never retired — so the marker correctly flipped to XPASS, which is a failure. The sibling test in that same class documents this exact lifecycle and *was* retired to a plain regression test; this one was missed. It is invisible to CI because dev's required checks do not run `tests/unit`.

**Not fixed here** — an unrelated backend test file has no business in a frontend PR. This branch touches **zero** `src/backend/` files, so it cannot be implicated. Per the operator's explicit call, **no issue was filed** for it; it is recorded here.

## Taste decisions surfaced for the human gate

- **T1 — fixture + backend pytest kept in this PR.** The mirror's whole liability is drift, and the backend-side assertion *is* the alarm; splitting it out would ship the mirror without its guard. Purely additive — no backend runtime change.
- **T2 — the pre-existing `initialMessage` / `editingSchedule` adjacency left alone.** That watcher path opens the form without going through `closeForm()`, so it needed a one-line `cronTouched` reset. Refactoring the adjacency itself would be scope drift in a shared file.
- **T3 — concise error strings rather than verbatim backend bytes.** The AC asks for *style*, and the messages are pinned loosely by substring (field name, "5 fields", example) rather than byte-for-byte, so wording stays editable without turning every copy tweak into a test failure.

- An optional `GET /api/validate-cron` thin endpoint — the only change that would fully retire the mirror-maintenance liability.
- Playwright e2e covering form gating + the row icon.
- The `initialMessage` watcher adjacency (T2 above).
- `raw-color-baseline.json` is stale tree-wide (~65 files above baseline on dev, pre-existing) — needs its own refresh PR.
- The stale `xfail(strict=True)` described above.

## Note on the test catalog

`.claude/agents/test-runner.md` needs a catalog row for `tests/unit/test_925_cron_grammar_fixture.py`, but it lives in the private `trinity-dev` submodule — a separate repo. It lands separately and is **not** part of this PR (no submodule pointer bump here).

## Scope

No backend runtime change (the 400 contract is byte-preserved), no schema change, no migration, no new endpoint, no new dependency, no server-side repair of stored-invalid rows.


vybe (approved):
Validated via /validate-pr. Frontend-only validation with a shared tests/fixtures/cron-grammar-cases.json driving BOTH the Vitest spec and a Python fixture-parity test, so the client-side grammar cannot drift from the server's accepted set — the right shape for a validator that duplicates a backend rule. Docs (scheduling flow + requirements) updated; closing keywords for #925 and #2158 both present. Resolved a feature-flows.md changelog-row conflict against dev (chronological placement preserved). All checks green incl. frontend build + e2e.

PR #2168, opened 2026-08-13
fix(operator-queue): budget the agent-influenceable platform alert emitter (#1677)

### LINKED ISSUE #1677: bug: platform operator-queue alert emitters bypass #1632 ingestion caps — skill-not-found is agent-triggerable (flood residual, gates pull default-ON)
## Summary

#1632 caps the **agent-authored file** create seam (`~/.trinity/operator-queue.json` → `OperatorQueueSyncService._sync_agent`): per-agent pending-DEPTH cap, per-agent+fleet RATE cap, field clamp, reserved-id guard. By design (#1632 AC#3) the **platform-internal** create callers bypass `_sync_agent` and create directly via `db.create_operator_queue_item`, so they are **exempt from every cap**. That exemption is correct for genuinely platform-driven alerts — but one of those "platform" emitters is actually **agent-influenceable**, which re-opens the exact flooding/operator-fatigue vector #1632 set out to close.

Surfaced by the adversarial `/cso --diff` pass while reviewing the #1632 branch. This is a **completeness residual** on the #1081 pull-mode default-ON gate list, not a defect in the #1632 fix itself.

## The exempt callers (all bypass the caps)

| Caller | id prefix | Agent-influenceable? |
|---|---|---|
| `lease_reaper_service._create_park_item` | `poison-` | No (platform-only) |
| `agent_client` circuit-breaker-dormant | `cb-dormant-` | No |
| `sync_health_service` (sync_failing / git_bloat) | `sync-failing-` / `git-bloat-` | No |
| `validation_service._notify_operator_on_failure` | `val_` | No |
| **`task_execution_service._alert_skill_not_found`** | **`skill-not-found-`** | **YES** |

## Exploit (skill-not-found)

`_alert_skill_not_found` (`src/backend/services/task_execution_service.py:341`) creates directly via `db.create_operator_queue_item` — uncapped. It fires when an execution whose `triggered_by ∈ _AUTONOMOUS_TRIGGERS` runs a slash command that doesn't resolve to an installed skill. `_AUTONOMOUS_TRIGGERS` **includes `loop`** (`task_execution_service.py:257`), and dedup is **per-command only** (`(it.get("context") or {}).get("command") == command`), with a timestamped id (`skill-not-found-{agent}-{now}`), so **distinct** commands each mint a new `priority:"high"` pending row.

A compromised / prompt-injected agent with loop capability:
```
run_agent_loop(message="/nonexistent-{{run}}", max_runs=100)
```
`{{run}}` templating makes each iteration a distinct unknown command → ~100 distinct uncapped `priority:high` operator-queue items. The per-command dedup never collapses them, and the loop's `no_progress` guard doesn't stop it (each "Unknown command: /nonexistent-N" response is a distinct fingerprint). None of these hit the #1632 depth/rate cap and none emit a `queue-flood` alert — they look like legitimate "your agent is misconfigured" alerts, i.e. exactly the plausible operator-fatigue vector #1402 fears.

**Caveats (honest scope):** requires (a) the agent to hold loop/schedule MCP capability and (b) it to be compromised/malicious; and it is a **slow** flood — throttled by execution throughput (each iteration is a full agent turn, consuming `max_parallel_tasks`), not an instant file-flood. So it accumulates over minutes, not seconds. It is nonetheless a real bypass of the #1632 mitigation via a sibling channel.

## Proposed direction

Bring the platform alert emitters under a shared per-agent bound before pull-mode default-ON — options to weigh:
- A shared per-(agent, alert_type) create-rate/dedup budget for the direct-create alert emitters (skill-not-found first; it's the only agent-influenceable one), OR
- Tighten `_alert_skill_not_found` dedup from per-command to per-(agent, window) so a templated-command loop can't mint unbounded rows, OR
- Route the direct-create alert emitters through a small shared "platform alert budget" helper (keeps the file-seam exemption intact while capping the aggregate).

Prefer the option that keeps genuinely platform-driven alerts (poison-park, breaker, sync-health, validation) unthrottled while bounding the agent-influenceable path.

## Also noted while reviewing #1632 (minor, fold in or split)
- **Reserved-id rejects emit no operator signal.** The reserved-prefix guard (`operator_queue_service.py:423`) `continue`s without incrementing `held`, so an agent flooding platform-reserved ids (`poison-*`, etc.) produces bounded WARN logs but **no** aggregated flood alert — an invisible compromise indicator. Deliberate per the #1632 code comment (un-guessable ids already defeat suppression); candidate observability enhancement.
- **`execution_id` unbounded on the under-cap context path.** `_valid_execution_id` (≤128, id-shaped) is applied only in the truncated-context marker branch; a normal under-cap context carries its `execution_id` (up to the 8 KiB context cap) straight into the `execution_id` column with no per-field belt. Bounded, low-impact, pre-existing — minor inconsistency worth closing.

## Refs
- Parent flooding-cap issue: #1632 (file-seam caps — shipped)
- Async human-gate: #1402 (makes this queue the irreversible-action approval channel)
- Umbrella: #1081 (pull default-ON gate list — this residual belongs here)

trinity-ability:
Fixes #1677

Platform operator-queue alert emitters bypassed #1632's ingestion caps entirely. #1632 bounded the **agent-authored** seam (`operator_queue_service._sync_agent` reading `~/.trinity/operator-queue.json`), and exempted platform creates on the reasoning that platform volume is platform-bound. That reasoning holds for *most* platform emitters and fails for one: `task_execution_service._alert_skill_not_found` (#1410) fires on an **unresolved slash command**, whose per-command dedup is defeated simply by varying the command name — so an agent could mint unbounded pending operator-queue items through a path with no cap on it, re-opening the operator-fatigue → reflexive-approval surface #1632 exists to close.

## What changed

The exemption is now scoped **by influence, not by caller location**.

- **`create_bounded_alert` + `_BUDGETED_ALERT_TYPES`** (`services/operator_queue_service.py`) — a DB-measured per-(agent, registered-type) pending-depth budget (`OPERATOR_ALERT_MAX_PENDING_PER_TYPE`, default 5). Type is derived from `item["type"]` against a frozenset, so a caller-interpolated string can't register itself.
- **The one agent-influenceable emitter routes through it** — `_alert_skill_not_found`. Genuinely platform-driven emitters (lease-reaper poison-park #1402, circuit-breaker, sync-health, validation, retention-guard, system-seed, base-image-stale) stay **direct and unthrottled**: their volume is bound by platform cadence — edge-triggered, idempotent/bucketed id, leader-locked, or operator-driven.
- **Fail-CLOSED on every arm.** An unreadable count, an unregistered type, or a failed create suppresses **the alert only**. The execution's own `FAILED` / `SKILL_NOT_FOUND` row is untouched and remains the primary surface, and the paired notification is gated on the same bool — a suppressed alert never suppresses the evidence.
- **At cap:** one cooldown-gated, time-bucketed `alert-budget-{agent}-{type}-b{bucket}` episode alert per window. The deterministic bucket makes the `(agent_name, request_id)` on-conflict target dedup cross-worker; no `held` count and no agent-controlled text ride along. The `alert-budget-` prefix is added to #1632's reserved-id guard so an agent can't pre-create — and via `on_conflict_do_nothing` silence — its own budget alarm.
- **An AST caller-parity test** (`tests/unit/test_1677_operator_alert_emitters.py`) requires every `create_operator_queue_item` call site to be either allowlisted platform-only or routed through the budget. **Every FUTURE emitter is classified at CI rather than remembered.**

A sink-level default bound was deliberately **rejected**: it fails *quietly* at the load-bearing poison-park create, where the parity test fails *loudly* at CI. (#1890's lesson, deliberately inverted.)

## No schema change — no migrations on either track

Verified, not assumed: **zero** matches for `db/migrations.py|migrations/versions/|db/schema.py|db/tables.py|alembic` across the diff. The sole DB-layer change is a **WHERE-clause append** in `count_pending_for_agent` (`db/operator_queue.py`). A clean boot on a freshly migrated DB returned `/health` **200** — not the 503 + `first_pending` shape — with zero migration errors. Stated explicitly to preempt the schema-parity reflex: there is no SQLite migration and no Alembic revision to look for, because there is no DDL.

## Closes the platform-emitter residual on the #1081 pull-mode default-ON gate list

`docs/planning/TARGET_ARCHITECTURE.md` still listed operator-queue flooding caps as an open blocker in **both** places it appears, while `architecture.md` on this branch already records the residual as closed. This PR updates both gate lines so the planning doc and the architecture doc agree. The remaining gate items (trace fidelity #548/#333, `prior_trace` #1401, fail-closed `execution_id` injection) are unchanged.

## Two honest corrections to the issue's exploit sketch

The vulnerability stands; only its **shape** was mis-stated in #1677, and the record should be accurate:

1. **It's a $0 turn, not an LLM turn.** An unresolved slash command never reaches the model, so the spray runs at **dispatch** throughput — not LLM throughput, and not at LLM cost. Reachable via the `agent`, `a2a`, `schedule`, and `reminder` triggers. This makes the surface *cheaper* to exploit than the issue implies, not more expensive.
2. **The literal loop PoC needs non-default flags.** It only works with explicit `on_failure="continue", max_consecutive_failures=N`; the **default** `on_failure="abort"` stops the loop at iteration 1. The PoC as written would not reproduce on a default-configured loop.

## Evidence

All on the production-pinned **Python 3.13.5**.

| Run | Result |
|---|---|
| Targeted battery (8 files) | **151 passed** |
| Full unit suite | **9,839 passed** |
| Integration | **70 passed / 13 skipped / 2 registry-deselected** |
| `verify-local` | **PASS** — `…/trinity-verify-cc463bb6/runs/20260813-160217/result.json` |

```
python3 -m pytest tests/unit/test_1677_operator_alert_budget.py \
  tests/unit/test_1677_operator_alert_emitters.py \
  tests/unit/test_1632_operator_queue_caps.py \
  tests/unit/test_1631_operator_queue_agent_scoped_ids.py \
  tests/unit/test_1525_operator_queue_quarantine.py \
  tests/unit/test_1410_unresolved_slash_command.py \
  tests/unit/test_1142_operator_queue_retention.py \
  tests/unit/test_1644_retention_guard.py -p no:warnings -q
# 151 passed
```

**Merge-forward verified.** `dev` advanced 5 commits (A2A outbound #736, workspace agent page ent#360/#2159/#2160/#2161) after this branch's base at `76d842da`. `git merge-tree` reports **no conflicts**, and against the *merged* tree both directions are green: the 151-test battery still passes, and dev's 183 new A2A/portal tests still pass. Dev's new code introduces **no** new `create_operator_queue_item` call site, so the AST parity guard stays green post-merge.

**Honest scope limits.** Agent stages were skipped (`--skip-agent`, pre-justified — no `docker/**`, no `requirements*`, no create-path changes), so this is **not** agent-side evidence. `test_alembic_postgres.py` is among those skips — harmless here precisely because there is no Alembic revision to exercise.

## Pre-existing `dev` reds, disclosed

The 3 `XPASS(strict)` in `tests/unit/test_pat_propagation_properties.py::TestKnownGaps` **reproduce on pristine `origin/dev` @ 76d842da** — they are not introduced by this PR. Cause: a stale `xfail(strict=True)`; #2017 fixed the underlying bug and #2018's marker was never flipped. Invisible to CI because dev's required checks don't run `tests/unit`. Not this PR's to fix — recorded here for the reviewer's awareness only.

## Review-stage fix included

`1119b5dc` closed a `ZeroDivisionError` that broke `create_bounded_alert`'s documented never-raises contract when the cooldown knob is set to `0` (the natural "disable cooldown" spelling). Two-layer fix — the bucket divisor is floored at 1, and the at-cap `await` is wrapped so the contract holds **structurally** against future edits of the episode helper. +2 regression tests.

## Follow-ups (listed, deliberately NOT filed)

- **Reserved-id rejects as a compromise-indicator signal** — deliberately deferred; an existing #1632 test pins the current no-alert behavior, so changing it is a scoped decision of its own.
- **`portal-inbox-collision-` prefix is still unreserved** (ent#308 residual).
- **The notification-sink (`create_notification`) emitter classification has no parity net** — the same reasoning applies to that sink, but this PR's guard covers `create_operator_queue_item` only.
- **An enterprise-tree twin of the caller-parity guard** — the guard is OSS-tree-only by construction; the private submodule owns its own twin.
- **`///` all-slash commands raise `IndexError`** — swallowed, pre-existing, no security delta.
- **`.claude/agents/test-runner.md` catalog rows** land separately in the private `trinity-dev` submodule (intentionally not committed here).


vybe (approved):
Validated via /validate-pr. Scopes the #1632 platform-emitter exemption by INFLUENCE rather than caller location — the right axis: _alert_skill_not_found is agent-influenceable (distinct unknown slash-commands defeat its per-command dedup at $0/turn), while edge-triggered/leader-locked platform emitters stay direct. Fail-closed on every arm (unreadable count, unregistered type, failed create) suppresses only the ALERT, with the FAILED execution rows left as the primary surface and the paired notification gated on the same bool. The AST caller-parity guard is what makes this durable — a sink-level default bound would fail quietly at the load-bearing poison-park create, where CI fails loudly. Reserved-id prefix added so an agent can't pre-create its own budget alarm. OPERATOR_ALERT_MAX_PENDING_PER_TYPE wired into both compose files and .env.example. Earlier base-job failure was a 25m hung-job cancellation with zero test failures; re-ran clean. All checks green.

Issue #471, opened 2026-04-23
feat: Subscription usage observability — 5h/7d usage dashboard, per-agent breakdown, and rate-limit event history


vybe:
## Summary

All the data needed to understand subscription pressure already exists in the database — token counts per message, cost per execution, rate-limit events per subscription — but none of it is surfaced in the UI. Operators have no way to see how close a subscription is to its 5-hour rolling limit, which agents are burning the most quota, or how frequently a given subscription has hit the wall. This issue wires the existing data to the front end.

## Context

Triggered by subscriptions exhausting their limits significantly faster than expected (#435, SUB-003). Investigation confirmed:

- `subscription_id` is already recorded on every `chat_messages` and `schedule_executions` row (wired in `task_execution_service.py`)
- `GET /api/subscriptions/{id}/usage` already aggregates 5h and 7d token windows from that data (SUB-004)
- `subscription_rate_limit_events` table is populated by SUB-003 on every 429, but is never exposed via API
- `Settings.vue` subscription list calls none of these endpoints — all data is dark

Anthropic does not expose remaining budget or proximity-to-limit. The only signal available is the 429 itself and our own token consumption records.

### Tier 1 — Surface existing data in Settings → Subscriptions

- [ ] Each subscription card in Settings.vue shows a usage summary for the last 5 hours: input tokens, output tokens, cost (USD), message count
- [ ] Each subscription card shows a 7-day aggregate for the same fields
- [ ] Rate-limit event count (last 24h) is shown per subscription with a visual warning indicator if > 0
- [ ] Data auto-refreshes on page load (no polling needed)

### Tier 2 — Per-agent breakdown endpoint

- [ ] New endpoint `GET /api/subscriptions/{id}/usage/breakdown` returns token consumption grouped by `agent_name` for the 5h and 7d windows
- [ ] Settings.vue subscription cards show an expandable per-agent breakdown (collapsed by default)
- [ ] Makes it clear which agent is consuming the most quota on a shared subscription

### Tier 3 — Subscription pressure indicator on Agents list

- [ ] Agents.vue shows a small warning badge on any agent card whose subscription had 1+ rate-limit events in the last 24h
- [ ] Hovering the badge shows the subscription name and event count
- [ ] Badge is sourced from a new lightweight endpoint (or piggy-backed on the existing agents list response)

### Tier 4 — Auto-assign to available subscription

- [ ] Settings → Subscriptions shows an "Auto-assign agents" button (or equivalent action per subscription card)
[...]

PR #2170, opened 2026-08-13
Show Claude subscription usage without metered cost

erichroepke:
## Summary

- add a discriminated `subscription` / `api` / `unconfigured` usage contract to agent auth, agent list/detail, and subscription responses
- show each Claude subscription-backed agent's named subscription and provider-backed allowance percentage/window/reset metadata when available
- return an explicit `provider_signal_unavailable` state today because Trinity has no reliable Claude allowance counter; never substitute token totals or Claude Code dollar estimates
- keep Anthropic API usage explicitly labelled `API · metered` with USD cost semantics
- update the agent header, Tasks summary, Dashboard grid/list/timeline, and Settings subscriptions table
- suppress Anthropic presentation entirely for non-Claude runtimes

## Provider data limitation

Claude setup-token authentication currently exposes no reliable allowance percentage/window/reset signal to Trinity. The new model supports an available state, but production responses remain honest and explicit until an authenticated provider signal can supply percent consumed, window identity, observation time, and reset time when defined.

## Verification

- `pytest tests/unit/test_subscription_usage_presentation.py -q` — 6 passed
- frontend unit suite — 311 passed
- focused Ruff — passed
- focused mypy — passed
- Python compileall — passed
- design-token check — passed
- production Vite build — passed
- staged Gitleaks scan — no leaks
- headed local Playwright review verified subscription, API-metered, and Settings presentation states

## Notes

- No credentials, subscriptions, or provider authentication were changed.
- No deployment or merge was performed.

### DIFF
--- a/docs/memory/feature-flows/subscription-usage-tracking.md
+++ b/docs/memory/feature-flows/subscription-usage-tracking.md
@@ -1,20 +1,20 @@
 # Feature: Subscription Usage Tracking
 
 ## Overview
-Per-subscription rolling token and cost usage across two time windows (5h and 7d), covering both chat messages and scheduled task executions.
+Per-subscription non-billing activity across two rolling windows (5h and 7d), covering both chat messages and scheduled task executions. Allowance utilization is a separate provider-backed signal: if Trinity cannot obtain a reliable percentage/window/reset, the API and UI return an explicit unavailable state instead of substituting token totals or Claude Code's dollar estimate.
 
 ## User Story
-As a platform admin, I want to see aggregate token usage per Claude Max/Pro subscription so that I can understand load distribution across shared subscriptions and detect overuse.
+As a platform admin, I want to see which named Claude subscription funds each agent and its provider-backed allowance utilization when available, without mistaking an estimated dollar value for subscription spend.
 
 ## Entry Points
 - **API**: `GET /api/subscriptions/{subscription_id}/usage`
-- No direct UI entry point — intended for admin inspection and future dashboard integration. Accepts subscription UUID or name as the path parameter.
+- **UI**: agent header, Dashboard timeline/grid/list, and Settings subscription table. Accepts subscription UUID or name as the API path parameter.
 
 ## Backend Layer
 
 ### Endpoint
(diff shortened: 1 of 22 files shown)


vybe:
Thanks for this — the diagnosis underneath it is correct and worth acting on. A bare dollar figure on a subscription-backed agent implies metered spend that doesn't exist, and that's a real defect. I'm going to fix it, but not this way, so I'm closing the PR.

The disagreement is about the cure. This PR removes the measurement: `cost_usd` comes out of the `SubscriptionUsage` aggregation SQL for both windows, and `AgentHeader.vue` loses the 7-day sparkline, today's cost, and the trend-vs-average for subscription agents. What replaces them is `usage unavailable`, and that state is permanent rather than transitional — the `available` arm needs an allowance percent/window/reset that Anthropic doesn't expose, as the PR body itself notes, so the progress bar behind `v-if="utilization?.status === 'available'"` can never render. The net effect across five surfaces is trading a real-but-mislabeled number for no number at all.

The number is worth keeping because of what it actually is. For Claude Code, `cost` isn't our estimate — it's the runtime's own `total_cost_usd`, real token counts at API prices. On a subscription that's counterfactual spend: what this work *would have* cost on the API. That's a legitimate and useful figure — it's the value being extracted from the flat-rate plan, it tracks load, and enterprise per-agent spend caps need it. So the fix is to label it honestly (`≈$3.12 API-equivalent`, plus a billing-mode badge) rather than suppress it.

Separately, the honest allowance signal already exists and is dark: `GET /api/subscriptions/{id}/usage` computes real 5h/7d token windows, and `subscription_rate_limit_events` is populated on every 429. Neither reaches the UI. That's the work, and it's tracked in #471 (P1, ready) — which concluded back in April that no provider-side allowance counter exists, so a contract built around one shouldn't ship ahead of it.

Two smaller notes for future PRs: `requirements/security.md` was edited to narrow SUB-004 to match the code, and per our Rules of Engagement requirements lead implementation rather than follow it; and the `subscription_credentials` join added to `get_all_agent_metadata` puts subscription names on the fleet roster, which also serves shared non-owner users.

I've written the approach up on #471. Genuinely useful contribution — you found the bug, and the labeling fix lands because of it.

Issue #2162, opened 2026-08-13
fix(workspace): Reports tab shows raw JSON to clients — reuse the display_hint renderers


vybe:
## Summary

The Workspace agent page's **Reports tab renders raw JSON** to external clients. The Workspace is a client-facing surface — a report must always render in a human-readable form; a JSON blob is never an acceptable presentation there.

## Context

The Agent Detail surface already has the full renderer set for reports (#918/#1535): `src/frontend/src/components/reports/` picks a renderer by `display_hint` → `report_type` prefix → fallback. The Workspace agent page (ent#360) fetches report metadata/detail through `src/backend/client_portal/agent_page.py::reports` / `report_detail` but does not reuse those renderers, so payloads land as raw JSON.

## Acceptance Criteria

- [ ] The Workspace Reports tab renders reports through the shared `components/reports/` renderer set (`display_hint` → `report_type` prefix → fallback), same behavior as Agent Detail
- [ ] The fallback for an unknown/malformed payload is a **human-readable presentation** (e.g. titled key-value summary), never a raw JSON dump — stricter than Agent Detail's JSON-viewer fallback, because the audience is an external client
- [ ] `table` payloads use the windowed rows endpoint pattern (#1537) so a large report doesn't transfer wholesale
- [ ] Verified in light and dark themes on the Workspace surface

## Technical Notes

- Frontend: `src/frontend/src/components/portal/PortalAgentPage.vue` (Reports tab) + shared `src/frontend/src/components/reports/` renderers
- Backend: `src/backend/client_portal/agent_page.py::reports` (metadata) / `report_detail` (payload) — the list-metadata / detail-payload split already matches the #918 convention
- Renderer keys are CI-pinned to the MCP `display_hint` enum (`tests/unit/test_1535_report_prompt_guidance.py`) — reuse, don't fork

PR #2171, opened 2026-08-13
fix(workspace): Overview layout — activity + recent work share a 50/50 row, asks below, avatars get an edge (#2169)

### LINKED ISSUE #2169: fix(workspace): Overview layout — activity + recent work share a 50/50 row, asks move below, avatars get an edge
## Summary

Three layout/polish defects on the Workspace client surface, all in the agent page Overview and the shared avatar:

1. **The two-column split is conditional on data.** In `PortalAgentPage.vue` the Overview grid is `<div class="mb-6 grid gap-6" :class="{ 'lg:grid-cols-2': asks.length }">` — the second column only exists when the agent has open asks. An agent with nothing waiting collapses to a single stacked column, so the page reflows into a different shape depending on whether a transient operator-queue item happens to be open. The structure should be stable; only the *content* should vary.

2. **Wrong things are sharing the row, in the wrong order.** Today the row is `Waiting on you` (left) | `Recent work` (right), with the activity chart sitting alone above in a `max-w-2xl` section. The intended layout is: **activity chart and recent work share one 50/50 row**, and **asks ("Waiting on you") move below that row**, full width, rendered only when there are asks.

3. **The agent avatar has no edge.** `PortalAvatar.vue` renders a bare `rounded-full` span with no border. Against the sidebar and chat backgrounds — and especially for image avatars whose own edges are light — the avatar bleeds into the surface. It needs a thin border in both places it appears (sidebar roster and chat rows), which is one change in the shared component.

## Context

Follow-up to #2161, which restructured this Overview (moved asks into the grid, reused the shared chart). The layout that landed keys its column count off `asks.length`, which is exactly the instability being reported here. Same file family, same reviewer context.

Target layout:

```
┌───────────────────────────┬───────────────────────────┐
│ Activity · last {window}  │ Recent work               │   <- always two columns
└───────────────────────────┴───────────────────────────┘
┌───────────────────────────────────────────────────────┐
│ Waiting on you                                        │   <- only when asks.length
└───────────────────────────────────────────────────────┘
┌───────────────────────────────────────────────────────┐
│ Your chats with {agent}                               │
└───────────────────────────────────────────────────────┘
```

Note that the target layout resolves the stability complaint by construction: both columns of the top row (activity, recent work) always exist and already have their own empty states ("No activity in this window." / "Nothing yet."), so the split never has to collapse.

## Acceptance Criteria

- [ ] The Overview top row is a 50/50 two-column grid holding **activity chart** (left) and **Recent work** (right); the column count does not depend on `asks.length` or on either section having data
- [ ] Both columns render their existing empty states in place rather than collapsing the grid — an agent with no executions and no activity still shows two columns
- [ ] "Waiting on you" moves **below** that row, full width, and is rendered only when `asks.length > 0` (no empty-state placeholder — an agent with nothing waiting should not advertise the section)
- [ ] The activity section's `max-w-2xl` constraint is removed or re-scoped so the chart fills its column
- [ ] The grid collapses to a single stacked column below the **`xl`** breakpoint (narrow viewports keep today's stacking behaviour) — layout stability applies within a breakpoint, not across them

  > **Amended from `lg` to `xl` during planning, on measurement.** `lg` was written from intent, not arithmetic. At 1024px, dropping the activity section's `max-w-2xl` and splitting 50/50 takes the chart from 672px to **332px** (sidebar `w-72` + `px-6` + `gap-6`) — the 30-day x-axis truncates to nothing and the 9-bucket legend wraps to 3–5 lines. `StackedBarChart` is pure CSS/flexbox, so the theme criterion below would still pass while the left column became unreadable. Recorded here so a later reader does not "fix" it back to `lg`.
- [ ] `PortalAvatar.vue` renders a thin border on every instance, visible in both light and dark themes, using semantic tokens (no raw palette classes, no hardcoded hex)
- [ ] The border does not change the component's outer footprint or clip the image fill — verify the avatar still aligns in the sidebar roster rows and the chat rows at every size it is used with
- [ ] Verified in the browser on the agent page in both themes: with asks and without, with executions and without

## Technical Notes

- `src/frontend/src/components/portal/PortalAgentPage.vue` — the Overview `<template v-else-if="tab === 'overview'">` block holds the activity `<section class="mb-6 max-w-2xl">`, the conditional grid, and the chats section
- `src/frontend/src/components/portal/PortalAvatar.vue` — single shared component; both the sidebar (`PortalSidebar.vue`) and chat (`PortalChatRow.vue` / `PortalConversation.vue`) consume it, so the border is one edit, not three
- Read `docs/memory/design-system-contract.md` before touching either file (semantic tokens only, both themes first-class, layout stability)
- No backend change; no new endpoint. Pure frontend layout.

---

### Planning decision — mobile ask-priority (recorded, reversible)

Moving "Waiting on you" below the row applies at **every** breakpoint, so on a stacked narrow viewport the asks sit under activity + recent work. This was raised as a blocking objection during plan review and is **implemented as specified** — the layout was an explicit instruction and an acceptance criterion is not the planner's to overrule.

Mitigations already present: the section is conditional (absent when there are no asks) and the Overview tab carries a count badge. The costed alternative (a ~3-line mobile-only order swap) is noted in the PR body so it can be reversed in one commit if the stacked ordering proves wrong in use.

trinity-ability:
Fixes #2169

The Workspace Overview's top row keyed its column count off `asks.length`, so the page changed shape whenever a transient operator-queue item opened or closed — and the agent this issue reports, the one with nothing waiting, collapsed to a single column. Portal avatars had no edge, so a light-edged image avatar bled into the surface behind it.

## What changed

The Overview top row is now an **unconditional** 50/50 grid — activity chart left, recent work right. It never has to collapse, because both occupants already own an empty state (*"No activity in this window." / "Nothing yet."*), so structure holds still and only content varies. The chart's `max-w-2xl` cap went with the move; inside a half-width column it does not bind below ~1656px.

**Asks moved below that row, full width**, still rendered only when there are any — an agent with nothing waiting must not advertise the section. Everything #2161 built into the card survives: compact cards, clamped question, first five plus a counted toggle, and no nested scroll region (#2101).

The **loading skeleton** was changed to match. It was one column in front of what is now a two-column row, so every load ended in a reflow — and the no-asks agent is exactly the case where skeleton and loaded state used to agree. Measured after: skeleton blocks and loaded sections land on identical boxes (x 312/876, y 198, w 540).

`PortalAvatar.vue` gained a 1px edge, and so did the hand-rolled `+N` chip in `PortalChatRow.vue:28` — without the second, a four-agent row draws three hairlined circles and one bare blob.

## The breakpoint is `xl`, not the `lg` in the issue text

Arithmetic, not taste. The Workspace holds a 288px sidebar plus page padding and a 24px gap, so at 1024px each column is 332px — where the 30-day x-axis (one truncating 9px label per day) reads as nothing and the nine-bucket legend wraps three to five lines. At 1280px the column is ~460px, which the legend needs and gets. A layout that technically satisfies "two columns" while making the left one unreadable fails the purpose.

**The issue's AC was amended by the orchestrator to `xl`.** Reviewers reading an older copy of the issue text will see `lg`; that is the discrepancy, and it is deliberate.

## Named tradeoff: asks now sit below the row at *every* breakpoint

This reverses #2161's deliberate "asks are first in DOM order so the mobile stack keeps the priority" decision. Implemented as specified, not by oversight — flagging it because it is a real priority inversion on narrow viewports.

The residual is bounded: the Overview tab's ask-count badge sits `shrink-0` in the header, **outside** the page scroller (verified live at 640 and 375), so a narrow viewport shows the count at every scroll position. Only the ask text moves below the fold.

**Costed reversal if you want it back:** keep the asks `<section>` first in DOM inside the same grid with `xl:col-span-2 xl:order-last`. Desktop renders identically, mobile keeps asks first, ~3 lines, one commit to undo — at the price of a DOM/visual order split.

## Accepted known residual: 30d x-axis clipping, ~1280–1680

The 30-day window's x-axis is ellipsis-clipped in a half-width column. Measured over a 1280→2000 sweep (Chromium, macOS system font — glyph widths are platform-dependent, so treat the band, not the counts, as the finding):

| width | ticks clipped (of 6) |
|---|---|
| 1280–1320 | 6 |
| 1360–1520 | 4 |
| 1600–1640 | 1 |
| ~1680+ | 0 |

Bounded on the other side too: **below 1280 the row stacks** and the chart runs full width — *wider* than the pre-#2169 `max-w-2xl` cap — so it is clean there as well. **7d (the default) and 14d measure zero clipped ticks at every width.** The clipping is ellipsis-marked rather than silently wrong; bars, legend totals and the hover tooltip's full date are unaffected.

The fix is width-responsive tick density in `StackedBarChart.vue::showLabel`, which the **operator** Overview (#1107) also consumes — deliberately out of scope for a client-facing P1.

## Raw-color delta is reviewer-enforced — CI cannot catch it

`raw_gray` **+2** on each of `PortalAvatar.vue` and `PortalChatRow.vue`; `raw_nongray` **0**, unchanged. Measured both sides:

```
origin/dev:  PortalChatRow.vue  12 gray   (PortalAvatar.vue: 0, not listed)
branch:      PortalAvatar.vue    2 gray
             PortalChatRow.vue  14 gray
```

It is `docs/memory/design-system-contract.md:15`'s `border-strong` gray-300/gray-700 recipe verbatim, and **no semantic neutral-edge token exists** — `status-*`/`action-*` would claim this decoration reports a result or affords a click.

Please eyeball it rather than trusting the checks, because two independent gaps mean a green run is not evidence here:

- **Both files are absent from `raw-color-baseline.json`** (it carries 6 portal files; neither of these), so the per-file ratchet has nothing to compare against.
- **`check:tokens` never inspects `border-gray-*`** — grepped the checker, no such rule. It passes: `Design-token check OK: 11 tokens equivalent to source palettes; all references resolve; dark ink ladder holds in 13 swept files`.
- **No CI workflow invokes `scan-raw-colors` at all** — grepped `.github/workflows/`, zero hits. The ratchet is advisory.

## Verification

`verify-local` **PASS** (`--skip-agent`; agent stages skipped, correct — this branch touches no `docker/base-image/**`).

Live both-themes visual pass on the branch's own dev server. An agent with **zero asks** renders 540/540 — the state that used to collapse. The reported light-theme defect quantified: a near-white avatar on white ground is **1.017:1 (invisible)**; the edge lifts it to **1.474:1**. Avatar footprint unchanged at all nine sizes in use (16, 18, 20, 22, 26, 28, 30, 52, 64); image insets 1px, never clipped.

Three geometry choices settled by measurement, not preference:
- **`border`, not `ring-inset`** — an inset ring paints below child content and the `` is exactly the padding box, so it is invisible on precisely the image avatars this fixes.
- **`border`, not an outer ring** — `PortalChatRow` already passes `ring-2` in as the stacked-avatar separator, and all Tailwind rings share one `box-shadow`. Border and ring are independent properties and compose.
- **`border-strong`, not `border`** — `gray-200` on the sidebar's `gray-50` ground is 1.19:1, an arc too faint to see at 26px. `gray-300` is 1.41:1.

### Tests

`cd src/frontend && npx vitest run` → **23 files / 342 tests passed**.

Backend was swept in full on this branch during verification: `9960 passed, 3 failed, 17 skipped, 1 xfailed`. The 3 failures are **pre-existing on `dev`** — `[XPASS(strict)]` reds in `test_pat_propagation_properties.py::TestKnownGaps::test_a_backslash_in_the_token_does_not_raise`, reproduced on a pristine `origin/dev` checkout. Not re-run here (concurrent ship workers share uncached settings state and manufacture phantom reds).

New coverage: `tests/unit/portalAvatarEdge.spec.js` (6 cases) and 6 added structure guards in `portalAgentPageUx.spec.js`. Note commit `775505ff` — the headline placement guard was passing against the *skeleton*, not the row; it now measures the row.

### e2e caveat — read before running

`e2e/portal-agent-page-overview.spec.js` needs `PORTAL_TEST_AGENT` pointed at an agent **with an open ask**, or the headline placement case (`asks, when present, sit below the row at full width`, line 143) **silently skips**:

- with such an agent → `expected 6, skipped 0`
- without → `expected 5, skipped 1`

The unconfigured default fails loud rather than skipping, so a misconfigured run is not mistaken for a pass.

## Rebase expectation

`PortalAgentPage.vue`, `docs/memory/requirements/core-agent.md` §5.11 and `docs/memory/feature-flows/workspace-agent-page.md` each have a second writer this wave (#2162). All edits here were surgical — verified disjoint hunk ranges, `<script setup>` untouched — so resolution is **take both sides**.

**After any rebase, re-run `npx vitest run`**: the structure guards read the whole file off disk and use offsets, so a clean textual merge can still move them.

- Width-responsive tick density in `StackedBarChart.vue::showLabel` — closes the 30d x-axis band; shared with the operator Overview, so it needs its own change.
- Retiring the 4 `raw_nongray` amber classes on the ask card — needs design sign-off, it is a real amber→yellow hue shift.
- Adding the three portal files to `raw-color-baseline.json` so the ratchet covers them.
- The stale `.claude/agents/test-runner.md` frontend counts (private submodule, separate change).

### DIFF
--- a/docs/memory/feature-flows/workspace-agent-page.md
+++ b/docs/memory/feature-flows/workspace-agent-page.md
@@ -195,11 +195,21 @@ column that cap does not bind below ~1656px.
 332px — where the 30-day x-axis (one truncating 9px label per day) reads as
 nothing and a nine-bucket legend wraps three to five lines. At 1280px the column
 is ~460px. A layout that technically satisfies "two columns" while making the
-left one unreadable fails the purpose. **Known residual:** the 30-day x-axis
-still truncates in a half-width column (measured 5/6 shown ticks at 1280, 3/6 at
-1600); 7d and 14d are clean. The fix is width-responsive tick density inside
-`StackedBarChart.vue`, which the operator Overview also consumes — out of scope
-here, deliberately.
+left one unreadable fails the purpose.
+
+**Known residual, accepted:** the 30-day window's x-axis is ellipsis-clipped in a
+half-width column. Measured over a 1280→2000 sweep (Chromium, macOS system
+font — glyph widths are platform-dependent, so treat the band, not the counts, as
+the finding): all six ticks clipped at 1280–1320, four at 1360–1520, one at
+1600–1640, **none from ~1680 up**. It is bounded on the other side too — below
+1280 the row stacks and the chart is full width, which is *wider* than the
+pre-#2169 `max-w-2xl` cap, so it is clean there as well. So the affected band is
+roughly **1280–1680 on the 30-day window only**: 7d (the default) and 14d measure
+zero clipped ticks at every width, and the clipping is ellipsis-marked rather
+than silently wrong — the bars, the legend totals, and the hover tooltip's full
+date are all unaffected. The fix is width-responsive tick density inside
+`StackedBarChart.vue::showLabel`, which the operator Overview also consumes —
+out of scope here, deliberately.
 
 **Moving asks below reverses #2161's DOM-order decision**, on instruction. The
 residual is bounded rather than a priority inversion: the header carrying the
--- a/docs/memory/feature-flows/workspace-agent-page.md
+++ b/docs/memory/feature-flows/workspace-agent-page.md
@@ -158,25 +158,104 @@ keep asks where they are: the page's own reason to exist is that an agent can
 reach you when no chat is open, and a tab is a place you have to *go*. The defect
 was that they were unbounded, not that they were present. So they are contained
 instead — compact cards, the question clamped, the first five with a counted
-"Show all" toggle, and a two-column split beside recent work on wide screens.
+"Show all" toggle.
 
 Two constraints on that layout:
 
-* **Asks are first in DOM order**, so the mobile stack keeps the priority the
-  page was built around.
+* ~~**Asks are first in DOM order**, so the mobile stack keeps the priority the
+  page was built around.~~ **Superseded by #2169** — see below. Asks now sit
+  below the top row on every breakpoint, which on a narrow viewport puts them
+  third.
 * **No nested scroll region.** #2101 settled this on this surface: the page has
   one scroll axis, and a pane that scrolls inside a page that scrolls traps the
   gesture on touch. Containment is first-N-plus-toggle, not `overflow-y-auto`.
+  **Still true after #2169** — the section moved, its containment did not change.
 
 The badge reads `20+` at the cap, because `MAX_ASKS` truncates server-side — a
 bare "20" against 50 pending is a wrong number, not a rounded one.
+
+### The Overview row is unconditional, and asks moved below it (#2169)
+
+#2161 put asks and recent work in a grid whose column count was **bound to
+`asks.length`** (`:class="{ 'lg:grid-cols-2': asks.length }"`), with the chart
+alone above in a `max-w-2xl` section. That made the page's shape a function of
+its data: an agent with nothing waiting collapsed to one column, so the layout
+changed whenever a transient operator-queue item opened or closed.
+
+The top row is now **unconditional** and its occupants are the chart (left) and
+recent work (right); asks sit **below it at full width**, still only when there
+are any and still with no empty-state placeholder — an agent with nothing
+waiting must not advertise the section. Nothing has to collapse, because both
+row occupants already own an empty state ("No activity in this window." /
+"Nothing yet."). The chart's `max-w-2xl` went with the move: inside a half-width
+column that cap does not bind below ~1656px.
+
+**The split is `xl`, not `lg`, and that is arithmetic.** The Workspace holds a
+288px sidebar plus page padding and the 24px gap, so at 1024px each column is
+332px — where the 30-day x-axis (one truncating 9px label per day) reads as
+nothing and a nine-bucket legend wraps three to five lines. At 1280px the column
+is ~460px. A layout that technically satisfies "two columns" while making the
+left one unreadable fails the purpose. **Known residual:** the 30-day x-axis
+still truncates in a half-width column (measured 5/6 shown ticks at 1280, 3/6 at
+1600); 7d and 14d are clean. The fix is width-responsive tick density inside
+`StackedBarChart.vue`, which the operator Overview also consumes — out of scope
+here, deliberately.
+
+**Moving asks below reverses #2161's DOM-order decision**, on instruction. The
+residual is bounded rather than a priority inversion: the header carrying the
+Overview tab's ask-count badge is `shrink-0` and sits **outside** the page
+scroller, so a narrow viewport shows the count at every scroll position; only the
+ask text moves below the fold. If it needs reversing, the costed alternative is
+~3 lines — keep the asks section first in DOM inside the same grid with
+`xl:col-span-2 xl:order-last` — which reads identically on desktop and keeps asks
+first on mobile, at the price of a DOM/visual order split.
+
+The **loading skeleton** gained the same `xl` split. It was one column in front
+of what is now a two-column row, so every load ended in a reflow — and the
+no-asks agent this issue is about is exactly the case where skeleton and loaded
+state used to agree. Measured after the change: skeleton blocks and loaded
+sections land on identical boxes (x 312/876, y 198, w 540).
+
+### The avatar has an edge (#2169)
+
+`PortalAvatar` was a bare `rounded-full` span, so an image avatar with light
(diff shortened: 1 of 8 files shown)


vybe (approved):
Validated via /validate-pr. Merges clean onto dev; combined result verified locally (27 files / 550 frontend tests pass, prod build OK, check:tokens OK). Raw-color delta isolated dev→merged: raw_nongray -1, hardcoded 0, gray additions are the documented border-strong / dark-ink-ladder recipes with no semantic token available. Security scan clean.

PR #2172, opened 2026-08-13
feat(workspace): composer typeahead — / for playbooks, @ for agents (ent#392)

trinity-ability:
Fixes Abilityai/trinity-enterprise#392

> Note the qualified cross-repo form. `Abilityai/trinity#392` in **this** repo is an
> unrelated merged PR ("feat(ui): add floating help chat widget"), so a bare `#392`
> would point at the wrong thing. GitHub auto-closes same-repo only — the private
> issue needs a **manual `status-in-dev` bump and close at release**.

## What changed

The Workspace composer's two invocation syntaxes become discoverable.

- **`/`** opens the active agent's `playbooks[]` and splices the selected entry's
  **`starter_prompt` into the composer without sending** — the same prefill contract the
  ent#138 briefing cards already use, now reachable *after turn 1*, where those cards are
  gone.
- **`@`** opens reachable agents and inserts a mention token that round-trips through the
  existing `mentionedAgents()` parser (ent#361).

One new presentational component (`PortalTypeahead.vue`) plus a new pure-logic layer in
`portalUtils.js` (+22 exports), wired into the 1:1 composer (`PortalConversation.vue`) and
the room composer (`PortalRoom.vue`). `PortalBriefing.vue` now imports the `starterFor`
helper instead of defining its own copy.

All decidable logic is **pure and exported**, because `vitest` here runs
`environment: 'node'` with no component-mount harness — a decision made inside a component
is a decision no test can reach. The components are dispatchers; the panel is presentational.

**Backend, store and router are untouched: no new endpoint, no new table, no migration.**

## OSS-core by decision — deliberately ungated

No `requires_entitlement`, logic stays in the OSS tree.

Recording this explicitly because CLAUDE.md's default for an enterprise-tracker feature is
*gated unless ruled otherwise* — so the ruling must never be inferred later from the mere
fact that it merged (the ent#326 / ent#384 discipline). Rationale: it extends a surface
that is already OSS-core (the Workspace, ent#356) over data the client already holds.

The ruling is written into four places so it cannot be lost: requirements §5.13,
`architecture.md`, the flow doc, and the `portalUtils.js` header.

## `/`-in-room is deferred, explicitly

The issue's AC permits this. A room has N participants and **no active agent**, so
"whose playbooks?" has no answer without inventing a participant-picker the issue does not
specify. `@` ships in rooms; `/` does not.

## Room `@` ships — and here is exactly what justifies its scope

The candidate list is scoped to the room's **agent participants**. That scope was
established by **observing the running server**, not by reading the private rooms engine.
Carrying the observations as what they are:

| probe | server response |
|---|---|
| mention a **participant** | `{"mentions":["acme-scout"],"woke":["acme-scout"]}` |
| mention a **non-participant** | `{"mentions":[],"woke":[]}` |

Confirmed live in the UI: the sidebar listed 10 agents, and the room's `@` offered exactly
2 — the participants. So the list offers only names a pick is *known* to wake; offering the
whole roster would put names in front of the user with no evidence that choosing one does
anything.

### The recruiting question is OPEN — please don't tidy this hedge away

The probe only ever read **two response fields** and **never inspected the participant
list afterwards**. An empty `woke` therefore does **not** prove a non-participant was not
*recruited* into the room. Requirements §5.12 records an engine-side
`_join_mentioned_newcomers` path from ent#361.

If that path is live, our participants-only scoping is **narrower than the engine allows** —
which is the safe direction — but it is *not proven*, and the doc says so rather than
claiming a behaviour we did not observe. Recruiting stays with the explicit "+ Add agent"
control, the honest home for an action that spends money on another agent.

Settling this properly means observing the participant list after a non-participant
mention. Listed as a follow-up below, not filed.

## Deviations from the AC, named

**A-9 keyboard.** The AC says "Enter/Tab to accept". Shipped instead:

- plain **Enter** accepts **only when an item was explicitly selected**; otherwise it **sends**
- **Tab** accepts the top row
- Esc dismisses, and keeps the popup shut while the same token is still being typed

The harm is asymmetric. An accidental *accept* destroys typed work — a popup that merely
happens to be open (a paste, or prose like "check /status of the deploy") would splice up to
500 characters over the message. An accidental *send* is what the user was reaching for.
So the roving index starts at "nothing chosen". Verified live: typing `@scout` and pressing
Enter **sent it as a message** rather than splicing to `@acme-scout`.

## Two rules that are load-bearing, not stylistic

**The trigger rule is deliberately stricter than the parser.** `MENTION_RE` is unanchored,
so `user@example.com` *parses* as `@example`. The typeahead only fires on a trigger char at
a token start (preceded by a **non-word** char — not merely whitespace, or it could not fire
after CJK, an emoji, or punctuation). Asymmetric in the only safe direction: the popup can
never open on something the parser would not see, so `50/50`, `and/or` and email addresses
are left alone. Test #2 asserts this at **every caret index** of those strings, as a loop
rather than a sample.

**Some agents are structurally un-mentionable, and are excluded.** `sanitize_agent_name`
permits `.` and imposes no length cap; the mention grammar permits neither. So a slug like
`data.scout` parses as `@data`, matches nothing, and **silently degrades to plain text** —
the exact failure this feature exists to close. Nothing offers such names today, which is
why the failure is currently invisible; a list that included them would *manufacture* it.
The exclusion predicate is **derived by asking `mentionedAgents` itself**, never a second
hand-written copy of the grammar — a mutation that re-typed it as a copied regex which
drifted to allow dots fails 10 assertions.

**Agent matching is substring; playbooks stay word-start-prefix.** A strict prefix fails
`scout` → `acme-scout`, and a shared deployment prefix makes exactly that the common query.
Ranked whole-prefix → word-start → substring, so substring hits never outrank real prefixes.

## Verification

| gate | result |
|---|---|
| `cd src/frontend && npx vitest run` | **23 files / 444 tests passed** (111 of them new) |
| `npm run check:tokens` | OK — 11 tokens equivalent, all refs resolve, dark ink ladder holds |
| `npm run build` | ✓ built |
| `verify-local --skip-agent` | **PASS** |
| backend sweep (measured WAVE-4, this branch) | `9960 passed, 3 failed, 17 skipped, 1 xfailed` |

The 3 backend failures are **pre-existing on `dev`** — `[XPASS(strict)]` in
`test_pat_propagation_properties.py::TestKnownGaps::test_a_backslash_in_the_token_does_not_raise`,
reproduced on a pristine `origin/dev` checkout. Not re-run here: three ship workers are
running concurrently and share uncached settings state, which manufactures phantom reds.

`verify-local` was run with `--skip-agent`; the agent stages were skipped **correctly** —
the diff is `src/frontend/**` + docs with zero backend surface.

**Live visual pass**, both themes, 11 cases — including the mid-sentence-pick regression
that this branch's own review found (the caret lands *inside* the inserted token, which is
the exact re-detect condition, and the panel must stay closed; commit `c7520aa8`).

Honest caveat on the evidence: the **`/` list case was a route-injected probe**, because
every agent on the test instance is stopped, so `_agent_briefing` legitimately returns no
playbooks. The `@` cases were driven against real server state.

Seven mutations were run against the suite to prove the guards can actually fail — baking a
space into `buildMentionToken` (8 failures), accepting Enter without an explicit selection
(1), dropping the mentionable filter (6), dropping the forward scan (3), removing the
token-boundary rule (5), the drifted-regex copy (10), dropping the post-pick settle (1).

## One design-system note

The new `PortalTypeahead.vue` carries **26 raw `gray-*` Tailwind classes**, so
`scan-raw-colors.mjs` counts a new baseline entry. Deliberately not "fixed" here:

- it adds **0 raw non-gray** and **0 hardcoded hex** — the branch's repo-wide totals for
  both are byte-identical to `dev` (810 / 454)
- the three *modified* portal files add **zero** raw colors; the entire delta is the new file
- every color is either light/dark paired or the theme-neutral `text-gray-400` that is the
  **established idiom of this exact directory** (PortalSidebar 10 uses, PortalConversation 9,
  PortalFilesPanel 9); the new file's 26 is *below* its siblings (PortalRoom 46, PortalConversation 59)
- the scanner is measurement-only by its own docstring ("Exit code is always 0"), and is
  referenced by **no workflow**

Converting only this one file to tokens would make it diverge from its four immediate
neighbours. Flagging rather than churning; a directory-wide pass is the right shape and is
not this issue.

1. Delete the orphaned `src/frontend/src/components/rooms/` — zero importers.
2. Settle the recruiting question by observing the **participant list** after a
   non-participant mention (see above).
3. Unify the three typeahead implementations behind one primitive — requires first putting
   the untested operator surface under test.
4. The `MENTION_RE` vs `sanitize_agent_name` divergence (cross-repo: the grammar and the
   name sanitizer disagree about `.` and length).
5. A room-scoped variant of the "shared with you" empty-state wording.

## CI note

`frontend-build.yml` runs on this PR but is **not** a required check on `dev`. Its green
should be **read**, not relied on.

### DIFF
--- a/docs/memory/architecture.md
+++ b/docs/memory/architecture.md
@@ -1339,10 +1339,14 @@ degrade-to-plain-text this feature exists to close; and **a plain Enter never ac
 without an explicit selection**, since an accidental accept destroys typed work while an
 accidental send is what the user was reaching for. `@` is hidden — popup *and*
 placeholder — without the rooms capability, which it reads from the roster payload per
-the rule below. The **room** composer gets `@` scoped to its **agent participants**
-(the engine's `resolve_mentions` wakes only those; a non-participant mention is neither
-recognised nor recruited, verified against a running instance), while `/`-in-room is
-deferred — a room has no active-agent subject. **No backend change, no new endpoint, no
+the rule below. The **room** composer gets `@` scoped to its **agent participants** —
+established by *observing* the running server rather than reading the private rooms
+engine (`POST /api/rooms/{id}/messages` answered `woke: ["<participant>"]` for a
+participant and `woke: []` for a non-participant), so the list contains only names a
+pick is known to wake; whether a non-participant mention still joins someone by the
+ent#361 engine-side path (§5.12) is deliberately not claimed either way, and recruiting
+stays with the explicit "+ Add agent" control. `/`-in-room is deferred — a room has no
+active-agent subject. **No backend change, no new endpoint, no
 migration. OSS-core by decision (ent#392): deliberately ungated — no
 `requires_entitlement`, logic stays in the OSS tree. Recorded explicitly because
 CLAUDE.md's default for an enterprise-tracker feature is gated unless ruled otherwise, so
--- a/docs/memory/architecture.md
+++ b/docs/memory/architecture.md
@@ -1320,6 +1320,35 @@ in-place "Show all N" toggle — deliberately **no nested scroll region**: the c
 stays the single scroll axis, and the toggle counts the shipped list, never claiming the
 agent's full skill set). Hint *curation* stays the connector allow-list (ent#178 later).
 
+**Composer typeahead — `/` playbooks, `@` agents (ent#392).** The same briefing payload
+now also feeds a composer typeahead, which is what makes it reachable after turn 1 (the
+hint cards render on the empty-chat screen only). `/` lists the active agent's
+`playbooks[]` and splices the `starter_prompt` in **without sending** — the ent#138
+prefill contract; `@` lists reachable agents and inserts a token `mentionedAgents()`
+resolves (ent#361). All decidable logic is pure and exported from
+`components/portal/portalUtils.js` (`vitest` runs `environment: 'node'` with no
+component-mount harness, so a decision inside a component is one no test can reach); the
+components are dispatchers over it and `PortalTypeahead.vue` is presentational. Three
+properties are load-bearing rather than stylistic: the **trigger rule is stricter than
+the parser** (`MENTION_RE` is unanchored, so `user@example.com` parses as `@example` —
+the popup must never open on something the parser would not see); **un-mentionable slugs
+are excluded**, the predicate *derived* by asking `mentionedAgents` rather than copying
+the grammar, because `sanitize_agent_name` keeps `.` and caps nothing while the grammar
+allows neither, so listing `data.scout` would manufacture the silent
+degrade-to-plain-text this feature exists to close; and **a plain Enter never accepts
+without an explicit selection**, since an accidental accept destroys typed work while an
+accidental send is what the user was reaching for. `@` is hidden — popup *and*
+placeholder — without the rooms capability, which it reads from the roster payload per
+the rule below. The **room** composer gets `@` scoped to its **agent participants**
+(the engine's `resolve_mentions` wakes only those; a non-participant mention is neither
+recognised nor recruited, verified against a running instance), while `/`-in-room is
+deferred — a room has no active-agent subject. **No backend change, no new endpoint, no
+migration. OSS-core by decision (ent#392): deliberately ungated — no
+`requires_entitlement`, logic stays in the OSS tree. Recorded explicitly because
+CLAUDE.md's default for an enterprise-tracker feature is gated unless ruled otherwise, so
+the ruling must never be inferred later from the mere fact that it merged.** See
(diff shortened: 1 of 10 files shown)


vybe (approved):
Validated via /validate-pr. Cross-tracker Fixes Abilityai/trinity-enterprise#392 resolves; OSS-core ruling recorded in four places per the ent#326/ent#384 discipline. Merges clean onto dev after #2171; combined result verified locally (550 frontend tests, prod build, check:tokens all pass). No new v-html — the diff's v-html hits are docs plus guards that pin the conversation count at 1 and assert none in the popup. Security scan clean.

PR #2173, opened 2026-08-13
fix(workspace): Reports tab renders through the shared display_hint renderers, never a raw dump (#2162)

### LINKED ISSUE #2162: fix(workspace): Reports tab shows raw JSON to clients — reuse the display_hint renderers

trinity-ability:
Fixes #2162

## This is a disclosure fix, not a cosmetic one

The Workspace Reports tab dumped an agent-authored payload key-for-key to an **external client**. That is the same class the sibling surface already refuses: `agent_page.py` deliberately withholds the free-form `asks.context` because it "has been a credential-leak surface before (canary G-04)". Reports carry an equally free-form, equally agent-authored blob, and shipped it whole to the agent's customer.

The tab now renders through the shared `display_hint` renderers, with a client-facing summary fallback instead of a `<pre>`.

## What this does NOT fix — stated precisely

Typed renderers **narrow** what crosses: each reads only its declared keys, so an unexpected sibling key is not rendered. The fallback **bounds, humanises and redacts**. Neither **eliminates** the class:

- a key-value summary still names every top-level key;
- a well-shaped `markdown` / `table` / `kpi` / `timeline` report still renders its values **as authored**.

The universal control is a **boundary-side scrub** at the portal serialization edge, deliberately out of scope here and listed as the first follow-up. Please do not read this PR as closing the leak class.

The redactor is **defense-in-depth on the fallback path only**. It mirrors the prefix set in `canary/invariants/g04_*.py` and consumes the whole token. That is a limit, not protection — it does nothing for the four typed paths above.

## Operator surfaces are behaviourally unchanged

`git diff origin/dev...HEAD -- ReportsPanel.vue ReportsPanelFleet.vue` is **empty**. `ReportRenderer`'s new `fallbackComponent` prop defaults to `null` and resolves `props.fallbackComponent || ReportJson`, so operator call sites — which pass nothing — reduce exactly to pre-branch behaviour.

The human-readable fallback is **portal-scoped by design**. AC #2 asks for a fallback *"deliberately stricter than the operator side, because the audience is an external client."* A raw dump is a feature when you are debugging your own agent and a defect when the reader is that agent's customer.

The portal override also covers an agent-chosen **`display_hint: "json"`** — a valid value in the MCP tool's enum — not merely a shape mismatch. That closes a second route to a raw dump.

## The only operator-visible deltas

Three dark-ink meta-text pairs and one `bg-blue-500 → bg-status-info-500` swap that resolves to the **identical** colour (`status-info` aliases `colors.blue`).

The dark fix was **necessary, not cosmetic**. Measured on rendered DOM, those sites **failed WCAG AA in dark**:

| site | before | after | threshold |
|---|---|---|---|
| table thead / footer | 3.67 | **6.99** | 4.5 |
| KPI tile label | 3.04 | **5.78** | 4.5 |

Light is unchanged at 4.83.

## Backend surface

Two optional query params (`rows_offset`, `rows_limit`) on the **existing** portal detail route. No new route, gate, `db.` accessor, table or column ⇒ **CLAUDE.md Rule #9 dual-track does not engage** — no SQLite migration, no Alembic revision.

Verified mechanically: nothing touched under `db/`, `migrations/`, `schema.py`, `tables.py`, `models.py`, `main.py`; **zero `@router.` changes**; zero `Depends(` changes **in Python** (the 6 `Depends(` hits in the diff are all prose in docstrings and markdown explaining why the operator `/rows` route is JWT-gated).

`REPORT_ROWS_PAGE_MAX` is **imported** from `models.py:503` where it already lives — never re-typed — so the two page sizes cannot drift with each side's tests pinning its own version.

Auth and scoping are inherited from the existing route. The **uniform-404 property (Invariant #8) survives**: `report_detail` returns `None` on `row["agent_name"] != agent_name` *before* any windowing runs, so a foreign report id is still indistinguishable from a nonexistent one. Pinned by `test_missing_and_foreign_stay_indistinguishable_when_windowed`.

## The route gains a rate limiter it did not previously have

`services/rate_limiter.py::enforce`, key `portal_report_detail:{email}:{agent_name}`, 60/60s, placed **after** `_require_roster` so an unvalidated path param can never mint unbounded limiter keys.

The windowed read re-reads and re-parses the ≤5 MiB blob per page, so the route that exists to cut **transfer** raises **reads**. Amplification is therefore strictly *better* than pre-branch, which had no limit at all.

Residual, stated: fail-open on a Redis outage, and the key is per-agent, so a client rostered on N agents scales linearly.

## Verification

- `verify-local` **PASS** (`--skip-agent`; agent stages correctly skipped — no `docker/base-image/**` change).
- **Live both-themes visual pass** on this branch's own dev server, all six report shapes — **`<pre>` count 0 in every case**.
- **Row window end-to-end**: "Showing 100 of 260 rows" → Load-more → "Showing 200 of 260". **Without** `rows_limit` the response is byte-for-byte back-compatible.
- At **414px** a wide table absorbs 658px of overflow inside its own container while the page never scrolls horizontally.
- The operator/client split was demonstrated **on the same report simultaneously** — operator shows the raw `<pre>`, portal shows the summary.
- `npm run check:tokens` → OK, "dark ink ladder holds in 13 swept files".
- Production `vite build` succeeds.

### Tests

- `cd src/frontend && npx vitest run` → **25 files / 428 tests passed**
- `pytest tests/unit/test_2162_portal_report_window.py test_1535_report_prompt_guidance.py test_918_report_endpoint.py -q` → **32 passed**
- Full backend sweep on this branch (measured in WAVE-4): **9980 passed, 3 failed, 17 skipped, 1 xfailed**. The 3 are the **pre-existing** `[XPASS(strict)]` reds on `dev`, reproduced on a pristine `origin/dev` checkout — not from this branch. The +20 over sibling branches is this branch's own new tests.

The 14 backend tests are overwhelmingly **non-happy-path**: foreign report, missing/foreign indistinguishability, read-failure→404, limit clamping, negative offset, offset past end, non-mutation of the source payload, rate limiting, route param bounds, and FastAPI route buildability.

## AC #4 is not machine-guarded — please read this bit

`check:tokens`' `INK_LADDER_SWEPT` covers exactly **13** files and includes **neither** `components/reports/` **nor** `components/portal/`. CI therefore **structurally cannot see** the theme work in this PR; it was verified by hand with the measured contrast above.

Suggest adding both directories to the swept list so the fix is *guarded* rather than re-earned. They are clean now, so it should pass as-is.

## Honest note on the raw-colour ratchet

This branch **adds 21 raw-gray class occurrences** (`ReportSummary.vue` +14, `PortalAgentPage.vue` +4, `ReportTable.vue` +2, `ReportKpiTiles.vue` +1; `ReportTimeline.vue` −1 non-gray), which nudges the count of files exceeding `raw-color-baseline.json` from **19 → 20**.

That increase **is the arithmetic of the dark-mode fix itself**: repairing `text-gray-500` to `text-gray-500 dark:text-gray-400` necessarily adds a second raw-gray occurrence. The scanner counts occurrences, while the design system's both-themes rule *requires* adding the dark variant — so the metric moves the wrong way for a change that strictly improves compliance.

No conversion is available: `tailwind.config.js` defines semantic tokens for `status-*`, `state-*`, `brand-*`, `accent-*`, `action-*` — **semantic roles only**. Neutral ink has no token (`gray` is deliberately kept as the raw palette), which is exactly why `bg-blue-500 → bg-status-info-500` *was* converted (info has a token) and the gray pairs were not (they have none).

## Rebase expectation

`PortalAgentPage.vue`, `workspace-agent-page.md` (`## Files` table) and `core-agent.md` §5.11 each have a second writer this wave (#2169). Verified disjoint — my `PortalAgentPage.vue` hunks are all at old line ≥192, #2169's Overview block is 103–191. Resolution is **"take both sides"**.

1. **Boundary-side G-04 scrub** at the portal serialization edge — the only fix that covers `table` / `markdown` / `kpi` / `timeline`.
2. Operator stores branch on the **literal** `display_hint`, so a prefix-resolved table never pages.
3. The operator `/rows` **400 dead-ends its own UI**.
4. `scan-raw-colors.mjs` is **not CI-wired** and its baseline is stale (generated 2026-07-31); measured today, **19 files on pristine `origin/dev` already exceed it**. It also has no self-compare mode — only `--baseline` (write) and `--json` — so the "may only shrink" ratchet in CLAUDE.md Rule #10 is currently unenforced by anything.
5. Renderer-set fallthrough attrs stringify a function into a DOM attribute (pre-existing; fix is `inheritAttrs: false` across the set).

### DIFF
--- a/docs/memory/architecture.md
+++ b/docs/memory/architecture.md
@@ -650,7 +650,13 @@ directly). Agents call the MCP `report` tool, which POSTs to `POST /api/agents/{
   (`offset`/`limit`, true `total`); the UI fetches tabular reports through it, so expanding
   a 1.2 MB report transfers ~8 KB. Storage stays a single TEXT blob — no migration — and
   the slice is Python-side, so it bounds the response, not the read; off-row storage waits
-  on a payload distribution that justifies it.
+  on a payload distribution that justifies it. The **portal** detail route carries the same
+  window as two optional query params (`rows_offset`/`rows_limit`, #2162) rather than a second
+  route: `/rows` is `Depends(get_current_user)`, which a portal principal cannot satisfy, and
+  a clone on a client-facing prefix would need its own copy of the uniform-404 contract. There
+  the *server* decides tabularity from the real payload (a non-tabular one returns whole with
+  no `row_meta`, never a 400), and because paging re-reads the blob per request it is
+  rate-limited per (client, agent).
 - **Export** (#1536): `GET /api/reports/{id}/export?format=xlsx|pdf` →
   `services/report_export.py` (pure builders, lazily-imported `openpyxl`/`reportlab`, both
   pure-Python wheels). Shape mismatch degrades to a sensible sheet or JSON rather than
@@ -661,7 +667,21 @@ directly). Agents call the MCP `report` tool, which POSTs to `POST /api/agents/{
   `GET /api/reports/{id}` returns the full payload, lazy-loaded when a card expands.
 - **Fleet access**: `GET /api/reports` + `GET /api/reports/stats` filter via
   `accessible_agent_names` + `_narrow_to_agent` (admin = all). Renderers (`components/reports/`)
-  pick by `display_hint` → `report_type` prefix → JSON, with shape-validation fallback to JSON.
+  pick by `display_hint` → `report_type` prefix → fallback, with a shape check per hint.
+- **Three renderer surfaces, one fallback** (#2162): Agent Detail, the Operations fleet tab,
+  and the **Workspace agent page** all mount the same `ReportRenderer`. The third was added
+  after it shipped `JSON.stringify(payload)` to external clients — a disclosure defect
+  (`payload` is free-form agent JSON of the class `client_portal/agent_page.py` refuses to
+  expose for an ask's `context`, canary G-04), which a typed renderer narrows because it
+  reads only the keys its hint declares. This matters to the CI pin above: `test_1535`
+  regexes `payload.X` out of `ReportRenderer.vue`, so a third consumer widens that drift
+  guard's blast radius and **`shapeOk` must stay in that file** — extracting it is the
+  natural refactor and it empties the pinned set. The fallback is now `ReportSummary`
+  (bounded, humanised, credential-shaped tokens redacted at value level) with raw JSON behind
+  a collapsed operator-only disclosure; `allow-raw="false"` removes it entirely on the
+  client-facing surface. The prop names the POLICY, not the mechanism, because `json` is a
+  valid agent-chosen `display_hint` — a fallback-override prop or slot would have left an
+  agent able to request a raw dump in front of a client.
 - **Agent read-back** (#1538): `list_reports` / `get_report` MCP tools over the existing
   access-controlled REST endpoints — no new endpoint, no new tenant-boundary logic. The
   MCP layer adds the narrowing the backend cannot do (agent key → owner scope → `{self} ∪
--- a/docs/memory/architecture.md
+++ b/docs/memory/architecture.md
@@ -676,12 +676,15 @@ directly). Agents call the MCP `report` tool, which POSTs to `POST /api/agents/{
   reads only the keys its hint declares. This matters to the CI pin above: `test_1535`
   regexes `payload.X` out of `ReportRenderer.vue`, so a third consumer widens that drift
   guard's blast radius and **`shapeOk` must stay in that file** — extracting it is the
-  natural refactor and it empties the pinned set. The fallback is now `ReportSummary`
-  (bounded, humanised, credential-shaped tokens redacted at value level) with raw JSON behind
-  a collapsed operator-only disclosure; `allow-raw="false"` removes it entirely on the
-  client-facing surface. The prop names the POLICY, not the mechanism, because `json` is a
-  valid agent-chosen `display_hint` — a fallback-override prop or slot would have left an
-  agent able to request a raw dump in front of a client.
+  natural refactor and it empties the pinned set. The fallback is **per-surface**: the default
+  stays `ReportJson`, so both operator surfaces render exactly what they always did, and only the
+  Workspace passes `:fallback-component="ReportSummary"` (bounded, humanised, credential-shaped
+  tokens redacted at value level, no raw payload reachable behind it). AC #2 asks for a client
+  fallback "deliberately stricter than the operator side", so the split IS the design — a global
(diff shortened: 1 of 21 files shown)


vybe (approved):
Validated via /validate-pr. Merges clean onto dev after #2171 and #2172; combined result verified locally (27 files / 550 frontend tests, 20/20 test_2162_portal_report_window.py, prod build, check:tokens all pass). Backend reviewed: ownership check precedes windowing so the uniform-404 (Invariant #8) survives, rows_offset/rows_limit bounded at both route and service layer, shallow copy leaves the source payload intact, rate limiter placed after _require_roster. Operator surfaces confirmed byte-identical (empty diff on both panels; fallbackComponent defaults null and resolves to ReportJson). No new endpoint/table/column, so Rule #9 dual-track correctly does not engage. Security scan clean — flagged token patterns are all fake fixtures (AWS's documented AKIAIOSFODNN7EXAMPLE, .test TLD, redaction-test strings).

PR #2176, opened 2026-08-14
fix(voice): stop agents telling Workspace clients the surface is text-only (#2157)

### LINKED ISSUE #2157: bug(voice): agents tell Workspace clients the surface is text-only — portal TTS is never advertised, and the channel/portal voice gates disagree
## Summary

The Workspace (Client Portal) speaks agent replies aloud, but it does so **client-side**: after a reply arrives, the portal UI calls `POST /api/enterprise/client-portal/agents/{agent}/tts`, gets an MP3 and plays it in the browser, gated on a speaker toggle in the conversation header. The agent is not involved, produces no audio artifact, and is never told this surface exists.

The only voice capability an agent knows about is the `send_voice_reply` MCP tool, which delivers an **audio artifact** into a messaging channel and is hard-restricted to Telegram/Slack/WhatsApp. So when a portal client asks for a spoken reply, the agent tries that tool, gets `not_a_channel_turn`, and — reasoning correctly from what it was told — **over-generalizes from its own limitation to a false claim about the surface**: that the portal is text-only, and that the client must go to another channel to be spoken to. Neither is true.

## What is actually true, and what is false

This distinction is the whole issue, and getting it wrong in either direction produces a new bug. Observed live on v0.8.5, with the platform ElevenLabs key and the agent's `tts_voice_id` both set. Agent reply, verbatim excerpt:

> I gave it a try, but voice replies aren't available on this channel. Voice notes only work over messaging channels like Telegram, Slack, or WhatsApp — and right now you're talking to me through the **Client Portal (web)**, which is text-only. The system confirmed it: `delivered: false, reason: not_a_channel_turn`
>
> So I can't send you an audio reply here […] If you want a spoken reply, reach me on Slack, Telegram, or WhatsApp (if this agent is connected there) and I can voice it back.

| Claim | Verdict |
|---|---|
| "I can't send you an audio reply here" | **TRUE.** There is no path by which an agent delivers an audio artifact into the portal. `send_voice_reply` is channel-only by design, and nothing else exists. |
| "voice notes only work over messaging channels" | **TRUE** for voice *notes* — an audio artifact in the transcript. |
| "the Client Portal (web) … is **text-only**" | **FALSE.** The surface ships both TTS and STT. With the speaker toggle on, this same reply was read aloud a few seconds after it appeared. |
| "If you want a spoken reply, reach me on Slack, Telegram, or WhatsApp" | **FALSE remedy.** The client can hear the reply on the surface they are already using, by toggling the speaker icon. |
| `delivered: false, reason: not_a_channel_turn` | Internal reason string relayed verbatim to an external client. |

Observed behaviour with the toggle on: the agent replies in **text**, and a few seconds later that same text begins playing as speech. The narration is ephemeral browser playback of the assistant's text — not a file, not a message, and not something the agent triggers or can observe.

**So the fix is not to tell the agent it can send voice in the portal — it cannot.** The fix is to tell it that this surface *narrates its text when the client turns narration on*, so it stops describing the surface as text-only and stops routing clients to other channels for something available where they are.

## Root cause

Four facts, each individually correct, that combine into a wrong answer. Line numbers as of `dev` @ `d2f8e685`.

**1. The tool is gated to channel turns** — `src/backend/routers/agents.py:1267-1270`:

```python
# Only user-facing channel turns have a deliverable voice destination.
channel = execution.source_channel or execution.triggered_by
if channel not in ("telegram", "slack", "whatsapp"):
    return {"delivered": False, "channel": channel, "reason": "not_a_channel_turn"}
```

A portal turn runs `triggered_by="public"`, so it can never satisfy this — correctly, since there is no audio destination.

**2. The reason string reaches the agent verbatim** — `src/mcp-server/src/tools/voice.ts:111` returns `reason: result.reason` straight through, so `not_a_channel_turn` lands in the agent's context and it reasons from it. The name describes a *transport* fact and invites the inference "this surface has no voice at all."

**3. Nothing advertises the portal's narration.** `build_voice_capability_prompt()` (`src/backend/services/platform_prompt_service.py:818`) has exactly **one** caller — `src/backend/adapters/message_router.py:691`, the channel path. The portal path never calls it. `_build_portal_system_prompt()` (`src/backend/client_portal/service.py:799`) composes only the MEM-001 memory block and the #1205 public-channel prompt; voice is mentioned nowhere.

**4. Narration is client-side and off by default** — `src/frontend/src/components/portal/PortalConversation.vue`:

```js
const ttsEnabled = computed(() => !!props.agent.voice_available)
const voiceMode = ref(false)  // off on every load
// :435 — after the text reply is pushed:
if (voiceMode.value && ttsEnabled.value && data.response) speak(data.response)
```

### Secondary defect: the two voice paths gate differently, in both directions

| Gate | Channel (`services/voice_reply_service.py:71-79`) | Portal (`client_portal/service.py:305,487-491`) |
|---|---|---|
| Platform TTS key | checked | checked |
| Agent-level `tts_voice_replies_enabled` | checked | **ignored** |
| Per-channel flag | checked | n/a |
| Voice id | agent's own **or platform default** | agent's own **only** |

Two consequences:

- **This makes the primary bug worse than it first appears.** An agent relying on the platform default voice (`tts_voice_id` NULL) speaks fine on Telegram but its portal speaker toggle **never renders at all** — `voice_available` requires the agent's own id. So the agents most likely to produce the misleading answer are exactly the ones where the client has no toggle to reach for.
- **Inverse leak:** an agent whose operator explicitly *disabled* voice replies is still narrated in the portal, because the portal checks neither the agent-level enable nor any per-surface flag. Operator intent is silently ignored.

## Impact

- Clients are told a working, paid feature does not exist — by the agent itself, which is the most credible source in that UI. The agent's *self*-report is accurate, which is what makes the surrounding claim so convincing.
- The suggested workaround pushes an external client off the deliberately client-facing surface and onto operator channels, for something already available where they are.
- Discoverability is already weak: the toggle is off on every load, and invisible entirely for agents on the platform default voice. The agent's denial actively reinforces the wrong conclusion.
- Internal reason strings leak to external clients.

## Acceptance Criteria

- [ ] A portal/workspace turn injects a capability fragment telling the agent that this surface narrates its replies aloud when the client enables it, and that narration is client-controlled rather than agent-triggered
- [ ] **The fragment must not imply the agent can send an audio artifact in the portal** — it cannot, and claiming otherwise would replace one wrong answer with another
- [ ] Asked for a spoken reply in the portal, an agent no longer describes the surface as text-only and no longer redirects the client to another channel; it points at the speaker toggle instead
- [ ] The fragment is set by the platform from the turn's surface, never asserted by the caller
- [ ] Covered by a test asserting the portal prompt carries the fragment and a channel turn is unchanged
- [ ] `send_voice_reply` returns a portal-specific, actionable reason (e.g. `portal_client_narrated`) rather than `not_a_channel_turn` when the turn originates in the portal, so an agent reasoning from the tool result alone still answers correctly
- [ ] Agents explain the outcome to clients rather than echoing the machine reason string
- [ ] Decide and implement consistent gating across the two paths: whether portal narration honours `tts_voice_replies_enabled`, and whether `voice_available` should fall back to the platform default voice as the channel path does — with the chosen behaviour covered by tests in both directions
- [ ] Channel turns keep their current behaviour, verified by test
- [ ] *Optional UX:* the client's voice-mode choice persists per client/agent instead of defaulting to off on every page load

## Technical Notes

- The fix rides the existing platform prompt composition (`services/platform_prompt_service.py`), which is already runtime-aware — the portal counterpart to `build_voice_capability_prompt()`, but describing a *client-controlled surface affordance* rather than an agent-invocable tool. That difference is worth carrying in the helper's name.
- **Shares its mechanism with `trinity-enterprise#363`** ("tell an agent when a room is user-facing"), which injects a different surface-awareness signal into the same prompt seam. These should almost certainly land together or reuse one helper rather than growing two parallel injection paths.
- **`trinity-enterprise#285`** ("live voice conversation for external clients") is *not* a duplicate — it covers live back-and-forth Gemini Live conversation. It does corroborate the premise, recording that turn-based portal `/tts` and `/stt` already ship.
- The channel gate has been touched by exactly one commit since it was written (`8e03dacb`, voice replies v2), and no open PR in either tracker touches this path — so there is no in-flight or landed fix to rebase onto.
- Reproduces on any instance with the Workspace/portal surface, a platform ElevenLabs key, and an agent with a per-agent `tts_voice_id`. The core defect is static logic and is verifiable by reading the call graph — no ElevenLabs key is needed to confirm it.

dolho:
## What this fixes

Asked for a spoken reply in the Workspace, an agent tried `send_voice_reply`, got
`not_a_channel_turn`, and told the client the surface is **text-only** and that they
must go to Slack/Telegram/WhatsApp to be heard. Its own limit was real — no path
delivers an audio artifact into the portal. Both claims about the *surface* were false:
the browser reads the reply aloud whenever the client switches the speaker on.

The fix adds exactly one fact — *this surface narrates* — without adding a second false
one. Telling the agent it can send audio here would replace one wrong answer with another,
so the fragment says the opposite explicitly (AC guard from the issue's follow-up comment).

## Changes

| Seam | Change |
|---|---|
| Prompt | `platform_prompt_service.build_narrated_surface_prompt()` — the portal counterpart to the channel path's tool advertisement, describing a **client-controlled affordance**, not an agent-invocable tool. Composed into the portal caller prompt by `_build_portal_system_prompt`; platform-set from the turn's surface; unreachable from the channel path. |
| Tool result | `POST /voice-reply` answers a portal turn `portal_client_narrated` + a plain-language `guidance` line — or `portal_voice_not_configured` when no voice resolves, so the guidance never points at a speaker control that does not render. MCP tool surfaces `guidance` and is told never to quote a `reason` code to a user. |
| Gate | `tts_service.resolve_voice_id()` / `resolve_voice_from_config()` — one rule for the roster card, the portal `/tts` endpoint, the prompt, and the channel path: platform key AND agent-level `tts_voice_replies_enabled` AND own voice else the platform default. |
| Surface id | Portal executions carry `source_channel="portal"` (`config.PORTAL_SOURCE_CHANNEL`) at both creation sites. `triggered_by="public"` is shared with public links and x402 and cannot identify the surface. Not a messaging channel: no chat id, so channel consumers already ignore it. |
| UX | The client's speaker choice persists per client+agent instead of resetting to off on every load. |

## ⚠️ Behaviour change, both directions

The issue left the gating decision open ("decide and implement consistent gating"). This PR
makes the portal use the channel gate minus the per-channel flags:

- an agent riding the **platform default voice** now gets a speaker control — it had **none**
  before (`voice_available` required the agent's own `tts_voice_id`), which is why the
  misleading answer landed hardest on exactly the clients with nothing to reach for;
- an agent whose operator **turned voice off** is no longer narrated in the portal — it was,
  ignoring operator intent.

That second one is reachable in a single API call today (`PUT /voice-replies {"enabled": false,
"voice_id": "…"}` keeps the id), verified live below. **Worth a release note.**

Branch code run against the live Postgres/Redis with the live backend's env; state changed
through the real API and restored afterwards.

**Same execution row, before and after** (`triggered_by=public, source_channel=portal`):

```
live instance (dev)  → {"delivered": false, "channel": "portal", "reason": "not_a_channel_turn"}
this branch          → {"delivered": false, "channel": "portal", "reason": "portal_client_narrated",
                        "guidance": "…it is not text-only: the client hears your reply read aloud
                         when they switch on the speaker control… do not send them to another channel…"}
```

**Gate on real rows**, driven through `PUT /voice-replies`:

| live state | portal OLD | portal NEW | channel path |
|---|---|---|---|
| enabled, own voice id | True | True | True |
| disabled, voice id kept | **True** | **False** | False |
| enabled, no own voice, platform default set | **False** | **True** | True |

"portal gate == channel gate" held in every state. Portal `/tts` (vendor call stubbed):
synthesizes with the platform default when enabled, `404 'This agent has no voice configured'`
when the operator disabled it.

**End-to-end, real agent, real Workspace endpoint** — asked *"Can you read your replies out
loud to me here? If not, what should I do?"*:

> I can't control audio playback myself. To have my replies read aloud, look for the **speaker
> control** in this conversation interface — toggling it on will have your browser read my text
> replies as they arrive.

Its own limit kept, both false claims gone.

## Tests

- `tests/unit/test_2157_portal_narration.py` — 29 tests: gate truth table both directions,
  fragment present/absent + the two over-correction guards, portal prompt composition vs an
  unchanged channel fragment, both portal tool answers, channel turn still delivers, both stamp sites.
- `src/frontend/tests/unit/portalVoiceModePersistence.spec.js` — 5 source guards (no mount harness in repo).
- 420 related backend tests, 337 frontend tests, frontend build, MCP `tsc` build: green.
- Full backend unit suite: running at time of opening; will report.

## Known gaps

- **Rooms turns** (`triggered_by="room"`, private `shared_sessions`) carry neither the stamp nor
  the fragment — the shared-mechanism follow-up the issue names (`trinity-enterprise#363`).
- The final AC (an agent *behaving* correctly) is proven by the live transcript above, not by a
  deterministic test — an LLM turn cannot be pinned in CI.

Fixes #2157


dolho:
Full backend unit suite (promised in the PR body): **9991 passed, 14 skipped, 1 xfailed, 3 failed** in 55m.

The 3 failures are `test_pat_propagation_properties.py::TestKnownGaps::test_a_backslash_in_the_token_does_not_raise` — strict-`xfail` markers that now XPASS, i.e. the #1979 backslash-in-token bug was fixed but the markers were never removed. **Pre-existing on `dev`**, not caused by this branch: verified on a clean dev-based worktree with zero diff to `services/github_pat_propagation_service.py`, where the same 3 fail identically. Nothing in this PR touches that path.

Worth a small separate cleanup issue (stale strict-xfail markers fail every full-suite run).

vybe (approved):
Validated via /validate-pr. CI green across all 25 checks (3 pytest seeds base+head, e2e, frontend build, MCP tsc). Still mergeable and green after #2177 landed — no code overlap, architecture.md hunks are in disjoint regions.

The design call is right: the fragment states the surface narrates while explicitly denying the agent can send audio here, so it does not replace one false claim with its mirror image. `tts_service.resolve_voice_id` / `resolve_voice_from_config` giving the roster, portal /tts, prompt and channel path one gate is the durable part; the pure variant keeps the roster from becoming an N+1. Both portal tool answers split on the same gate the speaker control renders on, so guidance can never point at a control that is not there.

Acknowledging the two-directional behaviour change as deliberate and documented: an agent on the platform default voice gains a speaker control it never had, and an agent whose operator disabled voice stops being narrated in the portal. Both move the portal onto the channel gate, which is the consistent answer the issue asked for. Needs a release note — flagging for the cut, not blocking here.

PR #2177, opened 2026-08-14
fix(a2a): delete one endpoint, not two — a ref removes exactly what it resolves (#2174)

### LINKED ISSUE #2174: bug: deleting one A2A endpoint can remove two, destroying a stored credential
## Summary

`remove_endpoint(ref)` in `src/backend/services/a2a_outbound.py` filters out **every** record whose id **or** name matches the reference:

```python
keep = [
    r for r in records
    if str(r.get("id") or "") != wanted
    and str(r.get("name") or "").lower() != lowered
]
```

Nothing enforces uniqueness *across* those two namespaces, and endpoint ids are visible in the admin `GET /api/settings/a2a-endpoints` response. So an operator who has named one endpoint after another endpoint's id can issue a single-target `DELETE /api/settings/a2a-endpoints/{ref}` and silently destroy **two** endpoints — along with their AES-256-GCM-encrypted credentials — while the route's own docstring says "Remove **one** registered outbound A2A endpoint by id or name" and the function returns a single `True`.

## Context

Found by an `/edge-cases` sweep of the #736 A2A outbound path. This is the cross-namespace-collision class the tracker has hit before (#1631 operator-queue ids: uniqueness assumed globally but enforced per-agent).

The blast radius is what makes it worth fixing rather than documenting: the destroyed record's `credential` is a partner secret the operator may not still have a copy of, and the deletion is reported as success, so there is no signal to re-check. The next call against the collaterally-deleted endpoint fails with `endpoint_not_found`, which reads as a registration problem rather than a deletion.

`upsert_endpoint` enforces case-insensitive uniqueness on **name** only; it never checks a proposed name against existing **ids**, so creating the collision is a normal, permitted operation.

### Reproduction

```
alpha = upsert_endpoint("alpha", url, "alpha-secret")     # -> id a2aep_<hex12>
        upsert_endpoint("beta",  url, "beta-secret")
        upsert_endpoint(alpha["id"], url, "gamma-secret") # legal: name == alpha's id

list_oss_endpoints()          -> ['alpha', 'beta', 'a2aep_<hex12>']   (3)
remove_endpoint(alpha["id"])  -> True
list_oss_endpoints()          -> ['beta']                             (1)   <-- two gone
```

## Acceptance Criteria

- [ ] `remove_endpoint(ref)` removes at most one record per call
- [ ] With an id/name collision present, removing by id removes the id-matched record and leaves the name-matched one intact (resolution order is documented and the delete agrees with `resolve_endpoint`, which returns the *first* match)
- [ ] `remove_endpoint` still returns `False` on a genuine miss and for an empty/whitespace ref
- [ ] Removing by name still works for the ordinary no-collision case
- [ ] The strict-`xfail` on `tests/unit/test_736_a2a_outbound_edges.py::test_G7_removing_one_endpoint_removes_exactly_one_endpoint` is removed and the test passes plainly

## Technical Notes

Two candidate shapes, and they are not equivalent — pick deliberately:

1. **Remove the first match only**, mirroring `resolve_endpoint`'s first-match-wins semantics. This keeps delete and resolve consistent: whatever a `ref` *resolves* to is what it *deletes*. Preferred.
2. **Refuse an ambiguous ref** (400/409, naming both candidates). Safer, but introduces a state the operator can reach and cannot exit without renaming, so it needs a rename path to be usable.

Worth considering alongside: `upsert_endpoint` could reject a name that collides with an existing id, which removes the ambiguity at the source. That is a separate, additive guard — it does not fix already-stored collisions, so it is not a substitute for fixing the delete.

Related: the same sweep found `upsert_endpoint` treats a whitespace-only credential as "clear" while `None` and `""` mean "leave unchanged" — filed separately with the other low-severity findings.

dolho:
## What this fixes

`remove_endpoint(ref)` filtered out **every** record whose id **or** name matched the
reference; `resolve_endpoint` stopped at the **first**. Ids and names are separate
namespaces with no cross-uniqueness — ids are shown in the admin `GET`, and
`upsert_endpoint` enforced uniqueness on name only — so one single-target
`DELETE /api/settings/a2a-endpoints/{ref}` could destroy **two** endpoints, taking an
AES-256-GCM-encrypted partner credential the operator may hold no other copy of, while
the route returned one `{"success": true}`. The follow-on symptom misdirects as well:
the collaterally-deleted endpoint's next call fails `endpoint_not_found`, which reads as
a registration problem rather than a deletion. Same class as #1631.

## Changes

- **One shared predicate** `_record_matches_ref`, used by resolve *and* remove, so the two
  cannot drift about which record a ref means.
- **`remove_endpoint` deletes the first match and returns** — at most one record per call,
  same rule and order as resolve, so a ref deletes precisely what it resolves to
  (Technical Notes option 1). Still `False` on a genuine miss and on an empty/whitespace
  ref, and a no-op delete no longer rewrites the stored envelope.
- **Create-time guard** (additive, as the issue suggests): a **new** endpoint may not take
  an existing endpoint's id as its name → 422 through the existing
  `EndpointValidationError` mapping. Create path only — an already-stored collision stays
  editable and removable, or the guard would strand the operator in exactly the state it
  exists to prevent.

Refuse-if-ambiguous (option 2) was rejected: with no rename path it strands the operator in
a reachable, unexitable state, and it leaves the destructive behaviour in place for every
collision already stored — the case actually at risk.

**Before**, through the live admin API (URLs SSRF-validated, so `api.github.com` stands in
for a peer):

```
PUT alpha            -> a2aep_f163a8bc5e6d   (credential: alpha-secret)
PUT beta             -> a2aep_456c2e0c36ca   (credential: beta-secret)
PUT a2aep_f163a8bc5e6d -> a2aep_a35c58ec9953 (credential: gamma-secret)   # legal today
GET  -> 3 endpoints
DELETE /a2aep_f163a8bc5e6d -> {"success": true, "removed": "a2aep_f163a8bc5e6d"}
GET  -> 1 endpoint   # two gone, gamma-secret destroyed
```

**After**, same store, this branch:

```
guard: creating an endpoint NAMED after alpha's id -> refused: That name is already the id
       of another registered endpoint; pick a different name.
before : [beta, alpha, <alpha-id-named record>]  (3)     # collision planted directly,
resolve(<alpha id>) -> id=<alpha id> name=alpha          # as a pre-guard install would hold it
remove  -> True
after  : [beta, <alpha-id-named record>]  (2)
collateral record survived with its credential: True
```

The live endpoint store was empty before and is empty again after.

## Tests

`tests/unit/test_736_a2a_outbound_edges.py` — 17 cases:

- `test_G7_removing_one_endpoint_removes_exactly_one_endpoint` — the AC-named test, plain,
  no xfail. **Note:** the file the issue refers to does not exist on `dev` (`git log -S`
  finds it in no branch or commit — the `/edge-cases` sweep's file never landed), so it is
  created here rather than un-xfailed.
- delete/resolve agreement on the same ref; a second delete clears the survivor
- by-name, by-id, case-insensitive; other endpoints and their credentials untouched
- miss / empty / whitespace refs return `False` and leave the envelope byte-identical
- create-time guard, its case-insensitivity, the already-stored-collision escape hatch, and
  an ordinary name that merely looks id-ish

Green: 17 new, 130 in the three `736` A2A files, 404 across all A2A-related suites.

## Docs

`architecture.md` (the settings-route row) and `requirements/mcp.md` §32.5 gain **FR-5a**,
which records the first-match-wins rule, the rejected alternative, and the deliberate
asymmetry that remains: `upsert_endpoint` is update-**by-name** (its shipped contract) while
resolve/remove are by id-or-name, so for a colliding string the update edits the record
*named* that and the delete reaches the id-owning one. Both documented, neither destructive.

Fixes #2174


trinity-ability (approved):
Validated via /validate-pr. CI green across all 24 checks (3 pytest seeds, base and head). One shared `_record_matches_ref` predicate closes the resolve/remove divergence; first-match-wins matches resolve order so a ref deletes exactly what it resolves to. Create-time guard is correctly scoped to the create path so an already-stored collision stays removable. Docs (architecture.md settings row + requirements/mcp.md FR-5a) record the rule and the remaining upsert-by-name asymmetry. 17 named tests incl. the AC test. Security scan clean.

PR #2178, opened 2026-08-14
fix(security): mapped-CGNAT bypassed the SSRF destination predicate (ent#393)

vybe:
Fixes abilityai/trinity-enterprise#393

## Summary

- `_is_internal_address` gated its RFC 6598 CGNAT clause on `ip.version == 4`. The IPv4-**mapped** form `::ffff:100.64.0.1` reports version 6, so the clause never ran — and CGNAT is exactly the range Python's `ipaddress` answers `False` to for all six of its own predicates, which is why ent#14 S3 added the clause in the first place.
- RFC 6598 space was therefore accepted as a **public** destination by both validators built on `_validate_public_https_url`: `validate_a2a_endpoint_url`, which then sends a credentialed `Authorization: Bearer` POST (#736), and `validate_template_registry_url`.
- The clause now resolves the address's **v4 view** once (`ipv4_mapped` when v6) and tests `100.64.0.0/10` membership against that.

## Why only this one range drifted

Every *other* internal range survives its mapped form for free, because CPython short-circuits `is_private` / `is_reserved` **through `ipv4_mapped` before consulting its IPv6 tables**. CGNAT is the one range whose refusal is Trinity's own code, so it is the one that had to do that delegation itself.

The suite had already anticipated the evasion generically — it parametrised `::ffff:127.0.0.1` and `::ffff:10.0.0.1` — but those rows pass on *borrowed* delegation, so they proved nothing about the one clause that performs none. That is the durable lesson recorded in `docs/memory/learnings.md`.

## The fix is complete, not one shape in a queue

`::ffff:0:0/96` is the **only** v4-in-v6 embedding CPython can inherit an IPv4 verdict through. Verified empirically across every transition mechanism:

| shape | refused wholesale by the stdlib? | populates `ipv4_mapped`? |
|-------|----------------------------------|--------------------------|
| `::ffff:0:0/96` IPv4-mapped | **no** | **yes** — the gap |
| `::/8` IPv4-compatible (RFC 4291) / IPv4-translated (RFC 2765) | yes | no |
| `64:ff9b::/96` + `64:ff9b:1::/48` NAT64 | yes | no |
| `2002::/16` 6to4 | yes | no |
| `2001::/32` Teredo | yes | no |

Those refusals are the **interpreter's**, not ours, so they are now pinned — an interpreter change could otherwise reopen them with the whole suite still green (the #1891 class, which a version-parity check structurally cannot catch).

## A narrowing fix needs its widening guard

`test_C1b` asserts the mapped rows either side of the /10 stay public. Refusing every mapped address would be a strictly worse bug than the leak — it breaks working endpoints instead of leaking to unreachable ones.

## Changes

| File | What |
|------|------|
| `src/backend/utils/url_validation.py` | the fix, plus why `ipv4_mapped` is the complete answer |
| `tests/unit/test_736_a2a_outbound_edges.py` *(new)* | the `/edge-cases` sweep that found this; F0's strict-xfail removed, `test_C1b` added, 4 completeness rows |
| `tests/unit/test_736_a2a_outbound_properties.py` *(new)* | mapped CGNAT added to the Hypothesis `INTERNAL_ADDRESSES` list |
| `tests/unit/test_ent14_registry_url_ssrf.py` | template-registry path: 5 refusal rows + 4 mapped boundary rows |
| `tests/unit/test_736_a2a_url_validation.py` | 2 mapped rows on the shared internal-destinations parametrize |
| `docs/memory/learnings.md` | the durable bug class |

## Acceptance criteria

- [x] `_is_internal_address` refuses `::ffff:100.64.0.1`, `::ffff:100.127.255.254` and any address whose `ipv4_mapped` form falls in `100.64.0.0/10`
- [x] `validate_a2a_endpoint_url` returns `endpoint_private_address` for a host resolving only to mapped CGNAT
- [x] `validate_template_registry_url` refuses the same
- [x] Genuinely public v4 and v6 addresses still validate
- [x] The strict-xfail on `test_C1_the_mapped_form_of_cgnat_is_refused_like_every_other_mapped_form` is removed and the test passes plainly
- [x] The mapped forms are added to `test_internal_destinations_are_refused`'s parametrize list

## Test plan

- [x] Affected suites: **281 passed, 5 xfailed**
- [x] Adjacent SSRF / registry suites: **297 passed**
- [x] Full `tests/unit`: **10,173 passed**

Failure triple per `learnings.md` 2026-08-13 (a differing failure set is not by itself a regression):

| run | result |
|-----|--------|
| base `88a4e2f7` | 6 failed / 9,956 passed |
| branch − new test files | 5 failed / 9,970 passed |
| branch (full) | 3 failed + 2 errors / 10,173 passed |

`test_pat_propagation_properties::TestKnownGaps` ×3 fails in **all three** — pre-existing, order-independent. Everything else *moves* between runs: `test_1081_physical_meter` and `test_1966_ops_reset_audit` fail on **base** and pass here; `test_subscription_auto_switch_pingpong` errors only in the full run and passes 29/29 standalone. **This change introduces zero failures.**

## Known defects landing with the sweep (tracked, not silent)

This PR lands the edge-case file whose five remaining strict-xfails are real, unfixed defects. `strict=True` means none can rot silently in code, and each marker now names its tracker issue:

| finding | issue | one-line |
|---------|-------|----------|
| F1 | abilityai/trinity-enterprise#399 | `_same_origin` compares IPv6 literals textually (fail-closed) |
| F2 | abilityai/trinity-enterprise#398 | port `:0` normalised three ways along one call path (fail-closed) |
| F3 | abilityai/trinity-enterprise#397 | refusal `reason` reverse-engineered from its own message |
| F4 | abilityai/trinity-enterprise#395 | `remove_endpoint` can delete two endpoints and both credentials |
| F5b | abilityai/trinity-enterprise#396 | a whitespace-only credential clears the stored secret |

Also filed: abilityai/trinity-enterprise#394 — `mcp_validator._resolves_to_private_ip` carries **no** CGNAT clause at all, in either form. Deliberately out of scope here: it is not in this fix's blast radius, has a different threat model (config hygiene, fails closed on DNS failure, no credential on the wire), and is vendored byte-identically under Invariant #5, so fixing it is 2 files plus a parity assertion.

## Notes for review

- No new endpoint, no dependency change, no schema change, no migration. The diff only **narrows** an existing predicate.
- Refusal strings stay fixed literals — the topology-oracle property (`test_refusal_message_never_echoes_the_resolved_address`) is unchanged.
- Per CLAUDE.md tiered documentation, a bug fix needs a descriptive commit message only; `architecture.md`'s `url_validation.py` entry remains accurate as written.


trinity-ability:
Reopening under the `trinity-ability` account as requested — the local `gh-shim` routes `pr create` to the repo's commit identity (`vybe`), which overrode the active account. Same branch, same commit, no content change.

dolho:
Heads-up on a collision with what just landed on `dev` — partly my doing, so here is the exact state and the options.

**`tests/unit/test_736_a2a_outbound_edges.py` now exists on `dev`** (from #2177, merged ~09:49Z). I searched for it before writing it and found nothing — this branch was pushed the same morning and had no PR open at the time — so `dev` now carries a 17-test file at the same path this PR adds a 1356-test file to. This PR's version is the canonical one; mine should give way.

**Two pins in this PR will XPASS-strict once rebased**, because both defects are now fixed on `dev`:

- **G7** (finding F4, ent#395) — `remove_endpoint` is first-match-wins as of #2177. Its xfail must be dropped. One wrinkle: #2177 also added a create-time guard that refuses a *new* endpoint named after an existing id, which is exactly how G7's setup builds the collision (`store.upsert_endpoint(alpha["id"], …)`), so G7 would now raise `EndpointValidationError` at setup. Either plant the third record directly (`_store_endpoint_records`, which is what a pre-guard install would hold anyway — and the case actually at risk), or say the word and I will drop the guard from `dev`: it was additive, not part of the issue's AC, and I have the revert prepared.
- **D7** (finding F2, ent#398) — fixed in #2180, open now. Its xfail must be dropped when that lands. #2180 holds the same property in its own file meanwhile, so there is no unpinned window either way.

Happy to do the rebase + un-xfail work on this branch if that is easier than you carrying it.

PR #2179, opened 2026-08-14
fix(security): mapped-CGNAT bypassed the SSRF destination predicate (ent#393)

trinity-ability:
Fixes abilityai/trinity-enterprise#393

## Summary

- `_is_internal_address` gated its RFC 6598 CGNAT clause on `ip.version == 4`. The IPv4-**mapped** form `::ffff:100.64.0.1` reports version 6, so the clause never ran — and CGNAT is exactly the range Python's `ipaddress` answers `False` to for all six of its own predicates, which is why ent#14 S3 added the clause in the first place.
- RFC 6598 space was therefore accepted as a **public** destination by both validators built on `_validate_public_https_url`: `validate_a2a_endpoint_url`, which then sends a credentialed `Authorization: Bearer` POST (#736), and `validate_template_registry_url`.
- The clause now resolves the address's **v4 view** once (`ipv4_mapped` when v6) and tests `100.64.0.0/10` membership against that.

## Why only this one range drifted

Every *other* internal range survives its mapped form for free, because CPython short-circuits `is_private` / `is_reserved` **through `ipv4_mapped` before consulting its IPv6 tables**. CGNAT is the one range whose refusal is Trinity's own code, so it is the one that had to do that delegation itself.

The suite had already anticipated the evasion generically — it parametrised `::ffff:127.0.0.1` and `::ffff:10.0.0.1` — but those rows pass on *borrowed* delegation, so they proved nothing about the one clause that performs none. That is the durable lesson recorded in `docs/memory/learnings.md`.

## The fix is complete, not one shape in a queue

`::ffff:0:0/96` is the **only** v4-in-v6 embedding CPython can inherit an IPv4 verdict through. Verified empirically across every transition mechanism:

| shape | refused wholesale by the stdlib? | populates `ipv4_mapped`? |
|-------|----------------------------------|--------------------------|
| `::ffff:0:0/96` IPv4-mapped | **no** | **yes** — the gap |
| `::/8` IPv4-compatible (RFC 4291) / IPv4-translated (RFC 2765) | yes | no |
| `64:ff9b::/96` + `64:ff9b:1::/48` NAT64 | yes | no |
| `2002::/16` 6to4 | yes | no |
| `2001::/32` Teredo | yes | no |

Those refusals are the **interpreter's**, not ours, so they are now pinned — an interpreter change could otherwise reopen them with the whole suite still green (the #1891 class, which a version-parity check structurally cannot catch).

## A narrowing fix needs its widening guard

`test_C1b` asserts the mapped rows either side of the /10 stay public. Refusing every mapped address would be a strictly worse bug than the leak — it breaks working endpoints instead of leaking to unreachable ones.

## Changes

| File | What |
|------|------|
| `src/backend/utils/url_validation.py` | the fix, plus why `ipv4_mapped` is the complete answer |
| `tests/unit/test_736_a2a_outbound_edges.py` *(new)* | the `/edge-cases` sweep that found this; F0's strict-xfail removed, `test_C1b` added, 4 completeness rows |
| `tests/unit/test_736_a2a_outbound_properties.py` *(new)* | mapped CGNAT added to the Hypothesis `INTERNAL_ADDRESSES` list |
| `tests/unit/test_ent14_registry_url_ssrf.py` | template-registry path: 5 refusal rows + 4 mapped boundary rows |
| `tests/unit/test_736_a2a_url_validation.py` | 2 mapped rows on the shared internal-destinations parametrize |
| `docs/memory/learnings.md` | the durable bug class |

## Acceptance criteria

- [x] `_is_internal_address` refuses `::ffff:100.64.0.1`, `::ffff:100.127.255.254` and any address whose `ipv4_mapped` form falls in `100.64.0.0/10`
- [x] `validate_a2a_endpoint_url` returns `endpoint_private_address` for a host resolving only to mapped CGNAT
- [x] `validate_template_registry_url` refuses the same
- [x] Genuinely public v4 and v6 addresses still validate
- [x] The strict-xfail on `test_C1_the_mapped_form_of_cgnat_is_refused_like_every_other_mapped_form` is removed and the test passes plainly
- [x] The mapped forms are added to `test_internal_destinations_are_refused`'s parametrize list

## Test plan

Rebased onto `dev` at `ccc51ee` (which now carries #2180 and #2177).

- [x] Affected A2A / SSRF / registry suites: **381 passed, 3 xfailed**
- [x] Full `tests/unit` on the branch: **10,262 passed, 17 skipped, 4 xfailed, 3 failed**
- [x] Those 3 failures are `test_pat_propagation_properties::TestKnownGaps::test_a_backslash_in_the_token_does_not_raise` — **verified failing identically on base `origin/dev`** (3 failed / 19 passed standalone). Pre-existing, order-independent, untouched by this change.

**This change introduces zero failures.**

## Rebase: two of the six findings were fixed while this was open

The sweep originally landed five strict-xfails. Two are now gone because their
defects shipped to `dev` in the meantime, and a `strict=True` marker over a fixed
bug is a **CI failure**, not a stale comment — so removing them was required, not
cosmetic:

| finding | issue | fixed by | what changed here |
|---------|-------|----------|-------------------|
| **F2** | ent#398 | **#2180** | marker removed; `test_D7_…` now asserts validation, pinning and origin comparison agreeing on `:0` |
| **F4** | ent#395 | **#2177** | this sweep's xfailed `test_G7_…` removed entirely; #2177's own passing delete suite is kept **verbatim** at the end of the file (own `oss_store` fixture, constants renamed off the `PEER_URL` collision), including the create-time guard that now refuses the collision at the source |

The conflict was an add/add on `tests/unit/test_736_a2a_outbound_edges.py` — #2177
created its own file of that name after this branch was cut. Resolved by keeping
this sweep as the base and folding #2177's suite in, so neither PR loses its
regression coverage. The file header table now records per-finding state.

Three genuine defects remain pinned:

| finding | issue | one-line |
|---------|-------|----------|
| F1 | abilityai/trinity-enterprise#399 | `_same_origin` compares IPv6 literals textually (fail-closed) |
| F3 | abilityai/trinity-enterprise#397 | refusal `reason` reverse-engineered from its own message |
| F5b | abilityai/trinity-enterprise#396 | a whitespace-only credential clears the stored secret |

Also filed: abilityai/trinity-enterprise#394 — `mcp_validator._resolves_to_private_ip` carries **no** CGNAT clause at all, in either form. Deliberately out of scope here: not in this fix's blast radius, different threat model (config hygiene, fails closed on DNS failure, no credential on the wire), and vendored byte-identically under Invariant #5, so fixing it is 2 files plus a parity assertion.

## Notes for review

- No new endpoint, no dependency change, no schema change, no migration. The diff only **narrows** an existing predicate.
- Refusal strings stay fixed literals — the topology-oracle property (`test_refusal_message_never_echoes_the_resolved_address`) is unchanged.
- Per CLAUDE.md tiered documentation, a bug fix needs a descriptive commit message only; `architecture.md`'s `url_validation.py` entry remains accurate as written.


vybe:
Rebased onto `dev` (`ccc51ee`). Was `CONFLICTING`, now `MERGEABLE`.

**The conflict was an add/add** on `tests/unit/test_736_a2a_outbound_edges.py` — #2177 created its own file under that name after this branch was cut. Resolved by keeping this sweep as the base and folding #2177's delete suite in verbatim (own `oss_store` fixture; its `PEER_URL`/`PEER` renamed to `DELETE_PEER_URL`/`DELETE_PEER` off the collision), so neither PR loses regression coverage.

**Two strict-xfails had to go, and this is the part worth checking:** `strict=True` turns a fixed bug into a CI *failure*, so both of these would have reddened `dev` on merge —

- **F2 / ent#398** — fixed by #2180, merged an hour ago. Marker removed; `test_D7_…` now asserts the three sites agreeing on `:0` and passes plainly.
- **F4 / ent#395** — fixed on `dev` by #2177. This sweep's xfailed `test_G7_…` removed outright; #2177's passing version supersedes it. Its setup (`upsert_endpoint(alpha["id"], …)`) is now refused by #2177's create-time guard anyway, so it would have errored at setup rather than XPASSing.

F1 / F3 / F5b remain pinned — genuinely open. File header table records per-finding state.

**Verification**
- Affected A2A / SSRF / registry suites: **381 passed, 3 xfailed**
- Full `tests/unit`: **10,262 passed, 17 skipped, 4 xfailed, 3 failed**
- The 3 failures are `test_pat_propagation_properties::TestKnownGaps`, **verified failing identically on base `origin/dev`** — pre-existing and order-independent, per the triple this PR's own `learnings.md` entry prescribes.

`url_validation.py` auto-merged cleanly against #2180 (different functions), so the diff vs `dev` is the CGNAT one-liner only — no duplication of #2180's `effective_port`.

Commit authorship preserved; force-pushed with `--force-with-lease`.

vybe (approved):
Validated via /validate-pr. CI 23/23 green.

The predicate change is correct and minimal: `v4 = ip if ip.version == 4 else ip.ipv4_mapped` resolves the v4 view once, so the one hand-rolled clause in a borrowed stack redoes the normalisation the stdlib performs for the other six.

What makes this mergeable rather than merely plausible is that the completeness claim is *pinned* — `test_C1_ipv6_transition_forms_*` asserts the stdlib's own wholesale refusal of `::/8`, NAT64, 6to4 and Teredo, so an interpreter change can't reopen the gap with the suite still green. And `test_C1b` supplies the widening guard a narrowing security fix needs: refusing every mapped address would be strictly worse than the leak.

Note for the merger: cross-tracker ref, so ent#393 needs a manual status-in-dev bump.

PR #2180, opened 2026-08-14
fix(a2a): one port normalisation for validation, pinning and origin comparison (ent#398)

dolho:
## What this fixes

Port `:0` was normalised three different ways along one outbound A2A call path:

| Step | Treatment of `:0` |
|---|---|
| `_validate_public_https_url` | coalesced to 443 (`parsed.port or 443` — `0` is falsy) |
| `_pinned_url` | dropped; connects to 443 |
| `_same_origin` | compared **literally as port 0** |

So a `:0` endpoint validated, would have connected correctly, and was then permanently
refused `card_origin_mismatch` against any card declaring the ordinary form. Fail-closed —
nothing leaks or misroutes — so the damage is a permanently broken endpoint with a reason
code that points at the wrong thing. The hazard worth fixing is the three independent
normalisations: today they disagree harmlessly, and an edit to any one of them is what turns
that into something else.

## Changes

- **`utils.url_validation.effective_port(port, scheme)`** + `SCHEME_DEFAULT_PORTS` — one
  answer: `0` and absent both mean the scheme default (port 0 is not a connectable
  destination, and browsers refuse `:0` rather than dialing it), an explicit port is
  preserved, an unknown scheme yields `None` for the caller to interpret.
- **All three sites consume it.** `_same_origin` keeps `-1` for "unknown scheme, no default"
  so two such URLs stay comparable without ever matching a real port; the junk-port read
  stays inside its `except` clause, so a bad authority is still a named refusal
  (`endpoint_invalid`) rather than a traceback.
- **`_pinned_url` omits a port equal to the scheme default** instead of spelling it out —
  same destination, and consistent with the default-port equivalence the card comparison is
  built on. This is the one behaviour change beyond `:0`; the `Host` header still preserves
  what the operator typed, which is the only place an explicit `:443` is observable to a peer.

The AC offered "refuse `:0` at registration" as the alternative. Not taken: the pinned D7
test asserts the coalescing reading (`validated.port == 443` and the origin comparison
agreeing), `:0` is already accepted today so refusing it would break endpoints that
currently validate, and the fix wanted is the single normalisation either way.

Same input, same DNS stub, run against the live instance's own backend code:

```
dev baseline (worktree identical to dev on both files)
  validator port : 443
  pinned url     : https://93.184.216.34/a2a
  same_origin    : False   -> REFUSED card_origin_mismatch

this branch
  validator port : 443
  pinned url     : https://93.184.216.34/a2a
  same_origin    : True    -> would proceed
```

## Tests

`tests/unit/test_ent398_port_normalisation.py` — 21 cases: the normalisation truth table
(explicit / default / absent / `0` / unknown scheme / case), the three consumers agreeing on
one `:0` endpoint, the end-to-end card comparison against all three spellings of the card,
the `Host` header, **default-port equivalence both ways** (the #736/#738 property Trinity's
own card depends on), non-default ports surviving pinning incl. IPv6, path/query/fragment
preserved around the netloc change, junk ports still named refusals, unknown-scheme
comparison.

Green: 21 new, 658 across every A2A / URL-validation / registry suite.

## Coordination with PR #2178

The AC's fourth item — un-xfail `test_736_a2a_outbound_edges.py::test_D7_…` — cannot be done
here: that file is **not on `dev`**, it arrives with PR #2178 (still open), where D7 is pinned
strict-xfail. When #2178 lands, its D7 must be un-xfailed or it will XPASS-strict and redden
CI. This PR holds the same property on `dev` in the meantime, so there is no window where the
behaviour is unpinned.

Same applies to that PR's **G7** for the sibling finding: the delete fix landed on `dev` in
#2177, so G7 will XPASS too — and #2177 also added a create-time guard that refuses G7's own
setup (it creates the collision through `upsert_endpoint`). Flagged on #2178.

Fixes abilityai/trinity-enterprise#398

### DIFF
--- a/docs/memory/requirements/mcp.md
+++ b/docs/memory/requirements/mcp.md
@@ -504,6 +504,27 @@ remove are by id-or-name, so for a colliding string the update edits the record
 *named* that and the delete reaches the id-owning one first. Both are documented;
 neither is destructive.
 
+#### FR-4a — Port normalisation happens once (ent#398)
+The port of an endpoint authority is read by validation, by connection pinning and
+by the card's origin comparison. Each spelled its own normalisation, and they
+disagreed on `:0`: the validator coalesced it to the scheme default
+(`parsed.port or 443` — `0` is falsy), `_pinned_url` dropped it and therefore
+connected to 443, and `_same_origin` compared it literally as port 0. A `:0`
+endpoint validated, would have connected correctly, and was then permanently
+refused `card_origin_mismatch` against any card declaring the ordinary form —
+fail-closed, but permanently broken with a reason code pointing at the wrong thing.
+
+All three now consume `utils.url_validation.effective_port(port, scheme)`: `0` and
+absent both mean the scheme default (port 0 is not a connectable destination, and
+browsers refuse `:0` rather than dialing it), an explicit port is preserved, and an
+unknown scheme yields `None` for the caller to interpret. The default-port
+equivalence FR-2 relies on — Trinity's own card emits no explicit port, so
+`https://h` and `https://h:443` must be one origin — is preserved and pinned by
+test. `_pinned_url` now omits a port equal to the scheme default rather than
+spelling it out (same destination, consistent with that equivalence); the `Host`
+header still reflects what the operator typed, which is the one place an explicit
+`:443` is observable to a peer.
+
 #### FR-6 — `message/send` only; no SSE; no fake `stream` parameter
 Filed AC4 (stream token chunks back to the calling agent over SSE) is
 **rejected**: an MCP tool call is one request and one response — a FastMCP
--- a/src/backend/services/a2a_client.py
+++ b/src/backend/services/a2a_client.py
@@ -79,8 +79,10 @@ from utils.credential_sanitizer import (
 )
 from utils.url_validation import (
     A2AEndpointUrlError,
+    SCHEME_DEFAULT_PORTS,
     ValidatedPublicUrl,
     canonical_host as _canonical_host,
+    effective_port,
     validate_a2a_endpoint_url,
 )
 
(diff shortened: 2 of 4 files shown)


vybe (approved):
Validated: one shared `effective_port` normalisation replaces three divergent readings across validate → pin → compare, so the `:0` endpoint that validated, would have connected, and was then permanently refused `card_origin_mismatch` now agrees on all three sides.

Reviewed the code path directly — the junk-port read stays inside its `except ValueError` (bad authority is still a named `endpoint_invalid` refusal, not a traceback), `_same_origin` keeps `-1` for the unknown-scheme case so two such URLs stay comparable without matching a real port, and `_pinned_url` omitting a default-equal port is consistent with the default-port equivalence the card comparison is built on. Security scan clean; no new endpoint, config var, or top-level backend module, so no packaging surface.

Note for the follow-up: PR #2179 pins this exact bug as a `strict=True` xfail (finding F2, `test_D7_…`), so once this lands that marker must be removed or it XPASSes strict and reddens CI. #2179 is separately conflicting on that same file.

dolho:
Full backend unit suite on this branch: **10050 passed, 14 skipped, 1 xfailed, 3 failed** (37m53s).

The 3 failures are `test_pat_propagation_properties.py::TestKnownGaps::test_a_backslash_in_the_token_does_not_raise` — strict-`xfail` markers that now XPASS because the #1979 backslash-in-token bug was fixed and the markers were never removed. **Pre-existing on `dev`**, unrelated to this branch (verified earlier on a clean dev-based worktree with zero diff to `services/github_pat_propagation_service.py`; identical 3 failures). Same 3 reported on #2176.

Nothing in this branch touches that path — it is `utils/url_validation.py` + `services/a2a_client.py` only.

PR #2181, opened 2026-08-14
fix(a2a): compare IP-literal origins as addresses, not as text (ent#399)

dolho:
> **Stacked on #2180** (ent#398) — both change `_same_origin`. Base is
> `fix/ent398-port-normalisation`; GitHub retargets to `dev` when #2180 merges.
> The reviewable diff is the last commit.

## What this fixes

`_same_origin`'s docstring claimed *"IPv6 bracket forms compared after normalisation"*.
The code compared `urlsplit().hostname` **textually**: `canonical_host` leaves an IP literal
untouched — `idna.encode` rejects it and the ASCII fallback returns it verbatim — so no
normalisation ever happened. `[2606:4700:4700::1111]` and `[2606:4700:4700:0:0:0:0:1111]`,
one address written two ways, were refused `card_origin_mismatch`.

Fail-closed, hence P3. Two things make it worth fixing rather than redocumenting, both named
in the issue: an IPv6-literal endpoint is unusable whenever the peer's card and the
registration disagree on spelling (they have no reason to agree), and a docstring asserting a
property the code lacks is what the next reader builds on.

## Changes

- **`utils.url_validation.canonical_origin_host`** — the origin key. An IP literal is parsed
  through `ipaddress` and compared canonically; a name still goes through `canonical_host`
  (UTS-46 IDNA, the SV-7 clause); anything neither path canonicalises keeps the existing
  textual comparison, so an unusual host stays comparable to itself rather than becoming
  un-callable.
- **Scope id is part of the key** — `fe80::1%eth0` and `fe80::1%eth1` stay different
  destinations (AC: "bracket form and zone-id handling are covered").
- **Ambiguous IPv4 spellings are not folded** — `ipaddress` *refuses* leading-zero and integer
  forms rather than normalising them, so they fall through to the textual path. This is what
  keeps the change from ever equating two addresses a resolver would treat differently.
- **The docstring now states what the code does.**

## Verification against a real IPv6 peer

Lab peer on a globally-routable IPv6 (`2606:4700:4700::1111`), real TLS with an IP-SAN cert,
serving a genuine agent card that declares the **expanded** spelling:

```
                     dev                          this branch
compressed      REFUSED card_origin_mismatch      OK completed
expanded        OK completed                      OK completed
mixed groups    REFUSED card_origin_mismatch      OK completed
host uppercase  REFUSED card_origin_mismatch      OK completed
```

**The property the card check exists for is unweakened** — a peer that *listens* at `::1111`
while its card declares `::2222` is refused `card_origin_mismatch` on both dev and this branch.
`endpoint_invalid` for a malformed literal, and a card differing in port, scheme or path
(`card_url_ambiguous`), all still refuse.

## Tests

`tests/unit/test_ent399_ipv6_origin.py` — 29 cases:

- `test_E6_ipv6_literal_origins_compare_equal_across_spellings` — the AC-named property,
  plain. **The sibling pin is on the PR #2178 branch**, not on `dev` (that file arrives with
  #2178, where E6 is the strict-xfail); it must be un-xfailed when #2178 lands. This file
  holds the property on `dev` meanwhile, so there is no unpinned window.
- symmetric equivalence table (compressed/expanded, hex case, leading-zero groups, `::1`,
  literal + non-default port, and the ent#398 `:0` interaction)
- the must-stay-different table: different address, port, scheme, scope id, scoped vs
  unscoped, IPv4 vs its mapped IPv6 form, literal vs name
- ambiguous IPv4 spellings not folded; the IDN clause still holding; the textual fallback
- a cross-origin card still refused for a literal endpoint

Green: 29 new, 708 across every A2A / URL-validation / registry suite.

Related to trinity-enterprise#399


vybe (approved):
Validated via /validate-pr, plus merge-integration work.

**The fix is correct and I verified its claims empirically** rather than taking the description's word: `ipaddress` does preserve the scope id (`fe80::1%eth0` stays distinct), does canonicalise `2606:4700:4700:0:0:0:0:1111` → `::1111`, and does *refuse* leading-zero IPv4 spellings rather than folding them — which is the property that stops this from ever equating two addresses a resolver would treat differently. `canonical_origin_host` falls back to `canonical_host` for names, so the SV-7 IDN clause is untouched, and the textual fallback keeps an unusual host comparable to itself.

Security clean, no packaging gaps, 4 files.

**Three things I had to do to land it — all consequences of the stack, not of the fix:**

1. **Retargeted to `dev`** (base `fix/ent398-port-normalisation` merged as #2180) and merged `dev` in. Two conflicts, both adjacent-line churn: the `a2a_client.py` import block (dev never touched it — took `canonical_origin_host`, and confirmed `_canonical_host` has no remaining caller) and `requirements/mcp.md` (pure add-vs-nothing; FR-4b now sits after #2180's FR-4a, no duplication).

2. **Un-xfailed F1 in `test_736_a2a_outbound_edges.py`.** #2178 landed on dev while this was stacked, carrying F1 as a `strict=True` xfail — so this fix made it XPASS and CI would have gone red on merge. Exactly the obligation your description flagged. Marker removed, tracker row flipped to FIXED, prose updated, per the file's own convention. F3 and F5b stay xfail for #2182 and #2183.

3. **`lint (sys.modules pollution check)` was failing** — `test_ent399_ipv6_origin.py` had the sanctioned snapshot/restore pattern built correctly but named the list `_STUBBED`; the linter's AST check keys on the exact name `_STUBBED_MODULE_NAMES`. One-token rename. Worth noting this was invisible until retargeting, because that check only runs on a `dev` base — the standing hazard with stacked PRs.

Green locally: 906 passed across the a2a / url-validation / registry suites; lint back to 169/240 baseline with no new violations.

Note: `Related to trinity-enterprise#399` is cross-tracker and the status workflow's regex is same-repo bare-`#N` only, so ent#399 will not auto-promote — I'll bump it manually.

PR #2183, opened 2026-08-14
fix(a2a): a blank credential leaves the stored secret alone, in every spelling (#2175 F5b)

### LINKED ISSUE #2175: bug: four low-severity edge-case defects in the A2A outbound path (origin pin, port 0, reason codes, blank credential)
## Summary

Four low-severity defects found by an `/edge-cases` sweep of the #736 A2A outbound path. Grouped into one issue because they are all low, all **fail-closed** (each produces a refusal or a no-op, never a wrong-destination call or a leaked credential), and all live in the same feature — four separate P3s would be backlog noise.

Each is pinned by a strict-`xfail` in `tests/unit/test_736_a2a_outbound_edges.py`, so whichever ones get fixed will turn that marker red and force it to be retired.

---

### F1 — `_same_origin` does not normalise IPv6 literals, contrary to its docstring

`src/backend/services/a2a_client.py::_same_origin` documents "IPv6 bracket forms compared after normalisation". It does not: it compares `urlsplit().hostname` textually, and `canonical_host` leaves an IPv6 literal untouched (idna rejects it, the ASCII fallback passes it through).

So a registered IPv6-literal endpoint whose peer card declares the same address in expanded form is refused `card_origin_mismatch`:

```
registered: https://[2606:4700:4700::1111]/a2a
card url:   https://[2606:4700:4700:0:0:0:0:1111]/a2a   -> card_origin_mismatch
```

An IPv6-literal endpoint is registrable (it passes `validate_a2a_endpoint_url`), so this is reachable. Fail-closed, and the docstring is the part that is wrong.

Test: `test_E6_ipv6_literal_origins_compare_equal_across_spellings`

---

### F2 — port `:0` means three different things in one call path

- `_validate_public_https_url` coalesces it to 443 (`parsed.port or 443` — `0` is falsy), and resolves on 443
- `_pinned_url` drops it, so the connection also goes to 443
- `_same_origin` compares it **literally as port 0**

A `:0` endpoint therefore validates, would connect correctly, and is then permanently refused `card_origin_mismatch` against any card declaring the ordinary form — with an error that blames the peer for something the registration did. This is the §null/optional conflation: `0`, `None` and "absent" are three different values and two of the three helpers fold them together differently.

Test: `test_D7_a_zero_port_is_normalised_consistently_across_the_call_path`

---

### F3 — refusal reason codes are derived by substring-matching the human message

`validate_a2a_endpoint_url` maps `ValueError` → `A2AEndpointUrlError.reason` with `if "must use HTTPS" in message` / `"internal address" in message` / `"could not be resolved" in message`.

This is the exact fragility `A2AEndpointUrlError`'s own docstring warns against — *"string-matching a human message is how those two drift apart"* — happening in the function that raises it.

It is reachable because the DNS-failure message interpolates the hostname and `canonical_host`'s ASCII fallback passes hosts containing spaces:

```
https://an internal address.example/a2a   (DNS fails)
  -> reason=endpoint_private_address      (should be endpoint_dns_failure)
```

The router maps both to 400, so the HTTP status is unaffected and there is no bypass. What breaks is the machine-readable `reason` the MCP tool branches on, plus an operator being told the host resolves to an internal address when it does not resolve at all.

Test: `test_D12_a_dns_failure_is_never_misreported_as_a_private_address`

---

### F5b — a whitespace-only credential silently clears the stored one

`upsert_endpoint` skips the header-safety check on a blank credential (`if credential.strip() and ...`) but then takes the `elif credential:` branch, because `"   "` is truthy, and writes `""`.

So `None` and `""` preserve the stored secret; `"   "` destroys it — a fourth path through a model the docstring describes as three (omit / set / `clear_credential`). Not reachable over HTTP: `A2AOutboundEndpointUpsert._validate_credential` normalises a blank `SecretStr` to `None`, so the API surface is honest. This is a public-module-function defect.

Test: `test_G4_a_whitespace_only_credential_does_not_silently_clear_the_stored_one`

## Acceptance Criteria

- [ ] F1: `_same_origin` normalises IPv6 literals before comparing (or its docstring is corrected to state that it does not, and the limitation is recorded)
- [ ] F2: the three helpers agree on `:0` — either all treat it as the scheme default, or registration refuses it by name
- [ ] F3: `validate_a2a_endpoint_url` derives its `reason` from structured state rather than by substring-matching prose (e.g. the inner helper raises a reason-carrying error directly)
- [ ] F5b: `upsert_endpoint` treats a whitespace-only credential the same as `None`/`""` (leave unchanged); `clear_credential` remains the only removal path
- [ ] Each fixed defect has its strict-`xfail` marker retired in `tests/unit/test_736_a2a_outbound_edges.py` and the test passes plainly
- [ ] Partial fixes are fine — the markers are independent, so this issue can be closed incrementally

## Technical Notes

F1 and F2 are both in `_same_origin`, which is the **card same-origin pin** — the single control that stops a hostile agent card from redirecting a credentialed POST. Both current behaviours are conservative (refuse), so neither weakens that control; they make legitimate endpoints unusable rather than illegitimate ones reachable. Any fix must keep the refuse-on-unparseable property: `_same_origin` returns `False` for junk input on both sides, including junk-vs-identical-junk, and that is load-bearing.

F3's fix should be structural rather than a better set of substrings — the reason codes already exist as `A2A_URL_REASONS`; the inner `_validate_public_https_url` just doesn't carry them. Note the same sweep added a parity guard (`test_H1_the_status_map_has_no_dead_entries`) that fails if a reason is renamed at the raise site while the router's status map keeps the old spelling, so a refactor here is covered.

dolho:
Closes the last open item of #2175. **F1, F2 and F3 are already fixed** — each was filed on
the enterprise tracker and has its own PR:

| Finding | Issue | PR |
|---|---|---|
| F1 — IPv6 literal origins compared textually | ent#399 | #2181 |
| F2 — port `:0` normalised three ways | ent#398 | #2180 |
| F3 — reason derived by substring-matching prose | ent#397 | #2182 |
| **F5b — whitespace-only credential clears the secret** | — | **this PR** |

## What this fixes

`upsert_endpoint` documents three credential paths: omit (leave), set,
`clear_credential` (remove). A whitespace-only value was a fourth — it skipped the
header-safety check (`if credential.strip() and …`) and then took the `elif credential:`
branch, because `"   "` is truthy, writing `""`:

```
update with None     -> PRESERVED
update with ""       -> PRESERVED
update with "   "    -> DESTROYED
update with "\t"     -> DESTROYED
```

Not reachable over HTTP (`A2AOutboundEndpointUpsert._validate_credential` normalises a blank
`SecretStr` to `None`, so the API surface is honest) — which is what makes it a
public-module-function defect. The value at stake is a partner secret the caller may hold no
other copy of, and "blank clears it" is the wrong default for that.

## Change

One normalisation — `clean_credential = (credential or "").strip() or None` — applied after
the validity checks and before **either** write path, so a future third write site cannot
reintroduce the fourth path.

Deliberately unchanged:
- `clear_credential` remains the only removal path;
- the **length cap is still measured on the raw value**, so the bound an operator is told
  about does not move when their secret carries surrounding whitespace;
- the header-safety guard still refuses a credential with an *interior* space — only an
  entirely blank one means "leave it alone".

## Verification on real code

```
[dev]      "   " -> DESTROYED (now None)    "\t" -> DESTROYED
[branch]   "   " -> PRESERVED               "\t" -> PRESERVED
           None and "" preserved on both
```

## Tests

`tests/unit/test_2175_blank_credential.py` — 14 cases: every blank spelling, omission, the
create path storing no `credential` key at all (rather than an empty string
`has_credentials` then reports False for), a real credential still set / replaced / stripped,
`clear_credential` precedence, the header-safety guard, the raw length cap, and that no
empty-string credential can reach the caller as an empty `Authorization: Bearer ` header.

Green: 14 new, 264 across the A2A suites.

## Note on the xfail marker

The AC asks for `test_G4_a_whitespace_only_credential_does_not_silently_clear_the_stored_one`
to be retired. That test is in `tests/unit/test_736_a2a_outbound_edges.py`, which is **not on
`dev`** — it arrives with PR #2178, where G4 is the strict-xfail. It must be un-xfailed when
#2178 lands (alongside G7, D7, E6 and D12, all now fixed); this file holds the property on
`dev` meanwhile. Flagged on #2178.

Closes #2175

> Closing keyword added at merge time: F1 (#2181), F2 (#2180) and F3 (#2182) have all
> landed, so this PR is the last of the four findings and #2175 is fully resolved.
> A bare `Related to` would have left it stranded in `status-in-progress` —
> `issue-status-on-merge.yml` promotes only on `Fixes`/`Closes`/`Resolves` + `#N`.

### DIFF
--- a/docs/memory/requirements/mcp.md
+++ b/docs/memory/requirements/mcp.md
@@ -546,6 +546,24 @@ are *refused* by `ipaddress` rather than folded, so the normalisation can never
 equate two addresses a resolver would treat differently. The card stays a hint: a
 card declaring a different address, port or scheme is still refused.
 
+#### FR-5b — A blank credential means "leave it alone", in every spelling (#2175 F5b)
+`upsert_endpoint` documents three credential paths — omit (leave), set,
+`clear_credential` (remove). A whitespace-only value was a fourth: it skipped the
+header-safety check (`if credential.strip() and …`) and then took the
+`elif credential:` branch, because `"   "` is truthy, writing `""`. So `None` and
+`""` preserved the stored secret while `"   "` destroyed it. Not reachable over
+HTTP (`A2AOutboundEndpointUpsert._validate_credential` normalises a blank
+`SecretStr` to `None`, so the API surface is honest) — a public-module-function
+defect, over a partner secret the caller may hold no other copy of.
+
+Every blank spelling now collapses to `None` in ONE normalisation applied after
+the validity checks and before either write path, so a future third write site
+cannot reintroduce the fourth path. `clear_credential` stays the only removal
+path. The length cap is still measured on the raw value, so the bound an operator
+is told about does not move when their secret carries surrounding whitespace, and
+the header-safety guard still refuses a credential with an *interior* space —
+only an entirely blank one is "leave it alone".
+
 #### FR-6 — `message/send` only; no SSE; no fake `stream` parameter
 Filed AC4 (stream token chunks back to the calling agent over SSE) is
 **rejected**: an MCP tool call is one request and one response — a FastMCP
--- a/src/backend/services/a2a_outbound.py
+++ b/src/backend/services/a2a_outbound.py
@@ -349,9 +349,18 @@ def upsert_endpoint(
     regardless: a stored row is not trusted, a DNS record can move, and the
     settings row could be written by some future path that skips this function.
 
-    A credential is **write-only**: omitted leaves an existing one untouched
-    (so an operator can rename or repoint without re-typing a secret they may
-    not have), and `clear_credential=True` removes it.
+    A credential is **write-only**, and there are exactly THREE paths (#2175 F5b):
+    omitting it leaves an existing one untouched (so an operator can rename or
+    repoint without re-typing a secret they may not have), passing one sets it,
+    and `clear_credential=True` is the only way to remove it. **Every blank
+    spelling — `None`, `""`, `"   "` — means "leave it alone."** A whitespace-only
+    value used to be a fourth path that silently destroyed the stored secret: it
+    skipped the header-safety check (`if credential.strip() and …`) and then took
+    the `elif credential:` branch, because `"   "` is truthy, writing `""`. Not
+    reachable over HTTP — `A2AOutboundEndpointUpsert._validate_credential`
+    normalises a blank `SecretStr` to `None` — but this is a public module
+    function, and "blank clears the secret" is the wrong default for a value the
+    caller may hold no other copy of.
     """
     import uuid
 
@@ -384,6 +393,13 @@ def upsert_endpoint(
                 "Endpoint credential contains characters that are not valid in an "
                 "HTTP header (whitespace, line breaks or control characters)."
             )
+    # #2175 F5b: ONE normalisation, immediately after the checks above — every
+    # blank spelling collapses to None ("leave it alone") before either write
+    # path can see it. Done here rather than at each branch so a future third
+    # write site cannot reintroduce the fourth path. The length cap is still
+    # measured on the raw value, so the bound an operator is told about does not
+    # move when their secret has surrounding whitespace.
+    clean_credential = (credential or "").strip() or None
     try:
         validate_a2a_endpoint_url(clean_url)
     except A2AEndpointUrlError as exc:
@@ -397,8 +413,8 @@ def upsert_endpoint(
             record["url"] = clean_url
             if clear_credential:
                 record.pop("credential", None)
-            elif credential:
-                record["credential"] = credential.strip()
+            elif clean_credential:
+                record["credential"] = clean_credential
             _store_endpoint_records(records)
             return _public_record(record)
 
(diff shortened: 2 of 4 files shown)


vybe (approved):
Rebased onto current dev (`requirements/mcp.md` conflict was purely additive — dev gained FR-4a/FR-4b from the merged #2180/#2181, this branch adds FR-5b; all three kept in numeric order).

**Fixed a defect that would have merged red.** `test_G4_a_whitespace_only_credential_does_not_silently_clear_the_stored_one` carried `@pytest.mark.xfail(strict=True)` documenting exactly this bug. Fixing the bug makes it XPASS, and a strict XPASS is a hard failure. The PR body reports "Green: 264 across the A2A suites", but the subset run excluded `test_736_a2a_outbound_edges.py` — and this PR had **no CI checks at all** ("no checks reported"), so nothing caught it. The assertion was already written as the correct behaviour, so dropping the marker converts it in place into a live regression test. Both head-seed pytest jobs are now green, which are the ones that would have failed.

Also validated: security scan §4.1–4.10 clean, no new env vars or packages, `clear_credential` remains the only removal path, length cap still measured on the raw value.

Two bookkeeping items that do **not** block this merge:
- `Closes #2175` will auto-close an issue covering four findings, and **F3 (#2182) is still open** — its CodeQL failure is a genuine new high-severity alert on its own diff, not a stale-run artifact, so it is not one re-run away. I'll reopen #2175 after merge.
- **ent#396** tracks F5b on the private tracker; cross-repo refs neither auto-close nor auto-promote, so it needs a manual bump.

PR #2184, opened 2026-08-14
feat(grid): Executions info tile — 24h hourly, stacked by trigger (ent#96)

dolho:
The Grid had no fleet-level execution chart — per-agent tiles carry 14d activity, but "what has
the system been doing today" had no answer at fleet scope. This adds the **Executions** tile:
24 hourly columns stacked by trigger bucket, a failure rail, headline totals, and live
running/queued chips. Default-on, toggled in Tiles ▾ like any other tile.

Unblocked by the foundation landing: #325's chassis and #326's timeline endpoint are both on
`dev` (the issue's own comment said "promote once #326 lands").

## Backend — `split=trigger`, extending ent#326

The stack needs **hour × trigger** and `/api/executions/timeline` grouped one dimension at a
time, so the alternative was one request per bucket name — and a column whose segments came
from ten responses can disagree with its own total.

`split=trigger` folds the second dimension server-side and **re-sums each bucket's totals from
the split rows**, so that disagreement is unrepresentable. Each bucket gains
`by_trigger: {label: {total, failed}}`; the response carries `trigger_order` (the db layer's own
`_BUCKET_ORDER`) so the tile cannot hold a stale copy of the stack order — AC1's "tile, legend
and Overview never disagree". Gap-filled intervals carry `{}` rather than a missing key, so the
renderer never distinguishes "no runs" from "no field". `split` is a **named 422** over a
categorical `group_by` and for an unknown value. The unsplit response is unchanged.

## Frontend

- `ExecutionsTile.vue` + `utils/executionsTile.js` — all shaping pure, since vitest runs
  node-environment here and a pure module is the only part of a tile that can be unit-tested.
- **Failures are a rail, not a stack segment** (AC2). As a segment, "Failed" would have to be
  subtracted from its trigger's segment to keep the column equal to runs — which silently
  redefines every other segment as *"runs that succeeded"*. The rail keeps the column honest
  and the failures visible; per-label `failed` still rides in the payload, so the hover
  breakdown names which trigger failed.
- A bucket present in the data but missing from `trigger_order` is **appended, not dropped** —
  otherwise it counts toward a column total while missing from its stack.
- Honest states: terminal-based success rate reports `—` not 0% when nothing has terminated;
  "no executions in 24h" requires a successful read (ent#100's manufactured-green rule); a
  failed background refresh keeps the last good chart with a `24h · stale` stamp; the live chips
  degrade to *absent*, never to zero, since a failed `/stats` is not evidence that nothing runs.
- Both GETs ride the existing 60s batch poll, gated on the tile being enabled. No new timer, and
  no fetch on mount — viewport culling unmounts tiles, so that would re-issue on every pan.
- Seven new `--gv-bk-*` bucket colours in **both** theme blocks. `AgentTile` collapses ten
  buckets to three because a 60px sparkline cannot carry ten; a fleet tile stacks all ten.

## Verification on the live instance

```
tile headline (computed from the buckets) : runs=747  ok=98%  failed=17
GET /api/executions/stats?hours=24        : total=747          failed_count=17
```
The chart and the stat card agree — the failure mode ent#326 was written to avoid. Also
verified live: 25 gap-filled buckets over 24h, every bucket's `by_trigger` summing to its own
`total`, quiet hours carrying `{}`, the unsplit response byte-identical to before, and all
three 422s (unknown split, `group_by=trigger`, `group_by=agent`).

## Tests

- `tests/unit/test_ent96_timeline_split.py` — 14: the fold sums to its column, per-label
  failures survive, an unmapped trigger lands in `Other` rather than vanishing, two raw triggers
  sharing a bucket merge rather than overwrite, gap-filled intervals carry `{}`, the unsplit
  contract is untouched, and every meaningless `split` is a named 422.
- `src/frontend/tests/unit/executionsTile.spec.js` — 20: segments sum to their column, failures
  stay out of the stack, an unknown bucket is appended and coloured with the catch-all, bars
  scale against the tallest column, a real execution never rounds to 0px, zero hours get the
  baseline stub, and the four tile states incl. stale-while-revalidate.
- Frontend suite **576 green** (incl. `gridTokens` + `gridWidgets` guards), build clean.
  Backend `ent#96 + ent#326` suites 48 green.

## Two things worth a reviewer's eye

1. **I touched ent#326's test file.** Two stubs mirror the db signature and gained the new
   optional parameter. Related: the route normalizes `split` with `isinstance` rather than
   `is not None` — called directly (as those tests do) an omitted parameter arrives as the
   `Query(...)` default *object*, which is truthy, so a bare None-check turned every such call
   into a 422. That is a trap for any future optional query param on a directly-called route.
2. **The token guard earned its keep**: it caught `--gv-fg` — read by my tile, defined in
   neither theme — before it shipped as a permanently-live fallback.

Fixes abilityai/trinity-enterprise#96


vybe (approved):
Validated via /validate-pr. CI 25/25 green (incl. e2e, pg-migrations, schema-parity).

Two design calls are right, not just present: failures as a **rail** rather than a stack segment (as a segment, every other segment silently becomes "runs that succeeded"), and an unknown bucket **appended** rather than dropped (otherwise it counts toward a column total while missing from that column's stack). `trigger_order` served from the db layer's own `_BUCKET_ORDER` is what keeps tile, legend and Overview from ordering the same buckets differently.

Server-side re-summing of each bucket's totals from its split rows makes stack-vs-total disagreement unrepresentable — the exact failure a second query would have invited.

The `isinstance(split, str)` normalisation reads odd but is load-bearing and documented at the site: a directly-called route receives the `Query(...)` default *object*, which is truthy, so a bare `is not None` would 422 every direct call.

Body patched pre-merge from `Related to` to `Fixes abilityai/trinity-enterprise#96`. Cross-tracker, so ent#96 still needs a manual status-in-dev bump.

PR #2187, opened 2026-08-14
fix(agents): starting a stopped agent no longer 500s when a recreate is needed (#2186)

### LINKED ISSUE #2186: bug: starting a stopped agent 500s when a recreate is needed — require_running guard not satisfied by start_agent_internal
## Summary

`POST /api/agents/{name}/start` returns **HTTP 500** for a **stopped** agent whenever any recreate predicate fires. The `require_running=True` default added to `recreate_container_with_updated_config` by #2092 (PR #2155) is not satisfied by its own most important caller — `start_agent_internal`, which recreates stopped containers by design.

Unreleased: the regression is on `dev` only (`v0.8.5` is unaffected), so it is a **release blocker for the current freeze**, not a live production outage.

## Context

#2092 hardened the helper so it can no longer silently start a container the caller believed was stopped:

```python
# src/backend/services/agent_service/lifecycle.py:768
async def recreate_container_with_updated_config(
    agent_name, old_container, owner_username, *,
    full_capabilities=None, env_overrides=None,
    require_running: bool = True,      # <-- new
    preserve_run_state: bool = False,  # <-- new
):
```

The guard reads the precondition before doing any work and raises:

```python
# lifecycle.py:840
was_running = getattr(old_container, "status", None) == "running"
if require_running and not was_running:
    raise ValueError(...)
```

The PR states every current caller already enforces this for itself. That holds for two of the three, but **not** for `start_agent_internal`:

- `services/agent_mcp_key_service.py` — gates on `status == "running"` before recreating (stopped agents take a DB-only path). ✅
- `services/agent_service/repo_binding.py` — worth re-confirming, but not the failing path here.
- **`services/agent_service/lifecycle.py:414` — recreates stopped containers deliberately.** ❌

No caller passes either new keyword (`grep -rn 'require_running\|preserve_run_state' src/` returns only the definition), so all three take the `True` default.

### Why the lifecycle path is guaranteed to hit it

`start_agent_internal` records the run state, then reaches the recreate with the container still stopped:

```python
lifecycle.py:299   was_already_running = getattr(container, "status", None) == "running"
lifecycle.py:331   if not needs_recreation and not was_already_running:   # base-image drift (#1809)
                       ...  needs_recreation = True
lifecycle.py:414   await recreate_container_with_updated_config(agent_name, container, "system", ...)
lifecycle.py:419   await container_start(container)                        # it starts it right after
```

Line 331 is the decisive one: the #1809 base-image-drift adoption branch is gated on `not was_already_running`, so it fires **only** when the container is stopped — the exact state the new guard refuses. That makes cold-start base-image adoption **100% unreachable**, not merely occasionally broken.

Config drift reaches the same line by the other route: the predicates above (shared-folder mounts, API key/subscription, resource limits, capabilities, guardrails, per-agent PAT, agent MCP key) are evaluated regardless of run state, so drift accumulated while an agent was stopped also lands on line 414 with a stopped container.

### What the user sees

`routers/agents.py` catches the `ValueError` in its broad handler and reports it as an opaque 500:

```
500  Failed to start agent (ValueError — details in backend logs)
```

Same failure via the UI Start toggle and the MCP `start_agent` tool.

## Impact

- **After any base-image rebuild** (`./scripts/deploy/build-base-image.sh`, a routine documented operation), every stopped agent fails to start until it is started by some other path.
- Any agent whose config drifted while stopped cannot be started.
- The failure is a generic 500 with no actionable message, so it will be reported as "agent won't start" with no pointer to the cause.

## Why CI didn't catch it

- `tests/unit/test_2092_recreate_run_state.py` exercises the guard directly with synthetic containers; its own docstring notes the container-creation body is not exercised (it needs Docker). It never drives `start_agent_internal`.
- `tests/unit/test_1809_image_drift_recreate.py` is a source/AST test asserting the gate *exists*; it does not assert the runtime outcome.

So both halves are individually green while the composition is broken.

## Acceptance Criteria

- [ ] `POST /api/agents/{name}/start` succeeds for a **stopped** agent when a recreate predicate fires (config drift **and** base-image drift), and the agent ends up **running**
- [ ] `start_agent_internal` passes the run-state intent explicitly at `lifecycle.py:414` rather than relying on the default
- [ ] The #1809 cold-start base-image adoption path is reachable again — a stopped agent on a stale base image adopts the rebuilt image on its next start
- [ ] A regression test drives `start_agent_internal` (not the helper in isolation) with a **stopped** container and a firing recreate predicate, and asserts the agent starts
- [ ] `agent_mcp_key_service` and `repo_binding` are re-confirmed to satisfy the precondition, or pass it explicitly too
- [ ] `ValueError` from this guard no longer surfaces as an opaque 500 — either it cannot escape, or the start endpoint maps it to a message naming the cause

## Technical Notes

The minimal fix is to make the intent explicit at the call site, since `start_agent_internal` legitimately intends to start the agent (it calls `container_start` on the very next line):

```python
# lifecycle.py:414
await recreate_container_with_updated_config(
    agent_name, container, "system",
    env_overrides=mcp_env_overrides,
    require_running=False,      # this path recreates stopped containers by design
)
```

`preserve_run_state` should stay `False` here — the caller starts the container immediately afterwards, so leaving the replacement stopped would be wrong.

Worth considering as a follow-up: the guard's value is real, but a precondition that three callers must each remember is the same shape as the bug it was added to prevent. Deriving the flag from the observed run state (with an explicit opt-out for callers that must refuse) would remove the class rather than this instance.

Related: #2092 (the hardening), #1809 (base-image drift adoption).

dolho:
**P0 release blocker for the current freeze.** Regression is on `dev` only; `v0.8.5` is unaffected.

## What broke

#2092 gave `recreate_container_with_updated_config` a `require_running=True` default so it can
no longer silently start an agent the caller believed was stopped — a real guard, and its PR
noted that every caller already enforced the precondition. That held for the two running-only
callers and **not** for `start_agent_internal`, which recreates stopped containers by design and
calls `container_start` on the very next line.

The #1809 branch makes it worse than intermittent: base-image adoption is gated on
`not was_already_running`, so it fires **only** for a stopped container — precisely the state the
guard refuses. Cold-start image adoption was **100% unreachable**, so after any
`build-base-image.sh` every stopped agent failed to start.

## Reproduced live, then fixed live

```
# dev code, stopped agent, memory limit changed 4g -> 2g while stopped
POST /api/agents/ws-scout/start
  -> 500 {"detail":"Failed to start agent (ValueError — details in backend logs)"}
     container: still exited

# this branch, same agent, same drift
POST /api/agents/ws-scout/start
  -> 200 {"recreated": true, "recreate_reason": "config_drift"}
     container: running, mem=2147483648   (the drift actually applied)

# this branch, stopped agent on a rebuilt base image
POST /api/agents/ws-scout/start
  -> 200 {"recreated": true, "recreate_reason": "image_drift"}
     container: running, image = the new base image id
```

(The base-image tag was restored and the agent returned to its original limits and stopped state
afterwards.)

## The fix

- **`lifecycle.py`** start path passes `require_running=False` — the intent stated where the
  decision is made. `preserve_run_state` stays `False`: this is the one caller whose purpose is
  to start the container.
- **`repo_binding.py`** and **`agent_mcp_key_service.py`** re-confirmed running-only (both gate
  on `status == "running"` before recreating, as the issue asked) and now pass
  `require_running=True` **explicitly** rather than inheriting it.
- **A static guard asserts every call site states its run-state intent.** The guard's own failure
  mode was a precondition three callers each had to remember — this bug is an instance of that
  class, and the AST test closes the class cheaply. It is the lighter half of the issue's
  "derive the flag from observed run state" follow-up: the *decision* stays explicit, but it can
  no longer be made by accident.

**On the last AC (`ValueError` no longer an opaque 500):** taken via the "it cannot escape"
branch. The endpoint's generic message is deliberately left alone — it is the #1917
stack-trace-exposure shape, and widening it to carry an internal precondition string would trade
one problem for another.

## Tests

`tests/unit/test_2186_start_stopped_agent_recreate.py` — 8 cases driving **`start_agent_internal`
itself**, which is the gap that let this ship: `test_2092_recreate_run_state.py` drives the guard
with synthetic containers and never calls the start path, and `test_1809_image_drift_recreate.py`
is an AST test asserting the gate *exists*. Both were individually green while the composition
was broken.

Covered: a stopped agent with config drift starts; a stopped agent on a stale base image starts
and adopts it; the flags are passed explicitly; the replacement is not left stopped; **a running
agent is unaffected** (image drift still does not recreate it — #2092's property); config drift
still recreates a running agent; plus the two static guards.

One harness note for reviewers: the test stubs `wait_for_agent_ready`, which really polls the
agent's HTTP port — leaving it live makes the test hang rather than fail.

Closes #2186


dolho:
**Test status — settled, and it turned up more evidence for the fix.**

I first pushed a version that also passed `require_running=True` explicitly at the two already-correct call sites. That broke 12 tests: three in-tree doubles stub `recreate_container_with_updated_config` with a fixed 3-arg signature, so an added keyword raises `TypeError` inside the code under test. Found by diffing against clean `dev` rather than trusting the "looks environmental" read of the log (it wasn't — dev fails 2, that branch failed 14). Narrowed in the second commit; the AC allows either option and re-confirmation is the one that doesn't churn test doubles during a P0.

Final numbers on the five suites around this code (`start_agent_skip_inject`, `ent109_repo_binding`, `2092`, `1809`, `2186`):

| | result |
|---|---|
| clean `dev` | **2 failed**, 44 passed |
| this branch | **0 failed**, 86 passed |

The two failures on `dev` are `test_start_agent_skip_inject.py::test_injects_when_container_was_stopped` and `::test_injects_when_container_recreated_even_if_was_running` — both assert that post-start injection happens when the container was stopped or recreated. They fail on `dev` for exactly the reason this issue reports: the recreate raises, the start never completes, and the injection never runs. **This fix repairs them without touching them**, which is independent confirmation that the broken path is the one #2186 describes.

Live re-verification after narrowing: stopped agent + config drift → `{recreated: true, recreate_reason: "config_drift"}`, container running with the new limit. Test agent and base-image tag restored to their original state.

dolho:
**Full live matrix against the real instance** — same seven states, run twice: `dev` code, then this branch. Both runs on the running Trinity, driving `POST /api/agents/{name}/start` over HTTP.

| # | state | dev | this branch |
|---|---|---|---|
| A | stopped, no drift | 500 `ValueError`* | ok, `recreated=false` |
| B | stopped + config drift | **500 `ValueError`** | ok, `recreated=true reason=config_drift`, limit applied |
| C | running, start again | 500 `ValueError`* | ok, `recreated=false`, same container id |
| D | running + config drift | 500 `ValueError`* | ok, `recreated=true reason=config_drift`, new container, limit applied |
| E | running + image drift | 500 `ValueError`* | ok, `recreated=false`, **same container id** (#2092's property intact) |
| F | stopped + image drift | **500 `ValueError`** | ok, `recreated=true reason=image_drift`, new image id adopted |
| G | back to no drift | ok | ok |

\* On `dev` the rows marked with an asterisk are a **cascade**, not seven independent failures: once row B's start fails the drift is never applied, so every later row inherits it and the agent stays stopped. That is itself the operator-facing shape of this bug — an agent with drift can never start, and nothing clears the drift but a successful start.

Also verified on the branch:
- the agent is **functional**, not merely `running`: its own `/health` answers `{"status":"healthy","runtime":"claude-code","runtime_available":true}` after a recreate-on-start;
- the **workspace volume survives** the recreate (`agent-ws-scout-workspace` still mounted);
- **zero** `ValueError` / `would START agent` lines in the backend log across the whole matrix;
- reproduced independently on a **second agent** (`ws-sage`): stopped + drift → `recreated=true reason=config_drift`.

One correction to my own harness, since it affects how row D should be read: my first pass set memory to `3g`, which the API rejects (`Must be one of: 1g, 2g, 4g, 8g…`), so that row had no drift and its `recreated=false` meant nothing. Re-run with `2g`: the container id changes and the limit is applied, which is the result reported above.

Everything restored afterwards — both agents back to 4g and stopped on the original base image, base-image tag back to its original id, no stray lab images.

dolho:
Full backend unit suite on this branch: **10058 passed, 14 skipped, 1 xfailed, 3 failed** (37m27s).

The 3 are `test_pat_propagation_properties.py::TestKnownGaps::test_a_backslash_in_the_token_does_not_raise` — stale strict-`xfail` markers that now XPASS because #1979's backslash bug was fixed and the markers were never removed. Pre-existing on `dev`, unrelated to this branch (verified earlier on a clean dev worktree), and the same 3 reported on #2176 and #2180.

That closes the evidence set for this P0: one-line fix + narrowing commit, 8 targeted unit tests, the five-suite comparison (dev 2 failed / branch 0 failed of 86 — this fix repairs two pre-existing dev failures without touching them), the seven-state live matrix on a real instance, and now the full suite.

vybe (approved):
Validated via /validate-pr.

**Verified live-reproduced fix.** The #2092 `require_running=True` default collided with `start_agent_internal`, whose whole purpose is recreating a stopped container — and the #1809 base-image branch is gated on `not was_already_running`, so cold-start image adoption was 100% unreachable. Confirmed in the diff: only the start path flips to `require_running=False`; `preserve_run_state` correctly stays `False`.

Security clean, no packaging gaps, minimal scope. `test_2186_start_stopped_agent_recreate.py` drives `start_agent_internal` itself — the composition gap that let two individually-green tests ship a broken path — and the AST guard pins the audited call-site set so a new one fails loudly.

One non-blocking note: the PR body says repo_binding.py and agent_mcp_key_service.py "now pass `require_running=True` explicitly", but the second commit reverted that — they carry explanatory comments and inherit the default instead. The code is right; the body is stale.

PR #2054, opened 2026-08-06
docs: document /trinity:start-here, the guided first journey

vybe:
## What

Documents the trinity plugin's new seventh skill, `/trinity:start-here` — a guided, resumable journey that takes a newcomer from "what is Trinity?" to a first agent running on their own instance. Shipped in the marketplace as [abilities `629aabb`](https://github.com/Abilityai/abilities/commit/629aabb) (trinity plugin 2.6.1).

**Docs-only.** No code, no behavior change.

## Why now

`docs/user-docs/abilities/trinity-plugin.md` says "Six skills" and lists six; `overview.md` gives newcomers a bare install list with no suggested entry point. Both are now stale, and both are upstream of the two surfaces people actually read:

- the docs site, regenerated by the `trinity-docs` agent's `/sync-docs` from `docs/user-docs/`
- the **Trinity Docs Q&A** Vertex index, re-synced by `.github/workflows/sync-docs-to-vertex.yml` on every push to `main`

So until this lands, neither the site nor the docs assistant can tell anyone the skill exists.

## Changes

| File | Change |
|---|---|
| `abilities/trinity-plugin.md` | Skills-table row; new **Guided Journey** section (stage table, zero-commitment tour, resume/`reset`, live-docs grounding); intro banner linking to it; "Six" → "Seven" |
| `abilities/overview.md` | Quick Start leads with the guided path before the manual install list; plugin table 6 → 7 skills |

## Notes for review

- The skill routes to existing skills (`connect`, `deploy-new-instance`, `onboard`, the create-agent wizards) rather than reimplementing them — the docs describe it that way deliberately.
- It answers platform questions from live docs at run time (public Docs Q&A endpoint pre-connect, the instance's `ask_trinity` after), so this page is describing behavior that stays current on its own.
- Cross-link target `../getting-started/help.md` verified present.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1829, opened 2026-07-27
fix(helper-mcp): rename to unscoped trinity-docs-mcp (npm scope unavailable)

vybe:
The `@abilityai` npm scope is held by a legacy account we can't access (website login/recovery dead-ends — support recovery is being pursued out-of-band), so the v0.8.5 `Publish Helper MCP to npm` workflow could never bootstrap. Resolution: publish **unscoped** as `trinity-docs-mcp`.

- `trinity-docs-mcp@0.1.0` is **already live** on npm (manual first publish done today; Trusted Publisher being pointed at `publish-helper-mcp.yml` so future publishes are OIDC/tokenless)
- Renames the package in `package.json` + lockfile, README install snippets, the publish workflow (incl. its version-exists guard + bootstrap comment), and the three docs pages that referenced the scoped name
- The workflow's red X on main self-resolves once this reaches main at the next cut — the version-exists guard makes re-runs idempotent

Closes Abilityai/trinity-enterprise#330

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1777, opened 2026-07-26
chore: reconcile main into dev before the v0.8.5 cut

vybe:
## What

Merge-commit reconcile of `main` → `dev` (13 commits that landed directly on main): hono 4.12.31 (#1738), fast-uri 3.1.4 (#1736), body-parser 2.3.0 (#1733), llms.txt, the 264-QA FAQ + PAT-scope docs, open-research folder, explainer-video links, and main's earlier enterprise-pointer alignment.

## Why

Required by the v0.8.5 freeze-plan recheck (2026-07-26): main is ahead of dev, so cutting dev→main without this risks a conflicted release merge.

## Conflict resolutions

- `src/backend/enterprise` gitlink → **e898cef8** (enterprise-main TIP, same target as #1776 — the ent#220 rooms security fix; the two PRs merge cleanly in either order)
- `.claude` gitlink → dev's `de7fc71` (main's pointer is a strict ancestor — verified)
- `docs/user-docs/integrations/github-pat-setup.md` → **dev side**: main's older text pre-dates #1574 (gh CLI now ships in the base image and `GH_TOKEN`/`GITHUB_TOKEN` are wired); dev's version describes the current code
- `docs/user-docs/faq/channels-and-messaging.md` → dev side (dev-only added Q&A)

⚠️ **Merge with a MERGE COMMIT, not squash** — squashing a reconcile re-diverges main (the recorded stacked-squash revert trap).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2188, opened 2026-08-14
chore: reconcile main into dev (merge, not rebase)

trinity-ability:
## Why

`dev` and `main` diverged: **dev 425 ahead / 26 behind**. This merges main back into dev so main is an ancestor again and the next release PR is clean.

**This is a merge, not a rebase — deliberately.** A rebase of dev would rewrite 425 SHAs and needs a force-push, but `dev` has `allow_force_pushes: false` + `enforce_admins: true`, and **12 open PRs** are based on it. It is blocked *and* destructive.

## What main actually had that dev didn't

All **9 code+docs fixes** on main were already in dev — they are the promote-to-main cherry-picks of fixes dev already carries. (`git cherry` flags #1759 as missing; that's a false positive: dev has it as `e54ce3a9`, main as `86f8493e`.)

The genuine gap was small:

| Gap | Detail |
|---|---|
| Root lockfile | `hono` 4.12.25→4.13.0 · `@hono/node-server` 1.19.14→1.19.17 · `ip-address` 10.2.0→10.4.0 · `fast-uri` 3.1.4→3.1.5 |
| Docs #2054 | `/trinity:start-here` in `docs/user-docs/abilities/{overview,trinity-plugin}.md` |

Everywhere else **dev is ahead** — its own dependabot kept running.

## Net effect: 8 files

```
docs/user-docs/abilities/overview.md       | 12 ++-
docs/user-docs/abilities/trinity-plugin.md | 31 +++-
package-lock.json                          | 42 ++--
package.json                               |  2 +-
src/frontend/package-lock.json             |  2 +-
src/frontend/package.json                  |  2 +-
src/helper-mcp/package.json                |  2 +-
src/mcp-server/package.json                |  2 +-
```

## Conflict resolution (24 files)

dev's side everywhere except where main was genuinely newer, because main carries the **older cherry-picked form** of fixes dev has since extended.

- **`src/backend/services/agent_service/crud.py`** — dev is a strict superset: **THREE** seams sharing `_resolve_local_template_dir` (#1900/#1935) vs main's **TWO**, and `HardenedYamlError` in the `except` clause (ent#314). Both sides carry the same `UNKNOWN_LOCAL_TEMPLATE` / `LOCAL_TEMPLATE_INVALID` / `INVALID_LOCAL_TEMPLATE_NAME` gates.
- **`src/helper-mcp/package.json`** — kept dev's **unscoped** `trinity-docs-mcp`. #1829 renamed it on purpose ("npm scope unavailable"); main still has the scoped form and taking it would break `publish-cli`. Took main's `@modelcontextprotocol/sdk` `^1.30.0` alongside it.
- **Nested package/lock files** — dev's side. Main's would have downgraded `pinia` 4.0.2→3.0.4, `fastmcp` 4.13.1→4.8.0, plus vite/vue/axios/playwright, and dropped `vitest`. Main's two floor bumps that dev lacked were **re-applied**: frontend `mermaid ^11.16.1` (with the lock root mirror kept in sync) and mcp-server `hono >=4.13.0` override — both already satisfied by dev's resolved lock, so `npm ci` stays valid.
- **`tests/registry.json`** — rebuilt from the three merge stages rather than splicing markers. Main contributed **0** entries dev lacks, and both dual-edited entries were unchanged from base on main. Result: 176 entries, no duplicates.
- **Frontend components + e2e specs** — dev's side. Both sides keep all 4 `data-sep` hooks, and dev's spec additionally covers the ent#260 `list` mode.

## Verification

- Zero conflict markers tree-wide (unfiltered `git grep`)
- Every touched JSON parses; lockfile ranges vs resolved versions checked
- `check_alembic_heads.py` → 39 revisions, **1 head** (`0038_portal_chat_state`)

## ⚠️ Merge with a **merge commit**, not squash

Squashing collapses the merge and main stays a non-ancestor — the same 24 conflicts return at the next release cut.

<sub>Branch name reused from the merged reconcile #1777; the push was a fast-forward, that PR is unaffected.</sub>


vybe (approved):
Validated end-to-end: main becomes an ancestor, HEAD is byte-identical to dev in 1184 of the 1192 divergent files, zero conflict markers, registry.json intact (176 entries, no dupes), alembic single head, every declared range satisfied by its resolved version. The red `regression diff` is a degraded-matrix false positive — base seed 99999 was cancelled (#2019 flake), and the flagged test fails on plain dev under that same seed in two other recent runs. Merging with a merge commit.

Issue #1150, opened 2026-06-11
Timeline views: display loop executions as a distinct bucket/color instead of folding into Scheduled


vybe:
## Summary

Executions triggered by sequential agent loops (`triggered_by="loop"`, #740) are visually indistinguishable in the timeline views. In the Agent Detail Overview tab's executions-by-type stacked bar chart they are folded into the **Scheduled** bucket, and in the Executions list the `loop` trigger chip has no color mapping and falls back to gray. Loop runs should be a first-class category with their own color and legend entry.

## Context

`_TRIGGER_BUCKETS` in `src/backend/db/schedules.py` maps `"loop": "Scheduled"`, so the Overview timeline (#1107) merges loop iterations with cron-scheduled and webhook runs. An agent running a 50-iteration loop looks like a burst of scheduled work, which misrepresents what the agent is doing. With the Loops tab now shipped (#1106), loops are a visible product surface and deserve their own legend entry in analytics.

## Acceptance Criteria

- [ ] `loop`-triggered executions map to a new `Loops` bucket in `_TRIGGER_BUCKETS` / `_BUCKET_ORDER` (`src/backend/db/schedules.py`), surfaced by both the per-schedule analytics (#868) and agent Overview analytics (#1107) endpoints
- [ ] `BUCKET_COLORS` in `src/frontend/src/components/OverviewPanel.vue` assigns the new bucket a distinct design-system color; the StackedBarChart legend and per-segment tooltips show "Loops"
- [ ] `triggerLabelClass` in `src/frontend/src/components/ExecutionsPanel.vue` gets a `loop` entry with the same hue so the trigger chip matches the chart color
- [ ] Existing buckets/colors are unchanged; `Other` catch-all behavior preserved (a new unmapped trigger still surfaces as Other)
- [ ] Dark mode variants provided for the new color

## Technical Notes

- Bucket order is shared between backend (`_BUCKET_ORDER`) and frontend (`BUCKET_COLORS` comment notes they must match) — update both sides together.
- `src/frontend/src/components/StackedBarChart.vue` is generic (CSS/flexbox stacked bars) and should need no changes beyond the new bucket flowing through `buckets`/`colors` props.
- Check the Operations page Executions tab filter dropdown (`triggered_by` filter) also offers `loop` as an option if it enumerates triggers.

PR #2189, opened 2026-08-14
fix(analytics): give room turns their own bucket; keep them out of the alert set (ent#220)

dolho:
The OSS half of trinity-enterprise#220's cross-repo residuals. The enterprise half (chain
shielding + perf/atomicity) is trinity-enterprise#401.

Two residuals from the ent#169 review, decided in **opposite directions**, because they answer
different questions.

## `_TRIGGER_BUCKETS` — fixed

`room` was unmapped, so every room turn landed in `Other`. That bucket means *"a trigger nobody
has classified yet"*; letting the Workspace's entire traffic live there quietly redefines it as
"rooms", and hides a distinct spend shape — one human message can bill N agent turns.

Now a first-class `Rooms` bucket, placed in `_BUCKET_ORDER` before the catch-all — the same
treatment `loop` (#1150) and `reminder` (#1296) each got, for the same reason.

## `_AUTONOMOUS_TRIGGERS` — deliberately NOT changed

That set means *"no human will see the reply, so an unresolved slash command earns an Operating
Room alert"*. `room` does not belong in it:

- a room turn's reply — **including its failure line** — is posted into the room's own
  transcript, a durable client-facing surface the reader is already looking at. It is the
  `chat`/`public` case, not the `schedule` case;
- `room` is the one trigger an **untrusted external Workspace client** can drive. Alerting on it
  would route a client's typo into the operator queue, one alert per unresolved command —
  operator fatigue by construction (#1632's concern) for a message the room already shows.

The reasoning is now in the code next to the set, because the absence otherwise reads as an
oversight — which is exactly how it got filed as a residual.

## Tests

`tests/unit/test_ent220_room_trigger_bucket.py` — the bucket and its position before `Other`,
`Other` going back to meaning *unclassified*, every mapped bucket having a stack position (the
#1107 invariant that keeps a chart and its legend in agreement), and **both directions** of the
alert-set decision, including that nothing was silently dropped from it.

201 passed / 0 failed across the analytics, trigger-bucket and rooms suites.

## Note for the grid tiles (ent#96 / epic ent#94)

`Rooms` reaches the executions tile and the #1107 Overview chart through the backend-served
`trigger_order`, so neither needs a frontend change. It renders with the catch-all colour until
a `--gv-bk-rooms` token is added — I'll fold that into #2184 rather than here, since that PR
owns the tile's token map.

Related to trinity-enterprise#220

### DIFF
--- a/src/backend/db/schedules/analytics.py
+++ b/src/backend/db/schedules/analytics.py
@@ -41,6 +41,13 @@ _TRIGGER_BUCKETS = {
     "schedule": "Scheduled", "webhook": "Scheduled",
     "loop": "Loops",  # #1150: first-class bucket so loop bursts don't read as cron load
     "reminder": "Reminders",  # #1296: agent self-direction, not operator cron
+    # ent#220: a room turn is an agent woken by an @mention in a shared room.
+    # It was unmapped, so every one landed in `Other` — the catch-all that is
+    # supposed to mean "a trigger nobody has classified yet", quietly turned
+    # into "rooms". A first-class bucket for the same reason Loops and
+    # Reminders got one: room traffic is a distinct spend shape (one human
+    # message can bill N agent turns) and reading it as unclassified hides that.
+    "room": "Rooms",
     "agent": "Agent-to-agent", "fan_out": "Agent-to-agent",
     "self_task": "Agent-to-agent", "a2a": "Agent-to-agent",  # ent#157 inbound A2A tasks
     "voip": "Voice", "voice": "Voice",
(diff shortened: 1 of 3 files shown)


vybe (approved):
Validated via /validate-pr.

Verified `room` is a real `triggered_by` value (already in `_VALID_TRIGGERS`, produced by the enterprise shared-sessions path), so the unmapped-into-`Other` defect is real and the new bucket is placed correctly in `_BUCKET_ORDER` before the catch-all — same treatment `loop` (#1150) and `reminder` (#1296) got.

The `_AUTONOMOUS_TRIGGERS` non-change is the right call and the reasoning belongs in the code: a room turn's reply, failure line included, lands in the transcript the reader is already watching, and `room` is the one trigger an untrusted Workspace client drives — alerting on it would route a client's typo into the operator queue (the #1632 concern).

Security clean, 3 files, no packaging gaps. Tests pin both directions of the decision plus the bucket/order invariant.

Note: `Related to trinity-enterprise#220` is cross-tracker, so `issue-status-on-merge.yml` (same-repo regex) will not promote it — I'll bump ent#220 manually once its enterprise half (ent#401) also lands.

PR #2190, opened 2026-08-14
chore(enterprise): bump submodule to ent#401 (rooms chain shield + residuals)

trinity-ability:
Moves the enterprise submodule pointer `dcf0b528` → `65182c14` — exactly one commit, the [ent#401](https://github.com/abilityai/trinity-enterprise/pull/401) merge.

## Why this is a separate PR

The public tree pins `src/backend/enterprise` by SHA, so merging a PR in `trinity-enterprise` does not by itself ship that code. Until the pointer moves, the rooms chain shield and the ent#220 item-6 transactionality exist in the private repo and are **absent from every install built from this tree** — with CI green throughout, because nothing in the public suite can observe the gap. Closing that lag promptly is the whole point.

## What it carries

`fix(rooms): shield the mention chain from client disconnect; residual perf/atomicity (ent#220)`

- **Item 2** — `asyncio.shield` around each mention wake, so a poster closing their browser can no longer truncate a conversation other participants are watching (`CancelledError` is a `BaseException` and sailed past every `except Exception`; a turn already dispatched and billed lost its reply, silently).
- **Item 6** — room creation collapsed into one transaction (a half-created room 404'd its own creator), batched sidebar reads, and a `_require_membership` reordering that closes a timing oracle on the existence bit the uniform 404 withholds.

Pairs with #2189, the OSS half of ent#220 (the `Rooms` analytics bucket), already on `dev`.

Related to abilityai/trinity-enterprise#220


vybe (approved):
Approving. Verified the three things a pointer bump can get wrong:

**1. The pin is the enterprise `main` tip, not a stale or side-branch SHA.** `65182c14` is byte-identical to `Abilityai/trinity-enterprise@main` right now (`compare/main...65182c14` → `identical`). This is the failure mode that ships green and 403s every entitled feature in production, so it is worth restating that it was checked against the *current* tip and not merely against the PR that introduced it.

**2. It is exactly one commit, fast-forward.** `compare/dcf0b528...65182c14` → `ahead_by=1, behind_by=0` — no unrelated work riding along, nothing dropped. The single commit is the ent#401 merge the body describes.

**3. Nothing public is touched.** The three changed files are all inside the private tree (`backend/shared_sessions/db.py`, `backend/shared_sessions/service.py`, `tests/test_220_chain_shield_and_residuals.py`), so the OSS build surface is unchanged and `backend boots without enterprise submodule` remains meaningful.

On the red `pytest (head, seed 99999)`: it is not a regression and not a hang. The job was **cancelled**, not failed — `"Run unit suite"` shows `conclusion: cancelled` at 25m17s against a 25-minute cap, while the other five shards on the same run passed in ~11–12 min. A genuine per-test hang would have *named itself* since #2021 put `--timeout=300 --timeout-method=signal` on this suite precisely so a stall becomes an ordinary named failure instead of a job cancellation; that flag was present on this run. What is left is the runner variance #2019 documents (a shard running ~2.5× its siblings pushes an 11-min suite past 25). I have updated the branch onto current `dev` so it re-runs from scratch rather than re-litigating a stale result.

No closing keyword here, and that is correct rather than an oversight: this PR does not fix ent#220 — enterprise PR #401 did, in its own repo, where the same-repo automation applies. ent#220 is already `status-in-dev`, so nothing strands. A pointer bump is a shipping vehicle, and "Related to" is the honest verb.

Issue #2045, opened 2026-08-06
bug(ci): test_1081_physical_meter ceiling-clamp test fails under full-suite ordering — settings_service caches a pre-eviction db


obasilakis:
## Summary

`tests/unit/test_1081_physical_meter.py::TestPerAgentMeter::test_available_floors_at_zero_over_ceiling[sqlite]`
fails when the full unit suite runs, and passes 14/14 when its file runs alone.
`pytest-randomly` is installed, so file order is randomized per run — which makes
this a coin flip on ordering rather than a deterministic failure.

Found while verifying an unrelated branch (#2044 / ent#243). **Pre-existing on
`dev`** — confirmed by running the identical command against a clean, unmodified
`origin/dev` worktree.

## Evidence

Clean worktree at `origin/dev` @ `47f8843b`, no changes of any kind:

```
$ git worktree add ~/wt-base --detach origin/dev
$ cd ~/wt-base/src/backend
$ python -m pytest ../../tests/unit -q -p no:warnings -x --timeout=120

FAILED ../../tests/unit/test_1081_physical_meter.py::TestPerAgentMeter::test_available_floors_at_zero_over_ceiling[sqlite]
1 failed, 2657 passed, 12 skipped in 92.58s
```

In isolation, same worktree:

```
$ python -m pytest ../../tests/unit/test_1081_physical_meter.py -q -p no:randomly
14 passed in 3.51s
```

Two independent full-suite runs both stopped at this same test after **different**
pass counts (2657 and 3821) — consistent with `pytest-randomly` drawing a
different seed each run and `-x` halting at the same landmine from different
positions.

## Likely mechanism (structurally verified; not yet reduced to a minimal pair)

- `src/backend/services/settings_service.py:18` is `from database import db` — it
  binds the `db` **object** at module-import time.
- `tests/unit/test_1081_physical_meter.py` deliberately evicts `database` (and
  `db.connection`, `db.schedules`, …) from `sys.modules` so production code
  re-resolves against the `db_harness` engine, and snapshots/restores those names
  around every test (Issue #762 hygiene, `_STUBBED_MODULE_NAMES`).
- **`services.settings_service` is not in that snapshot list.** If any earlier
  file in the session has already imported it, it keeps a reference to the
  *pre-eviction* `db`.
- The failing test then does `set_ceiling(2)` — writing `max_parallel_tasks_ceiling`
  through the re-resolved harness `database` — while the clamp is read through
[...]

Issue #686, opened 2026-05-06
bug: MCP-triggered chat executions fail with 'Silent launch failure' (regression / parallel codepath of #279)


pandore:
## Summary

Same symptom as the closed #279 — `Silent launch failure: no Claude session created within 60 seconds` — but the **failing execution was MCP-triggered**, not Chat-UI-triggered. #279 was scoped to `routers/chat.py`, so this is either an incomplete fix or a parallel dispatch path with the same bug.

The agent (`tom-marketing-ops`) is fully healthy, claude-runtime-available, and 4 schedule-triggered executions in the same 1-hour window all succeeded normally (12–15s each). The single MCP call hung 70s with no Claude session ever created.

## Component

Backend / MCP execution dispatch (`mcp__trinity__chat_with_agent`)

## Priority

P2 — affects agent-to-agent / external MCP automation, including any client that uses `chat_with_agent` from outside the Chat UI. Reproducible in production today.

## Error

```
Silent launch failure: no Claude session created within 60 seconds
```

## Concrete reproduction (production)

| | |
|---|---|
| Agent | `tom-marketing-ops` |
| Trigger | MCP (`triggered_by: "mcp"`) |
| MCP key | `QId1YLMRh00gZUP8LaKnCg` (name: `Test1`) |
| Source user | `o.nikitin@ability.ai` |
| Execution ID | `19PQqFfJznqXQf0o7DVkVA` |
| Started | 2026-05-06T19:10:33.307092Z |
| Completed | 2026-05-06T19:11:42.997219Z |
| Duration | 69690 ms |
| Status | failed |
| `claude_session_id` | `null` (never created) |
| `model_used` | `null` |
| `business_status` | `null` |
| Message size | ~5 KB (multi-section markdown brief) |

### Same agent, same hour, schedule-triggered → succeeds

| Execution ID | Trigger | Duration | Status |
|---|---|---|---|
| `ZwT7DiuPCOYdf6UeK-ZXfw` | schedule | 14596 ms | success |
| `RtythunivKGbkuk0WQSOVw` | schedule | 12808 ms | success |
| `8hPn1vLunHEV7497Wk1aLg` | schedule | 11689 ms | success |
| `cmO03cBzpBXkmRvf3zX1Xg` | schedule | 11952 ms | success |

Schedule path → fine. MCP path → silent fail.

### Health check at time of failure

```json
{
  "aggregate_status": "healthy",
[...]

PR #2192, opened 2026-08-14
fix(agent-runtime): stop killing a fan-out turn mid-wait and calling it success (#2127)

### LINKED ISSUE #2127: bug: turn ending with background subagents still in flight records the execution as SUCCESS and silently discards the work
## Summary

In a headless execution, if the model spawns background subagents (the `Task`/`Agent` tools, or a background workflow) and then **ends its turn intending to wait for the completion notification**, the CLI exits normally with `stop_reason=end_turn`. Trinity's teardown then kills every one of those subagents (pgid SIGTERM/SIGKILL followed by `kill_cgroup_orphans`), and the execution is recorded as **`status=success`** with the model's "I'll wait for the notification" placeholder as its stored `response`.

No error is raised, nothing is logged as a failure, cost is billed, and the actual work is lost. The operator has no signal at all: the run looks like a normal successful execution in the UI.

This is the sibling branch of #1870. There, a late `<task-notification>` *does* arrive after `end_turn`, the CLI reports `error_during_execution`, and a completed turn is wrongly discarded. Here nothing arrives, so a turn that completed *no* work is wrongly recorded as a success.

## Component

Agent Runtime (agent base image headless executor)

## Priority

P2 — a workaround exists (instruct the agent to work inline only), but it is undiscoverable without reading the `schedule_executions` row by hand, and each occurrence is billed.

## Observed

A single MCP-triggered async execution:

| Field | Value |
|---|---|
| `status` | `success` |
| wall clock | 909 s |
| `cost` | ~$0.65 |
| `execution_log` | 392 KB |
| `response` | 221 characters |

The entire stored response was the announcement, not a result:

```
The workflow is running — 6 parallel agents reading all corpus files
simultaneously. I'll wait for the notification and then write [...]
```

The six subagents were killed at teardown. Subsequent runs of the same task, with the prompt explicitly forbidding background subagents, completed normally — which confirms the work itself was viable and only the turn structure caused the loss.

## Location

- Result classification: `docker/base-image/agent_server/services/headless_executor.py` (the `error_during_execution` / empty-result / clean-exit branches, ~L1090-1180)
- Teardown: `docker/base-image/agent_server/utils/subprocess_pgroup.py`, `docker/base-image/agent_server/utils/orphan_sweep.py`

## Root Cause

The result classifier has branches for non-zero exit, `error_during_execution` (#1870 recovery), a lost `result` line (#520/#640/#678), and an empty response on clean exit (#160). None of them inspect whether the final assistant turn left background subagent invocations without matching completions.

A clean `end_turn` is therefore trusted unconditionally, which makes a turn that merely *announced* future work indistinguishable from a turn that *did* the work.

Note that the teardown itself is correct and should not change: cgroup-membership killing is the #817 guarantee, and background work legitimately cannot outlive the process. The defect is that the platform reports success for a turn whose work it just killed.

## Reproduction Steps

1. Trigger any agent execution (schedule, chat, or MCP) with a prompt whose natural shape makes the agent fan out to background subagents, for example "run this multi-file ingestion to completion".
2. The agent spawns the subagents and ends its turn to wait for the notification.
3. The execution completes as `success` well before the subagent work could have finished.
4. Inspect the row: `status=success`, `cost` populated, `execution_log` large, and `response` containing only the announcement.

## Suggested Fix

Either or both:

**1. Detect it in the result path.** Before accepting a clean `end_turn`, scan the session transcript for background-subagent invocations with no corresponding completion. On a mismatch, either fail the execution with an explicit error, or store the response with a prepended warning so it is visible in the UI and in `/error-analysis`.

**2. Prevent it at the source.** State the constraint in the base-image agent instructions — background work cannot outlive the turn, so everything must run inline — and consider adding the background/subagent tools to the default `disallowed_tools` for headless executions, since they cannot function correctly there by construction.

## Environment

- Trinity: post-v0.8.5 dev
- Runtime: Claude Code headless (`-p`), agent base image

## Related

- #1870 — completed turn discarded on `error_during_execution` (the opposite branch of this same race)
- #1849 — `[ede_diagnostic]` marker masks the real error text
- #1853 — `error_during_execution` runs store zero telemetry
- #817 — cgroup orphan sweep (the teardown mechanism; working as designed)
- Upstream: `anthropics/claude-code#82235`

webmixgamer:
## Summary

An agent that fans out to background subagents announces *"I'll wait for the notification"* and ends its turn. Trinity read that as "the turn is over", killed the process group 2s later, and recorded the run as **success** — with the announcement stored as the answer, the work killed, and the cost billed. No error, no log, indistinguishable in the UI from a real result.

`claude --print` emits one `{"type":"result"}` per turn **segment**, not per run, and deliberately stays alive between them to await background subagents ("their result is part of the final output" — headless docs). #970's early-completion treated any result line as definitive and killed the CLI *during a wait it was performing correctly*.

**#1870 is the same root cause** with different timing (the notification lands early enough that the interrupted follow-on turn surfaces as `error_during_execution`), so one defect had been filed twice from opposite ends. The issue's own stated root cause — that the classifier trusts a clean `end_turn` — is wrong: the classifier never runs, because the process is killed first. Building its suggested transcript-scanner would have been a detector layered on our own kill.

## The fix

Early-finalize now requires **all three**:

```python
result_seen  AND  not waited_bg_tasks  AND  (now - last_stdout_at) >= _IDLE_FINALIZE_S
```

**Ledger half** — `system/background_tasks_changed` carries a full snapshot of in-flight work, so the latest event is the whole truth. `_NON_WAITED_BG_TASK_TYPES` is a **denylist** (`local_bash`, `local_shell`): an allowlist that hasn't heard of a future waited type restores the bug *silently*, while a denylist miss costs a bounded wait — and a background shell must never hold the finalize, since the CLI kills it after ~5s and gating on one re-opens #970's lingering-child class.

**Idle half, not optional** — the ledger drains to `[]` while the model is still composing the synthesis the subagents were spawned to produce. Measured: 5 subagents all reported by t+27.8s, real `FINAL REPORT` at **t+55.9s**, so a ledger-only gate stands aside and the kill lands on an interim line. Idleness is what #970's failure *is*: claude alive, blocked on an MCP child, emitting nothing.

`_IDLE_FINALIZE_S` = **300s** (`AGENT_IDLE_FINALIZE_S`; `<=0` refused, it would restore the exact kill). Set from the longest silence a *healthy* run produces — without `--include-partial-messages` one long message is a single stdout line emitted only when complete, measured at **40.54s of dead air for 9,840 chars**. ~7.4x headroom, biased large because the errors are asymmetric: too large only delays a pathological finalize, too small destroys a deliverable and calls it success.

Free on the happy path — the branch is only reachable when the process does **not** exit — and `effective_timeout` remains the unconditional backstop.

Also raises `CLAUDE_CODE_PRINT_BG_WAIT_CEILING_MS` to the execution timeout at **both** spawn sites (the chat path never had the kill but inherits the same 10-minute CLI truncation; fixing only the reported path is the recurring one-of-two-call-sites escape, #686/#1264), and prepends a visible notice + `background_tasks_pending_at_exit` when the CLI stops waiting on its own.

**Nothing is taken away from agents** — subagents keep working exactly as in a local session.

## Changes

| File | What |
|---|---|
| `agent_server/services/headless_executor.py` | ledger tracking, idle clock, the three-part gate, CLI ceiling, pending-at-exit notice |
| `agent_server/services/claude_code.py` | CLI ceiling on the chat spawn site |
| `agent_server/models.py` | `background_tasks_pending_at_exit` (additive; backend reads metadata as a dict) |
| `tests/unit/test_2127_background_task_gate.py` | 34 tests |
| `docs/memory/feature-flows/parallel-headless-execution.md` + index + `learnings.md` | flow, revision row, bug-class entry |

## Test Plan

- [x] `pytest tests/unit/test_2127_background_task_gate.py` — 34 passed
- [x] Both halves verified **failing without their own condition**, replaying streams measured from a live container:
  - no ledger gate -> stores `'Waiting for the second notification (BRAVO)'` instead of `'RESULTS: ALPHA BRAVO'`
  - no idle gate -> stores `'ECHO reported.'` instead of `'FINAL REPORT ...'`
- [x] Headless regression suites — 152 passed
- [x] Full `tests/unit` — 10,049 passed (residual failures are pre-existing: the #2045 order-dependent flake, and `test_pat_propagation_properties::TestKnownGaps`, which exercises `src/backend` code this PR doesn't touch)
- [x] Live smoke on a patched agent container: fan-out turns complete and return the full consolidated answer

**Honest caveat:** the end-to-end failure is a race that needs a specific CLI flush timing; four live runs never reproduced it, before or after. The deterministic unit tests are the proof — the live runs are a non-regression check only.

## Deployment

Requires a **base-image rebuild + cold agent recreate**. A restart is not enough.

**Known bound:** agents with a timeout under ~300s lose #970's early finalize — a rare stuck-process case now returns an honest 504 instead of a rescued success. Accepted over clamping the ceiling to a fraction of the budget, which would re-admit silent truncation on the agents least able to afford it. Tunable via `AGENT_IDLE_FINALIZE_S`, and pinned by a test.

**Consequence:** fan-outs now run to completion, so they spend more tokens and can newly reach `--max-turns`.

Fixes #2127


vybe (approved):
Validated via /validate-pr. 20/20 checks green; re-verified clean against dev@a1356143 after #2194 landed (the nightly bot's conflict comment is a stale false positive — branch is 0 commits behind dev).

The diagnosis is the valuable part: `claude --print` emits one result per turn SEGMENT, so #970's early-completion was killing the CLI during a wait it was performing correctly — and #1870 is the same root cause filed from the other end. Notably the issue's own stated root cause (classifier trusts a clean end_turn) is wrong, and building its suggested transcript-scanner would have been a detector layered on our own kill. Good catch to check rather than implement.

Both halves of the three-part gate were verified failing without their own condition, replaying streams measured from a live container — the ledger-only gate provably stands aside during the 28s window between the ledger draining and the real FINAL REPORT. The denylist direction is argued from failure cost rather than taste, and `effective_timeout` stays the unconditional backstop, so this can defer the kill but never remove it.

Config packaging is the part that usually fails on a change like this and it's complete: all three env writers with the right per-site semantics (set-only on create, set-or-CLEAR on the config-drift recreate since that dict is seeded from the old container, set-only on rebuild-from-nothing), both compose files, .env.example — and test_1811_recovery_container_parity.py extended, so the CLASS is guarded, not just this instance.

The unlocked reader-thread -> wait-loop sharing is fine: both fields are single-attribute rebinds with no invariant spanning them, and last_stdout_at is seeded at spawn before the readers start.

Two things for the release notes, neither blocking:
- Requires a base-image rebuild + cold agent recreate; a restart is not enough, so this ships inert on the running fleet without it.
- Accepted deltas, honestly disclosed: agents with a timeout under ~300s lose #970's early finalize, and fan-outs now run to completion (more tokens, can newly reach --max-turns).

Not blocking on the missing /review — the diff touches no SQL and no auth surface.

Issue #250, opened 2026-04-03
fix: Replace budget display with actual token usage tracking


vybe:
## Problem

The current budget widget on the agent detail page is meaningless:
- It estimates costs based on assumptions about API keys
- It resets on agent/container restart
- It displays fabricated data that doesn't reflect reality

## Solution

1. **Remove** the current budget/cost display entirely
2. **Replace** with actual **tokens used** display
3. Pull token usage from execution history data already stored in the database

## Acceptance Criteria

- [ ] Current budget widget removed from agent detail view
- [ ] New "Tokens Used" display shows actual token consumption from execution records
- [ ] Data persists across agent restarts (sourced from DB, not container state)
- [ ] Data persists across container restarts (verify token counts survive `docker restart`)
- [ ] Validated: stop/start agent and confirm token usage data is unchanged

vybe:
Shipped in **v0.6.0** (released via #998).

PR #2194, opened 2026-08-14
fix(stats): 7d cost sparkline was flat on PostgreSQL — DATE() bucket never matched the axis (#2193)

### LINKED ISSUE #2193: bug(stats): agent-header 7d cost sparkline is permanently flat on PostgreSQL — `DATE()` bucket key never matches the axis
## Symptom

On the Agent Detail header, the yellow 7-day cost sparkline renders as a flat line at zero, while every scalar beside it on the same row is correct (`Today $100`, `^56% vs avg`, `Lifetime $1,218 · 1002 runs`).

Only reproduces on PostgreSQL instances. SQLite installs are unaffected, which is why it never showed up locally.

## Root cause

`db/schedules/stats.py::get_agent_token_stats` buckets the 7-day series with a **date function** on a TEXT column:

```sql
SELECT DATE(started_at) as day, ... GROUP BY DATE(started_at)
```

`started_at` is an ISO-Z **TEXT** column (Invariant #16).

* **SQLite** — `DATE()` returns TEXT `'2026-08-14'`.
* **PostgreSQL** — `date(x)` is the type-name-as-function cast, so the column comes back as a `date`, which psycopg maps to `datetime.date`.

The gap-filled axis right below is keyed by `strftime("%Y-%m-%d")` **strings**:

```python
raw_days = {row["day"]: row for row in day_rows}
...
d = (now_utc - timedelta(days=i)).strftime("%Y-%m-%d")
if d in raw_days:      # str vs datetime.date -> never matches on PG
```

A `str`/`date` dict lookup does not raise — it **misses**, and a missed bucket is indistinguishable from a day with no executions. So the whole series reads as legitimate zeros. The scalar `cost_24h` / `cost_7d` / `lifetime_*` fields come from a *different* query that never touches `DATE()`, which is why they stayed right and the failure looked cosmetic.

### Verified

Same input rows, same code, both backends:

```
[sqlite]   daily_breakdown costs = [0.0, 0.0, 0.0, 25.0, 0.0, 40.0, 100.0]   total=165.0
[postgres] daily_breakdown costs = [0.0, 0.0, 0.0,  0.0, 0.0,  0.0,   0.0]   total=0.0
           (cost_24h=100.0  cost_7d=165.0  lifetime=165.0 on BOTH)
```

## Why it survived

This is the last dialect-dependent date function in the backend. Every sibling series already uses the substring idiom — `get_agent_analytics` (#1107, `substr(started_at, 1, 10)`), `get_fleet_execution_timeline` (ent#326), the audit heatmap (Python-side bucketing) — so the #1107 Overview chart on the *same page* renders correctly while this one is flat. `get_agent_token_stats` (#250) predates the convention and was missed by the #1540 sweep, which repointed the canary collectors at the live engine but not this accessor.

Unit tests are SQLite-only unless `TEST_POSTGRES_URL` is set, and no workflow sets it — so a behavioural dual-backend test passes vacuously in CI.

## Fix

1. Bucket with the house idiom: `substr(replace(started_at, ' ', 'T'), 1, 10)` — dialect-agnostic, returns text on both backends, slices the same UTC string the row was written with. `replace` also covers pre-#1474 scheduler rows (space separator, no `Z`), which clear the lexicographic cutoff and are counted by the scalar query — bucketing them as `2026-08-06 ` would drop them from the chart alone.
2. Normalize the join key in Python (`_day_bucket_key`) so a key-type mismatch can never again read as data.
3. Guard the class, not just the instance: a static check that this accessor uses no SQL date function, so CI catches a reintroduction on the SQLite-only path.

## Acceptance criteria

- [ ] `daily_breakdown` is identical on SQLite and PostgreSQL for the same rows
- [ ] `sum(daily_breakdown[].cost)` reconciles with `cost_7d` (modulo the rolling-168h vs 7-calendar-day edge)
- [ ] A pre-#1474 space-separator `started_at` lands in its day bucket instead of vanishing
- [ ] CI fails if a SQL date function returns to this accessor

trinity-ability:
Fixes #2193

## The bug

The yellow 7-day cost sparkline in the Agent Detail header renders as a flat zero line, while every scalar beside it on the same row is correct — `Today $100`, `^56% vs avg`, `Lifetime $1,218 · 1002 runs`.

PostgreSQL only. SQLite installs are unaffected, which is why it never showed up locally.

## Root cause

`get_agent_token_stats` (#250) bucketed the 7-day series with a **date function on a TEXT column**:

```sql
SELECT DATE(started_at) as day, ... GROUP BY DATE(started_at)
```

`started_at` is an ISO-Z **TEXT** column (Invariant #16).

* **SQLite** — `DATE()` returns TEXT `'2026-08-14'`.
* **PostgreSQL** — `date(x)` is the type-name-as-function cast, so psycopg hands back a `datetime.date`.

The gap-filled axis immediately below is keyed by `strftime("%Y-%m-%d")` **strings**:

```python
raw_days = {row["day"]: row for row in day_rows}
...
d = (now_utc - timedelta(days=i)).strftime("%Y-%m-%d")
if d in raw_days:      # str vs datetime.date -> never matches on PG
```

A `str`/`date` dict lookup does not raise — it **misses**, and a missed bucket is indistinguishable from a day with no executions. So the whole series read as legitimate zeros. The scalar `cost_24h` / `cost_7d` / `lifetime_*` fields come from a *different* query that never touches `DATE()`, which is why they stayed right and the failure looked cosmetic rather than like a broken query.

Nothing raised, on either backend.

## Verified on a real engine

Same rows, same code, both backends. Pre-fix:

```
[sqlite]   daily_breakdown = [0.0, 0.0, 0.0, 25.0, 0.0, 40.0, 100.0]   total=165.0
[postgres] daily_breakdown = [0.0, 0.0, 0.0,  0.0, 0.0,  0.0,   0.0]   total=0.0
           (cost_24h=100.0  cost_7d=165.0  lifetime=165.0 on BOTH)
```

Post-fix both backends return `[0.0, 0.0, 0.0, 25.0, 0.0, 40.0, 100.0]`.

## Why it survived this long

This was the **last dialect-dependent date function in the backend**. Every sibling series already uses the substring idiom — `get_agent_analytics` (#1107, `substr(started_at, 1, 10)`), `get_fleet_execution_timeline` (ent#326), the audit heatmap (Python-side bucketing) — which is why the #1107 Overview chart on the *same page* renders correctly while this one is flat. #250 predates the convention and was missed by #1540, which repointed the canary collectors at the live engine but not this accessor.

## The change

1. **Bucket with the house idiom** — `substr(started_at, 1, 10)`, byte-identical to the expression `get_agent_analytics` already uses, so the two day series cannot drift. Dialect-agnostic, returns text on both backends, and slices the same UTC string the row was written with. Pre-#1474 scheduler rows (`YYYY-MM-DD HH:MM:SS`, space separator) need no special handling: the separator sits at position 11, outside a 10-char slice. (ent#326's **hour** bucket does wrap the column in `replace(..., ' ', 'T')` — its 13-char slice spans the separator. An earlier revision of this PR copied that wrapper here on the strength of the similar name; see the correction note below.)
2. **Normalize the join key** (`_day_bucket_key`) so a key-type mismatch can never again read as *data*. The SQL fix is the fix; this is the belt, because the failure mode here is silent zeros rather than an exception.
3. **Guard the class, not the instance.**

## On the tests — read this before trusting the green check

`db_backend` yields **SQLite only** unless `TEST_POSTGRES_URL` is set, and **no CI workflow sets it**. So the nine behavioural cases in this PR *cannot* catch a reintroduction in CI — on SQLite the broken query passes every one of them. They are the proof on a real engine (run here against PostgreSQL 16; they fail 8/10 against the pre-fix query there), not the guard.

The guard that actually protects the class on SQLite-only CI is `TestNoDateFunctionInDayBucket`, an AST check over the **SQL string literals** the accessor issues, plus the pure `_day_bucket_key` normalizer exercised with the `datetime.date` a PG driver returns. It reads literals rather than source text on purpose: the first draft grepped the function source and failed against its own comment explaining the bug — a guard that fires on its own rationale is one nobody keeps. A meta-test pins that property so a later "simplification" back to a grep can't pass silently.

Verification that the suite is not vacuous:

| | pre-fix | post-fix |
|---|---|---|
| SQLite (CI default) | **8 failed**, 9 passed | 17 passed |
| SQLite + PostgreSQL | **8 failed** (PG behavioural) | 26 passed |

Neighbouring suites unaffected: `test_1771c_schedules_analytics_edges`, `test_agent_analytics`, `test_schedule_analytics`, `test_1115_schedules_summary` — 101 passed. Broader `-k "schedule or stats or executions or timeline"` sweep on the working branch: 680 passed.

## Correction after review

`/code-review` was run on this branch after the first push and caught a real error, kept here rather than force-pushed away because the reasoning is the point:

The day bucket originally shipped as `substr(replace(started_at, ' ', 'T'), 1, 10)`. The `replace` is a **provable no-op** for a 10-character slice — the separator is at position 11 — and its rationale had been transplanted from ent#326's *hour* bucket, where the 13-char slice genuinely spans it. Verified in Python and SQLite across ISO-Z, space-separator and `+00:00` shapes: identical output with and without.

It was not merely redundant. The AST guard asserted `"substr(replace(started_at" in sql`, so **the guard rejected `substr(started_at, 1, 10)`** — the codebase's own idiom — and would have handed anyone normalizing the three copies a red check with a false reason. The claim had also propagated into the feature-flow doc and a test billed as its coverage that passed identically without it.

Also fixed from the same review:

* **The guard now sweeps the whole `db/schedules/` package**, not one accessor. Per-accessor scoping protects the instance, not the class — and #250 surviving #1540 is exactly that failure. Matching moved from literal prefixes to a `\b`-anchored regex, catching the alias / whitespace / `CAST` / `::date` evasions a prefix match misses (the `\b` keeps `update(` from reading as `date(`).
* **Two cases were clock-flaky** — rows at `now-2h`/`now-1h` asserted against the *last* axis slot, which is a different day when CI runs between 00:00–02:00 UTC (~8% of clock positions). They now resolve the bucket by its own UTC date. Simulated at 00:30Z: the old assertion reads `0.0` against an expected `100.0`; the new one holds.
* `test_the_guard_reads_sql_not_prose` asserted on the *wording* of a source comment; it now proves the extractor property against a synthetic module.
* `_day_bucket_key`'s docstring overclaimed ("degrades to correct"); the call site is a dict comprehension, so colliding sub-day buckets would be last-wins. It now states the guarantee it actually provides.
* Removed a dead local re-import shadowing the module-level `datetime` import.

**Three pre-existing defects** the review surfaced in this accessor are filed as #2203 rather than folded in — the 168h-vs-7-calendar-day axis skew, the `status IN ('success','failed')` filter dropping legacy `error` rows its siblings count, and `avg_daily_cost` including today while its comment says it excludes it. None was introduced here; #2 and #3 change displayed spend figures and want their own review.

## Scope

Backend + the feature-flow doc that documented the broken query (`token-usage-display.md` stated `GROUP BY DATE(started_at)` as the bucketer — the line that would lead the next person to reintroduce it, sitting directly above an Invariant #16 note that covered only the cutoff half). No frontend change (`AgentHeader.vue` and `SparklineChart.vue` were drawing exactly the zeros they were handed). No schema change, so no migration on either track. Swept the rest of `src/backend` and `src/scheduler` for the same class — no other SQL date functions remain, and `get_agent_analytics`'s Python-side `.date().isoformat()` axis already joins string-to-string.

### DIFF
--- a/docs/memory/feature-flows/token-usage-display.md
+++ b/docs/memory/feature-flows/token-usage-display.md
@@ -59,9 +59,9 @@ Component: AgentHeader.vue TOKEN USAGE ROW
 
 Two SQL queries:
 1. Single-pass aggregation for lifetime, 24h, and 7d windows using `CASE WHEN started_at > ?` with `iso_cutoff()` helpers
-2. `GROUP BY substr(replace(started_at, ' ', 'T'), 1, 10)` for the per-day
-   breakdown (last 7 days) — a SUBSTRING, never a date function (see the
-   time-window invariant below)
+2. `GROUP BY substr(started_at, 1, 10)` for the per-day breakdown (last 7
+   days) — a SUBSTRING, never a date function, spelled exactly as
+   `get_agent_analytics` spells it (see the time-window invariant below)
 
 Gap-filling: iterates days 6..0 (oldest→today), zero-fills missing dates so sparkline always has exactly 7 data points.
 
@@ -102,10 +102,12 @@ bucket read as a legitimate zero, so the sparkline was permanently flat on every
 PostgreSQL install while `cost_24h` / `cost_7d` / `lifetime_*` (query 1, no date
 function) stayed correct. Nothing errored on either backend.
 
-`replace` covers pre-#1474 scheduler rows (`YYYY-MM-DD HH:MM:SS`, space
-separator, no `Z`): they clear the lexicographic cutoff and ARE counted by query
-1, so bucketing them as `2026-08-06 ` would drop them from the chart alone.
-`_day_bucket_key` normalizes the join key on the Python side as a belt, and
+Pre-#1474 scheduler rows (`YYYY-MM-DD HH:MM:SS`, space separator, no `Z`) need
+no special handling here — the separator sits at position 11, outside a 10-char
+slice, so both shapes yield the same day. ent#326's **hour** bucket does wrap the
+column in `replace(..., ' ', 'T')`, because its 13-char slice spans the
+separator; do not copy that expression into a day bucket on the strength of the
+similar name. `_day_bucket_key` normalizes the join key on the Python side, and
 `tests/unit/test_2193_token_stats_day_bucket.py` AST-guards the SQL literals —
 the behavioural cases cannot catch a reintroduction in CI, which runs SQLite-only
 unless `TEST_POSTGRES_URL` is set.
--- a/docs/memory/feature-flows/token-usage-display.md
+++ b/docs/memory/feature-flows/token-usage-display.md
@@ -53,13 +53,15 @@ Component: AgentHeader.vue TOKEN USAGE ROW
 
 ## Backend Layer
 
-### DB Method (`src/backend/db/schedules.py`)
+### DB Method (`src/backend/db/schedules/stats.py`)
 
 `ScheduleOperations.get_agent_token_stats(agent_name: str) -> Dict`
 
 Two SQL queries:
 1. Single-pass aggregation for lifetime, 24h, and 7d windows using `CASE WHEN started_at > ?` with `iso_cutoff()` helpers
-2. `GROUP BY DATE(started_at)` for per-day breakdown (last 7 days)
+2. `GROUP BY substr(replace(started_at, ' ', 'T'), 1, 10)` for the per-day
+   breakdown (last 7 days) — a SUBSTRING, never a date function (see the
+   time-window invariant below)
 
 Gap-filling: iterates days 6..0 (oldest→today), zero-fills missing dates so sparkline always has exactly 7 data points.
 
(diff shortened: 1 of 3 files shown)


vybe (approved):
Validated via /validate-pr. All 21 checks green; merges clean against dev@58df9aed (the nightly bot's conflict comment is a stale false positive — branch is 0 commits behind dev).

The fix is one expression, but the test design is what earns the merge: the PR states outright that CI is SQLite-only unless `TEST_POSTGRES_URL` is set, so the behavioural cases cannot catch a reintroduction — and then builds the guard that can (AST sweep over SQL string literals across the whole `db/schedules/` package, \b-anchored, catching the alias/whitespace/::date/to_date evasions a prefix match misses). Package-scoped rather than accessor-scoped, because #250 surviving #1540 is exactly what per-accessor scoping fails to prevent.

The on-branch /code-review correction is kept in history rather than force-pushed away, and it caught a real one: the transplanted `replace(...)` no-op that the guard itself asserted on, which would have rejected the codebase's own idiom with a false reason. Two clock-flaky assertions fixed alongside.

The three pre-existing defects in this accessor being filed as #2203 rather than folded in is the right call — two of them move displayed spend figures.

Issue #2207, opened 2026-08-15
bug: Codex runtime pins a retired model and mis-estimates cost for every current GPT-5.x model


vybe:
## Summary

Trinity's Codex runtime (#1187) still points at the `gpt-5.1-codex` generation. Two things followed from that and both are now wrong:

1. **The pinned model is not in OpenAI's current Codex catalog.** `config/agent-templates/test-codex/template.yaml` pins `runtime.model: gpt-5.1-codex`, and the flow/requirements docs repeat it. The current lineup is GPT-5.6 Sol / Terra / Luna (5.5, 5.4, 5.4-mini legacy; 5.4 retires 2026-08-31), and `gpt-5.2` / `gpt-5.3-codex` are already deprecated for ChatGPT sign-in. `gpt-5.1-codex` is not listed at all. Any agent created from a template carrying that pin risks failing at the model call.
2. **`CODEX_PRICING` is materially wrong for every model an agent would actually run today.** The table tops out at the gpt-5.1 family, so everything newer prefix-matches `gpt-5` and bills at $1.25 / $10 per M.

| Model | Trinity estimate (in / out per M) | Actual | Error |
|---|---|---|---|
| `gpt-5.6-sol` | $1.25 / $10 | $5 / $30 short ctx, $10 / $45 long | **4–8x understated** |
| `gpt-5.5` | $1.25 / $10 | $5 / $30 | 4x understated |
| `gpt-5.3-codex` | $1.25 / $10 | $1.75 / $14 (fast: $3.50 / $28) | up to 2.8x understated |
| `gpt-5.2` | $1.25 / $10 | $1.75 / $14 | 1.4x understated |
| `gpt-5.6-luna` | $1.25 / $10 | $0.20 / $1.20 | 6x **over**stated |

Codex has no native cost reporting, so this estimate *is* the number Trinity records. It feeds execution rows, agent analytics, cost threshold alerts, and — the sharp edge — loop cost budgets (`max_cost_usd`, #1155). A loop budgeted against a 5.6-sol agent runs roughly 4x past its ceiling before the boundary gate trips.

Two smaller items in the same blast radius:

3. **`CODEX_CONTEXT_WINDOW = 272_000`** (`src/backend/services/model_context.py`) is applied to every `gpt-5*` id. On the current rate card 272K is the *threshold between short- and long-context pricing tiers*, not a ceiling, so the `% context used` denominator understates capacity for the 5.6 family and makes agents look closer to a compaction wall than they are.
[...]

Issue #2208, opened 2026-08-15
bug: Codex API-key auth is inert — every turn 401s because nothing writes $CODEX_HOME/auth.json


vybe:
## Summary

Trinity's Codex runtime resolves an OpenAI API key and passes it to `codex exec`
**via the environment only**. That is no longer sufficient: current Codex CLI
authenticates its `wss://api.openai.com/v1/responses` transport from
`$CODEX_HOME/auth.json`, and with only `OPENAI_API_KEY` set every turn fails with
repeated `401 Unauthorized` and a non-zero exit.

Net effect: **an API-key Codex agent cannot complete a single turn.** Only
subscription agents (which already ship an `auth.json`, #1971) work.

## Evidence

Same key, same prompt, same image, both CLI versions — reproduced while
validating #2207:

| Setup | Result |
|---|---|
| `-e OPENAI_API_KEY=…`, codex **0.139.0** (current pin) | `exit=1`, 7x `401 Unauthorized` |
| `-e OPENAI_API_KEY=…`, codex **0.147.0** | `exit=1`, 7x `401 Unauthorized` |
| `printenv OPENAI_API_KEY \| codex login --with-api-key` first, then same turn | `exit=0`, result `PONG`, 0x 401 |

```
ERROR codex_api::endpoint::responses_websocket: failed to connect to websocket:
HTTP error: 401 Unauthorized, url: wss://api.openai.com/v1/responses
```

The key itself is fine — `GET /v1/models`, `POST /v1/responses` and
`POST /v1/chat/completions` all return 200 with it. The failure is specific to
the CLI's websocket transport, which reads `auth.json`.

**Not a regression from the #2207 CLI bump** — 0.139.0 and 0.147.0 fail
identically, so this predates it.

## Why it was invisible

- `_codex_auth_ready()` accepts *either* an API key *or* an `auth.json`, so the
  pre-flight gate passes and the failure surfaces later as a CLI auth error.
- `responses_websockets` / `responses_websockets_v2` are listed as **removed**
  in `codex features list`, i.e. the transport is no longer toggleable — there
  is no flag to fall back to the HTTP path that does accept a bare env key.
- Nothing in `startup.sh` or `codex_runtime.py` ever runs `codex login`
  (`grep -rn "codex login"` → no hits).

## Proposed fix

Materialise the resolved key into `$CODEX_HOME/auth.json` before the first
`codex exec`, e.g. run `codex login --with-api-key` (key on **stdin** — the
[...]

Issue #1155, opened 2026-06-11
feat(loops): per-loop cost budget enforced as a hard stop (max_cost_usd)


vybe:
## Summary

Sequential agent loops (#740) record per-run cost on `agent_loop_runs.cost` but never enforce a spend limit. A loop with `max_runs=100` on an expensive model has no cost bound at all. Add an optional `max_cost_usd` parameter to the loop API, enforced as a **hard ceiling checked before each iteration** — not an alert after the fact.

## Context

2026 loop-engineering practitioner consensus (and the canonical $47k runaway-loop incident) is unambiguous: cost budgets for autonomous loops must be hard limits enforced before the next call, not monitoring alerts that fire after spend occurs. We already persist cost per run, so accumulation is nearly free — the gap is purely enforcement.

## Acceptance Criteria

- [ ] `POST /api/agents/{name}/loops` accepts optional `max_cost_usd` (float > 0); persisted on `agent_loops`
- [ ] Before starting each iteration, the runner sums completed-run costs; if the next iteration would start with the budget already met/exceeded, the loop stops with `stop_reason="budget_exhausted"` and terminal status `stopped` (the in-flight run is never killed mid-turn — check is at the iteration boundary)
- [ ] Runs with NULL cost (failed/unknown) count as 0 toward the budget — fail-open on accounting, the run cap still bounds the loop
- [ ] `GET /api/loops/{loop_id}` returns `max_cost_usd` and accumulated `total_cost`
- [ ] MCP tool `run_agent_loop` exposes the new parameter (three-surface sync: router, MCP tool; no agent-server change needed)
- [ ] Loops UI (LoopsPanel) shows budget + spend when set
- [ ] Schema migration for the new column; tests for budget stop, no-budget unchanged behavior, NULL-cost runs

## Technical Notes

- `src/backend/services/loop_service.py` — iteration-boundary check in `_run()`
- `src/backend/routers/loops.py`, `models.py`, `db/` loop module + migration
- `src/mcp-server/src/tools/loops.ts`
- New `stop_reason` value `budget_exhausted` joins the existing enum (max_runs_reached | stop_signal_matched | user_stopped | error | interrupted)

Fame510:
> "$47k runaway-loop incident is unambiguous: cost budgets for autonomous loops must be hard limits enforced before the next call, not monitoring alerts that fire after spend occurs."

[...]

PR #2209, opened 2026-08-15
fix(codex): price every current GPT-5.x model, and fix the matcher that let the table rot (#2207)

trinity-ability:
## Summary

- **The rows were the symptom; the matcher was the defect.** `CODEX_PRICING` was stranded on the gpt-5.1 generation, so every newer model prefix-matched `gpt-5` and billed at $1.25/$10. Correcting only the rates would have left a `gpt-5.7-*` resolving to the oldest, cheapest entry — the identical bug, pre-armed, with the fix already merged. `_resolve_pricing`'s prefix match is now **boundary-aware**: `-` is a variant/date suffix, `.` is a new version.
- **This is a spend-control bug, not a metric bug.** Codex reports no cost, so this estimate *is* the recorded number, and it gates the loop budget `max_cost_usd` (#1155). Measured understatement: **3.6x for `gpt-5.6-sol`**, and **21x for `gpt-5.5-pro`** — a model the issue didn't list.
- **Cache writes are now priced.** They were going to be documented as unobservable; a real turn showed `cache_write_input_tokens` present, a SUBSET of `input_tokens` (the `reasoning_output_tokens` trap again) and 99.97% of that turn's input.

## Verification against the live rate card

The issue said *"re-check at implementation time rather than trusting this table."* Doing that changed the plan — its table was stale in two ways, and a live `GET /v1/models` **corrected its headline premise**:

| | Issue's claim | Verified 2026-08-15 |
|---|---|---|
| `gpt-5.1-codex` | "not in the catalog… risks failing at the model call" | **Still resolves for API-key callers** — which is Trinity's auth mode. Stale pin, not broken. |
| `gpt-5.6-terra` | absent from the table | $2 / $12 — a *recommended, everyday* Codex model |
| `gpt-5.4` / `-mini` | written off as legacy | Still on the card, Codex-supported until 08-31 |
| `-pro` tiers | not mentioned | **$30 / $180** — the worst mis-billing in the file |
| `gpt-5.3-codex` | "$1.75 / $14" | Not on the card → resolves to `default` rather than carrying an unsourceable rate |

## Changes

| File | What |
|---|---|
| `codex_runtime.py` | Rates restated **per 1M** (diffable against the card by eye); 5.6 / 5.5 / 5.4 / `-pro` added; long-context tier (>272K in → 2x in / 1.5x out); cache-write surcharge; boundary-aware matcher; `default` → flagship rate; `get_default_model` → `gpt-5.6-sol` |
| `model_context.py` (+ vendored copy) | 5.6 family → **1,050,000**. The old flat 272K is a *price break*, not a window — conflating them made the gauge read ~3.9x too full |
| `template.yaml` | Pins `gpt-5.6-sol` |
| `Dockerfile` | `@openai/codex` 0.139.0 → **0.147.0** (needs a base-image rebuild) |
| docs ×3 + `learnings.md` | Retired pin removed; cost section added |
| tests ×2 | +14 tests |

### Two decisions worth reviewing

**`default` is the flagship rate, not the cheapest.** An unpinned agent reaches it on *every* turn — `_CodexParseState.model` is caller-supplied and `thread.started` carries no model id, so Trinity never learns what the CLI resolved. Since that default *is* `gpt-5.6-sol`, this is the accurate answer, not a pessimistic one. It's also why the template **pins** rather than omitting, which the issue offered as an equal option.

**The two prefix matchers are deliberately asymmetric.** `model_context._FAMILY_PREFIX_WINDOWS` keeps its bare `startswith` on purpose: there, falling through picks the *smallest window* (over-reports usage = safe), while here it picks the *cheapest rate* (under-bills = dangerous). Same mechanism, opposite risk. Both files carry a comment forbidding "unifying" them.

## Test Plan

- [x] `pytest tests/unit/test_codex_runtime.py` — 122 passed
- [x] `pytest tests/unit/test_model_context_{catalog,parity}.py` — parity byte-identical (Invariant #5)
- [x] Full unit suite — 10,226 passed; the 6 failures (`test_ent96_timeline_split`, `test_ent314_hardened_yaml`) reproduce on clean `dev` with this diff stashed
- [x] Every rate asserted against **literal** card values, not derived from `CODEX_PRICING` — the existing cost tests read their expectations *from the table*, so they were structurally blind to a wrong rate, which was the whole defect
- [x] **AC #10 — real turns on the rebuilt image, priced by Trinity's own parser:**

| Turn | Exit | Result | Side effect | Cost | Context |
|---|---|---|---|---|---|
| chat | 0 | `PONG` | — | $0.071401 | 1,050,000 |
| task | 0 | `DONE` | wrote `hello.txt` = `trinity` | $0.079468 | 1,050,000 |

The task turn exercises all three input components (plain 6 + cached 11,402 + writes 11,486 = 22,894) and reconciles to the card exactly. The old table billed it $0.029268 — **2.7x under**.

## Found along the way — filed separately

**#2208: Codex API-key auth is inert.** Nothing writes `$CODEX_HOME/auth.json`, so with only `OPENAI_API_KEY` set the CLI's websocket transport 401s on every turn; `codex login --with-api-key` first makes it work. **Not a regression from the CLI bump** — 0.139.0 and 0.147.0 fail identically. Out of scope here (auth, not pricing), but it means API-key Codex agents can't currently complete a turn.

## Reviewer notes

- The Dockerfile bump requires a base-image rebuild; existing Codex agents adopt it on next **cold start** via the image-drift predicate (#1809), not a plain restart.
- `gpt-5.4` retires from Codex **2026-08-31** — included deliberately (still on the API rate card, and API-key callers may outlive the ChatGPT-sign-in retirement).
- Cache-write rates are present only for models OpenAI lists as supporting prompt caching. The 1.25x rule is published, but so is the model list — inferring the surcharge onto `-mini`/`-nano` would be the "never add a rate that is not on the card" violation this table exists to prevent.

Fixes abilityai/trinity#2207


vybe (approved):
Rebased onto current dev (`learnings.md` conflict resolved chronologically — the file is ascending, so this branch's two 08-15 entries precede dev's 08-16 one).

Validated:
- Vendored `model_context.py` is byte-identical to the backend copy (Invariant #5); parity test green.
- AST scan for the unreachable-after-`return` class this PR's own learnings entry warns about: none.
- Boundary-aware `_resolve_pricing` verified — a future `gpt-5.7-*` falls through to `default` (over-reports) rather than matching the cheapest legacy entry.
- Security scan §4.1–4.10 clean; no new backend env vars or top-level packages.
- Infra changes justified inline (codex CLI 0.139.0→0.147.0, template model pin).

Fixed on the branch: `architecture.md` line 416 still asserted a flat `Codex→272K`, which this PR makes family-dependent — and `gpt-5.6` (pinned as the test-codex default by this same commit) is 1.05M. Two further copies of the same claim in `feature-flows/model-selection.md` and `faq/chat-and-sessions.md` were corrected too; a catalog restated in four files that disagree is how the original value went stale.

Note for deploy: the pricing table lives in the agent image, so existing Codex agents keep the old estimate until a base-image rebuild + cold start.

Issue #2198, opened 2026-08-15
bug: Agent Detail and Workspace over-fetch on load (21 redundant requests, N+1 roster)


vybe:
## Summary

Opening one Agent Detail page issues **62 API calls, 41 unique — 21 redundant**. `/api/agent-dashboard/{agent}` alone is fetched **3× per poll cycle**. Separately, the Workspace bootstrap is N+1 across the roster: one `/sessions` call per agent on every load.

## Context

Found during an exploratory E2E sweep of `dev` @ `c1cc22bc`, measured via `performance.getEntriesByType('resource')`. Request start times separate genuine duplicates from legitimate polling.

### Agent Detail — one page load

| Endpoint | Calls | Start times (ms) | Verdict |
|---|---|---|---|
| `/api/agent-dashboard/{a}` | 9 | 220,488,548 · 3506,3568,3574 · 9548,9593,9611 | **3× per poll cycle** |
| `/api/agents/{a}` | 2 | 93, 93 | concurrent duplicate |
| `/api/agent-dashboard/{a}/exists` | 3 | 145, 158, 504 | duplicate |
| `/api/agents/{a}/info` | 3 | 156, 158, 9660 | duplicate |
| `/api/users/me`, `/api/settings/feature-flags`, `/api/agents/{a}/playbooks`, `/api/agents/{a}/avatar/emotions` | 2 each | ms apart | duplicate |
| `/api/agents/{a}/activity` | 4 | 145, 5148, 9612, 14614 | ✅ 5s poll — correct |

The duplicates fire **milliseconds apart** (two at the identical 93ms), so they are concurrent double-invocation on mount, not polling.

### Workspace — bootstrap

`/workspace` issues one `GET /api/enterprise/client-portal/agents/{name}/sessions` **per roster entry** (10 calls for 10 agents) on every load. Scales linearly with fleet size.

### Related smaller case

Navigating to a **non-existent** agent issues `GET /api/agents/{name}` twice (both 404) — same double-invocation pattern on mount. The 404 page itself renders correctly and is enumeration-safe.

## Context: why now

`c1cc22bc` — _"the agent page built the whole roster to render one agent (#2160)"_ — just fixed roster over-fetching on this same page. The N+1 is gone; the per-agent duplicate fetches remain.

## Acceptance Criteria

- [ ] Agent Detail issues each unique endpoint **once** on mount; no two identical requests within the same tick
[...]

PR #2219, opened 2026-08-15
fix(workspace): roster shows container availability honestly — never hides an agent (#2196)

### LINKED ISSUE #2196: bug: Workspace roster lists agents with no container (DB-sourced roster vs Docker-as-truth)
## Summary

`GET /api/enterprise/client-portal/my-agents` returns agents that `GET /api/agents` does not. The rows are **not** soft-deleted — they are live `agent_ownership` rows that simply have **no container**. An end user in the Workspace sees agents that do not exist anywhere else and cannot be used.

## Context

Found during an exploratory E2E sweep of `dev` @ `c1cc22bc`.

On the test instance the fleet list returned 7 agents while the Workspace roster returned 10, including 4 leftover rows from pytest runs:

```
/api/agents            → 7 agents
/api/enterprise/client-portal/my-agents → 10 agents (4 extra)
```

Checking `agent_ownership` for the extras:

```
deleted_at = NULL, is_system = 0, is_ephemeral = 0
```

So the #834 soft-delete filter is not the gap. The divergence is architectural:

- **Fleet list** is Docker-as-truth (Invariant #11) — it enumerates containers
- **Workspace roster** is DB-sourced — it enumerates ownership rows

…so any ownership row that outlives its container leaks straight through to what is the **customer-facing** surface.

The leaked rows in this instance are pytest residue, so the *trigger* was environmental. The product behaviour worth fixing is that the roster never reconciles against container existence. Canary **L-03** covers orphan rows pointing at *missing* agents, but not live rows with no container.

## Acceptance Criteria

- [ ] The Workspace roster does not list agents that have no container, OR lists them with an explicit unavailable state that explains why and offers a next action (no dead entry)
- [ ] Behaviour is decided deliberately and documented: reconcile against Docker, or treat the ownership row as authoritative and surface the state honestly
- [ ] Selecting an agent in that state fails gracefully rather than erroring opaquely
- [ ] The fleet list and the Workspace roster agree on which agents exist, or the intended difference is documented
- [ ] Consider whether a canary invariant should cover "live ownership row with no container" (L-03 currently does not)

## Technical Notes

- Roster endpoint: `GET /api/enterprise/client-portal/my-agents` (`src/backend/client_portal/`)
- Fleet list: `GET /api/agents` (`src/backend/routers/agents.py`)
- Related: Invariant #11 (Docker as source of truth), #834 (soft delete), canary L-03

Separately worth considering: pytest runs leaving `agent_ownership` rows behind is itself a test-hygiene gap that made this visible.

trinity-ability:
Fixes #2196

## Summary

The Workspace roster stays **membership-honest and availability-honest at the same time**: the `agent_ownership` row remains authoritative for *who is on the roster*, while container state is projected as a 4-state `availability` field (`running | stopped | no_container | unknown`) on each roster card. Resolution happens in `get_roster` / `get_agent_card` via **one sparse batched Docker read** for the whole roster (not N per-card inspections). `_agent_briefing` loses its N discarded per-card Docker calls; `portal_chat` gains the previously **missing liveness gate BEFORE user-turn persistence** (a turn against a container that cannot run no longer persists the user message and then fails opaquely); and the `_agent_is_running` fail-closed bug is fixed (a Docker read failure no longer reads as "not running").

## Design decision: project availability, never hide the row

Hiding containerless agents from the roster was considered and **rejected**:

- `list_all_agents_fast()` returns `[]` on **any** Docker fault. A roster filtered through it would empty a paying customer's entire Workspace on a Docker blip, daemon restart, or `DOCKER_GID` change — indistinguishable from "you have no agents". Membership must not be a function of Docker health.
- The **narrowed #1747 argument**: agent identity lives in `agent_ownership`, not in the container. #1747 fixed the DELETE handler treating "no container" as "agent not found"; this PR extends the same principle to the roster read path — the row is the agent, the container is its runtime state. (Deliberately *not* overclaiming that #1747 ruled on listings; it ruled on the identity model, which listings inherit.)

So the roster always shows every shared/owned agent, and each card says honestly whether it can currently serve a turn.

## Trap note: `sparse=True` + `attrs["Names"][0]`

The batched read uses `containers.list(all=True, sparse=True)`. Under sparse listing, docker-py's `container.name` is **`None`** and `container.labels` **raises** — code written against the non-sparse attribute surface fails *silently green* (every name comparison is `None != anything`, so the availability map comes back empty and every card degrades). Names are therefore read from `attrs["Names"][0]` (leading-`/` stripped) and labels from `attrs`. This exact failure mode is pinned by a test that exercises **real docker-py sparse fixtures**, not mocks of our own wrapper.

## Acceptance-criteria notes

- **AC #4 (fleet endpoint and roster agree literally): not achievable as written.** `GET /api/agents` has an admin-only branch that surfaces **orphan containers** (containers with no ownership row) — the roster, keyed on ownership rows, structurally cannot show those. The intended difference is documented in the feature flows; the two surfaces agree on every row that has an ownership row.
- **AC #5 (canary invariant): answered as "no canary".** L-03 keys on rows referencing an agent *absent from `agent_ownership`* — a containerless agent with a live ownership row cannot fire it. A new invariant would need Docker evidence (the H-01 evidence class) to avoid paging on every stopped-but-healthy agent. Instead: an **operator-notification follow-up** (listed below) — the batch availability map makes it nearly free.

## Deviations from plan (5, from the review audit)

1. Compat-shim `_agent_is_running` **kept** (ent#286 streaming coverage still calls it) rather than deleted.
2. `portal_chat` takes an `availability` parameter (resolved by the caller) instead of re-reading Docker inside the turn path.
3. Owner naming for the chip states resolved **frontend-side** (`portalUtils.js`) rather than shipping display strings from the backend.
4. The availability classifier is exercised by a **classifier-driven test** (table of Docker states → expected 4-state) instead of per-endpoint duplication.
5. `.claude` test-runner catalog update ships via **trinity-dev** separately (submodule pointer deliberately not bumped here).

## Verification

WAVE-4 `verify-local` (full stack: prod images + base image + real agent boot, at base `58df9aed`):

| Surface | Result |
|---|---|
| Unit suite | **10,358 passed** (8 pre-existing reds — see below) |
| Integration suite | **70 passed** |
| Boot + image build | green |
| `latest-result.json` | `status: pass` |

The 8 unit reds are **pre-existing**: proven byte-identical at base `58df9aed` (same 8 failures with this branch's commits absent).

Post-rebase drift guard (on `origin/dev` @ f5fe15f3): `tests/unit/test_2196_roster_availability.py` → **36 passed**; seam-pinned files `test_ent357_workspace_owned_roster.py` + `test_2128_workspace_rooms_capability.py` → **22 passed**.

### Unverified-by-machine (for the human reviewer)

- **§11.15** — a fail-open turn against an unavailable container degrades to a *clean* FAILED via the transport breaker; proving it end-to-end needs a branch-serving stack (not reproducible in the unit/integration harness).
- **§11.8** — the visual reflow of the header availability fact: no reserved footprint, matches the file's existing `last_active` idiom; needs an eyeball, not a test.

- Operator notification for containerless **shared** agents — the batch availability map makes it nearly free.
- `is_ephemeral` roster exclusion — an access change across ~10 routes with a NULL-safety trap; out of scope here.
- `/my-agents` rate limiter — the endpoint now carries a free liveness map worth bounding.
- Portal test hygiene — pytest residue rows in `agent_ownership` from older portal suites.

## Merge order

**This PR merges BEFORE #2198's** — both touch `client_portal/service.py`; #2198 rebases after this lands.

### DIFF
--- a/docs/memory/architecture.md
+++ b/docs/memory/architecture.md
@@ -155,7 +155,7 @@
 **Services (`services/`)** — 67 service modules:
 
 *Core:*
-- `docker_service.py` - Docker container management (single point of Docker interaction, Invariant #11)
+- `docker_service.py` - Docker container management (single point of Docker interaction, Invariant #11). Also owns the **tri-state** container-state pair `agent_container_states()` (batch, one `sparse=True` `containers.list()`) / `agent_container_state(name)` (single) added by #2196: `None` means *Docker could not be asked*, distinct from `{}` / `"missing"` meaning *Docker answered: no container*. Every older helper collapses those two into one falsy value, which is why a naive "drop agents with no container" reconcile empties a customer's roster on any Docker fault. `sparse=True` is load-bearing and constrains the code: the SDK default full-inspects **every** container, and under sparse `.name` returns `None` while `.labels` raises — so the key comes from `attrs["Names"][0]` and the status is classified **explicitly** to `running`/`stopped` rather than reusing `list_all_agents_fast`'s raw-status fall-through (which would pass `paused`/`removing` through to a consumer's `Literal`). `list_all_agents_fast`'s `[]`-on-fault contract is unchanged.
 - `docker_utils.py` - Docker utility helpers
 - `template_service.py` - GitHub template cloning and processing; owns the tolerant `credentials:` readers (`credential_shape_errors` / `credential_mcp_server_names` / `credential_env_file_names` / `credential_mcp_env_vars` / `declared_credential_names`) — read paths degrade + surface `credential_errors`, the `.env` writer raises `CredentialDeclarationError` → 400 (ent#128) — see [template-processing.md](feature-flows/template-processing.md). **ent#89:** both builders also surface the declared `schedules:` (normalized) + `schedule_errors` via `services/template_schedules.py`, the same tolerant-reader shape; **both GitHub catalog list paths are now fenced** per-template (`_safe_build_github_template`) — they were bare list comprehensions, so a raise in the untrusted GitHub builder 500'd the whole GitHub half of `GET /api/templates` (ent#128 PR-A fenced only the local path). `fetch_template_metadata_for_create(repo, pat, ref)` is the **creation-path** metadata read: resolved PAT + parsed ref, cache-bypassed, loud on any failure — the catalog cache uses the global platform PAT off the default branch and would silently return `{}` for a private repo read with a per-user token. **ent#14:** the GitHub half of the catalog resolves **admin override → remote registry (`template_registry_service`) → `DEFAULT_GITHUB_TEMPLATE_REPOS`** in **both** `get_all_templates()` and `get_github_template()` (the second feeds `GET /api/templates/{id}` *and* agent creation, so one ladder or the display fields diverge by surface); registry entries arrive as `admin_override` dicts and are fenced so any failure degrades to the bundled floor. The per-repo fetch *reason* is now cached beside the metadata and surfaced as `metadata_unavailable`, because "declares no `template.yaml`" and "could not read it" were the same value — and `fork_to_own: required` silently read as absent under a rate limit. `fetch_template_metadata_result_for_create` is the reason-preserving form of the creation read (the plain wrapper drops the reason, which is fine for schedules and wrong for a security gate): the `fork_to_own` gate decides from IT, not from `gh_template`, because the catalog read answers 404 for a private repo the platform PAT cannot see and would classify a `required` template as absent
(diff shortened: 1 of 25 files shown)


vybe (approved):
Validated via /validate-pr: security scans clean, all CI green (3 head seeds + regression diff), `Fixes #2196` closing keyword present, docs updated (architecture.md + feature flows). Merges clean against current dev.

The core design call is right and well-argued: **project availability, never filter the roster**. `list_all_agents_fast()` returns `[]` on any Docker fault, so a roster filtered through it would empty a customer's whole Workspace on a daemon blip — membership must not be a function of Docker health. Fixing `_agent_is_running` to stop reading a Docker read failure as "not running" is the same fail-closed correction.

Good catch on the `sparse=True` trap — under sparse listing `container.name` is `None` and `.labels` raises, so code written against the non-sparse surface fails *silently green* with an empty availability map. Pinning it with real docker-py sparse fixtures rather than mocks of our own wrapper is the only version of that test that would actually catch it.

ACs #4 and #5 answered honestly rather than forced (the admin-only orphan-container branch structurally can't appear in an ownership-keyed roster; L-03 can't fire on a live row). Verified locally: 470 tests green across the full six-PR merged tree.

Issue #2199, opened 2026-08-15
bug: 7 frontend e2e specs fail on stale constants and a missing fixture agent


vybe:
## Summary

The frontend e2e suite is red on `dev`: **15 failed / 93 passed / 11 skipped**. Seven of those failures are **test-only** — stale constants and a missing fixture — and should be fixed so the suite reflects real product state. (The other 8 are genuine product defects, tracked separately.)

## Context

Found during an exploratory E2E sweep of `dev` @ `c1cc22bc`. All 15 failures were root-caused:

| # | Spec(s) | Cause | Product bug? |
|---|---|---|---|
| 7 | `dashboard-type-filter` (all) | hotkey registration race | ✅ yes — tracked separately |
| 1 | `dashboard-stats-overflow` @640 | real control/stat overlap | ✅ yes — tracked separately |
| 2 | `dashboard-grid-view` drag + tidy | **stale spec constant** | ❌ no — this issue |
| 1 | `system-install` preview | **spec strict-mode drift** | ❌ no — this issue |
| 4 | `loops-panel` ×2, `workspace-absorbs-session`, `continue-as-chat` | **missing fixture agent** | ❌ no — this issue |

### 1. Stale layout key (2 specs)

`src/frontend/src/stores/fleetGrid.js` uses `LAYOUT_KEY = 'trinity-grid-layout-v2'`; `e2e/dashboard-grid-view.spec.js` still asserts `'trinity-grid-layout-v1'`, so it reads `null` and both assertions fail.

The bump landed in `309c9ef1` (grid widget chassis, #2042) and the spec was never updated — `git log -S "trinity-grid-layout-v2"` on the spec file returns nothing.

### 2. Strict-mode drift (1 spec)

`e2e/system-install.spec.js` expects a single match for the previewed agent name; the preview now renders it in **4 places** (2 table cells + 2 spans):

```
strict mode violation: ...getByText('<agent>') resolved to 4 elements
```

### 3. Missing fixture agent (4 specs)

All four default to `LOOPS_TEST_AGENT` / `SESSION_TEST_AGENT` = **`testfix`**, which does not exist on a typical instance. `continue-as-chat` fails earlier still, with `task trigger failed: 404`.

Both affordances these specs assert on **do exist** and were verified manually — the Loops tab renders, and `Continue in Workspace → /workspace?agent={name}` is present on the Chat tab. The specs are producing **false failures**.

## Acceptance Criteria

[...]

PR #2220, opened 2026-08-15
fix(dashboard): arm the / hotkey at mount, not behind five fetches (#2200)

### LINKED ISSUE #2200: bug: '/' filter hotkey is silently dropped for ~50ms after the dashboard paints
## Summary

The Dashboard type-to-filter hotkey (`/`, trinity-enterprise#261) is not listening when the fleet first paints. A `/` pressed in that window is silently dropped. User impact is small (~50ms); test impact is total — all **7** `dashboard-type-filter` specs fail deterministically.

## Context

Found during an exploratory E2E sweep of `dev` @ `c1cc22bc`, measured by instrumenting `document.addEventListener`:

```
[data-agent="trinity-system"] visible at   t = 261 ms
handleDashboardKeydown registered at       t = 311 ms
```

The fleet is painted by a **child** component; the hotkey is registered in the **parent's** `onMounted`, which Vue always runs *after* children mount. So there is a window in which the dashboard looks fully interactive but the document-level handler does not exist yet. The event reaches `document` with `defaultPrevented: false` and nothing consumes it.

Probe matrix (fresh browser context per row):

| Scenario | Pill opens? |
|---|---|
| press `/` immediately | ❌ |
| press `/` twice | ❌ then ✅ |
| wait 6s, then press `/` | ✅ |
| click body, then press `/` | ❌ — it is time, not focus |
| click the toolbar `/` button | ✅ (direct `@click`, bypasses the document handler) |

The feature itself is correct once armed: it filters by slug **and** display label, reports "1 of 7 match", and Esc restores.

**Why it matters beyond the 50ms:** Playwright acts the instant its `waitFor` resolves, so the suite hits this window every run. All 7 specs in `e2e/dashboard-type-filter.spec.js` fail on `getByTestId('filter-pill')` not found. The spec's own docstring warns about a *focus* race one level lower down; the actual race is the listener registration.

## Acceptance Criteria

- [ ] Pressing `/` immediately after the fleet renders opens the filter pill
- [ ] The document keydown handler is armed no later than the point at which agent tiles/rows are visible
- [ ] All 7 specs in `e2e/dashboard-type-filter.spec.js` pass without an artificial `waitForTimeout`
- [ ] Existing guards preserved: `/` in an editable field still types a literal slash; Esc backstop still clears; nothing persists across reload

## Technical Notes

- `src/frontend/src/views/Dashboard.vue` — `handleDashboardKeydown` registered in `onMounted` (~line 868); guards 0–5 are correct and not the problem
- Options: register the listener earlier in setup rather than `onMounted`, or expose a readiness signal the tests (and the UI) can rely on
- **The root pattern is reusable:** registering a document-level hotkey in a parent's `onMounted` while children paint first will reproduce this anywhere. Worth a note in the design-system contract if the fix is a convention rather than a one-off.
- Related: trinity-enterprise#261

trinity-ability:
Fixes #2200

## Corrected root cause (differs from the issue body)

Not parent/child mount ordering — that gap is sub-millisecond. The listener registration sat **after `await Promise.allSettled([...5 fetches])`** in `Dashboard.vue`'s `onMounted`, so the dead window is `max(five fetches) − fetchAgents()` — **unbounded on a slow instance**, not "~50ms" (sibling #2130 measured ~10s for the same structural shape). The fleet paints as soon as `fetchAgents()` alone resolves (the `agents.length > 0` template gate), while the `/` hotkey waited on the slowest of five. The dashboard rendered as interactive; every `/` in that window was silently dropped.

## The fix

Both listeners (`keydown` + `click`) hoisted to the **very top** of `onMounted` — top, not merely above the await, because the drift that caused this was a slot that silently *became* post-await (PERF-269 introduced the parallel fetch under a registration that predated it). `handleClickOutside` is hoisted too — outside the stated ACs, but leaving it would re-create the exact "inherited slot beside an existing listener" configuration that produced this bug. Guards 0–5 in `handleDashboardKeydown` verified to read only `setup()`-created refs — arming early is safe by construction. `handleDashboardKeydown` itself is byte-identical.

## Recurrence class

**#2130 was the same class, fixed 3 days before this bug shipped, and its guard hardcoded one file path** (`agentDetailDeepLink.spec.js`) — so the class recurred. This PR's vitest guard (`tests/unit/mountListenerOrdering.spec.js`) is **repo-wide**: every `.vue` + `.js` under `src/frontend/src`, with one justified allowlist entry (`AgentWorkspace.vue` resize handler — self-correcting on the next resize, no input dropped) and `MAX_ALLOWLIST=1` direction-pinned (may shrink, never grow without a deliberate reviewed edit). Honest limitation, documented in-file: it keys on `addEventListener` and has two rare parser blind spots (unbalanced `}` or `//` inside a string literal in a hook body) — the AST-rule follow-up closes them for free.

## Tests

- **Forced-race e2e** (`e2e/dashboard-hotkey-readiness.spec.js`): `page.route` stalls `**/api/system-views` 2000ms; assertions run within 1200ms — a late listener **structurally cannot pass**. Fails 100% unfixed, passes 100% fixed. **Kept `@smoke`**: deterministic by construction, and it gives the smoke tier its first hotkey coverage that actually exercises the race.
- **Existing spec untouched — zero diff on `dashboard-type-filter.spec.js` by design** (a spec edit would destroy the evidence that the product, not the test, was wrong): pre-fix **32 failed / 8 passed** × `--repeat-each=5` → post-fix **40/40**. All e2e runs on a branch-serving Vite server, not localhost.
- **Vitest guard**: red-before/green-after; 6/6 post-rebase; full frontend suite 582/582.

**Evidence-honesty note**: the gate is **8/8** — the existing spec has 8 tests, all pressing `/`; the issue's "7" was a transcription of one race run. A green advisory `frontend-e2e` is corroboration, never proof: it ran green **12/12 over this live bug** (`retries: 2`, smoke tier covers 1-of-8, the least-exposed test).

## Docs

Interaction-readiness clause added to **Principle 23** in `design-system-contract.md` + the `design-system.md` §7 mirror (no renumbering; CLAUDE.md untouched): handlers are armed at mount, above every `await`, never behind fetched data — a surface that renders as interactive must already be interactive (the input-side twin of #15's honest-state rule).

- General-class AST/ESLint rule — would have caught **both** #2130 and #2200; the repo currently has no lint toolchain (no dep, no script, no workflow).
- `AgentWorkspace.vue:838-845` — same shape, lower stakes (resize handler; allowlisted with reason).
- Whether `connectWebSocket()` + the three pollers sit post-await deliberately or by the same drift.
- test-runner catalog drift — claims 52 tests / 3 missing files; actual is 30 files / 582 tests.

## Coordination

Zero file overlap with #2199's PR (verified file-by-file); independently mergeable in either order.

### DIFF
--- a/docs/memory/design-system-contract.md
+++ b/docs/memory/design-system-contract.md
@@ -75,7 +75,7 @@ Honest state:
 18. Verbs acknowledge in ~100ms and confirm completion; toasts for completed verbs, never errors. A failed verb surfaces an `InlineError` next to its control and persists until dismissed — never `console.error` alone, never `alert()` (#1926).
 19. Destructive actions: named verb, restated consequence, safe action focused first.
 22. Times are honest: relative for recency, absolute + timezone on hover/detail.
-23. Keyboard baseline: visible focus, Esc closes overlays, modal focus trap, tab order = visual order.
+23. Keyboard baseline: visible focus, Esc closes overlays, modal focus trap, tab order = visual order — and handlers are armed at mount, above every `await`, never behind fetched data. A surface that renders as interactive must already be interactive: the input-side twin of #15 (#2200).
 24. Encode identity in shape/icon as well as color — never hue alone.
 25. Errors say what happened, what it means, what to do — user vocabulary; codes/traces behind a disclosure.
 26. Set expectations: non-instant actions say what happens, how long, and where the result appears.
--- a/docs/memory/design-system.md
+++ b/docs/memory/design-system.md
@@ -356,7 +356,7 @@ The behavioral half of the standard (the visual half is §1–6). Confirmed in t
 20. Color via semantic tokens only; shared primitives; both themes first-class (#1430).
 21. Frontend invariants hold: domain-scoped stores, a single API client; loading flags live in stores, not components.
 22. Times carry honest context — relative for recency, absolute with timezone on hover or detail (see `docs/TIMEZONE_HANDLING.md`).
-23. Keyboard baseline: visible focus everywhere, Esc closes overlays, modals trap focus, tab order matches visual order.
+23. Keyboard baseline: visible focus everywhere, Esc closes overlays, modals trap focus, tab order matches visual order. Handlers are armed at mount, above every `await` — never behind fetched data. A surface that renders as interactive must already be interactive; this is the input-side twin of #15's honest-state rule for output. (#2200: the Dashboard `/` hotkey registered after `await Promise.allSettled([...5 fetches])` while the fleet painted on `fetchAgents()` alone, so keystrokes were silently dropped for an unbounded window. Guarded by `src/frontend/tests/unit/mountListenerOrdering.spec.js`.)
 24. Identity is encoded in form as well as color — classes and states are distinguishable by shape or icon, never by hue alone.
 25. Meaningful errors everywhere: every failure surface says what happened, what it means, and what to do — in user vocabulary; stack traces and HTTP codes go behind a details disclosure, never the headline.
 26. Expectation setting: non-instant actions state what will happen and roughly how long; multi-step operations show staged progress; say where the result will appear.
--- a/src/frontend/e2e/dashboard-hotkey-readiness.spec.js
+++ b/src/frontend/e2e/dashboard-hotkey-readiness.spec.js
@@ -1,0 +1,94 @@
+import { test, expect } from '@playwright/test'
+
+/**
+ * Dashboard `/` hotkey interaction readiness (#2200).
+ *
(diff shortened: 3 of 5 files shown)


vybe (approved):
Validated via /validate-pr: security scans clean, all CI green (incl. all 3 head seeds + regression diff), `Fixes #2200` closing keyword present, docs updated (design-system-contract Principle 23 + design-system.md §7 mirror). Merges clean against dev; full six-PR sequence simulated locally — 470 tests pass on the merged tree. Approving as part of the #2220 → #2221 → #2226 → #2225 → #2219 → #2222 sequence.

Issue #2210, opened 2026-08-15
bug(e2e): circuit-breaker-badge specs mock an endpoint the agent page never calls — one fails, one passes vacuously


vybe:
## Summary

`e2e/circuit-breaker-badge.spec.js` mocks **a standalone endpoint the agent page never calls**. One of its two tests fails permanently; the other passes **vacuously**. Neither asserts anything about the badge.

This is a **test-only** defect — the product behaviour is correct. It was initially mis-diagnosed as a product bug (the `data-testid` exists at `AgentHeader.vue:139`, so "the element should render"); an empirical check refuted that.

## Evidence

```
$ E2E_BASE_URL=http://localhost ADMIN_PASSWORD=… npx playwright test circuit-breaker-badge --reporter=list
  ✓ circuit-breaker-badge.spec.js:66 › badge is absent when the breaker is closed (386ms)
  ✘ circuit-breaker-badge.spec.js:49 › @smoke renders distinct danger badge when dispatch breaker is open (15.6s)
    Error: expect(locator).toBeVisible() failed
    Locator: getByTestId('circuit-open-badge')
    Error: element(s) not found
  1 failed, 2 passed
```

## Root cause

The badge reads `agent.circuit_breaker_state` (`AgentHeader.vue:138`), which `stores/agents.js:110-117` derives from a `circuit_breaker` block **embedded in the `GET /api/agents/{name}` response** — its own comment says *"no second round-trip"* (the #526 design).

The spec mocks `**/api/agents/*/circuit-breaker`, the **standalone** endpoint. Only `ReliabilityPanel.vue` calls that, for the Settings toggle — the agent page never does.

Two diagnostics confirmed it:

```
DIAG 1: /circuit-breaker requests during agent-page load: 0
        → the route mock can never fire, so it cannot influence the badge
DIAG 2: injecting circuit_breaker:{dispatch:{state:'open'},open:true} into the
        AGENT response renders getByTestId('circuit-open-badge') immediately
```

So the render condition is **fully reachable and correct** — it is simply never met by this spec. Live API confirms the benign real state: `circuit_breaker` present but `null` (dispatch breaking off fleet-wide) → `cb?.dispatch?.state || 'closed'` → badge correctly hidden.

## Both tests are inert, not just the failing one

The sibling *"badge is absent when the breaker is closed"* **passes vacuously** — its mock never fires either, so it asserts the absence of something nothing was ever asked to render. It would keep passing if the badge were deleted outright.

[...]

Issue #2197, opened 2026-08-15
bug: page body scrolls horizontally on Settings and Agent Detail below 1024px


vybe:
## Summary

Two pages break the design-system rule that the page body must never scroll horizontally (wide content scrolls inside its own container). Both tab strips overflow the viewport at tablet and phone widths. Separately, at 640px the Dashboard grid toolbar **overlaps** the stats cluster.

## Context

Found during an exploratory E2E sweep of `dev` @ `c1cc22bc`, measured with `document.documentElement.scrollWidth - clientWidth`.

### Horizontal body overflow

| Viewport | Page | Overflow | Worst offender |
|---|---|---|---|
| 375×812 | `/settings` | **+563px** | `BUTTON :: "Activation"` (tab strip) |
| 375×812 | `/agents/{name}` | **+517px** | `BUTTON.ml-auto :: "More"` |
| 768×1024 | `/settings` | **+178px** | tab strip |
| 768×1024 | `/agents/{name}` | **+132px** | `NAV.-mb-px.flex :: "Overview Tasks Chat Repor…"` |
| ≥1024px | all pages | 0 | — |

Dashboard (all three modes), Library, Operations and Workspace are **clean at every size** — so this is specific to these two tab strips.

Note that Agent Detail already uses `OverflowTabs` (#1114), which is meant to collapse overflow into a "More ▾" menu. At 375px the overflowing element **is** the "More" button, so the collapse is not preventing body overflow at narrow widths.

### Toolbar/stats overlap at 640px (grid mode)

Reproduced independently via `getBoundingClientRect()`:

```
viewport 640 · grid mode
stat "agents"  x  94.5 → 132.9   (y 79)
button "Tags"  x 103.4 → 178.6   (y 77)
→ ~30px horizontal overlap in the same vertical band
```

`Create Agent` also ends at x=95.4 against a stat starting at x=94.5 — touching.

This is exactly what `e2e/dashboard-stats-overflow.spec.js` (#1830) exists to catch; the spec measured 9px at its own layout and currently **fails** on `dev`.

## Acceptance Criteria

- [ ] `document.documentElement.scrollWidth === clientWidth` on `/settings` and `/agents/{name}` at 375px and 768px
- [ ] Tab strips scroll inside their own `overflow-x: auto` container, or collapse, rather than widening the body
- [ ] `OverflowTabs` collapses correctly at 375px — the "More" trigger itself must fit
[...]

PR #2221, opened 2026-08-15
test(e2e): derive the grid layout key, make missing fixtures skip honestly (#2199)

### LINKED ISSUE #1830: bug: Dashboard stats bar — right controls overlap host telemetry below ~1170px effective width

### LINKED ISSUE #2199: bug: 7 frontend e2e specs fail on stale constants and a missing fixture agent

trinity-ability:
Fixes #2199

## Summary

6 failing tests fixed across 11 files.

- **Grid layout key derived, not hand-copied**: the layout localStorage key is now imported from a new zero-dependency `src/utils/gridStorageKeys.js` — the store imports it via the `@` alias, the Playwright specs via a plain relative path (Playwright resolves neither `vite.config.js` nor a tsconfig `paths` map, hence zero imports by documented design). A vitest source guard (`tests/unit/gridStorageKeys.spec.js`, 5 tests) pins store-imports, values, and the zero-import property so the next key bump propagates instead of silently desyncing the specs again (#2042's v1→v2 bump is how this broke).
- **Missing fixture agents skip honestly** via ONE shared **authenticated** probe (`e2e/helpers/agent-probe.js`) with a two-sided contract: a definitive **404 → skip** with an honest reason; **401/403/5xx/transport → THROW** with the real diagnosis. The anti-pattern this PR eradicates — an unauthenticated probe whose 401 read as "agent not found" — produced silent skips with false reasons on every run.

## Scope corrections vs the issue

- `system-install` was **NOT** already fixed upstream — `0f36de8f` (#2125) was necessary but insufficient once ent#384 wrapped the Library in tab `<section>`s. Fixed here with an `ancestor::section[1]` anchor (positional `.last()` rejected — it's how this broke twice), with red-before/green-after AND a non-vacuity probe.
- Also included: the 5th `testfix` consumer the issue missed (`portal-agent-page-overview` — didn't exist at the sweep's commit) and the 6th stale v1-key holder (`grid-org-overlay` cleanup — silently cleared nothing).

## Found in passing, fixed

A 3rd instance of the unauthenticated-probe anti-pattern in `honest-failed-states`: 401 → `[]` → skip claiming "has no schedules" — a false diagnosis.

## Evidence

Full-suite: **95 → 97 passed / 8 failed / 19 skipped**. All 8 residual failures map 1:1 to open issues:

- 7× `dashboard-type-filter` → #2200 (PR in flight from this same run)
- 1× `dashboard-stats-overflow` → #2197 (its "Toolbar/stats overlap at 640px" half)

Five-run verification protocol, including proof the skip guards aren't permanently skipping: **9 passed / 0 skipped** with `*_TEST_AGENT=acme-sage`.

## Note for #2197's owner

`dashboard-stats-overflow.spec.js:4,:100` cites the CLOSED #1830 — anyone tracing the failure lands on a closed issue; the one-line citation fix belongs in #2197.

## Cross-refs

- #2210 — the circuit-breaker spec defect (mocks the standalone `/circuit-breaker` endpoint the agent page never calls; one test fails, one passes vacuously). The README known-gap note now points at it; `circuit-breaker-badge.spec.js` is left **byte-identical** here.

## CI honesty note

None of the 6 fixed tests are `@smoke`; `frontend-e2e` is smoke-only + advisory (#1526) with `retries: 2` — the evidence above is local full-suite runs on a branch-serving dev server, not CI.

- `grid-org-overlay` zone assertions are strict-mode-ambiguous on any stack with pre-existing `dept-*` tags (two independent base reproductions)
- `loops-panel:93` has a 90s assertion timeout inside a 30s test timeout (unreachable by construction)
- convert `circuit-breaker-badge` probes per #2210 then `test.skip(true, 'blocked by #2210')`

### DIFF
--- a/src/frontend/e2e/README.md
+++ b/src/frontend/e2e/README.md
@@ -52,13 +52,67 @@ test('@interactive create agent end-to-end', async ({ page }) => { ... })
 
 ## CI
 
-CI runs e2e only on PRs **labeled `ui`** — add the label to any frontend PR
-that should be exercised end-to-end. The workflow lives at
-`.github/workflows/frontend-e2e.yml` and stands up the full Trinity stack
-before running tests (~5 min total).
+The workflow lives at `.github/workflows/frontend-e2e.yml` and stands up the
+full Trinity stack before running tests (~5 min total). Since #1526 the `ui`
+label is **no longer the gate** — the suite runs on:
 
-To make e2e a required check on a PR, add the `ui` label and wait for the
-workflow to complete.
+1. **Nightly** on `dev` (07:00 UTC) — a red night opens/updates a tracking issue.
+2. **Any PR touching `src/frontend/**`** — automatically, via the `changes` job.
+3. **The `ui` label** — still a manual opt-in on NON-frontend PRs, or to force a run.
+4. **`workflow_dispatch`** — on demand.
+
+Two things follow, and both matter when you are judging whether a change is
+covered:
+
+- **CI runs `npm run test:e2e:smoke` only.** `@visual` and `@interactive`
+  specs never run there, so a green CI says nothing about them.
+- **The workflow is advisory, not a required merge gate** (an explicit decision
+  in #1526 — promoting it needs a flake budget first, #596). `dev`'s required
+  checks are `Analyze (python)`, `Analyze (javascript-typescript)`,
+  `schema-parity`, and `verify-non-root`.
+
+So a claim about the **full** suite has to be evidenced by a local run against
+a live stack, not by a CI badge.
+
+## Fixture agents
+
+Some specs need a real agent on the stack. Each reads an env var with a
+`testfix` default — an agent that does **not** exist on a typical instance:
+
+| Spec | Env var | Needs |
+|---|---|---|
+| `loops-panel` | `LOOPS_TEST_AGENT` | exists, running, **Claude runtime** |
+| `continue-as-chat` | `SESSION_TEST_AGENT` | exists, running, **Claude runtime** |
+| `workspace-absorbs-session` (`@interactive` only) | `SESSION_TEST_AGENT` | exists |
+| `portal-agent-page-overview` | `PORTAL_TEST_AGENT` | exists, visible to admin |
+| `timeline-cancelled-bar`, `honest-failed-states` | `TEST_AGENT` | exists |
+| `circuit-breaker-badge` | `TEST_AGENT` | exists |
+
+**The contract (#2199): a missing fixture reads as SKIPPED, never broken —
+and the probe must be authenticated.** `GET /api/agents/{name}` is behind
+`AuthorizedAgent` and Trinity's JWT lives in localStorage, not a cookie, so an
+unauthenticated probe 401s and the test skips on *every* run: silent false
+confidence, which is worse than a red test. Use the shared
+`e2e/helpers/agent-probe.js` — do not hand-roll a probe.
+
+> **Known gap:** `circuit-breaker-badge.spec.js` still carries the legacy
+> unauthenticated probe, so both its tests skip on every run. It was left as-is
+> deliberately: with an authenticated probe its `@smoke` test **fails** (the
+> `circuit-open-badge` element never appears, though the `data-testid` is still
+> present in `AgentHeader.vue`), which looks like a product defect. Un-skipping
+> it would turn the nightly signal red for a cause that belongs in its own
+> issue. Convert it to an explicit `test.skip(true, 'blocked by #NNNN')` once
+> that issue exists.
+
+The probe checks **existence only**. If a spec needs the agent running or on a
+particular runtime, say so in the spec header: reporting a *present but
+unsuitable* fixture as "absent" would be a false diagnosis.
+
+```bash
+# point the fixture-dependent specs at a real agent
+LOOPS_TEST_AGENT=my-agent SESSION_TEST_AGENT=my-agent PORTAL_TEST_AGENT=my-agent \
+  ADMIN_PASSWORD=<pw> npm run test:e2e
+```
 
 ## Adding tests
 
--- a/src/frontend/e2e/README.md
+++ b/src/frontend/e2e/README.md
@@ -92,7 +92,11 @@ Some specs need a real agent on the stack. Each reads an env var with a
 and the probe must be authenticated.** `GET /api/agents/{name}` is behind
 `AuthorizedAgent` and Trinity's JWT lives in localStorage, not a cookie, so an
 unauthenticated probe 401s and the test skips on *every* run: silent false
-confidence, which is worse than a red test. Use the shared
+confidence, which is worse than a red test. The inverse holds too: **only a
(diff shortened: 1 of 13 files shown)


vybe (approved):
Validated via /validate-pr: security scans clean, all CI green (3 head seeds + regression diff), `Fixes #2199` closing keyword present. Test-only + a new zero-dependency `gridStorageKeys.js` constant module with a vitest source guard pinning the store-import and zero-import properties — the right shape to stop the #2042 key-bump desync recurring. Merges clean against dev; six-PR sequence simulated locally (470 tests green on the merged tree).

PR #2222, opened 2026-08-15
fix(perf): Agent Detail + Workspace stop over-fetching — dedupe, batch sessions, close the portal agent-key hole (#2198)

### LINKED ISSUE #2198: bug: Agent Detail and Workspace over-fetch on load (21 redundant requests, N+1 roster)

trinity-ability:
Fixes #2198

## Summary

Agent Detail and the Workspace issued 21+ redundant requests on load. This PR takes **all 8 duplicate classes to exactly 1 request each**, replaces the Workspace sidebar's per-agent session fan-out with **one viewer-scoped batch call** (`GET /api/enterprise/client-portal/sessions`), and kills the dashboard 9s retry ladder at its root with a backend `settled` flag. `/activity` stays ×4 by design (per the issue's AC — four distinct windows).

| Duplicate class | Before | After |
|---|---|---|
| `/api/agents/{name}/agent-dashboard` | 9 (retry ladder ×3 callers) | **1** |
| `/api/agents/{name}/agent-dashboard/exists` | 3 | **1** |
| `/api/agents/{name}/info` | 3 | **1** |
| `/api/agents` | 2 | **1** |
| `/api/agents/{name}/emotions` | 2 | **1** |
| `/api/agents/{name}/connector/playbooks` | 2 | **1** |
| `/api/settings/feature-flags` | 2 (once per store) | **1** |
| `/api/users/me` | 2 (NavBar duplicate) | **1** |

Workspace: the sidebar previously fanned out one `/sessions` request **per roster agent × six `refreshThreads` call sites** (N = fleet size). It now makes **one** batch call returning the viewer's whole thread list, with a per-agent fallback if the batch fails (`sessionsFailed` flag) and a TRANSITIONAL 404 → fan-out fallback for deploy skew. The `settled` flag (backend-stamped, spoof-proof — overwritten server-side immediately after `response.json()`, and every inconclusive path returns backend-built dicts) means a dashboard probe that already answered conclusively is never re-probed, ending the `[0, 3000, 6000]` ladder on first load too.

## Security: the portal agent-key hole (E7) — closed at the dependency

**This PR closes a PRE-EXISTING hole its own route would have amplified.** Code trace established the hole: `get_portal_principal`'s platform branch → `get_current_user` → an agent-scoped key resolves to its owner **carrying the owner's role**, unfiltered (the ent#293/#297 trap) — so any agent's injected `TRINITY_MCP_API_KEY` reached **all 30 portal routes** as `is_platform=True`. Decisive fact in the grant-vs-use analysis: the portal is a REST path **around the MCP layer's agent-permission matrix** — an agent key could read the owner's conversations with agents it holds **no `agent_permissions` edge to**, access the MCP surface explicitly denies (`get_chat_history` is `{self} ∪ permitted`). The batch route added here amplified exfiltration from N calls against N *discovered* names to ONE call returning the owner's whole cross-agent thread index — which is why this PR had to close it rather than ship the amplifier silently.

Fix: `reject_agent_principal(user)` in the platform branch, **at the dependency** so every current and future portal route inherits it. Deliberately NOT `reject_non_interactive_principal` (a user's own user-scoped key scripting their own Workspace is legitimate; `scope='system'` keeps breadth by design; connector/portal_delegate are already fenced upstream in `get_current_user`). **Zero blast radius: no legitimate agent caller exists** — no MCP tool targets the portal, and neither agent images nor the base image call it (verified by grep). 3 new tests (agent key → 403 at the boundary; user key and JWT still pass as platform; portal session tokens never touch the fence); 306 portal-adjacent unit tests green; the fence broke zero of 10,338 in the full unit suite.

## Invariant #13 (MCP third surface)

No MCP tool is added for the batch route, deliberately: no portal route has one, an MCP key cannot hold a `PortalPrincipal`, and agent principals are now rejected outright at the portal boundary — a tool would be advertising a surface its own callers are fenced from.

## Corrected diagnosis

The issue body's "3× per poll cycle" was wrong — there is no timer; it was a retry **ladder** (`[0, 3000, 6000]`) fired by three independent callers. Root causes per class were corrected before implementation: https://github.com/Abilityai/trinity/issues/2198#issuecomment-5301899739

## AC edits

- **AC #2** reworded: the dashboard is *probed once per page load; the ladder is bounded at most once* (the `settled` flag makes the first conclusive answer terminal).
- **AC #4**'s "lazy per-agent fetch" alternative is struck: the sidebar renders a merged cross-agent recency list, so only a batch endpoint can serve it — a lazy per-agent fetch would have to fetch every agent anyway before it could sort.

## Measurements (origin-attributed)

Baseline measured at `58df9aed` on a **branch-serving dev server** for this worktree: **64 requests on mount / 27 on revisit** → post-dedupe **48 / 22** → with the `settled` flag **47 on mount** (`/agent-dashboard` = 1). Honest limits: the batch endpoint could **not** be exercised end-to-end locally — the live stack serves the main checkout — so it is verified by pytest (22 tests incl. the roster-equivalence guard) and a route probe instead; the live stack usefully proved the deploy-skew fallback (404 → fan-out → Workspace renders). Note the batch does **not** make the Workspace faster — the roster briefing fan-out still dominates load time (#2163); this PR removes request volume, not the critical path.

## C3 / KeepAlive record

Investigated per instruction; the removal premise was **falsified empirically**: removing `AgentDetail` from KeepAlive deletes only 2 of the 8 duplicate classes (not 6), costs ~19 requests per revisit, two shipped regression specs state the caching as a load-bearing premise, and un-caching kills in-flight chat SSE streams + voice sessions. KeepAlive stays; the dead `'SystemAgent'` include entry (matches no component; `/system-agent` is a redirect) is removed and the stale comment corrected.

## Deviations from plan (3, verified sound in review)

1. **`settled` flag instead of the planned `/exists` tri-state** — judged better: 1+1 requests at t≈0 vs 1+3 over 9s, no DB change, correct on every deployed image.
2. **T4 implemented with `once()`** rather than the planned shape.
3. **Revisit double-probe left in place** — explicitly scoped out (plan §7).

## Known gap (deliberate)

`sessionsFailed` has **no UI consumer yet** — the plan wanted an honest degraded-state banner; the one-liner is deferred past the #2196 rebase (design-contract pass needed). Listed as a follow-up below.

- `search_chats` missing `include_owned` — one-line fix now available via `roster_agent_names`; full issue text in the review report.
- A→B agent-switch leaves stale `hasBrainOrb`/emotions state (drift on switch without remount).
- Sessions cap + starred-pinning: product question on the batch list's bound.
- `sessionsFailed` honest-banner UI consumer.
- Remove the TRANSITIONAL 404 → fan-out fallback after one release.
- Enterprise-submodule confirmation that the E7 fence now covering room routes is intended (it is — rooms are human viewer surfaces — but enterprise CI should confirm).

## Merge order

**#2196 merges FIRST; this PR rebases after.** Shared file: `client_portal/service.py` (expected import/context-level conflicts only — this branch's `roster_agent_names` extraction deliberately avoids #2196's `get_roster`/`get_agent_card` functions). The roster-equivalence test (`tests/unit/test_2198_portal_batch_sessions.py`) is the post-rebase drift guard.


vybe (approved):
Validated via /validate-pr: security scans clean (`ALICE@EXAMPLE.COM`/`Owner@Example.com` are example.com placeholders), `Fixes #2198` closing keyword present, docs updated (architecture.md, requirements/core-agent.md, feature flows).

**The security fix is the headline and it is placed correctly.** `get_portal_principal`'s platform branch now runs `reject_agent_principal` — **at the dependency**, so all 30 portal routes inherit it rather than each remembering. This closed a pre-existing hole (an agent-scoped key resolves to its owner *carrying the owner's role* — the ent#293/#297 trap) that the new batch route would have amplified from N calls against N discovered names to one call returning the owner's whole cross-agent thread index. The grant-vs-use reasoning is right: the portal is a REST path *around* the MCP agent-permission matrix. Choosing `reject_agent_principal` over `reject_non_interactive_principal` is also correct — a user's own user-scoped key scripting their Workspace is legitimate.

Invariant #13 correctly answered as "no MCP tool, deliberately" — adding one would advertise a surface its own callers are now fenced from.

Also good: the C3/KeepAlive removal premise was **falsified empirically** rather than assumed, and the corrected diagnosis (a retry ladder from three callers, not a timer) was posted on the issue before implementation. The `settled` flag being backend-stamped after `response.json()` makes it spoof-proof.

Conflict resolution (docs-only): merged `origin/dev` and resolved one add/add `architecture.md` conflict against #2196 (just landed) — both paragraphs kept, since #2196's text already references "#2198's batch-sessions gate" (written to coexist). Verified the auto-merged `client_portal/service.py` + `models.py` retain both #2196's availability projection and this PR's batch route + auth fence; 251 portal tests green on the merged tree.

Issue #2217, opened 2026-08-15
bug: canary harness left disabled on a retired constraint, and nothing reports whether it is running


vybe:
## Summary

The canary invariant harness was disabled on the dev instance (`CANARY_ENABLED` 1 → 0) on the grounds that its snapshot reader is SQLite-only, when the instance migrated to PostgreSQL. **That constraint no longer exists** — #1540 finished migrating every SQL-tier collector to the `get_engine()`/`DATABASE_URL` seam, and #300 is closed.

Two things to fix: the harness should be re-enabled where it was switched off, and — the part that actually matters — **nothing anywhere reports whether the canary is running.**

## The stale premise

`src/backend/canary/snapshot.py:50-62` documents the current state explicitly:

> Backend seam (#300 / #1450 / #1540): all SQL-tier collector reads route through the `get_engine()`/DATABASE_URL seam (`db/engine.py` + `db/tables.py`), so the harness reads the live configured backend — SQLite OR PostgreSQL — not a stale `/data/trinity.db`. […] The collector no longer imports `db.connection`.

Confirmed: the `canary/` package contains **zero** `sqlite3` imports. #300 (`Configurable database backend: SQLAlchemy Core abstraction`) is **closed**. The pre-#1540 behaviour — where `_collect_known_agents` returned zero rows on PostgreSQL and S-01/E-05/B-02/E-04/G-04/B-01 all went **vacuously green** — is exactly the failure mode #1540 fixed.

So the harness is backend-agnostic today, and turning it off on PostgreSQL was unnecessary.

## The real gap: a disabled canary is indistinguishable from a healthy one

The canary exists to notice that something has gone quiet. It has no signal of its own for the case where **the canary itself is not running** — H-01 covers collector *blindness* while a cycle is executing, but nothing covers "no cycle is executing at all."

Consequences, all of which held on the affected instance:

- `CANARY_ENABLED=0` produces no violations, which is byte-identical to a clean fleet
- No surface reports last-cycle time, so staleness is invisible
- The harness can be switched off during an incident and silently never switched back on

This is the same class as #2204 (39 consecutive red deploys with no alert) and the same class the harness's own H-01 was written for (`"Ran and was fine" and "never ran" are the same bytes in a failure list`) — applied one level up, to the detector itself.

## Acceptance criteria

- [ ] `CANARY_ENABLED` restored on the dev instance; harness confirmed cycling against PostgreSQL
[...]

Issue #2216, opened 2026-08-15
bug: shipped database backup tooling is never invoked — no scheduled backups, no recovery point


vybe:
## Summary

Trinity ships both `scripts/deploy/backup-database.sh` and `scripts/deploy/restore-database.sh`, but **nothing ever invokes them**. There is no cron entry, no compose service, no workflow, and no in-process job. The only guidance is "run it before major changes" — a manual step, explicitly marked **Manual** in the process catalogue.

The result: an instance's entire database is one stray write away from unrecoverable loss, and the tooling that would have prevented it is already in the repo, unused.

## Context

This was not hypothetical. An instance database was destroyed by a single non-truncating write at offset 0 — an HTTP response body written over the SQLite header:

```
00000000: 4854 5450 2f31 2e31 2032 3030 204f 4b0d  HTTP/1.1 200 OK.
00000010: 0a64 6174 653a 204d 6f6e 2c20 3130 2041  .date: Mon, 10 A
          ... server: uvicorn / content-length: 13 ...
```

Every worker then died at import with `sqlite3.DatabaseError: file is not a database`, uvicorn never bound its port, and the healthcheck failed 893 consecutive times. Because the other services declare `depends_on: backend: condition: service_healthy`, the whole stack sat in `Created` behind a gate that could never open.

The file was *recoverable* — `sqlite3 .recover` returned ~1.37M rows and all 73 `CREATE TABLE` / 161 `CREATE INDEX` statements survived in overflow pages — but only because someone thought to try. There was **no backup to restore from**, so the instance was rebuilt empty and all prior data was written off by decision.

Worth stating plainly: a codebase scan found **no Trinity code path capable of this write**. There are zero update-mode file opens (`r+b`, `rb+`, `w+`, `a+`) and no `conv=notrunc` anywhere under `src/` or `docker/base-image/`. The leading hypothesis is an operator action during hands-on debugging. **That is precisely the point** — the backup gap is what turns an ordinary human slip into total data loss, and no amount of code hardening addresses it.

## Impact

Applies to every self-hosted install, which is the whole OSS audience. `CLAUDE.md` tells operators their data lives in a single SQLite file at `~/trinity-data/trinity.db`; the platform provides backup tooling and then never runs it. An operator who never reads the process docs has no recovery point at all, and will not discover this until the day they need one.

PostgreSQL installs are not exempt — `pg_dump` is likewise documented and likewise unscheduled.

## Acceptance criteria

[...]

Issue #2215, opened 2026-08-15
bug: fresh-install seeding races itself across workers — two seeders allocate the same agent SSH port


vybe:
## Summary

On a **fresh install with `--workers 2`**, the two first-run seeders race each other and allocate the **same agent SSH port**. The losing container fails to bind, agent creation fails, and the failure is logged rather than raised — so the deploy reports success with a missing agent.

Observed on a fresh PostgreSQL instance:

```
Failed to create agent acme-scout: ... Bind for 0.0.0.0:2227 failed: port is already allocated
ent#313: removed the orphaned container for acme-scout after a failed creation
```

`cornelius` took 2227; the `acme` system seed independently assigned 2227 to `acme-scout`. The second bind lost. The ent#313 reclaim did its job (no leaked container), but the agent was simply never created and nothing surfaced it.

## Root cause

Two independent defects compose:

**1. The two seeders hold different locks.** Both run at boot in **every** worker:

- `services/cornelius_agent_service.py:72` — `_PROVISION_LOCK_KEY = "cornelius:provision"`
- `services/system_seed_service.py:74` — `_PROVISION_LOCK_KEY = "system_seed:provision"`

Each SETNX lock serialises its own seeder against itself across workers — the logs confirm both guards firing (`"Another worker is provisioning Cornelius"` / `"Another worker is seeding the default system"`). Neither serialises **cornelius against system_seed**, so the two run concurrently by design.

**2. Port allocation is an unlocked check-then-act.** `services/docker_service.py:260` `get_next_available_port()`:

```python
agents = list_all_agents_fast()
existing_ports = set(a.port for a in agents if a.port)
start_port = max(existing_ports or {2221}) + 1
for port in range(start_port, start_port + 100):
    if port not in existing_ports and is_port_available(port):
        return port
```

It reads live container state, computes a port, and **returns** — the caller creates the container afterwards. Two concurrent callers observe the same `existing_ports`, compute the same `max+1`, and both attempt the bind. `is_port_available()` narrows the window but cannot close it: nothing holds a reservation between the check and `containers.run()`.

## Impact

Hits **every fresh multi-worker install**, which now includes every new PostgreSQL instance. The result is a silently incomplete default fleet — the operator's first impression of a new instance is a system missing one of its agents, with the reason buried in container logs.

The allocator is shared by all agent creation, not just seeding, so any two concurrent creations can collide; first-run seeding is simply the case that reliably triggers it.

## Acceptance criteria

[...]

PR #2223, opened 2026-08-15
fix(ci): dev deploy serialises, self-heals, and can no longer fail silently (#2204)

### LINKED ISSUE #2204: bug(ci): Deploy to Dev has failed on every run since 2026-08-10 — 39 consecutive silent failures, dev instance 5 days stale
> **⚠️ Corrected 2026-08-15.** The original diagnosis below (a self-perpetuating
> orphaned container) was **wrong**. Host inspection found no orphan at all. The
> real blocker was a **corrupted database**. The original text is preserved at the
> bottom for the record; read the corrected section first.

## Summary

`Deploy to Dev` failed on **every run from 2026-08-10T15:50Z to 2026-08-15T18:12Z** — ~5 days, 39+ consecutive failures, no successful deploy in between and no alert of any kind. Every merge to `dev` in that window shipped nothing. CI was green, PRs merged cleanly, issues auto-promoted to `status-in-dev` — and the instance stayed on the 2026-08-10 build.

**Status: resolved.** dev is green and current as of 2026-08-15T18:12Z.

This is the **second occurrence of this class**. The previous one went unnoticed for four days and thirty merges.

## Corrected root cause

Not the orphaned container. `~/trinity-data/trinity.db` had an HTTP response written over its header:

```
00000000: 4854 5450 2f31 2e31 2032 3030 204f 4b0d  HTTP/1.1 200 OK.
00000010: 0a64 6174 653a 204d 6f6e 2c20 3130 2041  .date: Mon, 10 A
          ... server: uvicorn / content-length: 13 ...
```

Every uvicorn worker then died at import:

```
File "/app/db/migrations.py", line 148, in run_all_migrations
    CREATE TABLE IF NOT EXISTS schema_migrations (
sqlite3.DatabaseError: file is not a database
```

uvicorn never bound `:8000` → the healthcheck failed **893 consecutive times** → `trinity-{scheduler,frontend,mcp-server}` sat in `Created` indefinitely, because they declare `depends_on: backend: condition: service_healthy`. The deploy was waiting on a health gate that could never open.

**Timeline:** first deploy failure 2026-08-10T15:50Z → database clobbered 16:33:05Z, ~43 minutes *later*, during hands-on debugging of that first failure. So the name conflict was the **trigger** for the incident, not its cause.

### On the container-name conflict

The conflict was real, not imaginary — it appears verbatim in run `31878949909` immediately before `Process exited with status 1`, so on that run compose exited at the rename, *before* reaching the health gate. But it was a **downstream blocker, not the root cause**: clearing it would have moved the failure four lines down to `curl -sf http://localhost:8000/health`, which could never succeed. Fixing only the orphan would not have restored dev.

Host inspection at remediation time found **no** `<12-hex>_trinity-*` container, and the running backend was the exact container ID the conflict error named as the winner — so the conflict had cleared on its own before anyone acted on it.

### Disk was never a factor

Explicitly measured, because the fix plan depended on it: **79% before, 80% after, with no prune run at any tier.** The 39 `--no-cache` builds had not accumulated meaningfully — 4 tagged Trinity images, ~2.4GB total. A pre-build prune step was planned on this hypothesis and **dropped** once measured.

## What this issue now covers

AC #1 is satisfied (dev restored, manually). The remaining work is **prevention**, and the alerting criterion is the one that matters — it is the only one that would have shortened a 5-day silent outage.

- [x] Dev instance restored and `Deploy to Dev` green on the current `dev` tip
- [ ] Deploy restart step reaps stale compose backup containers, so an interrupted run cannot wedge subsequent ones. **Hygiene, not the fix for this outage** — there was no orphan to reap
- [ ] Deploys are serialised, so multiple merges cannot run `docker compose up -d` against one host concurrently
- [ ] A failing `Deploy to Dev` raises a visible alert, so a red deploy cannot run silently for days a third time
- [ ] The alert asserts **post-deploy state**, not merely the job's conclusion (see below)

### Why conclusion-only alerting is insufficient

The successful remediation run **also failed to create an agent, and still reported success**:

```
Failed to create agent acme-scout: ... Bind for 0.0.0.0:2227 failed: port is already allocated
```

Agent-seeding errors are logged, never raised. So `deploy-dev.yml` can go green with the fleet incomplete. Any alerting added here must assert on observable post-deploy state — the workflow already sets this precedent by grepping boot logs for the enterprise-registration string.

## Spun out

Findings from this investigation that are their own defects:

- **#2215** — fresh-install seeding races itself across workers; two seeders allocate the same agent SSH port (the `acme-scout` failure above)
- **#2216** — shipped database backup tooling is never invoked; no scheduled backups, no recovery point (the destroyed database was recoverable to ~1.37M rows, but there was nothing to restore *from*)
- **#2217** — canary harness left disabled on a retired constraint, and nothing reports whether it is running

## Instance state changes

Recorded because they affect anyone testing against dev:

- dev now runs **PostgreSQL 16.15, not SQLite**; Alembic at head `0038_portal_chat_state`
- The database is **empty** — all prior dev data gone by decision. `email_whitelist` empty (email-OTP login dead until re-whitelisted; admin/password works), `mcp_api_keys` = 0, so **any stored dev MCP API key is now invalid**
- `CANARY_ENABLED` set to 0 — see #2217, the stated reason no longer applies
- There is **no rollback to SQLite** on dev; clearing `DATABASE_URL` returns the backend to the crash-loop

---

<details>
<summary><b>Original diagnosis (superseded — preserved for the record)</b></summary>

The original report attributed the outage to a self-perpetuating orphaned container:

> Compose's recreate renames the outgoing container to a `<short-id>_<name>` backup before starting the replacement. An earlier deploy died between the rename and the cleanup, leaving `ef62a976ed1d_trinity-backend` resident on the host. Every subsequent deploy now tries to claim that same backup name and hits the conflict at the identical point — so the state is **self-perpetuating**: it cannot clear itself.

This was inferred from the run log alone, without host access. It was wrong in its central claim: the state was **not** self-perpetuating, the orphan had already cleared, and removing it would not have restored the instance. The conflict was a genuine but secondary blocker sitting in front of a crash-looping backend.

The proposed immediate fix (`docker rm -f ef62a976ed1d_trinity-backend`) was therefore a no-op — the container did not exist by the time anyone looked.

</details>

vybe:
## Summary

`Deploy to Dev` failed on **every run from 2026-08-10T15:50Z to 2026-08-15T18:12Z** — 39+ consecutive failures, no successful deploy in between, and **no alert of any kind**. CI stayed green and PRs merged cleanly while the dev instance sat on a five-day-old build. Second occurrence of the class; the first ran four days and thirty merges.

dev has since been restored manually (AC #1). This PR is the **prevention** half.

## The issue's diagnosis was wrong, and the record is corrected

#2204 attributed the outage to a self-perpetuating orphaned container. Host inspection found **no orphan at all**. The real blocker was a corrupted database — an HTTP response written over the SQLite header — which crash-looped every uvicorn worker at import, so the healthcheck failed 893 times and `trinity-{scheduler,frontend,mcp-server}` sat in `Created` behind `depends_on: backend: condition: service_healthy`.

The name conflict *was* real (it appears verbatim before `Process exited with status 1` in run `31878949909`), but it was a **downstream blocker, not the cause** — clearing it would have moved the failure four lines down to the health check, which could never pass. The issue body has been rewritten with the corrected postmortem, original diagnosis preserved in a collapsed section.

A corrupted file is not something a workflow can prevent. So this PR targets the properties that decide whether the *next* such outage is noticed in minutes or in days.

## Changes

**Serialise deploys** — measured up to **four** runs executing concurrently against the one dev VM (2026-08-12T12:49:51–12:50:15, overlapping 960–1215s), each running `docker compose build --no-cache` then `up -d` on the same host and project. `cancel-in-progress` is explicitly `false`: cancelling mid-run is exactly what strands a half-recreated container.

**Reap stale compose backups** before `up -d`, so one interrupted run cannot wedge its successors. Both guards are load-bearing under the script's `set -e` — `grep` exits 1 when nothing matches and `docker rm -f` exits 1 with no arguments, so an *unguarded* reaper would abort every **healthy** deploy, which is the common path.

**Assert post-deploy state**, not the job's conclusion. Agent-creation failures are logged and never raised, so a deploy can report success with the fleet incomplete — observed on the remediation run itself, where a seeded agent lost an SSH port race (#2215) and the deploy stayed green. Follows the existing enterprise-registration boot-log grep precedent; the single log read is hoisted so both consumers share it.

**File an issue on failure**, deduped to one open at a time.

## Two deliberate rejections

**Slack webhook → GitHub issue.** #2204 suggested reusing `CANARY_SLACK_WEBHOOK_URL`. That variable lives on the *instance*, not in Actions, and this repo has **no** webhook secret (repo secrets are `DEV_*`, `GCP_*`, `HOMEBREW_TAP_TOKEN`, `TAILSCALE_AUTH_KEY`; zero org secrets). `secrets.SLACK_WEBHOOK_URL` would render empty and no-op silently — the packaging-gap class recorded **six** times in `learnings.md` (#1039, #1056, #1707, #1871, ent#14, ent#31). An alert that silently never fires is worse than none. It is a separate **job**, not a step in the SSH script, because an alert sent from the host shares a failure domain with the thing it reports on.

**Pre-build prune → dropped.** Planned on the hypothesis that 39 `--no-cache` builds had filled the disk. Measured: **79% before, 80% after, no prune run at any tier.** The hypothesis was false, so the step was cut rather than shipped on speculation.

## Test plan

- [x] `pytest tests/unit/test_2204_deploy_self_heal.py -q` → **13 passed**
- [x] Six **meta-tests** mutate the workflow back to its pre-fix content and assert each check goes red (`learnings.md` 2026-08-05: a guard asserted only against the fixed tree documents today's state instead of closing the class)
- [x] Reaper shell logic executed both ways: zero matches survives `set -e` (exit 0); with matches it selects only `<12-hex>_trinity-*` and leaves `trinity-backend` / `agent-test` untouched
- [x] `notify-failure` script `bash -n` clean; heredoc rendered with test env — backticks literal, vars expanded
- [x] YAML parses; both jobs present with expected `if`/`permissions`
- [x] Other workflow-scanning guards still pass (`test_1891`, `test_2089`, `test_canary_env_prod_parity`) → 43 passed
- [ ] **Verified on merge**: the post-merge `Deploy to Dev` run goes green, and the reaper logs "No stale compose backup containers" rather than aborting

## Honest scope

The reaper is **hygiene, not the fix for this outage** — there was no orphan to reap. Both the inline comment and the guard's docstring say so, because recording it as the fix would bury the real cause.

## Spun out

- #2215 — fresh-install seeding races itself across workers; two seeders (different SETNX keys) over an unlocked check-then-act port allocator
- #2216 — shipped database backup tooling is never invoked; no scheduled backups, no recovery point
- #2217 — canary harness left disabled on a retired constraint (#1540 made it backend-agnostic), and nothing reports whether it is running

Fixes #2204

### DIFF
--- a/.github/workflows/deploy-dev.yml
+++ b/.github/workflows/deploy-dev.yml
@@ -8,6 +8,23 @@ on:
 permissions:
   contents: read
 
+# #2204: serialise deploys. Measured up to FOUR runs executing concurrently
+# against the one dev VM (2026-08-12T12:49:51-12:50:15, overlapping 960-1215s),
+# each running `docker compose build --no-cache` then `up -d` on the same host
+# and compose project: concurrent recreates of the same container, and four
+# concurrent full image builds on one box.
+#
+# cancel-in-progress MUST stay false. Cancelling mid-run is precisely what
+# strands a half-recreated container -- compose renames the outgoing container
+# to a <12-hex>_<name> backup and removes it once the replacement is up, so a
+# run killed in between leaves the backup resident (see the reaper below).
+# GitHub supersedes the *pending* run rather than the running one, so this
+# converges on "current deploy finishes, then the newest commit deploys"
+# without ever interrupting a deploy in flight.
+concurrency:
+  group: deploy-dev
+  cancel-in-progress: false
+
 jobs:
   deploy:
     runs-on: ubuntu-latest
@@ -170,6 +187,32 @@ jobs:
               BUILD_DATE="${BUILD_DATE}" \
               docker compose ${COMPOSE_FILES} build --no-cache backend frontend mcp-server scheduler
 
+            echo "=== Reap stale compose backup containers (#2204) ==="
+            # Compose renames the outgoing container to <12-hex>_<name> before
+            # starting its replacement, then removes the backup once the new one
+            # is up. A run that dies in between leaves the backup resident, and
+            # every later deploy hits "container name is already in use" at the
+            # identical point.
+            #
+            # HONEST SCOPE: this is hygiene, NOT the fix for the #2204 outage.
+            # That outage was a corrupted database crash-looping the backend;
+            # host inspection found no orphan at all, and clearing one would
+            # only have moved the failure to the health check below. Do not
+            # record this block as the fix for that incident.
+            #
+            # Both guards are load-bearing under the `set -e` at the top of this
+            # script: `grep` exits 1 when nothing matches, and `docker rm -f`
+            # exits 1 when given no arguments. Unguarded, either would abort
+            # EVERY healthy deploy -- the no-orphan path is the common one.
+            STALE=$(sudo docker ps -a --format '{{.Names}}' | grep -E '^[0-9a-f]{12}_trinity-' || true)
+            if [ -n "$STALE" ]; then
+              echo "Reaping stale compose backup containers:"
+              printf '%s\n' "$STALE"
+              printf '%s\n' "$STALE" | xargs -r sudo docker rm -f
+            else
+              echo "No stale compose backup containers."
+            fi
+
             echo "=== Restart ==="
             sudo docker compose ${COMPOSE_FILES} up -d backend frontend mcp-server scheduler
 
@@ -179,6 +222,27 @@ jobs:
             SCHED=$(sudo docker inspect trinity-scheduler --format='{{.State.Health.Status}}')
             echo "Scheduler: $SCHED"
             [ "$SCHED" = "healthy" ] || echo "WARNING: scheduler not healthy yet"
+
+            echo "=== Backend boot log ==="
+            # ONE read, several consumers (enterprise registration below, and
+            # the #2204 seeding check). Scope it to the CURRENT container run,
+            # so a line left over from a previous boot cannot fail today's
+            # deploy.
+            #
+            # The --tail fallback does NOT carry that property, and is sound
+            # only here: `up -d` just recreated this container, so its whole log
+            # is a few hundred lines and every boot line is inside the window.
+            # On a long-running container it would be wrong in both directions
+            # -- deep enough to catch a previous boot's failure line, and too
+            # shallow to reach the current boot's success line (a live instance
+            # measured 14k lines since start). Don't lift this block somewhere
+            # the container isn't newly started.
+            BOOT_SINCE=$(sudo docker inspect -f '{{.State.StartedAt}}' trinity-backend 2>/dev/null || echo "")
+            if [ -n "$BOOT_SINCE" ]; then
+              BOOT_LOG=$(sudo docker logs trinity-backend --since "$BOOT_SINCE" 2>&1 || true)
+            else
+              BOOT_LOG=$(sudo docker logs trinity-backend --tail 2000 2>&1 || true)
+            fi
 
             echo "=== Enterprise registration (#2068) ==="
             # The enterprise version-line is upgraded during module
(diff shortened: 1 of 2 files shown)


trinity-ability (approved):
Validated via /validate-pr.

Independently reproduced: `test_2204_deploy_self_heal.py` → 13 passed; sibling workflow scanners (`test_1891`, `test_2089`, `test_canary_env_prod_parity`) → 43 passed. Workflow YAML parses with the expected two jobs, `concurrency.cancel-in-progress: false`, and job-scoped `issues: write` on `notify-failure` only. All five labels the notifier passes to `gh issue create` exist in the repo, so the alert job cannot fail on a missing label.

Security scan clean (no keys, secrets, emails, public IPs, symlinks, or host paths). No backend package or env-var packaging surface touched.

One non-blocking note: the seeding check greps `Failed to create agent`, which has two emitters — `system_service.py:1629` (any `deploy_manifest`) and `crud.py:2910` (any agent create) — not just first-run seeding. `BOOT_LOG` is scoped to the current container run, so exposure is limited to the window between container start and the check, and the failure is loud and self-clearing. Worth a tighter anchor later; not a merge blocker.

PR #2225, opened 2026-08-16
fix(#1704): agent plugin manifest — declared, committed, self-healing

### LINKED ISSUE #1704: bug: agent plugin install-state is lost on container recreate (~/.claude.json is gitignored)

AndriiPasternak31:
Fixes #1704

## Summary

Makes an agent's Claude Code marketplace-plugin selection a first-class, **committed, secret-free, self-healing** piece of config — the agent-local half of the incubating global plugin-management model (trinity-enterprise#192).

**Reframe first.** The issue's literal premise ("a recreate loses plugins") is not reproducible — `/home/developer` **is** the durable `agent-{name}-workspace` volume, no recreate removes it, and `startup.sh` preserves untracked files, so a plain recreate keeps both `~/.claude.json` and the `~/.claude/plugins/` cache. The real gap is **git-based reconstitution** onto a fresh/empty volume or a new host (the #1169 move exports `data/` only; the #834/#1581 hard-purge removes the volume), where the gitignored manifest + cache are exactly what a clone drops — a gap #1705 completed by removing the last incidental crutch (the cache used to be auto-committed).

## What it does

- **Declaration + materialization** — a template's `template.yaml plugins:` block is read by the **total** (never-raising) `services/template_plugins.py` (the ent#89 reader shape) into `_TemplateResolution.declared_plugins`, fed by all three resolver branches; `crud._materialize_agent_files` → `git_service.materialize_plugins()` writes nested `~/.trinity/plugins.yaml` via the shared injection-safe heredoc writer (`sort_keys=True` for byte-stability so the 15-min auto-sync loop never re-commits a churning manifest). Opt-in, ghost-skipped, non-fatal inside the creation rollback fence.
- **Committed** — `.trinity/plugins.yaml` joins `_TRINITY_AUTHORED_PATHS`, so it rides the #2070 contents-only `!` re-include + `git rm --cached` exemption; `.claude.json`/`.claude/plugins/` stay gitignored (#1705 intact).
- **Self-heal at boot** — `startup.sh` runs `agent_server.plugins_reinstall` **after** credential injection; it hardened-parses the untrusted manifest (`AliasPolicy.REJECT`), re-charset-validates every name and the marketplace `source`, reads current state via `claude plugin [marketplace] list --json`, and re-installs only what's missing (**zero subprocesses** on a volume-persisting restart).
- **Source-mode / Cornelius fallback** — when the committed manifest never materialized, `load_manifest` falls back to the re-cloned `template.yaml`'s `plugins:` block.

## Security hardening (this branch's last commit)

The agent-side `_is_source` gate — the **sole** validator on the untrusted `template.yaml` fallback a source-mode agent's clone carries — accepted any `scheme://` (`ftp://`/`file://`/`ssh://`/`data://`) while the backend writer `_validate_source` accepts only `https://`. The looser copy gated untrusted input. Tightened to **https-only** and pinned cross-boundary agreement with a shared accept/reject table run against **both** copies (`test_1704_source_validator_parity.py`), encoding the invariant `agent_accepts ⊆ backend_accepts`. Learnings entry records the re-implemented-cross-boundary-policy drift class.

## Docs

- `docs/memory/requirements/content-files.md` §42 (reframe, functional reqs, security, honest scope)
- `docs/memory/architecture.md` (Agent Plugin Manifest subsystem + validator https-only note)
- `docs/memory/feature-flows/agent-plugin-manifest.md` (new) + index row + `template-processing.md`
- `docs/TRINITY_COMPATIBLE_AGENT_GUIDE.md` (Declaring Plugins section)

## Tests

121 test cases across 5 files — reader/materialize/commit-seam/boot-hook/wiring + the cross-boundary source-validator parity. All green locally.

## Deploy note

Base-image change — old images silently skip the boot hook (release-note ordering requirement).


vybe (approved):
Validated via /validate-pr: security scans clean (the `git@github.com`/`user:tok@github.com` hits are **rejection fixtures** asserting `False` against the https-only validator — correct test design), `Fixes #1704` closing keyword present, docs updated (requirements §42, architecture.md, new feature flow + index, agent guide).

The cross-boundary hardening is the right call: the agent-side `_is_source` gate was the sole validator on an untrusted `template.yaml` and accepted any `scheme://` while the backend accepted https-only — the looser copy gating untrusted input. Pinning `agent_accepts ⊆ backend_accepts` with a shared table run against both copies is the correct shape for an Invariant #5 mirror.

Conflict resolution (docs-only): merged `origin/dev` and resolved one `architecture.md` conflict against #2069 (just landed). Both edits preserved — dev's creation-time-seed block plus this PR's `plugins.yaml` addition to the `_TRINITY_AUTHORED_PATHS` enumeration. Verified `.trinity/plugins.yaml` is present in the runtime constant and #2069's merge functions are intact; 241 affected tests green on the merged tree.

Issue #1703, opened 2026-07-20
refactor: agent git repo root is $HOME — 22 gitignore patterns exist only to compensate, and every new home-writing feature repeats the dance


vybe:
## Summary

Trinity agent containers set `ENV HOME=/home/developer` and `WORKDIR=/home/developer` (base image), and the agent's **git repo root is that same directory**. Every piece of home-directory furniture and every byte of Claude Code runtime state therefore lands inside the agent's git working tree, and the platform compensates with a fleet-wide ignore list that keeps growing.

**22 of the current `_GITIGNORE_PATTERNS` entries exist solely because of this layout**: `.bashrc`, `.bash_logout`, `.profile`, `.bash_history`, `.sudo_as_admin_successful`, `.cache/`, `.local/`, `.npm/`, `.ssh/`, `.env`, `.tmp/`, `.claude.json`, `.claude/projects/`, `.claude/statsig/`, `.claude/todos/`, `.claude/debug/`, `.claude/sessions/`, `.claude/shell-snapshots/`, and friends. None of these would be candidates for commit if the repo lived in a subdirectory.

The cost is not the list itself — it is that **every new platform feature that writes into the agent home must rediscover the rule and add its own ignore entry**, and the ones that don't get discovered ship as repo-bloat bugs.

## Context

Surfaced during the full-directory skill-injection work (abilityai/trinity-enterprise#183): platform-injected skill packages land in `~/.claude/skills/`, which is inside the repo, so injection had to append per-skill `.gitignore` lines and untrack already-committed matches — otherwise the 15-minute auto-sync commits multi-MiB packages into every agent's GitHub repo (the #1595/#1596 class).

Adjacent instance found the same day: #1702 (`.claude/plugins/` never added to the list, so plugin caches are committed today).

Two facts that constrain the solution space:

1. **`~/.claude` is not relocatable.** Claude Code has no supported config-directory environment variable; the user-level path is fixed. So the skills/config cannot move out from under git — only the repo can move.
2. **Trinity used to have the good layout.** Legacy agents keep their repo at `/home/developer/workspace` with `$HOME` outside it (`git_service._detect_git_dir` still branches on this), and for those agents the problem does not exist.

## Acceptance Criteria

This is a decision issue first, an implementation second.

- [ ] Decide explicitly: **keep** home-as-repo-root and adopt the ignore list as stated policy, or **move** the repo root to a subdirectory
[...]

PR #2226, opened 2026-08-16
fix(git-sync): merge the canonical .gitignore at agent creation (#2069)

### LINKED ISSUE #2069: bug: the fleet-wide .gitignore merge never runs at agent creation — in-container auto-sync commits .trinity/ runtime state before any Push can migrate
## Summary

The fleet-wide `.gitignore` list (`_GITIGNORE_PATTERNS`, `src/backend/services/git_service.py:1269`) is never applied at agent creation. It is applied at exactly two later moments, both operator-initiated — so an agent created from an external `github:` template whose repo ships no `.gitignore` (or a partial one) starts life with **no** platform exclusions, while the in-container 15-min auto-sync loop is already committing with `git add -A`. Runtime state — `.trinity/`, `.claude/projects/`, `content/`, and the credential files — reaches the user's own repo before any Push can migrate the list.

## Where the canonical list is actually applied

| Call site | Trigger | Reaches a new template agent? |
|---|---|---|
| `initialize_git_in_container` step 2 (`git_service.py:1545`) | `POST /api/agents/{name}/git/initialize` — "create a new repo for this agent" | ❌ not on the `github:`-template path |
| `_migrate_workspace_gitignore` (`git_service.py:1440`, called at `git_service.py:1112`) | `sync_to_github` — the operator **Push** button / MCP `git_sync` | ❌ only once a human pushes |

Creation itself (`services/agent_service/crud.py`) calls neither. `_materialize_agent_files` touches `.gitignore` only via `_append_agent_gitignore` (`git_service.py:709`), and only for `data_paths` entries (#1169) — opt-in, and unrelated to the canonical list. `startup.sh` deliberately never writes `.gitignore` (#953, comments at `docker/base-image/startup.sh:239`, `:446`, `:561`). So a template agent's `.gitignore` is exactly what the template repo shipped.

## Why the Push-time fallback loses the race

The fallback assumed by #1908 ("it depends entirely on the fleet-wide `.gitignore` merge … at sync time") is not the default writer. The default writer is the **in-container** auto-sync loop, which never merges the list:

- `agent_server/auto_sync.py:44` → `agent_server/routers/git.py::_run_auto_sync_once`
- stages with a bare `git add -A` (`agent_server/routers/git.py:485`) against whatever `.gitignore` exists in the container

and it is **on from creation** for the two paths that push to a repo the user owns — `env_vars['GIT_SYNC_AUTO'] = 'true'` at `crud.py:1567` for non-source-mode (legacy working-branch) and fork-to-own (trinity-enterprise#93) agents. So such an agent auto-commits every 15 minutes from birth, and converges only if and when someone clicks Push.

## Impact

- Platform runtime state committed into a user-owned repo: `.trinity/pending-results/` (#1083), `.trinity/pipeline-state/` (#919), `.trinity/operator-queue.json`, `.trinity/persistent-state.yaml` (#383) — churny content, so the same unbounded-`.git`-growth class as #1596/#1702.
- Credential exposure is the stated risk of the same missing patterns in the compatibility checks (`.env`, `.mcp.json` live at the repo root because `$HOME` **is** the repo root — #1703).
- Silent: nothing fails, and the compatibility check that would surface it (`c_g002`, `services/compatibility/static_checks.py:752`) is advisory and per-agent.

## Scope — what this is *not*

#1908 (`status-in-dev`) fixed the **bundled** templates and proposes a CI guard scoped to them. Verified locally that this landed: 16 of the 17 bundled templates carrying a `.gitignore` now list `.trinity/`, and the remainder are hidden fixtures plus `trinity-system`. This issue is about the **platform-side timing**, which no template-side fix can reach — external `github:` templates are the general case and are outside any bundled-template CI guard.

## Proposed fix

Apply the existing merge at creation (and/or at start, before the auto-sync loop can run a cycle) for any agent that has a `.git` directory — reusing `_build_gitignore_merge_command` rather than adding a fourth copy of the pattern list.

## Acceptance Criteria

- [ ] An agent created from an external `github:` template whose repo ships no `.gitignore` (or one lacking the canonical patterns) has the merged list on disk **before** the first auto-sync cycle can commit
- [ ] The first in-container auto-sync of such an agent commits no `.trinity/`, `.claude/projects/`, `content/`, `.env` or `.mcp.json` path
- [ ] The merge reuses `_build_gitignore_merge_command` — `_GITIGNORE_PATTERNS` stays the single source of truth (no fourth call site with its own list)
- [ ] Idempotent: an agent whose template already ships the patterns shows no `M .gitignore` drift against `origin` (the exact regression #953 removed the shell-level append for)
- [ ] Existing agents keep converging via the Push-time migration — no behaviour change on `sync_to_github`
- [ ] Regression test covers the fork-to-own / non-source-mode path, since that is where `GIT_SYNC_AUTO` is on from creation

## Technical Notes

- Ordering matters: the merge must land after `startup.sh` finishes cloning and before the first auto-sync cycle. `auto_sync.run_auto_sync_loop` sleeps one full interval before its first cycle (`auto_sync.py:56`), which is the window to use — but the fix should not depend on that sleep for correctness.
- `_detect_git_dir` (`git_service.py:1416`) already resolves the legacy `workspace/` vs `$HOME` split; reuse it so init, Push-migration and this new call site agree.
- Consider whether the untrack half (`_build_rm_cached_ignored_command`) should run at creation too — probably not, and see the sibling issue on its exemption list before wiring it anywhere new.
- #1703 (repo root is `$HOME`) is the structural fix that would retire this whole compensation layer; this is the bounded fix until then.
- Related: #1908 (bundled templates ship no `.gitignore` — template-side half), #1702, #1596, #462, #2036 (prior instances of runtime state reaching agent repos), #1655 (bundled-template CI validation).

AndriiPasternak31:
Fixes #2069

## Problem
The fleet-wide `_GITIGNORE_PATTERNS` list was applied at exactly two operator-initiated moments — the on-Push migration and the init-path merge — and **never at agent creation**. The in-container 15-min auto-sync loop, however, is on **from birth** for the `GIT_SYNC_AUTO` set (non-source / fork-to-own `github:` agents, ephemeral ghosts included), so a fresh agent auto-committed `.trinity/` runtime state, `.claude/projects/`, `content/` and the root-level `.env` / `.mcp.json` into its **user-owned** repo on the first cycle — before any Push could migrate the list. That's a credential-leak-into-a-private-repo hygiene hazard plus the #1595/#1596 unbounded-`.git`-bloat class. (#1908 fixed only the bundled templates; external `github:` templates are the general case.)

## Fix
Land the merged list on disk **after** startup.sh finishes ALL its git setup and **before** the first auto-sync cycle can commit — the bounded compensation until #1703 (repo root ≡ `$HOME`) retires the layer.

- **`git_service.merge_gitignore_after_clone`** — readiness-gated (direct agent `/health` probe, not the masking proxy) ∧ `.git`-present; then reuses **`_build_gitignore_merge_command`** verbatim (`_GITIGNORE_PATTERNS` stays the single source of truth — no fourth call site, nothing crosses into the agent-server / Invariant #5). **Merge-only = PREVENT** (no `rm --cached` untrack at creation; the generated `.env`/`.mcp.json` are untracked, so a merge-installed `.gitignore` stops `git add -A` from ever staging them).
- **`spawn_gitignore_merge_after_clone`** — fire-and-forget (zero creation latency), monotonic deadline, `asyncio.Semaphore`-capped **on the Docker-exec section only** (the readiness poll is pure HTTP and runs outside the cap, so slow-booting agents can't head-of-line-block a healthy one), every exec/HTTP `wait_for`-bounded, non-fatal.
- **Gated on `_git_auto_sync_baked`** — the single owner of "does this agent bake `GIT_SYNC_AUTO`", shared with `crud.py::_apply_github_env`, so the merge covers exactly the auto-committing population (ephemeral non-source `github:`+PAT ghosts included; source-mode excluded).
- **Wired** at creation (`crud.py::_materialize_agent_files`) and — **T1 fleet remediation** — at start/recreate (`lifecycle.start_agent_internal`, gated on the DB `auto_sync_enabled` flag) so existing leakers converge on their next recreate/restart. No behaviour change on Push (`sync_to_github`) — Push migration remains the backstop.
- **Idempotent (#953)**: `grep -qxF` gate → no `M .gitignore` drift for an already-compliant template; a stale wholesale `.trinity/` line gets a legitimate supersede→append (#2070).

## Docs
- `requirements/github.md` §11.14 (new)
- `architecture.md` — Git Sync Health gained "merged into each agent on sync **and at creation**" + a Creation-time seed block
- `feature-flows/git-sync-health.md` — new §0 execution flow + files-touched table
- `feature-flows/github-sync.md` — REMEDIATE / PREVENT cross-reference; `feature-flows.md` index row

## Tests (local: all green)
- `test_2069_gitignore_at_creation.py` — merge behaviour against real tmp git repos, readiness gating, the `_git_auto_sync_baked` matrix, creation + start/recreate wiring, and the poll-outside-semaphore concurrency invariant (**30 passed**)
- `test_2069_gitignore_merge_caller_guard.py` — AST writer-SET guard on the merge-command callers + `_GITIGNORE_PATTERNS` name-refs (no fourth pattern list)
- Adjacent `test_ent123_tokenless_clone.py` / `test_fork_to_own.py` extended (**90 passed**)

## Follow-ups (filed separately)
- `GIT_SYNC_AUTO` is baked for ephemeral ghosts at all, contradicting "ghosts never auto-push" — a distinct pre-existing bug; #2069 only ensures the merge *covers* everyone who auto-syncs.
- #1703 (repo root ≡ `$HOME`) is the structural fix that retires this compensation layer.


vybe (approved):
Validated via /validate-pr: security scans clean, `Fixes #2069` closing keyword present, docs updated (requirements/github.md §11.14, architecture.md, two feature flows + index). Reuses `_build_gitignore_merge_command` verbatim so `_GITIGNORE_PATTERNS` stays the single source of truth, with an AST caller-set guard pinning it — correct shape for this class.

CI note: the one red is `pytest (base, seed 12345)` = **CANCELLED**, and it is the *baseline* arm, not the head. All three head seeds and `regression diff` are green; not a product failure.

⚠ Non-blocking follow-up (packaging, #1056 class): `TRINITY_GITIGNORE_MERGE_TIMEOUT_SECONDS` / `_INTERVAL_SECONDS` / `_CONCURRENCY` are read via `os.getenv` in `git_service.py` but wired into neither `docker-compose.yml`, `docker-compose.prod.yml`, nor `.env.example` — so the documented tuning levers are inert on deploy. Not a merge blocker: all three carry safe defaults (1800/5/6) and the feature is default-ON and fully functional at them; nothing is stuck OFF. Worth a one-line follow-up so the knobs are actually reachable.

PR #2227, opened 2026-08-16
test(ci): collect self-contained root tests + make them island-safe (#1895)

### LINKED ISSUE #1895: bug: 108 test files sit in tests/ root, where no CI job runs them

AndriiPasternak31:
Fixes #1895

## Problem

108 self-contained `test_*.py` files sit in `tests/` root, which **no per-PR CI job collects** — the unit job runs `cd tests && pytest unit/`, and `tests/unit/pytest.ini`'s `norecursedirs = ..` seals the root out. That structural gap is how `test_canary_invariants` (guarding the #1880 canary alert parity) shipped gated nowhere.

## What this does

1. **Relocates** the self-contained root tests into `tests/unit/` so the per-PR unit job collects them; splits clearly-mixed files (unit half moved, live-API half stays in root); pure live-API tests stay in root.
2. **Adds a two-part CI guard** (`tests/lint_root_test_placement.py`, wired into the unconditional `lint-sys-modules` job): (a) no self-contained test may sit in `tests/` root, (b) no unmarked `async def test_` under `tests/unit/` (the island runs pytest-asyncio in **strict** mode). Durable placement rule in `tests/README.md` + the unit conftest header.
3. **Makes the relocated files island-safe** under `pytest-randomly` (the bulk of the review surface — see below).

## Island-safety fixes (the hard part)

Relocating files reshuffles the whole `pytest-randomly` order, which exposed pre-existing latent order-fragilities. Three root causes, all confirmed by object-id instrumentation (never guessed):

- **del/reload+reimport of a service module → zombie module.** Three fixtures injected a stub by evicting a service module from `sys.modules` and re-importing it. In a shared island this leaves a fresh module object; a later test's captured instance (or `from services.X import fn` consumer) then diverges from the entry a sibling patches → the patch silently no-ops (broke test_117 `no_voice_id`, test_1609 rate limit, test_1081 506 clamp). **Fix:** patch the captured reference **in place** (`monkeypatch.setattr(module, "db"/"platform_audit_service"/…)`) — never reimport. Snapshot/restore is *not* enough (instances keep referencing the fresh module).
- **Collection-time config leak** (test_voice_auth WS 4001): a `pytest_collectstart` hook restores the **stateless** shared modules (config/models/database/utils*) before each Module import, so a test binding `from config import SECRET_KEY` at import captures the real key. It excludes singleton modules by design.
- **Import-time utils stubs** (test_self_execute / test_watchdog_unit): confined with the lint-recognised `_STUBBED_MODULE_NAMES` + `_restore_sys_modules` pair.

## Verification

Ran the **exact CI command** (`cd tests && pytest unit/ -m "not slow" -p randomly --randomly-seed=<seed>`) across the 3 CI seeds (12345 / 67890 / 99999), diffing head-vs-base JUnit XML per seed. Adding files reshuffles the order, so base-seed-X ≠ head-seed-X — the only sound check is **NEW-in-head == [] on all three**, which holds. Guards + their unit tests green; `lint_sys_modules` clean.


vybe (approved):
Validated via /validate-pr. Ratchet lint (baseline + `--regenerate-baseline`) wired unconditionally into the sys.modules lint job so it can't be path-filtered away; 18 relocations verified island-safe across all three CI seeds with NEW-in-head == [] each. Docs (tests/README.md + unit conftest header) carry the durable rule. All checks green.

Issue #1710, opened 2026-07-20
refactor(auth): audit and retire the slack.py inv8 carve-out left by #1310


vybe:
## Summary

`#1310` consolidated Trinity's inline auth/dependency patterns behind five leaf guard helpers in `src/backend/dependencies.py` (`assert_admin`, `assert_agent_access`, `assert_agent_owner`, `assert_owns_or_admin`, `assert_owns`). It deliberately left **`src/backend/routers/slack.py` out of scope** — its ~10 inline auth sites carry `# noqa: inv8` markers and a line-exemption in the static guard `tests/unit/test_1310_auth_wiring.py`.

The commit says to "coordinate with the channel-adapter owner", but **no follow-up issue was ever filed**. The result is an expiry-less waiver on an auth-boundary invariant, with an in-repo suppression that reads as permanent.

## Context

Found by `/release-plan` during the v0.8.5 pre-freeze coherence review (2026-07-20). Nothing is known to be broken today — this is a governance gap, not a live vulnerability. The risk is specifically that a suppression with no ticket behind it becomes invisible: the next person to read `# noqa: inv8` has no way to tell whether it is a considered exception or an abandoned TODO.

Invariant #8 also requires enumeration-safety (self-uniform handlers — never an existence-404 followed by an access-403 in the same function). The `slack.py` sites have not been audited against that rule, only exempted from the wiring guard.

## Acceptance Criteria

- [ ] Each `# noqa: inv8` site in `src/backend/routers/slack.py` is audited and classified: migrate to a shared guard helper, or document why it is a legitimate permanent exception
- [ ] Every site that can migrate does so, and its `# noqa: inv8` marker is removed
- [ ] Any site that stays exempt carries an inline comment naming the reason (following the pattern used by the documented permanent exceptions in `nevermined.py` / `reports.py` / `sessions.py`)
- [ ] The line-exemption list in `tests/unit/test_1310_auth_wiring.py` shrinks to exactly the remaining documented exceptions
- [ ] Each audited site is confirmed **self-uniform** per Invariant #8 — access checked before existence, so no 404/403 differential leaks whether a resource exists
- [ ] No behavioral change to Slack auth outcomes for existing installs (this is a wiring consolidation, not a policy change)

## Technical Notes

- Guard helpers and the invariant text: `src/backend/dependencies.py`, `docs/memory/architecture.md` → Architectural Invariants #8
[...]

PR #2229, opened 2026-08-16
fix(canary): run-state observability — /status + canary_enabled flag (#2217)

### LINKED ISSUE #2217: bug: canary harness left disabled on a retired constraint, and nothing reports whether it is running

AndriiPasternak31:
Fixes #2217

## What & why

A disabled canary emits **zero violations** — byte-for-byte identical to a clean fleet — so nothing reported whether the harness was even running. This is the **H-01 class one level up, applied to the detector itself**: H-01 catches a blind collector *while a cycle runs*; it structurally cannot catch "no cycle is running at all" (a dead loop emits nothing). The harness had also been switched OFF on the dev instance on the retired belief that its snapshot reader is SQLite-only — a constraint **removed by #1540** (every SQL-tier collector now reads the configured backend through the `get_engine()`/`DATABASE_URL` seam).

## Changes

- **`GET /api/canary/status`** (admin-only) → `CanaryService.get_run_status()` (Invariant #1 — logic in the service, thin router). Returns `{enabled, status, last_cycle_at, seconds_since_last_cycle, interval_seconds, stale_after_seconds, alert_sink_configured, redis_available}`.
- **`status` ∈ `disabled|healthy|stale|unknown`** — all three non-healthy states distinct from *enabled + fresh + zero-violations*. Reads the **shared** `canary:last_cycle_at` cursor (never the per-worker counters, stale/zero on a non-leader under #1881's lease).
- **Fail-open so default-OFF never alarms**: `disabled` short-circuits before Redis (`redis_available=None`); any Redis/parse failure → `unknown`, never `stale`.
- **Threshold** `stale_after_seconds = _max_failover_seconds() + _MAX_CYCLE_LEASE_SECONDS` (≈1680s) — both the service's own constants, provably above the ~780s failover window AND a maxed-but-healthy ≤900s cycle.
- **`alert_sink_configured`** is a **separate** field (liveness vs can-it-alert are orthogonal).
- **`canary_enabled`** boolean added to `GET /api/settings/feature-flags` (observability-only flag home, any authed user; detail stays admin-only), backed by public `CanaryService.is_enabled()`.
- No schema change, no migration, no new config knob, no compose change.

## Acceptance criteria

| AC | Status |
|----|--------|
| 1 — `CANARY_ENABLED` restored on dev instance, cycling against PostgreSQL | ⚠️ **Ops step — see hand-off below** (code PR cannot perform it) |
| 2 — run-state observable without shell access | ✅ `/status` + `canary_enabled` flag |
| 3 — disabled/stale distinguishable from no-violations | ✅ `status: disabled\|stale\|healthy\|unknown` |
| 4 — PostgreSQL docs state harness is backend-agnostic | ✅ `docs/POSTGRESQL_SETUP.md` + `docs/migrations/SQLITE_TO_POSTGRES.md` |

## ⚠️ Ops hand-off (required for AC#1 — manual, not in this PR)

**AC#1 is not code.** To complete it an operator must, on the **dev instance**:

1. Set `CANARY_ENABLED=1` in the instance `.env`.
2. Redeploy the backend (compose wiring already shipped in #1876; the harness is backend-agnostic on PostgreSQL since #1540, so this is safe).
3. Confirm cycling via the new surface: `GET /api/canary/status` → `{"enabled": true, "status": "healthy", ...}` once the first cycle lands.

This is a manual environment change this code PR cannot perform.

## Docs updated (pre-push gate)

- `docs/memory/architecture.md` — Canary Harness "Run-state observability (#2217)" subsection + API Endpoints table (`/status` row, `/run-cycle` cursor note) + feature-flags row
- `docs/memory/requirements/infrastructure.md` — §31.2
- `docs/memory/feature-flows.md` — index row
- `docs/POSTGRESQL_SETUP.md`, `docs/migrations/SQLITE_TO_POSTGRES.md` — backend-agnostic note (AC#4)

## Tests

`tests/unit/test_2217_canary_status.py` — 21 tests, all green: disabled short-circuit, unknown fail-open, stale/healthy boundary, Redis-down vs missing-cursor, unparseable cursor, clock-skew clamp, full model round-trip, feature-flags `canary_enabled`, and route-registration + `require_admin` gate. Sibling canary guards (73 tests) unaffected.


vybe (approved):
Validated via /validate-pr. Thin router over CanaryService.get_run_status (Invariant #1); /status is a distinct root literal with no /{param} sibling in this router, so no Invariant #4 collision. Fail-open ladder is right: disabled short-circuits before Redis, any read/parse failure reads unknown never stale, so a default-OFF install cannot alarm. stale_after derived from the service's own _max_failover_seconds + _MAX_CYCLE_LEASE_SECONDS so it can't drift from the timing it guards. Resolved a feature-flows.md changelog-row conflict against #1710 (both rows kept). All 21 checks green.

Issue #831, opened 2026-05-13
feat: platform default model setting + UI clarity for null model on schedules


vybe:
## Summary

Add a configurable platform-wide default model setting (in Settings) and fix the UI so users can clearly see when a schedule has no model explicitly set and what runtime behaviour to expect.

## Context

Discovered via an incident (related bug: headless_executor.py defaults to `sonnet` when `model=NULL`, while the platform's intended default is Opus 4.5). All 44 production schedules have `model=NULL`, which caused Sonnet quota exhaustion because:

1. The `ModelSelector` component labels `claude-opus-4-5-20251101` as "Default — most capable" — users assume null = Opus 4.5.
2. Inside the agent container, `headless_executor.py:143` resolves null to `"sonnet"` (a stale Issue #81 workaround).
3. There is no single authoritative platform default model. The UI, the base image, and the scheduler all make independent assumptions that have drifted out of sync.

The fix has two layers:
- **Settings**: expose a `platform_default_model` setting admins can configure (stored in `system_settings`, surfaced via `GET /api/settings/feature-flags` or a new settings key).
- **UI clarity**: schedules and the model selector should visibly distinguish "explicitly set to X" from "inheriting platform default (currently Y)".

## Acceptance Criteria

- [ ] Admin can set a platform default model in Settings (e.g. `claude-opus-4-5-20251101`); stored in `system_settings` table under key `platform_default_model`.
- [ ] `GET /api/settings/feature-flags` (or a dedicated endpoint) returns the resolved platform default model so the frontend can display it.
- [ ] `headless_executor.py` reads the platform default from an env var (injected by the backend at dispatch time) instead of hardcoding `"sonnet"`. The env var falls back to `claude-opus-4-5-20251101` if not set, matching the UI label.
- [ ] `claude_code.py` (interactive chat path) applies the same resolved default for parity.
- [ ] `SchedulesPanel.vue`: when `schedule.model` is null, the model column shows e.g. `platform default (claude-opus-4-5-20251101)` rather than being blank or omitting the field entirely.
- [ ] `ModelSelector.vue`: when the bound value is empty/null, the placeholder shows the current platform default (fetched from backend) rather than the generic "Select or type a model..." string.
- [ ] Schedule create/edit form: add a "Use platform default" toggle or explicit null-state so users can intentionally leave the model unset without confusion.
- [ ] No migration required for existing null-model schedules — they automatically pick up the platform default at dispatch time once the backend passes the resolved model.

## Technical Notes

[...]

PR #2230, opened 2026-08-16
refactor(#2086): centralize the selectable model catalog — one source of truth

### LINKED ISSUE #2086: refactor: centralize the selectable model catalog — one source of truth for three hand-synced lists
## Summary

The selectable Claude model catalog is maintained by hand in **three independent
places** with no shared registry and no CI guard, so the copies drift silently.
They are drifted right now: `claude-opus-5` — the current Opus tier — is absent
from all three.

| Location | Role |
|---|---|
| `src/frontend/src/components/ModelSelector.vue` (`PRESET_MODELS`) | The per-agent / schedule / loop model picker |
| `src/backend/services/settings_service.py` (`PUBLIC_CHANNEL_MODELS`) | Server-side allow-list validating the #894 per-agent public-channel override |
| `src/frontend/src/views/Settings.vue` | Admin platform-default-model dropdown |

The hazard is already documented in the code — `settings_service.py` carries the
comment *"Mirrors the current presets in `ModelSelector.vue` (kept in sync by hand
— there is no shared model registry across the Python/Vue boundary)"* — but the
comment is the only enforcement, and a comment is not a guard.

## Context

Surfaced by a routine "is the model list current?" check. Two properties make the
drift worse than cosmetic:

1. **It is not symmetric.** `PUBLIC_CHANNEL_MODELS` is a *validation* set, so a
   model missing there is rejected with a 422 on
   `PUT /api/agents/{name}/public-channel-model` — an owner cannot select a valid,
   shipping model even via the API. The frontend lists degrade quietly (the picker
   accepts free text); the backend list fails loudly and blocks the operation.
2. **Nothing announces the drift.** There is no parity test, no lint, and no
   release-time checklist item. The `ModelSelector.vue` header comment carries a
   hand-typed `Last updated:` date, which is the same class of manually-maintained
   state the issue is about.

Model releases are frequent enough that a hand-sync step across a Python/Vue
boundary will keep being missed. This issue is the structural fix; adding
`claude-opus-5` to the current three lists is the separate immediate fix and does
not depend on this work landing.

## Acceptance Criteria

- [ ] A single source of truth defines the selectable model catalog (id, display
      label, note/tier, and whether the entry is current or legacy).
- [ ] All three consumers above read from it; no consumer carries its own literal
      model list.
- [ ] A CI guard fails the build when a consumer drifts from the source of truth
      (the mechanism depends on the approach chosen below — a parity test if the
      lists stay materialized, or a codegen freshness check if generated).
- [ ] `claude-opus-5` is present and selectable end-to-end (picker → save →
      `PUT /api/agents/{name}/public-channel-model` returns 200, not 422).
- [ ] Existing graceful-degradation behavior is preserved, not regressed:
      - `db.get_public_channel_model` still treats a model absent from the
        allow-list as unset → platform default (the #1080 posture).
      - `ModelSelector.vue` still accepts free-text entry for ids outside the
        preset list.
- [ ] If the frontend resolves the catalog at runtime rather than at build time,
      a fetch failure degrades to a usable picker (free-text entry still works;
      no empty dropdown and no dead end).
- [ ] Adding a future model is a **single-file** edit, and the PR checklist / docs
      say where that file is.

## Technical Notes

**Design decision to make first** — the reason this isn't already centralized is
the Python/Vue boundary. Three plausible shapes, in rough order of cost:

1. **CI parity guard only.** Keep the three lists; add a test asserting they
   agree. Cheapest, preserves all current behavior, but still three edits per
   model release — it converts silent drift into a loud failure without removing
   the sync work.
2. **Backend-served catalog.** Backend owns the list and exposes it (extend the
   existing `GET /api/settings/feature-flags`, which the frontend already reads
   for gating, or add a dedicated endpoint). Removes the duplication properly;
   needs the fetch-failure fallback in the criteria above.
3. **Generated from one source.** A JSON/YAML manifest generating both the Python
   frozenset and the Vue constant, with a CI freshness check. No runtime
   dependency, but adds a codegen step to the build.

Recommendation: (2) if the catalog should be admin-configurable later, (1) as a
stopgap if that's not wanted yet. Worth deciding explicitly rather than defaulting.

**Explicit non-goal — do not fold in `model_context.py`.** `src/backend/services/model_context.py`
also maps model ids, but to *context-window* values, and it is deliberately
prefix-based (`claude` → the safe 200K floor) so it already resolves new ids
correctly with no edit. It carries a documented `#1521` design invariant against
resolving a bare Claude id to its 1M capability ceiling, and it is vendored
**byte-identically** into `docker/base-image/agent_server/model_context.py` under
architecture Invariant #5 with a parity test. Merging it into a
display-catalog registry would break both the invariant and the vendoring
contract. Leave it alone.

**Also out of scope:** `PLATFORM_DEFAULT_MODEL_VALUE` in `settings_service.py`
and the matching agent-side runtime defaults are a cost/latency *policy* choice,
not catalog data — changing the fleet default is a separate decision.

**Related, lower priority:** the model-id prettifier in `src/backend/routers/ops.py`
matches on prefixes that predate the current generation, so several shipping
model ids already fall through to its generic fallback. Cosmetic and
pre-existing; fold it in only if the chosen approach makes it free.

AndriiPasternak31:
Fixes #2086

## What

Centralizes the selectable Claude-model catalog into **one source of truth**, replacing three hand-synced lists (the #894 backend allow-list, the `ModelSelector.vue` picker, the `Settings.vue` admin default dropdown) that were drifting silently — `claude-opus-5` was missing from all three, and because `PUBLIC_CHANNEL_MODELS` is a *validation* set, an owner could not even select it via the API (422).

Approach **(3) generated-from-one-source** with a CI freshness guard (no runtime fetch, no new endpoint).

## Changes

- **`src/backend/services/model_catalog.py`** — new stdlib-only leaf defining `MODEL_CATALOG` (id, label, note + `public_channel` / `admin_default_selectable` / `recommended` policy flags). Import-time invariant: exactly one recommended entry, pinned to `PLATFORM_DEFAULT_MODEL_VALUE` (#831).
- **`scripts/gen_model_catalog.py`** → emits the checked-in, do-not-edit `src/frontend/src/constants/modelCatalog.js` (`.gitattributes` pins LF for byte-stable regen).
- **Three consumers now derive, none keeps a literal list**: `settings_service.PUBLIC_CHANNEL_MODELS` re-exports the catalog set; `ModelSelector.vue` maps `MODEL_CATALOG` → presets (free-text + `[1m]` passthrough preserved); `Settings.vue` `v-for`s the `adminDefaultSelectable` set with `(recommended)` riding the flag.
- **`claude-opus-5` added** (public + admin) → selectable end-to-end (422→200); prior Opus gen (4.8/4.7/4.6) relabelled Legacy.
- **CI guard `tests/unit/test_2086_model_catalog_parity.py`** (rides `backend-unit-test.yml`, every PR): byte-freshness **and** structural validation, opus-5 e2e, per-flag (#1080 Haiku public-not-default), recommended==default, backend-derive. Retired the brittle `test_1660` Vue-regex guard (intent folded in); repointed `test_ent243` / `test_model_selection` / `test_platform_default_model` at the catalog.

## Adding a future model — single-file edit

Add one `ModelEntry` to `MODEL_CATALOG` in `src/backend/services/model_catalog.py` and re-run `python scripts/gen_model_catalog.py`. Documented in the module docstring, the codegen script, requirements `public-access.md` §47.2 (FR-6), and both feature-flow docs.

## Docs

`architecture.md` (#1521 note + #894 endpoint row), `requirements/public-access.md` §47.2, `feature-flows/model-selection.md`, `feature-flows/public-channel-model.md`.

## Validation

`/validate-pr` clean: docs/requirements/feature-flows/tests all updated; no new `os.getenv`/top-level module (services/ copied wholesale); no secrets/PII. Parity + prompt_tier unit guard: **47 passed** locally. Integration mirrors (`test_model_selection`, `test_platform_default_model`) need a running server — not run in this session.

Known-deferred (FR-7): `ops.py::_format_model_name` / `observability.js::formatModelName` display prettifiers (title-case fallback, no list) — same class, deliberately out of scope, tracked as follow-up.


vybe (approved):
Validated via /validate-pr. Genuine single-source-of-truth: settings_service.PUBLIC_CHANNEL_MODELS is now a re-export, ModelSelector.vue and Settings.vue derive from the generated mirror, and none of the three keeps a literal list. The byte-freshness parity test rides backend-unit-test on every PR, and .gitattributes pins LF so an autocrlf contributor's regen matches CI. Import-time invariant (exactly one recommended entry, pinned to PLATFORM_DEFAULT_MODEL_VALUE) keeps #831 honest. claude-opus-5 now selectable end-to-end (422→200). Deferred display prettifiers (FR-7) named rather than silently skipped. All checks green incl. frontend build + e2e.

PR #2231, opened 2026-08-16
refactor(auth): retire the slack.py inv8 carve-out left by #1310 (#1710)

### LINKED ISSUE #1710: refactor(auth): audit and retire the slack.py inv8 carve-out left by #1310

AndriiPasternak31:
Fixes #1710

## What

Retires the `slack.py` Invariant-8 carve-out that #1310 deliberately deferred. The 11 inline `if not db.can_user_*_agent(...): raise 403` sites — each carrying an expiry-less `# noqa: inv8` waiver — are migrated onto the shared #1310 imperative guards in `dependencies.py`:

- **3 read gates** → `assert_agent_access` (status, get-channel, list-channels)
- **8 owner gates** → `assert_agent_owner` (connect, disconnect, update, create, delete, dm-default, proactive-toggle, send)

**Behavior-preserving** (AC #6): same `403` + `detail` string byte-for-byte. Every site was already access-first, so no `404`-before-`403` enumeration oracle is introduced (AC #5). The helpers add `_enforce_connector_scope` first — redundant belt-and-braces, since a connector key is already 403'd at the auth entry point before reaching any slack handler. Site #10 (proactive-consent toggle) keeps its ent#223 human-only `reject_agent_principal` line (`assert_agent_owner` is agent-permissive). **Zero `# noqa: inv8` markers remain on any source line.**

## Docs (pre-push gate)

- **architecture.md Invariant #8** — the `slack.py … deferred follow-up (# noqa: inv8-marked)` clause is **retired**.
- **requirements/auth.md** — `slack.py deferred` bullet → `slack.py retired (#1710)`; static-guard note updated.
- **feature-flows/slack-integration.md** — Authorization Checks table rewritten to `assert_agent_*() → db.can_user_*()` with current line numbers; changelog row added.
- **feature-flows.md** index — #1710 row added.

## Tests

- `test_1310_auth_wiring.py` — static guard: slack.py line-exemption retired; a re-introduced inline gate now trips unless it carries a fresh, individually-reviewed marker.
- `test_1310_auth_consolidation.py` — behavioral proof (real-DB harness): status + detail + admitted-principal set per migrated site; stranger→403, shared-reader read/owner matrix, site-#10 human-only vs site-#11 send asymmetry.
- `test_223_slack_channel_proactive_consent.py` — runtime ordering proof for the kept `reject_agent_principal`.
- `test_1649_group_message_history.py` — **fix caught by /validate-pr**: the 3 `TestRoutersPersist` cases call the now-`assert_agent_owner`-gated `send_agent_slack_channel_message`, which reads `current_user.connector_agent`. The `SimpleNamespace` fakes lacked the field → `AttributeError`. Added `connector_agent=None` (a real non-connector `User` carries it). Proven load-bearing: the 3 cases fail on the pre-fix fixtures, pass with it.

**81 passed** across the four files (venv-test, bcrypt 4.2.1).

## Scope

IDOR / public-link / rate-limit logic untouched. No schema, endpoint, or config change.

---

@dolho — heads-up as channel-adapter owner: this only swaps the auth-gate *shape* in `slack.py` (inline `db.can_user_*` → the shared `assert_agent_*` helpers), behavior byte-preserved; no routing/adapter logic touched. Flagging for visibility, not blocking.


vybe (approved):
Validated via /validate-pr. Mechanical 1:1 migration of 11 inline slack.py gates onto assert_agent_access / assert_agent_owner — same 403 status and same detail strings, so no API contract change, and the migration additionally picks up _enforce_connector_scope (a tightening, which is the point of #1710). The inv8 static guard is updated to assert no noqa marker remains, so a re-introduced inline gate trips. All checks green.

PR #2232, opened 2026-08-16
fix(#1920): consolidate hand-rolled Redis single-flight locks onto ownership-checked SingleFlightLock

### LINKED ISSUE #1920: bug: hand-rolled Redis single-flight locks — 7 divergent copies; system_seed_service releases with an unconditional tokenless delete

AndriiPasternak31:
Fixes #1920

## What

Consolidates the hand-rolled Redis SETNX single-flight lock idiom — copied into **seven** sync sites with divergent behaviour — onto **one** ownership-checked, fail-open primitive, `redis_breaker_util.SingleFlightLock`. Fixes the live #1919 bug class (extend/release a lock you no longer own) in the three sites that carried it, worst being `system_seed_service`'s **unconditional, tokenless `r.delete(...)`** on release, which can delete a *successor's* lock after the original holder's 600s TTL lapsed.

`cornelius_agent_service` and `services/compatibility/fixes.py` were found (while building the anti-recurrence guard) to carry the **same** constant-`"1"` value + tokenless-delete bug verbatim, and are fixed alongside.

## The primitive (`redis_breaker_util.SingleFlightLock`)

- Mints a **unique per-acquire `uuid` token** — closes the constant-`"1"` bug where A's release matches B's lock.
- `acquire()` / `refresh_if_owned()` → `LeaseState{OWNED, REACQUIRED, LOST, DEGRADED}` / `release_if_owned()`.
- **Fail-open**: a `None` client (Redis unreachable) or any Redis error behaves as the sole worker (`acquire` returns True, `held` stays False, so refresh/release are no-ops).
- **Single-use state machine**: `_held` is set once on a real SETNX win and never cleared — a DEGRADED refresh must not abandon a live lock for the full TTL (the `ops:fleet_restart` 2100s-wedge class).
- Release is GET-then-DELETE **compare-and-delete**, deliberately non-atomic (the accepted TOCTOU window `ops.py` already shipped) so it stays fakeredis-unit-testable. The async Lua-CAD `ResumeLock` remains the stronger, **deliberately-separate** primitive — not merged onto this.
- Client is **injected** (no helper-owned factory) so every existing per-site `get_breaker_redis` monkeypatch keeps biting.

## Acceptance criteria

| AC | Status | Where |
|----|--------|-------|
| `system_seed_service` release is ownership-checked (token + compare-and-delete) | ✅ | commit `4038bcd1` (THE FIX); requirements/roadmap.md §16.5.1 |
| Shared `SingleFlightLock` in `redis_breaker_util` — `acquire` / `refresh_if_owned` / `release_if_owned`, fail-open | ✅ | commit `7c6a5ef9`; architecture.md `redis_breaker_util` row |
| `ops.py` and `ephemeral.py` consume it (behaviour-preserving; #1919's tests stay green) | ✅ | commits `a94d78d8` (ops), `cfe59221` (ephemeral); `test_1860_fleet_restart_adoption.py` green (27 passed) |
| Per-site deliberate properties preserved | ✅ | `ResumeLock` (async Lua), `monitoring`/`operator_queue` leader leases, canary leader lease all kept separate (static-guard `_ALLOWED` + architecture "four lock idioms remain") |

Beyond the four ACs, the consolidation was extended to `skill_service` ×2 and the two twins (`cornelius`, `compat_fix`), since they shared the same defect — all documented.

## Anti-recurrence

`tests/unit/test_1920_no_hand_rolled_single_flight.py` fails CI when a new hand-rolled `set(..., nx=True, ex=...)` single-flight lock appears outside `SingleFlightLock` with no allowlist entry, and pins the adopted files clean. A doc line doesn't fail CI; this does.

## Tests

`tests/unit/test_1920_single_flight_lock.py` (28 cases) pins the primitive's contract against fakeredis + MagicMock, headed by `test_release_never_deletes_a_successor_lock` (the exact #1920 bug). Per-site successor-safety + fail-open cases added to `test_ent107_cornelius_seed.py` and `test_compatibility_checks.py`.

**304 passed / 4 skipped** across every #1920-touched module under full-dir (CI-path) collection; `test_1860_fleet_restart_adoption.py` (the #1919 ops behaviours) green.

## Docs

- `architecture.md` — `SingleFlightLock` primitive + full consumer list; consolidation, compat-fix, cornelius Redis-key entries.
- `feature-flows/` — ephemeral-agents, skill-injection, system-manifest, cornelius-default-agent, agent-compatibility-validation flow notes.
- `feature-flows.md` — Recent-Updates index row.
- `requirements/roadmap.md` — §16.5.1 system-seed lock note.

## Safety

No schema change, no migration, no feature flag (Redis-only). Behaviour-preserving refactor; the #1920 lock fix is class hygiene — system_seed's real double-seed convergence guard remains the reserved-name existence backstop.


vybe (approved):
Reviewed the primitive and all six adoptions. This is a good change, and the part I want to single out is that **the guard is the deliverable, not the cleanup** — which this rebase proved on itself, see below.

### The primitive

`SingleFlightLock` is well-judged. Three decisions I checked specifically because they are the ones that would bite later:

- **`_held` is set once and never cleared.** `refresh_if_owned` returning `LOST`/`DEGRADED` must not clear it, or a caller that reclaims its own lock after a stale-replica false-foreign read abandons a live lock for the full TTL. That is the `ops:fleet_restart` 2100s-wedge class, and the docstring names it.
- **`release_if_owned` gates on the stable acquire-time `held`, not on a refresh result** — so it still runs after a detected `LOST`, and is foreign-safe by construction rather than by the caller remembering. Verified: with a foreign token in the key, `lock_token_matches` fails and nothing is deleted.
- **The client is injected, with no `client_factory` default.** Every existing lock test monkeypatches the *site* binding (`ops.get_breaker_redis`, `sss.get_breaker_redis`, …); a helper-owned factory would resolve `redis_breaker_util.get_breaker_redis` and silently no-op every one of those patches — a whole test suite going green while testing nothing. Worth the comment it got.

Also correct to keep `ResumeLock` out. Its atomic Lua CAD is strictly stronger and the async/blocking-poll shape is structurally different; merging it onto GET-then-DELETE would be a downgrade dressed as consolidation.

`ops.py` is faithful — the `OWNED`/`REACQUIRED`/`LOST`/`DEGRADED` branches map 1:1 onto the four old inline outcomes, including the "key vanished in the GET→EXPIRE sliver counts as absent, not foreign" subtlety and the `last_current` foreign-vs-absent log word. `compat_fix` preserves its three-way contract exactly (`from_url` raises → `None` fail-open; internal SETNX error → held-False lock that no-ops on release; real busy → `FixBusy`). `cornelius` and `system_seed` genuinely carried the constant-`"1"` + unconditional-delete bug verbatim, and both are now token-checked.

**One small regression, non-blocking:** the `DEGRADED` warning in `ops.py` loses its `exc_info=True`, because the Redis exception is now swallowed inside `refresh_if_owned` (which is deliberately silent). The message still fires once per run, so a full-run outage is still visible — but the traceback that told you *why* Redis was erroring no longer reaches the log. The PR is upfront about this in the comment. Fine to leave; worth knowing.

### What the rebase surfaced — the guard caught three new sites

`dev` moved 25 commits under this branch and landed **three new `set(..., nx=True, …)` sites** while it was open. `test_1920_no_hand_rolled_single_flight` failed on the merge and named all three. That is the anti-recurrence claim being demonstrated within days of being written, which is a better argument for the guard than the guard's own docstring. Resolved as:

- **`system_seed_service` (#2215's `first_run_seed:provision` pass lock) → ADOPTED.** It is in `_ADOPTED_MUST_BE_CLEAN`, and a hand-rolled lock sitting in the very file whose consolidation this PR claims would make that claim false. Its 3-state contract maps exactly onto the primitive, so this was a direct swap.
- **`db_backup_service` (#2216's lease) → ALLOWLISTED, consolidation-only.** It already mints a unique token and releases via `lock_token_matches` compare-and-delete, so it is *not* the #1919 bug class — it is stylistic divergence. Converting a day-old feature without its author, inside a merge resolution, is the wrong place for that.
- **`docker_service` (#2215's `port_alloc:{port}`) → ALLOWLISTED permanently, and this one should never be adopted.** It is an allocation over a keyspace of many keys, not a mutex: no release, no token, TTL-expiry only, and it deliberately *propagates* Redis errors so the caller decides. `SingleFlightLock` would be actively wrong there — its internal fail-open returns `True` on a Redis error, which a port reservation reads as "reserved", handing two agents the same SSH port. That is precisely the collision #2215 exists to fix. The reason is written into the allowlist entry so nobody "finishes the consolidation" later.

I also ported `test_pass_lock_release_leaves_a_foreign_token_alone`, which called the removed `_release_pass_lock` directly. The port is **stronger than the original**: the old test never acquired, so `release_if_owned` would no-op before it ever compared, and the assertion would have passed against a blind delete. It now acquires for real, loses the key to a sibling, then releases. Mutation-checked — replacing the compare-and-delete with a blind `delete` makes it fail, and restoring makes it pass.

### On the red `pytest (head, seed 12345)`

Not a regression and not a hang. The job was **cancelled** at 25m15s against a 25-minute cap while the other five shards passed in ~11–12 min. Since #2021 this suite runs `--timeout=300 --timeout-method=signal` specifically so a stalled test names itself as an ordinary failure instead of eating the job budget, and that flag was present on the run — so "a test hung" is ruled out by construction, leaving the runner variance #2019 documents.

I did not want to assume that on a PR that rewrites locking, so I ran the full unit suite locally at the exact failing seed: **11722 passed, 8 failed, 8m05s, no hang.** All 8 failures reproduce identically on plain `origin/dev` (`test_pat_propagation_properties::TestKnownGaps` ×3 — self-describing — and `test_ent96_timeline_split` ×5, a pre-existing dev failure worth its own issue, unrelated to this PR).

Approving.

vybe:
Merging with `pytest (head, seed 99999)` red. Recording the evidence, because this PR rewrites locking and "it's just the flake" is exactly the sentence that ships a hang.

**It is a cancellation, not a failure.** `"Run unit suite"` → `conclusion: cancelled` at 25m15s against a 25-minute cap, with `Upload JUnit XML` failing downstream because there was nothing to upload. The other five shards on the same run passed in ~11–12 min.

**A hang is ruled out by construction.** Since #2021 this suite runs `--timeout=300 --timeout-method=signal` precisely so a stalled test raises inside itself, the suite *continues*, and the diff job reports it as an ordinary named failure. That flag was present on this run. A test that hung would have named itself; nothing did.

**Verified rather than assumed.** I ran the full unit suite locally on this exact branch head at **both** the seeds CI flagged across these PRs:

| run | result |
|---|---|
| seed 12345 | 11722 passed, 8 failed, **8m05s**, no hang |
| seed 99999 | 11722 passed, 8 failed, **8m13s**, no hang |

Identical failure sets, and all 8 reproduce on a clean `origin/dev` checkout — `test_pat_propagation_properties::TestKnownGaps` ×3 (self-describing) and `test_ent96_timeline_split` ×5, which I filed separately as #2247 (a stale date fixture that has been red on `dev` since 2026-08-15, invisible because the base-vs-head diff cancels it out on PRs and does not run at all on pushes).

**`regression diff` is green and that green is real.** Its quorum rule (#1910) tolerates one lost seed only while that side retains at least one usable JUnit; a whole missing side stays fatal, and its own self-test case 10 pins that quorum must not mask a genuine regression. Head had 2 of 3 usable, base had 3 of 3.

So: base 3/3 green, head 2/3 green plus two clean local full-suite runs at the failing seed, no required check red.

PR #2233, opened 2026-08-16
fix(portal): the Workspace turn bound is the agent's own execution timeout (#2214)

### LINKED ISSUE #2214: bug: Workspace chat turns are hard-capped at 300s, ignoring the agent's configured execution timeout
## Summary

A Workspace chat turn is hard-capped at 300 seconds and dies there. The cap is a module constant with no setting, no env var and no per-agent override — and it silently overrides the agent's own configured execution timeout, which defaults to 3600s and is operator-settable from 60–7200s.

## Context

Reported from live use of `/workspace`: longer turns die after ~5 minutes and there is no way to regulate it.

`src/backend/client_portal/service.py`:

```python
PORTAL_TURN_TIMEOUT_SECONDS = 300
```

That constant is passed verbatim as `timeout_seconds=` into `run_resumable_turn` for every Workspace turn. Meanwhile the platform already has a first-class per-agent bound — `agent_ownership.execution_timeout_seconds` (TIMEOUT-001, default 3600s, `GET`/`PUT /api/agents/{name}/timeout`, range 60–7200) — which the Workspace path never consults. So an operator who deliberately raised an agent's timeout to an hour still gets 5 minutes in the surface people actually work in, and nothing tells them why.

Two things to keep intact while changing the number, both load-bearing and both documented in that file:

- **The derived bounds must move with it.** `PORTAL_ATTEMPT_CEILING_SECONDS` = turn timeout + 10 (HTTP slack) + 300 (the #678 reader-race auto-retry's second call), and `PORTAL_MAX_TURN_SECONDS` = 2 × attempt ceiling + 60 (a resume whose JSONL is gone re-runs the whole turn cold). `PORTAL_INFLIGHT_TTL_SECONDS` is sized off that. Raising the turn timeout without these moving reintroduces the #2120/#2133 failure: the in-flight marker expires while a live, already-billed turn is still running, and the client is told nothing is running.
- **The frontend's own wait ceiling is derived from the same constants** (`REPLY_MAX_WAIT_MS_FALLBACK` in `src/frontend/src/components/portal/portalUtils.js`), so the client must not keep waiting on a shorter or longer clock than the server enforces.

Raising the default alone is not the whole fix — an operator with a long-running agent needs the bound to follow that agent, and a client watching a 20-minute turn needs the surface to stay honest about it rather than presenting a dead spinner.

## Acceptance Criteria

- [ ] The Workspace turn timeout is materially larger than 300s by default, and the value is justified in the code comment rather than picked arbitrarily
- [ ] The bound respects the agent's configured `execution_timeout_seconds` rather than a constant that silently overrides it (clamped to a sane Workspace ceiling if one is wanted)
- [ ] `PORTAL_ATTEMPT_CEILING_SECONDS`, `PORTAL_MAX_TURN_SECONDS` and `PORTAL_INFLIGHT_TTL_SECONDS` stay derived from the turn bound — no constant left sized against the old 300s
- [ ] The client's wait ceiling stays derived from the server's bound (no independent number to drift)
- [ ] A turn that legitimately runs past the old cap completes and renders instead of dying; the existing `tests/unit/test_2133_*` pins (or their successors) still hold against the real constants
- [ ] A turn that does hit the new bound fails with an honest, user-readable reason rather than a silent stop
- [ ] If the bound becomes configurable, it is surfaced where operators already set the agent timeout — not `.env`-only — with a named validation error on out-of-range input

## Technical Notes

- Backend: `src/backend/client_portal/service.py` (`PORTAL_TURN_TIMEOUT_SECONDS` and the derived ceiling/marker constants; the `run_resumable_turn` call site), `src/backend/services/session_turn_service.py`
- Existing per-agent bound: `agent_ownership.execution_timeout_seconds` (TIMEOUT-001, #665), `GET`/`PUT /api/agents/{name}/timeout`, and the #929 rule that a schedule's `timeout_seconds` may not exceed the agent cap — the same relationship applies here
- Frontend: `src/frontend/src/components/portal/portalUtils.js` (`REPLY_MAX_WAIT_MS_FALLBACK`, `PORTAL_ATTEMPT_CEILING_S`)
- Rooms share the surface — check whether the room turn path carries the same constant

trinity-ability:
Fixes #2214

## What changed and why

Every Workspace turn was hard-capped at `PORTAL_TURN_TIMEOUT_SECONDS = 300`, while the platform's own per-agent knob — `agent_ownership.execution_timeout_seconds` (TIMEOUT-001/#665; default 3600, range 60–7200, `PUT /api/agents/{name}/timeout`) — was never consulted on this path. The asymmetry was inside one module: the resume **lock** already resolved per agent (`session_turn_service.resolve_lock_ttl`), while the **turn** it guards did not. An operator who raised an agent to an hour still got 5 minutes, silently, on the surface clients actually use.

**The turn bound is now a per-turn resolution of the agent's own timeout.**

- **New `resolve_turn_timeout` in `services/session_turn_service.py`, beside `resolve_lock_ttl`** — the fix lives where the asymmetry does. Read-side clamped to TIMEOUT-001's own range `[60, 7200]` (the #506 stray-row pattern), fail-open to the platform **default** 3600. Deliberately distinct from `resolve_lock_ttl`, which falls back to its **cap**: over-TTL on an auto-expiring key is harmless, whereas an over-long turn is billable work. Placing it in the engine also makes the enterprise-rooms follow-up an `import` + delete-a-constant instead of a cloned clamp/fallback.
- **The three derived module constants become pure functions** — `portal_attempt_ceiling_seconds(t) = t + 10 + _AUTO_RETRY_MAX_TIMEOUT_S` (imported, never copied) and `portal_max_turn_seconds(t) = 2 × ceiling + 60` (the cold retry re-runs the whole turn). The old names (`PORTAL_TURN_TIMEOUT_SECONDS`, `PORTAL_ATTEMPT_CEILING_SECONDS`, `PORTAL_MAX_TURN_SECONDS`, `PORTAL_INFLIGHT_TTL_SECONDS`) are **deleted, not aliased**, so any missed def-time consumer fails loudly at import rather than silently keeping the 300s arithmetic. `mark_turn_inflight`'s TTL keyword default — bound at *definition* time, which is exactly where a per-agent value cannot live — became a `None` sentinel resolved at call time.
- **`start_portal_turn` resolves ONCE** and threads that one value to the marker TTL, the 202's `wait_budget_seconds`, and the dispatch. Marker, client budget and actual turn cannot disagree, and the mid-turn `PUT /timeout` race between independent reads is gone. (Honest scope: the resume *lock* still resolves its own TTL at acquire — pre-existing, engine-owned; see follow-ups.)
- **Honest 504** — the timeout message names the bound that was enforced, with honest rounding (below 120s it speaks in seconds, so a 90s bound is never reported as "1-minute").

### Plus the gap found while planning

`reattach()` passed **no budget** to `awaitPersistedReply`, so `REPLY_MAX_WAIT_MS_FALLBACK` (1280s) was the *live* ceiling for every reloaded client on every backend — not just old ones. Once the server bound is per-agent, a client reloading into a 2-hour turn would give up at ~21 minutes while the turn is alive and billed. Closed by returning the marker's **remaining** Redis TTL as `in_flight_wait_budget_seconds` on the history response — the fetch a mounting client already makes, so reattach costs no extra round trip. Remaining TTL rather than a fresh full budget, because the budget was fixed at dispatch and recomputing would over-wait by the elapsed time; `loadThread` stamps `budgetReadAt` **beside the fetch**, since the remainder is only honest from its own read time. TTL mapping: `-2` (marker vanished between GET and TTL) → report nothing in flight; `-1` (unexpected, no expiry) → fail open to the full per-agent budget; exception → `None`, client falls back.

This needed the `client_portal/models.py` field: the route carries `response_model=PortalHistory`, which **strips undeclared fields**, so without the declaration the budget would silently never leave the server.

## AC #2 — the ceiling decision, stated explicitly

**The full 60–7200 range is respected with NO extra Workspace clamp. The clamp IS the TIMEOUT-001 range.**

AC #2's operative clause is *"respects the agent's configured `execution_timeout_seconds`"*; a clamp is optional ("if one is wanted"). Any Workspace clamp below 7200 re-introduces the silent-override bug for exactly the upper half of the range TIMEOUT-001 sells — so the decision is to add nothing tighter than the platform's own range.

The old `test_2133` `< 3600` marker-TTL pin encoded a **real** hazard and is not being waved away: a hard-killed backend leaves the composer dead for the marker's remaining life — ~4.2h at the 7200 cap (`portal_max_turn_seconds(7200)` = 15,080s). That is **accepted and documented**, on these grounds:

- **Exposure requires a HARD kill.** Graceful shutdown cancels the background task and the `finally` clears the marker during unwind; only SIGKILL/OOM/host-loss orphans it.
- **Blast radius is one thread's composer** on one agent — "+ New chat" and every other thread stay live. After a hard kill the agent-side work often continues, so "working" is then true; only delivery died.
- **Precedent**: the Session surface's own in-flight sentinel has run unclamped at `min(timeout+30, 7230)` ≈ 2h with identical orphan characteristics.
- **Operator recovery**: `DEL portal_inflight:{session}` — the same manual-DEL escape the engine already documents for its lock keys. Recorded in requirements and the feature flow.

**Mitigations weighed and rejected:**
- *Marker heartbeat* (short TTL + per-turn refresher) — a second task per turn plus refresh-vs-compare-and-delete races in a file two open PRs are touching, and it decouples the marker TTL from the client budget the whole #2133 design keys on. Complexity for a hard-kill-only case.
- *Row-terminal self-heal* — **UNSAFE**, not merely costly: the cold retry runs under a **new** execution id while the marker still names the original (now FAILED) row, so clearing on the original's terminal declares a live, billed retry "not running" — precisely the #2120 double-spend the marker exists to prevent. A boot-epoch discriminator would fix it but needs multi-worker-safe boot identity (prod runs `--workers 2`); listed as a follow-up, not smuggled in.

**Successor tests pin the DERIVED ceiling + the resolver clamp**, not an arbitrary hour literal: `test_the_bound_is_nowhere_near_the_old_hour` → `test_the_marker_is_bounded_by_the_agent_timeout_range` (999999→7200, 10→60, raise→3600; `portal_max_turn_seconds(7200) == 15080`). The pin moves from a magic number to a bound derived from the platform's own range — the "no independent number to drift" spirit, kept. AC #5 explicitly anticipates successors.

## ✅ RESOLVED: the merge-second collision, and the trap it hid

**Both #2219 (issue #2196) and #2222 (issue #2198) merged into `dev` while this branch was in flight**, moving `dev` from `24309246` to `6bd7c179` and putting this PR into `CONFLICTING`. Resolved with a **merge commit** (`c0a0dd48`), not a rebase — the branch already had an open PR, and siblings #2235/#2236 resolved the same class of collision the same way.

**Textual conflicts — 6 hunks, all "both sides added something adjacent", resolved as unions with neither side's behaviour changed:**

| File | Hunks | Resolution |
|---|---|---|
| `src/backend/client_portal/service.py` | 3 | The signature carries **both** kwargs. #2219's `availability` + its `_turn_failed_detail(availability)` **502** and #2214's `turn_timeout_seconds` + its named-limit **504** are orthogonal failure modes — one names *why the turn could not run*, the other *what bound killed it* — so both survive. `start_portal_turn` threads both from its single per-turn resolution. |
| `tests/unit/test_ent286_portal_streaming.py` | 2 | Both `portal_chat` fakes gain both kwargs. |
| `docs/memory/learnings.md` | 1 | Both entries kept, chronological, dev's first. |

**The dangerous part had NO textual conflict — and was worse than predicted.** #2219 moved the availability gate to the async `_agent_availability` (all three live gates: `:477`, `:1143` inside `portal_chat`, `:1707`) but left **`_agent_is_running` defined and uncalled** at `:1646`. This branch's `test_2133` fixtures stubbed only that dead name. Because the attribute still exists, `monkeypatch.setattr` **does not raise** — it silently patches dead code while the real async gate reaches Docker from a unit test. A green run is not evidence here; that is exactly the failure mode that hides.

**Fixed, and verified live:**

1. Both the `portal` and `real_chat` fixtures now stub `_agent_availability` with a correctly-shaped `async def _ready(name)` returning `"ready"`.
2. `_fake_chat` absorbs `availability=None`, matching the merged `portal_chat(..., turn_timeout_seconds=None, availability=None)`.
3. **Proof the stubs are load-bearing rather than decorative — falsification, not a green run:** flipping both stubs to return `"unavailable"` turns **4 tests red**, split across *both* fixtures (`portal`: `test_the_marker_ttl_is_that_bound`, `test_resolver_fails_open_to_the_platform_default`; `real_chat`: `test_the_turn_is_dispatched_with_the_timeout_the_bound_is_built_from`, `test_timeout_error_names_the_agents_bound`). Restoring `"ready"` returns all green. Had the fixtures still been patching dead code, breaking them would have changed nothing.

**Post-merge test evidence** (`-p no:randomly` throughout): `test_2133` + `test_ent286` + `test_ent358` → **59 passed** (55 pre-merge; `dev` contributed 4). The code merged with: `test_2196_roster_availability` + `test_2198_dashboard_settled` + `test_2198_portal_batch_sessions` → **58 passed**. Frontend `portalSidebarIA.spec.js` → **32 passed**. All 25 CI checks green.

**Residual, deliberately not fixed here:** `_agent_is_running` remains dead code on `dev`. Any future test that patches it will silently patch nothing, exactly as this one did. Deleting it belongs to #2196, not to this PR.

## Verification evidence

- **`/verify-local --skip-agent`** (correct mode — no `docker/base-image` change in this diff): **PASS**. Build + import-smoke OK, boot + health OK, integration **70 passed / 13 skipped / 2 deselected**.
- **Frontend vitest run separately: 588 passed / 31 files.** Stated explicitly because `verify-local`'s unit stage is **pytest-only** — vitest is a blind spot in that gate, and this PR changes three frontend files.
- **Backend neighbourhood** (`-p no:randomly`): `test_2133_bounded_reply_poll.py` + `test_ent286_portal_streaming.py` + `test_ent358_workspace_absorbs_session.py` → **55 passed**. `test_ent358` stays green with **zero edits** (the engine change is purely additive); `test_ent286` touched minimally — the two `portal_chat` fakes gain `turn_timeout_seconds=None`, which the new kwarg would otherwise `TypeError` on inside `_run()`.
- **Frontend neighbourhood**: `portalSidebarIA.spec.js` → **32 passed**.
- **e2e watch item CLEARED**: both #2221 specs read in full — `workspace-absorbs-session.spec.js` is pure routing assertions; `portal-agent-page-overview.spec.js` is pure layout geometry (its only timeout is a 20s page-load wait on an Overview that never dispatches a turn). A grep across all 28 e2e specs found **no turn-budget assumption**.
- **`/cso --diff`**: no findings at the ≥8 confidence gate — report in `docs/security-reports/cso-diff-2026-08-16-2214-portal-turn-bound.md`, with refutation cites for each of the seven load-bearing claims.

### Pre-existing reds disclosure

The unit suite at dev tip carries **8 failures that are not from this PR**:
- 5 × `tests/unit/test_ent96_timeline_split.py` — hardcoded `2026-08-14` bucket keys (date rot).
- 3 × `tests/unit/test_pat_propagation_properties.py::TestKnownGaps` — stale strict-`xfail` after #2017.

Both predate this branch and are untouched by it. (`pytest-randomly` is installed, so all runs above used `-p no:randomly`.)

## Follow-ups (listed, deliberately NOT filed)

1. **Enterprise rooms carry the same bug** — `ROOM_TURN_TIMEOUT_SECONDS = 300` in `enterprise/backend/shared_sessions/service.py` (`:48` constant, `:631` TTL derivation, `:745` dispatch). Same class; after this PR it becomes `import resolve_turn_timeout` + delete-a-constant. **Enterprise tracker**; the submodule is not mounted in this worktree, so this OSS PR must not touch it.
2. **PRIORITY — engine lock-TTL sizing.** `resolve_lock_ttl = min(t+30, 7230)` is shorter than one worst-case attempt (`t + 310`, because the #678 auto-retry is an *additive* second HTTP call) for any `t ≥ 280`. The lock therefore lapses **mid-attempt**, and a headless `POST .../chat` re-send after the lapse can acquire it and run a concurrent `--resume` against the same JSONL — the corruption the lock exists to prevent. Pre-existing and engine-owned, but **this PR widens the exposed population** from ~zero (300s turns always fit) to every long turn on the mainline surface. Size the lock TTL from the attempt ceiling, not one bare timeout.
3. **Marker boot-epoch self-heal** — would make the row-terminal self-heal rejected above safe; needs multi-worker-safe boot identity.
4. **Dev-tip test rot** — the ent#96 hardcoded dates and the #2017 stale strict-xfail above.
5. **Feature-flows index bloat** — noted during the docs pass.

## Rollout note

**Proxy read-timeout must cover BOTH paths** — worth an eyeball before/at deploy. The SSE stream emits events during a turn (traffic acts as keepalive, low risk), but the synchronous `POST .../chat` — the documented ent#83 headless surface and the frontend's no-JS-timeout fallback — holds a **byte-silent** HTTP response for the whole turn, now up to 7200s against an nginx default `proxy_read_timeout` of 60s. That converts an honest 504-with-message into a connection reset. No code change here; the requirements delta steers long-timeout headless integrators to the streaming route.

## Scope

No migration, no feature flag, no env var, no new endpoint, no `docker/base-image` change. All new fields optional: old FE + new BE → the 202 budget has been consumed since #2133 (now per-agent, i.e. a longer honest wait) and reattach falls back exactly as today; new FE + old BE → the frozen fallback literal matches the old backend's real bound by construction.

Docs landed with the change: `requirements/core-agent.md` §5.9, one sentence in `architecture.md` ("Streaming a turn"), and the "Turn bound & wait budgets" section in `feature-flows/workspace-absorbs-session.md`.


vybe (approved):
Validated per /validate-pr. **I resolved the conflicts and the non-textual hazard on this branch** (merge `c0a0dd48`) — details below, please read that commit rather than assuming a mechanical rebase.

## The fix

`resolve_turn_timeout` lands beside `resolve_lock_ttl`, which is where the asymmetry lived — the lock already resolved per agent and the turn it guards did not. The two resolvers are deliberately **not** unified and the reason is right: an over-TTL on an auto-expiring key is harmless, whereas an over-long turn is billable work, so the lock falls back to its **cap** and the turn to the platform **default**. Read-side clamp to `[60, 7200]` follows the #506 stray-row pattern.

Deleting `PORTAL_TURN_TIMEOUT_SECONDS` et al. rather than aliasing them is the correct call — a def-time keyword default is a mirror with no owner, and the `None` sentinel on `mark_turn_inflight` is the only way a per-agent value can reach it. `start_portal_turn` resolving **once** and threading that value to marker TTL, 202 budget and dispatch closes the mid-turn `PUT /timeout` race between independent reads.

The `reattach()` gap found while planning is the more valuable half: `REPLY_MAX_WAIT_MS_FALLBACK` was the live ceiling, not a fallback, and returning the marker's **remaining** TTL (not a fresh full budget) is the honest number. The `PortalHistory` field declaration is load-bearing — `response_model` strips undeclared fields.

## AC #2 ceiling decision — accepted

No extra Workspace clamp; the clamp is the TIMEOUT-001 range. Any tighter bound re-introduces the silent-override bug for the upper half of the range the knob sells. The ~4.2h orphaned-marker exposure at the 7200 cap is accepted on the stated grounds: hard-kill only, blast radius one thread's composer, precedent in the Session sentinel, `DEL portal_inflight:{session}` recovery documented. The rejection of row-terminal self-heal is correct and not merely costly — the cold retry runs under a new execution id, so clearing on the original's terminal would declare a live billed retry "not running" (#2120).

## Conflict resolution (mine)

Three textual hunks, all unions, neither side's behaviour changed:
- `client_portal/service.py` — #2219's `availability` and #2214's `turn_timeout_seconds` are orthogonal (one names *why the turn could not run*, the other *what bound killed it*), so the signature carries both, the 504 stays #2214's and the 502 stays #2196's `_turn_failed_detail(availability)`.
- `test_ent286_portal_streaming.py` — both fakes take both kwargs.
- `learnings.md` — empty base, both appended; chronological with dev's entries first.

**The hazard the PR body flagged was real and I confirmed it**: `_agent_is_running` is still defined at `service.py:1646` with **zero callers** fleet-wide, so `monkeypatch.setattr` on it does not raise — `test_2133`'s `portal` fixture was silently patching dead code while the real async gate reached Docker from a unit test, and `real_chat` never stubbed it at all while driving the real `portal_chat`. Both fixtures now stub `_agent_availability`; `_fake_chat` takes `availability`.

## Verification

- `test_2133` + `test_ent286` + `test_ent358` → **59 passed** post-merge (55 pre-merge; dev contributed 4)
- `-k "portal or 2196 or 2198 or client_portal or 2214 or 2133"` → **335 passed, 1 skipped**
- Frontend `vitest run` → **654 passed / 35 files** (verify-local's unit stage is pytest-only, so this was run separately)
- CI green on the merge commit: all six pytest legs, `regression diff`, `e2e`, `prod-image-smoke`, `verify-non-root`, `gitleaks`
- Scan: no secrets/host-paths/mode-flips; no new env var, no new top-level package, no schema change

Follow-ups the PR names honestly and does not smuggle in (resume-lock TTL still self-resolving; boot-epoch discriminator for marker self-heal) are the right things to defer.

PR #2234, opened 2026-08-16
fix(security): mcp_validator refuses RFC 6598 CGNAT space, plain and IPv4-mapped (ent#394)

trinity-ability:
Fixes Abilityai/trinity-enterprise#394

> **Cross-repo note**: the issue lives on the private enterprise tracker; the code ships here. GitHub auto-closes **same-repo only**, so this keyword does **not** close ent#394 — it is closed **manually at release**. Live precedent: PR #2179 says `Fixes …#393` and ent#393 is still open.

## What changed, and why

`services/mcp_validator._resolves_to_private_ip` asked CPython's `ipaddress` six questions — `is_private`, `is_loopback`, `is_link_local`, `is_multicast`, `is_reserved`, `is_unspecified` — and **none of them covers RFC 6598** (`100.64.0.0/10`, CGNAT). The stdlib reports that range as neither `is_private` **nor** `is_reserved` (the fact that made ent#14 S3 add the clause to `utils/url_validation` in the first place), so an `.mcp.json` naming an http/sse server whose host **is** — or **resolves into** — CGNAT space passed SSRF validation, in **both** plain-v4 and IPv4-mapped (`::ffff:100.64.0.1`) form.

This is not the ent#393 defect (a version-*gated* clause that never ran for the mapped form). Here the clause was simply **absent** — the second instance of the durable bug class recorded in `docs/memory/learnings.md` (2026-08-14): *in a predicate stack that borrows N checks from a library and hand-rolls one, the hand-rolled one is where input normalisation drifts.* Not re-derived here; see that entry.

The fix is one module constant plus the merged #2179 v4-**view** idiom:

```python
_SHARED_ADDRESS_SPACE = ipaddress.ip_network("100.64.0.0/10")
...
v4 = ip if ip.version == 4 else ip.ipv4_mapped
if (... or (v4 is not None and v4 in _SHARED_ADDRESS_SPACE)):
```

The membership test is **never version-gated** — `ip.version == 4` appears only where it *selects the view*, never where it gates the check. The mapped form reports version 6, and the six borrowed stdlib predicates delegate through `ipv4_mapped` themselves, so CGNAT — the one range whose refusal is our own — must do the delegation too.

Applied to **both** copies via `cp` canonical → vendored (Invariant #5); `shasum` of `src/backend/services/mcp_validator.py` and `docker/base-image/agent_server/mcp_validator.py` both `9539937fd1db5243e5c8173dea46db2ea67dcb19`.

**Consumers inheriting the fix with zero call-site change**: `routers/credentials.py:392` (the `POST /credentials/inject` guard, #590/#598) and `docker/base-image/agent_server/mcp_template.py:215` (the #2007 boot-time `.mcp.json.template` renderer, which validates every declared server on every container start).

## Scope discipline

Deliberately **untouched**: the refusal message (`"resolves to a private/loopback/link-local address (SSRF guard)"` — caller-visible API under the #598 verbatim-surface contract, matched by three existing tests); the `except socket.gaierror: return True` exception contract (`mcp_template.py` wraps `render_server` in a broad `except Exception` *because* a `UnicodeError` can escape `getaddrinfo` — that division of labour is deliberate); the `except (ValueError, IndexError): continue` sockaddr tolerance; `_HttpValidator` / `_SseValidator`; and `utils/url_validation.py`. No new imports — the vendored copy stays stdlib-only, which is why the `/10` constant is duplicated rather than imported.

## Test evidence

**50 new rows** in `tests/unit/test_mcp_validator.py`:

`TestCgnatSsrfGuard` (11 tests) — plain-v4 literals at both `/10` edges; IPv4-**mapped** literals; **both widening guards** including the exact first-address-above boundary `100.128.0.0` (not `.1`, which leaves the real boundary untested) and `100.63.255.255` below; a public-IPv6 row pinning the `v4 is not None` **None-branch** (one "simplification" away from an `AttributeError` on every public-v6 endpoint — the suite previously had zero v6 rows); a **positive control** asserting the `getaddrinfo` mock was actually consulted (without it the hostname rows pass vacuously, since an NXDOMAIN host fails *closed* and is green with the clause absent); hostname-via-`getaddrinfo` rows for both AF_INET and AF_INET6 sockaddrs; mixed-resolution any-match semantics; and interpreter pins for NAT64 / IPv4-compatible / 6to4 (refused by the stdlib's own tables — pinned so a CPython change can't silently reopen them, the #1891 class). The class deliberately never uses `_public_dns_patch`, which stubs the predicate itself and would make every acceptance row vacuous.

`TestCrossValidatorAgreement` — **retires the drift class that caused this bug.** It asserts that `mcp_validator._resolves_to_private_ip` and `url_validation._is_internal_address` agree over one address corpus (every internal family, both `/10` edges, mapped forms, transition forms, publics). The two stacks share one membership *policy* behind deliberately different DNS wrappers (fail-closed here, fail-open there — the wrapper stays out of the corpus). Before this, nothing fired when a range landed in one stack only, which is exactly how ent#394 happened. A range added to `url_validation` and not here now turns CI red.

**Red-check**: 13 of the new tests fail against the unfixed validator. **Mutation check**: re-introducing the ent#393 version-gated form turns 5 rows red.

```
tests/unit/test_mcp_validator.py tests/unit/test_2007_mcp_template_render.py -q -p no:randomly
→ 188 passed
```

(`pytest-randomly` is installed in this repo's test venv — always pass `-p no:randomly`.)

## Verification evidence

`/verify-local` ran in **FULL** mode by its own rule (the branch touches `docker/base-image/**`). Branch verdict **PASS**:

- Stages green: build + import-smoke 8s · **agent-build + import-smoke 3s** · boot + health 14s. A `--skip-agent` run reached the integration suite: **70 passed / 13 skipped / 2 deselected**.
- **The built base image ships the fix**: `/app/agent_server/mcp_validator.py` inside `trinity-agent-base:verify-*` hashes to `9539937f…`, identical to the worktree copy, and contains `_SHARED_ADDRESS_SPACE` — no stale layer.
- **In-image predicate**: `100.64.0.1` → True · `::ffff:100.64.0.1` → True · `100.63.255.255` → False · `100.128.0.0` → False.
- **End-to-end in a real booted agent** (the #1929 withheld-with-reason contract), running exactly what `startup.sh` invokes:
  ```
  [mcp-template] withheld MCP server 'cgnat-probe': Server 'cgnat-probe': url hostname
  '100.64.0.1' resolves to a private/loopback/link-local address (SSRF guard)
  [mcp-template] rendered MCP server(s) into .mcp.json: public-probe
  ```
  The resulting `.mcp.json` holds `trinity` (token intact) + `public-probe`, and **not** `cgnat-probe` — one bad server never costs the good ones.
- **Negative control**: the pre-fix `origin/dev` validator, in the *same* container, answers False for both CGNAT forms and would have installed it.
- Three-way parity confirmed: worktree = base image = backend image; the app's own `from services.mcp_validator import …` passes all four addresses.

The harness `result.json` reported fail@agent-exercise; that was root-caused (reproduced 4×, including on a fresh VM) to a **Docker Desktop 4.36.0 / Engine 27.3.1** bug — `docker network connect` on a running container with an ephemeral `HostPort:"0"` silently drops the host binding (`com.docker.backend.gvisor … write unixgram: destination address required`). The application was healthy in-container throughout.

## Release notes

- **Agent-side enforcement lags the base image.** The `.mcp.json.template` renderer picks this up only after `./scripts/deploy/build-base-image.sh` and an agent recreate — adopted per agent on its next **cold** start (#1809/#1816). The backend inject path enforces immediately on backend redeploy.
- **MCP servers addressed in RFC 6598 CGNAT space — including Tailscale node IPs (`100.64/10`) — are now refused by `.mcp.json` validation. This is by design, with no escape hatch**, matching the posture `url_validation` took in ent#14 S3 and the A2A endpoint registry inherited. Supported shapes are a public HTTPS front or MagicDNS names that resolve publicly.

Already-rendered `.mcp.json` entries are not retro-scrubbed (the renderer is merge-only-missing); the next full validation touchpoint — an owner re-inject — refuses them.

## Schema

No schema change → **no Alembic revision, and `schema-parity` is N/A here.** Stated explicitly so reviewers don't go looking.

## Docs

`docs/memory/architecture.md` — Invariant #5's enumerated byte-identical-mirror list now names `mcp_validator.py` (#2007). It was omitted while the #2007 section of the same document already called it an Invariant #5 mirror.

`docs/memory/feature-flows/credential-injection.md` — the SSRF-guard scope line now names CGNAT (plain + IPv4-mapped).

## Pre-existing reds (not from this PR)

`dev` tip currently carries 8 unit failures: 5× `tests/unit/test_ent96_timeline_split.py` (hardcoded dates) and 3× `tests/unit/test_pat_propagation_properties.py::TestKnownGaps` (stale strict-xfail, #2017). Both predate this branch and are untouched by it.

1. The Docker Desktop 4.36.0 agent-exercise blocker above should be filed against the `verify-local` skill, so a future FULL run doesn't read it as a branch regression.
2. The 8 dev-tip test failures (ent#96 hardcoded dates; #2017 stale strict-xfail).
3. Optional: a one-line "second instance: ent#394" appendix under the `learnings.md` 2026-08-14 entry (the class itself is already recorded; non-blocking).

## Post-merge human action

```
gh issue edit 394 --repo Abilityai/trinity-enterprise \
  --add-label status-in-dev --remove-label status-in-progress
```

and close ent#394 **manually at release** — the cross-repo closing keyword above will not do it.


vybe (approved):
Validated per /validate-pr.

**Security**: byte-identical vendored parity confirmed (`9539937f` both copies, Invariant #5). The `v4 = ip if ip.version == 4 else ip.ipv4_mapped` view idiom is correct — the version check selects the view, never gates the membership test, which is exactly the ent#393 defect this avoids. No new imports; vendored copy stays stdlib-only.

**Scan**: no secrets/keys/host-paths/mode-flips. The three public-looking IPs in the diff (`100.0.0.0`, `198.18.0.1`, `240.0.0.1`) are boundary fixtures inside `tests/unit/test_mcp_validator.py` — not findings.

**Packaging**: no new top-level `src/backend/` package, no new env var, no schema change → schema-parity/Alembic correctly N/A.

**CI**: all six pytest legs (base+head × 3 seeds) green, plus `prod-image-smoke`, `verify-non-root`, `regression diff`, `gitleaks`, both CodeQL analyses.

**Tests**: `TestCrossValidatorAgreement` is the right structural addition — it retires the drift class that produced this bug rather than only patching the instance.

**Cross-tracker note**: `Fixes Abilityai/trinity-enterprise#394` will NOT auto-close (same-repo only) — ent#394 needs a manual close + `status-in-dev` bump at release.

Issue #1486, opened 2026-07-06
OPERATOR_INTAKE_ENABLED / DO_NOT_TRACK privacy kill switch is inert (not forwarded by compose)


vybe:
## Summary

The privacy kill switch for the first-run operator intake is inert: `.env.example` (lines 60–70) promises *"Set to false (or DO_NOT_TRACK=1) to fully disable the outbound submission for air-gapped / privacy-strict installs"*, but neither `docker-compose.yml` nor `docker-compose.prod.yml` forwards `OPERATOR_INTAKE_ENABLED` or `DO_NOT_TRACK` to the backend container. Setting either in `.env` silently does nothing.

## Context

Found by the 2026-07-06 automated `/validate-config` run (umbrella: #1485). Same packaging-gap class as #1039/#1056: a documented config lever that never reaches the container because compose lacks the `environment:` entry (compose reads `.env` only for `${VAR}` interpolation — container env comes solely from explicit `environment:` entries).

Both vars are read at `src/backend/config.py:33-35`:

```python
OPERATOR_INTAKE_ENABLED = (
    os.getenv("OPERATOR_INTAKE_ENABLED", "true").lower() == "true"
    and os.getenv("DO_NOT_TRACK", "0") not in ("1", "true", "True")
)
```

The setup-consent checkbox remains the primary gate (the intake POST only fires on affirmative opt-in), but the env-level kill switch is explicitly advertised for air-gapped installs and is broken. `OPERATOR_INTAKE_URL` (also read in `config.py`) has the same gap.

## Acceptance Criteria

- [ ] `OPERATOR_INTAKE_ENABLED` and `DO_NOT_TRACK` are forwarded in the backend `environment:` block of **both** `docker-compose.yml` and `docker-compose.prod.yml` (defaults `true` / `0`)
- [ ] `OPERATOR_INTAKE_URL` is forwarded the same way (default to the code default)
- [ ] Setting `OPERATOR_INTAKE_ENABLED=false` in `.env` verifiably disables the intake POST inside the running backend container (e.g. `docker exec trinity-backend env | grep OPERATOR_INTAKE`)

## Technical Notes

- Read sites: `src/backend/config.py:33-40`, consumed by `services/operator_intake_service.py`
- Two-line change per compose file, mirroring the existing `PUBLIC_CHAT_URL`-style forwarding comments ("#1039 packaging-gap class")

PR #2236, opened 2026-08-16
fix(backup): automatic database backups, both backends, zero setup (#2216)

### LINKED ISSUE #2216: bug: shipped database backup tooling is never invoked — no scheduled backups, no recovery point

trinity-ability:
Fixes #2216

## What changed

Database backups now happen by themselves, on both backends, with no operator setup — a daily job at 03:30 UTC plus a copy taken at boot whenever migrations are pending, under `/data/backups/`, bounded by a retention window with a floor, and loud when they fail.

- **New leaf `src/backend/db/backup_primitives.py`** (stdlib-only, no `database`/services import — it is called from `init_database()` at import time): the SQLite copy via `sqlite3.Connection.backup()`, verification, day-keyed naming, pattern-scoped prune, aged-tmp sweep, and the boot pre-migration hook.
- **New `src/backend/services/db_backup_service.py`**: the daily job (backend lifespan, beside `db_vacuum`), the PG arm, the free-space preflight, durable status, and the alarms.
- **Boot pre-migration copy**, SQLite arm only, inside the existing `init_database()` `flock` window before the runners, gated on migrations actually being pending.
- **Retention**: `backup_retention_days` joins the ops-settings model (default 14, bounds 1–3650) plus a fixed `MIN_KEEP = 3` floor; prune runs on **every** attempt.
- **Observability**: durable `system_settings` status, a `backup` block on the existing admin `GET /api/settings/retention`, an edge-triggered operator alarm plus a staleness re-alarm, the `db-backup-` reserved id prefix, and the `_db-backup` canary L-03 sentinel.

**No schema change** — status rides `system_settings`, config rides ops settings + env. No `db/migrations.py` entry, no Alembic revision, no feature flag.

## The honest reframe: this is new code, not "wire up what exists"

The issue's technical notes say the capability already exists and this is invocation, retention, and observability. Reading the shipped tooling does not support that, and the plan says so rather than inheriting it.

`scripts/deploy/backup-database.sh` is a **workstation-side GCP pull**: it hard-exits without a `deploy.config` naming a GCP instance/zone/project, then `gcloud compute ssh`es in to run a naive live `cp` of `trinity.db` — no `sqlite3 .backup`, no writer quiescing, no `-wal`/`-shm` handling, zero PostgreSQL support, zero retention, no status surface. It cannot be "the backup routine that just needs invoking" on any instance that is not that one original GCP deploy, and its copy step is the exact hazard a backup exists to avoid. The same naive-`cp` recipe was also published in the user doc and in a suggested operator crontab.

So the in-process copy step is genuinely new code. It is small — that was never the objection — but it had to be written, not called.

## Premise correction worth flagging to reviewers

The dossier framed the copy hazard "under WAL". **The database is not in WAL mode.** No `journal_mode` pragma is set anywhere (zero hits across `src/`, `scripts/`, `docker/`; the SQLAlchemy engine passes only `timeout`/`check_same_thread`), and the live dev DB reports `delete` when asked directly. Two consequences the design absorbs rather than papers over:

- `Connection.backup()` is the right primitive in **both** modes, so the choice survives the correction — but in DELETE mode a one-shot copy **holds a read lock for the copy duration**, so a concurrent writer waits up to its 30s busy timeout. Chosen deliberately over incremental `pages=N`, which releases the lock between steps but restarts from scratch whenever another connection writes, i.e. livelocks under sustained writes. The nightly 03:30 UTC window exists for this class, and the wall is made visible: every run logs copy duration and artifact size, and a copy past 20s logs a WARNING naming the approaching 30s limit and the remedies (WAL migration / PostgreSQL / #1278). A session-layer BUSY-retry loop was rejected as platform-wide blast radius for a nightly tail case.
- The per-cleanup-cycle `PRAGMA wal_checkpoint(TRUNCATE)` is a no-op today. Not touched here; noted so nobody reasons from it.

Switching the platform to WAL is a platform-wide concurrency-semantics change and is explicitly out of scope.

## Design points a reviewer should not have to reconstruct

**The lease is duplicate-I/O suppression, never a correctness boundary.** A Redis `SETNX` lease (`db_backup:running`, TTL derived from `PG_DUMP_TIMEOUT_SECONDS + 300` so retuning the dump timeout visibly re-opens the math) fails **open**. Correctness is carried entirely by pid-suffixed tmp files, atomic `os.replace`, day-keyed artifact names, and the pattern-scoped prune. With Redis down at exactly 03:30 both workers can run concurrently; the worst outcome is one of them hitting ENOSPC on its own tmp — a clean, loud failure with the tmp unlinked while the sibling completes. No lease-refresh loop; a run consuming >80% of the TTL logs a WARNING instead.

**Prune runs on every attempt, not only after a success.** The obvious "prune only after a verified backup" rule is a disk-full Catch-22: the disk fills, the preflight skips the backup, the prune is bypassed, and an operator lowering `backup_retention_days` to free space finds it frees nothing until someone shells in. Safety is carried by the prune's own bounds instead — never delete the newest `MIN_KEEP` verified artifacts regardless of age, never delete anything inside the window regardless of disk pressure (no prune-to-make-room), never touch a non-artifact pattern.

**Retention's fail-safe direction is inverted vs #1638.** For DB rows "keep forever" is the safe failure; for backups it is the #1871 disk-fill trap. So `0` is **invalid** for `backup_retention_days` (disabling backups is the separate, explicit `DB_BACKUP_ENABLED=false`), and the reader coerces garbage or out-of-bounds values → **14 + WARNING**, not → 0. Because that inverts the generic `_ops_int` reader (garbage → 0 → "sweep disabled"), the key is **excluded from the generic `windows` map** and rendered only inside the `backup` block through one shared helper — otherwise a single malformed row makes the same response state two different numbers. `MIN_KEEP = 3` is a fixed constant, deliberately not a knob (#1644: a control that must explain which way is safe is the wrong control).

**The boot hook's two skip classes are distinct branches, not one blanket except.** A fresh install skips **silently** at INFO — a first boot must not emit a scary ERROR — while a corrupt file is the loud path. And the corrupt case is pinned against the real incident: the test overwrites the source with the literal `HTTP/1.1 200 OK` / `server: uvicorn` bytes from the issue's hexdump and asserts `run_all_migrations` on the same connection still raises the **same exception type as before this feature existed**. The incident's fingerprint must survive the feature that exists because of it.

## Review-stage fixes (so they are not re-litigated)

Three defects found during review and fixed in `c17f8195`:

1. **`DB_BACKUP_ENABLED=false` did not gate the boot hook**, and prune lives only in the nightly tail. An opted-out install therefore accumulated one un-pruned full-database copy per upgrade, forever — the #1871 class, arriving through the very knob provided to avoid it. Both producers now read one shared `backup_enabled_from_env()`.
2. **Artifacts were written 0644.** They are full database copies; now `chmod 0600` before the atomic rename (best-effort, so a filesystem that refuses `chmod` cannot fail the backup), inside a `0700` directory.
3. **An unparseable `DATABASE_URL` echoed the password** into the ERROR log and the durable status via the exception chain. Now `raise BackupRunError(...) from None` with a URL-free message.

## Security posture

- No new endpoint, no new auth path, no WS, no MCP tool. Status rides the existing admin-only `GET /api/settings/retention`; the only write path is the validated `PUT /api/settings/ops/config` (1–3650, 422 on `0`/garbage), and the generic `PUT /api/settings/{key}` catch-all 422-blocks the key via `RETENTION_OPS_KEYS` membership.
- The password reaches `pg_dump` **only** via `PGPASSWORD` on the subprocess env — never in argv, which is world-readable through `/proc` — and `create_subprocess_exec` means no shell ever sees the conninfo URI. Timeout → `proc.kill()` → `await proc.wait()` (reap: `wait_for` cancels the await, never the child) → tmp unlinked in `finally`.
- The full `DATABASE_URL` is passed through with exactly two rewrites: driver suffix normalized (`postgresql+psycopg2://` → `postgresql://`, which libpq rejects) and password stripped. Query params (`sslmode`, `sslrootcert`, `options`) survive — re-parsing into `-h/-p/-U` flags would silently drop what managed PG mandates and break backups permanently for exactly the external-PG installs.
- Prune is scoped to three artifact globs among the direct children of the backup dir; deletions are logged by name only. Alarm context carries status, paths, and sizes — never row data (canary G-04).
- `db-backup-` is a reserved operator-queue id prefix, so an agent cannot pre-create the id and let `on_conflict_do_nothing` swallow the platform's own failure alarm; the `_db-backup` sentinel host is uncreatable (`sanitize_agent_name` strips the leading `_`) and excluded from canary L-03 with service↔canary parity pinned by test.
- Artifacts contain the AES-256-GCM envelopes but not `CREDENTIAL_ENCRYPTION_KEY`, which lives in `.env` — an artifact alone decrypts nothing. Backing up `.env` stays a documented operator step.

## Verification evidence

`/verify-local --skip-agent` **with the real backend image build** (the Dockerfile gained a package — the #1033 packaging class is exactly what it exists for). Verdict **PASS**.

- build + import-smoke OK (71s); boot + health OK; integration **70 passed / 13 skipped / 2 deselected**.
- `docker run --rm --entrypoint "" <built backend image> pg_dump --version` → `pg_dump (PostgreSQL) 17.11 (Debian 17.11-0+deb13u1)`; `dpkg -l` shows `postgresql-client-17` — the major pin holds on trixie.
- Boot log: `DB backup scheduler started: daily at 03:30 UTC → /data/backups`, **zero** `[DBBackup]` errors, zero ERROR lines in the whole boot; retention log shows `backup_retention_days=14d (code-default)`.
- `GET /api/settings/retention` → `backup` block with `scope: "same-disk"`, `enabled: true`, `retention_days: 14`, `min_keep: 3`; `backup_retention_days` correctly **absent** from the generic windows map.
- `run_backup(trigger='manual')` → `{"status":"ok","path":"/data/backups/trinity-backup-20260816.db","size_bytes":1433600}`; artifact **mode 600** in a dir **mode 700**, owned `1000:1000`, `PRAGMA integrity_check` → `ok` with **74 tables** — a real restorable copy, not a produced file. Second call → `{"status":"skipped_exists"}` (day-key idempotence).
- **The cross-version claim was tested, not assumed**: a throwaway `postgres:16-alpine` (server 16.15) was dumped with pg_dump 17.11 **from the built image** → exit 0, and `pg_restore --list` read the archive back.
- Unit, deterministic (`-p no:randomly`): branch **8 failed, 10510 passed**; base **8 failed, 10411 passed** → **+99 passing, 0 new failures**.
- Affected-file neighbourhood re-run at ship: `pytest tests/unit/test_2216_*.py tests/unit/test_retention_floor.py tests/unit/test_1644_retention_guard.py tests/unit/test_1771a_retention_edges.py tests/unit/test_297_ops_settings_validation.py -q -p no:randomly` → **312 passed**.

## Honest coverage gaps (by design, unit-covered)

- **The boot pre-migration hook did not fire on the verify stack.** A fresh install is a documented silent skip — no `schema_migrations` rows means there is nothing worth copying — so the verify stack exercised exactly the skip branch. The backup branch, the ordering guarantee (artifact exists before the migration mutates), the corrupt-DB fail-open, and the unchanged downstream fingerprint are covered in `test_2216_boot_pre_migration_backup.py`.
- **`_pg_dump_to` was not driven through a live PG-backed stack.** The binary was exercised directly from the built image instead (version probe + the PG16 round-trip above); the conninfo construction, `PGPASSWORD`-not-argv, timeout kill+reap, missing-binary, and non-zero-exit paths are unit-covered with the subprocess mocked.

## Pre-existing reds disclosure

The 8 unit failures above are **pre-existing dev-tip rot**, present identically on base and not introduced here: 5× `test_ent96_timeline_split.py` (hardcoded 2026-08-14 date keys) and 3× `test_pat_propagation_properties.py::TestKnownGaps` (stale strict-xfail after #2017). `pytest-randomly` is installed, so unit runs need `-p no:randomly` to be comparable.

## Release notes

- **Automatic database backups are on by default.** Daily at 03:30 UTC plus a pre-migration copy at any boot with pending migrations, written to `/data/backups/` on the same bind mount as the database. Scope is **same-disk**: this protects against the incident class (a stray write, a fat-fingered delete, a bad migration), **not** against disk loss.
- New env knobs, forwarded in **both** compose files and documented in `.env.example`: `DB_BACKUP_ENABLED` (default `true` — the documented way to disable), `DB_BACKUP_HOUR`, `DB_BACKUP_MINUTE`, `DB_BACKUP_PG_DUMP_TIMEOUT_SECONDS`.
- New ops setting `backup_retention_days` (default 14, bounds 1–3650, `0` invalid) written via `PUT /api/settings/ops/config`; a fixed floor keeps the newest 3 artifacts regardless of the window.
- **The PostgreSQL arm requires a rebuilt backend image** (it gained `postgresql-client-17`). An un-rebuilt image degrades **loudly** — a failed status plus an operator alarm naming the rebuild — never silently.
- Backup status is readable without shell access at `GET /api/settings/retention` → `backup` (admin-only).
- Failures raise operator-queue alarms on the `_db-backup` sentinel: edge-triggered on the success→failure transition and on a no-space skip, plus a staleness re-alarm when the last success ages past 3 days (re-fired at most weekly), so a sustained failure cannot go quiet.
- The legacy `scripts/deploy/backup-database.sh` / `restore-database.sh` no longer ship the naive-`cp` recipe; the restore script now stops the scheduler as well as the backend and clears stale `-wal`/`-shm`/`-journal` sidecars.

1. Off-site backup destination via the existing `archive_storage.ArchiveStorage` ABC — the ready seam for the disk-loss case this PR explicitly does not cover.
2. PostgreSQL boot-time pre-migration backup (v1 is SQLite-only; Alembic DDL is transactional and a boot-path subprocess adds import-time latency).
3. Compression of SQLite artifacts.
4. A manual "back up now" endpoint.
5. WAL journal-mode migration for the platform DB (would remove the DELETE-mode read-lock hold entirely).
6. The pre-existing unforwarded `DB_VACUUM_*` / `AUDIT_RETENTION_*` env knobs — still inert today for the same #1486/#1488 reason this PR avoided; untouched here.
7. `Settings.vue` rendering of the new `backup` block — the retention knob is API-only today.
8. The legacy GCP pull script may need `sudo` when the remote ssh user is not UID 1000, now that artifacts are 0600.
9. `architecture.md` still says `RETENTION_OPS_KEYS` carries 7 windows; it carries 9 (pre-existing staleness, not introduced here).
10. Dev-tip test rot: the ent#96 hardcoded dates and the #2017 stale xfail behind the 8 pre-existing failures.


vybe (approved):
Validated via /validate-pr. The description's honest reframe is correct and load-bearing: the shipped backup-database.sh is a workstation-side GCP pull that hard-exits without a deploy.config and does a naive live `cp`, so this genuinely is new code rather than wiring. Design points I checked rather than took on faith — Connection.backup() is right in both journal modes so the not-in-WAL correction doesn't invalidate the choice; the Redis lease is explicitly duplicate-I/O suppression with correctness carried by pid-suffixed tmp + atomic os.replace + day-keyed names; prune-on-every-attempt avoids the disk-full Catch-22 with safety carried by MIN_KEEP and the window bounds instead; and backup_retention_days correctly inverts #1638's fail-safe direction (0 invalid, garbage → 14 + WARNING) and is excluded from the generic windows map so one malformed row can't make the response state two numbers. postgresql-client-17 major-pin justified inline against the #1823 silent-distro-migration precedent. DB_BACKUP_* wired into both compose files and .env.example.

Merge resolution: two conflicts against dev. (1) feature-flows.md changelog row — both kept. (2) operator_queue_service._RESERVED_ID_PREFIXES against #2168 — union, both prefixes kept, and #2168's AST caller-parity guard then correctly failed on this PR's new DBBackupService._emit_alarm call site. Classified it platform-only in _ALLOWED_CALLERS after verifying the claim: call site 1 is edge-triggered on the durable prior-status transition (prior in (None,'ok')), call site 2 is a staleness re-alarm throttled to weekly, and no agent input reaches either. 62 affected tests pass locally; all 22 checks green.

PR #2237, opened 2026-08-16
fix(dashboard): Executions tile stops slicing its legend and says what it hides (#2228)

### LINKED ISSUE #2228: bug: Executions grid tile clips its legend mid-row, silently hiding trigger keys
## Summary

The Fleet **Executions** info tile on the Dashboard Grid clips its trigger legend **through the middle of the second row**, leaving a strip of half-rendered glyphs floating above the "Open executions →" footer. Any legend key that falls past the clamp is dropped with no indication, so a colour that is plainly visible in the stacked chart can have no decoder anywhere on the tile.

Observed with 8 keys present (Chat/Tasks · MCP · Channels · Public · Scheduled · Agent-to-agent · Other · failed): row 1 renders normally, row 2 renders as a sliced band of letterforms.

## Context

`src/frontend/src/components/tiles/ExecutionsTile.vue` renders one legend key per **present** bucket (`presentBuckets`, `src/frontend/src/utils/executionsTile.js`) plus a `failed` key, in a wrapping flex row bounded by:

```css
.ex-legend {
  display: flex;
  flex-wrap: wrap;
  gap: 2px 8px;
  font-size: 10px;
  overflow: hidden;
  max-height: 26px;   /* ~1.9 rows at this line-height */
}
```

Two distinct defects fall out of that clamp:

1. **The clamp lands mid-row.** `max-height: 26px` is not a whole number of legend rows, so `overflow: hidden` cuts row 2 horizontally instead of hiding it. The result reads as a rendering fault, not as truncation.
2. **Truncation is silent.** The vocabulary is backend-supplied (`trigger_order` from `GET /api/executions/timeline?split=trigger`) and can legitimately reach 10 buckets + `failed`. Whichever keys land past the clamp simply vanish — there is no `+N more` affordance and no tooltip — so the stacked chart becomes partially undecodable exactly on the busiest fleets, which are the ones the tile exists for.

The clamp itself is not gratuitous: grid tiles are a fixed `CELL_H = 216` (`src/frontend/src/utils/gridLayout.js`), the tile is `display: flex; column`, and `.it-body` is `overflow: hidden`. So an unbounded legend would push the chart or the footer link out of the tile. **Deleting the `max-height` is not the fix** — the legend needs a bounded presentation that stays honest about what it is not showing.

## Acceptance Criteria

- [ ] The legend never renders a partially-clipped row — the visible area is a whole number of rows in both themes and at the tile's real font metrics.
- [ ] When keys do not fit, the omission is **stated**, not silent (e.g. a `+N` chip whose `title` lists the hidden bucket names, or an equivalent affordance) — a user can always determine what every colour in the chart means without leaving the tile.
- [ ] Every colour rendered in a chart column is decodable from the tile at any bucket count the backend can return (up to the full `trigger_order` set plus `failed`).
- [ ] Tile total height is unchanged: the chart, the head line and the "Open executions →" footer all stay inside `CELL_H` with the maximum number of legend keys present.
- [ ] Works in light and dark (`--gv-*` tokens only — no raw Tailwind palette classes, no hardcoded hex) per `docs/memory/design-system-contract.md`.
- [ ] Layout is stable: adding or removing a bucket between polls must not reflow the chart or shift the footer.
- [ ] Pure shaping logic (which keys are shown vs. summarized) lives in `src/frontend/src/utils/executionsTile.js` and is unit-tested — vitest runs node-environment here, so a decision inside the component is one no test can reach.

## Technical Notes

- Component: `src/frontend/src/components/tiles/ExecutionsTile.vue` (`.ex-legend` block)
- Shaping: `src/frontend/src/utils/executionsTile.js` → `presentBuckets()`
- Chassis: `src/frontend/src/components/InfoTile.vue` (`.it-body { overflow: hidden }`), `src/frontend/src/utils/gridLayout.js` (`CELL_H = 216`)
- Bucket vocabulary/order is backend-owned (`_BUCKET_ORDER` → `trigger_order`); do not hardcode a second copy of the list or its order in the frontend.
- Worth checking whether `RecentFailuresTile.vue` or any other `InfoTile` consumer uses the same mid-row `max-height` clamp pattern.

trinity-ability:
## Summary

- The Executions tile's legend clamped with `max-height: 26px` — ~1.7 rows at its 15px pitch — so `overflow: hidden` **cut row two through the glyphs** instead of hiding it, and any key past the clamp vanished with no `+N` and no tooltip.
- Measuring the running tile found a **second clip** behind the same symptom: `.it-body` is `overflow: hidden` at a fixed `CELL_H` and offers **134px**, while the tile already asked for **~138 before this change**. Fixing the legend alone leaves the symptom on screen.
- The legend is now a fixed two whole rows, the vertical budget is stated explicitly and fits, and anything that still does not fit is handed to a `+N` chip that names it.

## The two clips

| | before | after |
|---|---|---|
| `.ex-legend` row 2 | **sliced** (cut 11px into a 13px row) | fully visible |
| `.it-body` overflow | **6px** (4px before this branch) | **0px** |

## Changes

**`src/frontend/src/components/tiles/ExecutionsTile.vue`**
- `.ex-head` pins `line-height: 1` — baseline-aligning 20px and 12px text made the box 22.1px, and every px here comes out of the legend at the far end of a fixed-height tile.
- `.ex-chart` 78px → 70px = `CHART_HEIGHT + COL_GAP + RAIL_HEIGHT + padding`. It stood 4px above its tallest possible column: dead space paid for at the legend's expense.
- `.ex-legend` takes a fixed **28px** (two rows). `height`, not `max-height` — a ceiling grows with the key count and would resize the chart every time a bucket appeared or went quiet.
- Renders `legend.shown` plus a `+N` chip carrying the hidden swatches and naming them in its `title`.

**`src/frontend/src/utils/executionsTile.js`**
- `legendFit()` packs keys into the available rows and returns `{shown, hidden}`. CSS can clamp but cannot report *what* it dropped.
- The width model is pessimistic **per key, not in aggregate**: greedy packing wastes space at the end of each row, so a model generous overall but tight on one name still lets a real third row form. Over-reserving shows an unnecessary `+N`; under-reserving loses keys silently.
- `LEGEND_WIDTH = CELL_W - 50`, not `- 46`: counting only InfoTile's padding and forgetting its 1px borders claims 338px for an element that measures **334** — error in the one direction that manufactures a third row.
- `CHART_HEIGHT` / `RAIL_HEIGHT` / `COL_GAP` / `CHART_PAD` make the chart's share of the budget explicit beside the legend's.

## Test Plan

- [x] `npx vitest run` — **670 passed**, incl. 40 for this tile
- [x] Source-parsing guards pin the CSS budget to the JS constants (a stylesheet and a constant cannot check each other; jsdom does no cascade resolution, so a mid-row clamp looks identical to a clean one under `getComputedStyle` — same rationale as `gridTokens.spec.js`)
- [x] `vite build` clean; raw-color ratchet unchanged (tokens only, no new token)
- [x] Verified in the running app with the full 11-key vocabulary:

| check | result |
|---|---|
| legend rows / sliced keys | `2` / `0` |
| keys hidden by the body | `0` |
| `.it-body` overflow | `0px` (was `6`) |
| legend height @ 2 keys vs 11 keys | `28px` / `28px` → chart never reflows |
| width model vs measured labels | **0 violations** across all 11 |
| light + dark | same geometry; chip resolves real tokens in both |

Calibration figures in the comments are **measured, not estimated** — my first pass had them wrong in the unsafe direction, which is how the 334-vs-338 width error surfaced.

## Notes

Per CLAUDE.md tiered docs, a bug fix needs a descriptive commit message only — no `architecture.md` / feature-flow change. Behaviour on today's fleets is unchanged apart from the fix: eight keys still render without a chip.

Fixes #2228

### DIFF
--- a/src/frontend/src/components/tiles/ExecutionsTile.vue
+++ b/src/frontend/src/components/tiles/ExecutionsTile.vue
@@ -59,11 +59,34 @@
       </div>
     </div>
 
+    <!-- Exactly LEGEND_ROWS whole rows (#2228). The clamp below hides a third
+         row outright rather than slicing one, and `legendFit` guarantees there
+         is nothing in it to hide — anything that would not fit is handed to the
+         `+N` chip, which names it rather than dropping it silently. -->
     <div class="ex-legend">
-      <span v-for="b in legend" :key="b.name" class="ex-key" :title="`${b.name}: ${b.total}`">
-        <i :class="'ex-dot ex-bk-' + b.token"></i>{{ b.name }}
+      <span
+        v-for="k in legend.shown"
+        :key="k.fail ? '\u0000failed' : k.name"
+        class="ex-key"
+        :class="{ 'ex-key-fail': k.fail }"
+        :title="`${k.name}: ${k.total}`"
+      >
+        <i class="ex-dot" :class="k.fail ? 'ex-dot-fail' : 'ex-bk-' + k.token"></i>{{ k.name }}
       </span>
-      <span v-if="head.failed" class="ex-key ex-key-fail"><i class="ex-dot ex-dot-fail"></i>failed</span>
+      <span v-if="legend.hidden.length" class="ex-key ex-key-more" :title="hiddenTitle">
+        <!-- Dots only while they still identify WHICH buckets are hidden. Past
+             three the swatches stop being a mapping and become decoration, so
+             the count carries it alone and the title stays authoritative. -->
+        <template v-if="legend.hidden.length <= 3">
+          <i
+            v-for="k in legend.hidden"
+            :key="k.name"
+            class="ex-dot"
+            :class="k.fail ? 'ex-dot-fail' : 'ex-bk-' + k.token"
+          ></i>
+        </template>
+        +{{ legend.hidden.length }}
+      </span>
     </div>
   </InfoTile>
 </template>
@@ -98,9 +121,12 @@ import { computed } from 'vue'
 import InfoTile from '../InfoTile.vue'
 import { useFleetGridStore } from '@/stores/fleetGrid'
 import {
+  CHART_HEIGHT,
   chartColumns,
   headline,
-  presentBuckets,
+  legendFit,
+  legendKeys,
+  RAIL_HEIGHT,
   tileState,
 } from '@/utils/executionsTile'
 
@@ -115,9 +141,23 @@ const gridStore = useFleetGridStore()
 const buckets = computed(() => gridStore.execTimeline || [])
 const head = computed(() => headline(buckets.value))
 const columns = computed(() =>
-  chartColumns(buckets.value, { triggerOrder: gridStore.execTriggerOrder }),
+  chartColumns(buckets.value, {
+    triggerOrder: gridStore.execTriggerOrder,
+    chartHeight: CHART_HEIGHT,
+    railHeight: RAIL_HEIGHT,
+  }),
 )
-const legend = computed(() => presentBuckets(gridStore.execTriggerOrder, buckets.value))
+const legend = computed(() =>
+  legendFit(legendKeys(gridStore.execTriggerOrder, buckets.value)),
+)
+
+/**
+ * What the `+N` chip is standing in for. Bucket names and counts only — the
+ * same backend vocabulary the columns already report on hover.
+ */
+const hiddenTitle = computed(
+  () => `Not shown: ${legend.value.hidden.map((k) => `${k.name}: ${k.total}`).join(' · ')}`,
+)
 
 const state = computed(() =>
   tileState({
@@ -168,6 +208,10 @@ function retry() {
   align-items: baseline;
   gap: 4px;
   font-size: 12px;
+  /* Pinned, not inherited: with the default line-height the baseline-aligned
+     mix of 20px and 12px text made this box 22.1px, and every px here comes
+     straight out of the legend at the other end of a fixed-height tile. */
+  line-height: 1;
   color: var(--gv-muted);
   margin-bottom: 6px;
 }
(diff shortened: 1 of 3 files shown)


vybe (approved):
Validated per /validate-pr.

**Correctness of the fix**: two clips, both real. The `max-height: 26px` clamp landing mid-row (26px is not a whole multiple of the 15px pitch) is the reported symptom; the `.it-body` overflow at fixed `CELL_H` is the one that would have left the symptom on screen after a legend-only fix. Fixing both together is the right call.

**The width model's bias is correct and correctly argued.** `legendFit`/`keyWidth` are pessimistic **per key**, not in aggregate — which is the load-bearing distinction, since greedy packing wastes space at the end of each row and an aggregate-only margin still lets a real third row form. The asymmetry is stated and the safe direction chosen: over-reserve shows an unnecessary `+N`, under-reserve loses keys silently. `LEGEND_WIDTH = CELL_W - 50` (padding **plus** the 1px borders) rather than `- 46` is the same discipline — 46 claims 338px for an element that measures 334.

**`pack()` reviewed for termination**: the `used === 0` force-place means a key wider than a whole row is clipped by CSS rather than spinning the packer; the two-pass chip reservation avoids truncating sets that fit without a chip. No unbounded loop.

**Accepted trade-off**: at the full 11-key vocabulary the model shows `+2` where the browser would fit them. Documented, and it gives up the last keys in stack order (`Other`, `failed`) — the right end to lose from, with the failure count already stated in the headline and both named in the chip's `title`.

**CI**: fully green, all 20 non-skipped checks including `e2e` and all six pytest legs. Frontend-only diff, zero backend surface.

**Scan**: no secrets, no host paths, no mode/symlink changes. Bug fix → descriptive commit only per the tiered-docs rule; correctly no architecture/feature-flow change.

Source-parsing guards pinning the CSS budget to the JS constants are the right shape here — jsdom does no cascade resolution, so a mid-row clamp and a clean one are indistinguishable under `getComputedStyle`.

PR #2239, opened 2026-08-17
fix(workspace): dictate on the path we control, and say why voice failed (#2212)

### LINKED ISSUE #2212: bug: Workspace voice mode starts the mic and stops instantly, with every failure silent
## Summary

The Workspace voice control turns the mic on and stops again immediately, with nothing shown to the user. Dictation never produces text in the composer.

## Context

Reported from live use of `/workspace`: clicking voice mode "kind of turns on the mic and dies instantly". The relevant code is the STT/dictation block in `src/frontend/src/components/portal/PortalConversation.vue` ("Voice: speak replies (TTS) + dictate (STT)").

**Likely mechanism — the browser Web Speech path is preferred, and it commonly ends instantly.**

```js
const micMode = SpeechRec ? 'speech' : (canRecord ? 'record' : null)
```

When `window.SpeechRecognition`/`webkitSpeechRecognition` exists (Chrome), the MediaRecorder fallback is never used. That API is server-backed and routinely ends within a second with `network`, `no-speech`, `aborted`, or `service-not-allowed` — exactly the reported symptom. It is also not set `continuous`, so it ends at the first pause regardless.

**Every failure path is silent, which is why it presents as "just dies".**

- `recog.onerror = () => { listening.value = false }` — the error code is discarded
- `try { recog.start() } catch { listening.value = false }`
- `catch { return }` around `getUserMedia` — a denied mic permission is indistinguishable from success-then-stop
- `if (blob.size < 1500) return` — a short recording is dropped with no feedback
- `catch { /* keep text mode */ }` around the transcription call — a server-side failure is invisible

**Possible second gap:** the mic button renders on `sttSupported` (a pure browser-capability check) and is not gated on the platform actually being able to transcribe, while the header speaker toggle gates on `agent.voice_available`. If the instance has no STT/voice provider configured, the control is a dead affordance.

## Acceptance Criteria

- [ ] Root cause confirmed with the real signal captured (the `SpeechRecognition` error code and/or the `getUserMedia` rejection), not inferred
- [ ] Dictation produces a transcript in the composer on a supported browser
- [ ] Failures are surfaced to the user with an actionable message — permission denied, no speech heard, network/service error, voice not configured on this instance — instead of silently reverting to idle
- [ ] The mic control is hidden or disabled with an honest reason when transcription cannot work on this instance (no dead control; honest status; graceful fallback to text)
- [ ] Verified in both engines, since they take different `micMode` branches (Chromium → `speech`, Safari/Firefox → `record`)
- [ ] The header speaker toggle (reply narration / TTS) verified separately — it is a different path gated on `agent.voice_available`
- [ ] Same review applied to the room composer if it shares the control

## Technical Notes

- File: `src/frontend/src/components/portal/PortalConversation.vue` — `micMode`, `sttSupported`, `toggleMic()`, `toggleSpeech()`, `toggleRecord()`, `cleanupVoice()`
- Store calls: `transcribeStt(agentName, blob)` and `synthesizeTts(agentName, text)` in `src/frontend/src/stores/clientPortal.js`
- Backend TTS side: `synthesize_portal_tts` in `src/backend/client_portal/service.py` (404s when no key / no voice resolves) — worth confirming the STT counterpart's failure shape too
- Consider preferring the MediaRecorder + server-transcription path over the browser Web Speech API when both exist, since only the former is under our control

dolho:
## What this fixes

The Workspace mic "kind of turns on and dies instantly", with nothing shown, and dictation never reaches the composer. Two independent causes.

### 1. The branch — we preferred the path we do not control

`micMode` chose the browser Web Speech API whenever the object merely **existed**, so Chrome never reached the MediaRecorder + `/stt` path.

**Real signal captured** (Playwright Chromium, `http://127.0.0.1` secure context, microphone permission granted, `--use-fake-device-for-media-stream`):

```
capabilities: {"hasSpeechRec":true,"hasMediaDevices":true,"hasMediaRecorder":true,"isSecureContext":true}
WEB SPEECH: { "continuousDefault": false, "threw": null, "endedBy": "timeout-15s", "events": [] }
MEDIARECORDER: { "ok": true, "mimeType": "audio/webm;codecs=opus", "bytesFor2s": 24236 }
```

`start()` does not throw and the engine then emits **no event at all** — no `start`, no `audiostart`, no `result`, no `error`, no `end` — so `listening` sticks on with no recourse. `continuous` also defaults to `false`, so even where the service *does* answer (official Chrome), recognition ends at the first pause. It is a browser-hosted, Google-backed service: not ours to fix, not ours to see into.

### 2. The silence — every failure path discarded its own signal

`onerror` dropped the error code; `getUserMedia` was wrapped in `catch { return }` (a denied mic is then indistinguishable from success-then-stop); a short blob was dropped with no feedback; the transcription call ended in `catch { /* keep text mode */ }`.

## Changes

- **Prefer the server path.** `resolveMicMode` picks `record` (MediaRecorder → `POST /stt`) whenever the platform can transcribe — that path is ours end to end: a real status, a real message, a real transcript. Web Speech stays the **no-key fallback** instead of the default.
- **A separate honest bit.** New roster-card field `stt_available` = exactly the `/stt` gate (platform ElevenLabs key resolves). It cannot be folded into `voice_available`, which additionally needs an *effective voice to speak with*: on a key-but-no-voice agent, dictation works and narration does not, and one bit has no correct answer for that. With neither path available the mic is **not rendered** — no dead control.
- **Every failure speaks**, in one inline `aria-live` notice beside a composer that still works: SpeechRecognition error codes (blocked permission vs unreachable service — the distinction the discarded code held), the `getUserMedia` rejection name, a too-short recording, the `/stt` `detail` verbatim, and a narration (TTS) failure.
- **A watchdog** for the engine that reports nothing (the measured case), so the button cannot stay lit forever.
- **`continuous = true`** plus final-result accumulation, so a pause is not the end of dictation.
- **Firefox upload mislabel, found while verifying.** Firefox reports `mediaRec.mimeType === ""` and puts `audio/ogg; codecs=opus` on the chunks, so every clip from the one engine that *always* records was uploaded as `voice.webm` labelled `audio/webm` while containing Ogg. `resolveRecordingMimeType` reads the chunk when the recorder is silent.

Decisions live in pure helpers (`portalUtils.js`) because this project has no component-mount harness; the wiring is pinned by source-structure guards in the shape of `portalVoiceModePersistence.spec.js`.

## Acceptance criteria

- [x] **Root cause confirmed with the real signal captured** — the probe above; not inferred.
- [x] **Dictation produces a transcript** — real end-to-end: ElevenLabs TTS → Chromium fake-mic capture → MediaRecorder (`audio/webm;codecs=opus`, 97 KB) → ElevenLabs Scribe returned `"Book the meeting for Thursday afternoon and send Rachel the agenda."`, the sentence spoken in.
- [x] **Failures surfaced with an actionable message** — permission denied / no speech / network-service / not configured / too short / provider `detail`, each its own line (asserted distinct in tests).
- [x] **No dead control** — mic hidden when neither path can work; the card bit and the endpoint's own 404 gate are asserted to be the **same condition**, since drift between them *is* the dead-control bug.
- [x] **Verified in both engines**, feeding each engine's real capabilities into the shipped resolver:

  | engine | `speechApi` | `canRecord` | with key | without key |
  |---|---|---|---|---|
  | Chromium | true | true | `record` | `speech` |
  | Firefox | false | true | `record` | *(no mic)* |

- [x] **Header speaker toggle (TTS) reviewed separately** — it stays on its own `voice_available` gate, untouched apart from now reporting a failure instead of silently doing nothing; the #2157 gate test still passes, and the TTS provider call returned 200 with this instance's key during the end-to-end above.
- [x] **Room composer** — `components/rooms/RoomComposer.vue` has no mic/voice control (the only STT call site in the frontend is `PortalConversation.vue`), so there is nothing to apply there.

## Verification

- `vitest run` — **841 passed / 37 files** (34 new in `portalVoiceDictation.spec.js`)
- `pytest tests/unit/test_2212_portal_stt_gate.py` — **6 passed**
- every `client_portal` backend unit test — **444 passed**
- `vite build` clean; `check-design-tokens` clean

Fixes #2212


vybe (approved):
Validated via /validate-pr. Root cause captured from the running browser rather than inferred; the `stt_available` split is argued from a case the single-bit design cannot express, and the card bit is literally `tts_service.is_available()` on both sides — pinned by a parametrized test, which is the drift that *is* the dead-control bug. All 22 checks green.

PR #2244, opened 2026-08-17
fix(tests): derive time from the clock, not the calendar; retire four stale xfails (#2243)

### LINKED ISSUE #2243: bug: calendar-expiring tests and stale strict-xfails keep 17 tests permanently red
## Summary

Two sets of tests fail purely because the calendar moved, and a third set fails because a bug they document has since been fixed. None indicate a product defect; all three make the suite permanently red until touched.

**1. Hard-coded time buckets (5 failures)** — `tests/unit/test_ent96_timeline_split.py` asserts against the literal bucket `"2026-08-14T09"`:
```
test_split_rows_fold_to_one_row_per_interval            KeyError: '2026-08-14T09'
test_the_breakdown_always_sums_to_its_own_column        StopIteration
test_failures_are_carried_per_label_not_only_in_the_column_total
test_an_unmapped_trigger_lands_in_Other_rather_than_vanishing
test_two_raw_triggers_in_one_bucket_merge_into_that_bucket
```
The endpoint gap-fills a continuous UTC axis relative to *now*, so once that hour falls outside the window the bucket simply does not exist. These passed when written and fail from then on.

**2. Hard-coded `started_at` (9 PostgreSQL failures)** — `tests/unit/test_1771c_schedules_cas_edges.py::insert_execution` defaults to `started_at="2026-01-01T00:00:00.000000Z"`. The terminal write computes `duration_ms = now − started_at`, which grows without bound as the fixed anchor recedes; on PostgreSQL it eventually overflows the int32 column and raises `NumericValueOutOfRange` (`test_a1b_*`, `test_a2_*`, `test_a3_*`, `test_a7_*`, `test_a8_*`).

The column width is a genuine product concern tracked separately, but the test's own fixed anchor is what makes these fail *now* and would keep them failing even after the column is widened — the value only ever grows.

**3. Stale strict-xfail markers (3 failures)** — `TestKnownGaps::test_a_backslash_in_the_token_does_not_raise` for params `tok\1`, `tok\g<1>`, `tok\slash` all report `[XPASS(strict)]`:
> BUG: the new line is used as an `re.sub` REPLACEMENT, so a backslash in the token is parsed as a group reference and raises re.error mid-rotation.

The documented bug is fixed; the `xfail(strict=True)` marker was never removed, so success is reported as failure. This is good news wearing a failure's clothes — and worth removing promptly, because a strict xfail is exactly the marker you want available for the *next* known gap.

## Context

Found by a full-suite run on `dev` (`691e7dc9`). Together these are 17 of the run's 44 failures, and none of them is a product regression — which is precisely the problem: a suite with a standing red block trains people to read past it, and the #2080 full-suite work exists to stop exactly that.

## Acceptance Criteria

- [ ] `test_ent96_timeline_split.py` computes its expected buckets relative to a controlled clock (frozen/injected now, or derived from the same window the endpoint uses) instead of literal date strings — and would still pass a year from now
- [ ] `test_1771c_schedules_cas_edges.py::insert_execution` anchors `started_at` relative to now (or per-test explicit), so `duration_ms` stays bounded regardless of run date
- [ ] The three `TestKnownGaps` backslash params no longer XPASS — the `xfail(strict=True)` marker is removed and the tests assert the fixed behaviour directly
- [ ] Grep for other literal future/past date strings in `tests/` and either fix or note them, so this class is closed rather than patched twice
- [ ] Full suite has no calendar-dependent failures on a machine whose clock is set a year forward

## Technical Notes

- The repo already depends on `hypothesis`; `freezegun` is not currently a test dependency — prefer injecting/deriving the reference time over adding one, unless a frozen clock is clearly cleaner.
- Reproduce: `cd tests && source .venv/bin/activate && python -m pytest unit/test_ent96_timeline_split.py -q`
- The PostgreSQL half needs the pg tier: `tests/run-full.sh --tier postgres`

dolho:
## What this fixes

17 tests were permanently red on `dev` for reasons unrelated to the product. All three sets are fixed, and a fourth stale marker turned up while fixing the third.

### 1. Hard-coded time buckets — 5 failures

`test_ent96_timeline_split.py` asserted against the literal hour `"2026-08-14T09"`. `_gap_fill` builds a continuous UTC axis relative to `datetime.now(timezone.utc)`, so once that hour left the window the bucket stopped existing — `KeyError` on the dict lookups, `StopIteration` on the `next(...)` ones.

The key is now derived from the same clock the axis is, at an offset kept strictly inside the narrowest window under test (6h), so an hour boundary ticking between the test's `now` and `_gap_fill`'s cannot move it out. The row variable was renamed `col` so nothing depends on a rebind inside a generator expression.

### 2. Hard-coded `started_at` — 13 PostgreSQL failures

Not the 9 the issue lists: **the count grows with the calendar**, and I reproduced 13 on unmodified `dev` before changing anything.

`insert_execution` defaulted to `2026-01-01T00:00:00.000000Z`. The terminal write computes `duration_ms = now − started_at`, so a fixed anchor gains ~31.5 billion ms a year; it had already passed int32 and raised `NumericValueOutOfRange` on PostgreSQL, while SQLite — untyped — kept quietly widening the value.

- the default is now a few minutes ago, so the duration is bounded whatever year the suite runs in — and would stay bounded after the column is widened, since a fixed anchor's duration only ever grows
- **A7** kept its four timestamp-format variants (naive / `Z` / `+00:00` / `+05:00`) but renders them from **one** recent instant: the formats were the point of that case, the calendar date was not. It also gained a bound on the resulting duration, so a form parsed with the wrong offset now fails loudly (±5h) instead of passing
- **`started_at=None` still means SQL NULL.** A5b passes it deliberately to prove the `NOT NULL` constraint is what stands between production and `parse_iso_timestamp(None)`, so "caller said nothing" needed a sentinel — a `None` default silently turned that probe into a passing insert (the suite caught this mid-change)

### 3. Stale strict-xfails — 3 failures, and a fourth found

The three backslash params reported `XPASS(strict)` because #2017 (PR #2024) fixed the `re.sub`-replacement bug they pinned. Retired to plain regression tests that assert the **round trip** rather than merely "does not raise" — not raising was the crash symptom; writing a backslash the agent decodes back to the same token is the contract.

That required fixing the oracle, which exposed the fourth. `agent_reads` in this file was a hand-written replica that only stripped quotes — but #2023 made the `.env` encoding reversible (escape the escape character, then the quote; `unquote_env_value` reverses both in one scan). So:

- value decoding now delegates to the **real** agent-side reader, loaded by path via the idiom `test_2023_env_quote_round_trip.py` already uses; the line scan stays local because last-wins is what makes a duplicate PAT line dangerous (#2016)
- against the real reader, `test_a_quote_in_the_token_round_trips` XPASSes. It had been reporting **XFAIL — green** — for a gap #2023 closed, and only the replica made the claim look true. Same class as the three above reached from the other side: a stale marker hidden by a stale copy rather than by an XPASS. Measured (`ghp_a"b` round-trips exactly), then retired.

### 4. Closing the class, not patching it twice — `tests/clockshift.py`

Grep finds **124** files under `tests/` carrying literal ISO timestamps, and no grep can tell an inert fixture value from one compared against *now*. Running the clock forward can.

```
PYTHONPATH=tests python -m pytest tests/unit -q -p clockshift
```

Opt-in pytest plugin, `CLOCKSHIFT_DAYS`-tunable (negative works too), **no new test dependency** — the issue asked to prefer deriving over adding `freezegun`, so this shifts only what test code reads (`datetime.now/utcnow/today`, `time.time`) and documents its own limits: it does not move a database's own clock, so a test whose expectations come from SQL-side time is out of its reach. It is not a `conftest` on purpose — it must not load for runs that did not ask for it.

## Verification

| what | result |
|---|---|
| `test_ent96_timeline_split.py` | 14 passed — today **and** at `+365d` |
| `test_1771c_schedules_cas_edges.py` | **86 passed** (43 SQLite + 43 PostgreSQL), from 13 PG failures reproduced on unmodified `dev` first |
| PostgreSQL tier | throwaway `postgres:16-alpine` + `alembic upgrade head`, per `run-full.sh --tier postgres` |
| PAT properties + `test_2023` parity | 97 passed, **zero** xfail/xpass |
| fixed files under a shifted clock | green at `+365d` and at `-400d` |

The full-unit `+1y` sweep (the empirical answer to "any other calendar-dependent failures") is running against unmodified `dev` as the baseline; I'll post the residual list as a comment.

Not fixed, noted: `test_ent96_timeline_split.py` has no entry in `tests/registry.json`. Pre-existing and unrelated to this issue, so I left it rather than widen the diff.

Fixes #2243

### DIFF
--- a/tests/clockshift.py
+++ b/tests/clockshift.py
@@ -31,6 +31,19 @@ Known limits, stated so a green run is not over-read: it moves
 database's own clock (`CURRENT_TIMESTAMP`, `now()`), nor any C-level time source
 a dependency reads directly, so a test whose expectations come from SQL-side
 time is out of its reach.
+
+One false-positive class is handled rather than documented, because a sweep whose
+output has to be triaged by hand is not a verdict. `datetime` cannot be patched
+in place (a C type with immutable attributes), so the shift installs a subclass —
+which splits the class identity: a library imported AFTER the swap type-checks
+against the subclass, while a value produced by code that imported `datetime`
+BEFORE it is an instance of the original. PyJWT does exactly this
+(`isinstance(payload["exp"], datetime)` in `api_jwt.encode`, guarding the
+datetime→epoch conversion), so under a naive swap ~30 auth/session tests failed
+with `TypeError: Object of type datetime is not JSON serializable` — a harness
+artifact indistinguishable, in a failure list, from a real calendar bug. The
+metaclass below makes `isinstance(any_real_datetime, _ShiftedDatetime)` true, so
+both identities satisfy such a check.
 """
 from __future__ import annotations
 
@@ -44,7 +57,23 @@ SHIFT = _dt.timedelta(days=_DAYS)
 _real_datetime = _dt.datetime
 
 
-class _ShiftedDatetime(_real_datetime):
+class _AnyDatetime(type):
+    """Make `isinstance(x, _ShiftedDatetime)` true for ANY real `datetime`.
+
+    Without this the swap narrows a widely-used type test: a real `datetime`
+    (built by a module that bound the name before the swap) is not an instance of
+    the subclass, so a library checking `isinstance(value, datetime)` silently
+    takes its else-branch. See the module docstring for the PyJWT case.
+    """
+
+    def __instancecheck__(cls, obj):
+        return isinstance(obj, _real_datetime)
+
+    def __subclasscheck__(cls, sub):
+        return issubclass(sub, _real_datetime)
+
+
+class _ShiftedDatetime(_real_datetime, metaclass=_AnyDatetime):
     """`datetime` with the three "what time is it" constructors moved.
 
     A subclass, not a stub: instances stay real `datetime`s, so comparisons and
--- a/tests/clockshift.py
+++ b/tests/clockshift.py
@@ -1,0 +1,74 @@
+"""Run the suite as if the machine clock were a year ahead (#2243).
+
+    PYTHONPATH=tests python -m pytest tests/unit -q -p clockshift
+
+(`PYTHONPATH=tests` because `-p` resolves plugins before pytest has put `tests/`
+on `sys.path`, and this file deliberately is not a `conftest` — it must not load
+for runs that did not ask for it.)
+
+Why this exists. Three sets of tests were permanently red on `dev` for one
+reason: they asserted against literal calendar strings — an hour bucket
+(`"2026-08-14T09"`) that the gap-filled axis stops reaching, and a `started_at`
+anchor (`"2026-01-01T00:00:00.000000Z"`) whose `duration_ms = now − started_at`
+grew past int32 and raised `NumericValueOutOfRange` on PostgreSQL. Both passed
+the day they were written and failed every day after. A grep can find date-shaped
+strings, but it cannot tell an inert fixture value from one that is compared
+against *now* — so the class is closed by running the clock forward and reading
+the failures, not by eyeballing literals.
+
+Not a test dependency and not a `freezegun`: it shifts the two clocks test code
+actually reads and adds nothing to `requirements`. Deliberately opt-in — it is a
+diagnostic for "is this suite date-independent?", not something CI runs by
+default.
+
+Loaded with `-p` so it lands BEFORE test modules run their
+`from datetime import datetime`, which is what makes the shift visible to code
+that binds the name at import time. Override the distance with
+`CLOCKSHIFT_DAYS` (e.g. `CLOCKSHIFT_DAYS=-400` to run a year in the past).
+
+Known limits, stated so a green run is not over-read: it moves
+`datetime.datetime.now/utcnow/today` and `time.time`. It does NOT move a
+database's own clock (`CURRENT_TIMESTAMP`, `now()`), nor any C-level time source
+a dependency reads directly, so a test whose expectations come from SQL-side
+time is out of its reach.
+"""
+from __future__ import annotations
+
+import datetime as _dt
+import os
+import time as _time
+
+_DAYS = int(os.environ.get("CLOCKSHIFT_DAYS", "365"))
+SHIFT = _dt.timedelta(days=_DAYS)
+
+_real_datetime = _dt.datetime
+
+
+class _ShiftedDatetime(_real_datetime):
+    """`datetime` with the three "what time is it" constructors moved.
+
+    A subclass, not a stub: instances stay real `datetime`s, so comparisons and
+    arithmetic against values parsed from the database keep working.
+    """
+
+    @classmethod
+    def now(cls, tz=None):
+        return _real_datetime.now(tz) + SHIFT
+
+    @classmethod
+    def utcnow(cls):
+        return _real_datetime.utcnow() + SHIFT
+
+    @classmethod
+    def today(cls):
+        return _real_datetime.today() + SHIFT
+
+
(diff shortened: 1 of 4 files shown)


dolho:
## Full-suite sweep result — AC5, with the residual named

Both legs finished (~54 min each, clean `origin/dev` vs this branch):

| leg | failures |
|---|---|
| `dev`, normal clock | **23** |
| this branch, clock **+365 days** | **19** |

### The 23 on dev decompose exactly

| file | n | class |
|---|---|---|
| `test_ent96_timeline_split.py` | 5 | calendar rot — fixed here |
| `test_pat_propagation_properties.py` | 3 | stale strict-xfail — fixed here |
| `test_736_a2a_outbound_edges.py` | 5 | environment |
| `test_mcp_validator.py` | 4 | environment |
| `test_ent14_registry_url_ssrf.py` | 4 | environment |
| `test_ent399_ipv6_origin.py` | 2 | environment |

The 8 is exactly the set this workflow's own JUnit artifacts report on `dev` under all three seeds, and this branch takes CI to **0** (verified from the artifacts of run 32012207851: base union 8, head union 0, 35,229 testcases). The 15 are local Python 3.12 mapped-address behaviour — they fail with **no** clock shift at all and do not fail on CI's 3.13.

### The +1y leg: 15 expected, plus 4 I did not predict

```
test_1595_git_maintenance.py::TestReapStaleGitLitter::test_fresh_locks_kept
test_2216_backup_primitives.py::TestTmpSweep::test_sweeps_aged_orphan_keeps_fresh
test_2216_db_backup_service.py::TestPreflightAndFailure::test_prune_runs_even_when_backup_fails
test_2216_db_backup_service.py::TestPreflightAndFailure::test_prune_runs_on_no_space_skip_too
```

All four are **the instrument, not the repo**, one root cause, and it breaks in both directions — the harness shifts the process clock but not the **filesystem's**:

- `test_fresh_locks_kept` / `TestTmpSweep` write a file *now* (real mtime) and expect it judged fresh; against a shifted `now` it looks a year old and is reaped
- the two `test_prune_runs_*` are the mirror: an artifact meant to be out-of-window looks freshly created, so the prune correctly declines and the assertion fails

Those four tests are correct. Recorded in `tests/clockshift.py` with each direction named (bcedefee), because four unexplained failures in a sweep's output are four false leads for whoever reads it next.

### So, honestly

- **Zero repo-side calendar rot remains at +365 days.** Every failure is either environment-specific or outside what the harness can see.
- **AC5 is satisfied for clock-derived tests and unproven — not disproven — for mtime-driven sweeps.** Shifting inode times means intercepting `os.stat`, a much larger instrument than the problem justifies, and it would put the harness itself under suspicion.

Earlier in this PR I also reported the naive-swap run's 74 failures; ~30 of those were a PyJWT `isinstance(x, datetime)` artifact, fixed by the metaclass in 3ccbac0e. The numbers above are from the corrected harness.

vybe (approved):
Validated via /validate-pr.

Test-only. Replaces calendar-literal fixtures with clock-derived ones and retires four stale strict-xfails whose underlying bugs were fixed by #2017/#2024 — the three backslash cases now assert the round trip rather than merely 'does not raise', which is the actual contract. Ships `tests/clockshift.py`, the harness #2249 builds on.

- Security scan: clean
- All checks green across three seeds, both sides
- Verified prerequisite ordering: #2249 carries these three commits, so this lands first
- Closing keyword patched in so #2243 auto-promotes

Issue #715, opened 2026-05-07
test: gate dev → open-PR merges on the unit-test suite


AndriiPasternak31:
## Why

PR #714 (closes #660 — *green the unit-test suite*) had to grow scope twice
because of dev-merge residuals only visible at PR review time:

1. **First scramble** (#714 v1): 5 unit-test files regressed silently when dev
   landed schema/import/path changes between this branch's base and merge time.
   Caught at review — fixed in 3df1d5dd.
2. **Second observation** (during the v1 fix): the broader Trinity unit suite
   has ~34 failures + 17 collection errors that pass individually but fail
   under the polluted full-suite import order. Files include
   `test_voice_tools`, `test_monitoring_router_signatures`,
   `test_skill_service_user_agent`, `test_sync_health_service`,
   `test_fleet_status_resilience`, `test_webhook_rate_limit_*`,
   `test_login_rate_limit_split`, `test_terminate_async_executor`,
   `test_inject_assigned_credentials`. Pre-existing on dev — not introduced
   by #714.

Both issues share a root cause: nothing runs the **full** unit suite in the
order CI runs it on a representative dev → open-PR-branch merge. Local devs
run individual files. CI runs the suite at PR-time. The gap between dev's
last green-on-CI state and an open PR's mergeability is invisible until
review.

## What to add

Two pieces of plumbing:

1. **Per-PR full-suite gate.** GitHub Actions workflow on `pull_request`
   triggers: run `uv run pytest tests/unit/ -q` against
   `merge(target, head)`. Fail the PR check if anything regresses vs the
   target branch's run. Already partially in place via existing
   `frontend-e2e` for UI labels — extend the pattern to backend unit tests.

2. **Nightly dev-vs-open-PRs run.** Periodic job that, for each open PR
   targeting `dev`, attempts the merge into a scratch branch and runs the
   suite. Surface red as a comment on the PR. Catches the residual-failure
   class earlier than waiting for the next reviewer.

Bonus: install `pytest-randomly` and run with 3 fixed seeds. The current
suite has order-dependent pollution (see point 2 above) — exposing it
deliberately at CI time gives us pressure to fix the root cause instead
of the symptom.

## Out of scope (for whoever picks this up)

- Actually fixing the 34 failed + 17 errored test files. That's a separate
  follow-up; this issue is just the CI gate to surface them earlier.

## Related

- PR #714 (closes #660)
- Commit 3df1d5dd (dev-merge residual greens)
[...]

PR #2249, opened 2026-08-17
fix(ci): make a wholly-red suite reportable, and detect calendar rot on the spot (#2247)

### LINKED ISSUE #2247: test_ent96_timeline_split has been red on dev since 2026-08-15 — stale date fixture, and no CI job can report it
## Summary

All five tests in `tests/unit/test_ent96_timeline_split.py` have been failing on `dev` since **2026-08-15**, and no CI job reports it.

```
FAILED tests/unit/test_ent96_timeline_split.py::test_split_rows_fold_to_one_row_per_interval
FAILED tests/unit/test_ent96_timeline_split.py::test_two_raw_triggers_in_one_bucket_merge_into_that_bucket
FAILED tests/unit/test_ent96_timeline_split.py::test_failures_are_carried_per_label_not_only_in_the_column_total
FAILED tests/unit/test_ent96_timeline_split.py::test_the_breakdown_always_sums_to_its_own_column
FAILED tests/unit/test_ent96_timeline_split.py::test_an_unmapped_trigger_lands_in_Other_rather_than_vanishing
```

Reproduces on a clean `origin/dev` checkout: `python -m pytest tests/unit/test_ent96_timeline_split.py`

## Root cause — a stale-fixture time bomb

The fixtures hardcode an absolute date and then ask for a **relative** 24-hour axis:

```python
rows = [
    _row("2026-08-14T09", "schedule", 10),
    _row("2026-08-14T09", "mcp", 4),
    _row("2026-08-14T10", "schedule", 2),
]
out = _shape_execution_timeline(rows, group_by="hour", hours=24, split="trigger")
assert by_bucket["2026-08-14T09"]["total"] == 14   # KeyError
```

`hours=24` gap-fills a continuous UTC axis over `[now-24h, now]`. The file landed on **2026-08-14** (#2184), so the fixture bucket sat inside the window that day and the tests passed. From 2026-08-15 onward `2026-08-14T09` is outside the axis, the gap-fill drops it, and every assertion keyed on that bucket raises `KeyError`.

Demonstrated directly — same rows, same function, only the window varies:

```
hours=  24 -> fixture bucket present: False
hours= 168 -> fixture bucket present: True
hours= 720 -> fixture bucket present: True
```

**The product code is fine.** Gap-filling a relative window is exactly what `_shape_execution_timeline` is specified to do (architecture.md: `hour`/`day` are gap-filled on a continuous UTC axis). Only the test is wrong.

## Why CI never said anything

Two mechanisms combine so this is invisible rather than merely tolerated:

1. **On PRs**, the unit suite runs with `|| true` and `regression diff` compares base against head. Both sides fail identically, so it reports no regression — green.
2. **On pushes to `dev`**, the diff job does not run at all (`if: github.event_name != 'push'`), so nothing evaluates the result.

The failure is therefore structurally unobservable: it will not surface on any PR, and it will not surface on any `dev` push. It was found by running the suite locally while validating an unrelated PR (#2232).

This is the more interesting half of the bug. A test that can rot into permanent red without a single red check is worth more than the five assertions it contains.

## Fix

Make the fixture relative to the same clock the function uses — derive the bucket keys from `utc_now_iso()`-based arithmetic instead of literals, so a 24h-window test always places its rows inside a 24h window. (Widening to `hours=168` would also go green, but it silently stops testing the 24h axis the ent#96 tile actually ships.)

## Acceptance criteria

- [ ] `tests/unit/test_ent96_timeline_split.py` passes on `dev` and keeps passing on any future date (no absolute date literals in fixtures whose assertions depend on a relative window)
- [ ] A test that would have caught the rot: the suite is exercised at a simulated later "now" (e.g. frozen clock a week on) and still passes
- [ ] Separately, decide whether a whole-suite red on `dev` should be reportable at all — today neither the PR diff job nor the push run can see it. A cheap option is a scheduled or push-triggered absolute-failure-count check against a committed baseline, distinct from the base-vs-head diff.

dolho:
> **Stacked on #2244** (issue #2243), which owns the fixture fix and ships `tests/clockshift.py`. Merge that first; until it lands this PR's diff shows its two commits too.

#2243/#2244 fixes the five rotted fixtures. This is the half #2247 calls *the more interesting one*: why nothing reported it, and what would have caught it.

## The two gaps, both structural

`backend-unit-test.yml` runs pytest with `|| true` and gates on a **base-vs-head diff** — when both sides fail identically the diff is empty and the check is green. And on a push to `dev`/`main`, `test` and `diff` are both `if: github.event_name != 'push'`, so nothing evaluates the suite at all.

A test can therefore rot into permanent red without a single red check: invisible on PRs by construction, invisible on pushes by construction.

## What the numbers actually are

I pulled the JUnit artifacts from a **real run of this workflow** rather than estimating:

| side | failing (union of 3 seeds) | testcases |
|---|---|---|
| `dev` (base) | **8** — the 5 ent#96 + 3 stale strict-xfails | 35,229 |
| #2244 branch (head) | **0** | 35,229 |

So the `|| true` comment citing *"the existing 34 failures + 17 errors documented in #660"* was years of drift. Corrected in this PR.

That measurement **changes the design the issue proposed.** With zero reachable, the cheapest gate that cannot itself rot is "no failures at all", not a failure *count*. `tests/unit_failure_baseline.txt` therefore ships **empty**: the ratchet mechanism exists (precedent: `lint_sys_modules_baseline.txt`) so a genuinely-expected failure can be tolerated deliberately — but a tolerated-failure list is where debt hides, so it starts empty, every entry has to be argued for in review, and the gate *reports entries that have started passing* so the file shrinks rather than grows.

## The gate

`scripts/ci/check-failure-baseline.py` + a new push-only `absolute` job:

- **shares test identity** with `diff-pytest-failures.py` `(classname, name, kind)`, so a test cannot be baselined under one spelling and reported under another
- distinguishes `failure` from collection `error` — "known failure flips to collection error" is a real regression
- **fails closed** on missing / unparseable / zero-testcase JUnit (#715's lesson: infra failure reporting green is the failure mode the gate exists to prevent)
- `--self-test` covers all nine behaviours, and the workflow runs it *before* trusting the verdict
- **one run, default order** — seeds catch order-dependent regressions *between commits*, which stays a PR-side property; tripling every dev push buys nothing, and this job should stay cheap enough that nobody deletes it

Verified against the real artifacts, both directions:

```
dev (base):  failing (union): 8 · baselined: 0  →  FAIL, naming all 8 ids
#2244 head:  failing (union): 0 · baselined: 0  →  PASS
```

## Detecting it at the source

`tests/unit/test_2247_calendar_rot_detector.py` re-runs the calendar-sensitive files under `clockshift.py` at **+7 days** and **−365 days**, in the ordinary unit suite, so a fixture pinned to a literal date inside a relative window fails on the first run rather than on a future Tuesday.

- subprocess, because the shift must land before the module binds `from datetime import datetime` — an autouse fixture cannot reach an already-imported module
- it guards **its own no-op failure mode**: asserts the plugin exists *and* that importing it actually moves `datetime.now()`. A vanished plugin would otherwise make every shifted run pass by shifting nothing
- **proven to detect, not merely to pass:** reverting the ent#96 fixture to its pre-#2243 state turns both directions red with the message naming the cause; restoring it turns them green. A rot detector never shown going red is decoration
- ~37s, nearly all pytest boot — which is why the list is short and excludes the ~176s `test_1771c_schedules_analytics_edges.py` (same property, second sample, 5× the cost)

Python pinned to **3.13** to match the sibling jobs, deliberately rather than by copy: the observed failing set is interpreter-dependent (locally on 3.12 the SSRF/IPv6 suites fail; on CI's 3.13 they don't), so a gate on a different Python would judge against a baseline collected under different rules.

## Acceptance criteria

- [x] fixtures pass on any future date — **#2244** (this PR's base), verified there at +365d
- [x] a test that would have caught the rot, exercised at a simulated later "now" — the detector above, shown red against the original fixture
- [x] decide whether a whole-suite red on `dev` should be reportable — **yes**, implemented as the push-only `absolute` job with an empty committed baseline; reasoning recorded in the workflow comment and the script docstring

Fixes #2247


vybe (approved):
Validated via /validate-pr.

This is the more important half of the pair: #2244 fixed the rotted fixtures, this explains why nothing reported them. The gap is real and correctly stated — the per-PR gate compares BASE against HEAD, so a suite failing identically on both sides produces an empty diff and a green check, and on a push both the `test` and `diff` jobs skip entirely. A test can rot into permanent red with no red check anywhere, which is exactly what ent#96 did for two days.

The mechanism is well chosen: absolute comparison against a committed baseline, fail-closed on missing or empty JUnit (a run that produced nothing must not read as 'nothing failed' — #715's lesson), identity deliberately kept in step with `diff-pytest-failures.py` so a test cannot be baselined under one spelling and reported under another, and a checker self-test inside the job. The baseline ships EMPTY on measured evidence rather than aspiration — real JUnit artifacts show dev failing exactly 8 under all three seeds, and 0 with #2243 applied.

The calendar-rot detector is the right complement: a subprocess (the shift must be installed before the module binds `from datetime import datetime`, which an autouse fixture cannot do), and it guards its own no-op failure mode by asserting the plugin actually moves `datetime.now()` — a vanished plugin would otherwise make every shifted run pass by shifting nothing.

**Verified the empty baseline is achievable, not aspirational**: ran the full unit suite locally on this branch merged with dev — **11769 passed, 29 skipped, 0 failed** (9m11s). So the new push gate lands green rather than instantly red.

- Security scan: clean
- Ordering respected: stacked on #2244, which merged first; after rebasing, this PR's diff is exactly its own 5 files
- Merged `origin/dev` and rebuilt `tests/registry.json` from git stages (never spliced)
- Deliberately merged LAST of the batch so the gate switches on against a fully-cleaned dev
- Closing keyword patched in so #2247 auto-promotes

PR #2250, opened 2026-08-17
fix(tests): re-point five stale mocks at the seams they claim to stub (#2242)

### LINKED ISSUE #2242: bug: stale test mocks drifted from #1804, #2015 and ent#237 — 9 failures of lost coverage
## Summary

Nine tests fail because their mocks/fixtures still describe an older shape of the code they stub. In every case the production code is correct and the test is stale — but each one silently stopped testing what its name claims, so the failures are lost coverage rather than noise.

**1. Circuit-breaker fast-fail (3 tests)** — `tests/test_cb_probe_execution_close.py::TestCircuitBreakerFastFail` (`test_cb_open_fails_execution_record`, `…_does_not_call_agent`, `…_does_not_mark_dispatched`):
```
TypeError: object MagicMock can't be used in 'await' expression
  at src/backend/services/task_execution_service.py:856 in _write_terminal_and_gate
```
The mock db predates the async call #1804 added to the terminal path (the `close_execution_activity` / terminal-event seam), so a plain `MagicMock` is returned where the code awaits.

**2. Audit hash chain (2 tests)** — `tests/test_audit_log_unit.py::test_hash_chain_produces_hashes` (`assert None is not None`) and `::test_verify_chain_valid_with_details` (`assert 0 == 2`). The `audit_service` fixture builds a fake `database.db` wiring only `create_audit_entry`, `get_audit_entries`, `get_audit_entries_range`. #2015 moved the chain from an in-process flag to durable `system_settings`, so it now needs `set_setting` / `get_setting` / `create_audit_entry_chained`. The MagicMock swallows the write, the (correct, deliberate) fail-closed reader answers "off", and nothing is ever hashed.

**3. GitHub PAT propagation (4 tests)** — `tests/test_github_pat_propagation_unit.py`, all four failing with `cannot import name 'AgentGitConfig' from 'database' (unknown location)` surfacing as empty `updated` lists.

**4. Skills-library sync (1 test)** — `tests/test_skills.py::TestSkillsLibrarySync::test_sync_requires_url_configured` expects `400` for "no library URL configured", but ent#237 replaced the single `skills_library_url` setting with the multi-source `skill_sources` table. A migrated source now exists, sync succeeds, and the endpoint correctly returns `200`. The test asserts a contract that no longer exists.

**5. PostgreSQL tier isolation (1 error)** — `test_migration_is_conditional_inside_the_container` fails at setup with `UniqueViolation: duplicate key value violates unique constraint "users_pkey", Key (id)=(1)`. The tier runs many tests against one disposable database and this fixture inserts a fixed-id user a prior test already seeded.

## Context

Found by a full-suite run on `dev` (`691e7dc9`): unit 8 failed / api 21 failed + 14 errors / postgres 15 failed + 2 errors. This issue covers the stale-mock cluster; the `assert_*`/MagicMock collision and the calendar-expiring tests are filed separately.

The pattern worth naming: each of these is a **test that keeps passing an outdated contract** until the stubbed seam changes shape. #1804 and #2015 both moved a seam and left the mock behind, so the tests went red only when someone ran the full suite.

## Acceptance Criteria

- [ ] `TestCircuitBreakerFastFail` (3 tests) pass — the db mock provides the awaitable(s) `_write_terminal_and_gate` now uses
- [ ] `test_hash_chain_produces_hashes` and `test_verify_chain_valid_with_details` pass against the #2015 persisted-setting design (fixture wires `set_setting`/`get_setting`/`create_audit_entry_chained` to the temp DB)
- [ ] The 4 `test_github_pat_propagation_unit.py` tests pass — the `database` stub exposes `AgentGitConfig`
- [ ] `test_sync_requires_url_configured` is rewritten for the ent#237 multi-source model, or deleted with a note saying which test now covers "no sources configured"
- [ ] The PostgreSQL tier no longer fails on `users_pkey` — the fixture does not assume an unseeded database
- [ ] Each fix asserts the *current* contract, not a re-mocked version of the old one

## Technical Notes

- The audit fixture is the most delicate: #2015's fail-closed read is deliberate (`hash_chain_enabled` answers "off" on a settings-read failure because a settings error is not evidence an integrity record exists). The fixture must make settings genuinely readable rather than force the flag on.
- Reproduce: `cd tests && source .venv/bin/activate && source setup-env.sh && python -m pytest test_cb_probe_execution_close.py test_audit_log_unit.py test_github_pat_propagation_unit.py -q`
- PostgreSQL tier: `tests/run-full.sh --tier postgres`

dolho:
Nine failures, none of them a product defect — and every one a test that had quietly stopped asserting its own name. Each fix targets the **current** contract; where the old assertion no longer describes anything real, it is replaced rather than re-mocked.

## 1. Circuit-breaker fast-fail (3)

`#1804` added `await activity_service.close_execution_activity(...)` to `_write_terminal_and_gate`, and the activity double was inlined at **seven** call sites wiring only `track_activity`/`complete_activity` — so the new await got a plain `MagicMock`. One `_activity_double()` factory now serves all seven, so the next seam change is one edit instead of seven silent drifts.

The fast-fail test also now asserts #1804's actual rule — **the CAS winner owns the close**. Restoring the await is not the same as testing it: a fast-fail that writes FAILED and orphans the activity is precisely the bug #1804 fixed, and it would satisfy every other assertion in that test.

## 2. Audit hash chain (2)

`#2015` moved the toggle from an in-process attribute to a durable `system_settings` row. The fixture now creates that table (schema.py's DDL) and wires `set_setting`/`get_setting_value` through a real engine-bound `SettingsOperations` on the same temp DB, plus the real `create_audit_entry_chained`.

Deliberately **not** `get_setting_value.return_value = "true"` — forcing the flag on would bypass the fail-CLOSED reader, which is the behaviour least worth faking (the issue's own technical note). And the chained writer is the real one because the chain head is read *inside* the insert transaction: a stub would have to reimplement the part that makes the chain a chain.

## 3. GitHub PAT propagation (4)

`_propagate_to_agent` lazily runs `from services import git_service`, whose header is `from database import db, AgentGitConfig, GitSyncResult`. A stub carrying only `db` raised `ImportError` **inside** `asyncio.gather(return_exceptions=True)`, so it surfaced as every agent landing in `failed` — four tests named after propagation outcomes were asserting an import failure.

Fixed with the real pydantic models plus one more docker symbol, so the **real** git_service loads and `update_remote_pat`'s signature is genuinely covered. Stubbing git_service itself would have let that drift unnoticed — this issue's own bug class.

**Fixing the import then exposed two things it had been hiding:**

- `assert all("a1" not in str(c) ...)` searched the *whole call repr* for an agent name, and that repr now carries the #1159 `X-Trinity-Agent-Token` — a 64-char HMAC hex containing "a1" about half the time (observed: `...db83a1d7ab65a1a2239d...`). It only looked stable because no request was ever made, so the generator was vacuously true over an empty list. Now it asserts the URL.
- `db` as a blanket MagicMock made `get_git_config(...).github_repo` truthy, so **#1967's eligibility signal silently flipped**: the "skips an agent without GITHUB_PAT" test was exercising the WRITE path while asserting the skip. It now states its condition (no git config), and a **new test** covers the other branch — an agent Trinity manages a repo for gets the line created — which nothing covered before.

## 4. Skills-library sync (1)

ent#237 replaced `skills_library_url` with the `skill_sources` table, so `status["url"]` is `None` on every modern install — *including ones with sources*. The guard `if status.get("url"): skip` stopped firing, sync ran against the migrated source, succeeded, and the test failed asserting 400 against 200.

Swapping the key does not repair it: an API test cannot dictate the target instance's configuration — observed live, `/status` reported `configured: false` moments before `/sync` found an enabled source and cloned it. So:

- the deterministic contract moved to a unit test that owns its configuration: `test_ent237_skill_sources.py::TestSyncBookkeeping::test_sync_with_no_sources_configured_fails_by_name`, asserting the error string the router returns verbatim as `detail`
- the API test now checks what holds on **any** instance: the route degrades by name and never 500s
- its sibling `test_sync_returns_structure_when_configured` gated on the same dead key and had been skipping **unconditionally** — it now gates on `configured`

## 5. PostgreSQL tier (1 error)

`portal_db` was written for its own temp SQLite file, but `get_engine()` honours `DATABASE_URL` and the PG tier sets it — so the fixture seeded `users(id=1)` into the tier's **one shared** database and hit `UniqueViolation: users_pkey` against a user an earlier test had inserted, failing in *setup*.

Seeding is now present-or-insert, and the fixture clears its own `agent_sharing` rows at both ends — the two-claimant test compares an exact tuple, so leftover shares would break it on a shared DB even after the pkey clash was gone. The failing test also **never touched the database at all** (it base64-decodes a shell script), so it and one sibling no longer request the fixture. It reaches that tier only because the tier selects on `-k "…or migration or…"` and these test names contain "migration" — an incidental keyword match, not a dual-backend claim.

## Verification

| what | result |
|---|---|
| all five affected files together | **136 passed** (SQLite + the live instance for the API tier) |
| `test_ent237_skill_sources.py` with the new case | 95 passed |
| **PG cluster, proven the hard way** | reproduced on unfixed `dev` against a deliberately pre-seeded shared PostgreSQL → **3 setup errors**; then **23 passed** on this branch against that same seeded database, **twice in a row** to prove the share-row cleanup |
| `tests/lint_sys_modules.py` | clean — 140/206 baseline, one violation removed |
| `tests/lint_root_test_placement.py` | clean |

Fixes #2242

### DIFF
--- a/tests/test_audit_log_unit.py
+++ b/tests/test_audit_log_unit.py
@@ -97,6 +97,21 @@ def audit_db(monkeypatch):
         END
         """
     )
+    # #2242/#2015: the hash-chain toggle is a durable `system_settings` row now,
+    # not an instance attribute. Without this table the fixture's `db` was a
+    # MagicMock that swallowed the write, the (deliberately fail-CLOSED) reader
+    # answered "off", and nothing was ever hashed — so two tests named after the
+    # hash chain were asserting that a disabled chain produces no hashes.
+    # Same DDL as db/schema.py's `system_settings`.
+    cursor.execute(
+        """
+        CREATE TABLE system_settings (
+            key TEXT PRIMARY KEY,
+            value TEXT NOT NULL,
+            updated_at TEXT NOT NULL
+        )
+        """
+    )
     conn.commit()
     conn.close()
 
@@ -365,6 +380,26 @@ def audit_service(audit_db, monkeypatch, audit_ops):
     fake_db.create_audit_entry = audit_ops.create_audit_entry
     fake_db.get_audit_entries = audit_ops.get_audit_entries
     fake_db.get_audit_entries_range = audit_ops.get_audit_entries_range
+    # #2242: the three seams #2015 added, wired to the SAME temp DB rather than
+    # left as MagicMocks.
+    #
+    # `create_audit_entry_chained` is the real one because the chain head is read
+    # inside the insert transaction — a stub would have to reimplement exactly the
+    # part that makes the chain a chain.
+    #
+    # Settings go through a real `SettingsOperations` (engine-based, so it binds to
+    # the temp file via the DATABASE_URL the audit_db fixture sets) so
+    # `enable_hash_chain()` genuinely persists and `hash_chain_enabled` genuinely
+    # reads it back. Deliberately NOT `fake_db.get_setting_value.return_value =
+    # "true"`: that would force the flag on and bypass the fail-closed reader the
+    # #2015 design turns on, which is the one behaviour most worth not faking.
+    monkeypatch.delitem(sys.modules, "db.settings", raising=False)
+    from db.settings import SettingsOperations
+
+    settings_ops = SettingsOperations()
+    fake_db.create_audit_entry_chained = audit_ops.create_audit_entry_chained
+    fake_db.get_setting_value = settings_ops.get_setting_value
+    fake_db.set_setting = settings_ops.set_setting
     fake_db_module.db = fake_db
     monkeypatch.setitem(sys.modules, "database", fake_db_module)
 
--- a/tests/test_cb_probe_execution_close.py
+++ b/tests/test_cb_probe_execution_close.py
@@ -128,6 +128,25 @@ def _restore_complete_stubs(monkeypatch):
 # Fix A: Circuit breaker fast-fail
 # ===========================================================================
 
+def _activity_double(*, activity_id=None):
+    """The activity-service seam, with EVERY awaited method awaitable.
+
+    #1804 added `close_execution_activity` to the terminal path
+    (`task_execution_service._write_terminal_and_gate`) — and this double was
+    inlined at SEVEN call sites, each wiring only `track_activity` and
+    `complete_activity`. So the new await got a plain `MagicMock` and three tests
+    died with "object MagicMock can't be used in 'await' expression": not noise,
+    but three fast-fail assertions that had quietly stopped running (#2242).
+
+    One factory, so the next seam change is one edit instead of seven silent
+    drifts. Anything the service awaits belongs here.
+    """
+    return MagicMock(
+        track_activity=AsyncMock(return_value=activity_id),
+        complete_activity=AsyncMock(),
+        close_execution_activity=AsyncMock(),
+    )
+
 class TestCircuitBreakerFastFail:
     """execute_task closes the execution record immediately when CB is open."""
 
@@ -183,9 +202,7 @@ class TestCircuitBreakerFastFail:
         mock_capacity.acquire = AsyncMock(return_value=admitted)
         mock_capacity.release = AsyncMock()
 
-        mock_activity_svc = MagicMock()
-        mock_activity_svc.track_activity = AsyncMock(return_value="act-001")
-        mock_activity_svc.complete_activity = AsyncMock()
+        mock_activity_svc = _activity_double(activity_id="act-001")
 
         mock_circuit = MagicMock()
         mock_circuit.allow_request.return_value = False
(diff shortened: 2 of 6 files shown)


vybe (approved):
Validated via /validate-pr.

Test-only change; no product behaviour touched. Re-points five stale doubles at the seams they actually stub (#1804 `close_execution_activity`, #2015 durable `system_settings` toggle, ent#237 source rows) and consolidates the seven inlined activity doubles onto one factory, so the next seam change is one edit rather than seven silent drifts.

- Security scan: clean (no keys, emails, IPs, host paths, mode/symlink changes)
- Base `dev`; all checks green across three seeds, both sides
- Closing keyword patched in so #2242 auto-promotes to status-in-dev

PR #2251, opened 2026-08-17
fix(git): let repo discovery cross a mount boundary, so a home-rooted repo is found (#2245)

### LINKED ISSUE #2245: `_git_toplevel` misses a home-rooted repo when `workspace/` is a separate mount (no GIT_DISCOVERY_ACROSS_FILESYSTEM)
## Problem

`_git_toplevel` (shipped in #2077 for #2075) asks git for the agent's repo root:

```python
"git -c safe.directory='*' -C \"$start\" rev-parse --show-toplevel 2>/dev/null"
```

`$start` is `/home/developer/workspace` when that directory exists, `/home/developer` otherwise.

**Git stops repository discovery at a filesystem boundary by default.** If an agent has `/home/developer/workspace` as a separate mount — a bind mount, a distinct volume, an overlay — then `rev-parse --show-toplevel` started inside it will **not** walk up into `/home/developer`, even when the repository is genuinely rooted there.

`_git_toplevel` then returns `None`, and `_detect_git_dir` falls through to `_detect_git_dir_fallback` — the pre-#2075 content heuristic. On that topology, a populated `workspace/` answers `/home/developer/workspace`, which is exactly the misclassification #2075 exists to eliminate. Every path-based consumer (the compatibility collector, the `.gitignore` migration and its `.git` guard, the compatibility fix endpoint) then reads and writes a subdirectory's files as if they were the repo root's.

The fallback is correct as a *no-repository* path. The defect is that a boundary makes a repository that *does* exist look like it does not.

## Fix

Set `GIT_DISCOVERY_ACROSS_FILESYSTEM=1` for this probe:

```
GIT_DISCOVERY_ACROSS_FILESYSTEM=1 git -c safe.directory='*' -C "$start" rev-parse --show-toplevel
```

Safe here specifically because `_git_toplevel` already refuses any answer outside `/home/developer` — discovery may cross the boundary, but the result is still confined before it is trusted.

## Provenance

Spotted in #2076, a competing implementation of the #2075 fix that carried this flag and was closed as superseded by #2077. #2077 is the better change overall (docs + a 251-line test file); this one flag is the single thing it does not have. Recorded so the idea is not lost with the closed PR.

## Acceptance criteria

- [ ] `_git_toplevel` sets `GIT_DISCOVERY_ACROSS_FILESYSTEM=1`
- [ ] A regression test in `tests/unit/test_2075_detect_git_dir.py` covers the boundary case: git answers empty/non-zero from inside `workspace/` while a home-rooted repo exists, and the result is `/home/developer` rather than the fallback's `workspace/`
- [ ] The containment check (answer must be `/home/developer` or beneath) is unchanged and still pinned by a test

dolho:
One flag, plus the tests that make it stick.

## The bug

`_git_toplevel` (#2077, for #2075) starts inside `workspace/` when it exists and asks git to walk up. Git halts repository discovery at a filesystem boundary by default — and says so in as many words:

```
fatal: not a git repository (or any parent up to mount point /home/developer)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

So on an agent whose `workspace/` is its own mount (bind mount, distinct volume, overlay), the probe never reaches a repository rooted at the home directory. `_git_toplevel` returns None, `_detect_git_dir` falls through to the pre-#2075 content heuristic, and a populated `workspace/` answers `/home/developer/workspace` — the exact misclassification #2075 exists to eliminate, reintroduced by a mount instead of by a heuristic. Every path-based consumer then reads and writes a subdirectory as if it were the repo root: the compatibility collector, the #462 `.gitignore` migration *and its `.git` guard*, the compatibility fix endpoint.

Crossing is safe because the containment check is untouched — discovery may walk out of the mount, but an answer outside the agent home is still refused before it is trusted.

## Verified in the real topology, not argued

A unit test cannot create a mount (user namespaces are blocked in the sandbox, `mount` needs privilege), so I built the topology in a `trinity-agent-base` container with a tmpfs at `/home/developer/workspace` and a repo at `/home/developer`:

```
st_dev home      : 220
st_dev workspace : 1048621        <- a genuine boundary
pre-fix  probe   : exit=128, "Stopping at filesystem boundary ..."
post-fix probe   : exit=0,   "/home/developer"
```

Then the **shipped** script, extracted from the source so the string under test is the one that ships:

```
start=/home/developer/workspace; [ -d "$start" ] || start=/home/developer; \
GIT_DISCOVERY_ACROSS_FILESYSTEM=1 git -c safe.directory='*' -C "$start" rev-parse --show-toplevel
-> exit=0 output=[/home/developer]
```

And the direction that must **not** change — a genuinely workspace-rooted legacy repo inside the mounted workspace:

```
legacy workspace-rooted repo still wins: [/home/developer/workspace]
```

So the flag lets the walk continue past a boundary without letting it skip a nearer repository.

## Tests (3 new, in the file the AC names)

- **the plumbing, behaviourally**: a `git` shim on PATH records the environment git *actually received*. A static substring check on the script would pass on a version where the assignment landed after a pipe or inside the wrong quoting layer — both leave the flag unset exactly where it matters.
- **containment, separately pinned**: asserted over five paths a boundary-crossing walk could now produce (`/`, `/home`, `/home/developer-other`, `/etc`, `/opt/repo`), because that guard is the entire safety argument for the flag.
- **the failure mode, documented**: one test records the degraded answer a boundary-failed probe produces, so the comment explaining why the probe must not fail cannot go stale silently.

The container-verified boundary evidence is written into the file's comments, since it is the one part the test suite structurally cannot reproduce.

## Verification

| check | result |
|---|---|
| `test_2075_detect_git_dir.py` + `test_2069_gitignore_at_creation.py` | **40 passed** |
| real mount boundary, container | pre-fix exit 128 → post-fix `/home/developer` |
| legacy workspace-rooted repo | still `/home/developer/workspace` |
| `tests/lint_sys_modules.py` | clean (140/206 baseline, no new) |

Also registered the test file in `tests/registry.json` — a pre-existing omission from #2077, worth closing while adding coverage to it.

Fixes #2245

### DIFF
--- a/src/backend/services/git_service.py
+++ b/src/backend/services/git_service.py
@@ -1601,12 +1601,30 @@ async def _git_toplevel(container_name: str) -> Optional[str]:
     otherwise make git refuse to answer and silently drop the caller onto the
     fallback heuristic that this function exists to replace.
 
+    ``GIT_DISCOVERY_ACROSS_FILESYSTEM=1`` because the walk up has a second way to
+    stop that has nothing to do with repositories (#2245): git halts discovery at
+    a filesystem boundary by default, so on an agent whose ``workspace/`` is its
+    own mount — a bind mount, a distinct volume, an overlay — a probe started
+    inside it never reaches a repository rooted at the home directory. Git says so
+    in as many words: *"Stopping at filesystem boundary
+    (GIT_DISCOVERY_ACROSS_FILESYSTEM not set)"*, exit 128. This function would then
+    return None and the caller would fall through to the content heuristic, which
+    answers ``/home/developer/workspace`` for exactly that topology — the
+    misclassification #2075 exists to eliminate, reintroduced by a mount.
+
+    Crossing the boundary is safe here specifically because the containment check
+    below is unchanged: discovery may walk out of the mount, but an answer outside
+    the agent home is still refused before it is trusted. (Carried over from #2076,
+    a competing #2075 fix closed as superseded — this flag was the one thing it had
+    that #2077 did not.)
+
     Returns None when there is no repository at or above the probe point, or
     when git answers with a path outside the agent home (never trusted).
     """
     script = (
         f"start={shlex.quote(LEGACY_WORKSPACE_DIR)}; "
         f'[ -d "$start" ] || start={shlex.quote(AGENT_HOME_DIR)}; '
+        "GIT_DISCOVERY_ACROSS_FILESYSTEM=1 "
         "git -c safe.directory='*' -C \"$start\" rev-parse --show-toplevel "
         "2>/dev/null"
     )
--- a/tests/registry.json
+++ b/tests/registry.json
@@ -2310,6 +2310,18 @@
         "voice"
       ],
       "description": "Portal dictation gate (#2212). `stt_available` is a SEPARATE roster-card bit from `voice_available`: transcription needs only the platform ElevenLabs key, narration additionally needs an effective voice, so the single-bit design could not express \"can dictate, cannot narrate\" — which is why the mic rendered on a pure browser capability check and was a dead control on a provider-less instance. Asserts the truth table in both directions (key without voice → stt yes / tts no; no key → both no, failing closed like `voice_available`), that the model default is absent-not-present so an older payload hides the mic rather than rendering a dead one, and — the load-bearing one — that the card bit and `transcribe_portal_audio`'s own 404 gate are the SAME condition, since drift between what the client sees and what /stt will do IS the dead-control bug. Pure unit: no DB, no HTTP, no ElevenLabs."
+    },
+    {
+      "file": "unit/test_2075_detect_git_dir.py",
+      "feature": "#2075, #2245",
+      "added": "2026-08-17",
+      "categories": [
+        "backend",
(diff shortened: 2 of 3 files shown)


vybe (approved):
Validated via /validate-pr.

Correct root cause: git halts repo discovery at a filesystem boundary, so an agent whose `workspace/` is its own mount made a home-rooted repo look absent (exit 128) and fell back to the very content heuristic #2075 removed. `GIT_DISCOVERY_ACROSS_FILESYSTEM=1` is applied as a command-scoped prefix on the git invocation itself.

Crossing the boundary is safe here because the containment check is unchanged — discovery may walk out of the mount, but an answer outside the agent home is still refused before it is trusted, and the test asserts that over five paths a boundary-crossing walk could now produce.

- Security scan: clean
- All checks green including `prod-image-smoke`
- Honest scoping noted: a unit test cannot create a mount, so the boundary itself was verified in a container and recorded in comments; what is pinned is the plumbing (a git shim proving the env var git ACTUALLY received) plus the containment guard
- Closing keyword patched in so #2245 auto-promotes

PR #1683, opened 2026-07-17
refactor(auth): INV-8 imperative auth-guard family — consolidate inline auth wiring (#1310)

AndriiPasternak31:
## Summary

Fixes #1310 (INV-8, split from #654's architecture-invariant drift report; carved off from INV-14/#1308 **because it is security-sensitive**).

**Behavior-PRESERVING** consolidation of duplicated inline auth wiring in ~20 FastAPI routers behind a small **imperative-guard family** in `dependencies.py`. Every consolidated site enforces the *same* verdict (status + detail + admitted-principal set) it did before — proven behaviorally, not by an OpenAPI snapshot.

### The family (all raise **403**, access-first → self-uniform per #186)

| Helper | Wraps | Replaces |
|---|---|---|
| `assert_admin` | `_reject_connector_principal` + `role != "admin"` | inline admin + 4 router-local `require_admin` dupes |
| `assert_agent_access` | `_enforce_connector_scope` + `can_user_access_agent` | inline `can_user_access_agent` → 403 |
| `assert_agent_owner` | `_enforce_connector_scope(owner_op=True)` + `can_user_share_agent` (owner-or-admin; **NOT** delete-auth — no `is_system` guard) | inline `can_user_share_agent` → 403 |
| `assert_owns_or_admin` | `id != owner AND role != "admin"` | strict-self-**or-admin** session gates (voice/chat) |
| `assert_owns` | `id != owner` (**no admin bypass**) | strict-self session gate (public-link session) |

Each agent-name helper runs `_enforce_connector_scope` first, matching the path-dependencies' second fence (kills the two-convention divergence). `_narrow_to_agent` relocated `routers/executions.py` → `services/agent_service/helpers.py` (removes a router→router import edge).

**Preserve-403, not 404:** access-first inline handlers are already self-uniform (they check access before any existence lookup → no 404-then-403 enumeration oracle). Precedent: `schedules.py` `create_schedule` (#1445).

## Security hard-gates (the merge-blockers the review flagged — each verified in this branch)

- **Anti-widening (public.py):** `public.py:1134` is **strict-self, no admin bypass** → uses **`assert_owns`** (NOT `assert_owns_or_admin`, which would let an admin read another user's public-link session). ✅ verified
- **Anti-exfil (agent_files):** the agent-scoped-key defense-in-depth sibling check (`:470+`) survives the `:468` migration to `assert_agent_owner`. ✅ verified
[...]

PR #2252, opened 2026-08-17
fix(tests): a bare MagicMock cannot stub a module exporting assert_* (#2241)

### LINKED ISSUE #2241: bug: MagicMock blocks assert_* attributes — the #1310 auth-guard names break 14 api-tier tests at import
## Summary

`unittest.mock.MagicMock` deliberately raises `AttributeError` for any attribute whose name starts with `assert` — a typo guard against `mock.assert_called_once` and friends:

```
AttributeError: 'assert_owns' is not a valid assertion.
Use a spec for the mock if 'assert_owns' is meant to be an attribute.
```

The #1310 imperative auth-guard family is named exactly that way — `assert_admin`, `assert_agent_access`, `assert_agent_owner`, `assert_owns_or_admin`, `assert_owns`. So any test that stubs a module exporting one of them with a bare `MagicMock()` fails at **import time**, before a single assertion runs.

`tests/test_ip_rate_limit_fix.py` does exactly this: it installs `MagicMock()` as `sys.modules["dependencies"]`, then imports `routers/public.py`, whose line 27 is `from dependencies import get_current_user, assert_owns`. Result: **14 setup errors** in the api tier — every test in `TestIsTrustedProxy`, `TestGetClientIp`, and `TestPerTokenRateLimit`, including `test_pentest_bypass_scenario`.

The production code is fine. The lost coverage is the point: these are the trusted-proxy / client-IP tests that pin the rate-limit bypass fix.

## Context

Found by a full-suite run on `dev` (`691e7dc9`); all 14 api-tier errors share this single cause.

Latent since `b2db1c87` (#1310 / #1683), which introduced `assert_owns` into `routers/public.py` — so this has been silently red for any run that reaches the api tier.

The broader risk is that this is a **trap for future tests**, not a one-off: the guard family is a documented, encouraged pattern (Invariant #8), and every new stub of a module that re-exports one will fail the same way, with an error message that points at mock internals rather than at the stub.

## Acceptance Criteria

- [ ] The 14 errors in `tests/test_ip_rate_limit_fix.py` are gone and the tests actually execute
- [ ] The `dependencies` stub can resolve `assert_*` names — via `MagicMock(unsafe=True)`, or a `spec`'d/`create_autospec` stub, or a real import
- [ ] A short comment at the stub records *why* (`assert_*` + MagicMock's typo guard), so it is not "tidied" back
- [ ] Any other test that stubs a module exporting an `assert_*` guard is found and fixed in the same pass (grep `setitem(sys.modules` against the guard names)
- [ ] Considered: a guard test or lint that fails when a bare `MagicMock()` is installed for a module exporting `assert_*` — or an explicit note that the cost is not worth it

## Technical Notes

- Minimal fix is one line: `MagicMock(unsafe=True)` (the `unsafe` flag exists precisely to disable the `assert`/`assret` prefix check).
- `create_autospec(real_module)` is stricter and would catch stub drift generally, but pulls in the real import the stub exists to avoid — probably not worth it here.
- Reproduce with: `cd tests && source .venv/bin/activate && python -m pytest test_ip_rate_limit_fix.py -q`

dolho:
## The bug

`MagicMock` raises `AttributeError` for any attribute whose name starts with `assert` — its typo guard against `mock.assert_called_once`:

```
AttributeError: 'assert_owns' is not a valid assertion.
```

The #1310 imperative auth-guard family is named exactly that: `assert_admin`, `assert_agent_access`, `assert_agent_owner`, `assert_owns_or_admin`, `assert_owns`. `tests/test_ip_rate_limit_fix.py` installs `MagicMock()` as `sys.modules["dependencies"]` and then imports `routers/public.py`, whose line 27 is `from dependencies import get_current_user, assert_owns` — so the import raised and **all 14 tests became setup errors before one of them ran**. Those are the trusted-proxy / client-IP tests pinning the rate-limit bypass fix, `test_pentest_bypass_scenario` among them, silently red since `b2db1c87`.

**The fix is one line** — `MagicMock(unsafe=True)`, the flag that exists to disable the prefix check — with a comment recording why it is load-bearing rather than tidy, since the entire failure mode is that the next reader deletes it.

## Sweep (AC#4)

Derived the exporter set rather than guessing: `dependencies` (the five #1310 guards) and `services/mcp_auth_service.py` (`assert_email_may_reach_agent`). Exactly **one** bare-MagicMock stub of either exists — the one fixed here.

`tests/unit/test_inter_agent_timeout_unit.py` also stubs `dependencies`, but with a real `types.ModuleType` carrying explicit attributes, so it is immune — and would fail with an honest *"module has no attribute"* if it ever needed a guard.

## The guard (AC#5) — built, with the reasoning

The issue asked me to consider one or say why not. **Built**, because this is a trap by construction rather than an incident:

- Invariant #8 documents and *encourages* the guard family, and #1683 added a fifth member
- every future test stubbing a module that re-exports one breaks identically, with a message about mock internals rather than about the stub
- it breaks in a tier **per-PR CI never reaches**, so rediscovery costs a full-suite run — that is how this stayed hidden for weeks

Against that, the guard is a text scan with no imports and no runtime. `tests/unit/test_2241_magicmock_assert_guard.py`:

- **derives** the exporter set from `src/backend` — never hardcoded, so a sixth guard is covered the day it lands
- flags `sys.modules` stubs of those modules built with a bare MagicMock, accepting `unsafe=True` / `spec=` / `create_autospec` / a real `types.ModuleType` (each resolves `assert_*` on its own terms)
- **guards its own vacuity**: one test asserts the exporter scan still finds the family, because a scan that silently matches nothing passes forever
- **pins the upstream behaviour it rests on**, so if `mock` ever drops the prefix check this file fails and can be *deleted* rather than maintained on faith

Stated limit: source text only. A stub built across several lines or through a helper is out of reach; it catches the shape that actually shipped.

## Verification

| check | result |
|---|---|
| `test_ip_rate_limit_fix.py` | **14 passed** (was 14 setup errors) |
| the new guard | 3 passed |
| **guard proven to detect** | reverting the one-line fix makes it fail, naming `tests/test_ip_rate_limit_fix.py:88 stubs 'dependencies'` |
| `lint_sys_modules.py` / `lint_root_test_placement.py` | clean |

One note on the transcript, in case it matters to a reviewer: immediately after the fix the file still errored for me — that was my own missing `TRINITY_TEST_PASSWORD` (the session `api_client` fixture 401'd against the local instance), not the fix. With real credentials, 14/14.

Fixes #2241


vybe (approved):
Validated via /validate-pr.

Correct root cause and a genuinely one-line fix. `MagicMock` raises `AttributeError` for any attribute starting with `assert` (its typo guard), and the #1310 imperative auth guards are named `assert_admin` / `assert_agent_access` / `assert_agent_owner` / `assert_owns` / `assert_owns_or_admin` — so stubbing `dependencies` with a bare `MagicMock()` made `routers/public.py`'s import raise and turned all 14 trusted-proxy / client-IP tests into setup errors, silently red since b2db1c87. `MagicMock(unsafe=True)` is exactly the documented escape.

The accompanying guard is the right shape: static text scan with no imports, exporter set **derived** from `src/backend` rather than hardcoded (so a sixth guard is covered the day it lands), it guards its own vacuity, and it pins the upstream mock behaviour it depends on so the file fails loudly rather than being maintained on faith.

- Security scan: clean
- The earlier `pytest (head, seed 67890)` red was a CANCELLED job — a 25-minute stall in `Run unit suite`, the known hung-job flake, not an assertion failure. Re-ran it; all six seed/side jobs now pass and `regression diff` is green.
- Closing keyword patched in so #2241 auto-promotes

PR #2254, opened 2026-08-17
fix(workspace): make the `/` typeahead able to find every skill the agent offers (#2213)

### LINKED ISSUE #2213: bug: Workspace composer '/' typeahead lists only part of an agent's skills
## Summary

Typing `/` in the Workspace composer lists only part of an agent's skills. Skills the agent demonstrably has are missing from the popup, while others appear.

## Context

Reported from live use of `/workspace`. The `/` typeahead is fed from the roster payload's `playbooks[]` (the new-chat briefing), so the list passes through four independent narrowing steps — the fix starts by determining which one is dropping the missing entries.

**1. Server-side hint cap (24).** `src/backend/client_portal/service.py` bounds the briefing at `_MAX_BRIEFING_HINTS = 24` via `_bound_briefing_hints()`. An agent with more client-visible skills silently loses the tail, and because the popup searches only what shipped, those skills are unreachable by typing their name.

**2. Exposure filter.** `services/connector_service.py::resolve_exposed_playbooks` drops any skill whose frontmatter sets `user-invocable: false`, and — when a connector allow-list is configured for that agent — everything not on the list. A skill available *to the agent* is not automatically client-visible, so part of this may be correct behaviour that is simply never explained.

**3. Source directory scan.** The agent-server endpoint that feeds this (`GET /api/skills`, `docker/base-image/agent_server/routers/skills.py`) scans exactly two directories — `.claude/skills/` and `~/.claude/skills/` — one level deep, requiring a `SKILL.md` in each immediate subdirectory. Skills provided by an installed **plugin**, or nested any deeper, are invisible to that endpoint and therefore to the Workspace, even though the agent can run them.

**4. Popup render bound.** `TYPEAHEAD_LIMIT = 8` in `src/frontend/src/components/portal/portalUtils.js` renders 8 rows plus an overflow count. If that count is not obvious, a bounded list reads as a broken one.

## Acceptance Criteria

- [ ] Determined which tier drops the missing skills for a real agent, by comparing three observations: `GET /api/skills` from inside the container, the `playbooks[]` in the roster payload, and what the `/` popup renders
- [ ] Skills the agent can run but the Workspace does not offer are either included, or the omission is explained where the user can see it (honest status — e.g. an overflow/`N not exposed` affordance rather than a silently short list)
- [ ] Searching with `/` matches across the agent's full client-visible set, not only the first 24 that happened to ship in the briefing
- [ ] Plugin-provided skills are either enumerated by the agent-server skills scan or explicitly declared out of scope and documented
- [ ] Regression test over `filterPlaybookCandidates` / `boundCandidates` in the `portalUtils` unit tests covering a set larger than both bounds

## Technical Notes

- Frontend: `src/frontend/src/components/portal/portalUtils.js` (`filterPlaybookCandidates`, `boundCandidates`, `TYPEAHEAD_LIMIT`), `PortalTypeahead.vue`, `PortalConversation.vue`
- Backend: `src/backend/client_portal/service.py` (`_agent_briefing`, `_bound_briefing_hints`, `_use_case_hints`), `src/backend/services/connector_service.py` (`resolve_exposed_playbooks`)
- Agent container: `docker/base-image/agent_server/routers/skills.py` (`scan_skills_directory`, `list_skills`)
- The same `playbooks[]` payload also drives the new-chat briefing hint cards, so a cap change affects both surfaces — the hint grid is deliberately bounded and collapses past 6; the `/` search bound is the part that should not truncate silently

dolho:
## Diagnosis first (AC#1), measured on a live instance

I planted 30 probe skills in a real agent container so all four tiers could be compared against one agent:

| observation | result |
|---|---|
| agent-server `GET /api/skills` (in container) | **33** skills, all `user_invocable` |
| roster payload `playbooks[]` | **24** — exactly `_MAX_BRIEFING_HINTS` |
| popup, query `"probe 27"` | **0 matches**, nothing on screen saying why |

So of the four candidates: the container scan found everything (tier 3), the exposure filter dropped nothing (tier 2 — no allow-list, everything invocable), and the popup's own limit of 8 was **already honest**, rendering *"N more — keep typing to filter"* (tier 4).

**Tier 1, the payload bound, is the defect.** 24 is right for the hint-card grid it was written for (#2101 — a card surface has a layout limit, and cards carry descriptions) and wrong as the search corpus, because search cannot reach what never shipped and the popup has no way to know its list is truncated.

## The fix: two surfaces, two bounds, one honest count

- **`playbooks`** — unchanged. 24, rich, feeds the cards.
- **`searchable_playbooks`** — the same client-visible set at **200**, title + starter only. Dropping descriptions is what makes 200 entries cheap: ~40 chars each instead of ~540.
- **`playbooks_total`** — the count *before* either bound.

The popup now keeps the two omissions distinct instead of blurring them:

- `overflow` — rows it chose not to draw; **typing reaches them**
- `hiddenCount` — skills that never reached the browser; **typing cannot** — so it says *"N more not listed here — ask for it by name"*, and the screen-reader status says the same

A bounded list is fine. A bounded list that looks complete is the bug.

## Two things worth a reviewer's attention

**Widening the briefing's return broke five test files, two of which FAKE `_agent_briefing` with a 2-tuple** — the #2242 stale-double class, and it would have surfaced as a 500 in the roster build rather than a failed assertion. So the return is a named `AgentBriefing` NamedTuple (still a tuple, so the existing `isinstance` guards hold) and one `_apply_briefing` helper copies it onto the card positionally-tolerantly, absorbing any stub still returning the old shape. That took the fallout from **26 failures to 5**, and those 5 were asserting the old shape directly — updated to read fields *by name*, so a future field cannot silently shift what they assert.

**My own test caught a real defect mid-change.** The availability early-return still yielded a 2-tuple, which would have raised `ValueError` on unpack for **every stopped or unavailable agent** — breaking the roster on exactly the agents that early return exists to serve cheaply.

## Plugin skills (tier 3): out of scope, documented

Taking the AC's second option, with the reasoning recorded at `scan_skills_directory`. Enumerating plugin-provided skills means reading the plugin cache layout in `agent_server/routers/skills.py`, which **ships in the base image** — so an instance only gets it after a rebuild and old images silently keep the old surface, the release-note trap ent#123 already paid for once. The Workspace consequence is now bounded and honest rather than invisible, which is what makes deferring it defensible.

## Verification

| check | result |
|---|---|
| backend: every briefing / roster / connector suite | **180 passed** |
| frontend unit | **851 passed** (10 new) |
| the new specs | exercise a set larger than **both** bounds — the case the AC names, and the case the old tests never had |
| design-token check · prod build | clean |

**Not verified live post-fix, deliberately:** the running instance bind-mounts a *different* worktree (#2157's), and I am not editing a colleague's checkout or repointing their stack to prove my own branch. The pre-fix measurement above is real; the post-fix confirmation belongs to the deploy.

Fixes #2213


dolho:
## Review round applied (7 findings)

Pushed `538f309f`. Five were real defects introduced by the first round:

- **Field caps bypassed.** `searchable_playbooks` was built from `playbooks` *before* `_bound_briefing_hints` applied its title/starter caps, so up to 200 uncapped strings shipped per card per roster load — and skill `name` comes from agent-author-controlled YAML, which also feeds what a pick inserts into the composer. Now clamped with the same constants.
- **Every `/` row lost its subtitle.** Searchable entries carry no description by design, and the popup renders `secondary: c.description || ''` — so pointing search at them blanked the subtitle for *all* agents, including those under the card bound whose descriptions did ship. `playbookSearchSource` merges them back by title: restored where one exists, blank only for the tail that genuinely has none, payload not mutated.
- **`playbooks_total` under-reported on the use-case tier**, because `_use_case_hints` truncates to 6 before the count was taken: an agent declaring 10 use-cases reported 0 hidden while 4 were dropped — the same "looks complete" bug this issue exists to fix, one tier over.
- **Shared mutable defaults.** `AgentBriefing`'s `playbooks: list = []` is class-level, so every stopped/failed card aliased one list — in a module whose `_bound_briefing_hints` mutates hints in place. Immutable `()` now.
- **The hidden count rendered mid-query**, where "N more not listed here" reads as "N more results matching what you typed" — false, since typing can never reach them. Bare `/` only.

Plus the orphaned #2101 comment moved back to its constant, and **my own test tautology**: `_briefing(n) if n else (...)` short-circuited `n == 0`, so the empty-skills path where the fallback tier takes over was never exercised.

New regression tests: caps hold under 4000-char names, the use-case tier's hidden count is visible, defaults are immutable, and the subtitle merge is correct in all four directions. 127 tests across this file plus the ent#79 / #2196 / ent#357 briefing suites.

vybe (approved):
Validated via /validate-pr.

The defect is correctly located at the payload bound, not the popup: `_MAX_BRIEFING_HINTS` (24) is right for the hint-card GRID it was written for (#2101 — a card surface has a layout limit and cards carry descriptions) and wrong as the search corpus, because search cannot reach what never shipped. Splitting into `playbooks` (cards) / `searchable_playbooks` (200, title+starter only, which is what keeps it cheap) / `playbooks_total` (so remaining truncation is sayable rather than silent) is the right shape, and the exposure policy still bounds both surfaces.

Two details worth calling out as good:
- `AgentBriefing` uses **immutable** tuple defaults — a NamedTuple's defaults are class-level, so `[]` would be one shared list aliased by every stopped agent's card, in a module that mutates hint lists in place.
- `_apply_briefing` is positional-tolerant on purpose, absorbing test doubles still returning the pre-#2213 2-tuple. That is the #2242 class, and without it a stale double becomes a 500 in the roster build rather than a failed assertion — it took the fallout from 26 failures to 5.

- Security scan: clean
- `ui` label present, so `build` and `e2e` both ran
- Merged `origin/dev` and resolved three conflicts by hand: `portalUtils.js` and `PortalConversation.vue` against #2256 (both append-only; kept both functions and both imports) and `tests/registry.json` against #2251/#2252/#2255 (rebuilt from git stages, never spliced — verified no entry modified or deleted on either side). Post-resolve: full frontend suite 875/875, portal backend tests 128/128.
- Verified the resolved branch diffs against dev in exactly its own 11 files — no leakage from the PRs it merged over.
- Closing keyword patched in so #2213 auto-promotes

PR #2255, opened 2026-08-17
fix(deploy): give /data/archives an owner, so log archival stops dying silently (#2205)

### LINKED ISSUE #2205: bug: log archival is dead — /data/archives is root-owned while the backend runs as UID 1000 (11 GB unpruned)
## Summary

`/data/archives` is owned by `root:root` while the backend runs as `uid=1000(trinity)`, so **`log_archive_service` fails on every run** with `[Errno 13] Permission denied`. Archived originals are never pruned, and `/data/logs` grows without bound.

Observed on a local instance: **`/data/logs` = 11 GB**, largest single file `platform-2026-07-06.json` at **4.6 GB**. Failing since at least 2026-06-23.

## Evidence

```
$ docker exec trinity-backend id
uid=1000(trinity) gid=1000(trinity) groups=1000(trinity),0(root)

$ docker exec trinity-backend ls -ld /data /data/logs /data/archives
drwxr-xr-x 9 trinity trinity  /data
drwxrwxr-x 2 trinity trinity  /data/logs
drwxr-xr-x 2 root    root     /data/archives      <-- root-owned

$ docker exec trinity-backend touch /data/archives/.perm-probe
touch: cannot touch '/data/archives/.perm-probe': Permission denied
```

Live errors, recurring every archival run (34 occurrences in one day's log):

```
ERROR services.archive_storage: Failed to store archive agents-2026-06-23.json.gz:
  [Errno 13] Permission denied: '/data/archives/agents-2026-06-23.json.gz'
ERROR services.log_archive_service: Storage failed for agents-2026-06-23.json.gz: ...
INFO  services.log_archive_service: Archiving platform-2026-07-06.json (4647.4 MB)
```

## Root cause — a missed sibling of #1478

#1478 established that the non-root backend needs write access on the shared logs volume, and added a one-shot init container for exactly this (`docker-compose.yml:613-627`):

```yaml
  # #1478: one-shot that makes the shared logs volume dir owned by UID 1000 so the
  # backend (non-root) can unlink retention-archived originals.
  logs-init:
    command: ["sh", "-c", "chown 1000:1000 /data/logs && chmod 775 /data/logs"]
```

**There is no equivalent `archives-init`** for `trinity-archives:/data/archives` (`docker-compose.yml:349`). Named Docker volumes are root-owned by default, and `scripts/deploy/start.sh:183-191` chowns only the bind-mounted `$data_path` — and only on Linux.

`archive_storage.py:84` does `self.archive_path.mkdir(parents=True, exist_ok=True)`, which silently no-ops on the existing root-owned directory, so the failure surfaces only at write time.

`docs/migrations/NON_ROOT_CONTAINERS_2026-05.md` makes no mention of archives.

## Why it stayed invisible

The ERROR is written into the very log files that are not being pruned. The symptom (disk growth) and the diagnosis (permission denied) are in the same unbounded file nobody tails. Same silent-failure class as #1478 and #1871.

## Acceptance Criteria

- [ ] The backend (UID 1000) can write `/data/archives` on a fresh install and on an upgrade
- [ ] `log_archive_service` completes successfully; archived originals are removed from `/data/logs`
- [ ] A pre-existing root-owned `/data/archives` is repaired, not just fixed for new installs
- [ ] The failure is surfaced somewhere an operator sees it (not only in the unpruned log)
- [ ] Both compose files covered; `docs/migrations/NON_ROOT_CONTAINERS_2026-05.md` updated to name the archives volume
- [ ] Consider a guard so a future volume added without an ownership story fails loudly (the #1871 parity-test pattern)

## Technical Notes

- `docker-compose.yml:349` (`trinity-archives`), `:613-627` (`logs-init` precedent)
- `src/backend/services/archive_storage.py:76,84`; `src/backend/services/log_archive_service.py`
- Related: #1478 (logs volume, same class), #1871 (bounded container logs), Invariant #17 / #874 (non-root containers)
- Found during an audit of the 2026-08-14 exploratory E2E sweep; the sweep noted the 11 GB as a minor observation (L-6) without root-causing it, and it was never filed.

dolho:
P0. `/data/archives` was root-owned while the backend runs as UID 1000, so `log_archive_service` failed on **every** run and archived originals were never unlinked. Reproduced on a live instance before touching anything:

```
$ docker exec trinity-backend ls -ld /data /data/archives
drwxr-xr-x  trinity trinity  /data
drwxr-xr-x  root    root     /data/archives
$ docker exec trinity-backend touch /data/archives/.perm-probe
touch: cannot touch ...: Permission denied
/data/logs = 8.5 GB · 114 archives frozen since 2026-05-18
```

## The mechanism is subtler than "named volumes are root-owned"

`docker/backend/Dockerfile` runs `mkdir -p /data /agent-configs && chown -R trinity:trinity …` **before** `USER trinity`, and Docker seeds an *empty* named volume from the image's directory **including ownership** — which is why `trinity-data`, mounted **at** `/data`, is fine. `trinity-logs` and `trinity-archives` mount at **nested** paths the image never creates, so there is nothing to inherit from. #1478 hit this on the logs volume and added a one-shot; archives never got one.

`archive_storage.__init__` then hid it: `mkdir(parents=True, exist_ok=True)` **succeeds** on someone else's directory, so the class logged "initialized" and only the first *write* failed — with a message about a file, not about ownership. And that error went into the very log files archival exists to prune, so symptom (disk growth) and diagnosis (permission denied) sat in the same unbounded file nobody tails.

## Fixes

- **`archives-init`** one-shot in the dev compose, mirroring `logs-init`, with the backend gated on `service_completed_successfully` so a chown cannot race the first archival run. It runs on every `up`, so it **repairs existing installs** — verified by running its exact command against the live volume: `root:root` → `1000:1000`, and the backend's `touch` then succeeded.
- **Prod needs no one-shot, and that is now stated at the mount** instead of left to be rediscovered: prod has no archives volume, `/data/archives` lives inside the `${TRINITY_DATA_PATH}` bind mount that `start.sh` chowns recursively. The guard fails if prod ever grows one.
- **Operator-visible failure** (AC#4): the storage constructor probes writability and files **one** operator-queue alarm naming the cause *and the remedy*, instead of leaving the truth in an unpruned log. Plumbing mirrors #2216 exactly — direct platform create (bypassing #1632 ingestion caps by construction), reserved `log-archive-` id prefix so an agent cannot pre-create and thereby silence it, `_log-archive` sentinel registered canary-L-03-exempt and uncreatable as an agent name. Fail-soft throughout: a maintenance job's broken directory must never take the backend down.
- **The guard** (AC#6, #1871 pattern): every named volume mounted into a UID-1000 service must be covered by an init one-shot, by the image having prepared that exact path, or by an explicit exemption **with a reason**. Proven to detect by excising the one-shot and re-running the rule.
- **`docs/migrations/NON_ROOT_CONTAINERS_2026-05.md`** now names both nested volumes and explains why nesting defeats the image's chown.

Two of the guard's own bugs surfaced while writing it and are worth noting: a block parser that reported *covered* volumes as uncovered, and a substring check that tripped over the comment documenting the rule it enforces.

## Verification

| check | result |
|---|---|
| new guards + alarm parity suites (#2216, #1644, #1632) | **102 passed** |
| both compose files | render valid; `archives-init` + both `depends_on` gates present |
| live repair | `root:root` → `1000:1000`; backend `touch` OK |
| guard proven to detect | excise the one-shot → rule rejects |

⚠️ **I changed your live dev instance**: applying the one-shot command re-owned that volume's directory (ownership only — no data touched). Its 8.5 GB of logs will begin pruning on the next archival run.

Fixes #2205


dolho:
## Review round applied (7 findings)

Pushed `e4774be2`. The sharpest finding said this fix would not have pruned the file the issue was filed about — **substantially right**, and I verified the numbers rather than taking them:

**Staging sat on a 100 MB tmpfs.** Both compose files mount `- /tmp:noexec,nosuid,size=100m`, and archival gzips into `/tmp/archives`. I measured the real compression ratio: platform JSON logs are **~42:1** (50 MB sample → 1.2 MB), so the review's 300–500 MB estimate was ~4× pessimistic — the 2.0 GB file on my instance stages to ~47 MB and fits. But the **4.6 GB file named in this issue** stages to ~110 MB and does **not**: ENOSPC, swallowed by `logger.error` + `continue`, and the largest file never archives. Staging now resolves beside the destination, which removes the ceiling and turns the final store into a same-filesystem rename.

Also fixed:
- **`mkdir` could crash-loop the backend** — it runs at import time (module-level singleton imported by `main.py`) and was unguarded, so an unwritable parent raised straight out of the import: no API, no alarm, the opposite of this PR's stated guarantee.
- **False alarm when `LOG_ARCHIVE_ENABLED=false`** — storage is built unconditionally while the flag is only read in `start()`, so a deliberately-disabled install got a high-priority alarm every boot about a job that never runs.
- **Never re-armed** — the alarm fired once per *process*; under `restart: unless-stopped` an acknowledged-but-unfixed fault went silent for weeks. Now probed per run, with the per-day id bucket doing the deduplication (#2216's `_staleness_check` precedent).
- **My guard test was a tautology** — it split on `"\\n  vector:"`, which is defined near the end of the file, so its "backend" span covered six services and was satisfied by the one-shots' own definitions. Both gates could have been deleted with it still green.
- **A false exemption** — `_EXEMPT["agent-configs"]` claimed the backend doesn't write it; it does. Safe only because the image prepares the path, and since `_EXEMPT` is checked first, the untrue reason was the one carrying it. Dropped.
- **A guard loophole** — `_image_prepared_paths()` harvested every `mkdir -p` while only checking that a chown appeared *somewhere*, so a future unchowned `mkdir -p /data/archives` would read as "prepared".

Five new tests; 18 across the two #2205 files. One of my own new assertions was also wrong (`"/tmp" not in str(staging)` is meaningless when `tmp_path` is under /tmp) and now asserts what matters.

vybe (approved):
Validated via /validate-pr. P0.

Root cause is right and well-evidenced: Docker creates named volumes root-owned, and `trinity-archives` mounts at a NESTED path (`/data/archives`) the image never creates — so unlike `trinity-data` there is no image ownership to inherit. `mkdir(exist_ok=True)` then silently no-ops on the root-owned dir and nothing fails until the first write. The `archives-init` one-shot mirrors `logs-init` (#1478), is idempotent, and repairs existing installs rather than only fresh ones.

Also good: the constructor was kept cheap and DB-free after the CI finding (a module-level singleton must not put database init on the import path), with the alarm moved to `probe_archive_writability()` at the start of an archival run — which is also where re-arming belongs — and `LOG_ARCHIVE_ENABLED=false` correctly suppresses an alarm about a job that never runs.

**One fix applied during validation** (commit `8768e1b1`): the new `_alarm_unwritable_archive_dir` create site was unclassified, so the #1677 AST emitter-parity guard failed under HEAD on all three seeds while passing under base — the guard doing exactly its job. I classified it **platform-only** rather than routing it through `create_bounded_alert`, having verified: it is raised from a scheduled archival run (daily cadence), its id is bucketed per path per UTC day and `create_operator_queue_item` is ON CONFLICT DO NOTHING (so a restart loop yields ≤1 row/path/day), and no agent-controlled input reaches it — `path` is the configured dir and `detail` an OSError string. Same shape as the #2216 sibling this plumbing mirrors. `regression diff` is green on the re-run.

- Security scan: clean
- All checks green incl. `prod-image-smoke`, `verify-non-root`, both guards
- Closing keyword patched in so #2205 auto-promotes

PR #2256, opened 2026-08-17
fix(workspace): looser chat rhythm, a wider column, no scrollbar in an empty composer (#2211)

### LINKED ISSUE #2211: bug: Workspace chat readability — tight line spacing, narrow message column, scrollbar in the empty composer
## Summary

Three readability defects in the Workspace chat surface (`/workspace`), all in `src/frontend/src/components/portal/PortalConversation.vue` (and its twin `PortalRoom.vue`): message lines sit too tightly together, the message column uses only a narrow slice of the available panel width, and the composer renders a vertical scrollbar while it is empty.

## Context

Reported from live use of the Workspace. The text reads "stuffy" — dense line spacing plus a narrow column — and the empty composer shows a scrollbar it has nothing to scroll.

**1. Line spacing is too tight.** The thread is `space-y-4` with bubbles at `px-3.5 py-2 text-sm`, and the assistant's rendered markdown is squeezed further by the scoped rule `.prose-portal :deep(p) { margin: 0.25rem 0; }` — 4px between paragraphs, at the default `text-sm` leading. Needs a modest increase in line-height and inter-message/inter-paragraph rhythm (a tiny bit larger, not a redesign).

**2. Message column is too narrow.** Both the thread and the composer are capped at `max-w-3xl mx-auto` (48rem), and bubbles cap again at `max-w-[85%]` of that. On a wide viewport most of the panel is empty gutter. The column should take a higher percentage of the available panel width while staying readable.

**3. Empty composer shows a vertical scrollbar.** `autoGrow()` does:

```js
el.style.height = 'auto'
el.style.height = Math.min(el.scrollHeight, 160) + 'px'
```

The textarea is `box-sizing: border-box` with a 1px border on each side (Tailwind `border`), and `scrollHeight` **excludes** borders — so the applied height is 2px short of the single line it must hold, and the browser shows a scrollbar even at zero characters. `overflow-y` is also never toggled to `hidden` below the `max-h-40` ceiling. Reproduces after any input event that leaves the field empty (send, select-all-delete). The identical composer + `autoGrow()` exists in `PortalRoom.vue`, so any fix must land in both.

## Acceptance Criteria

- [ ] Chat message line spacing is visibly looser (line-height and inter-message/paragraph rhythm), without changing the bubble language or font size
- [ ] The message/composer column takes a larger share of the available panel width on wide viewports, with a sensible readable ceiling; narrow/mobile layout unchanged
- [ ] An empty composer never shows a vertical scrollbar; the scrollbar appears only once content exceeds the `max-h-40` ceiling
- [ ] The composer fix is applied to both `PortalConversation.vue` and `PortalRoom.vue` (same code, two files)
- [ ] Changes use semantic tokens only per `docs/memory/design-system-contract.md` — no raw Tailwind palette classes, no hardcoded hex; verified in light **and** dark
- [ ] No layout shift or scroll-position reset introduced in the thread on send/receive

## Technical Notes

- Files: `src/frontend/src/components/portal/PortalConversation.vue` (thread `max-w-3xl` + `space-y-4`, composer textarea, `autoGrow()`, scoped `.prose-portal` block), `src/frontend/src/components/portal/PortalRoom.vue` (same composer and `autoGrow()`)
- Scrollbar root cause: under `box-sizing: border-box`, `scrollHeight` omits border width — the assigned height needs the border box accounted for (or the textarea needs `overflow-y` managed explicitly below the ceiling)
- Read `docs/memory/design-system-contract.md` before touching any file under `src/frontend/`
- Raw-color ratchet: `node src/frontend/scripts/scan-raw-colors.mjs src/frontend` must not regress per-file counts

dolho:
Three readability defects, all measured against `docs/memory/design-system-contract.md` (4px grid: 4 tight · 8 related · 12 grouped · 16 card · 24 section; semantic tokens only).

## 1. Rhythm

Bubbles were `py-2` at default `text-sm` leading, with rendered markdown squeezed to `margin: 0.25rem` — **4px** between paragraphs. Now:

- `py-2` → **`py-3`** (8 → 12px, the next grid step) and `leading-relaxed` on the message text
- paragraph margin **4 → 8px**, with first/last paragraph margins zeroed so the looser rhythm does not double up with the bubble's own padding and read as a lopsided bubble
- `space-y-4` → **`space-y-6`**: 20px is not on the contract's scale, so 24px is the next legal step

Bubble language and font size untouched, as the AC requires.

## 2. Width

`max-w-3xl` → **`max-w-4xl`** (48 → 56rem) on both the thread and the composer, in both files.

Deliberately not wider: at `text-sm`, a full-width line in a 4xl column is ~95 characters, which is already at the readable ceiling. 5xl would be ~110ch and would trade the reported "stuffy" complaint for a harder-to-track one. Narrow/mobile is unchanged — the cap only binds above it.

## 3. Empty-composer scrollbar

`scrollHeight` **excludes** the border under `box-sizing: border-box`, and the textarea has a 1px border per side — so `height = min(scrollHeight, 160)` left the field **2px short** of the single line it must hold, and the browser drew a scrollbar at zero characters. `overflow-y` was never managed either, so the scrollbar stayed available at every height below the ceiling.

Fixed by adding the **measured** border box (`offsetHeight - clientHeight`, so a future `border-2` cannot reintroduce it) and toggling `overflow-y` explicitly.

## The issue's premise was wrong on one point

**`PortalRoom.vue` does not carry a copy of `autoGrow()` — it has none at all.** Its composer never grew past its single `rows="1"` line, so a long message scrolled inside a one-line box. It therefore gains the **corrected** implementation rather than a patched copy, plus a shrink-back on send.

And rather than land the same arithmetic in two files — the duplication that lets one copy silently keep the bug — it is extracted as the pure **`resolveComposerGrowth()`** in `portalUtils.js`, which is also what makes it testable without a DOM.

## Verification

| check | result |
|---|---|
| frontend unit | **849 passed** (8 new) |
| new spec pins | one-line case (40px content + 2px border = 42, where the old code assigned 40), exactly-full (160 is *not* overflowing), border measurement at 0/2/4px, degenerate input |
| `check:tokens` | clean |
| raw-color ratchet | **byte-identical to dev** (809/8474/468/3936) — these changes add no color classes |
| prod build | clean |

**Not verified: the visual result in a browser, in either theme.** The running instance bind-mounts a different worktree, so seeing this needs a deploy or a local dev server on this branch. A spacing change is exactly what a test cannot judge, so the light/dark pass belongs to whoever runs it — it's in the checklist on the issue.

Fixes #2211

### DIFF
--- a/src/frontend/src/components/portal/PortalConversation.vue
+++ b/src/frontend/src/components/portal/PortalConversation.vue
@@ -73,7 +73,7 @@
 
     <!-- Messages -->
     <div ref="scrollEl" class="flex-1 min-h-0 overflow-y-auto px-3 sm:px-6 py-5">
-      <div class="max-w-3xl mx-auto space-y-4">
+      <div class="max-w-4xl mx-auto space-y-6">
         <div v-if="loadingHistory" class="text-center text-sm text-gray-400 mt-10">Loading…</div>
 
         <!-- Briefing (new-chat state) rendered by the parent via slot -->
@@ -83,7 +83,7 @@
           <PortalAvatar v-if="m.role !== 'user'" :name="agent.name" :avatar-url="agent.avatar_url" :size="28" class="mt-0.5" />
           <div v-if="m.role === 'user'" class="max-w-[85%] flex flex-col items-end gap-1">
             <div
-              class="rounded-2xl rounded-br-md px-3.5 py-2 text-sm whitespace-pre-wrap"
+              class="rounded-2xl rounded-br-md px-3.5 py-3 text-sm leading-relaxed whitespace-pre-wrap"
               :class="m.failed ? 'bg-status-danger-50 dark:bg-status-danger-900/30 text-status-danger-800 dark:text-status-danger-200 ring-1 ring-status-danger-300 dark:ring-status-danger-800' : 'bg-action-primary-600 text-white'"
             >{{ m.content }}</div>
             <p
@@ -101,7 +101,7 @@
           </div>
           <div
             v-else
-            class="max-w-[85%] rounded-2xl rounded-bl-md bg-gray-100 dark:bg-gray-800 text-gray-900 dark:text-gray-100 px-3.5 py-2 text-sm prose-portal"
+            class="max-w-[85%] rounded-2xl rounded-bl-md bg-gray-100 dark:bg-gray-800 text-gray-900 dark:text-gray-100 px-3.5 py-3 text-sm leading-relaxed prose-portal"
             v-html="render(m.content)"
           ></div>
         </div>
@@ -130,7 +130,7 @@
 
     <!-- Composer -->
     <div class="shrink-0 border-t border-gray-200 dark:border-gray-800 bg-white dark:bg-gray-900 px-3 sm:px-6 py-3">
-      <div class="max-w-3xl mx-auto">
+      <div class="max-w-4xl mx-auto">
         <p v-if="offline" class="mb-2 text-xs text-status-warning-600 dark:text-status-warning-400 flex items-center gap-1.5">
           <span class="w-1.5 h-1.5 rounded-full bg-status-warning-500"></span>
           You appear to be offline — messages will send once you're reconnected.
@@ -251,6 +251,7 @@ import {
   dismissAfterInsert,
   filterAgentCandidates,
   filterPlaybookCandidates,
+  resolveComposerGrowth,
   isSuppressed,
   nextActiveIndex,
   nextDismissState,
@@ -424,11 +425,19 @@ async function scrollDown() {
   await nextTick()
   if (scrollEl.value) scrollEl.value.scrollTop = scrollEl.value.scrollHeight
 }
+// #2211: the composer's growth ceiling, matching the `max-h-40` class on the
+// textarea (40 * 4px). Named so the class and the JS cannot drift apart.
+const COMPOSER_MAX_PX = 160
+
 function autoGrow() {
   const el = textarea.value
   if (!el) return
   el.style.height = 'auto'
-  el.style.height = Math.min(el.scrollHeight, 160) + 'px'
+  // The three measurements the pure resolver needs; `height: auto` first so
+  // `scrollHeight` reports the content's natural height rather than the current one.
+  const { height, overflowY } = resolveComposerGrowth(el, COMPOSER_MAX_PX)
+  el.style.height = height + 'px'
+  el.style.overflowY = overflowY
 }
 
 // ---- ent#392: composer typeahead (`/` playbooks, `@` agents) -----------------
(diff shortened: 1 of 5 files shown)


dolho:
## Review round applied (6 findings)

Pushed `09d19852`. Round one extracted `resolveComposerGrowth()` and tested the arithmetic thoroughly — and **all six findings still slipped through, because all six were in the wiring**, which my verification table implied was covered and was not. That is the fair criticism of the first round.

**One root cause behind three of them:** `autoGrow()` reads the DOM, but Vue patches `v-model` on the next microtask — so every call following a programmatic `input.value = …` measured the *old* content. Before this PR that only meant "did not resize"; now that `overflow-y` is pinned it also means text left **clipped and unscrollable**. Rather than ask each call site to remember `nextTick`, there is now `autoGrowAfterUpdate()`, used by room send, conversation send, the failure-restore path, dictation, and mount-with-prefill. The direct `@input` path stays synchronous, where the DOM is already current.

- **paste/drop** — growth now runs *before* the early-return in PortalRoom, as it already did next door.
- **typeahead pick** — PortalRoom's `acceptActive` never regrew, so a mention wrapping to a second line stayed one line tall.
- **rewrap** — a pinned `hidden` never recomputed on width change, so narrowing the window could leave a fitting draft invisible where it previously scrolled. Both files recompute on `resize` and unregister on unmount.
- **`prose-portal` was inert in PortalRoom** — applied in the template, defined only in PortalConversation's scoped block. So the paragraph-rhythm half never reached room transcripts and `pre { overflow-x: auto }` was missing, letting wide code blocks overflow the bubble.

Copying that block would have grown the hardcoded-colors ratchet 468 → 469 (its `rgba(0,0,0,0.06)`), which the contract forbids. Both copies now use `@apply bg-gray-100 dark:bg-gray-800` — the same tokens the bubbles use — so **hardcoded goes 468 → 467**, at the cost of `raw_gray` +4. Gray is the contract's sanctioned default, and I'd rather state that trade than hide it.

**New `portalComposerWiring.spec.js`** covers what a pure-function test structurally cannot: no programmatic `input.value =` followed by a synchronous grow, the deferred variant exists, growth precedes the paste return, resize registered *and* unregistered, a pick regrows, and any file applying `prose-portal` defines it. 861 frontend tests, tokens clean, build clean.

vybe (approved):
Validated via /validate-pr.

Frontend-only. Correctly diagnoses the empty-composer scrollbar as `scrollHeight` excluding the border under `box-sizing: border-box`, and fixes it with a pure `resolveComposerGrowth()` that measures the border (`offsetHeight - clientHeight`) rather than hardcoding 2px — so a future `border-2` cannot reintroduce it. Extracting it to `portalUtils` is the right call given PortalRoom had no autoGrow at all; two copies of that arithmetic is how one composer silently keeps the bug.

- Security scan: clean
- `ui` label present, so `build` and `e2e` both actually ran and passed
- Non-happy-path coverage present (portalComposerGrowth/Wiring specs)
- Noted from the PR body: the light/dark visual result is deliberately unverified and belongs to whoever deploys — a spacing change is not something a test can judge
- Closing keyword patched in so #2211 auto-promotes

PR #2260, opened 2026-08-17
fix(workspace): align the composer buttons to the input, not to its wrapper (#2259)

### LINKED ISSUE #2259: bug(workspace): composer buttons align to the typeahead wrapper, not to the input — Send hangs 6px low
## Summary

On the Workspace composer (`/workspace`), the action buttons do not align with the input they sit beside: the Send button's bottom edge hangs **6px below** the input's bottom edge, and the attach/mic icons sit visibly low against the text line. The row is `flex items-end`, so the buttons *are* being bottom-aligned — just to the wrong box.

## Root cause

The typeahead wrapper introduced by ent#392 (`<div ref="composerWrap" class="relative flex-1 min-w-0">`) is a **block** container, and a `<textarea>` is **inline-block**. The textarea therefore sits on the wrapper's baseline, and the line box reserves descender space beneath it — the wrapper renders 6px taller than the textarea it contains.

`items-end` aligns the flex items, i.e. the **wrapper**, so the buttons align to the wrapper's bottom while the visible input edge is 6px higher. Measured live on `dev` (Chromium, 1440px):

```
wrapper  : top 693  bottom 745  h 52
textarea : top 693  bottom 739  h 46   <- 6px of dead space below the visible field
sendBtn  : top 705  bottom 745  h 40   <- flush with the wrapper, 6px below the input
```

Independently, the buttons are 40px tall (`p-2.5` + a 20px icon) against a 46px single-line composer, so even with the gap removed the icon's centre lands 3px below the text line's optical centre.

Note this is **not** #2211 — that fixed the composer's own height/scrollbar arithmetic. The gap is between the textarea and its wrapper and survives that fix (the numbers above are from a tree that has it).

## Acceptance Criteria

- [ ] The composer wrapper's height equals the textarea's height — no descender gap — so `items-end` aligns the buttons to the visible input edge
- [ ] Button bottoms are flush with the input's bottom edge in the collapsed (single-line) state
- [ ] Icon centres sit on the composer's text line (within ~1px) in both the collapsed and the grown (multi-line) state
- [ ] Buttons stay pinned to the bottom as the composer grows toward its `max-h-40` ceiling
- [ ] Applied to **both** `PortalConversation.vue` and `PortalRoom.vue` (same composer, two files — the #2211 lesson)
- [ ] The `PortalTypeahead` popup (`absolute bottom-full`) still clears the field; no regression to its anchoring
- [ ] Semantic tokens only per `docs/memory/design-system-contract.md`; verified light **and** dark
- [ ] Raw-color ratchet does not regress (`node src/frontend/scripts/scan-raw-colors.mjs src/frontend`)

## Technical Notes

- Files: `src/frontend/src/components/portal/PortalConversation.vue` (attach / mic / send), `src/frontend/src/components/portal/PortalRoom.vue` (send)
- `components/rooms/RoomComposer.vue` carries the same shape but is unreferenced — out of scope, do not revive it
- The 44px button box keeps the row on the 4px grid the design contract asks for; the composer's own 46px collapsed height is `leading-6` + `py-2.5` + a 1px border and is deliberately left alone

trinity-ability:
Fixes #2259

## What was wrong

The Workspace composer's buttons bottom-aligned to a box the user cannot see.

ent#392 wrapped the textarea in a `relative` div so the typeahead popup had something to anchor to. A `<textarea>` is **inline-block**, so inside that block wrapper it sat on the baseline and the line box reserved 6px of descender space beneath it. The row is `flex items-end`, which aligns the flex **item** — the wrapper — so Send hung 6px below the visible input edge. Measured on `dev` before the fix (Chromium, 1440px):

```
wrapper  : top 693  bottom 745  h 52
textarea : top 693  bottom 739  h 46   <- 6px of dead space
sendBtn  : top 705  bottom 745  h 40   <- flush with the wrapper, not the field
```

This is **not** #2211 — that fixed the composer's own height/scrollbar arithmetic, and the numbers above are from a tree that already has it. The reporter was running a stale checkout, which is why the scrollbar was in the screenshot alongside this.

## The fix

**`block` on the textarea** removes the line box, so the wrapper is exactly as tall as the field and `items-end` finally aligns the buttons to the input. It also re-anchors the popup's `absolute bottom-full` to the real field — it now clears it by its declared `mb-2` rather than 8px + a phantom 6px.

That alone left the icons 3px low, because the buttons were `p-2.5` around a 20px icon (40px, off the 4px grid) against a 46px single-line composer. They are now a **44px box with a centred glyph**: bottoms flush, icon centre within 1px of the text line in the collapsed **and** grown states — `items-end` measures from the bottom either way, so the icons track the last line as the field grows. 44px is also the WCAG 2.5.5 minimum target size, which 40px missed.

Landed in **both** composers (`PortalConversation` + `PortalRoom`) — the #2211 lesson.

### One extra, in-class

`PortalRoom` never initialised its field at all. It gained `autoGrow()` in #2211, but nothing called it at startup: the composer sits behind `v-if="!isClosed"` and does not exist when the component mounts, so an eager call no-ops on a null ref. It therefore rendered with `overflow-y` at its stylesheet `auto` — the state #2211 replaced with an explicit `hidden` ("the scrollbar must be gone rather than merely unused"). Invisible at rest only because the untouched `rows="1"` box happens to fit its one line, i.e. a coincidence rather than the contract. Now initialised after `load()` resolves, on mount and on room switch.

`components/rooms/RoomComposer.vue` carries the same shape but is unreferenced — deliberately left alone rather than revived.

## Verification

Measured in Chromium against this branch (local Vite → the running backend), on both surfaces:

| | 1:1 collapsed | 1:1 grown (3 lines) | room |
|---|---|---|---|
| wrapper gap | 0px | 0px | 0px |
| button bottom vs input bottom | 0px | 0px | 0px |
| icon centre vs text line | 1px | 1px (last line) | 1px |
| `overflow-y` | `hidden` | `hidden` | `hidden` |
| scrollbar rendered | no | no | no |
| typeahead clears field | yes (`mb-2`) | — | — |

Checked in light and dark. Typeahead popup opens, clears the field and stays in the viewport.

- `npx vitest run` — **884 passed (41 files)**
- New `tests/unit/portalComposerAlignment.spec.js` (9 assertions across both files), confirmed **red** on the pre-fix source and green after
- One existing assertion in `portalComposerTypeahead.spec.js` was anchored on `class="w-full` and read the added `block` as the field losing its width — rematched on the class list, same intent
- `node src/frontend/scripts/scan-raw-colors.mjs src/frontend` — no per-file regression (the change adds no colour classes)

Semantic tokens only; no new colours, no backend change, no migration.

### DIFF
--- a/src/frontend/src/components/portal/PortalConversation.vue
+++ b/src/frontend/src/components/portal/PortalConversation.vue
@@ -166,9 +166,17 @@
         </div>
         <form class="flex items-end gap-2" @submit.prevent="send">
           <input ref="fileInput" type="file" class="hidden" @change="onPickFile" />
+          <!-- #2259: the composer's action buttons are `h-11 w-11` (44px, on the
+               4px grid) rather than `p-2.5` around a 20px icon (40px, off it).
+               `items-end` pins them to the bottom so they stay beside the LAST
+               line as the field grows, and at 44px against the 46px single-line
+               composer the icon's centre lands within 1px of the text line in
+               both states. Sizing the box explicitly (instead of padding an
+               icon) also keeps the three buttons identical when one of them
+               swaps its glyph. -->
           <button
             type="button"
-            class="shrink-0 p-2.5 rounded-xl text-gray-400 hover:text-gray-700 dark:hover:text-gray-200 hover:bg-gray-100 dark:hover:bg-gray-800 transition"
+            class="shrink-0 h-11 w-11 flex items-center justify-center rounded-xl text-gray-400 hover:text-gray-700 dark:hover:text-gray-200 hover:bg-gray-100 dark:hover:bg-gray-800 transition"
             title="Attach a file for the agent"
             @click="fileInput?.click()"
           >
@@ -177,7 +185,7 @@
           <button
             v-if="sttSupported"
             type="button"
-            class="shrink-0 p-2.5 rounded-xl transition disabled:opacity-50"
+            class="shrink-0 h-11 w-11 flex items-center justify-center rounded-xl transition disabled:opacity-50"
             :class="listening ? 'text-status-danger-600 dark:text-status-danger-400 animate-pulse' : 'text-gray-400 hover:text-gray-700 dark:hover:text-gray-200 hover:bg-gray-100 dark:hover:bg-gray-800'"
             :title="micTitle"
             :aria-label="micTitle"
@@ -191,7 +199,16 @@
                wrapper is new. It must inherit the flex sizing the textarea used
                to carry (`flex-1 min-w-0`) — a bare `relative` div collapses the
                field to content width — and it deliberately carries no z-index,
-               so it creates no stacking context of its own. -->
+               so it creates no stacking context of its own.
+
+               #2259: it must ALSO not be taller than the textarea it wraps. A
+               `<textarea>` is inline-block, so inside this block wrapper it sat
+               on the baseline and the line box reserved 6px of descender space
+               below it. `items-end` aligns the flex ITEM — this wrapper — so the
+               buttons bottom-aligned to that dead space and Send hung 6px below
+               the visible input edge. The `block` on the textarea removes the
+               line box entirely; it also re-anchors the typeahead's
+               `absolute bottom-full` to the real field. Do not drop it. -->
           <div ref="composerWrap" class="relative flex-1 min-w-0">
             <PortalTypeahead
               v-if="typeaheadOpen"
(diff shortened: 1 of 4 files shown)


vybe (approved):
Validated via /validate-pr: bug fix with a named regression test (#2259) that was red-then-green, browser-measured on both composers, raw-colour scan clean, e2e + unit green. The one CodeQL hit was on the spec's own comment-stripper (test code, not a sanitizer) and is replaced by an index walk in ad237ad4. No docs required (bug fix). Approving.

Issue #2261, opened 2026-08-17
fix(workspace): a client session that EXPIRES on a browser holding a platform login falls back to the platform identity


trinity-ability:
## Summary

Residual of #2258, stated there rather than fixed. #2258 made the Workspace's **user-initiated** "Sign out" end whichever credential is live (`store.signOutEverywhere()` → `authStore.logout()` then `signOut()`). The **expiry** path was deliberately left alone: `fetchRoster`'s 401 → `endSession({expired:true})` → `signOut()` clears only the portal token, and if a platform JWT is also present in the browser, `isPlatformSession` re-derives and the client silently becomes the operator — the same class as #2258, by a different door.

## Why it was left

- Expiry is not a user act. Ending the platform session there would log out an operator working in another tab because a client's idle timeout lapsed.
- The alternative — a persisted "suppress the platform fallback" flag — was rejected in #2258 on evidence: the platform JWT is an axios DEFAULT header and per-request headers merge over defaults, so a flag hides the disclosure while every portal request still carries the operator's credential (see `docs/memory/feature-flows/workspace-session-signout.md` → *Why not the suppression flag*).

## Reachability

Narrow. The Workspace OTP form never renders while a platform JWT exists (`Portal.vue` gates it on `!isClientSignedIn`, and `isPlatformSession` is already true), so this needs the platform login to arrive **after** the client signed in on the same browser (a shared machine where an operator logs in later), and then the client session to idle out (7d default) with that tab still open or revisited.

## Acceptance Criteria

- [ ] After a client session expires on a browser that also holds a platform login, the Workspace shows the "session expired" sign-in state — not the operator's roster — and a refresh keeps it there.
- [ ] An operator's platform session in another tab is **not** ended by a client's idle timeout.
- [ ] Whatever the mechanism, the wire agrees with the screen: no portal request carries the platform JWT while the UI says "signed out" (the #2258 constraint).
- [ ] Unit coverage in `src/frontend/tests/unit/workspaceSession.spec.js`.

## Technical Notes

The honest options are both non-trivial, which is why this is filed rather than folded into #2258:
1. A **positive** in-memory entry marker (`workspaceEntered`, NOT persisted) that `isPlatformSession` requires and that expiry clears — fails closed to the sign-in form when lost, and needs no localStorage; but it needs a "Continue as <email>" escape for the operator, and a live-request guard equivalent to the flag's (the axios-default problem still applies to any state that hides a live JWT).
[...]

PR #2262, opened 2026-08-17
fix(workspace): "Sign out" signs out — end the live credential, not a derivation of it (#2258)

### LINKED ISSUE #2258: fix(workspace): Sign out doesn't sign out — the platform JWT silently re-authenticates on refresh
## Summary

Signing out of the Workspace does not produce a signed-out state. `signOut()` clears only the portal session; the platform JWT in `localStorage['token']` is left in place, and because a platform session is an *implicit* way to be signed into the Workspace (ent#357), removing the portal token is precisely what **activates** the platform fallback. Refreshing the page re-enters the Workspace under the platform identity.

## Context

Reported from live use: "click sign out, refresh the page, and you're signed in again — it looks like it's taking the login from the main system." Confirmed in code; no credential leak, but "Sign out" is a button that does not sign the user out, and on a shared machine the next person is one refresh away from the previous user's agents.

Two variants, both from the same cause:

1. **Signed-in platform user in the Workspace.** There is no portal token to clear, so `isClientSignedIn` stays `true` and the app shell stays mounted — but with an emptied roster, blank email and no threads. It looks half-signed-out; a refresh restores everything.
2. **External client signing out in a browser that also holds a platform login.** Their portal token *is* dropped, then the store silently falls back to the platform `Authorization` header — re-authenticating as the platform user, with that user's email and roster.

## Root cause

`ent#357` made a platform session implicit but never updated the sign-out path:

- `src/frontend/src/stores/clientPortal.js` — `isPlatformSession()` is `!portalToken && authStore.isAuthenticated`; `isClientSignedIn()` is `!!portalToken || isPlatformSession`; `authHeader()` falls back to the platform header when no portal token exists.
- `src/frontend/src/stores/clientPortal.js` — `signOut()` clears `portalToken`, `clientEmail`, `agents`, `rosterLoaded` and removes the portal token key from `localStorage`. It never touches the platform session.
- `src/frontend/src/components/portal/PortalSidebar.vue` — the footer Sign out button renders unconditionally, for both principal kinds.
- `src/frontend/src/views/Portal.vue` — the shell gates on `store.isClientSignedIn`, and `onMounted` re-runs `bootstrap()` whenever it is true.

The store comment reasons only about the opposite direction ("signing out of the *platform* ends the Workspace session, by construction" — which is true). Signing out *of the Workspace* was never handled.

## Severity

Not a credential leak: the platform JWT is the browser owner's own credential, the client's portal token really is removed, and the backend re-scopes the roster to whichever identity the header carries (`get_portal_principal`). This is an identity/UX defect plus a shared-machine hazard.

## Acceptance Criteria

- [ ] Sign out from a **platform session** ends that session — no state where the shell is mounted with a blank email and an empty roster, and no reappearance on refresh.
- [ ] Sign out from a **client session** in a browser that also holds a platform login lands on the Workspace sign-in form, not on the platform user's roster. Refresh keeps it there.
- [ ] The suppression survives a reload (the fallback cannot re-derive from `localStorage` on the next `bootstrap()`), and is cleared by the next explicit sign-in of either kind.
- [ ] Signing out of the platform still ends the Workspace session (the ent#357 property is preserved, not regressed).
- [ ] The button's label/affordance matches what it does for each principal kind — an operator is never shown "Sign out" for an action that only leaves the page (honest status; no dead end).
- [ ] Unit coverage in `src/frontend/tests/unit/workspaceSession.spec.js` pins both directions: post-sign-out `isClientSignedIn === false` with a platform JWT present, and the existing implicit-entry behaviour unchanged for a user who never signed out.

## Technical Notes

Frontend-only; no backend change. Two decisions to make explicit rather than infer:

- **Platform session:** simplest correct behaviour is for Sign out to call `authStore.logout()` and route to the platform login — same person, and the Workspace is the platform's own surface since ent#356. Relabelling to "Leave workspace" (→ `/`) is the weaker alternative: it still leaves a Sign-out affordance elsewhere that doesn't sign out.
- **Client session over a live platform JWT:** needs a sticky, persisted suppression flag that `isPlatformSession` honours until the next explicit sign-in — clearing the portal token alone will always re-derive the platform identity.

Related but distinct: ent#281 covers *operator-initiated* client revocation (server-side session cutoff). This is the *user-initiated* client-side path and is not covered by it.

trinity-ability:
## Summary

- **"Sign out" now signs out — for both ways of being signed in.** `signOut()` cleared only the portal token, and since ent#357 a platform JWT is an implicit workspace session (`isPlatformSession = !portalToken && isAuthenticated`), removing the portal token was precisely what *activated* the fallback: a platform user's refresh re-entered the Workspace, and a client on a browser also holding a platform login was re-authenticated as that operator.
- **Fix: destroy the credential, never a derivation of it.** `store.signOutEverywhere()` ends the platform session (`authStore.logout()`) **first**, then clears portal state, and routes by principal — operator → `/login`, client → the Workspace OTP form. Order is load-bearing (no window where `portalToken` is gone while `isAuthenticated` is true). `signOut()` stays the plain primitive so `endSession({expired})` never ends a platform session.
- **The issue's suggested suppression flag was rejected on evidence** (recorded in the flow doc + CSO report): the JWT is an axios DEFAULT header and per-request headers merge over defaults, so a flag hides the disclosure while every portal request still carries the operator's credential; `streamPortalExecution` (bare `fetch`) would 401 while axios paths stayed authorized; a wrong OTP digit would trip the global bounce onto the operator `/login`.
- `auth.logout()` clears local state **before** the network revoke — the 401 interceptors and the `/login → /` guard both key on it (NavBar's un-awaited `logout(); push('/login')` landed on the dashboard since #187). The revoke still carries the token.
- One footer button (no v-if swap, #2159), labelled per principal: **"Sign out of Trinity"** / **"Sign out"**. A "Signing out…" frame holds the shell so a departing operator never sees the client OTP copy.

## Changes

- `src/frontend/src/stores/clientPortal.js` — `signOutEverywhere()`, `PLATFORM_LOGIN_ROUTE`
- `src/frontend/src/stores/auth.js` — `logout()` local-clear-before-revoke
- `src/frontend/src/views/Portal.vue` — `onSignOut` delegates; `signingOut` frame
- `src/frontend/src/components/portal/PortalSidebar.vue`, `portalUtils.js` — `signOutLabelFor`
- `src/frontend/tests/unit/workspaceSession.spec.js` (+8), new `workspaceSignOut.spec.js` (7)
- Docs: new `docs/memory/feature-flows/workspace-session-signout.md`, index, `architecture.md` Workspace block, `learnings.md` (2), `docs/security-reports/cso-diff-2026-08-17-2258-workspace-signout.md`

## Residuals (stated, not hidden)

- A **client** session that *expires* on a browser which later gained a platform login still falls back to it — same class by an ordering the UI does not produce. Filed as #2261 (incubating).
- The client portal token stays **server-side valid** after sign-out (client-side disposal; no self-service revoke — the ent#281 primitive is per-email). Deferred by decision.
- A client signing out on a browser that also holds an operator's login ends the operator's session too — by design.

## Test Plan

- [x] `npx vitest run` — 890/890 green (15 new)
- [x] Both variants browser-verified against a live backend from this branch: platform user → `/login`, revisit `/workspace` shows the OTP form with no platform token; client + platform JWT (real minted portal token, revoked after) → stays on `/workspace` OTP form, refresh keeps it there
- [x] `/review` 0 critical · `/cso --diff` 0 findings · enterprise-docs guard clean over every doc addition
- [ ] Reviewer: click Sign out as a platform user in the Workspace; confirm `/login` and no reappearance on refresh

Fixes #2258


vybe (approved):
Validated via /validate-pr: bug fix (#2258) with named regression tests (workspaceSignOut.spec.js + 8 in workspaceSession.spec.js), architecture + feature-flow + learnings + CSO diff report all in-diff, /review + /cso recorded 0 findings, browser-verified for both principal kinds, security greps clean, 754/754 frontend unit tests locally + full CI green. Approving.

PR #2263, opened 2026-08-17
fix(dashboard): keep the Executions tile's failure rails on the chart baseline

vybe:
## Summary
- The fleet **Executions** grid tile drew its red failure rails hanging **below** the baseline of the other columns.
- Root cause: `.ex-col` spaces stack + rail with `gap: 2px`, and the rail `<i>` was rendered for every column even at `height: 0`. Flex gap applies to a zero-height item too, so every failure-free stack floated 2px above the chart bottom while a failed column's rail reached it.
- Fix: `v-if` the rail so it exists only when there is one to draw. Every column now ends on one baseline; failed columns show the red rail flush on it with the stack lifted above. Chart budget (`CHART_HEIGHT + COL_GAP + RAIL_HEIGHT + pad`) is unchanged.
- Spec pin added in `executionsTile.spec.js` (a source guard — a 0px rail and an absent one are indistinguishable to the pure shaping module). Verified it fails on the pre-fix template.

## Test plan
- [x] `npx vitest run tests/unit/executionsTile.spec.js` → 38 passed
- [x] Live check on the local dashboard grid: measured every `.ex-col` — rail-less stacks and failure rails all end at the chart bottom (previously stacks ended 2px above it)

### DIFF
--- a/src/frontend/src/components/tiles/ExecutionsTile.vue
+++ b/src/frontend/src/components/tiles/ExecutionsTile.vue
@@ -54,8 +54,13 @@
           ></i>
         </span>
         <!-- The failure rail sits BELOW the stack on its own scale, so failures
-             are never hidden inside a column and never steal height from it. -->
-        <i class="ex-rail" :style="{ height: col.failPx + 'px' }"></i>
+             are never hidden inside a column and never steal height from it.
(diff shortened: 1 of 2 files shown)


trinity-ability (approved):
Verified: 38/38 vitest, live geometry check confirms every column now ends on the chart baseline.

PR #2278, opened 2026-08-18
feat(onboarding): ambient activation guidance — checklist surface + schedules empty-state action (ent#238)

dolho:
## Summary

The OSS half of the ambient onboarding layer (ent#238, epic ent#54). The setup wizard gets a user to a first agent; nothing afterwards carries them to chat → schedule → channel, and one first-hour view still dead-ended.

Two pieces:

**1. Activation checklist** — `components/onboarding/ActivationChecklist.vue` + `stores/onboarding.js`, mounted on the Dashboard. A dismissible, non-gating card listing the four first-value milestones. It ships in the OSS bundle and reads a server-driven surface: **404** (router not mounted) or **403** (not entitled) leaves it invisible *and silent*. That silence is the design — ambient guidance that shouts about its own absence is worse than no guidance. Only a genuine failure logs.

The store keeps **no progress of its own**: milestones are derived server-side from verified rows on every read, so an item unticks when the state behind it goes away. The card hides itself entirely once the last milestone is reached — the reward for finishing is that it stops asking.

**2. Schedules empty state** gets the primary action its siblings already have. Dashboard ("No agents yet" → Get started), Chat (`ChatEmptyState`) and Operations ("All caught up") already carried explainer + action; the Schedules panel explained and then dead-ended. Scope stayed the first-hour path the issue specifies, not an app-wide audit.

## Design notes

- **Dismissal is optimistic on purpose** — it hides before the write lands. A dismissal that waits on the network feels broken, and a failed write costs the user nothing worse than seeing the card again next load.
- **Only the next undone item carries an action**, so the card reads as one step to take rather than four competing buttons.
- **Never a mandatory tour** (the ent#54 principle): dismissible, gates nothing, self-hiding.

## Already satisfied on dev — recorded, not re-done

Two of the issue's acceptance criteria were already true before this branch, verified before planning:

- **Dead-code removal is done.** Neither `components/OnboardingChecklist.vue` nor `composables/useOnboarding.js` exists on dev, and nothing references either name. Nothing to delete — saying so beats pretending to do work.
- **ent#184 landed**, so the milestone signal source the issue requires ("no parallel tracking store") is in place.

## Testing

`tests/unit/onboardingChecklistStore.spec.js` — 10 tests covering the two load-bearing behaviours no structural check can see: silence on OSS/unentitled (404 and 403 log nothing; a 500 logs once and stays unavailable), and visibility going false independently on completion vs dismissal. Plus fetch-once-unless-forced, optimistic hide before the write resolves, hide surviving a failed write, and the restore seam.

```
$ npx vitest run
Test Files  43 passed (43)
     Tests  910 passed (910)

$ npx vite build
✓ built in 3.40s
```

## Depends on

The gated backend lives in the private repo (`trinity-enterprise#412`). The **submodule pin is deliberately not bumped here** — it would pin an unmerged commit. Order: merge the enterprise PR → bump the pin → merge this.

Until the pin moves, this PR is inert by construction: the endpoint 404s, the checklist renders nothing, and only the Schedules empty-state action is user-visible.

Fixes Abilityai/trinity-enterprise#238


vybe (approved):
Validated: fail-silent entitlement surface, derivation-only store, non-happy-path store tests; Dashboard.vue/architecture.md conflicts vs ent#319 resolved keeping both panels.

PR #2279, opened 2026-08-18
feat(onboarding): first-run front desk — three doors into a seeded install (ent#319)

dolho:
## Summary

First-run seeding silently broke the onboarding wizard it shipped alongside.

`Dashboard.vue::maybeAutoOpenOnboarding` fires only when there are **zero non-system agents**. Since trinity-enterprise#124, a fresh install boots with Cornelius plus the bundled fleet already running — so the trinity-enterprise#52 *"one question → a tailored agent"* wizard **stopped auto-opening on exactly the install it was written for**. A new user meets four agents they did not choose and no route into anything.

This adds the routing surface the epic (trinity-enterprise#54) is missing: a dismissible Dashboard card with the three doors.

```
┌────────────────────────────────────────────┐
│  Start here                             ✕  │
│  Trinity runs agents. Here are three ways  │
│  in — none of them commit you to anything. │
│                                            │
│  [ Show me ]   watch cornelius work        │
│  [ Make me one ]  one question → an agent  │
│                                            │
│  Already run a fleet? Bring it over →      │
└────────────────────────────────────────────┘
```

**Bring mine is deliberately a de-emphasised link, not a peer button** — the issue's own instruction: fleet migration is our strongest capability and our worst first impression, asking the most from the person who knows us least.

## Scope note — read this before reviewing against the issue text

trinity-enterprise#319 was filed proposing an agent that **replaced** the seeded fleet ("a fresh install deploys no agent the user did not choose"). That premise was **reversed** when trinity-enterprise#322 was closed not-planned: a fresh install *does* roll out an agentic system, and *which* agents it contains belongs to trinity-enterprise#137.

So:

- The front desk here is a **surface, not an agent** — nothing new is seeded, no manifest changes, zero collision with ent#137's held branch.
- The issue's *"exercises the Trinity MCP in its first exchange — operate, don't narrate"* criterion is a property of the **seeded agent's template**, which is ent#137's scope. It is **not delivered here**, and is flagged rather than silently dropped.

## How "is this still the first run?" is answered

The seed deploys under the **admin account** (`system_seed_service.SEED_OWNER`), so it is indistinguishable from a human admin's own work in `audit_log` — no actor-based test can answer this. What *can* is the seeder's own naming contract, which is deterministic:

| Piece | Role |
|---|---|
| `system_seed_service.seeded_agent_names()` *(new, public)* | `{system}-{short}` from the manifest **actually in force** — honours `TRINITY_DEFAULT_SYSTEM_MANIFEST` and its disable sentinels (identical check to the seeder's own). Read-time derivation: no new persisted state, no migration, and an operator seeding their own manifest gets their own names. |
| `services/onboarding_service.py` *(new)* | Composes that with `CORNELIUS_AGENT_NAME` and answers the predicate: `first_run` is true while **every agent the caller can see is one Trinity seeded**. |
| `GET /api/onboarding/first-run` *(new)* | `{first_run, seeded_agents, own_agent_count, demo_agent}`. OSS-core, un-gated — this is the open install's front door. |

Both halves of the seeded set stay owned by the service that seeds them, so this composes rather than re-derives.

## Decisions worth reviewing

- **DB-only read.** Visibility comes from `db.get_all_agent_metadata` (owner / shared / is_system), *not* the Docker-backed `get_accessible_agents`: the question is about ownership rows, so the card renders truthfully when a container is down.
- **The two first-run surfaces never stack.** The card requires a non-empty seeded set, so a genuinely empty install (seeding disabled) still gets the wizard's auto-open, **unchanged**. That includes CI e2e, which sets `TRINITY_DEFAULT_SYSTEM_MANIFEST=disabled` — so no visual-regression surface here.
- **Everything fails toward hidden.** Any read failure resolves to `first_run: false`. A missed nudge is a non-event; a "Start here" card over somebody's mature fleet is noise no user can dismiss on the install's behalf.
- **Dismissal is per-browser localStorage**, matching the wizard's existing `trinity_onboarding_dismissed` key. ent#238's checklist persists dismissal server-side because that was an explicit AC there; a per-user table for a card that stops appearing the moment you create an agent would be more machinery than the behaviour is worth.
- **A non-admin's view is their own.** A colleague's agent is not this user's first-run evidence — bob has built nothing, so bob is still on his first run (tested).

## Testing

`tests/unit/test_ent319_first_run_front_desk.py` — 13 tests: seeded-name derivation (system-prefixed + Cornelius, disable sentinel, unreadable manifest, operator override), the predicate (seed-only → first run, one own agent ends it, empty install has nothing to show, system agents don't count, non-admin scoping, shared-agent case, demo-agent preference), and both failure paths.

`src/frontend/tests/unit/firstRunFrontDesk.spec.js` — 9 tests: visibility incl. the stand-down-on-empty-install rule, no flash before the answer arrives, fail-toward-hidden, dismissal + persistence + storage refusal.

```
$ pytest tests/unit/test_ent319_first_run_front_desk.py -q
13 passed

$ pytest tests/unit/test_ent124_default_system_seed.py tests/unit/test_ent107_cornelius_seed.py \
         tests/unit/test_models_centralized.py tests/unit/test_ent319_first_run_front_desk.py -q
88 passed

$ npx vitest run
Test Files  43 passed (43)
     Tests  909 passed (909)

$ npx vite build
✓ built in 3.04s
```

Full local backend suite: **4909 passed, 1 failed** — `test_736_a2a_outbound_edges::test_C1b_mapped_addresses_either_side_of_the_cgnat_block_stay_public`. That failure is **pre-existing and environmental**, not from this branch: it reproduces identically on an unrelated branch that does not touch A2A (all 5 params fail there in 0.58s). CI is the authority on it.

## Docs

`docs/memory/requirements/roadmap.md` §16.5.1a (new, next to the seeder it depends on) and two `architecture.md` entries (frontend surface + the endpoint).

Fixes Abilityai/trinity-enterprise#319


vybe (approved):
Validated: DB-only first-run read, fail-toward-hidden, both first-run surfaces never stack; e2e + pytest matrix green; scan clean.

PR #2284, opened 2026-08-18
feat(agents): pre-install the Trinity plugin so a bare repo can onboard itself (ent#411)

dolho:
## Summary

`/trinity:onboard` is what turns a deployed agent into a Trinity-compatible one — it writes `template.yaml`, `.env.example`, `.gitignore`, `.mcp.json.template` and pushes them back. But #1704 reads the plugin set from `template.yaml`, and **the agent that needs onboarding most is a bare GitHub repo that has no `template.yaml`**. It declares nothing → nothing installs → the plugin that would write the declaration is the one thing missing. The escape was a prose instruction telling the agent to run the CLI itself.

`create_agent(template: github:owner/repo)` already tolerates a missing `template.yaml`, so *deploy-as-is* worked; only *onboard-in-place* was blocked. This makes the Trinity plugin something the platform **provides** rather than something you declare.

## What's here

| | |
|---|---|
| `docker/base-image/Dockerfile` | pre-installs `trinity@abilityai` (`ARG TRINITY_PREINSTALL_PLUGINS=1`). Docker populates an empty named volume from the image on first mount, so a new agent inherits it and boots with **zero subprocesses** |
| `plugins_reinstall.py` | ensures the platform set every boot regardless of what is declared — the self-healing path for a volume that predates the image, an air-gapped build, or a failed install |
| `startup.sh` | **guard widened** — see below |
| `I-006` (new INFO check) + `.trinity/plugins-state.json` | the honest-status surface |
| docs | agent guide, `architecture.md`, requirements §42.5, `agent-validation-spec.md` |

## The part worth reviewing: the startup guard

#1704 ran the reconciler **only when a manifest or a `plugins:` block existed**. Under this feature that means the bare repo is the single agent that never runs the hook — the pre-install would have been the only delivery path, and every agent on an older volume would have stayed broken *silently*, since nothing errors. The guard now also fires when platform defaults are on; `TRINITY_PLATFORM_PLUGINS=0` restores the old skip exactly. A static test pins it so it cannot narrow back.

## Properties, each pinned by a test

- **Additive, never subtractive** — a `plugins:` block that omits `trinity@abilityai` does not uninstall it. Nothing in this module uninstalls anything.
- **The platform marketplace name is pinned to its source** — the manifest is on the agent-writable volume, so a declaration re-pointing `abilityai` at another repo is ignored. A redefinable platform marketplace is not a convenience, it is an arbitrary-code-fetch primitive.
- **Never fatal** — unreachable marketplace at build time logs and the image still builds; at boot it is withheld with a reason and startup continues.
- **Opt-out at both ends** — `TRINITY_PLATFORM_PLUGINS=0` (runtime) and `--build-arg TRINITY_PREINSTALL_PLUGINS=0` (air-gapped builds).

## Honest status (the issue's AC)

Each reconcile is recorded to `~/.trinity/plugins-state.json` and surfaced by **I-006**. "The marketplace was unreachable" and "the operator never wanted it" are different facts a bare presence flag cannot separate — so the check reports the reconciler's own withheld reason, treats a switched-off platform set as a **skip**, and treats a missing state file as a **skip** (an image predating this is not a failed install). The file is agent-writable, so the check cross-checks the claim against the recorded lists rather than trusting a free-text `status`.

## Testing

```
$ pytest tests/unit/test_ent411_platform_plugins.py -q          # 19 passed
$ pytest tests/unit/test_ent411_platform_plugins.py \
         tests/unit/test_1704_plugins_reinstall.py \
         tests/unit/test_compatibility_checks.py -q             # 152 passed
$ pytest tests/unit/test_2070_trinity_authored_paths.py \
         tests/unit/test_ent128b1_compat_gates.py \
         tests/unit/test_1908_bundled_template_gitignore.py -q  # 158 passed
```

Shell paths were executed, not just linted: both Dockerfile branches run (opt-out prints and skips; a `claude` that exits 1 still exits 0 with the withheld line), and all three startup-guard paths were exercised against the extracted guard (bare+on → runs, bare+off → skips, declared+off → still runs).

**Honest gap:** a full local unit-suite run was in flight when I re-organised the worktree, so its result is invalid rather than green — CI is the authority here. Targeted suites above are real. The base image itself is not built by CI, so the Dockerfile change is verified by shell execution and review, not by a build.

## Out of scope

The in-place mode of `/trinity:onboard` — detect-in-container, write, push back or emit a patch, verify via `get_agent_compatibility_report` — is owned by the marketplace (`abilityai/abilities`), as the issue states.

**Release-note ordering caveat:** an agent on a base image built before this lacks the pre-install and pays one install at next boot; rebuild the base image for the zero-subprocess path.

Fixes Abilityai/trinity-enterprise#411


vybe (approved):
Validated: base-image plugin pre-install with pinned marketplace source, additive-only reconcile, honest-status surface; full pytest matrix green; security scan clean.

PR #2285, opened 2026-08-18
feat(settings): room budget defaults panel (ent#387)

dolho:
## Summary

The OSS half of ent#387 — the Settings panel where multi-agent **room budgets** are set now that ent#381 retired the Sessions page (and with it `NewRoomDialog`, the only surface that ever set one).

Sits in `Settings → Data & retention`, beside the ent#375 Workspace session policy, because both bound what a client engagement consumes.

The gated backend half is `Abilityai/trinity-enterprise#416`, which is also where the sharper finding lives: `POST /api/rooms` accepted budget fields from **any** principal including a workspace client, so the party being bounded could set its own bound. That is closed server-side; this panel is where the person with the authority exercises it.

## Design notes

- **An operator surface, not a Workspace field.** A budget bounds what a client conversation may spend, so putting the dial in the client's "start a chat" dialog hands it to the person being bounded. The backend draws the same line, so this UI is not the enforcement — it is the control.
- **Self-hiding when the rooms module is not entitled.** Unlike the session policy, whose *mechanism* is OSS and therefore worth showing in every edition, there is nothing here for a community install to look at.
- **The empty cost box is the one subtle case.** The PUT is a partial update, so an omitted field means "leave it alone" — read naively, emptying the box omits the field and the previous cap silently survives a change the operator believes they made. Clearing is therefore explicit, and only sent when there is something to clear.
- That logic lives in `utils/roomBudgets.js`, **not inline in the component**, precisely so it can be tested.
- Per-field `configured` / `default` labels follow the #1638 rule: a shipped value a future default change may move must not look like a deliberate setting.

## Testing

```
$ npx vitest run tests/unit/roomBudgets.spec.js     # 8 passed
$ npx vitest run                                    # 43 files, 908 passed
$ npx vite build                                    # ✓ built in 3.02s
```

The 8 cover: nothing-changed sends nothing, only-changed-fields are sent, emptying the box clears, an already-absent cap is **not** re-cleared, whitespace counts as empty, `ttl 0` is *sent* rather than omitted (0 means "no expiry"), dirty-on-clear, and the no-server-state-yet case.

**Merge order:** the enterprise PR first, then the submodule pin, then this — the panel 404s until the endpoint exists (it fails to a named load error, not a blank card).

Related to ent#387


vybe (approved):
Validated via /validate-pr: OSS panel half of ent#387 (enterprise side ent#416 + pin #2297 merged first, per stated order). Full check suite green after dev merge (pytest ×6, e2e, build, CodeQL, gitleaks); security scan clean; 8 unit tests cover the clear-vs-omit partial-update semantics incl. non-happy paths. learnings.md conflict with #2287 resolved keeping both entries.

PR #2286, opened 2026-08-18
chore(enterprise): bump submodule pin to main tip (ent#238 backend + ent#323 design)

trinity-ability:
## Summary

Moves `src/backend/enterprise` 65182c1 → 9401b75 — exactly the two enterprise merges of today:

- trinity-enterprise#412 — ent#238 per-user activation checklist (gated `onboarding` module)
- trinity-enterprise#414 — ent#323 zero-commitment-door design doc (docs-only)

Required before #2278 (the OSS half of ent#238) merges, per its stated order: enterprise PR → **pin bump** → OSS PR. Until #2278 lands, the new module simply registers; no OSS surface reads it.


vybe (approved):
Pin gap verified: exactly the two enterprise merges (ent#412, ent#414).

PR #2287, opened 2026-08-18
docs(learnings): global-hook stubs must honour their argument (ent#379)

dolho:
## Summary

One ledger entry, captured while fixing `Abilityai/trinity-enterprise#379` (fix PR: Abilityai/trinity-enterprise#417).

A test patched `importlib.import_module` with a fake that **ignored its argument** and returned a stub for everything. That hook is process-global, so the fake answered for every lazy import inside the window — including `requests`, whose `compat.py:43` resolves chardet through `importlib.import_module` and whose `packages.py:17` then reads `chardet.__name__`. The test scheduled *first* died; the rest passed only because requests was already imported by then. 8 of 11 failed under one seed, and the suite had looked green for weeks.

Two sibling shapes in the same repo are recorded with it: a `sys.modules` swap that a lazy `from pkg import mod` bypasses (the attribute on the package wins), and a provider left in a module-global by one test and inherited by every later one.

The tell for all three is the same and costs one command: **run the single test alone.** A test that passes only in company is depending on a sibling's side effect.

Docs-only — no code changes in this repo. `/autoplan` reads this file before planning, which is the reader it is written for.

Related to ent#379


vybe (approved):
Validated via /validate-pr: docs-only ledger entry (learnings.md), all checks green (pytest ×6, CodeQL, gitleaks), security scan clean. Cross-tracker sibling of trinity-enterprise#417 (merged).

PR #2298, opened 2026-08-19
fix(workspace): an expired client session no longer becomes the operator (#2261)

### LINKED ISSUE #2261: fix(workspace): a client session that EXPIRES on a browser holding a platform login falls back to the platform identity

dolho:
Fixes #2261.

## Summary

#2258's stated residual. Expiry cannot end the platform credential — that would log an operator out of another tab over a client's idle timeout — so `endSession → signOut` cleared only the portal token and `isPlatformSession` re-derived from the platform JWT still in the browser. The client's tab silently became the operator's.

## Why the flag the issue suggested works now

#2258 rejected it on evidence, and the evidence was about the **wire**: `auth.js` installs the platform JWT as an `axios.defaults` header, per-request headers *merge* over defaults, so a flag hides the operator's roster on screen while every workspace request still carries the operator's credential. So this fixes the wire first:

| Half | What it does |
|---|---|
| **The wire** | Every workspace call moves onto a dedicated instance (`portalHttp`) whose request interceptor decides the credential and strips whatever `axios.defaults` merged in. `authHeader` returns `{}` while suppressed. |
| **The screen** | `platformFallbackSuppressed`, set when a client session **expires** here; `isPlatformSession` returns false while set. sessionStorage — survives a refresh of *this tab* and nothing wider, which is also what leaves the operator's other tab alone. |

Neither half works alone, and that is the whole design. `streamPortalExecution` uses bare `fetch` with `authHeader` only, so it was already honest and stays so.

**The escape hatch is mandatory, not a nicety.** The suppression fails closed, so it necessarily catches the operator when the browser is genuinely theirs — "Continue as `<email>`" is the one explicit click back in. Re-deriving automatically is the bug.

## Three things surfaced while building it

**1. The 401 bounce had to move with the transport — and off the wrong question.** `main.js` decided with `onWorkspace && localStorage['token']`. Workspace calls no longer reach it, so the operator bounce is re-registered via `setPlatformSessionLostHandler` and fires only when `isPlatformSession`. That's false in exactly the case the old predicate got wrong: a client at the OTP form on a browser holding an operator's JWT, where **one mistyped digit would 401, bounce them to /login, and destroy the operator's session** — objection 3 in `workspace-session-signout.md`, now closed by construction. Without the move, an operator whose JWT expired on the workspace would have sat on "Failed to load your agents" forever.

**2. `sessionExpired` was never rendered.** The store has set it since ent#375 and no view read it, so an idle-out was indistinguishable from "you were never signed in" — the form just reappeared. It says which now.

**3. `resumePath` was never read either**, so the notice's "pick up where you left off" was a promise the app didn't keep. The sign-in that consumes the expiry navigates there.

## Acceptance criteria

| AC | Where |
|---|---|
| Expired client on a platform browser sees the expired sign-in, and a refresh keeps it | `shows the expired sign-in state`, `and a refresh keeps it there` |
| Operator's session in another tab is not ended | `leaves the operator's platform session alone` (+ `__logout` never called) |
| The wire agrees with the screen | `sends no credential while the UI says signed out`, `strips a platform JWT merged in from axios defaults` |
| Unit coverage in `workspaceSession.spec.js` | 37 tests (was 28) |

## A test-harness bug worth flagging

The spec's axios mock returned **one shared instance** for every `create()`, so `api.js`'s PERF-269 dedupe wrapper (which replaces `.get` on its own instance) was also replacing the workspace store's — a stub on it was silently not the code path, and it surfaced only because this PR moved the store onto an instance. `create()` now returns a fresh instance, like real axios.

## Testing

```
$ npx vitest run tests/unit/workspaceSession.spec.js   # 37 passed
$ npx vitest run                                       # 45 files, 940 passed
$ npx vite build                                       # ✓ built
```

Not covered by the unit suite (node-only, no DOM): the two new template branches render only in a browser. Their conditions are pure computeds over tested store state, and both carry `data-testid`s for an e2e follow-up.

## Docs

`docs/memory/feature-flows/workspace-session-signout.md` — the residual is struck through with the reason the flag became viable, plus a new "The expiry path (#2261)" section covering both halves, the escape, the bounce move, and the private-mode degradation.

Related to #2261

### DIFF
--- a/docs/memory/feature-flows/workspace-session-signout.md
+++ b/docs/memory/feature-flows/workspace-session-signout.md
@@ -107,14 +107,57 @@ in lockstep, `authHeader` divergence, an escape hatch for an operator dead-ended
 the OTP form, storage-write failure, cross-tab staleness). Credential destruction
 carries none, and leaves the 401 predicate correct **by construction**.
 
+
+## The expiry path (#2261)
+
+Expiry still cannot end the platform credential — an operator working in another
+tab must not be logged out because a client's idle window lapsed — so the fix
+breaks the *derivation* instead, for the tab where the client's session died.
+
+**Two halves, and neither works alone.**
+
+1. **The wire.** Every workspace call moved onto a dedicated axios instance
+   (`portalHttp`), whose request interceptor decides the credential and strips
+   whatever `axios.defaults` merged in. `authHeader` returns `{}` while the
+   fallback is suppressed. This is the half that answers the objection above: with
+   it, "the workspace is signed out" is a statement about the wire, not only the
+   screen. (`streamPortalExecution` uses bare `fetch` with `authHeader` only, so it
+   was already honest and stays so.)
+2. **The screen.** `platformFallbackSuppressed` — set when a client session
+   EXPIRES here, cleared by `signOut()`, by a client signing in again, and by the
+   operator's explicit "Continue as <email>". `isPlatformSession` returns false
+   while it is set. Stored in **sessionStorage**: it must survive a refresh of this
+   tab and nothing wider, which is also what keeps an operator's other tab
+   untouched.
+
+**The escape hatch is mandatory, not a nicety.** The suppression fails closed, so
+it necessarily catches the operator when the browser is genuinely theirs. One
+explicit click (`continueAsPlatform`) is the way back; re-deriving automatically is
+the bug.
+
+**The 401 bounce moved with the transport, and onto a better question.** The global
+`axios` interceptor in `main.js` decided with `onWorkspace && localStorage['token']`.
+Workspace calls no longer pass through it, so the operator bounce is re-registered
+via `setPlatformSessionLostHandler` and fires only when `isPlatformSession` — which
+is false in exactly the case that predicate got wrong: a client at the OTP form on a
+browser holding an operator's JWT, where one mistyped digit would have 401'd, bounced
+them to `/login` and destroyed the operator's session (objection 3 above).
+
+**`resumePath` is finally read.** `endSession` has recorded where the client was
+since ent#375 and nothing consumed it, so the expired notice's "pick up where you
+left off" was a promise the app did not keep; the sign-in that consumes the expiry
+now navigates there. The expired notice itself was also unrendered until #2261 — the
+form simply reappeared, indistinguishable from "you were never signed in".
+
+**Degradation:** if `sessionStorage` is unavailable (private mode), the marker reads
+as absent — pre-#2261 behaviour, rather than a workspace nobody can enter.
+
 ## Stated residuals (not hidden)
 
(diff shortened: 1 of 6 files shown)


vybe (approved):
Validated via /validate-pr: CI fully green (6 pytest seeds, e2e, build, gitleaks), security scans clean, feature-flow doc updated, regression specs named to the issue. Approving.

Issue #887, opened 2026-05-18
bug: Read-only mode hooks don't block Write/Edit tools, only Bash


vybe:
## Summary

When an agent is placed in read-only mode, it can still write arbitrary files using the native Claude Code `Write` and `Edit` tools. Read-only mode is enforced via hooks that intercept Bash commands, but `Write`, `Edit`, and `NotebookEdit` are separate tool calls that don't go through Bash — so they bypass the restriction entirely. An agent explicitly configured as read-only can freely create and modify files in its workspace.

## Component

Agent Runtime / Read-Only Mode / Hooks

## Priority

P1

## Observed Behavior

An agent with read-only mode enabled received a user request to save a memory. The agent called the `Write` tool directly and successfully created files at `/home/developer/.claude/projects/-home-developer/memory/`. No permission denial was recorded (`permission_denials: []`).

The execution init event shows the full `Write`, `Edit`, and `NotebookEdit` tools present in the tool list, with no hook blocking them.

## Root Cause

Read-only mode is implemented via hooks that scan/deny `Bash` tool calls. These hooks fire on `PreToolUse` (or equivalent) for Bash only. The native file tools — `Write`, `Edit`, `NotebookEdit` — execute directly against the filesystem without going through a shell, so they are never intercepted by the bash-scanner hooks.

Additionally, agents run with `permissionMode: bypassPermissions`, which disables the permission prompt system entirely. Even if write tools would normally prompt for approval, they proceed without pause.

## Reproduction Steps

1. Create an agent and enable read-only mode via the UI or API (`PUT /api/agents/{name}/read-only`)
2. Send the agent a task that asks it to write or edit a file (e.g. "save a note about me")
3. Observe the execution transcript — the agent calls `Write` and the file is created successfully
4. Check `permission_denials` in the execution result — it will be empty

## Suggested Fix

Add `PreToolUse` hooks for `Write`, `Edit`, and `NotebookEdit` (and any other file-mutating tools) in the read-only mode hook injection logic, mirroring the existing Bash hook. When read-only mode is enabled, these hooks should deny the tool call and return an error to the agent explaining the restriction.

Alternatively, the hook could be a single wildcard `PreToolUse` hook that checks the tool name against a denylist of write-capable tools: `Write`, `Edit`, `NotebookEdit`, plus any Bash command that modifies files.

```python
# Conceptual read-only PreToolUse hook logic
WRITE_TOOLS = {"Write", "Edit", "NotebookEdit"}

[...]

PR #2299, opened 2026-08-19
fix(security): move guardrail hook registration out of the agent's write reach (ent#345)

dolho:
## Summary

The guardrail hooks were always root-owned in `/opt/trinity/hooks/`. Their **registration** was not:

```dockerfile
COPY --chown=developer:developer ./hooks/claude-settings.json /home/developer/.claude/settings.json
RUN chmod 0644 /home/developer/.claude/settings.json
```

Owned by `developer`, mode `0644`, **inside the git-synced working tree** (HOME is the repo root), read live by Claude Code. So the guarded agent had write access to the file that decides whether it is guarded — one `Edit` removes the `PreToolUse`/`PostToolUse` entries and the bash deny-list, file guard, read-only guard and credential output-scanner stop running for every subsequent tool call, with no privilege escalation. Observed in the wild: an agent edited this file autonomously and committed it through git sync during the #2036 investigation. Being synced also made it settable from *outside*, by a push to the agent's repo.

§28.1's design principle says *"agents cannot opt out"*. That was not true of the mechanism enforcing it.

## The fix

Registration moves to **`/etc/claude-code/managed-settings.json`** — Claude Code's admin-controlled managed-settings path, which takes precedence over user and project settings and sits outside the synced tree, closing the self-edit and inbound-git vectors together.

**Verified, not assumed:** the CLI bundle (2.1.235) resolves `/etc/claude-code` with `managed-settings.json` and a `managed-settings.d` drop-in dir, and its own help text says such settings are *"Honored only from admin-controlled managed settings … ignored in user, project"* scopes.

Shipped `root:root 0444` inside a **root-owned `0755` directory**. The directory mode matters as much as the file's — a writable dir lets the file be replaced outright, or shadowed via a `managed-settings.d` drop-in. No platform-owned `settings.json` is shipped into `~/.claude` any more, so there is nothing there to edit away; the directory stays the agent's, since Claude Code writes its own state there.

## The new failure mode is the dangerous one, and it is handled

If the registration is missing or writable, Claude Code runs **no hooks — silently**, with nothing in the log. Trading "the agent can disable its guardrails" for "the guardrails quietly never ran" would be a worse bug, so `startup.sh` asserts both properties every boot:

```
GUARDRAILS: registration verified — root-owned, read-only, outside the synced tree
GUARDRAILS: ERROR — … is MISSING; Claude Code will run NO guardrail hooks (ent#345). Rebuild the base image.
GUARDRAILS: ERROR — … is WRITABLE by the agent user; the registration is not a boundary (ent#345).
```

All three branches were **executed**, not linted. It reports and continues rather than refusing to boot: a registration fault must not become a fleet outage. `-w` runs as `developer`, so the check asks exactly the right question. An operator-visible signal on `/health` (the #1439 `clone_status` pattern) is the tracked follow-up — deliberately not half-built here.

## Interaction checked before moving anything

Read-only mode *looked* like a second registration (`read_only.py` writes `.claude/settings.local.json`), which a managed `hooks` block could have clobbered — silently disabling read-only enforcement. Since **#887** it writes only a **config** file that the baked `read-only-guard.py` reads; `_remove_legacy_settings_hook` merely cleans up pre-#887 leftovers. So there is exactly one live registration, and this move strands nothing.

Also corrected: `READ_ONLY_MODE.md`'s "Hook Bypass" mitigation, which still described deleting `settings.local.json` as a live bypass.

## Note on the deny lists

Both paths stay in the three write-deny lists (`_FILE_WRITE_DENY_PATTERNS` / `guardrails-baseline.json::path_deny` / `EDIT_PROTECTED_PATHS`) as defence in depth, not as the primary control. Worth knowing: **`bash-guardrail.py` does not consult `path_deny`**, so before this change the Bash route to the registration (`sed -i`, `python -c`) was open even though the Edit route was denied.

## Testing

`tests/unit/test_ent345_guardrail_registration.py` — 9 tests pinning the shipped artifact, since CI does not build the base image:

- the managed COPY exists; no registration lands in the agent's HOME under **either** filename (a revert can't return quietly under the old name)
- root ownership + both modes (file `0444`, dir `0755`)
- the image still ends as the non-root user (Invariant #17 — the block runs as root)
- every registered hook command exists in `hooks/` **and** is COPYed into the image (a registration pointing at a missing script is a hook that fails on every call)
- all four guardrails are still registered
- the startup assertion's shape, and that it never `exit`s
- both deny-list entries survive

```
$ pytest tests/unit/test_ent345_guardrail_registration.py -q          # 9 passed
$ pytest <guardrail-adjacent: guardrails, read_only_guard, files_protected_paths,
          credential_inject_allowlist, 2070_trinity_authored_paths> -q  # 201 passed
```

**Honest gap:** the one property CI cannot check is that the CLI *in the built image* honours the managed path — I verified the CLI on this host, not in the image, because building the base image isn't something CI does. The startup assertion is the runtime backstop for exactly that, and a `verify-local --skip-agent=false` run (which boots a real agent) would confirm it end to end.

Full backend unit suite was still running when this was opened; will report.

Related to ent#345


vybe:
Hold from /validate-pr — the `src/backend/enterprise` gitlink in this PR moves `1dd28bc` → `3b41eab`, which is the head of **unmerged** enterprise PR Abilityai/trinity-enterprise#427 (the ent#364 asks *design doc* commit — unrelated to this guardrail fix, and that PR is currently red on all four pytest seeds). Looks like the submodule was checked out at the ent#364 branch when this was committed and `git add -A` picked up the pointer.

Two asks before merge:
- [ ] Reset the gitlink to the enterprise `main` tip (`git -C src/backend/enterprise checkout 1dd28bce && git add src/backend/enterprise`) — or, if the bump is intentional, land ent#427 first so the pinned SHA is reachable from `main`.
- [ ] Add a qualified closing ref to the body (`Fixes Abilityai/trinity-enterprise#345`) for traceability — cross-tracker, so the status bump on ent#345 stays manual either way.

Everything else validates clean: head pytest green across seeds (the base seed-67890 failure is a dev-baseline flake; regression diff passed), security scans clean, docs + tests present. Happy to approve as soon as the pointer is sorted.

dolho:
## CI unblocked — merged `dev` in, and dropped a bad submodule pin

CI was never actually failing here. The blocker was `mergeable: CONFLICTING` (17 commits behind `dev`), with the only real check anomaly being one *cancelled* `pytest (base, seed 67890)` shard — a runner timeout, which `regression diff` tolerates in degraded mode.

**Now: 21 success / 4 skipped / 0 failures, `regression diff` green, `MERGEABLE`.**

### The conflict was worth reading, not just resolving

The single conflicted path was the `src/backend/enterprise` submodule pointer. Resolving it surfaced a genuine defect in the PR rather than a mechanical merge artifact:

| pointer | commit | where it lives |
|---|---|---|
| this branch | `3b41eab` — *docs(workspace): design of record for agent-initiated asks (ent#364)* | **only** `origin/docs/ent364-workspace-asks-design` (unmerged) |
| `dev` | `a419812` — *fix(migrations): 0007/0008 no longer import the deleted client_portal (ent#431)* | `main` |

So `7e7d2ac6` had swept the submodule pointer onto an **unmerged feature branch of an unrelated issue**. Neither pointer contains the other, so this was never going to auto-resolve.

Resolved to dev's `a419812`, on two checks: there is **no enterprise-side ent#345 work at all** (grepped the submodule's full history), and the rest of this PR is OSS-only. So the bump was incidental.

Net effect: `src/backend/enterprise` no longer appears in this PR's diff — it is 10 OSS files, exactly what a guardrail-registration change should touch.

Merge commit rather than rebase, deliberately: no history rewrite, no force-push on a branch that already has review history, and it squashes away on merge to `dev` anyway.

Sanity-checked the PR's own suites after the merge — `test_ent345_guardrail_registration.py` + `test_2036_claude_settings_leak.py`, 22 passed.

Nothing about the substance of the change was touched.

vybe (approved):
Validated via /validate-pr: security scans clean, docs updated (architecture.md + requirements/security.md + READ_ONLY_MODE.md), regression tests present (test_ent345_guardrail_registration.py, test_2036_claude_settings_leak.py), all CI green including prod-image-smoke and verify-non-root. Base-image change — fleet needs a base-image rebuild to pick up the managed-settings registration.

Issue #1142, opened 2026-06-11
Add operator_queue retention sweep to cleanup_service


vybe:
## Summary

`operator_queue` has no retention sweep — it is absent from the #772 cleanup sweeps, so terminal rows (acknowledged/cancelled/expired) accumulate forever. #1017 added a manual **Clear All** (`POST /api/operator-queue/clear-resolved`), but unbounded growth should be handled automatically like the other tables.

## Proposal

Add an `operator_queue_retention_days` sweep to `cleanup_service.py` following the #772 pattern (configurable, default e.g. 90, `0` = disabled, 5000 rows/cycle cap):
- DELETE terminal rows (`acknowledged`, `cancelled`, `expired`) past the window.
- `responded` rows only past a generous age (the 5s write-back loop re-scans `responded` — young rows must survive until delivered; agents stopped at clear time pick the response up on restart).

## Context

Identified during #1017 plan review as the "10x" complement to the manual button. The `clear_resolved_items` DB primitive from #1017 (`db/operator_queue.py`) can be reused/extended.

## Acceptance Criteria
- [ ] Sweep wired into the cleanup service cycle with config knob + docs (architecture.md background-services table)
- [ ] Terminal statuses purged past window; `responded` rows protected by an age guard
- [ ] Capped per cycle; `0` disables

PR #2300, opened 2026-08-19
feat(workspace): agent-initiated asks — addressed column, validation, and three renderings (ent#364)

dolho:
## Summary

The OSS half of ent#364: an agent raises a question, an approval request or an update **addressed to a specific workspace user**; it renders in three places and is cleared everywhere when answered.

Gated surface half: `Abilityai/trinity-enterprise#427` (which also carries the design of record). **Merge order:** this PR → submodule pin → enterprise PR, since the enterprise module reads this column.

## One row, not three

An ask **is** an `operator_queue` row with an addressee — `addressed_to_email`, nullable, NULL meaning "the operator answers" so every existing row keeps its meaning. Both migration tracks (invariant #3). No new table: ent#164 builds a hard human-approval gate on these same rows, and two approval systems is the failure mode.

## Why the addressee is a column, not a key in `context`

It decides **who may answer** an ask and **whose sidebar it appears in** — an authorization decision. `context` is free-form agent-authored JSON whose hygiene clamp bounds size and type only, so an addressee living there is an agent picking its own audience.

So it is validated at that same #1632 ingestion boundary against `agent_on_roster(agent, email, include_owned=False)`: **an agent may address only someone already on its roster.** `include_owned=False` is the rule, not a default — flipping it on for a portal principal would hand a client agents nobody shared with them (see that function's own docstring).

Fails closed throughout — non-string, address-shaped-nothing, over-length, off-roster, or a roster read that raises all resolve to NULL, i.e. an ordinary operator ask. Malformed values are refused **before** the lookup so garbage can't amplify into a DB read per item. A `context.addressed_to_email` is never honoured, pinned by a test, because that is exactly the hole the column exists to close.

## Three renderings, one list

`PortalAsks.vue` renders on the agent page and inline above the chat composer; the sidebar shows `store.askCount`. All three come from one `fetchAsks()` — which is what makes *"answering anywhere clears it everywhere"* true with no sync step. Answering removes the row from local state rather than patching it, so a stale `pending` copy can't offer to answer the same ask twice.

An **expired** ask renders as expired and is excluded from the count: #1142 deletes terminal rows, and an ask that simply vanishes reads as "answered" to the person who didn't answer it — but it isn't something to nag about either.

**Poll, not a socket.** The Workspace has no WebSocket (`operator_queue_new` goes out on the platform `/ws`, which a portal client isn't on) and already polls elsewhere. 20s, visibility-aware, torn down with the view.

## Honest limits

- **AC #5 is only half-deliverable today.** Answering writes back to the agent's queue file through the OSS respond path, which the agent processes at its **next turn** — so scheduled agents act on answers now, and an agent with no schedule does not until **ent#329** lands (tracked as ent#430, `status-blocked`). That's the prerequisite ent#364 names; nothing here pretends otherwise.
- **Raise-time chat attachment** (AC #4) reads `context.workspace_session_id` when the platform has written one; the writer is ent#429's slice. An ask without it still renders in the sidebar and on the agent page, so it is never homeless — just not yet threaded.
- These two new calls use the bare `axios` export like the other 29 in that store. #2261 is moving the whole file onto a dedicated instance; half a file on each transport would be worse than either, and that PR owns the migration.

## Testing

```
$ pytest tests/unit/test_ent364_addressed_asks.py -q                       # 16 passed
$ pytest <operator-queue suites: 1632, 1631, 1525, 1426, 1142> -q          # 62 passed
$ npx vitest run tests/unit/workspaceAsks.spec.js                          # 9 passed
$ npx vitest run                                                           # 937 passed
$ python3 scripts/ci/check_alembic_heads.py src/backend/migrations/versions # 1 head
$ npx vite build                                                           # ✓
```

Not covered by the unit suite (node-only, no DOM): the three template branches. Their conditions are computeds over tested store state and each carries a `data-testid`.

Related to ent#364


vybe (approved):
Validated via /validate-pr: dual-track migration present (migrations.py + Alembic 0039, pg-migrations + schema-parity green), all pytest seeds + e2e + build green after the dev merge-resolution, security scans clean, architecture.md updated, tests on both layers (test_ent364_addressed_asks.py + workspaceAsks.spec.js — 48 specs pass locally post-portalHttp migration). Conflict with #2298 resolved on-branch per the fetchAsks comment's own plan; OPERATOR_QUEUE_EMAIL_MAX wired into compose + .env.example per the #1632 sibling convention. Approving.

PR #2307, opened 2026-08-19
fix(base-image): feature-detect `--yes` on `claude plugin install` (#2305)

### LINKED ISSUE #2305: bug: plugin install always withheld — `claude plugin install --yes` uses a flag the CLI doesn't have (#1704/ent#411 delivery broken)
## Summary

Every plugin install performed by the agent base image is withheld with `error: unknown option '--yes'`: the invocation `claude plugin install <ref> --yes` passes a flag that Claude Code CLI **2.1.227** (the version in the current base image) does not support. This breaks **both** delivery paths of ent#411 (#2284) — the build-time pre-install bakes only the marketplace into the image, never the plugin, and the #1704 boot-hook reconciler withholds the install on every container start — and it equally breaks #1704's original self-healing for any *declared* plugin that is missing from the volume. Observed fleet-wide on a production instance after updating to `70e0587e` and rebuilding all agents: 100% of agents report the plugin withheld, `claude plugin list` shows "No plugins installed", and `/trinity:onboard` is unavailable everywhere. The "never fatal" design worked as intended (boot continues, reason recorded), which is also why nothing errors loudly.

## Component

Agent Runtime / base image (`plugins_reinstall.py` + Dockerfile pre-install)

## Priority

P1 — the feature is entirely non-functional with no workaround; agents otherwise unaffected.

## Error

Recorded per agent in `~/.trinity/plugins-state.json`:

```json
{
  "added_marketplaces": ["abilityai"],
  "installed": [],
  "platform_plugins": ["trinity@abilityai"],
  "status": "platform_defaults_only",
  "withheld": { "plugin:trinity@abilityai": "error: unknown option '--yes'" }
}
```

Boot log:

```
[plugins-reinstall] withheld plugin:trinity@abilityai: error: unknown option '--yes'
[plugins-reinstall] declared 1 marketplace(s) / 1 plugin(s); added 1 marketplace(s), installed 0 plugin(s), skipped 0 present, withheld 1
```

Compatibility check I-006 surfaces it correctly (`status: fail — "trinity@abilityai could not be installed"`), so the observability half of ent#411 works.

## Location

- **File**: `docker/base-image/agent_server/plugins_reinstall.py`, line ~480: `["claude", "plugin", "install", ref, "--yes"]` (introduced by #1704 / `8d5cea44`, PR #2225)
- **File**: `docker/base-image/Dockerfile`, line ~167: `claude plugin install trinity@abilityai --yes` (introduced by ent#411 / PR #2284)

## Root Cause

CLI flag drift: `claude plugin install` in Claude Code 2.1.227 accepts only `--config` and `--scope` (per `claude plugin install --help` inside the image) — there is no `--yes`. The unit tests mock the CLI, so the invalid flag was never exercised against the real binary. Two aggravating details:

1. The Dockerfile's fallback message mislabels the failure: `|| echo "withheld trinity@abilityai: marketplace unreachable at build time"` fires for *any* non-zero exit, so a flag error is reported as network unreachability.
2. Because #1704 introduced the flag, the boot-hook's self-healing install path has been broken since #2225 for any declared-but-missing plugin — ent#411 just made every agent hit it on every boot.

## Reproduction Steps

1. Build the current agent base image (`docker/base-image/`) — build log prints the mislabeled "marketplace unreachable" line even online.
2. Start any agent container; check `~/.trinity/plugins-state.json` → `withheld: error: unknown option '--yes'`.
3. Inside the container: `claude plugin install trinity@abilityai --yes` → `error: unknown option '--yes'`; `claude plugin list` → "No plugins installed".
4. `claude plugin install --help` → no `--yes` option exists.

## Suggested Fix

Remove `--yes` from both call sites, verifying against the real CLI that the install is non-interactive under the reconciler's existing conditions (`stdin=DEVNULL`, hard timeout) — and if a confirmation path exists in some CLI versions, feature-detect via `claude plugin install --help` output rather than passing an unconditional flag. Consider:

```python
# plugins_reinstall.py
rc, _out, err = _run(
    ["claude", "plugin", "install", ref],   # drop --yes; stdin is already DEVNULL
    timeout=_INSTALL_TIMEOUT,
)
```

```dockerfile
# Dockerfile — also split the fallback so a flag/exit error is not reported as network unreachability
&& claude plugin install trinity@abilityai \
```

Additionally worth pinning by a test that runs the **real** CLI's `--help` (the base image build has the binary available): assert every flag the module passes is advertised, so the next CLI bump cannot silently re-break delivery. A CI grep for `withheld` in the base-image build output would catch this class at build time despite the never-fatal design.

## Companion finding (separate root cause, same validation pass)

On the platform-seeded **system agent** (`trinity-system`), even the marketplace add fails: `✘ Failed to add marketplace: EACCES: permission denied, mkdir '/home/developer/.claude/plugins'` → plugin then withheld as `marketplace 'abilityai' unavailable`. The system agent's `~/.claude` volume apparently carries ownership that blocks the reconciler from creating `plugins/`. Regular agents do not exhibit this. May deserve its own issue if not trivially fixed alongside.

## Environment

- Trinity: `70e0587e` (dev, post-v0.9.0), base image built from it
- Claude Code CLI in image: 2.1.227
- OS: Ubuntu (GCP), Docker prod compose

## Related

- ent#411 / PR #2284 (pre-install), #1704 / PR #2225 (boot-hook reconciler, introduced the flag)
- `docker/base-image/agent_server/plugins_reinstall.py`, `docker/base-image/Dockerfile`
- Compatibility check I-006 (reports the withheld reason correctly)

trinity-ability:
## Summary
- Every plugin install was withheld fleet-wide with `error: unknown option '--yes'` — CLI 2.1.227 (the image version) rejects the flag, breaking both ent#411 delivery paths (build-time pre-install + #1704 boot-hook reconciler) and #1704's original self-healing.
- The flag drifted **both ways**: 2.1.235+ advertises `-y, --yes` and *requires* it for a plugin installed via a marketplace-declared command when stdin is not a TTY. So neither passing nor dropping it unconditionally is safe — both call sites now **feature-detect** it via the CLI's own `claude plugin install --help` (the fix shape the issue itself suggests).
- The Dockerfile fallback no longer mislabels every failure as "marketplace unreachable" — it names which step failed (the mislabel that hid this bug at build time).

## Changes
- `docker/base-image/agent_server/plugins_reinstall.py` — `_install_supports_yes()` probe; detected lazily, **once per reconcile, only when an install is actually needed** (the zero-subprocess contract for a satisfied boot holds); fail-closed to a bare install (works for normal plugins on every CLI version; a prompting path still can't hang — `stdin=DEVNULL` + hard timeout ⇒ withheld with a reason).
- `docker/base-image/Dockerfile` — same detection in shell for the ent#411 pre-install; fallback split into per-step messages.
- `tests/unit/test_1704_plugins_reinstall.py` — fake CLI gains a `help_yes` knob + help-probe dispatch; the old unconditional-`--yes` argv pin is replaced by both detection branches (`…install <ref>` on a 2.1.227-shaped CLI, `…install <ref> --yes` on 2.1.235+); probe-at-most-once and zero-probe-when-satisfied pinned; **plus a real-CLI parity test** (skips when no `claude` on PATH) so the next CLI flag drift fails on a dev machine instead of fleet-wide at boot — the mocked suite structurally couldn't catch this class.
- `tests/unit/test_ent411_platform_plugins.py` — fake dispatches the help probe; `installs()` excludes it.
- Docs: `architecture.md`, `feature-flows/agent-plugin-manifest.md`, `requirements/content-files.md` — the four spots that pinned `install --yes` now describe the feature-detected behavior.

## Notes
- The issue's optional "CI grep for `withheld` in base-image build output" has no home — no workflow builds the base image today; the split fallback messages make the failure legible in any manual/scripted build log instead.
- The companion `trinity-system` EACCES finding (`mkdir '/home/developer/.claude/plugins'` permission denied) is a **separate root cause** — filed as its own issue.
- Requires a base-image rebuild to reach agents (same release-note ordering caveat as #1704/ent#411).

## Test Plan
- [x] `pytest tests/unit/test_1704_plugins_reinstall.py tests/unit/test_ent411_platform_plugins.py tests/unit/test_1704_source_validator_parity.py` — 78 passed
- [x] Real-CLI parity test exercised against local `claude` 2.1.235 (detection returns True, matching the advertised flag)
- [x] Dockerfile RUN block reconstructed and `sh -n` syntax-checked; detect snippet dry-run prints `yes_flag='--yes'` on 2.1.235 (and greps nothing on a 2.1.227-shaped help)
- [ ] Manual: rebuild base image, boot an agent, check `~/.trinity/plugins-state.json` → `installed: ["trinity@abilityai"]` / `skipped`, and I-006 passes

Fixes #2305


vybe (approved):
Validated via /validate-pr: all pytest seeds green (base+head), gitleaks/CodeQL/schema-parity/verify-non-root green, security scans clean, base-image change justified and documented (architecture.md + feature-flow + requirements), regression tests name #2305/#1704/ent#411. Approving.

PR #2310, opened 2026-08-19
feat(skills): assign and unassign from the Library (ent#386)

dolho:
Closes the write half deferred from ent#384. The Library could report that a skill had no holders and was the one page unable to act on it — the next action was a link out to an agent's Skills tab.

## What it does

Each skill block now carries an inline assign dropdown, and every holder chip the caller may modify carries an unassign control. Zero-holder skills get the same control instead of the link-out.

## The decision the issue deferred

**Assign targets are a different set from holders**, and that difference is the whole feature:

| | scope | source |
|---|---|---|
| holder chips | owned ∪ shared | `GET /api/skills/assignments` (ent#384) |
| assign targets | **owner-or-admin** | `assignable_agents`, new on the same response |

So a shared agent appears as a holder, is never offered as a target, and shows no unassign × — rather than an affordance that exists only to 404.

The issue floated deriving this client-side from `GET /api/agents`'s `is_owner`. **That field does not exist on the response**, and deriving it in the browser would put a second copy of an authorization predicate one refactor away from disagreeing with the gate the write route enforces. It is computed server-side, on the read the Library already performs — no extra round-trip, no per-block N+1.

## Deliberate choices

- **No skill-keyed writer.** Assign/unassign call the existing `POST`/`DELETE /api/agents/{name}/skills/{skill}`, where the owner gate already lives (ent#182: one skill model).
- **The two lists share their exclusions by construction** — `get_assignable_agents` sits next to `get_all_skill_assignments` and filters identically. Drift means the dropdown offers an agent the chips can never show, so assigning looks like a no-op.
- **Ghosts excluded**, matching ent#384: hard-discarded at budget, so the row stops meaning anything in minutes.
- **Patch, don't refetch** — keeps scroll position and tab state (AC 3), and avoids re-running a fleet-wide O(agents × skills) read for one row.
- **Errors render inline**, carrying the server's own wording (`Skill 'x' not found in library`, `Agent not found`) rather than invented text (AC 4).

## Acceptance criteria

- [x] Dropdown lists only modifiable agents, excluding current holders
- [x] Unassign from the same block
- [x] No page reload; scroll and tab state preserved
- [x] Named validation error, never a generic 500 or silent no-op
- [x] Zero-holder skill's next action is an inline assign

## Verification

- 12 backend tests — access asymmetry, admin tri-state (`None` ≠ empty set), ghost + soft-delete exclusion, `response_model` allow-list, and a guard that no skill-keyed assignment writer exists
- 19 store tests — target derivation, in-place patching, key removal at zero holders, server-reason propagation, path encoding
- 408 skill-related backend tests pass; frontend suite 961 passed; `vite build` clean

Delivery stays out of scope (ent#385) — this writes the `agent_skills` row and claims nothing about injection.

Related to Abilityai/trinity-enterprise#386


vybe (approved):
Validated via /validate-pr — see the validation summary on this run.

PR #2312, opened 2026-08-19
feat(operator-queue): an operator answer can wake a parked agent (ent#329)

dolho:
An operator's answer reaches the agent's queue file in ~5s, but is only **processed** at the agent's next turn. A scheduled agent has one. An agent started by a one-shot webhook or chat task does not — so an approved action silently never runs until somebody re-triggers it by hand. trinity#1402 documented the gap and told agents to work around it in prose; this is the platform-side fix.

It is also the hard prerequisite for ent#430 / ent#364: every Workspace ask no-ops without it, which is why that flag ships OFF.

## Shape

**Opt-in, per agent, default OFF.** A dispatch spends money, so it is never unconditional — a respond-storm must not fan out executions. Deliberately *not* a per-request `resume: true` the agent sets: that lets an agent decide that answering costs the answerer money, which is unacceptable once the answerer is an external Workspace client (ent#430 AC #3). Confirmed with the maintainer before building.

**One dispatch surface.** Everything runs through `execute_task(triggered_by="operator_response")`, so capacity admission, the dispatch breaker, cost accounting and the terminal appliers are the ones the platform already uses. No queue- or workspace-specific execution path.

**Hangs off the CAS win.** The respond endpoint raises 409 when the item left `pending` under the caller; the dispatch sits after that raise, so an answer that was never recorded can never spend (the #1083 rule that side effects follow the CAS result, not the attempt).

**Idempotent** over (item id, answer) per Invariant #18. **Never silent**: a failed dispatch is a FAILED execution row plus an `operator_resume_dispatch` audit entry carrying ids and status — never the answer text, since that row is broadly readable and the answer is whatever a client typed.

## Trigger registration

`operator_response` is added to all three sets a trigger must appear in: `_VALID_TRIGGERS` (Executions filter), `_TRIGGER_BUCKETS` (its own **Operator queue** bucket — unmapped triggers silently become "Other", which then quietly means "resumes"), and `_AUTONOMOUS_TRIGGERS`.

#2048's tripwire fired and demanded the pull-mode classification, correctly. It is **stranded**: dispatched by a direct backend call, never by `POST /task`, so `_derive_task_trigger` can never emit it — same shape as `schedule` and `reminder`.

## Verified on real PostgreSQL

The full Alembic chain applies on a fresh PG database, the column lands with default `0`, and a re-run is a no-op. The revision is **guarded**, and that guard is load-bearing: `0001_baseline` emits the same DDL as `init_schema_postgres`, which this change edits — so on a fresh PG the column exists *before* the revision runs, and an unguarded `add_column` aborts boot with `DuplicateColumn` (exactly the failure #2300 hit before its guard).

## ⚠️ Merge ordering

`dev` does not yet carry #2300's `0039_operator_queue_addressed_to`, so both PRs currently claim `0039`. **Whichever merges second must be re-pointed** (renamed to `0040`, `down_revision` chained onto the other) or the two become sibling heads and boot fails. I'll do that rather than leave it to CI.

## Verification

- 21 new tests: default-OFF for pre-migration rows, soft-deleted agents unflippable, one-execution-per-answer, replay does not double-dispatch, answer framed as data, failed dispatch audited and non-raising, unreadable flag fails safe, audit carries no answer text, CAS ordering, owner-only knob, and the three trigger-set registrations
- 463 tests pass across the related capacity/1083/idempotency/operator-queue suites
- Frontend 927 passed; `vite build` clean

**One honest note on the tests**: three of them initially passed alone and failed in a full run. `sys.modules` stubbing is not enough once another test has really imported `services.x` — the parent package attribute shadows it, the real service runs, raises, and the module's own except-clause swallows it. The harness now patches both, and the previously-failing slice passes.

Related to Abilityai/trinity-enterprise#329


dolho:
## CI unblocked — merged `dev` in; one conflict was cosmetic, the other was not

CI was not failing here either. The blocker was `mergeable: CONFLICTING` (12 commits behind `dev`); the only check anomaly was one *cancelled* `pytest (base, seed 99999)` shard (runner timeout).

**Now: 23 success / 3 skipped / 0 failures, `regression diff` green, `MERGEABLE`.**

### 1. `db/migrations.py` — routine

Both sides appended to the `MIGRATIONS` list. Kept **both**, dev's two entries first and this branch's last, so the runner applies them in the order they landed. 110 entries, no duplicates.

### 2. Alembic multi-head — this one was silently dangerous

`dev` added `0039_operator_queue_addressed_to` and `0040_rl_events_failure_kind` while this branch carried its own `0039_agent_ownership_operator_resume` off the same `0038` parent. Git merges that without a murmur — two files, each individually valid, no textual conflict. The defect exists only in the *relationship*.

Per Architectural Invariant #3, that is the bad case: `alembic upgrade head` is **singular** and resolves its target before applying anything, so with two heads it applies **zero** revisions. Not just the forking one — **every revision merged since the fork stops arriving**, while CI stays green because each file parses.

Renamed to `0041_agent_ownership_operator_resume` and re-parented onto `0040_rl_events_failure_kind`. `scripts/ci/check_alembic_heads.py` now reports one head; `pg-migrations` and `schema-parity` both pass, and dual-track parity holds (SQLite entry + Alembic revision).

### ⚠️ Heads-up: `0041` now collides with #2330

Abilityai/trinity#2330 also claims `0041` (`0041_secret_settings_encryption`, same `0040` parent). There is no ordering-free fix — a `down_revision` can only point at a revision that already exists — so **whichever of the two merges second must re-parent onto the other** (`0042_…`, `down_revision` = the first one's id).

That is a one-line change, and it fails loudly rather than silently: `check_alembic_heads` is wired **unconditionally** into `schema-parity`, so the second PR goes red on the multi-head before it can ship an inert chain. Flagging it so nobody debugs it cold.

Merge commit rather than rebase: no history rewrite, no force-push, and it squashes away on merge to `dev`.

Ran the PR's own tests plus the migration/schema/alembic guards after the merge — 209 passed, 3 skipped. Nothing about the substance of the change was touched.

vybe (approved):
Validated via /validate-pr: security scans clean, docs updated (architecture.md), tests present incl. the #2048 pull-pilot tripwire and test_ent329_operator_resume.py, all 23 checks green.

Alembic revision re-parented by the reviewer during merge: 0041_agent_ownership_operator_resume -> 0042, chained off 0041_secret_settings_encryption (#2330), since three PRs each claimed 0041 off 0040_rl_events_failure_kind. dev has strict:false, so a stale PR is not re-checked and the fork would not have been caught by check_alembic_heads at merge time. pg-migrations and schema-parity both pass on the re-parented chain. Cross-tracker ent#329 needs a manual status-in-dev bump.

PR #2315, opened 2026-08-19
chore(enterprise): bump submodule pin — client_portal migration fix (ent#431)

dolho:
Bumps `src/backend/enterprise` from `1dd28bce` → `a4198127`, picking up Abilityai/trinity-enterprise#432 (ent#431).

## Why this is needed on top of the enterprise merge

The deploy runs `git submodule update --init --recursive` — **no `--remote`** — so it checks out exactly the SHA `dev` pins. Merging the enterprise PR alone changes nothing for the instance.

## What it unblocks

`ENT_SUBMODULE_PAT` fixed the clone, and the deploy immediately failed one layer later:

```
ENTERPRISE: using HTTPS transport (ENT_SUBMODULE_PAT configured)
ENTERPRISE: initialized at 1dd28bc
Trinity Enterprise registration FAILED — continuing OSS-only:
  ModuleNotFoundError("No module named 'enterprise.backend.client_portal'")
```

ent#356 moved the workspace tables to the OSS track and deleted the package; revisions `0007`/`0008` kept importing it. On any database not already stamped past `0008` the enterprise chain raises, which aborts `register_enterprise` **entirely** — not just the portal. `enterprise_features` empty, every `requires_entitlement` route 403ing.

## Verification at this exact pin

Against a real PostgreSQL:

| check | result |
|---|---|
| enterprise chain | reaches head `0016_onboarding_state` |
| modules registered | **17** (`2fa, a2a, audit, compliance_export, cross_model_validation, ephemeral_agents, onboarding, permissions_matrix, portal_session_policy, retention, shared_sessions, siem, skill_runner, slack_per_agent_bots, sso, telemetry, user_management`) |
| legacy `0007`-stamped DB | backfill still folds pre-session messages — `sessions=1 orphaned=0` |

## What to watch after merge

Merging pushes `dev`, which triggers Deploy to Dev automatically. Green looks like three lines, the third of which has never yet appeared:

```
ENTERPRISE: using HTTPS transport (ENT_SUBMODULE_PAT configured)
ENTERPRISE: initialized at a419812
ENTERPRISE: modules registered
```

Then confirm on the instance itself — the deploy gate only checks registration:

```bash
curl -s https://<dev-host>/api/version | jq '.edition, .enterprise_features'
```

Note the dev DB's enterprise line has **never** run, so revisions `0009`–`0016` execute there for the first time. That whole chain was verified on a fresh PG, but a fresh PG is not identical to the dev DB; if something fails it will be one of those revisions and the traceback will name it.

Diff is the submodule pointer only.


dolho:
## /review Report

**Branch**: `chore/bump-enterprise-pin-431` → `dev` (merge-base `16caacb0`)
**Files Changed**: 1 (+1/−1) · **Scope**: CLEAN · **Critical**: 0

### Scope check

The diff is the pointer and nothing else:

```
-Subproject commit 1dd28bce20d089353991151d8dfdfa5d9f5847b8
+Subproject commit a4198127da107d0749b6cc098f427434d691ff2d
```

For a pin bump the diff isn't the interesting part — what the pointer *drags in* is:

```
$ git log --oneline 1dd28bce..a4198127
a419812 fix(migrations): 0007/0008 no longer import the deleted client_portal (ent#431) (#432)
```

One commit, and it touches only migrations + tests:

```
$ git diff --name-only 1dd28bce..a4198127 | grep -v '^tests/\|^backend/migrations/versions/'
(nothing outside migrations + tests)
```

No module registration, no feature ids, no `tables.py` — so the OSS side's contract with the submodule is unchanged.

### Informational findings

**[I1] No CI job in this repo exercises the new pin.**
`enterprise-tests.yml` lives in the enterprise repo, not here. The nearest job on this board — `backend boots without enterprise submodule` — validates the *OSS-only* path, i.e. the opposite configuration. A green board here says nothing about the enterprise code being pinned.

Mitigated for this bump, though not structurally. The incoming commit passed its own repo's CI:

```
alembic-single-head                          pass
pytest (seed 106 / 12345 / 67890 / 99999)    pass
```

and the pin was verified against a real PostgreSQL:

| check | result |
|---|---|
| enterprise chain | reaches head `0016_onboarding_state` |
| modules registered | **17** |
| legacy `0007`-stamped DB | backfill intact — `sessions=1 orphaned=0` |

This is inherent to the two-repo split rather than a defect in this PR, and the deploy's own #2068 gate is the backstop. Not filing it.

**[I2] The dev DB runs revisions `0009`–`0016` for the first time.**
Its enterprise line has never executed — the submodule hasn't mounted for the whole retained run history. That chain was verified on a *fresh* PG, which is not the dev DB (real OSS tables, real history). If it fails it will fail at a revision this PR does not touch, and the traceback will name it.

### Clean categories

- **SQL / data safety** — no SQL in this diff; the incoming migration change was reviewed on Abilityai/trinity-enterprise#432, where one critical finding (a guard that also swallowed `0008`'s backfill) was found and fixed before merge.
- **Auth boundaries, credentials, frontend, enums, performance** — N/A for a single gitlink change.
- **Docs staleness** — no endpoint, schema or flow described in `architecture.md` changes.
- **§4.14 incomplete-fix completeness** — the regression guard ships in the enterprise repo; re-grepped for other references to the deleted package across the submodule: none.

### Verdict

**Merge-ready.** Merging pushes `dev` and triggers Deploy to Dev automatically. Green looks like three lines, the third of which has not appeared in any run to date:

```
ENTERPRISE: using HTTPS transport (ENT_SUBMODULE_PAT configured)
ENTERPRISE: initialized at a419812
ENTERPRISE: modules registered
```

Then confirm on the instance itself, since the deploy gate only asserts registration:

```bash
curl -s https://<dev-host>/api/version | jq '.edition, .enterprise_features'
```

*CI at time of writing: 8 pass / 8 pending / 4 skipping, 0 failures.*

obasilakis (approved):
Approving. Verified the substance independently rather than just the pointer diff — this also unblocks an ops-side enablement that hit the identical failure today.

**Ancestry is clean.** `a4198127` is strictly ahead of every pin in play: +1 over `1dd28bce`, +9 over `2d64baa`, +5 over `65182c14`, zero behind in all three. No rewind, and the `2d64baa` security floor (ent#218/#220 rooms fixes, ent#375) is preserved.

**The skip has a real creator upstream.** 0007/0008's early return is only safe if something else still makes those tables — `src/backend/migrations/versions/0036_client_portal_oss.py` is present on `dev`, so this isn't dropping DDL on the floor.

**`find_spec` over `try/except ImportError` is the right call**, and for the reason the comment gives: a bare except would also swallow a genuine broken import *inside* a module that does exist, converting a real failure into a silent skip. That distinction matters more than usual here, given the failure mode being fixed.

**0008 keeping the backfill outside the guard is the subtle part, and it's right.** A database stamped at exactly `0007` carries pre-session messages with NULL `session_id`, and nothing in either chain folds them. An early return there would orphan client history silently — invisible until someone opened the Workspace and found the conversation gone. Scoping the backfill to `session_id IS NULL` makes it a no-op everywhere else, which is the correct trade.

**The test is a class-of-bug harness, not a two-file patch test.** `test_client_portal_migration_imports.py` scans every revision for imports of packages that don't exist and requires a `find_spec` guard, asserts the per-revision guard shape, asserts no bare-except swallowing, and asserts the backfill stays outside the guard. That prevents ent#431 recurring, not just this instance of it.

Two non-blocking notes:

1. **The load-bearing ordering assumption isn't asserted anywhere.** 0007's skip is correct *because* OSS `init_database()` runs to head before `register_enterprise()`. The tests check guard shape, not that ordering. If it ever changes, 0007 becomes a silent no-op and the tables simply never exist — the same silent-failure property that made ent#431 expensive to find. A single test pinning the call order, or an assert that the table exists when the guard skips, would close it cheaply. Fine as a follow-up.

2. **Merge-time watch item** (already called out in the description): the dev DB's enterprise line has never run, so `0009`–`0016` execute there for the first time. Verified on a fresh PG, and a fresh PG is not the dev DB. Worth watching that deploy rather than assuming it.

Ops-side corroboration: enabling enterprise on `tapestry` today failed exactly this way at pin `2d64baa` — `ModuleNotFoundError: No module named 'enterprise.backend.client_portal'` out of `register_enterprise()`, entitlement empty, instance reverted to OSS. Existing entitled instances are unaffected (stamped past `0008`, so those revisions never re-import), and `ceoci` enabled fine the same day on the pre-ent#356 pin `e898cef8`. tapestry stays OSS until this lands and the pin reaches the instance.

PR #2316, opened 2026-08-19
feat: subscription usage observability + live headroom — 5h/7d windows, per-agent breakdown, pressure badges (#471)

### LINKED ISSUE #471: feat: Subscription usage observability — 5h/7d usage dashboard, per-agent breakdown, and rate-limit event history

trinity-ability:
## Summary

- **Data-layer fixes first**: failure events recorded BEFORE the auto-switch enabled gate (opted-out operators previously had permanently-zero observability); `failure_kind` persisted on `subscription_rate_limit_events` (the table conflated auth failures with 429s — dual-track migration, SQLite + Alembic `0040` (re-parented onto dev's `0039_operator_queue_addressed_to` at merge)); **SUB-004 dedup** — a modern turn writes both an execution row and (for `/chat` + persisted `/task`) a cost-bearing chat message, so executions are now the sole cost/context/turn source (verified live: one $0.0687 turn previously reported as $0.137).
- **Live headroom** (`services/subscription_headroom_service.py`): actual 5h/7d utilization % + reset times from the `anthropic-ratelimit-unified-*` headers of a `max_tokens=1` micro-ping probe. Established empirically with a real stored setup token: `GET /api/oauth/usage` is 403-scope-dead for `sk-ant-oat01-` tokens (the closed-PR #2170 mechanism); the header channel works. Click-to-refresh + default-ON ambient refresh behind a Settings toggle (15-min floor, demand-driven, single-flighted, **fail-closed to observed-only when Redis is unreachable**, stale-while-revalidate on the dashboard path). Every reading carries `source: anthropic|observed` + snapshot age; the DB-derived arm is always populated (the #2170 inversion).
- **Surfaces**: Settings subscriptions section extracted to `SubscriptionsPanel.vue` (partial #717/#1030 paydown) with usage windows, failure-event warnings, live gauge + Refresh, lazy per-agent breakdown (cost-ranked — model-weighted by construction, resolving the 2026-07-28 model-mix research item); `GET /api/agents/subscription-pressure` batch endpoint (pure-DB visible set via the shared ent#384 helper, `AgentAuthStatus` vocabulary, explicit `response_model`) feeding AgentTile chips + AgentListPanel badges; Tier 0 `≈ API-equivalent` cost relabel on AgentHeader + fleet surfaces, with the Step-0 zeros-fork (honest dash when cost is structurally unreported).

## Rulings recorded at the build gate (2026-08-19, on-issue)

OSS-core by explicit decision (visibility free, governance paid — ent#166). Tier 4 auto-assign cut (SUB-003 covers the reactive per-agent case; bulk proactive migration is a real gap — filing separately). MCP tools + `~/.trinity/usage.json` stay in trinity-enterprise#351. Ambient probing default-ON by operator ruling with cost disclosure on the toggle.

## Test Plan

- [x] `tests/unit/test_471_subscription_usage_observability.py` — 26 tests: failure_kind counters + boundaries, usage extension + model defaults, dedup pins (cost counted once), breakdown ranking, batch map scoping, record-before-gate, headroom header parsing against the live spike payload, fail-closed ambient probing, stale-while-revalidate, route-order + admin-gate static pins
- [x] Blast-radius suites green (pingpong — one pin deliberately flipped to the new record-before-gate contract, concurrency, #2114 shadow guard, #792 retry, BOLA, models-centralized, #1310 auth wiring, #1560 key parity)
- [x] Frontend: 954 vitest (incl. 12 new for the shared pressure predicate) + production build clean; raw-color ratchet exit 0
- [x] **Live demo on local Docker**: registered a real subscription, assigned an agent, ran a turn — Settings panel shows `5h: 27% · 7d: 7%` with `actual (Anthropic)` source + reset times; pressure endpoint carries per-agent rows; three bugs caught live and fixed with regression pins (microsecond age parse, SUB-004 double-count, plus review-caught Redis fail-closed + batch-path blocking)
- [x] `/cso --diff`: zero findings at the daily gate (`docs/security-reports/cso-2026-08-19.md`)

## Release notes needed

Operators will see tiny platform-initiated messages (~1 token) in the Anthropic console when live headroom is enabled (default ON; Settings → Integrations toggle turns it off).

Fixes #471


vybe (approved):
Validated: all 4 required contexts green, full pytest matrix green on head and base, pg-migrations + schema-parity + prod-image-smoke pass. `Fixes #471` resolves same-repo so status auto-promotion fires on merge. Approving for merge.

vybe:
Merge was blocked by a **real** `regression diff` failure, not the cancelled base seed — flagging it here since it was invisible in the per-job conclusions.

All six `pytest` seed jobs reported **success**, but the head XMLs each carried `Fail: 1`; the gate is `regression diff`, which correctly flagged:

```
❌ New failures introduced by HEAD (1)
- [F] test_384_skill_assignments_read::test_access_set_is_db_derived_not_docker_derived
```

Deterministic on all three head orders (12345/67890/99999), green on base — a genuine new failure. It passed in isolation (12/12), which is what made it look like flake.

**Cause** — the guard added in 059b0abc asserted object identity:

```python
assert router_mod._visible_agent_names is _helpers.visible_agent_names
```

`router_mod` is the module object bound at *its* import, while the `from services.agent_service import helpers` inside the test resolves at call time. The suite contains tests that purge `sys.modules` (the reason CI carries a dedicated sys.modules-pollution lint); after a purge, `helpers` is re-imported from the same path under the same key, so `is` compares two distinct function objects. The CI output shows exactly that — two `visible_agent_names` at different addresses, both from `src/backend/services/agent_service/helpers.py`.

Reproduced deterministically: with a purge between import and assert, the identity assert is `False` while the origin assert is `True`.

**Fix** (f8b29203, test-only) — pin the *origin* (`__module__` + `__qualname__`) instead. Stable across re-import, and it still defeats what identity was there to defeat: a router-local `def _visible_agent_names` reads `routers.skills`. Verified by negative control — swapping the import for a local re-implementation fails with `got routers.skills._visible_agent_names`.

The product refactor itself is sound: one definition (`services/agent_service/helpers.py:206`), two consumers (`routers/skills.py`, `routers/agents.py`).

PR #2321, opened 2026-08-19
fix(pull): claim envelope reads the flat backlog_metadata keys the producer writes (#2317)

### LINKED ISSUE #2317: bug(pull): claim envelope reads task_overrides, a key no producer writes — every pulled turn drops the row's model, tools, turn cap, timeout and session
## Summary

The pull claim envelope reads per-task settings from a `task_overrides` key inside `backlog_metadata`. **No producer ever writes that key.** Every pulled turn therefore runs with the row's own model, tool allow-list, turn cap, timeout and session identity silently dropped, falling back to agent/global defaults.

The push path is unaffected — it reads the same fields, flat, and applies them correctly. This is a pull-path-only divergence, and it is silent: nothing logs, nothing fails, and the execution row records a normal `success`.

## Mechanism

`backlog_service.enqueue` (`src/backend/services/backlog_service.py:92`) persists the queued row's settings as **flat** keys:

```python
metadata = {
    "message": request.message,
    "model": request.model,
    "allowed_tools": request.allowed_tools,
    "system_prompt": request.system_prompt,
    "timeout_seconds": effective_timeout,
    "max_turns": request.max_turns,
    "save_to_session": request.save_to_session,
    "user_message": request.user_message,
    "create_new_session": request.create_new_session,
    "chat_session_id": request.chat_session_id,
    "resume_session_id": request.resume_session_id,
    "inject_result": request.inject_result,
    ...
}
```

The **push** drain reads them flat and reconstructs the request faithfully — `_spawn_drain` (`backlog_service.py:289`): `model=metadata.get("model")`, `allowed_tools=metadata.get("allowed_tools")`, `timeout_seconds=metadata.get("timeout_seconds")`, and so on.

The **pull** envelope does not. `_build_claim_response` (`src/backend/services/pull_coordination_service.py:224`):

```python
overrides: Dict[str, Any] = dict(meta.get("task_overrides") or {})
```

`meta` here is the parsed `backlog_metadata`. There is no `task_overrides` key in it, so `overrides` is always `{}`. #1633 then writes exactly one field into it (`system_prompt`, composed explicitly), which is why the platform prompt is the only setting that survives the pull path.

The same mismatch applies to session identity one line earlier (`pull_coordination_service.py:215`): the envelope reads `meta.get("session_id")`, while the producer writes `chat_session_id` / `resume_session_id`.

## Fields dropped on the pull path

`model`, `allowed_tools`, `max_turns`, `timeout_seconds`, `chat_session_id`, `resume_session_id`, `save_to_session`, `inject_result`, `user_message`, `create_new_session`.

Only `system_prompt` survives, via #1633's composer.

## Evidence (production, `eu2`)

Measured against the live PostgreSQL, `cornelius-oracle`, covering the 7 days before and 6 days after the #1766 pilot flip (`2026-08-12T12:44:59Z`):

- `backlog_metadata LIKE '%task_overrides%'` matches **0 of 580 rows**. No producer has ever written the key.
- Executions failing on their own configured timeout: **18 in the 7 pre-flip days, 0 after.** Those rows carried 60/90/120/300s budgets against a ~282s average successful turn. Under pull they run to `TRINITY_PULL_TURN_TIMEOUT` (default 900s) instead and mostly land `success`.

That second number is the visible symptom: it looks like a reliability improvement, and it is actually the row's timeout no longer being enforced.

## Impact

1. **Correctness.** A caller that asks for a specific model, a restricted tool allow-list, or a bounded turn count does not get it on the pull path. The tool allow-list case is the sharpest — a caller narrowing an agent's capability for one task is silently ignored.
2. **Session continuity.** `chat_session_id` / `resume_session_id` are dropped, so a pulled turn cannot resume the session the caller named.
3. **Measurement.** This makes M6 (outcome regression vs baseline) uninterpretable for any soak window — the failure class that disappears is the one the missing timeout stopped enforcing. Recorded as a scope limit on #1766's open window.

## Suggested fix

Have the pull envelope read the same flat keys the push drain reads, rather than a nested `task_overrides` object that nothing produces — and align `session_id` with `chat_session_id` / `resume_session_id`. Keeping #1633's composed `system_prompt` behaviour on top. A structural test asserting that the pull envelope and `_spawn_drain` consume the same key set from one `backlog_metadata` fixture would stop this recurring.

## Notes

- Present on `dev` tip at time of filing.
- Found during #1766 pre-flight; recorded in that issue's 2026-08-12 flip comment as a scope limit but never filed separately until now.
- Blocks widening the pull pilot; does **not** block closing #1766, whose remaining ACs are independent of the overrides.

obasilakis:
Closes #2317. Unblocks widening the #1766 pull soak (that issue's remaining ACs are independent; nothing here closes it).

## Mechanism

`backlog_service.enqueue` is the only producer of a queued row's `backlog_metadata`, and it writes every per-task setting as a **flat** key (`model`, `allowed_tools`, `system_prompt`, `timeout_seconds`, `max_turns`, `resume_session_id`, …).

The **push** drain reads them flat and rebuilds the request faithfully (`_spawn_drain`, `backlog_service.py`).

The **pull** claim envelope did not. `_build_claim_response` had:

```python
overrides: Dict[str, Any] = dict(meta.get("task_overrides") or {})
```

There is no `task_overrides` key in `backlog_metadata` — no producer has ever written one — so `overrides` was **always `{}`**. #1629 then wrote exactly one field into it (`system_prompt`, composed explicitly), which is why the platform prompt was the only setting that survived a pull-claimed turn. Everything else fell back to agent/global defaults, silently: nothing logged, nothing failed, the execution row recorded a normal `success`.

Same shape of bug one line earlier: the payload's `session_id` was read from `meta["session_id"]`, a key the producer never writes (it writes `resume_session_id`).

Evidence from the issue, measured on `eu2`/`cornelius-oracle`: `backlog_metadata LIKE '%task_overrides%'` matched **0 of 580 rows**, and the 18 executions that failed on their own configured timeout in the 7 pre-flip days became 0 after the pilot flip — the row's timeout no longer being enforced, presenting as a reliability improvement.

## Wire-shape decision: keep `task_overrides`, populate it

The agent-side consumer is `docker/base-image/agent_server/services/pull_worker.py::_run_and_report`:

```python
session_id = payload.get("session_id")
overrides  = payload.get("task_overrides") or {}
... execute_headless(model=overrides.get("model"),
                     allowed_tools=overrides.get("allowed_tools"),
                     system_prompt=overrides.get("system_prompt"),
                     max_turns=overrides.get("max_turns"),
                     resume_session_id=session_id, persist_session=bool(session_id))
```

So the consumer genuinely expects the nested object, and `MESSAGE_ENVELOPE_SCHEMA.md` §2.2 defines `task_overrides` as the deliberate quarantine sub-object for per-task knobs. **The wire field is therefore unchanged** — the fix populates it from the flat keys the producer actually writes, rather than renaming anything. Renaming would have meant a lock-step backend + agent-base-image change for no gain.

Two properties preserved on purpose:

- **A nested `task_overrides` still overlays the flat keys and wins.** Nothing writes it today, but it is the §2.2 target shape, and keeping it honoured means this change is purely additive (it also keeps `test_1081_pull_endpoints.py`'s existing #1629 fixtures meaningful).
- **#1629/#1633's composed `system_prompt` is byte-identical in behaviour.** The only difference is that the caller override handed to `compose_system_prompt(caller_prompt=…)` is now read from the flat `system_prompt` key it is actually stored under, and `model` (the ent#243 prompt-tier signal) now resolves from the row rather than always falling through to `model_used`.

Absent settings are **omitted** rather than sent as `null`, so a claim for a bare row still yields `{"system_prompt": …}` only.

## `timeout_seconds` needed the consumer too

The envelope alone would not have fixed the timeout: **the worker never read `timeout_seconds` at all**, running every pulled turn to `TRINITY_PULL_TURN_TIMEOUT` (900s default). `pull_worker` now resolves the row's budget with a fail-soft fallback to the pool default (missing / non-numeric / non-positive → default; never raises). The **result-delivery deadline stays keyed to the pool budget** (`max(row, pool)`), so honouring a short per-task timeout cannot shrink the retry window the lease reaper backstops — the backend lease derives from the agent's `execution_timeout_seconds`, not the row's.

This half lives in the agent base image, so it takes effect after `/rebuild-agent`. The backend half is live on deploy.

## Test

`tests/unit/test_2317_pull_envelope_parity.py` (new, 9 tests). The reason this bug survived is that both consumers were only ever tested against **hand-written** metadata fixtures shaped like whatever the consumer under test happened to read. So every test here runs off **one fixture built by the real producer** — an actual `BacklogService.enqueue` call against a fake db:

- `test_producer_keys_are_exhaustively_classified` — every key `enqueue` writes is either runtime-facing or backend-only. Adding a key to `enqueue` reds this until someone classifies it, which is the moment to ask whether the pull path needs it.
- `test_pull_envelope_and_push_drain_apply_the_same_settings` / `test_both_paths_apply_the_producers_recorded_values` — both consumers resolve the same producer keys to the same values, and to the producer's own values (so the pair cannot drift together onto the wrong thing).
- Plus: `task_overrides` populated not empty; `session_id` from `resume_session_id` and never the `chat_session_id` row id; `system_prompt` still composed; nested overlay still wins; absent settings omitted not nulled.

**7 of the 9 fail on the pre-fix code** (verified by stashing the fix). 4 more tests in `tests/unit/test_1081_pull_worker.py` cover the worker half; 3 of those fail pre-fix.

The file is order-safe under `pytest-randomly` (it imports every backend module *before* installing either stub, and carries the `_STUBBED_MODULE_NAMES`/`_restore_sys_modules` snapshot pair) — an earlier draft without that reddened `test_1081_pull_endpoints.py` under some seeds, which is noted in the file.

## Commands

```
cd tests
python -m pytest unit/test_2317_pull_envelope_parity.py unit/test_1081_pull_worker.py \
  unit/test_1081_pull_endpoints.py unit/test_1081_pull_mode.py unit/test_1081_lease_reaper.py \
  unit/test_1081_lease_sweep_exclusion.py unit/test_1081_physical_meter.py unit/test_backlog.py \
  unit/test_1449_backlog_metadata_scrub.py unit/test_1766_pull_pilot_exclusive.py \
  unit/test_2048_pull_pilot_reach.py unit/test_chat_sync_backlog.py -q
# 229 passed, 1 skipped

python -m pytest unit/ -q --randomly-seed=1     # 21 failed, 11862 passed, 30 skipped
python tests/lint_sys_modules.py                # OK, no new violations
python tests/lint_root_test_placement.py        # OK
```

The 21 failures are **identical on `origin/dev` with no changes applied** (same 21 ids, byte-for-byte) — all IPv4-mapped-IPv6 SSRF-guard cases (`test_736_a2a_outbound_*`, `test_ent14_registry_url_ssrf`, `test_ent399_ipv6_origin`, `test_mcp_validator`) that fail on a local Python 3.11/macOS interpreter; CI pins 3.13. Nothing in the pull/backlog area failed on either side. The pull-adjacent files were additionally run green under 8 `pytest-randomly` seeds.

## Notes on the issue's analysis

Three corrections, none of which change the fix:

1. **`max_turns` is dropped on the push path too**, so this was never a pull-only loss. `_spawn_drain` rebuilds `ParallelTaskRequest(max_turns=…)`, but `chat_execution_service.run_async_task` never forwards it to `TaskExecutionService.execute_task` (which has no `max_turns` parameter) and the `/api/task` payload the backend sends the agent contains no `max_turns` — it dies in `run_async_task`. After this PR the **pull** path honours it (the worker does pass it to `execute_headless`), so the asymmetry now runs the other way, in favour of the caller's stated intent. The push-side gap is real and worth its own issue; it is out of scope here.

2. **The issue's suggested fix is necessary but not sufficient for the timeout.** "Have the pull envelope read the same flat keys" fixes model/tools/turn-cap/session, but the worker never read `timeout_seconds` from the envelope, so the field it names as the visible production symptom needed the consumer change above.

3. **`chat_session_id` must NOT source `session_id`.** The issue says "align `session_id` with `chat_session_id` / `resume_session_id`"; only `resume_session_id` is correct. `chat_session_id` is a Trinity `chat_sessions` row id, while the envelope's `session_id` is the Claude Code session UUID that the worker feeds straight into `execute_headless(resume_session_id=…)` — putting the row id there would resume a session that does not exist. The test pins this explicitly.

Related, and deliberately **not** fixed here: `save_to_session` / `chat_session_id` / `user_message` / `create_new_session` / `inject_result` are applied post-turn by `run_async_task` on the push path, and the pull sink (`apply_task_result`) has no equivalent — so a pulled turn is not persisted into the caller's chat session and its result is not injected. That is a separate gap in the pull sink, not an envelope-shape bug, and the test classifies those keys as backend-only with that reasoning written down.

### DIFF
--- a/docker/base-image/agent_server/services/pull_worker.py
+++ b/docker/base-image/agent_server/services/pull_worker.py
@@ -194,6 +194,25 @@ def _turn_timeout() -> int:
     return max(1, n)
 
 
+def _resolve_turn_timeout(overrides: Dict[str, Any]) -> int:
+    """The claimed row's own turn budget when the envelope carries one, else the
+    pool default (#2317).
+
+    The backend records a concrete ``timeout_seconds`` on every queued row
+    (``backlog_service.enqueue`` — the caller's override, or the agent's
+    ``execution_timeout_seconds``) and the PUSH path enforces it. Before #2317 the
+    claim envelope never carried it and this pool ran every pulled turn to
+    ``TRINITY_PULL_TURN_TIMEOUT`` instead, so a row asking for 60s got 900s.
+    Fail-soft: a missing / non-numeric / non-positive value falls back to the
+    pool default, never raises."""
+    raw = overrides.get("timeout_seconds")
+    try:
+        n = int(raw)
+    except (TypeError, ValueError):
+        return _turn_timeout()
+    return n if n > 0 else _turn_timeout()
+
+
 # ---------------------------------------------------------------------------
 # Result body (§3.3 reply payload) construction
 # ---------------------------------------------------------------------------
@@ -373,7 +392,7 @@ async def _run_and_report(
     message = payload.get("message") or ""
     session_id = payload.get("session_id")
     overrides = payload.get("task_overrides") or {}
-    turn_timeout = _turn_timeout()
+    turn_timeout = _resolve_turn_timeout(overrides)
 
     agent_state.record_task_start()
     try:
@@ -416,7 +435,10 @@ async def _run_and_report(
         _persist_pull_result(execution_id, {"execution_id": execution_id, "body": body})
 
     # Deadline = now + (turn timeout + buffer) = the slot-lease TTL window.
-    deadline = time.monotonic() + turn_timeout + _SLOT_TTL_BUFFER_SECONDS
+    # #2317: keyed to the POOL's budget, not the row's — honouring a short
+    # per-task timeout must not also shrink the result-delivery retry window
+    # (the backend lease derives from the agent's execution_timeout, not the row's).
+    deadline = time.monotonic() + max(turn_timeout, _turn_timeout()) + _SLOT_TTL_BUFFER_SECONDS
     delivered = await _deliver_result(client, execution_id, body, backend_url, mcp_key, deadline)
     if delivered and persisted:
         _delete_pull_result(execution_id)
--- a/docs/planning/MESSAGE_ENVELOPE_SCHEMA.md
+++ b/docs/planning/MESSAGE_ENVELOPE_SCHEMA.md
@@ -177,6 +177,16 @@ Model selection is **not** a task field: per-call `model` is demoted to a
 session/schedule attribute (map #2, MODEL-001). `async_mode` disappears —
 everything is async; sync is the `?wait=true` edge adapter (map #4).
 
+> **Pilot reality (#2317).** The table above is the TARGET shape. Until the
+> demotion PRs land, the producer of a queued row (`backlog_service.enqueue`)
+> still records a per-task `model` and `timeout_seconds`, the push path enforces
+> both, and the pull worker reads both off `task_overrides` — so the claim
+> envelope built by `pull_coordination_service._build_claim_response` carries
+> `model` and `timeout_seconds` alongside the three fields above. They leave the
+> payload when the demotion that removes them from `ParallelTaskRequest` lands,
+> not before; dropping them from the wire earlier is a silent correctness
+> regression on the pull path (that was #2317).
+
 ### 2.3 `kind: event`
 
 Agent event pub/sub (EVT-001). Emitted once by a source agent; the projector
--- a/src/backend/services/pull_coordination_service.py
+++ b/src/backend/services/pull_coordination_service.py
@@ -155,6 +155,27 @@ class ResultApplyOutcome:
 
 
 # ---------------------------------------------------------------------------
+# Per-task settings carried on the claim envelope (#2317)
+# ---------------------------------------------------------------------------
+# `backlog_service.enqueue` persists a queued row's per-task settings as FLAT
+# keys inside `backlog_metadata`, and the PUSH drain (`_spawn_drain`) reads them
+# flat. The pull envelope must consume the SAME keys or a pulled turn silently
+# loses them. These are the keys that reach the RUNTIME (they map 1:1 onto
+# `execute_headless` kwargs the pull worker passes); the remaining metadata keys
+# are backend-side concerns (chat-session persistence + provenance) that neither
+# path sends to the agent. `tests/unit/test_2317_pull_envelope_parity.py` pins
+# that split against the producer itself, so a key added to `enqueue` cannot
+# silently belong to neither set.
+_TASK_OVERRIDE_KEYS = (
+    "model",           # --model            (push: ParallelTaskRequest.model)
+    "allowed_tools",   # --allowedTools     (push: ParallelTaskRequest.allowed_tools)
+    "max_turns",       # --max-turns        (push: ParallelTaskRequest.max_turns)
+    "timeout_seconds", # turn budget        (push: ParallelTaskRequest.timeout_seconds)
+    "system_prompt",   # --append-system-prompt; recomposed below (#1629)
+)
+
+
+# ---------------------------------------------------------------------------
 # Claim (GET /api/internal/next-task) — §3.1 / §3.2
 # ---------------------------------------------------------------------------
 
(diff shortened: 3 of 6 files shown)


vybe (approved):
Validated after rebase onto `dev`. **Closes #2317** (closing keyword present). All 21 non-skipped checks pass; regression diff clean across all six seeds (base 12130 → head 12143, 0 failures either side, **No new failures**).

**Rebase note.** The branch conflicted on `tests/registry.json` after #2333 landed. Resolved by rebuilding from the three git stages rather than splicing markers: `ours` (dev) and `theirs` (this PR) add disjoint entries — dev added `test_433_headroom_history` + `test_471_subscription_usage_observability`, this PR adds `test_2317_pull_envelope_parity` — with **no removals and no modified entries** on either side, so the union is unambiguous. The new entry is inserted at the position `theirs` placed it (after `test_2246_oss_only_deploy_is_loud`). Verified the resolution is a **+13-line insert and nothing else**, and that a round-trip re-serialization of dev's own blob is byte-identical, so no reformatting rode along.

**Substance.** `test_2317_pull_envelope_parity.py` is the named regression test, and it is the right shape: every case runs off one fixture built by the **real producer** (`BacklogService.enqueue`), with exhaustive key classification that reds the suite when a new key appears unclassified — so the pull and push consumers cannot silently diverge again. Both sides of the vendored boundary are touched (`pull_coordination_service.py` and `agent_server/services/pull_worker.py`), which is what the fix requires. `MESSAGE_ENVELOPE_SCHEMA.md` updated.

Security scan clean, no new env var, no new backend package, no host paths or mode flips. This is dark until an agent is piloted (`PULL_MODE_PILOT_AGENTS`), so blast radius is nil today.

PR #2327, opened 2026-08-19
feat(dashboard): Subscription pressure grid tile — per-subscription 5h/7d limit spend, colour-banded (abilityai/trinity-enterprise#259)

trinity-ability:
## Summary

One row per **subscription** on the Dashboard Grid: how much of each rolling limit is spent, colour-banded, plus 429 history and honest degradation. Answers *"how much is left on any of my subscriptions?"* without opening Settings.

It is the **inverse unit** of #471's per-agent pressure chip. That chip says whether *an agent's* funding is strained; it structurally cannot say *which subscription* is the bottleneck, because agents share a subscription and burn one 5h window between them — reading it off the board otherwise means grouping every agent chip into buckets by eye.

**Zero backend change**, per the operator ruling on the issue (2026-08-19): *"a small build on `GET /api/agents/subscription-pressure` + the extended `GET /api/subscriptions/{id}/usage` once #471 merges"*. I built a batched admin endpoint first and **reverted it** — at the real fleet size (one subscription per Claude account) it optimizes ~0.5 q/s, and `assert_admin` rejects agent principals (#1890), so ent#351 could not have reused it anyway. The fan-out is fewer requests than `fetchOpQueuePending` already makes on the same poll.

## What a row may claim

All decisions live in `utils/subscriptionPressureTile.js` — vitest is node-environment here, so a rule inside the SFC is a rule no test can reach.

| Shown | Rule |
|---|---|
| `5h 97%  7d 88%` | **Both** limits, each on its own band (green < 60, amber < 85, red ≥ 85). Two windows because either can be the wall you hit — a subscription routinely sits at 20% of its 5h while its weekly is nearly gone. Rendered **only** while the provider snapshot is fresh (`source=anthropic`, ≤ 30 min). #471 made the number real, so the AC's *"never a fake-precise X% left"* governs **framing** — utilization consumed, source-labelled — not disclosure. Labelled `5h`/`7d`, never "daily"/"weekly": rolling windows with a visible `resets_at`. |
| `near` chip | A red window raises severity to `warn` even with a spotless failure history — the only signal that exists **before** the first 429. Caught during local verification: without it, a subscription at 93% of its weekly ranked *below* one whose usage read merely failed and got pushed into the overflow. The chip separates `429s` (happened) from `near` (has not). |
| `3× 429` | The `rate_limit` kind **alone**, never `failure_events_24h`, which also carries `auth` + pre-#471 `unknown`. Migration 0040 split them at the data layer; reporting the total under a "429" label would undo that fix in the UI. |
| `≈$3.12 · 1.2k out` | **`input_tokens` is deliberately off the row face**: it is `SUM(schedule_executions.context_used)` — context-window *occupancy*, not tokens consumed — the ruling `architecture.md` already records for the sibling ent#101 tile. It appears in the tooltip, labelled an estimate. Cost is always API-equivalent, never a bill. |
| `unavailable` | A failed per-subscription read renders that row unavailable — never zeroes, never a healthy-looking row. |

## Honest status

- **Empty** requires a fetch that *succeeded and returned zero*, never `rows.length === 0`.
- A **wrong-shaped 200 is a fault**, not an empty fleet (`Array.isArray` guard, mirroring the ent#100 fetchers) — laundering it would tell an operator their subscriptions are *not configured*, a different and much worse claim than "could not read them".
- A poll failing **after** a good one keeps its rows and marks the stamp `stale`, rather than replacing real data with an error panel.
- **Overflow is disclosed**: a `+N more` row plus a `3 of 9` stamp, because past the fixed track count a row is clipped by `InfoTile`'s `overflow: hidden` with no scroll, ellipsis or trace.

## First `adminOnly` tile

Load-bearing, not cosmetic: every endpoint it reads is admin-gated (per-subscription **spend**) and its footer link targets `Settings`, whose route is `requiresAdmin`. The flag doubles as the fetch gate — the widget key is absent for non-admins, so the poll never fires and there is no 403 loop. Because `isAdmin` is only confirmed once `fetchUserProfile()` lands, the store also watches the key appearing and re-polls; otherwise a cold load leaves the tile blank for 60s.

`fetchSubscriptionPressure()` (the per-agent chip feed) stays **ungated** — chips and list badges must keep working while this tile is off.

## Changes

- `src/frontend/src/utils/subscriptionPressureTile.js` — new pure module (all decisions)
- `src/frontend/src/components/tiles/SubscriptionPressureTile.vue` — new tile
- `src/frontend/src/components/tiles/catalog.js` — registration (`adminOnly: true`, `defaultOn: true`, both recorded as explicit decisions)
- `src/frontend/src/stores/subscriptions.js` — `fetchPressureData()` + roster/`listLoaded`/`listError`
- `src/frontend/src/stores/fleetGrid.js` — gated batch-poll entry + key-appearance watch
- `src/frontend/tests/unit/fleetGridFailuresFetch.spec.js` — mocks the new store, matching how it already mocks `@/stores/executions` (the real store imports `api.js`, which replaces `.get` on the axios instance and would break that file's mock)
- docs: `dashboard-grid-view.md`, `subscription-usage-tracking.md`, `architecture.md`

## Test Plan

- [x] `npm run test:unit` — **1037 passing** (73 new across two specs; baseline was 964)
- [x] Route + tab-id pinned against the real router and `Settings.vue`, and **mutation-tested** (flipping `tab` to `subscriptions` turns it red) — `gridTileLinks.spec.js` validates route *names* only, so a wrong tab would otherwise land the operator on an unrelated Settings tab silently
- [x] `npm run check:tokens` clean; new files scan at **zero** raw colors (no `var(--x, #hex)` fallbacks — `gridTokens.spec.js` already proves every token is defined in both themes, so a fallback is dead code that still counts)
- [x] `npm run build` clean
- [x] **Verified against the live local stack** (dev server from this checkout — the running container mounts a *different* checkout): real Anthropic headroom `5h 15% / 7d 10%` rendering green in **both** light and dark; multi-state fleet exercising red/amber/green bands, the `near` and `429s` chips, the unavailable row and the overflow link

## Out of scope

Trend line / sparkline is **ent#433's** (headroom history), by explicit coordination in that issue: *"if #259 builds first it ships point-in-time only and gains the trend when this lands."*

Fixes abilityai/trinity-enterprise#259


vybe (approved):
Validated via /validate-pr. Zero backend change, all decisions in a pure testable module, 76 new tests weighted to non-happy paths. Verified locally: 1041 tests pass, check:tokens clean, raw-color ratchet holds with the new tile at zero.

Independently confirmed `adminOnly` is honored at gridWidgets.js:120 rather than an inert flag — that was the one way this could have shipped a 403 loop.

Approving.

PR #2331, opened 2026-08-20
feat(subscriptions): headroom history — durable probe snapshots for utilization trends (abilityai/trinity-enterprise#433)

trinity-ability:
## Summary

- **The durable half of #471.** The live headroom probe keeps exactly ONE last-known-good Redis snapshot per subscription, overwritten on every probe — so *"how close did we run to the 5h wall this week"* was structurally unanswerable. This persists each probe as a row and exposes a bounded, gap-honest time series.
- **Adds no probe.** The write hook sits inside `_probe_and_store`, the single chokepoint every probe passes through, so it inherits #471's entire rate-bounding envelope (60s floor, cross-worker single-flight, fail-closed ambient gate). Probe cadence and spend are unchanged, which the issue explicitly requires.
- **Also converts a hardcoded 24h sweep into a real retention window** — a deliberate, operator-approved scope expansion (see below).

## ⚠️ Deliberate scope expansion — please read before flagging scope drift

This PR contains **two** related changes. The second was raised as a separate concern during planning, and the operator explicitly chose to fold it in rather than file it separately:

`subscription_rate_limit_events` — the platform's **only durable record of real agent work hitting a provider rate limit**, timestamped and attributed to the causing agent — was swept at a **hardcoded `iso_cutoff(24)`** with no operator-visible window, no #1644 blast-radius guard, and no `GET /api/settings/retention` entry, while every sibling table had all three. It was doubly capped: `pressure_states` also hardcodes `hours=24`. It now has `subscription_failure_event_retention_days` (default 30).

Widening 1 day → 30 days is the **#1638-safe direction**: no install loses data, and it cannot change any existing answer because every consumer already filters by time itself.

**The series is `last`-per-bucket, never `max`** — three independent arguments, any one sufficient:
1. **Observer effect.** Probes are demand-driven (they fire only on an HTTP request), so samples-per-bucket ∝ operator attention. `E[max of n]` rises with `n`, so a watched hour out-reads an identical unwatched one — and the unattended overnight burn, the thing most worth seeing, gets the fewest samples and the LOWEST reading.
2. **Two-peak ambiguity.** 5h and 7d peak at different instants inside a bucket, so "the peak sample's timestamp" is undefined for a two-column response. `last` gives one *correlated* snapshot of both.
3. **Invisible 429s.** A 429 carries `status='rate_limited'` with `utilization_pct = NULL`; a max drops the most important sample in the series and flatlines the chart through an outage.

The headline question is still answerable — max **across** buckets, a far less biased estimator.

**Honest gaps need TWO timestamps.** Emitting only non-empty buckets with real timestamps *cannot* support gap detection: jitter and gaps are indistinguishable from timestamp deltas (10:05→11:55 is 1h50m with **no** gap; 10:55→12:05 is 1h10m **with** one). Each bucket carries the logical `bucket_start` beside the real `fetched_at`; nothing is interpolated or zero-filled, and `coverage_pct` lets a thin series report itself as thin.

**Three load-bearing write properties**, each pinned by test:
- **Off the event loop** (`asyncio.to_thread`) — the DB is DELETE-journal with a 30s busy timeout, and a sync write landing during the 03:30 backup or 04:30 VACUUM would stall `/health`, the WS dispatcher and every in-flight request. `try/except` handles *errors*, not *blocking*.
- **After** the Redis write — the snapshot is what every live surface reads.
- `except Exception`, never `BaseException` — shutdown's `CancelledError` must propagate.

**No FK in the Alembic revision.** `_PG_TABLE_SUBS` regex-strips every `FOREIGN KEY … REFERENCES` clause before DDL reaches PostgreSQL, so the platform has **zero enforced FKs**. The natural `op.create_table(..., ForeignKeyConstraint(...))` would create the first one and diverge the backends on the in-flight-probe delete race. Cascade is explicit inside `delete_subscription`'s transaction.

**`ROW_NUMBER()`, not a bare column beside an aggregate** — `MAX(x)` with a bare `fetched_at` in a `GROUP BY` is a SQLite-only extension that raises `GroupingError` on PostgreSQL. Since the PG matrix only runs with `TEST_POSTGRES_URL` set, the dialect is pinned by **compiling** the query for PG, which runs everywhere.

**Storage:** dedicated table, not `product_events` — that table has no `subscription_id`, binds these rows to the activation-funnel's retention window, and **egresses** (`telemetry_sharing_service.build_aggregate_payload` counts it by type and POSTs it on Tier-2 opt-in).

## Changes

**Schema (four-file dual-track):** `db/schema.py`, `db/tables.py`, `db/migrations.py`, `migrations/versions/0041_subscription_headroom_history.py` (single head verified)
**DB:** `db/subscriptions.py` — insert, bucketed read, two candidate counts, two chunked prunes, two shared predicates, explicit cascade
**Facade:** `database.py` — 5 hand-written delegations (no `__getattr__`)
**Service:** `services/subscription_headroom_service.py` — `_history_row`, `_record_history`, `HISTORY_WINDOWS`, `get_history`
**Retention:** `config.py`, `services/settings_service.py`, `services/cleanup_service.py` (new `_read_retention_setting` helper, two guarded sweeps, report fields; also fixes a pre-existing omission of `agent_reminders_pruned` from the WAL-checkpoint sum, in the line being touched)
**API:** `routers/subscriptions.py`, `models.py`
**Docs:** requirements §20.5b, architecture.md (incl. the retention counts it states explicitly: 8→10 row windows), new feature flow

## Test Plan

- [x] `pytest tests/unit/test_433_headroom_history.py -v` — **41 passed**
- [x] Full unit suite — **11,935 passed** (2 failures are local-venv false positives in `test_1920` / `test_ent314`; both **pass on clean `origin/dev`** — verified in a temp worktree — because those guards walk a local `venv/` and the mounted private submodule)
- [x] Retention gates: `test_1771a` (the AST guard requiring exactly one `_guard_allows` per key), `test_1644`, `test_1638`, `test_297`
- [x] `check_alembic_heads.py` — 42 revisions, 1 head
- [x] `test_schema_parity`, `test_models_centralized`, `test_alembic_revision_id_length`
- [x] Enterprise-docs guard run locally — no disclosure

**Four pre-existing tests updated**, all characterizations of the old behaviour:
- `test_subscription_auto_switch_pingpong` asserted the hardcoded 24h sweep — that assertion documented the defect, not a contract
- `test_297` pinned `len(RETENTION_OPS_KEYS) == 9`
- `test_1834` counted alarms across the whole cycle; now scoped to its own window so a future retention key can't break it
- `test_cleanup_inner_sweeps` needed the new prune return values in its mock harness

## Notes for the reviewer

- **OSS-core by explicit decision** (the #471 gate ruling carries over), recorded in requirements §20.5b so it is never inferred backwards from the mere fact that it merged (the ent#326 discipline).
- **Backend only.** AC-7 names ent#259's tile as the consumer; that tile is pinned to one cell with `overflow: hidden` and no scroll, so the viable consumer is an in-tile **sparkline**, not a labelled trend chart. Recorded rather than silently ignored.
- Cross-repo: the closing keyword does not auto-close a private-tracker issue — ent#433 is closed manually at release.

Fixes abilityai/trinity-enterprise#433


vybe (approved):
Validated via /validate-pr: security scans clean, dual-track migration present, docs updated (architecture.md, feature-flows index + new subscription-headroom-history.md, requirements/security.md), tests present incl. test_433_headroom_history.py, all 22 checks green.

Reviewer work during merge: two dev merges (post-#2330, post-#2354, post-#2312). Alembic revision re-parented 0041_subscription_headroom_history -> 0043, chained off 0042_agent_ownership_operator_resume — three PRs each claimed 0041 off 0040_rl_events_failure_kind, and with dev at strict:false a stale PR is not re-checked, so the fork would not have been caught at merge time. Three stale doc references to the old revision id updated with it. The architecture.md conflict was non-additive (#2352 inserted a sentence mid-paragraph, ent#433 appended one) — both preserved. 119 subscription/retention unit tests pass on the merged tree. Cross-tracker ent#433 needs a manual status-in-dev bump + manual close at release.

PR #2333, opened 2026-08-20
fix(codex): materialise API-key auth.json, or every turn 401s (#2208)

### LINKED ISSUE #2208: bug: Codex API-key auth is inert — every turn 401s because nothing writes $CODEX_HOME/auth.json

obasilakis:
Closes #2208

## The bug

Trinity resolved the OpenAI API key and passed it to `codex exec` **via the environment only**. That is no longer sufficient: the CLI authenticates its `wss://api.openai.com/v1/responses` transport from `$CODEX_HOME/auth.json`, and that transport is no longer toggleable (`responses_websockets` / `..._v2` are listed as *removed* in `codex features list`). Nothing in `startup.sh` or `codex_runtime.py` ever ran `codex login`, so an API-key Codex agent could not complete a single turn — only subscription agents (#1971, which ship their own `auth.json`) worked.

## The fix

`_materialize_api_key_auth` runs in `_execute_codex` **before** the spawn — after would still 401 — shelling out to `codex login --with-api-key` with the key on **stdin, never argv**.

Three properties carry it:

* **Subscription auth wins** (#1971) — only an `auth_mode: apikey` file is overwritten. A `chatgpt` file, or one that will not parse, is left alone. Classification is tri-state (`absent` / `unreadable` / `parsed`) so "no file" and "a file I cannot read" never collapse into "safe to write".
* **Rotation propagates** — since #1999 the execution env is rebuilt from `.env` per spawn, so the key can change under a live container; a stored key that no longer matches is re-logged-in. Steady state is one small file read per turn, not a subprocess.
* **It never raises** — a login failure logs at ERROR and the turn proceeds to fail with the CLI's own auth error, rather than converting an auth problem into a Trinity 503 on the subscription path too.

## The second defect, created by the first fix

`_has_subscription_auth` was an existence test. Once Trinity writes an `auth.json` of its own, that reports **every API-key agent as a subscription agent** — and the #1971 credential gate would then let a container whose key was *removed* from `.env` keep running on the stale file it left behind: silent failure of credential revocation, the #1999 class. It is now keyed on `auth_mode`.

An `auth.json` with no `auth_mode` (older CLI), or an unreadable one, still counts as a subscription credential — preserving pre-#2208 behaviour exactly. #1971's point is that a bad `auth.json` surfaces as the CLI's own auth error, not as a Trinity 503 claiming nothing is configured.

## Verification

Live OpenAI key, rebuilt base image (**codex 0.147.0**), key present **only** in `~/.env` and absent from the container env — i.e. AC #1's setup:

| module | `POST /api/chat` | `POST /api/task` | 401s | `auth.json` |
|---|---|---|---|---|
| patched | `"PONG"` | `"TASKPONG"` | 0 | `auth_mode=apikey`, cost recorded |
| `origin/dev` (unpatched) | 503 `Codex authentication failure … 401 Unauthorized` | — | present | absent |

Same image, same key, same container setup; only the module differs. The key appears in no log line and in no argv.

Isolated A/B on the CLI itself, model held constant, only `auth.json` differing — reproduced on **both** pinned versions, matching the issue's own table:

```
codex 0.139.0, gpt-5.3-codex   A) no auth.json → exit=1, 401s=18   B) after login → exit=0, PONG
codex 0.147.0, gpt-5.6-sol     A) no auth.json → exit=1, 401s=19   B) after login → exit=0, PONG
```

Unit: `tests/unit/test_codex_runtime.py` +14 tests (mode classification, subscription/unclassifiable preservation, rotation, idempotence, stdin-not-argv, failure logs without the key, and login-before-spawn ordering). `695 passed, 4 skipped` across the codex / agent_server / vendor-parity selection.

## Notes for review

* Two **existing** test helpers now neutralize the login spawn: `_install_fake_codex` (same reason it neutralizes the pgid helpers — there is no real codex binary) and #1971's `_captured_env`, where the first `Popen` captured would otherwise be the login rather than the turn, asserting credential forwarding against the wrong process.
* **Requires a base-image rebuild.** Existing Codex agents adopt it on their next *cold* start via the image-drift predicate (#1809), not on a plain restart.
* Chose the CLI over hand-writing `auth.json`: the file format is undocumented, and `--api-key` is removed in current versions, so the ticket's `--with-api-key` stdin form is the stable seam.
* The issue's other candidate site was `startup.sh`. Lazy-in-runtime was chosen instead — it also covers a wiped `CODEX_HOME` and key rotation, and keeps one writer rather than two.
* `gpt-5.1-codex` 404s on `/v1/responses` for API-key callers — the thing #2207 explicitly left unverified ("treat the pin as suspect until a real turn proves otherwise"). Already moot on `dev`: #2207 shipped `gpt-5.6-sol`, and `get_default_model()` only feeds the context-window fallback, never `-m`. Recorded here so the question is closed rather than re-opened later.


obasilakis:
## `/review` pass — findings and what changed

Ran the pre-landing review after opening this PR. It found one credential-exposure defect and four smaller items; all but one are fixed here, the remaining one is filed as #2334.

### Critical — `sk-svcacct-` keys were not redacted

`credential_sanitizer` (both the backend and agent-server copies) carried three rules: `sk-[a-zA-Z0-9]{20,}`, `sk-proj-…`, `sk-ant-…`. An OpenAI **service-account** key matched none — the generic rule stops at the first hyphen, the other two are literal prefixes. Proven, not inferred:

```
LEAKED   sk-svcacct-AAA...
REDACTED sk-proj-BBBBBB...
REDACTED sk-CCCCCCCCCCC...
REDACTED sk-ant-DDDDDDD...
```

Pre-existing and independent of the auth fix, but it is precisely this issue's AC #3 ("the key never appears in argv, logs, or an error body"), and `_login_with_api_key` promises "sanitized, never the key". Replaced with one shape-based rule per family in both copies, parametrized over five variants, plus an over-redaction guard — `sk-` is also an ordinary prefix (`sk-SK` locale tags, `sk-42` ticket ids) and the length floor has to keep those readable.

### Also fixed

- **`_materialize_api_key_auth` claimed "never raises" but didn't** — it caught only `OSError`/`ValueError`/`TimeoutExpired`. On the turn path, anything else escaped into `_execute_codex` and would 500 the turn, *including for a subscription agent that needed nothing from the function*. Now total, with a test.
- **The login spawn snapshotted `{**os.environ}`** instead of `build_execution_env` — the #1999 rule, caught by that issue's own guard test. Re-verified live afterwards: a dropped `PATH` there would make `codex` unfindable.
- **Rotation rationale named the wrong mechanism** — I credited #1999's per-spawn env rebuild; it is actually `_load_api_key_with_source` re-resolving per turn, and it prefers the process env over `.env`, so `.env` rotation only reaches agents with no baked key. Corrected in the docstring and flow doc.
- **`harness-authoring-guide.md` §6 carried the same stale contract FR-5 did** ("inject it into the subprocess env", full stop) — the doc a future harness author follows. Generalized.

### Filed, not fixed

**#2334** — a persistently failing `codex login` re-runs on every turn, up to the 30s timeout each time. Negative caching must not outlive a corrected `.env`, and the failure is already loud (ERROR per attempt), so the tradeoff is its own decision rather than a drive-by here.

### Test state

`11,724 passed` on the branch. The failure set is **identical** to clean `origin/dev` (26 pre-existing, incl. the `test_736_a2a_outbound_edges` IPv6 group and the known `test_subscription_auto_switch_pingpong` ordering errors, #1846) — verified by diffing both runs rather than eyeballing, after two genuine regressions from this branch showed up in that diff and were fixed (the two above).

Learning recorded in `docs/memory/learnings.md`: prefix-enumerated secret patterns are a blocklist, and the entry you have not heard of is the one that leaks.

vybe (approved):
Validated: Codex API-key auth.json materialisation. All CI green (12145 tests, regression diff clean, prod-image-smoke, non-root, schema-parity). Closing keyword present (Closes #2208). No security findings, no new backend package or env var, no host paths. Merging.

Issue #444, opened 2026-04-21
bug(SUB-003): auto-switch ping-pongs between rate-limited subscriptions due to clear_rate_limit_events destroying detection signal


vybe:
## Summary

When multiple Claude Max subscriptions are simultaneously exhausted (e.g. during a weekly usage-limit window), the SUB-003 auto-switch service repeatedly re-assigns agents between two already-exhausted subscriptions every ~15–60 minutes instead of detecting that no viable alternative exists. Each cycle causes: 2× failed scheduled executions with 429, a DB write, and a full container restart. Over ~90 minutes this was observed to cause 8+ switches across 4 agents, all looping between subscriptions that each report the same weekly reset time. The bug is a logic defect where the detection signal is wiped on every switch, so the "is this subscription rate-limited?" check always returns `False` on the next cycle.

## Component

Backend / Subscription Auto-Switch (SUB-003) — `services/subscription_auto_switch.py` + `db/subscriptions.py`

## Priority

P1 — major feature broken (auto-switch actively worsens behavior when all subs are exhausted). Workaround exists: disable the `auto_switch_subscriptions` system setting.

## Observed behavior

Sample timeline for one agent over ~75 minutes (all three subscriptions report `resets <date>, 3pm (America/New_York)` — i.e. truly out of quota):

| Time | Action | Sub before → after |
|------|--------|-------------------|
| T+0   | auto-switch | sub-A → sub-B |
| T+29m | auto-switch | sub-B → sub-A |
| T+50m | auto-switch | sub-A → sub-B |
| T+74m | auto-switch | sub-B → sub-A |

After 10+ 429 errors and 8+ switches, `subscription_rate_limit_events` contains **zero rows** in the last 2-hour window — events are deleted faster than the window can hold them.

## Error

Underlying 429 body (from Anthropic, faithfully surfaced by agent):
```
Subscription usage limit: You've hit your limit · resets <date>, 3pm (America/New_York).
To resolve: (1) wait for the usage reset, (2) set an ANTHROPIC_API_KEY on this agent
for pay-per-use billing, or (3) assign a different subscription token in
Settings → Subscriptions.
```

Backend INFO log pattern (repeats):
```
[SUB-003] Rate-limit event #1 for agent '<name>' on subscription <id> — waiting for 2+ before auto-switch
[SUB-003] Auto-switching agent '<name>' from '<sub-A>' to '<sub-B>' after 2 consecutive rate-limit errors
[...]

PR #2354, opened 2026-08-20
fix(subscriptions): a rejected token is not a rate limit (#2352, #2353)

### LINKED ISSUE #2352: bug: subscription reports "rate-limited" for auth failures — is_subscription_rate_limited ignores failure_kind
## Summary

`is_subscription_rate_limited()` counts **any** row in `subscription_rate_limit_events` within the 2h window, with **no `failure_kind` filter**. An `auth` failure (401/403 — dead, expired, or shadowed token) therefore sets `rate_limited_now = true`, and every consuming surface reports the subscription as *rate-limited* when the real problem is a credential.

This is the exact conflation that #471's migration 0040 introduced `failure_kind` to fix. The fix landed at the **display** layer — `rateLimitEventCount()` in `src/frontend/src/utils/subscriptionPressureTile.js` deliberately reads only `failure_kind === 'rate_limit'`, and `failureKindLabel()` refuses to fold `unknown` into `rate_limit` — but the **predicate** was never updated, so the badge that outranks all of it still says "rate-limited" for an auth failure.

## Context

Observed live on the Dashboard "Subscription pressure" tile: **all four** subscriptions carried the red `LIMIT` badge, three of them showing `5h ≈$0 · 0 out` (i.e. no work attempted in five hours), and one simultaneously showing a healthy **9% / 5%** provider-truth reading. A subscription at 9% of its 5h window is not rate-limited — the DB arm of the OR was firing on auth-class events.

Two aggravating factors make this fleet-wide rather than one-off:

1. **The predicate is subscription-scoped, the events are agent-scoped.** One agent's failures mark the subscription limited for every other agent on it.
2. **SUB-003 auto-switch cascades the same failure across the fleet.** `handle_subscription_failure()` records the event, then switches the agent to the next subscription, which fails identically and records there too — so one bad credential can light up every subscription in sequence. This is the #2114 story (a stale `.env` `ANTHROPIC_API_KEY` shadowing subscription auth, "which SUB-003 then mis-attributed to each healthy subscription in turn").

Impact: the operator surface built to answer *"which subscription is the bottleneck"* misdiagnoses a credential problem as quota exhaustion, and points the operator at the wrong remedy (wait for the window to reset vs. re-register the token).

## Acceptance Criteria

- [ ] `db.is_subscription_rate_limited()` no longer counts `failure_kind = 'auth'` rows — a subscription with only auth failures in the window does **not** report `rate_limited_now`
- [ ] `failure_kind IS NULL` (pre-#471 rows) is handled explicitly and the choice is stated in the docstring — these are genuinely unknown, and silently promoting them to `rate_limit` is the same false-precision this issue is about
- [ ] Auth-class pressure is still **visible**, not silently dropped: an agent/subscription whose token is failing surfaces a distinct state (e.g. an `auth` severity feeding a different badge word), so the fix trades a wrong claim for a right one rather than for silence
- [ ] The candidate-skip behaviour in `select_best_alternative_subscription()` is preserved or deliberately re-decided: `subscription_auto_switch.py` currently *relies* on the kind-blind predicate to skip a subscription that recently failed for **either** reason (see the comment at `services/subscription_auto_switch.py` step 4). If the predicate narrows, that skip needs its own kind-blind read so auto-switch does not start cascading onto a subscription with a known-dead token
- [ ] `pressureBadge()` and the tile row keep the one-gate rule — no second derivation of `rate_limited_now` is introduced (#2157)
- [ ] Unit tests cover: rate_limit-only → limited; auth-only → not limited; mixed → limited; unknown-only → the documented choice; and that auto-switch candidate selection still skips an auth-failing subscription

## Technical Notes

- Predicate: `src/backend/db/subscriptions.py` → `is_subscription_rate_limited()`
- Writer: `src/backend/services/subscription_auto_switch.py` → `handle_subscription_failure(failure_kind=...)`
- One-gate derivation: `src/backend/services/subscription_headroom_service.py` → `decorate_usage()` / `pressure_states()` (`db.is_subscription_rate_limited(sid) or _headroom_indicates_limited(headroom)`)
- Consumers of the flag: the per-agent chip (`pressureBadge`), the fleet batch endpoint `GET /api/agents/subscription-pressure`, and the Grid "Subscription pressure" tile
- Column added by SQLite migration `rate_limit_events_failure_kind` + Alembic `0040_rl_events_failure_kind`
- Sibling display defect: the tile shows "rate-limited" in front of a rejected provider token (filed separately)

### LINKED ISSUE #2353: bug: subscription pressure tile shows "rate-limited" for a rejected provider token (invalid_token hidden in tooltip)
## Summary

`pressureHeadline()` tests `rate_limited_now` **before** it ever looks at `headroom.status`, so a subscription whose provider token was **rejected** (`invalid_token`, from a probe that got 401/403) renders on the tile as **"rate-limited"**. The actionable state — *re-register this subscription* — exists in the payload and is written into `rowTooltip()`, but the row face claims the opposite kind of problem, and a tooltip is not where an operator looks first.

The same row also shows no percentages, because `windowReadings()` returns `null` unless `headroomIsFresh()` (`source === 'anthropic'` **and** snapshot ≤ 30 min) — and `decorate_usage()` only promotes `source` to `anthropic` when `headroom.status == "ok"`. That part is correct and deliberate: the observed arm has consumption but no denominator, so there is no honest percentage to show. The defect is that the qualitative fallback then picks the **wrong word**, and the difference between "your quota is spent" and "your credential is dead" is the difference between waiting and acting.

## Context

Observed live: three rows reading `LIMIT … rate-limited` with `5h ≈$0 · 0 out · 7d ≈$X`. A subscription that has run **zero** work in five hours is not being throttled — its probe is failing auth. Nothing on the tile face said so; the only distinguishing signal was a hover.

`invalid_token` is described in `rowTooltip()` as "the most actionable state the payload can carry, and it is otherwise indistinguishable from 'no provider data' on the row face" — this issue is that observation cashed out: it is *also* indistinguishable from real throttling, which is worse than indistinguishable from nothing.

## Acceptance Criteria

- [ ] `pressureHeadline()` reports a distinct headline for `headroom.status === 'invalid_token'` (e.g. `token invalid`) that is **not** the word used for genuine throttling
- [ ] Precedence is explicit and tested: a rejected token outranks `rate_limited_now`, since a subscription whose probe cannot authenticate has produced no evidence about its quota
- [ ] `headroom.status === 'error'` (transport failure / provider drift) is likewise distinguishable from a real limit — it must not read as either "rate-limited" or "ok"
- [ ] The distinct state carries a next action (Settings → Integrations is already the row's `to:` target); no dead end
- [ ] The row's severity still ranks it at the top of the tile — a dead token is at least as urgent as a full window, and must not sort below a healthy row
- [ ] `subscriptionSeverity()` keeps a single source of truth for the badge word and the severity — no second derivation
- [ ] Unit tests in the existing node-env suite cover each `headroom.status` value against both `rate_limited_now` states

## Technical Notes

- `src/frontend/src/utils/subscriptionPressureTile.js` → `pressureHeadline()`, `subscriptionSeverity()`, `rowTooltip()`
- `src/frontend/src/utils/subscriptionPressure.js` → `headroomIsFresh()`, `usageSourceLabel()`
- Status values are set in `src/backend/services/subscription_headroom_service.py` → `_probe()`: `ok` | `invalid_token` (401/403) | `rate_limited` (429) | `error` (transport, non-200, or 200 with no unified headers)
- `SubscriptionHeadroom.status` already reaches the frontend via `decorate_usage()`; no backend change should be needed
- The `LIMIT` badge itself is driven by `subscriptionSeverity()` → `crit`, which comes from `rate_limited_now` — the sibling backend defect (auth failures setting that flag) is filed separately; this issue stands on its own because a valid 429 on one subscription and a dead token on another must still read differently

trinity-ability:
## Summary

The Subscription pressure tile reported **"rate-limited"** for three states that are not throttling. On a live instance all four subscriptions carried the red `LIMIT` chip — three with `5h ≈$0 · 0 out` (no work attempted in five hours) and one *simultaneously* showing a healthy 9%/5% provider reading.

Two independent defects, one symptom:

- **#2352 (backend)** — `is_subscription_rate_limited()` counted every row in its 2h window regardless of `failure_kind`, so an auth failure (401/403: a dead, expired, or `.env`-shadowed token) set `rate_limited_now` on every surface. That is the conflation `failure_kind` was added for in #471; the display layer honoured the split (`rateLimitEventCount` reads the `rate_limit` kind alone) and the predicate never caught up. SUB-003 auto-switch then spreads one bad credential across every subscription in turn — the #2114 pattern, which is why *all four* rows lit up.
- **#2353 (frontend)** — `pressureHeadline()` tested `rate_limited_now` before it ever read `headroom.status`, so `invalid_token` — described in the row tooltip as *"the most actionable state the payload can carry"* — was reachable only by hovering, behind a chip claiming the opposite kind of problem.

Net effect for an operator: a credential problem was diagnosed as quota exhaustion, and the remedy on offer ("wait for the window to reset") was the wrong one.

## The predicate change is a split, not a filter

`is_subscription_rate_limited` had three consumers with **two incompatible meanings**:

| Consumer | Wants | Now uses |
|---|---|---|
| `decorate_usage` / `pressure_states` / `get_subscription_usage` | "is this throttled *right now*" | `is_subscription_rate_limited` (rate_limit kind only) |
| `select_best_alternative_subscription` | "did this fail recently for *any* reason" | `has_recent_subscription_failures` (kind-blind) |
| `select_subscription_for_new_agent` | same | `has_recent_subscription_failures` (kind-blind) |

Narrowing the shared predicate in place would have passed every display assertion and quietly started offering auth-failing subscriptions as switch candidates — an outage dressed as a remedy, and the #444 ping-pong class by a new route. The kind-blind half is byte-for-byte the pre-split behaviour, so **switch safety is unchanged**, and `test_auto_switch_skips_an_auth_failing_candidate` pins it.

`NULL failure_kind` (pre-#471 rows) is excluded from the narrowed predicate — unknown is never promoted to "429", matching what the frontend already refuses to do — and still counted by the kind-blind one, whose job is caution rather than attribution.

## Changes

**Backend**
- `db/subscriptions.py` — the predicate split, both over one private `_failure_event_count` so the NULL rule is a property of a single comparison; `get_failure_event_counts_by_subscription` returns `{total, by_kind}`, the same shape as its single-subscription sibling, in one query
- `services/subscription_headroom_service.py` — `pressure_states` emits `auth_failures_24h` beside the total
- `models.py` / `routers/agents.py` — the batch row gains `auth_failures_24h` + the probe's `token_status`. **Fields on an existing endpoint — no new route**, so ent#259's "zero new endpoints" property holds. Both default absent-safe
- `services/subscription_auto_switch.py` — three comments that named the old predicate now name the right one (comment-only)

**Frontend**
- `utils/subscriptionPressureTile.js` — new `auth` severity from a live `invalid_token` verdict, ranked **above** `crit` in the row sort (the row needing a person sorts above the row needing a wait); headline precedence `unavailable → token invalid → rate-limited → N× 429 → failures → no provider data → ok`
- `headroomStatus` is read **regardless of snapshot freshness** — the deliberate inverse of `windowReadings`. A *number* decays; "this credential was refused" does not become false by sitting still
- `utils/subscriptionPressure.js` — `pressureBadge` says `sub auth` instead of `sub 429s` for an auth-only history (strict equality against the total, so the tile and the chip cannot disagree about the same subscription), and degrades to the pre-#2352 wording on a payload without the new fields
- `SubscriptionPressureTile.vue` — `auth` chip + a `WARN_CHIP` map (three reasons now; a chained ternary silently makes the last one the default)

**Docs** — `architecture.md`, `dashboard-grid-view.md`, `subscription-usage-tracking.md`, `subscription-auto-switch.md` (new dated section for the split), `subscription-management.md`, `SUB-003-*.md`

## Test Plan

- [x] `pytest tests/unit/test_2352_subscription_failure_kind_predicates.py -v` — **19 passed**: both predicates, their disagreement, the auto-switch candidate-skip regression, NULL handling, window boundaries, by-kind batch counts
- [x] `pytest tests/unit -k "subscription or pressure or 471"` — **167 passed, 4 skipped**
- [x] `npx vitest run` (src/frontend) — **1087 passed**, incl. 27 new cases across every `headroom.status` × `rate_limited_now` combination
- [x] Full `pytest tests/unit` — 11943 passed; the only two failures are pre-existing local-environment noise (static AST guards walking a local `venv/` and the private enterprise submodule, neither present in CI)

One existing assertion changed meaning: `test_usage_carries_counters_and_2h_predicate` asserted `rate_limited_now is True` for an **auth** event. That assertion *was* the bug, pinned as behaviour; it now asserts `is False` with the reasoning inline. `_seed_event` in the ping-pong suite now writes `failure_kind` explicitly — it was leaving every row NULL, a shape the production writer never produces.

## Manual verification

On an instance with a dead subscription token, the row should read `AUTH · token invalid` and sort above a genuinely rate-limited neighbour; hovering still gives the full remedy. A subscription with real 429s is unchanged.

Fixes #2352
Fixes #2353

https://claude.ai/code/session_01DHEVtdzDkB6ybTK5PUykS8


vybe (approved):
Validated via /validate-pr: security scans clean, docs updated (architecture.md + 4 feature flows + SUB-003 requirement), named regression tests present for both linked bugs (test_2352_subscription_failure_kind_predicates.py, subscriptionTokenHonesty.spec.js) plus the switch-safety pin in test_subscription_auto_switch_pingpong.py, all CI green incl. e2e and pg-migrations. Closing keywords present — #2352/#2353 auto-promote on merge.

PR #2357, opened 2026-08-21
fix(auth): a login that issued no session must not look like one that did (#2322)

### LINKED ISSUE #2322: bug: /token returns HTTP 200 with an unusable access_token while a 2FA challenge is pending — silent break for password-grant machine clients
## Summary

When enterprise 2FA applies to an account (the user is enrolled, or a role policy requires enrollment), `POST /token` with a **correct** username/password still returns **HTTP 200** with an `access_token` + `token_type: "bearer"` — but that token carries the MFA challenge scope and is rejected with 401 by every endpoint except `/api/enterprise/2fa/login/*`. Any OAuth2 password-grant client (monitoring dashboards, ops tooling, scripts) reads `access_token` from a 200 and believes authentication succeeded; the failure only surfaces as unexplained 401s on later calls. The break is silent and presents as instance malfunction rather than an auth-flow change.

## Component

Backend / Auth (`routers/auth.py` login response + `dependencies.py` `MFA_CHALLENGE_SCOPE` rejection; enterprise `two_factor` module)

## Priority

P1

## Error

```
POST /token  →  200
{"access_token": "<jwt>", "challenge_token": "<jwt>", "token_type": "bearer",
 "mfa_required": true, "mfa_enrolled": true, "enrollment_required": false}

GET /api/version  (Authorization: Bearer <that access_token>)  →  401
{"detail": "Could not validate credentials"}
```

## Root Cause

The MFA-challenge response reuses the success shape of the OAuth2 password grant. A field named `access_token` with `token_type: "bearer"` in a 200 response is, by contract, a usable bearer token — here it is a challenge-scoped token that `get_current_user` deliberately rejects everywhere (the `MFA_CHALLENGE_SCOPE` check). Verified empirically on an enrolled admin account; per the router code the same shape is returned for a policy-required-but-unenrolled user (forced-enrollment flow), so machine clients break the moment the policy is enabled, before anyone even enrolls.

## Reproduction Steps

1. On an enterprise-entitled instance, enroll TOTP 2FA on an account (or enable "Require 2FA for admins")
2. `POST /token` with that account's correct credentials → observe 200 + `access_token`
3. Use the returned `access_token` on any authenticated endpoint (e.g. `GET /api/version`) → 401

## Suggested Fix

Signal the pending second factor as an **error for the grant**, per RFC 6749 error semantics: return 401/403 with `{"detail": "mfa_required", "challenge_token": "..."}` and no `access_token` field. If the frontend depends on the current 200 shape, keep it only on a UI-specific login endpoint and fix `/token` (the OAuth2-shaped surface) to fail loudly. At minimum, remove/null the `access_token` field in the challenge response so clients fail at login time, not downstream.

## Environment

- Trinity version: v0.8.5 (`19e41fff`), enterprise edition, PostgreSQL backend

## Related

- `src/backend/dependencies.py` (MFA challenge-scope rejection)
- `enterprise/backend/two_factor/router.py` (login verify / forced enrollment flow)

obasilakis:
## Summary

When enterprise 2FA defers a login, `POST /token` answered **HTTP 200** carrying `access_token: null` **and** `token_type: "bearer"` — a password grant that issued no session, describing itself as a bearer grant. Any client reading the token field without checking it stored nothing, reported success, and produced unexplained 401s on its next call, far from the cause.

Trinity ships three such clients: the MCP server (checked `response.ok`, never the payload) and both CLI login paths. All three are fixed here.

## The contract now

A password grant that issued no session is an **error for the grant** (RFC 6749 §5.2). Three outcomes, three status codes:

| Status | Meaning | Body |
|---|---|---|
| 200 | Session issued | `{access_token, token_type}` |
| **403** | Credentials correct, second factor required | `{detail: "mfa_required", mfa_required: true, mfa_enrolled, enrollment_required, challenge_token}` |
| 401 | Credentials rejected | `{detail: "Incorrect username or password"}` — never a challenge |

**403 and not 401**, though the issue permits either — 401 is actively wrong here, and both first-party clients are the proof:

- `trinity_cli/client.py::_handle_response` treats 401 as *"your session expired, run `trinity login`"* and hard-exits. During login that message is nonsense.
- `main.js` registers a **global** axios 401 interceptor that calls `authStore.logout()` and redirects to `/login`. Safe today only because of a `currentPath !== '/login'` guard — wiring a mid-flight login refusal into "your session died" machinery is a trap for whoever next moves the login form.

Semantically the credentials were *correct*: authentication is incomplete, not rejected, which is why OAuth providers answer `mfa_required` with 403. Pinned by `test_the_403_is_not_a_401` so a later tidy-up can't silently re-arm either behaviour. Clients branch on the **boolean** `mfa_required`, never on parsing `detail`.

## Root-cause correction (posted on the issue)

The issue's stated mechanism — a challenge-scoped token that `get_current_user` rejects — is **not** what fires. `mfa_gate.gate_login` never set `access_token`; `Token` defaulted it to `None` and the route had no `response_model_exclude_none`, so the field was **null** and `MFA_CHALLENGE_SCOPE` was never reached. A naive client never presents `challenge_token` as a bearer.

The real chain: `resp.json()["access_token"]` → `None` → `Authorization: Bearer None` → 401. Verified empirically against the real router, and against `v0.8.5` (the reporting instance's version), which is byte-identical in both files. The issue's stated *minimum* fix — "remove/null the `access_token` field" — was already shipped; what remained was the contract, which is what this PR delivers.

## Changes

- **`models.py` / `routers/auth.py`** — the deferred case returns a `JSONResponse` 403 so the challenge fields sit beside `detail` (the shape the issue specifies), with the 403 declared in the OpenAPI. `Token.token_type` becomes `Optional` and the success path serializes with `response_model_exclude_none=True`, which fixes the **second half of the bug**: every *successful* login, OSS-only builds included, used to carry all four 2FA fields as `null`, contradicting `Token`'s own docstring.
- **`stores/auth.js`** — the challenge now arrives on the error path; read off `error.response.data`.
- **`cli/`** — `TrinityAPIError` carries the parsed `body`, and `_exit_on_login_error` maps a 403 + `mfa_required` to an actionable message across all four login handlers. `_require_access_token` stays as a belt for any future 2xx-without-a-session.
- **`mcp-server/src/client.ts`** — parses the body **before** the status check; otherwise the one failure with a specific, actionable cause reports a bare "Forbidden". `TokenResponse` fields are optional so TypeScript enforces the check at the next call site.
- **Docs** — both feature flows (neither mentioned 2FA at all; both documented the old contract), the user-facing API reference (now shows the three-status table and a snippet that checks the status), and the two getting-started snippets that were teaching a blind `['access_token']` read.

### One deliberate asymmetry

`/api/auth/email/verify` **keeps its 200**, omitting `access_token` as it always has. The issue scopes the RFC change to *"`/token` (the OAuth2-shaped surface)"*, and that route is a bespoke JSON endpoint, not a grant. The invariant both satisfy is **no response ever presents a session that was not issued** — via the mechanism each contract calls for. Pinned on both sides so the asymmetry is deliberate rather than drift.

## Worth noting for triage

The **policy arm** is the worse trigger and isn't what the issue leads with. `gate_decision` returns `required=True` for every admin/creator-role account the moment the role policy is enabled — so flipping the policy defers every unattended admin/creator credential **instantly, before anyone has enrolled**. One toggle, no user action, and the rollout hasn't started.

## Test Plan

28 tests across three surfaces, **each guard mutation-verified** — removed, watched it go red, restored:

| Suite | Tests | Guard removed → |
|---|---|---|
| `tests/unit/test_2322_mfa_challenge_response.py` | 13 | revert to 200 body → 6 fail · 403→401 → 6 fail · drop `exclude_none` → 6 fail · `token_type` non-Optional → 4 fail · `response_model` on the email route → 2 fail |
| `tests/test_cli_admin_login.py` (extended) | 10 (3 new) | CLI blind read → 3 fail |
| `src/mcp-server/src/auth-challenge.test.ts` | 5 | MCP guard → 2 fail |

- [x] Backend 13 passed. Edition-agnostic — the provider registers through the OSS `mfa_gate` seam, so they run with **no enterprise submodule present**. Both trigger arms covered (enrolled; policy-required-but-not-enrolled), plus *wrong password stays 401 and carries no challenge*.
- [x] CLI 10 passed (3 new + 7 existing, unmodified).
- [x] MCP 269 passed, `tsc --noEmit` clean.
- [x] **Full unit suite vs base `76615f26`: zero regressions.** Base 23F/12042P, head 21F/12057P; `comm` empty on the HEAD side. The pre-existing failures are IPv6-mapped-address and capacity-meter tests on this machine; two of them didn't reproduce this run (flaky, not fixed here).
- [x] `lint_root_test_placement` + `lint_sys_modules` clean; `enterprise-docs-guard` PCRE run locally — clean.

**Test placement note:** the CLI tests went into the existing `test_cli_admin_login.py` rather than a new file. A new root-level test is rejected by `lint_root_test_placement.py` (#1895 — root tests are collected by no per-PR job), but `tests/unit/` is also wrong: `requirements-test.txt` deliberately excludes `trinity_cli`, and the #660 contract is that a bare `pytest unit/` runs green after installing only that file. `test_cli_admin_login.py` is already grandfathered in the baseline. Honest consequence: like every existing CLI test, these run in the nightly integration job, not the per-PR gate — pre-existing boundary, tracked by #1896.

## Reviews run

- **`/review`** — 0 critical, 3 informational, all fixed in `05e84bfc`: a doc example that echoed the response body (which carries the short-lived `challenge_token`), two more docs teaching the blind read, and the email route being cited as the reference behaviour while pinned by nothing.
- **`/cso --diff`** — 0 findings. Report at `docs/security-reports/cso-diff-2026-08-21-2322-mfa-challenge-shape.{md,json}`, with an addendum recording R1 as closed by this PR.
- **`/validate-pr`** — APPROVE. One warning, fixed: this description had gone stale against the 403 change.
- **`/sync-feature-flows`** — run late, and it earned its place. It found **nine** more copies of the blind `access_token` read that `/review`'s sweep missed, because that sweep was scoped to `src docs/user-docs scripts` while the defect sat one directory over in `docs/memory/` and `docs/testing/`. The worst of them: `feature-flows/mcp-orchestration.md` transcribed the **old MCP client verbatim** — `this.token = data.access_token;`, the exact line this PR replaces. Also updated `role-model.md`'s test plan, `AGENT_NETWORK_DEMO.md`, `user-docs/quick-start.md`, `user-docs/guides/deploying/monitoring.md`, three `docs/testing/` plans, and `AGENTS.md`, plus the feature-flows index changelog row.

  Three copies are **deliberately left**: `.github/workflows/{build-without-submodule,container-security}.yml` and `tests/test_ip_rate_limit_fix.py` authenticate against throwaway instances we control where 2FA cannot be enabled. Docs teach patterns humans copy; those don't — and editing CI workflows on an auth PR widens the blast radius for nothing.

## Still deferred to @vybe

Whether a role policy should sweep **machine credentials** into a human-interactive second factor at all. That's the root of the policy-arm trigger and a new capability rather than a fix — Trinity has no service-account principal, and any exemption is a second-factor bypass by construction. The supported answer today is that automation uses an MCP API key (`trinity_mcp_*`), which isn't subject to the flow; the docs in this PR now say so.

Fixes #2322


vybe (approved):
Validated after rebase onto `dev`. **Fixes #2322** (closing keyword present). All 21 non-skipped checks pass; regression diff clean across all six seeds (base 12130 → head 12143, 0 failures either side, **No new failures**). `e2e` green, and the PR is correctly unlabelled-`ui`-free — it does touch `src/frontend/src/stores/auth.js`, and `frontend-e2e` ran anyway.

**Substance.** The fix is right and, more importantly, it is applied at all three surfaces that were reading the field without checking it — Invariant #13's whole point. A password grant that issues no session now answers **403** rather than `200 {access_token: null, token_type: "bearer"}`, and the MCP server (`client.ts`, which checked `response.ok` and nothing else), both CLI login paths (`trinity_cli/client.py`, `commands/auth.py`) and the frontend store all learn about it. A one-sided fix here would have left the exact silent-401-far-from-the-cause failure the issue describes, just from a different client.

`test_2322_mfa_challenge_response.py` is the named regression test; `test_cli_admin_login.py` covers the CLI half; `auth-challenge.test.ts` the MCP half. A CSO `--diff` audit rode along (`docs/security-reports/cso-diff-2026-08-21-2322-mfa-challenge-shape`, 0 findings) with an addendum recording R1 closed — appropriate for an auth-path change, and the review pipeline asks for it on this PR type.

**Rebase note.** Conflicted on `docs/memory/learnings.md` after #2333 landed — a pure append-vs-append at the same insertion point. Resolved keep-both, ordered by entry date (the file is chronological ascending); verified no duplicate headings and no surviving markers. `tests/registry.json` came through clean and still parses at 200 entries.

Security scan clean — the two secret-shaped hits are test fixtures (`eyJ-challenge`, `eyJ-real`). No new env var, no new backend package, no host paths or mode flips.

PR #2359, opened 2026-08-21
fix(workspace): a failed turn says so, and Retry follows the billing evidence (#2320)

### LINKED ISSUE #2320: bug(workspace): failed turns are invisible to the client — fast failures misreported as "lost track", backend error discarded, Retry suppressed where it is safe
## What happened (observed on a live instance, evidence in the execution row)

A Workspace message created an execution that **failed 4.6 seconds later** with HTTP 429 — the agent's Claude subscription hit its weekly limit, and the SUB-003 auto-switcher correctly reported no viable alternative (the whole pool was exhausted). The backend composed a precise, actionable error (limit + reset time + remediation), wrote it to `schedule_executions.error`, logged it, and ended the turn cleanly.

The Workspace client showed:

> Still no reply — we've lost track of this turn. It may still finish; check the conversation shortly.

Every clause of that is wrong for this case: the turn **did** finish (failed), it will **never** produce a reply, and the composed error was discarded. Retrying "as-is" would 429 again, but the user has no way to know that.

## Mechanism

The client learns a turn's outcome exactly two ways (`PortalConversation.vue::awaitPersistedReply`):

1. a **new assistant message** in persisted portal history, or
2. the **in-flight marker** (`in_flight_execution_id`) still being set.

A failed turn produces **neither**: no assistant/error message is persisted to `enterprise_portal_messages`, and `start_portal_turn._run`'s `finally` clears the marker on every exit path (`client_portal/service.py:1843-1846`). The failure record lives only in `schedule_executions.error` (`_fail_unstarted_execution`, `service.py:1702`), which no portal read surface exposes — the history response carries `in_flight_execution_id` / `in_flight_wait_budget_seconds` but no last-turn outcome.

So after the server reports "nothing running" for `REPLY_IDLE_GIVE_UP_MS` (6s), the client returns `{ lost: true, idle: true }` (`PortalConversation.vue:899`) — and `deliver()` (line 754) renders the #2133 "lost track" message designed for turns that *ran and got lost*. The `idle: true` flag that distinguishes "server says nothing is running" from "wait budget exhausted" is **returned but never read** — even the distinction the client already has goes unrendered.

Net effect, on every fast-failing Workspace turn (subscription/auth 429·503, agent stopped mid-dispatch, resume-lock contention, any pre-start `ClientPortalError`):

- misreported as "lost, may still finish" when the server knows it failed and why
- the actionable error the backend composed is discarded
- **Retry is suppressed on the one path where retrying is safe** — the #2120/#2133 no-Retry rule exists for turns that may have run and billed; a turn that failed **before start** was never billed

## Constraints on the fix

- `schedule_executions.error` must NOT be exposed verbatim to portal clients — it is free-form runtime text on a client-facing surface (the #2162/G-04 discipline). Persist/expose a **typed, sanitized** failure: a category (`capacity`, `auth`, `agent_unavailable`, `busy`, `internal`) plus a client-safe sentence; the full detail stays on the operator surfaces (Executions) where it already lands. Subscription-limit remediation ("add an API key / register a subscription") is operator guidance, not client guidance.
- Retry may only be offered when the failure is provably pre-start/unbilled (the typed record can carry that bit). The lost/idle paths keep no-Retry.

## Suggested shape

1. On the `_run` failure paths, persist a failure record the client can see — either a typed error message row in portal history, or a `last_turn_outcome: {execution_id, status, category, message}` field on the history response (the client already polls it).
2. Client: render "This turn failed: <sanitized reason>" instead of the lost-track message when an outcome is present; offer Retry iff the record says pre-start/unbilled.
3. Minimal interim step (frontend-only): read the existing `idle` flag and word the two give-ups differently — "The agent did not reply" (server says nothing running) vs today's "lost track" (budget exhausted while the marker lived).

## Acceptance criteria

- [ ] A turn that fails before/at start renders its failure (sanitized category + client-safe message) in the thread instead of "lost track"
- [ ] Raw `schedule_executions.error` text never reaches a portal client verbatim
- [ ] Retry appears only for failures recorded as never-started/unbilled; the lost/idle paths keep suppressing it (existing #2120/#2133/#2214 pins hold)
- [ ] The `idle` vs budget-exhausted give-ups are worded distinctly
- [ ] Regression test naming this issue covers the fast-fail path (marker cleared + no persisted reply + failed execution row)

Related: #2120, #2133, #2214, #2319 (the *slow* twin: post-lost recovery for turns that really ran), abilityai/trinity-enterprise#286, abilityai/trinity-enterprise#358

obasilakis:
## Summary

A Workspace turn that fails before or at start persists **no assistant message** and clears its in-flight marker on every exit path. The client learns an outcome exactly two ways — a new assistant row, or the marker still being set — so a fast failure produced neither, and after the 6s idle give-up it rendered the #2133 *"we've lost track of this turn — it may still finish"* copy for a turn the backend had diagnosed precisely and written to `schedule_executions.error`. Every clause of that message was false, and Retry was suppressed on the one path where re-sending is safe.

Two things turned out to be worse than the issue describes:

- The `idle` flag that distinguishes "the server says nothing is running" from "we ran out of budget" was returned and never read.
- `deliver()`'s retryable `if (!data)` branch — which PR #2150's body describes as preserved (*"a genuine no-answer keeps its existing message and its Retry"*) — is **unreachable**. `awaitPersistedReply` has three returns and null is not one, so `if (data?.lost)` swallowed the idle case. That is why Retry was suppressed on the path #2150 believed was retryable.
- `reattach()` was the silent surface: it checked only for a reply, so a failed or lost turn rendered **nothing at all** — no message, no Retry, the spinner just stopped. Refreshing mid-turn showed less than staying put. Not mentioned in the issue.

## Changes

**The two bits live on the exception, decided at the raise site.** `ClientPortalError(status, detail, *, category, retryable)`. Not inferred downstream: `_fail_unstarted_execution` is reached from the pre-start branch **and** the generic `except Exception`, which can fire after `execute_task` already returned — so "was this billed" is not a property of the row being written. `retryable` defaults **False**.

| Raise site | category | retryable |
|---|---|---|
| roster miss / stopped / containerless | `agent_unavailable` | ✗ unbilled, but ent#286 settled that retrying cannot work |
| `ResumeLockBusy` | `busy` | **✓** never reached the agent |
| `CAPACITY` | `capacity` | **✓** admission refused; the queue drains |
| `AUTH` / `BILLING` | `auth` | ✗ retry re-fails |
| `TIMEOUT` | `timeout` | ✗ ran to the bound |
| generic turn failure | `agent_error` | ✗ ran |
| uncaught crash | `internal` | ✗ fixed sentence; raw text stays operator-only |

**Classification reads `TaskExecutionResult.error_code`** instead of substring-matching the human-readable error, keeping the substring tests as the `None`-code fallback (additive — nothing that classified before stops classifying). `AUTH`/`BILLING` had no branch at all and fell through to the generic 502: that is the subscription-limit case this issue was reported from. `_error_code_name` reads `.name` rather than comparing members, because `TaskExecutionErrorCode` is `@dataclass`-decorated and `AUTH == TIMEOUT` is **True** (verified) — the #1085 footgun.

**The verdict rides Redis beside the marker it is the terminal half of** — `portal_turn_outcome:{session_id}`, TTL 900s, written in `_run`'s except branches **before** the `finally` clears the marker. That ordering is the contract: the client's give-up timer starts when the marker vanishes, so an outcome written after it races a 6s window. Cleared at dispatch (turn N+1 never inherits turn N's verdict) and on success. Redis down ⇒ no outcome ⇒ the pre-#2320 message, never worse. Surfaced as `PortalHistory.last_turn_outcome`, **declared** on the model — the route's `response_model` strips undeclared keys, so a service-layer-only change would be a no-op.

**No schema change.** An error message row in `enterprise_portal_messages` was the obvious shape and is poison: `_format_history_context` replays any non-`user` role to the agent **as its own words**, and `_persist_user_turn`'s dedupe reads `recent[-1].role == "user"`, so Retry would duplicate the user message — breaking a #2120 pin that has a test.

**Client** believes a verdict only for the turn it is awaiting (`outcome.execution_id === executionId`), words the two give-ups distinctly, and offers Retry iff the verdict says nothing reached the agent. `markLastUserTurnFailed` marks only the thread's **unanswered tail** — two raise sites record a verdict without persisting a user row, and a backwards walk would pin the failure onto an earlier, answered turn.

## Acceptance criteria

- [x] A turn that fails before/at start renders its failure instead of "lost track"
- [x] Raw `schedule_executions.error` never reaches a portal client verbatim (fixed sentence for the uncategorised path; raw text to log + execution row only)
- [x] Retry appears only for failures recorded as never-started/unbilled; lost/idle keep suppressing it
- [x] `idle` vs budget-exhausted worded distinctly
- [x] Regression test naming this issue covers the fast-fail path

## Test Plan

- [x] `cd tests && python3 -m pytest unit/test_2320_portal_failed_turn_visibility.py -q` → **42 passed**
- [x] `cd src/frontend && npx vitest run tests/unit/portalFailedTurn.spec.js` → **25 passed**
- [x] Portal suites (`ent286`, `2133`, `ent358`, `2196`, `2320`) → **137 passed**, no regressions
- [x] Full frontend suite → **53 files, 1112 passed**
- [x] Mutation-verified: 10 backend + 11 frontend mutations, each caught; restores byte-identical
- [x] **Reproduced and fixed on a live instance.** Held the resume lock, dispatched a Workspace turn: before, `in_flight=None` + no reply + lost-track copy. After:

```json
{"execution_id":"qT4FcFCfiwgn0asbQQC0Nw","category":"busy",
 "message":"This conversation is already handling a message. Please try again shortly.",
 "retryable":true}
```

Next turn: verdict cleared at dispatch, `assistant "FIXED"`, outcome stays null.

Note: the frontend spec is the **first** to exercise `PortalConversation.vue`, and there was no executable coverage of the no-Retry rule anywhere before this — only a docstring at `test_2133_bounded_reply_poll.py:15`. There is no component-mount harness in the project (`@vue/test-utils` is not a dependency and vitest runs `environment: 'node'`), so rather than re-implement the decision expressions, the spec extracts the shipped ones and runs them.

Two entries added to `docs/memory/learnings.md`: the dead-caller-branch class, and applying a verdict to "the last matching row" instead of the row actually waiting.

Fixes #2320


vybe (approved):
Validated. Fixes #2320 (closing keyword present). Named regression test `test_2320_portal_failed_turn_visibility.py` + frontend `portalFailedTurn.spec.js`; feature flow `workspace-absorbs-session.md` updated. Security scan clean (the two secret-shaped hits are obvious test fixtures). No new backend package, no new env var, no host paths or mode flips.

CI note for the record: `pytest (base, seed 99999)` shows red but is a **runner timeout on the base side** — `junit-base-99999.xml` never materialised. All three head seeds are fully green (12145 passed / 0 failed) and `regression diff` reports **No new failures** on the surviving seeds. Not a regression signal.

vybe (approved):
Validated after three rebases onto `dev`. **Fixes #2320** (closing keyword present). All checks pass on a fully clean board this time — no degraded seed: regression diff across all six seeds (base 12195 → head 12237, **0 failures either side**, No new failures), `e2e` green (6m55s).

Worth recording since it was red twice on the way here: the earlier `pytest (base, seed …)` failures were **cancelled** base-side jobs whose `junit-base-*.xml` never materialised — runner timeouts on `dev`, not this PR. The head side was green all three times. This run confirms it against a full six-seed board.

**Substance.** The PR fixes a real dead-branch: `awaitPersistedReply` has exactly three returns and never returns null, so `deliver()`'s `if (!data)` branch and its retryable throw were unreachable from the day #2150 wrote them — while `if (data?.lost)` sat above and swallowed the `idle` case into the pessimistic "we've lost track of this turn" copy. Every failed Workspace turn for months was reported as possibly-still-running with Retry suppressed, including turns the backend had diagnosed precisely and written to `schedule_executions.error`. The `idle` discriminator was computed, returned, and read by nobody.

Good sign for the fix's own quality: the second learnings entry records a defect found in `/review` **of this PR** — `markLastUserTurnFailed` walking backwards to the last user row, which for the two raise sites that fire before `_persist_user_turn` would have pinned the failure onto an *earlier, already-answered* turn and offered a Retry that re-sends it. Not live, but correct only by a property of the current call graph. Fixing that before merge rather than shipping on "unreachable today" is the right call, and the same family as the #1804 lookup-vs-write rule.

Tests: `test_2320_portal_failed_turn_visibility.py` (named regression, 931 lines) and `portalFailedTurn.spec.js` (443). `workspace-absorbs-session.md` flow updated.

**Rebase note.** Three rebases — after #2333, #2357 and #2356 in turn. The first two were pure `learnings.md` append conflicts. The third overlapped #2356 on three real source files (`client_portal/service.py`, `PortalConversation.vue`, `stores/clientPortal.js`); git auto-merged all three, and I verified rather than assumed — ent#428's ask code is intact in each, this PR's own diff is unchanged at 8 files / 1808 insertions, and `learnings.md` ordering breaks still match `origin/dev`'s exactly with no duplicate headings.

Security scan clean (`portal-token`, `SUPER-SECRET-INTERNAL-abc123` are test fixtures). No new env var, no new backend package, no host paths or mode flips.

Issue #2377, opened 2026-08-21
bug: Operations feed never shows "your response was not recorded" after a respond race — fetchItems() wipes the copy and the UI only renders it as a refresh-failure detail


webmixgamer:
## Summary

On the desktop Operations feed, when `POST /api/operator-queue/{id}/respond` is refused because the item is no longer pending (409 lost race, #1017; or 400 already terminal), `stores/operatorQueue.js::respondToItem` sets `error` to the "your response was not recorded" copy and calls `fetchItems()` — whose first line sets `error = null`. The copy is therefore wiped synchronously and is effectively never visible. Even when `error` does survive, `Operations.vue` renders it only as the *detail* line under the "Couldn't refresh the queue — showing the last items that loaded" banner, i.e. under a headline about refreshing, not about an unrecorded answer. The operator's card simply disappears and nothing says their answer was dropped.

## Context

Found while pinning the shared respond payload for #2370 (the `/m` fix). `/m` routes the same refusal to a persistent page-level notice and scrolls it into view; desktop has no equivalent surface. The copy itself is now the shared, attribution-free `QUEUE_RESPONSE_NOT_RECORDED` in `src/frontend/src/utils/operatorQueue.js`; what is missing is a channel that survives the refetch and a headline that names the event. The behavioural store test in `src/frontend/tests/unit/operatorQueueResponse.spec.js` deliberately does NOT pin the wipe (it asserts only the refetch) so this stays fixable.

## Acceptance Criteria

- [ ] After a 409/400 refusal the Operations feed shows a persistent, dismissible notice with the `QUEUE_RESPONSE_NOT_RECORDED` copy (an `InlineError` with its own headline — not the detail of the refresh-failed banner), and it survives the immediate refetch and the 10 s poll until dismissed
- [ ] A transient refusal does not masquerade as a refresh failure (the refresh-failed banner appears only when a fetch actually failed)
- [ ] The behavioural store test asserts the notice is observable after `respondToItem` on a 409 and a 400
- [ ] No change to the respond payload or the agent write-back

## Technical Notes

- `src/frontend/src/stores/operatorQueue.js` — `respondToItem` catch branch, `fetchItems` (`error.value = null` at its start), `error` ref
- `src/frontend/src/views/Operations.vue` — the `queueRefreshFailed` `InlineError` that renders `operatorQueueStore.error` as `:detail`
[...]

Issue #2376, opened 2026-08-21
bug: POST /api/operator-queue/{id}/respond accepts any string as an approval decision — no layer checks response ∈ options


webmixgamer:
## Summary

`OperatorResponse.response` is a bare `str`; the router, the DB write and the agent write-back pass it through verbatim, and no layer checks that an **approval** item's recorded decision is one of the options the agent offered. #2370 shipped `response: "approved"` against `options: ["Approve", "Deny"]` for five months without a single 4xx — a server-side membership check would have turned the first tap into a named 422.

## Context

The operator queue is the approval channel for irreversible agent actions (#1402 poison-park, ent#329 respond→resume, the TARGET_ARCHITECTURE v2 human-gate). Four producers exist today (desktop store, `/m`, the Workspace asks panel, the MCP tool) and the next one — a channel button, an ops agent — can re-ship the class unless the sink refuses. Defence in depth at the boundary (Product Quality Bar #6: a named validation error, never a silently broken state). Surfaced by the #2370 plan review.

## Acceptance Criteria

- [ ] For an item with `type == 'approval'` and real options (a non-empty list that is not the size-cap marker `_OPTIONS_DROPPED_MARKER`), a `response` not in `options` is rejected with a named 422 (e.g. `response_not_an_offered_option`) that lists the offered options; exact string match (options are agent-authored)
- [ ] Question items, alert items, approval items without usable options, and `acknowledged` are unaffected
- [ ] The check lives in ONE place reached by both `POST /api/operator-queue/{id}/respond` and the Workspace asks answer path (`src/backend/client_portal/asks/service.py`) — a shared validator at the service/db chokepoint, not a copy per router
- [ ] MCP `respond_to_operator_queue` surfaces the 422 as a structured error (no silent success)
- [ ] Tests: accept / reject / exempt cases on both entry points, including the marker-list case

## Technical Notes

- `src/backend/models.py` (`OperatorResponse`), `src/backend/routers/operator_queue.py` (respond), `src/backend/db/operator_queue.py` (`options` stored as JSON and loaded back in the row mapper), `src/backend/services/operator_queue_service.py` (`_OPTIONS_DROPPED_MARKER`), `src/backend/client_portal/asks/service.py`, `src/mcp-server/src/tools/operator_queue.ts`
- Sibling of the Workspace asks issue filed alongside; both are follow-ups to #2370

Issue #2375, opened 2026-08-21
bug: Workspace asks panel sends a typed answer as response_text (the agent reads response: "") and answers approval asks one-tap — same class as #2370


webmixgamer:
## Summary

On the Workspace asks surface (OSS core since ent#428) a client's typed answer to a **question** ask is posted as `{ response: null, response_text: "<answer>" }`; the backend writes `response = response or ""`, so the agent's queue file receives `response: ""` with the answer in `response_text`. The agent protocol (platform prompt + MCP tool doc) tells agents to read `response` — a Workspace client answering a question therefore hands the agent an **empty answer**. The same panel answers **approval** asks with a one-tap option button (no note, no explicit submit) and carries a third type-label set. This is the same bug class #2370 fixes on `/m`: a hand-built respond payload with no shared builder.

## Context

Found by the independent plan review of #2370. That PR introduces `src/frontend/src/utils/operatorQueue.js` (`queueResponseBody` / `buildQueueResponse` / `queueTypeLabel`) as the single home of the respond payload, type labels and controls-kind rule for the desktop store and `/m`; the Workspace asks panel is the remaining frontend producer that does not go through it. It was split out of #2370 rather than folded in because it posts to a different endpoint with its own model — `POST /api/enterprise/client-portal/asks/{id}/answer`, `WorkspaceAskAnswer` (both fields optional, at least one required) — and has its own test suites.

## Acceptance Criteria

- [ ] A typed answer to a question ask is sent as `response` (trimmed), never as `response_text`; the agent's queue file shows it under `response`
- [ ] Approval asks: select an option → optional note → explicit submit (no one-tap irreversible answer), mirroring desktop `QueueCard` and `/m` after #2370; a note travels as `response_text`
- [ ] The panel renders the shared type labels (`queueTypeLabel` from `src/frontend/src/utils/operatorQueue.js`) — one label set across desktop, `/m`, Workspace
- [ ] `WorkspaceAskAnswer` / `asks/service.py` no longer coerce a missing `response` to `""` silently: either require `response` (named 422 when absent) or document the note-only answer as a distinct, intended case
- [ ] Regression tests pin the payload for question, approval (+note) and alert asks: `src/frontend/tests/unit/workspaceAsks.spec.js` and `tests/unit/test_ent428_workspace_asks.py`
- [ ] Honest failure preserved (the existing per-card error rendering)

## Technical Notes

[...]

PR #2378, opened 2026-08-21
fix(ui): /m approval card sends the tapped option as the decision — select → note → explicit Send, alert Got it, real type labels; one respond-payload builder shared with desktop (#2370)

### LINKED ISSUE #2370: bug: /m mobile admin approval card records every option tap as response:"approved" (a Deny becomes an approval) and lacks the desktop note field
## Summary

On the mobile admin page (`/m`, `src/frontend/src/views/MobileAdmin.vue`) every tap on an operator-queue option button POSTs a hard-coded `response: "approved"` and puts the tapped option into `response_text`. Tapping **Deny** is therefore recorded as an approval with the word "Deny" in the note — the opposite of what the operator did. The same card lacks the desktop `QueueCard`'s note field and explicit submit step (a one-tap, no-confirm answer on an irreversible action), and renders a blank type line (`item.request_type`; the API field is `type`).

## Context

Found while eyeballing #1927 on `/m` (that PR fixed only the card's loading gates / polling and the `{items,count}` response-shape bug that had hidden the queue tab entirely; this bug is pre-existing and independent).

Desktop `src/frontend/src/components/operator/QueueCard.vue` selects an option, takes an optional note and submits `{ response: <option>, response_text: <note> }` (`submitApproval` → `store.respondToItem(id, selectedOption, responseText)`); a free-text answer is sent as `response`. Mobile `MobileAdmin.vue::respondToQueueItem(id, response)` POSTs `{ response: 'approved', response_text: response }` for **every** path — option taps and typed answers alike.

The operator queue is the approval channel for irreversible agent actions (#1402 poison-park, ent#329 respond→resume), so a Deny recorded as an approval is an integrity bug, not a cosmetic one.

## Acceptance Criteria

- [ ] A tapped option is sent as `response` (the option string) with any note in `response_text` — the same payload desktop `QueueCard` sends; a typed free-text answer is sent as `response`, never as `response_text`
- [ ] Approval items on `/m` reach desktop parity: select an option → optional note → explicit submit (no one-tap irreversible answer); question items keep the text input + Send
- [ ] The card's type line renders the item's `type` (question / approval / alert) instead of a blank
- [ ] A regression test pins the payload shape for both item kinds — a vitest spec over a pure `buildQueueResponse(...)` helper and/or a Playwright `@smoke` test that fulfils `/api/operator-queue` and captures the `POST …/respond` body
- [ ] Honest failure is preserved: a failed respond still surfaces the existing `reportActionFailure` banner
- [ ] Design-system principle 19 for the new submit step: a destructive/irreversible answer restates the consequence and focuses the safe action

## Technical Notes

- `src/frontend/src/views/MobileAdmin.vue` — the `ops-options` buttons (respond on tap), `respondToQueueItem` (hard-coded `response: 'approved'`), the `item.request_type` line
- `src/frontend/src/components/operator/QueueCard.vue` — `submitApproval` / `submitAnswer` (the shape to mirror); `src/frontend/src/stores/operatorQueue.js::respondToItem`
- `src/backend/routers/operator_queue.py` — `POST /api/operator-queue/{id}/respond` accepts `response` + `response_text` and writes both back to the agent's queue file, so the agent reads `response`

webmixgamer:
## Summary

- **The bug**: the `/m` mobile admin's queue card POSTed `{ response: 'approved', response_text: <tapped option> }` for **every** answer — a Deny was recorded as an approval, a typed answer became a note, the card answered on **one tap** with no note and no explicit submit (desktop `QueueCard` is select → optional note → Send), and the type line read a nonexistent `request_type` (blank). The agent reads `response` as the decision (sync write-back + the ent#329 resume framing), so the field is load-bearing. Live artifact: the two cards answered on `/m` during the #1927 eyeball sit in the agent's queue file as `response='approved'`, `response_text='Approve'`.
- **The fix**: `utils/operatorQueue.js` is now the ONE builder of the respond payload, shared by the desktop store and `/m` (the bug class was a second hand-built producer). `/m` reaches desktop parity: select (`aria-pressed`, inset ring) → restated consequence → optional note (no Enter-submit) → `Cancel` (focused on reveal, p19) + `Send: <option>`; question = text + Send; alert = `Got it`; title line; real type labels (`Needs approval` / `Question` / `Heads up`, shared with `QueueCard`).
- **Honest state**: success drops the card locally before the refetch (`fetchQueue` swallows its own errors) and a monotonic guard stops a stale poll from resurrecting an answered card; a 5xx/transport failure renders an `InlineError` **inside the card** (p18) with the selection + note kept for a retry; a 409/400/404 "no longer pending" refusal drops the card and shows a persistent page-level notice, scrolled into view. Per-card state is keyed by item id, pruned on every poll, cleared on logout. Colour-free CSS only.

## Changes

- `src/frontend/src/utils/operatorQueue.js` (new) — `queueResponseBody` / `buildQueueResponse` / `queueResponseKind` (by item **type**: approval-with-options → buttons, alert → Got it, else text answer) / `optionsOf` (the DB stores whatever JSON the agent wrote) / `queueTypeLabel` / `QUEUE_RESPONSE_NOT_RECORDED` + `respondRefusedAsNotPending` (409/400/404)
- `src/frontend/src/stores/operatorQueue.js` — body via the shared builder (note trimmed, empty → `null`; a blank decision never reaches the wire); 400/404 join 409 with the shared attribution-free copy
- `src/frontend/src/components/operator/QueueCard.vue` — imports the shared type label
- `src/frontend/src/views/MobileAdmin.vue` — the card rebuild above; `.ops-card-type` off gray-500 (the contract's dark-ink floor); `data-testid` hooks
- `src/frontend/tests/unit/operatorQueueResponse.spec.js` (new) — decision table, behavioural store test (`vi.mock('axios')`), source-wiring + residue guards
- `src/frontend/e2e/mobile-admin-queue-respond.spec.js` (new, `@smoke`) — one method-aware `page.route` serves the GET and **captures the real POST body**; "no POST on tap" is proved by order (tap → form → Send → exactly one captured body)
- Docs: `feature-flows/mobile-admin-pwa.md`, `feature-flows/operating-room.md`, `requirements/public-access.md` §27.3, flows index, 3 `learnings.md` entries, CSO diff report

### Review round (`253c45cd`) — three findings from @dolho, all in code this PR introduced

- `stores/operatorQueue.js` — the not-recorded copy was assigned and then destroyed by an **un-awaited** `fetchItems()`, whose synchronous prologue nulls `error`. Harmless while 409 was the only refused status; but widening `respondRefusedAsNotPending` to **400/404** moved two statuses that previously reported via `apiErrorMessage` into a branch that reported nothing — so the widening and the ordering were one change. The refetch is now awaited and the copy assigned after it.
- `views/MobileAdmin.vue` — `fetchQueue`'s `if (seq !== queueFetchSeq) return` sat inside `try` only; the `catch` still wrote `fetchError.queue` and the `finally` still cleared `loading.queue`, so a superseded poll rejecting painted "couldn't refresh" over a list a newer fetch had just proved fresh. The guard now covers **every exit**.
- `views/MobileAdmin.vue` — `pruneQueueItemState` cleared `respondingItems`, which **is** the in-flight Send guard. The GET is a `limit: 100` window (ordered status → priority → created_at) an item can fall out of and return to, so the guard could be cleared mid-POST, re-enabling Send and reporting a *recorded* answer as not recorded. An in-flight id is now exempt — the **whole per-card record**, not just the flag, since the retryable-failure path keeps the selection and note for a retry. `resetQueueItemState` no longer delegates to the prune (logout is a wipe, not a reconcile; delegating would inherit the exemption and strand a disabled Send).

## Test Plan

- [x] `vitest run` — **57 files / 1232 tests** (54 new: 45 for the original fix + 9 for the review round, each **mutation-verified** — reverting any one fix turns them red)
- [x] Playwright against the live stack — **16/16**: the new spec (7) + the #1927 `/m` (5) + Operations (3) regression
- [x] Re-verified after the review round — new spec **8/8**; full e2e suite **130 passed** (the 6 remaining failures are pre-existing `@interactive` specs needing agent/portal data, reproduced identically with the changes stashed)
- [x] **Proved to fail on the pre-fix code** (source stashed, spec kept): test 1 fails with `a tap must not answer; captured POSTs: [{"id":"e2e-2370-approval","body":{"response":"approved","response_text":"Deny"}}]`
- [x] `vite build` ✓ · `check:tokens` ✓ · loading-gate ratchet ✓ · raw-colour scan (review-enforced, not CI): `views/MobileAdmin.vue` **0 / 4 / 128** (baseline 0 / 4 / 130), `QueueCard.vue` 28 / 59 / 0, `stores/operatorQueue.js` 4 / 0 / 0 — all unchanged or smaller
- [x] Operator eyeball (review round) — `/m` approval answered **Deny** with a note recorded as `response='Deny'`, `response_text='<note>'`; offline/online poll banner appears and self-clears. Desktop refusal verified by resolving an item **directly in the DB** (no WebSocket broadcast, so the open tab still shows it pending): **without** the fix no banner renders at all, **with** it the banner renders carrying the copy.
- [x] Operator eyeball on localhost `/m` (three seeded cards), then the agent's own queue file: `approval → response='Approve', response_text='test'` · `question → response='test answer'` · `alert → response='acknowledged'` — never `'approved'`
- [x] `/review` — round 1: 0 critical, 7 low applied · round 2 (@dolho): 3 findings, all fixed in `253c45cd` · `/validate-pr` — 0 critical · `/cso --diff` — 0 findings (`docs/security-reports/cso-diff-2026-08-21-2370-mobile-approval-payload.md`)

## Notes for reviewers

- **AC #5 deviation (deliberate)**: a failed respond no longer goes through the old top-of-page `reportActionFailure` banner — 5xx/transport → per-card `InlineError` (design-system p18: next to the control; also drops the untrue "Nothing was changed" for a timed-out POST), 409/400/404 → the persistent page banner (the card is about to leave). Honest failure is preserved and more visible.
- **Desktop touches are DRY plumbing only**: the store now trims the note (`"  "` → `null`) and shares the refusal copy; `QueueCard` renders the same label set. No desktop UX change.
- Follow-ups filed: #2375 (Workspace asks panel — same inversion for typed answers, different endpoint/model), #2376 (server-side `response ∈ options` check — the guard that would have made this a 422 five months ago), #2377 (desktop never shows the not-recorded copy) — **partially addressed here**: the wipe is fixed and #2377's third acceptance criterion (the store test asserts the notice is observable after a 409 and a 400) is met. What remains is presentation: `Operations.vue` still renders `error` only as the `:detail` of "Couldn't refresh the queue", and `InlineError` folds `detail` into a **collapsed disclosure**, so the operator must expand "Technical detail" to read it — and the next 15s poll clears it. Residuals registered in trinity-dev `DEBT_INBOX`: options size-cap marker selectable on both surfaces (parity), and — from the review round — `/m` posts through the raw `axios` global with **no timeout**, so a never-settling POST leaves that card's Send disabled and its state unprunable (the prune exemption is the correct trade; the fix is a per-request timeout, not removing the exemption). Notes left on #1030 and #1924.

Fixes #2370


dolho:
## Review — #2378 (mobile approval payload)

Read the full diff plus the surrounding code: `MobileAdmin.vue`, `stores/operatorQueue.js`, `QueueCard.vue`, `QueueItemDetail.vue`, `Operations.vue`, `InlineError.vue`, backend `routers/operator_queue.py` + `db/operator_queue.py`, and both test files.

**The core fix is correct.** `response` now carries the decision on every path; `queueResponseKind` matches desktop's type switch (and is strictly more capable for approval-without-options, which desktop renders as nothing); `optionsOf` correctly defends against the untyped JSON the DB stores; and `respondRefusedAsNotPending`'s 400/404/409 set matches what `respond_to_queue_item` actually raises (validation errors are 422, access is 403 — neither is swallowed).

Three findings.

---

**[1] `stores/operatorQueue.js:147` — medium: 400/404 now lose the desktop's only failure message.**

Routing 400/404 into the `respondRefusedAsNotPending` branch is semantically right, but that branch calls `fetchItems()`, whose first synchronous statement is `error.value = null` (~line 100). The `QUEUE_RESPONSE_NOT_RECORDED` set one line earlier is wiped before anything renders.

Before this PR, 400/404 fell through to the `else` and left `apiErrorMessage(err, …)` in `error`, which `Operations.vue:144/169/225` renders as the InlineError/LoadFailed `:detail`.

Scenario: an operator answers a queue item whose row was purged (404) or is already terminal (400). Previously they saw an error string; now the item silently vanishes with no indication the response was not recorded. #2377 covers the 409 case, but this PR *widens* it to two statuses that previously did report.

Fix: set `error.value` after the `fetchItems()` call (or await it first).

---

**[2] `views/MobileAdmin.vue:824` — low: the stale-fetch guard covers only the success path.**

`if (seq !== queueFetchSeq) return` sits inside `try`, before the assignments; the `catch` still writes `fetchError.queue` unconditionally and `finally` still clears `loading.queue`.

Scenario: a 15s poll (seq N-1) is in flight when the operator answers a card. The success-path `fetchQueue()` (seq N) resolves, repaints a correct list and clears `fetchError.queue`; the superseded poll then rejects on a transient blip and sets `fetchError.queue` — so the Ops tab paints a refresh-failed/stale banner over demonstrably fresh data, and drops `loading.queue` while the newer request is still running.

Fix: extend the stale guard to the `catch` and `finally`.

---

**[3] `views/MobileAdmin.vue:1143` — low: `pruneQueueItemState` can clear an in-flight send guard.**

`pruneQueueItemState` deletes `respondingItems[id]` for any id absent from the freshly fetched list, but `respondingItems` is the in-flight guard `sendQueueResponse` checks (`if (!body || respondingItems[id]) return false`). The GET is `limit: 100` with no stable ordering guarantee, so on a busy queue an item with an in-flight POST can drop out of one poll's window and reappear in the next; its guard is cleared, the card re-renders with Send enabled, and a second tap issues a duplicate `POST …/respond` while the first is still outstanding.

The server 400s the loser so nothing is mis-recorded — but the operator gets the "no longer pending — your response was not recorded" banner for an answer that *was* recorded.

Fix: exclude ids with `respondingItems[id] === true` from the prune.

---

webmixgamer:
## All three findings addressed — `253c45cd`

Thanks, these were all real and all in code this PR introduced. Verified each against the running code before changing anything rather than taking them at face value; notes below where my reading differed from yours.

---

**[1] `stores/operatorQueue.js` — confirmed, and it is a regression this PR introduces.**

Fixed by awaiting the refetch and assigning the copy after it. Your diagnosis was exact: `fetchItems()`'s synchronous prologue nulls `error`, so the un-awaited call destroyed the notice in the same tick. The part worth stating explicitly is *why it had to be fixed here* — pre-PR only 409 entered that branch; widening `respondRefusedAsNotPending` to 400/404 moved two statuses that previously reported via `apiErrorMessage` into a branch that reported nothing. The widening and the ordering were one change.

A failed refetch's own message is deliberately overwritten by the not-recorded copy: the operator needs to know their answer was dropped more than they need to know the list is stale.

**Verified on a live stack, not just in tests.** Reproducing this by hand is impossible, which is worth recording for anyone who tries: `POST /respond` broadcasts `operator_queue_responded`, and the other tab's store marks the item responded *locally, sub-second* — the card leaves `openItems` before a human can click. So I resolved the item **directly in the DB** (no broadcast, so the open tab still shows it pending) and answered it, twice, identical except for the fix:

| | banner | message |
|---|---|---|
| without the fix | none at all | card vanishes, operator told nothing |
| with the fix | renders | "…your response was not recorded." |

**One correction to the shared premise.** You wrote that `error` "renders as the InlineError/LoadFailed `:detail`" — true, but `InlineError` puts `detail` inside a **collapsed `<details>` disclosure**. So even with the copy surviving, the operator sees a headline about *refreshing* and must expand "Technical detail" to learn their answer was dropped. Strictly better than before (it survives instead of being destroyed, which is the prerequisite for fixing the presentation), but not yet good. That's #2377, and this change now satisfies its third acceptance criterion — the store test asserts the notice is observable after a 409 and a 400. I've commented there and recorded the disclosure behaviour in `operating-room.md` so the next person doesn't re-derive it.

---

**[2] `MobileAdmin.vue:824` — confirmed.** The guard now covers `catch` and `finally`. `return` inside try/catch still runs `finally`, so the `finally` carries its own check rather than inheriting one. I walked every interleaving to confirm `loading.queue` cannot stick `true`: the newest sequence always reaches its own `finally` with `seq === queueFetchSeq`, and `fetchQueue` is the sole writer of that flag.

---

**[3] `MobileAdmin.vue:1143` — confirmed, with one deviation from your suggested fix.**

You suggested excluding ids with `respondingItems[id] === true` from the prune. I exempt the **whole per-card record** instead — selection and note included — because the retryable-failure path explicitly promises to keep them for a retry, and preserving only the guard flag would delete exactly what that promise needs.

Also: `resetQueueItemState` no longer delegates to `pruneQueueItemState([])`. Logout is a wipe, not a reconcile; delegating would silently inherit the new exemption and strand a card's Send disabled across the next sign-in.

One refinement to the mechanism, since it changes how narrow the window is: the list *is* deterministically ordered (`status → priority → created_at DESC`) — it isn't "no stable ordering guarantee". What's unstable is **window membership** at `limit: 100`: an item can be pushed past the 100th row by newer high-priority arrivals and return once those are answered. Narrower than "no ordering", still reachable, and the fix is right either way.

---

### Tests

9 new assertions in `operatorQueueResponse.spec.js` — behavioural for [1] (the copy is observable after 409/400/404, outranks a failed refetch, and a 500 still takes the generic path), source-pinned for [2] and [3], which are `<script setup>` internals against `environment: 'node'`. **All 9 mutation-checked**: reverting each fix individually fails them, so they're guards rather than tautologies. The spec's old comment documenting the wipe as a known limitation is gone — it was about to become false.

`1232/1232` unit · `8/8` mobile e2e · full e2e `130 passed` (the 6 failures are pre-existing `@interactive` specs on this instance, reproduced identically with my changes stashed) · production build clean.

### Residual, not fixed

`/m` posts through the **raw `axios` global, which carries no timeout** (only `api.js:17` sets one, and `/m` deliberately doesn't use that instance). So a POST that never settles now leaves that card's state unprunable as well as its Send disabled. The exemption is the correct trade — removing it re-opens [3] — and the guard fails *closed*, which is the safe direction. Captured as `debt:2026-08-24-m-respond-guard-pinned-by-hung-post` with the fix sketch (a per-request timeout on the respond POST, not removing the exemption).

dolho:
## /review — re-review after `253c45cd`

**All three findings are fixed, and I verified each rather than reading the diff and agreeing.** Two corrections in your reply are right and I'm adopting both.

---

### [1] — fixed, and the fix is sound for a reason worth writing down

The ordering is right, but the thing that makes it *safe* is one level down: `fetchItems` has its own `try/catch/finally` and **never rejects**. That matters because the new `await` sits inside a `catch` block — if the refetch could throw, it would propagate out of `respondToItem` and skip the `error.value = …` line entirely, silently reintroducing exactly the bug being fixed, on the failure path. It can't, so "the refetch's own error is deliberately overwritten" holds structurally and not merely by test.

**Mutation-checked, not taken on trust.** I reverted the fix in a worktree (`await fetchItems(); error.value = …` → `error.value = …; fetchItems()`) and ran the spec:

```
5 failed | 51 passed (56)
  ✗ a 409 refusal leaves the not-recorded copy OBSERVABLE after the refetch
  ✗ a 400 refusal …
  ✗ a 404 refusal …
  ✗ the not-recorded copy outranks a failed refetch
  ✗ the desktop store awaits the refetch BEFORE assigning the not-recorded copy
```

Guards, not tautologies. Your claim checks out.

**I also verified the refusal set against the backend** instead of accepting the mapping:

| status | `routers/operator_queue.py` | in the refusal set? |
|---|---|---|
| 404 | `if not existing` — row gone | ✅ |
| 400 | `if existing["status"] != "pending"` — already terminal | ✅ |
| 409 | lost the CAS race | ✅ |
| 403 | `_assert_agent_accessible` — access denied | ❌ correctly excluded |

And 400 is the **only** 400 on that route — a malformed body yields FastAPI's 422, so there is no path where "your payload was bad" gets reported as "no longer pending". The docstring is exact.

### [2] — fixed. `return` inside `try/catch` still runs `finally`, so the `finally` carrying its own check rather than inheriting one is the right shape.

### [3] — fixed, and **your deviation is better than what I suggested.**

I proposed excluding ids whose guard is set; you exempted the whole per-card record. You're right: the retryable-failure path explicitly promises to keep the selection and note for a retry, so preserving only the flag would delete precisely what that promise needs. Dropping the `resetQueueItemState` → `pruneQueueItemState([])` delegation is the other half — logout is a wipe, not a reconcile, and inheriting the exemption would strand a Send disabled across the next sign-in.

I checked the case that would have made the exemption leak: **the success path never sets `respondingItems[id] = false`.** It doesn't need to — `clearQueueItemState(id)` deletes from all four maps including that one, so an answered card's record is prunable on the next reconcile. Had it only removed the card, the new exemption would have pinned that record forever.

### Your two corrections — both accepted

- **`InlineError` puts `detail` in a collapsed `<details>`.** You're right, and it changes what [1] actually buys: the copy now *survives*, which is the prerequisite, but the operator still reads a headline about refreshing and must expand "Technical detail" to learn their answer was dropped. Recording the disclosure behaviour in `operating-room.md` so the next person doesn't re-derive it is the right move.
- **"No stable ordering guarantee" was wrong** — the list is ordered `status → priority → created_at DESC`. What's unstable is *window membership* at `limit: 100`. Narrower than I wrote, still reachable, same fix.

---

### One new informational finding

**[I1] The raw-source assertion is broad enough to edit prose, and it did** (Confidence: 8/10)

Dropping `stripComments` in favour of raw source is defensible — "the old literals may not survive anywhere, not even in a comment" is a real standard. But one assertion is unanchored:

```js
expect(mobile).not.toMatch(/request_type/)
```

so it matches the *comment explaining the bug*, which had to be rewritten:

> ~~The API field is `type`; a `request_type` read here rendered a blank line for months.~~
> The API field is `type`; a read of a misnamed field here rendered a blank line for months (issue 2370).

The name **was** the value of that comment. A future reader who sees `request_type` somewhere now has nothing to recognise it by, and "a misnamed field" is the kind of sentence that decays into noise. The sibling assertion shows the fix: `/response:\s*['"]approved['"]/` requires the key, so the comment can still quote `'approved'` freely. Narrow this one the same way — `/item\.request_type/` or `/\brequest_type\s*[:.]/` — and restore the field name.

Worth naming as a class, since this file now leans on source-matching in several places: **a source-regex guard whose pattern can match prose starts governing prose.** The pattern should be as specific as the code it forbids.

Non-blocking. The `[3]` pin (`/live\.has\(id\) \|\| respondingItems\[id\] === true/`) is exact-string and brittle to harmless reformatting, but it does catch deletion of the exemption, which is the regression — an acceptable trade given `environment: 'node'` and no mount harness, and you say so.

---

### Summary

- **Prior findings**: 3/3 fixed, all verified (one by mutation testing, one against the backend's own status codes, one by checking the leak case the fix could have opened)
- **New**: 1 informational, non-blocking
- **Spec**: 56/56 green in a clean worktree
- **Scope**: unchanged

Nothing here blocks. Nice work on the live reproduction — resolving the row directly in the DB to defeat the `operator_queue_responded` broadcast is exactly the right instinct for a race you can't click fast enough to hit, and it's the kind of note that saves the next person an hour.

*(Minor: your comment cites `32ded859`, which isn't in the repo — the commit that landed these is `253c45cd`. I did the same thing on my own PR yesterday, so: purely so nobody greps for a sha that was never pushed.)*

dolho (approved):

webmixgamer:
Thanks for the re-review — and for mutation-testing the fix in your own worktree rather than reading the diff and agreeing. Two things you added that I hadn't stated:

- **The structural reason [1] is safe.** You're right that the load-bearing property is `fetchItems` never rejecting: because the new `await` sits inside a `catch`, a throwing refetch would skip the `error.value` assignment entirely and silently reintroduce the bug *on the failure path*. I verified that non-rejection before making the change but wrote it as an aside; it's the whole argument.
- **The leak case for [3]** — that the success path never sets `respondingItems[id] = false`, and it doesn't need to because `clearQueueItemState(id)` deletes from all four maps. Had it only dropped the card, my exemption would have pinned that record forever. That's the check I should have run on my own fix.

Your backend cross-check on the refusal set is a better verification than mine: I reasoned from `respondRefusedAsNotPending`'s docstring, you went and read the route.

**[I1] — accepted, and captured rather than fixed here.**

You're right on both halves: the pattern governs prose, and the field name was the value of that sentence. Your narrowing works — I tested `/\.request_type\b|request_type\s*:/` against the restored comment (no match) and three regression shapes, `{{ item.request_type }}` / `props.item.request_type` / `{ request_type: … }` (all match), so it loses no coverage.

I'm **not** pushing it to this PR. It landed after your approval, and I can't read this repo's stale-review dismissal setting with my token — so the cost of a commit is somewhere between "one CI cycle" and "your approval is dismissed", for a code comment. Wrong trade against an approved PR.

Captured as `debt:2026-08-24-source-regex-guard-governs-prose` in trinity-dev `DEBT_INBOX` (alongside this PR's two other deferrals), carrying the verified pattern and the note that the fix should also add your class — *a source-regex guard whose pattern can match prose starts governing prose; the pattern must be as specific as the code it forbids* — to `docs/memory/learnings.md`. That's the durable home, since `/autoplan` reads it before planning and this spec now pins several view internals on source by design.

Also: fixed the `32ded859` citation in my earlier comment — you were right, that SHA was amended away when I stripped a session URL and never existed in the repo. Corrected to `253c45cd`.

PR #2382, opened 2026-08-24
fix(ui): a failed background refresh keeps the data and says so (ent#253)

dolho:
## Audit first (AC #1)

All 31 interval-refreshed files, each with its verdict and the evidence for it, posted on the issue **before** any change: [audit comment](https://github.com/Abilityai/trinity-enterprise/issues/253#issuecomment-5393186249). #1927 had already fixed four surfaces and shipped the rule (`utils/loadingState.js::viewState`) plus the bare-gate ratchet, so most of the inventory was already green. Three were not.

## What was actually wrong

The re-flash the issue describes was real — each of the three set its loading flag *inside the function the poll calls*, so the panel was replaced by a spinner on every tick. But the `catch` was the serious half:

```js
} catch (error) {
  console.error('Failed to load metrics:', error)
  metricsData.value = { has_metrics: false, message: 'Failed to load metrics' }   // ← overwrites good data
}
```

One transient 502 rewrote a working agent as one with **no metrics** / **no dashboard** / **an unreachable collector**, and the real values were gone until the next success. That is a claim about the subject manufactured out of a claim about the request — #1926 one layer down — and it also violates AC #5 outright. Each failure also went to `console.error` and nowhere else.

## The fix, per surface

- Gate on **"no data yet"** via `viewState`; `loading` keeps meaning "fetch in flight" so it can still drive the retry labels. The poll is now invisible.
- A **failed refresh keeps the data** and raises the shared stale banner *beside* it — a sibling, never an else-if arm, which would replace the data it promises to keep.
- A failed **first** load gets `LoadFailed` + retry, kept distinct from the empty state (a statement about the agent, not about the request).
- The bespoke spinners are replaced by the #245 `ScanlineReveal` primitive, which brings `prefers-reduced-motion` with it (AC #4, AC #6).
- Agent-switch watchers clear the error, so agent B's first failure is not reported with agent A's detail.
- In the store, a transport failure (`refreshError`) is held apart from `error`, which carries the collector's own message out of a **successful** response. Collapsing the two let a dropped request read as the collector declaring itself down, and that is what set `available = false`.

## Verified live, not just asserted

Against a local stack with this branch's source Vite-mounted, on the one mounted surface, driving the panel's **real 30 s poll** with the endpoint aborting (`page.route(... r.abort())`):

```
1. scanline primitive mounted   : true
   bespoke 8x8 spinner present  : false
   dashboard content rendered   : true
2. endpoint aborting; waiting out the panel's own 30s interval…
   loading face during the wait : false      ← no re-flash (AC #3)
3. after the failed refresh:
   content still on screen      : true       ← data survives (AC #5)
   "no dashboard" empty state   : false      ← the fabricated empty state is gone
   stale banner                 : true       ← and it says so
```

## Honest scope — this corrects the audit

`MetricsPanel` and `ObservabilityPanel` turn out to be mounted by **nothing**; `stores/observability.js` has exactly one consumer, that orphaned panel. Their defect is real in code and unreachable in the app — the live win is `DashboardPanel`. [Correction posted on the issue](https://github.com/Abilityai/trinity-enterprise/issues/253#issuecomment-5393278568); the convention note now says to record consumer-reachability during the audit, because `grep setInterval` finds code that polls, not surfaces a user can reach.

Worth someone's decision, filed rather than folded in: the observability store polls on an interval with an unmounted consumer — delete the surface or wire it back is a product call, not a refresh-treatment one.

## Tests

`tests/unit/refreshTreatment.spec.js`:

- **Behavioural** (Pinia + mocked transport) on the store, because a source regex cannot prove a branch: a later failed fetch keeps `metrics`/`totals` and `available`, sets `refreshError`, and leaves `hasLoaded` true; a **first** failed fetch still fails honestly (`available = false`, nothing to protect); a transport failure never overwrites the collector's own message; a later success clears the stale flag.
- **Source-asserted** on the components (no mount harness in this repo): no in-flight gate on rendered content, `hasLoaded`-derived loading face, no data-clobbering `catch`, `InlineError` + `LoadFailed` present, agent-switch clears the error, bespoke spinners gone.

Ratchet: `loading-gate-baseline.json` regenerated (63 files/80 gates → 60/77 — the ratchet spec pins the baseline exactly, so a fixed file's stale entry fails CI). `npm run check:tokens` passes; `ObservabilityPanel` also loses a raw `text-blue-600` (1 → 0).

## Docs (AC #7)

The convention already lived in `design-system.md` §6 / `design-system-contract.md` from #245/#1927. This adds the half ent#253 found: **a failed request never becomes a statement about the subject** — keep the data, set an error, raise the banner; a `catch` that assigns the data ref is the smell. Plus the reachability note above.

## Not in this pass

- `ReplayTimeline.vue` — excluded by the issue, untouched.
- Bare gates on surfaces that **do not poll** (`SkillsPanel`, enterprise `Audit`, the settings tabs): they re-flash only on a user-triggered refetch, a different and much weaker complaint. They stay in the ratchet baseline.

Related to Abilityai/trinity-enterprise#253


obasilakis (changes requested):
Reviewed against `/validate-pr`. The core fix is right and well argued — a `catch` that assigns the data ref turns a claim about the request into a claim about the subject (#1926), the stale banner is correctly a sibling rather than an `v-else-if` arm, and splitting `refreshError` from `error` (the collector's own verdict out of a *successful* response) is the distinction that was actually driving `available = false`. The store-level tests are behavioural rather than regex, which is the right call for the half where the honesty is decided. The audit correction on the issue — `grep setInterval` finds code that polls, not surfaces a user can reach — is a good durable output.

One blocking item, on the one surface here that is actually mounted.

## "Agent Not Running" is now unreachable; a stopped agent's Dashboard tab renders blank forever

`viewState()` deliberately ignores `loading` (`loadingState.js:37`, `void loading`). Its only inputs are `hasLoaded` and `error`:

```js
if (!hasLoaded) return { state: hasError ? 'failed' : 'loading', stale: false }
```

For a stopped agent, `onMounted` takes the `else { loading.value = false }` branch and never fetches. So `dashboardData` stays `null`, `loadError` stays `''`, and `firstLoad` is **`true` permanently**. The new placeholder is first in the chain:

```
v-if="firstLoad"                        ← wins forever
v-else-if="loadFailed"
v-else-if="agentStatus !== 'running'"   ← never evaluated
```

Result: an empty `min-h-[8rem]` `aria-hidden` div under a `ScanlineReveal` stuck at `:loading="true"` — a dimmed, animating, wordless panel. Previously `v-if="loading"` was false here and the chain fell through to "Agent Not Running · Start the agent to view its dashboard."

This is reachable on the default path: the Dashboard tab is gated on `hasDashboardFlag`, which comes from the DB-backed `/exists` probe (`AgentDetail.vue:883`), not on run state — so any stopped agent that has ever published a `dashboard.yaml` shows the tab. `DashboardPanel` is the only mounted surface this PR touches (`AgentDetail.vue:176`), so this is the whole live blast radius.

Either gate the placeholder on run state:

```diff
-      <div v-if="firstLoad" class="min-h-[8rem]" aria-hidden="true"></div>
+      <div v-if="firstLoad && agentStatus === 'running'" class="min-h-[8rem]" aria-hidden="true"></div>
```

…or seed `hasLoaded` on the non-running mount branch. Whichever you pick, `ScanlineReveal`'s `:loading` needs the same treatment or it animates behind the not-running copy.

Note the stop-while-viewing path is fine — `hasLoaded` is true by then and the chain falls through correctly. It is specifically **mounting on a stopped agent** that breaks, which is why the Playwright verification in the description read green: it only drove the running path.

`MetricsPanel.vue:32` has the identical construction. Unmounted today, so not user-visible, but worth fixing with it rather than leaving the pattern in the tree to be copied.

## Non-blocking

`stores/observability.js` — on a failed refresh with `hasLoaded` true, `error` is deliberately left alone (correct — that is the fix), but a stale collector message from an earlier successful response keeps rendering while the new failure only reaches `refreshError`. Narrow window (success → collector-reports-unavailable → transport failure), and arguably fine as-is.

## Process

Body says "Related to Abilityai/trinity-enterprise#253" rather than `Fixes`. Cross-tracker doesn't auto-promote either way, so ent#253 needs a manual `status-in-dev` at merge.

dolho:
## Code review — re-review at `536c660`

**Note on the target:** GitHub recorded the force-push (`head_ref_force_pushed` is in the timeline) but never advanced this PR's head — the API still reports `302df29` while the branch is at `536c660`. A first pass run against the PR ref therefore read pre-fix code and re-raised the already-fixed placeholder bug; this pass was run against `origin/dev...feature/ent253-refresh-treatment` instead. Anyone reviewing from the GitHub diff view is likely seeing the stale tree too.

**Scope:** 4 commits, 9 files. Every hunk read at the target revision in a scratch worktree, SFC templates compiled, the compiled `v-if` chain inspected to confirm branch ordering, full frontend unit suite run (58 files / 1246 tests green) plus the loading-gate ratchet.

The core change is right: the `catch` blocks no longer fabricate `{has_dashboard:false}` / `{has_metrics:false}` / `available:false`, so a transient 502 stops rewriting a working agent as a misconfigured one. What follows is the residue.

### Medium

**1. `DashboardPanel.vue:691` — a failed first load now outranks "Agent Not Running", and the status watcher never clears it.**

The `agentStatus` watcher's non-running branch calls `stopRefresh()` only; it never clears `loadError`. With the new ordering (confirmed in the compiled render: `awaitingFirstLoad ? … : loadFailed ? … : agentStatus !== 'running' ? …`) the `loadFailed` arm at :36 pre-empts the not-running arm at :45.

Scenario: agent running, operator opens the Dashboard tab, the first fetch fails (container still booting → 502), `LoadFailed` renders. The operator then stops the agent. `dashboardData` is still `null` and `loadError` is still set, so the panel keeps showing "Couldn't load the dashboard — The agent did not return its dashboard" with a Retry button that can only ever fail again, instead of "Agent Not Running · Start the agent to view its dashboard." Before this PR the chain put `agentStatus !== 'running'` ahead of the error state, so stopping the agent recovered the panel.

Same for the stale banner at :8 — with data on screen and a failed poll, stopping the agent leaves "Couldn't refresh the dashboard…" pinned above the not-running copy.

Fix: clear `loadError` (and arguably `lastLoadedAt`) in the watcher's `else` branch, mirroring what the `agentName` watcher already does.

### Low

**2. `MetricsPanel.vue:421` — the same bug, same shape.** `loadFailed` at :42 precedes `agentStatus !== 'running'` at :51, and the status watcher never clears `loadError`. Lower severity only because nothing mounts `MetricsPanel` today — `grep` across `src/frontend` finds it in `raw-color-baseline.json` and the new spec and nowhere else — but it should move with #1 rather than be left to diverge.

**3. `ObservabilityPanel.vue:157` — the stale banner can call a 20-minute-old reading "just now".** It interpolates `formatLastUpdated`, a computed whose only reactive dependency is `observabilityStore.lastUpdated`; the elapsed time comes from a non-reactive `Date.now()`. On a failed refresh `lastUpdated` does not change, so the computed is never invalidated and keeps serving its cached string. Last successful poll at 10:00 renders "just now"; the collector then goes unreachable for 20 minutes and the banner reads "Couldn't refresh — showing the reading from just now." for the whole outage — precisely the never-present-stale-data-as-fresh claim the line exists to make. `DashboardPanel` avoids this by using `staleBannerMessage`, which prints an absolute `toLocaleTimeString`; this panel should do the same, or tick a `now` ref.

**4. `ObservabilityPanel.vue:47` — a failed first fetch still renders as a claim about the platform's configuration.** `enabled` stays at its `false` default, and the chain's first arm is `v-if="!observabilityStore.enabled"` → "OTel not enabled. Set OTEL_ENABLED=1." That is the #1926 class this PR set out to remove, one arm further up. The store's new `if (!this.hasLoaded) { this.error = … }` branch (`stores/observability.js:196`) records the real reason and the spec asserts `store.error` is truthy ("still fails honestly when the FIRST fetch fails") — but no template arm can display it, because `!enabled` short-circuits first. The test gives coverage the UI does not have.

**5. `stores/observability.js:185` — `hasLoaded` means "the request succeeded", not "we have data".** It is set on any HTTP 200, including `{enabled:false}` or `{available:false, error:…}` responses carrying no metrics. Since `isStale = refreshError && hasLoaded`, a later transport failure on such an install makes the panel assert "Couldn't refresh — showing the reading from …" when it has never had a reading to show. If the flag is meant to gate "we have data", derive it from the presence of `data.metrics` / `data.totals` (or `available`), not from the response arriving.

**6. `DashboardPanel.vue:22` — the celebratory wipe plays over an empty state.** `:reveal="view.state === 'ready'"`, but `viewState` is called without `count`, which defaults to `1`, so `state` is `'ready'` for every successful response including `has_dashboard:false` (and `has_metrics:false` at `MetricsPanel.vue:26`). `ScanlineReveal`'s documented contract is `reveal:false` "when loading ended without real data (error / empty): snap to loaded instead of playing the celebratory pass", so the 550 ms wipe runs over "No Dashboard Defined". Cosmetic, but a deviation from the primitive this PR is adopting; `count: dashboardData.value?.has_dashboard ? 1 : 0` would honour it.

### Confirmed fixed since the last review

The `firstLoad` placeholder no longer swallows the chain on a stopped agent: `awaitingFirstLoad = firstLoad && agentStatus === 'running'` gates both the placeholder arm and `ScanlineReveal :loading`, in both panels, pinned by a spec case that asserts the rule rather than the spelling.

obasilakis (changes requested):
Re-reviewed at `536c6605`, the tip of `origin/feature/ent253-refresh-treatment`. Every hunk read at that revision via `git show`, the branch chains traced arm by arm in both panels, `loadingState.js` / `ScanlineReveal.vue` / `InlineError.vue` / `LoadFailed.vue` read for the contracts this pass adopts, both ratchet scanners run against a `git archive` of the branch tree, and the new spec read in full.

## Before the code: this PR's GitHub record is wedged, and the green checks belong to a tree that no longer exists

The branch ref is `536c6605`. The PR record's head is `302df29`, and that commit is **not an ancestor of the branch** — it was orphaned by the force-push. `mergeable` is `null`/`unknown`, and there are **zero check-runs against `536c6605`**. The full green matrix everyone can see (pytest, e2e, build, CodeQL, gitleaks) ran against `302df29`, i.e. against the code *before* the fix for the blocking finding from the last round.

Three consequences worth stating plainly: the GitHub diff view is showing reviewers the pre-fix tree, no CI has evaluated what is actually on the branch, and the two nightly "merge conflict against `dev`" comments predate the rebase (the branch's merge-base is now exactly `origin/dev`, so there is no conflict left). Please unwedge the PR — an empty commit, or a close/reopen — before anyone tries to merge it, so the checks and the diff describe the tree that would land.

## Earlier findings

**Blocking — "Agent Not Running" unreachable, blank panel forever: fixed.** `awaitingFirstLoad = firstLoad && props.agentStatus === 'running'` now gates both the placeholder arm and `ScanlineReveal`'s `:loading`, in both panels. I re-derived the stopped-agent mount: `onMounted` takes the `else` branch, `dashboardData` stays `null`, `firstLoad` stays true, `awaitingFirstLoad` is false, arm 1 and arm 2 both fall through, and `agentStatus !== 'running'` renders. The primitive gets the same gate, so nothing animates behind the not-running copy. The spec case pins the rule rather than the spelling, and the assertion that the not-running arm stays downstream of the two request-shaped arms is the right thing to have written down.

**Non-blocking — a stale collector message outliving its response: unchanged.** Still the narrow success → collector-reports-unavailable → transport-failure window. Fine as-is; noting only that it was not silently claimed as fixed.

**Process — `Related to` rather than `Fixes`.** Unchanged, as expected across trackers. ent#253 still needs its `status-in-dev` label applied by hand at merge.

## New findings

### Blocking: the `agentStatus` watcher never clears `loadError`, so a request-shaped arm outranks the run-state arm again

`DashboardPanel.vue:691` — the watcher's non-running branch calls `stopRefresh()` and nothing else. `loadError` and `lastLoadedAt` survive the transition, and both feed arms that sit *above* `v-else-if="agentStatus !== 'running'"` (`:45`). Two reachable outcomes, both on the one surface this PR actually ships to users:

1. Agent running, Dashboard tab opened, first fetch fails — a container that is up while its agent-server is still booting returns 502, and this endpoint proxies to the agent. `LoadFailed` renders, correctly. The operator then stops the agent. `dashboardData` is still `null` and `loadError` is still set, so `loadFailed` (`:36`) keeps winning and the panel says *"Couldn't load the dashboard — The agent did not return its dashboard. This is not the same as an agent without one."* with a Retry button that cannot succeed, where `dev` says *"Agent Not Running · Start the agent to view its dashboard."*

2. Agent running with a rendered dashboard, one poll fails, banner appears, operator stops the agent. `hasLoaded` is true so `view.stale` is true and the `InlineError` at `:9` stays pinned, asserting *"showing data from 10:42"* directly above a panel body that is rendering "Agent Not Running" and showing no data at all.

This is the same shape as the item I blocked on last round — an arm that describes the *request* pre-empting the truer statement about the *subject* — and (2) in particular is the PR's own thesis inverted: a banner that promises the data is still on screen, printed over a panel where it is not. It is a regression against `dev` on `DashboardPanel`, which the description correctly identifies as the whole live blast radius.

Clearing `loadError` in the watcher's `else` branch fixes both: `view` falls back to `loading`, `awaitingFirstLoad` is false because the agent is not running, `view.stale` goes false, and the chain reaches arm 3. One line, mirroring what the `agentName` watcher already does. `lastLoadedAt` does not need clearing — `staleMessage` only renders under `view.stale`.

You have already written this up yourself as Medium #1 in the review comment on this PR; I am confirming it independently and treating it as blocking rather than as a note, because it is a user-visible regression on the only mounted surface and the fix is two lines plus a spec case. `MetricsPanel.vue:421` carries the identical construction and should move with it for the same reason the last fix did — not because it is reachable, but so the pattern is not left in the tree to be copied.

### Low: `ObservabilityPanel` hand-rolls the stale banner, and the hand-rolled version is the one that lies

`ObservabilityPanel.vue:156` interpolates `formatLastUpdated` (`:185`), whose only reactive dependency is `observabilityStore.lastUpdated`; the elapsed time comes from a bare `Date.now()`. On a failed refresh `lastUpdated` does not change, so the computed is never invalidated and keeps serving its cached string — a reading taken at 10:00 renders "just now" for the entire outage, and the banner reads "Couldn't refresh — showing the reading from just now." That is the never-present-stale-data-as-fresh rule broken by the very line written to honour it.

The structural point underneath it: every other adopter of this pass — `InfoPanel`, `SchedulesPanel`, `MobileAdmin`, and now `DashboardPanel` and `MetricsPanel` — renders `<InlineError :message="staleBannerMessage(...)">`, and `staleBannerMessage` prints an absolute `toLocaleTimeString`, which is immune to this by construction. Skipping `ScanlineReveal` on a 192px chip is a justified call and the comment makes the case; skipping the shared banner helper is a separate decision that was not argued and is what produced the bug. Reaching for `staleBannerMessage` here would fix the honesty defect and the consistency gap in one edit, and would also give this panel the `role="alert"` the bare `<p>` lacks.

### Low: a failed first fetch still renders as a claim about the platform's configuration

`ObservabilityPanel.vue:46` — the chain opens with `v-if="!observabilityStore.enabled"` → "OTel not enabled. Set `OTEL_ENABLED=1`." `enabled` defaults to `false` and the `catch` never touches it, so a failed first fetch renders a misconfiguration instruction. That is the #1926 class this pass exists to remove, one arm further up than the arm that was fixed. The store's new `if (!this.hasLoaded)` branch (`stores/observability.js:196`) records the real reason and the spec asserts `store.error` is truthy, but the arm that would display it is a `v-else-if` behind `!enabled` and can never be reached — the test has coverage the UI does not.

### Low: `hasLoaded` means "the request returned 200", not "we have data"

`stores/observability.js:185` sets `hasLoaded` on every successful response, including `{enabled: false}` or `{available: false, error: …}` payloads carrying no metrics — and `lastUpdated` with it. Since `isStale = refreshError && hasLoaded`, a later transport failure on an OTel-disabled install renders "Couldn't refresh — showing the reading from …" when there has never been a reading to show. Deriving the flag from the presence of `data.metrics` / `data.totals` would make the name true.

### Low: `viewState` is called without `count`, so the celebratory wipe plays over the empty state

`DashboardPanel.vue:479` and `MetricsPanel.vue:234` pass `loading` / `hasLoaded` / `error` but no `count`, which defaults to `1` — so `view.state` is `'ready'` for every successful response, `has_dashboard: false` and `has_metrics: false` included. `ScanlineReveal`'s documented prop contract is that `reveal` is "false when loading ended without real data (error / empty): snap to loaded instead of playing the celebratory pass", so the 550 ms wipe currently runs over "No Dashboard Defined". Cosmetic, but it is a deviation from the primitive this pass is adopting, and `count: dashboardData.value?.has_dashboard ? 1 : 0` honours it. The failure path is already correct here — `'failed'` gives `reveal: false` and snaps.

### Low: the raw-color ratchet moved the wrong way on `ObservabilityPanel`, and the description reports only half of it

Running `scan-raw-colors.mjs` over the merge-base and over `536c6605` for the three touched files:

| file | base | head |
|---|---|---|
| `DashboardPanel.vue` | 8 nongray / 101 gray | 8 / 101 |
| `MetricsPanel.vue` | 0 / 63 | 0 / 63 |
| `ObservabilityPanel.vue` | 1 / 80 | 0 / **82** |

The `raw_nongray: 1 → 0` the description claims is real — that is the `text-blue-600` on the removed overlay spinner. But the replacement `<span class="text-xs text-gray-500 dark:text-gray-400">Loading…</span>` adds two `raw_gray` hits, taking that file from 80 to 82 against a committed baseline of 80, and `raw-color-baseline.json` is untouched by this PR. `CLAUDE.md` §9 states per-file counts may only shrink. Nothing in CI enforces this today (`frontend-build.yml` runs `check:tokens` only, and the scanner always exits 0), so it is a documented-rule miss rather than a red build — but the description asserts the file improved, and on the tracked-gray axis it did not.

## Verified clean

- **Loading-gate ratchet is exact.** The regenerated baseline (60 files / 77 gates) matches `scan-loading-gates.mjs` run against the branch tree with zero per-file mismatches, and the three fixed files are correctly absent.
- **The fix is not inert.** `agents.js` `getAgentDashboard` / `getAgentMetrics` have no `catch`, so the rejection genuinely reaches the component's `catch` — worth having confirmed, since a swallowing store action would have made the whole change a no-op.
- **Polling teardown.** `onUnmounted → stopRefresh` in both panels, the interval body re-checks run state on every tick, and the `agentName` watcher stops before restarting. No leaked intervals; the `agentStatus` watcher's `stopRefresh` is correct as far as the timer goes — it is only the error state it forgets.
- **Stale-banner treatment matches its siblings** on the two panels that matter: `InlineError` + `staleBannerMessage` gated on `view.stale`, as a sibling above the primitive rather than an arm inside it, identical to `InfoPanel` / `SchedulesPanel` / `MobileAdmin`.
- **Accessibility of the failure states.** `InlineError` and `LoadFailed` are both `role="alert"` with the failure named in text and the trace behind a disclosure, so nothing is conveyed by colour alone; `ScanlineReveal` carries `aria-busy` during loading and the placeholder is correctly `aria-hidden`.
- **Store behaviour.** The five Pinia cases are behavioural rather than regex, and they test the right distinctions — a later failure keeps `metrics`/`totals`/`available`, a first failure still fails honestly, and a transport failure never overwrites the collector's own message.
- **Scope.** `ReplayTimeline.vue` untouched, and the non-polling bare gates correctly left in the baseline rather than swept.

## Verdict

Requesting changes on the `agentStatus` watcher clearing `loadError` in both panels — it is a live regression on `DashboardPanel` and it re-opens the exact class this PR closes. The five Low items are yours to take or leave, though the `formatLastUpdated` one is worth taking with it: it is a one-line switch to `staleBannerMessage` and it removes a banner that currently states the opposite of the truth.

Separately and independently of the code: please unwedge the PR head so CI runs against `536c6605` and the diff view stops showing the orphaned tree.

dolho:
## Re-review — ent#253 refresh treatment

Fresh pass over the merge-base diff (`88efbe7d` → `1e4db339`). Frontend-only, 9 files.

**Verdict: approve with nits.** All four prior findings genuinely hold in the code — verified by reading, and the blocking one by mutation. Two new Low findings below are worth acting on; the rest are nits.

---

### Prior findings — verified

**1. BLOCKING — `agentStatus` watcher never cleared `loadError` → FIXED (verified by mutation).**
Both panels now clear it on the way to stopped (`DashboardPanel.vue:696-711`, `MetricsPanel.vue:426-441`). Not taken on trust: I deleted `loadError.value = ''` from the `else` branch of both watchers and re-ran the spec — `lets go of the error when the agent stops` fails (`21 passed / 1 failed`), and passes again on restore. The guard is real, not decorative. *(But see NEW-1 — the same symptom survives via a narrower path.)*

**2. `ObservabilityPanel` hand-rolled a relative-time stale banner → FIXED.**
`const staleReadingTime = computed(() => staleBannerMessage('metrics', observabilityStore.lastUpdated))` — the shared helper prints an absolute time *by construction* (`loadingState.js` `defaultFormatTime`), so the "stuck at just now during an outage" defect is unreachable rather than merely un-triggered. `role="alert"` is on the element. The re-pin as a rule (`not.toMatch(/v-if="isStale"[\s\S]{0,200}formatLastUpdated/)`) is the right shape.

**3. Failed FIRST fetch rendered as a config claim → FIXED.**
The request-shaped arm is now first and the config arm is demoted:
```
<div v-if="!observabilityStore.hasLoaded && observabilityStore.error" …>
<div v-else-if="!observabilityStore.enabled" …>OTel not enabled. Set OTEL_ENABLED=1.
```
Pinned by an index comparison (`expect(errArm).toBeLessThan(cfgArm)`), not a string — good: that survives rewording and still catches a reordering.

**4. `hasLoaded` meant "200", not "we have data" → FIXED in the store, but the fix contradicts its own justification.** See NEW-2.

---

### New findings

**NEW-1 — Low (high confidence): `loadError` can still outlive its request; the stop-clear runs *before* the in-flight rejection lands.**

The watcher clears synchronously on the prop change; the `catch` writes later and unconditionally:

```js
// DashboardPanel.vue:461-475
const loadDashboard = async () => {
  loading.value = true
  try { … } catch (error) {
    loadError.value = error?.response?.data?.detail || error?.message || 'Request failed'
  } finally { loading.value = false }
}
```
```js
// DashboardPanel.vue:696-711
watch(() => props.agentStatus, (newStatus) => {
  if (newStatus === 'running') { … } else {
    stopRefresh()
    loadError.value = ''
  }
})
```

Stop the agent while a dashboard fetch is in flight — the common case, because a wedged agent is exactly the one whose fetch is slow *and* the one you press Stop on — and the order is: watcher clears → request rejects (the backend proxy answers 502/503 once the container is gone) → `loadError` is set on a stopped agent. `loadFailed` then outranks the not-running arm, and the operator gets **"Couldn't load the dashboard"** with a Retry that can only fail again: the reviewer's blocking symptom, reached by ordering instead of by state.

Same shape on the `agentName` watcher, whose own comment claims it prevents this: *"otherwise agent B's first load, if it fails, is reported with agent A's error detail"* — but A's rejection landing after the switch does exactly that. (The *response* half of that race — A's dashboard rendering under B's name — is pre-existing on `dev`; `loadError` is new here.)

Fix is one guard, e.g. capture the request's identity and drop a stale outcome:
```js
const forAgent = props.agentName
… catch (error) {
  if (props.agentName !== forAgent || props.agentStatus !== 'running') return
  loadError.value = …
}
```

**NEW-2 — Low (high confidence): `hasLoaded = this.hasData` contradicts the comment that justifies it, and re-arms the chip's overlay on every poll for a no-data install.**

`ObservabilityPanel.vue:203-206` still says:
> `hasLoaded` **is deliberately left as-is**: it also retires the first-load placeholder, and an OTel-disabled install would otherwise sit behind that placeholder forever — which is the bug this branch fixed one commit ago, by a new route.

The next commit (`1e4db339`) did the opposite and left the comment standing:
```js
// stores/observability.js:192
this.hasLoaded = this.hasData
```

Two consequences:

- An OTel-disabled install answers 200 with `{enabled:false}` and no metrics → `hasData` false → `hasLoaded` false permanently. So `const firstLoad = computed(() => observabilityStore.loading && !observabilityStore.hasLoaded)` (`:196`) **reduces to `loading`** — the bare in-flight gate this PR exists to delete — and the `v-if="firstLoad"` overlay covers *"OTel not enabled. Set OTEL_ENABLED=1."* for the duration of every 60s poll, indefinitely. (`scan-loading-gates.mjs` cannot see it: `/loading/i` doesn't match `firstLoad`.)
- `hasLoaded` is now **non-monotonic**, so the catch's guard no longer means what its comment says:
  ```js
  // :203 — comment: "Before a first success there is nothing to protect"
  if (!this.hasLoaded) { this.error = this.refreshError; this.available = false }
  ```
  It now means *"whenever we currently show zero"*, so on an install whose window has emptied a transport failure **does** retract `available` — the thing this catch was rewritten to stop doing.

Cheapest fix preserving both properties is the one `:203-206` already describes: keep `hasLoaded = true` on any success, and give `isStale` a separate `hasReading = this.hasData`. Mitigating: `grep -rn "ObservabilityPanel" src` matches only its own file — nothing mounts it, so this is latent, not a live symptom.

**NEW-3 — Low: Dashboard/Metrics can still print "showing data from 10:42" over an empty state.**

`view.stale` is `hasLoaded && error` with `hasLoaded: dashboardData.value !== null`. A success returning `{has_dashboard:false}` followed by a failed poll raises `staleBannerMessage('the dashboard', …)` → *"Couldn't refresh the dashboard — showing data from 10:42."* directly above **"No Dashboard Defined"**. That is the same class the chip's `isStale` fix just removed, and the two halves of this PR now disagree: the chip gates its banner on `hasData`, these two don't. The needed bit is already computed one line up — `count: dashboardData.value?.has_dashboard ? 1 : 0` — so `v-if="view.stale && view.state === 'ready'"` closes it.

**NEW-4 — nit: one event, two colour semantics.** The same failed background refresh renders `status-danger` red via `<InlineError>` on Dashboard/Metrics and `status-warning` amber on the chip. Both are semantic tokens (no ratchet issue), but the PR's thesis is *one consistent treatment* and design-system §6 shows one. Worth a deliberate answer rather than an accident of which primitive fit the 192px chip.

**NEW-5 — nit (test quality): one assertion pins the spelling while its comment claims the rule.**
`refreshTreatment.spec.js` says *"The rule, not the spelling: whatever gates the placeholder must also consider run state"*, then asserts:
```js
expect(src, name).toMatch(
  /const awaitingFirstLoad = computed\(\(\) => firstLoad\.value && props\.agentStatus === 'running'\)/)
```
An operand swap or a line wrap — both no-ops — fail it. Same for `/v-if="isStale" role="alert"/` (attribute order) and `toContain('ScanlineReveal :loading="awaitingFirstLoad"')` (exact spacing). Suggest asserting that the `awaitingFirstLoad` declaration mentions **both** `firstLoad` and `agentStatus`. The rest of the file is genuinely rule-shaped — the arm-order index comparison, the `catch[\s\S]{0,200}dataRef.value = \{` negative, and the store half run against a real Pinia store with a mocked transport rather than a regex.

---

### Checked clean, and what proved it

- **Loading-gate ratchet — regenerated legitimately.** Scanner on the branch reports `total: 77`; the baseline says `77 / 60 files`, down from `80 / 63`. The only per-file deltas are the three removals (`DashboardPanel`, `MetricsPanel`, `ObservabilityPanel`, each 1→0). No count grew, no file added. `loadingGateRatchet.spec.js` 6/6 pass.
- **Raw-color ratchet.** No non-gray palette class anywhere in the added lines (a `(text|bg|border)-(blue|red|green|amber|…)-NNN` grep over the `+` lines returns nothing); the diff *removes* one (`text-blue-600` on the chip's old spinner). `scan-raw-colors.mjs` runs clean.
- **Suite: 59 files / 1266 tests pass** — matches the PR's claim. (Reproduced in `node:20` with a scratch vitest install; `package.json`/lockfile untouched.) The one red CI check, `pytest (base, seed 67890)` cancelled at ~25m, is a base-side infra timeout on a frontend-only PR — unrelated.
- **Component contracts read, not assumed.** `environment: 'node'` with no mount harness means a wrong prop name fails silently at runtime, so every prop/emit used against `InlineError` (`message`/`detail`/`retryable`/`retryLabel`, `@retry`/`@dismiss`), `LoadFailed` (`title`/`message`/`detail`/`retrying`, `@retry`) and `ScanlineReveal` (`loading`/`reveal`) was checked against its `defineProps`. All correct.
- **ScanlineReveal adoption is terminal-safe.** `initialPhase(false) → PHASE_LOADED` (`utils/scanlinePhase.js:39-41`), so on a stopped agent — where `awaitingFirstLoad` is false from the first frame — the slot renders unclipped and "Agent Not Running" is reachable. The placeholder-swallows-not-running fix holds. `:reveal="view.state === 'ready'"` is false for empty/failed, so no celebratory wipe over an empty state.
- **The thesis itself.** Neither `catch` reassigns its data ref (pinned negatively), and the store's catch keeps `metrics`/`totals`/`available` once data exists — exercised for real against Pinia, not source-matched. The only surviving stale-as-fresh paths I could construct are NEW-2 and NEW-3, both narrow.
- **Docs** updated in the same pass (`design-system.md`, `design-system-contract.md`, `learnings.md`); no backend, enterprise or submodule paths touched.

---

*Reviewed at `1e4db339`. NEW-1 and NEW-2 are the two I'd act on before merge; NEW-3/4/5 are fine as follow-ups.*

obasilakis (approved):

The blocking item is fixed and verified: the `agentStatus` watcher now clears `loadError` in its `else` branch on both panels, so a request-shaped arm no longer outranks "Agent Not Running". The early return in the `catch` when `agentStatus !== 'running'` closes the in-flight-rejection race, which I hadn't asked for.

Four of the five low items are in too: the stale banner now goes through `staleBannerMessage` (absolute time by construction, plus `role="alert"`), the error arm precedes `!enabled` so a failed first fetch stops reading as a config claim, and `viewState` gets a real `count`. On `hasLoaded` you kept it monotonic and gated `isStale` on `&& hasData` instead — that answers the actual complaint (the banner claiming a reading that never existed), so I'm fine with the alternative.

Merge note: conflicts with `dev` on `docs/memory/learnings.md` (EOF append, no code). #2379 and #2386 conflict on the same file.

One thing I'd like picked up rather than dropped, since it won't surface on its own — CI is green here because the scanner isn't a failing gate: **the raw-color ratchet moved the wrong way.** `ObservabilityPanel.vue` measures `raw_gray 82` against a committed baseline of `80`, and `raw-color-baseline.json` is untouched. The design-system contract is that per-file counts may only shrink. The PR description also reports only the improving half ("loses a raw `text-blue-600` (1 → 0)"). Either bring the count back under 80 or update the baseline with the reason stated — but the description should match whichever you do.

PR #2389, opened 2026-08-24
feat(auth): bound, attribute and fail-closed the machine credential for admin/ops APIs (#2323)

### LINKED ISSUE #2323: feat: machine identities for admin/ops APIs — service credentials that survive enforced 2FA
## Summary

Admin-scoped surfaces (`/api/ops/*`, admin-bearer `GET /api/version`, user/agent/settings admin APIs) accept only interactive-user bearer tokens minted by `POST /token`. Monitoring dashboards and fleet ops tooling therefore authenticate with an admin username + password. The moment "Require 2FA for admins" is enabled, every such consumer breaks (#2322) — there is no service-account or API-key path to these endpoints, no per-identity 2FA exemption, and MCP API keys deliberately do not cover admin endpoints. Net effect today: enforcing a recommended security control costs platform observability and ops automation, so operators are pushed to keep 2FA off — the security feature competes with security.

## Proposal

First-class machine identities for admin/ops surfaces:

- Scoped, revocable **service credentials** (API keys or an OAuth2 client-credentials grant) usable on admin/ops endpoints
- **Exempt from interactive 2FA by design** — 2FA remains a human-login control
- Audited distinctly (`audit_log` actor_type = machine/service, key id attribution, as MCP operations already are)
- Rotation without fleet-wide breakage (overlapping validity or explicit re-issue flow)

An acceptable smaller shape: extend `mcp_api_keys` with an `admin`/`ops` scope tier gated to admin creation, reusing the existing key lifecycle + audit plumbing.

## Motivation / evidence

Observed in fleet operations: enabling `require_for_admin` instantly degraded an external health dashboard (password-grant auth) to liveness-only and broke ops tooling's admin API verification steps, with no error at login time (#2322). The only recovery without this feature is leaving policy-enforced 2FA off, or storing the TOTP secret next to the password for machine clients — which defeats the second factor for that account.

## Related

- #2322 (the silent-break shape of the current behavior)
- Enterprise `two_factor` module (policy semantics are otherwise sound)

obasilakis:
## The issue's premise is false, and the correction inverts its proposed shape

#2323 asks for a service credential that survives enforced 2FA, on the premise that none exists. One does — verified against a live instance with a `scope='user'` MCP key owned by `admin`:

```
200  GET /api/version           200  GET /api/users
200  GET /api/ops/fleet/status  200  GET /api/settings/retention
200  GET /api/ops/costs         200  GET /api/audit-log/stats
200  GET /api/ops/auth-report   200  GET /api/canary/status
200  GET /api/admin/soft-deleted/agents      200  GET /api/subscriptions
PUT /api/settings/max-parallel-tasks-ceiling → 200   (writes too)
```

| Issue claim | Verified on `dev` |
|---|---|
| No API-key path to admin/ops endpoints | **False** — `require_admin`/`assert_admin` reject only *agent* and *connector*; a user-scoped key inherits the owner's role |
| MCP keys deliberately don't cover admin endpoints | **False** — true of the MCP *tool* surface, not the backend bearer surface |
| `GET /api/version` is admin-bearer | **False** — plain `get_current_user` |
| 2FA exemption is work to do | **Already true** — `gate_login` has 2 call sites, both login routes |
| Rotation needs building | **Already works** — overlapping keys both 200 |
| Audited distinctly | **Missing** — the one real ask |

The credential exists; it is **unbounded, unattributable and non-expiring**. So the only way to enable 2FA and keep a dashboard alive was to hand it a permanent, invisible, unlimited admin key — worse than the control it works around. The issue's "acceptable smaller shape" (a tier that *unlocks* admin) buys nothing, because `user` already does. The tier needed **bounds** it.

Full premise-correction and consumer evidence: [comment](https://github.com/Abilityai/trinity/issues/2323#issuecomment-5395525257), [corrections](https://github.com/Abilityai/trinity/issues/2323#issuecomment-5395690770).

## Changes

**1 — the admin gate becomes an allowlist.** ent#293/#297 closed it against two scopes *by name*. `scope` is free-text with no CHECK constraint, so that is a denylist over an open space: anything setting neither `agent_name` nor `connector_agent` inherits the owner's role across ~163 sites. `models.User`'s docstring predicted it; this PR's `ops` scope is that sixth scope. `ADMIN_GATE_SCOPES = {None, "user", "system"}` is exactly what passes today — zero behaviour change except `portal_delegate`, previously contained *only* by its route fence, which gains a second layer. An absent `mcp_scope` fails **closed** via a sentinel, never a `None` default that would make the missing attribute the privileged JWT value.

**2 — attribution from the credential, not a header.** `validate_mcp_api_key` always returned the key id/name; `get_current_user` discarded them. Deriving the three `mcp_*` audit columns from `actor_user` fixes ~70 call sites with no diff at any, plus `mcp_key_id`/`mcp_scope` query filters. `actor_type` stays `user` — the owner is accountable, is the only branch yielding an email, and the enterprise user-activity view matches on it.

`X-MCP-Key-Id` is **removed** from the audit path, not out-ranked: `Header(None)` on six routes, validated nowhere, persisted into the backlog replay blob — so honouring it let any authenticated caller forge the credential named in the platform's two highest-volume audit events, surfacing minutes later on queue drain.

**3 — `ops`: read-only, route-fenced, self-authorizing.** Admin-minted and human-only. Fenced at the single auth entry point. Every entry is a `GET`, asserted by importing the constant — without that belt a prefix entry admits a future `POST /api/ops/*`, and that is where `emergency-stop` lives. Kept *out* of the admin allowlist; ops reads opt in with `allow_scopes={"ops"}`, so authority comes from being an ops key rather than from the owner's role — it survives that admin being offboarded, and a new ops route is inaccessible until granted rather than silently reachable.

The fence set is the **measured** read set of Trinity Control, not the issue's wording (which named only `/api/ops/*` and would have shipped a credential unable to run the dashboard it exists for).

## ⚠️ This carries a security fix

`routers/a2a.py::_a2a_idem_scope` builds an **idempotency scope** from `mcp_key_id` and fell through to `username` because the field did not exist. Two agent-scoped keys of one owner therefore shared a peer-controlled `messageId` namespace — caller B received caller A's full response text and B's task never ran. Its own docstring describes the failure it was not preventing. Reachable on an entitled install with ≥2 `a2a_exposed` agents under one owner (default OFF, OSS unaffected).

The test asserting the distinction used a stub carrying a field the real principal lacked — green over a property production never had. Re-pointed at a real `User`.

**Deploy note:** the scope string moves from `a2a:{agent}:{username}` to `a2a:{agent}:{key_id}`, so a `messageId` replayed across the deploy re-executes instead of replaying. Bounded by the 24h TTL. Worth a release note.

## Test plan

- [x] `pytest tests/unit/test_2323_machine_identities.py` — 41 passed, **0 skipped**
- [x] `pytest tests/unit/test_293_admin_gate_rejects_agent_keys.py` — 25 passed
- [x] Full unit suite: **12313 passed, 21 failed — byte-identical to the failure set on an unmodified worktree at this branch point** (IPv6/SSRF, environment-dependent). Zero regressions, confirmed by a full baseline run, not by inspection.
- [x] Mutation-tested three ways, each detected: add a write to the fence → method belt red; drop a route the consumer needs → admit-set red; widen the admin allowlist → closed-literal + future-scope red.
- [x] Live probe against a running instance for every premise above.

The route-parity guard initially **skipped** (importing the app needs the full stack). A guard that skips is not a guard, so it resolves against the router modules instead. It then caught two bugs in this PR's own code on its first real execution: a doubled router prefix, and a gate using direct attribute access where a principal might not carry the field.

## Not delivered — stated, not implied

- **Key expiry.** `mcp_api_keys` still has no `expires_at`. A leaked ops key is bounded in reach, not in time.
- **`user` stays unbounded.** This adds a tier; narrowing the existing one would break the fleet.
- **No write-capable ops tier.** The ops toolkit's 24 writes stay on password auth — **the read fence does not retire the admin password.**
- **Two costs inherited, not introduced:** key validation writes per request, and fleet-status fans out per agent. The Observatory already polls exactly these endpoints on an admin JWT at the same cadence.
- **Honest bound:** this narrows the API surface only. Ops tooling that mutates containers over SSH never touches the API; SSH remains the real privilege boundary there.

## Note on 44 test doubles

38 principal stand-ins lacked `mcp_scope`; 6 stubs replaced the admin gate with a narrower signature. Every one claimed to stand for the real principal while not matching it — the same defect that hid both live bugs above. That is why the gate fails closed rather than defaulting, and the stand-ins were fixed rather than the gate softened.

Fixes #2323


dolho:
## Code review — `feature/2323-machine-identities`

Reviewed the bounded `ops` MCP-key scope, the admin-gate allowlist, and the credential-derived audit attribution. Six findings; verified-clean list at the end.

### High

**1. `src/backend/routers/subscriptions.py:166` — admitted by the fence, rejected by the gate.**
`GET /api/subscriptions/{id}/usage` is in `_OPS_ALLOWED_ROUTES` and is asserted by `test_ops_fence_admits_the_measured_consumer_read_set`, but the handler still calls bare `assert_admin(current_user)` with no `allow_scopes={"ops"}`. `ADMIN_GATE_SCOPES = {None, "user", "system"}` excludes `"ops"`, so `_reject_scope_at_admin_gate` raises 403.

Concretely: an ops key passes the route fence for `/api/subscriptions/sub-1/usage` and is then rejected at the admin gate — the "subscription pressure" read the fence was explicitly measured for cannot work. The admit-set test gives false assurance because it exercises only `_enforce_ops_key_fence`, never the endpoint's own gate. Every other admitted path is either `get_current_user`-only or got the opt-in; this is the one gap.

Fix: `assert_admin(current_user, allow_scopes={"ops"})` there, and extend the test to drive the endpoint gate rather than just the fence.

### Medium

**2. `src/backend/main.py:1284` — `/ws/events` bypasses the fence.**
The handler calls `db.validate_mcp_api_key(token)` directly and never runs `_enforce_ops_key_fence`, so the fence's own docstring claim ("enforced HERE at the single auth entry point") is not true. An ops key can open `wss://.../ws/events?token=trinity_mcp_*` and receive the fleet-wide `agent_activity` / execution event stream, scoped by the *owner's* accessible agents (everything, for an admin owner) — a surface outside the declared read allowlist. The same hole already exists for `connector` and `portal_delegate` (pre-existing), but the fence is this PR's value proposition, so it's worth either closing or documenting explicitly.

**3. `src/backend/services/chat_execution_service.py:176` (also `:1113`, `:1464`) — forgeable header still reaches durable storage.**
The PR removes `X-MCP-Key-Id` from the *audit* path but leaves `source_mcp_key_id=x_mcp_key_id` / `source_mcp_key_name=x_mcp_key_name` feeding `db.create_task_execution`, and `backlog_service.py:109` still persists both into the replay blob.

Concretely: any authenticated caller sends `X-MCP-Key-Id: <someone-else's-key>` on `POST /api/agents/{n}/chat`; the audit row now correctly names the presented bearer while `schedule_executions.source_mcp_key_id` durably records the forged value. Two provenance surfaces for one request disagree, and the threat model this PR states ("validated nowhere") applies unchanged to the column that survives longest.

### Low

**4. `src/backend/db/mcp_keys.py:139` — the `_AGENTLESS_SCOPES` guard can never fire.**
It reads `getattr(key_data, "agent_name", None)`, but `McpApiKeyCreate` (`db_models.py:87`) declares only `name` / `description` / `scope`, and Pydantic silently drops unknown fields. A caller POSTing `{"scope":"ops","agent_name":"atlas"}` never reaches the guard. `test_ops_keys_must_never_carry_an_agent_name` only asserts membership in the constant, not the behaviour — so it documents a protection that does not exist.

**5. `src/backend/services/dispatch_admission_service.py:124-125` — dead forgeable parameters.**
`admit_chat_request` still declares `x_mcp_key_id` / `x_mcp_key_name` and `routers/chat.py:172` still passes them, but nothing in the body reads them. This contradicts the rule stated one function below ("a parameter nothing reads is how a forgeable value quietly finds a new consumer later") — the same reason `audit_idempotent_replay` dropped them from its signature.

**6. `tests/unit/test_1310_auth_consolidation.py:172` — helper kwarg silently ignored.**
`_user()` gained an `mcp_scope=None` keyword that is never forwarded to the `User(...)` construction on the next line, and no caller passes it. A future test writing `_user("bob", mcp_scope="ops")` silently gets `mcp_scope=None` — the privileged JWT value in `ADMIN_GATE_SCOPES` — and asserts a pass the real principal would not get. That's precisely the stand-in-does-not-match-production defect this PR exists to eliminate.

### Verified clean

- Every `_OPS_ALLOWED_ROUTES` entry resolves to a real registered `GET` route (`/api/agents/execution-stats`, `/slots`, `/{n}/executions` in `schedules.py`, `/{n}/executions/{id}/stream` in `chat.py`, both telemetry routes, `/api/monitoring/status`, both executions routes).
- All six `routers/ops.py` opt-ins are on GET handlers; every write stays bare.
- `validate_mcp_api_key` really returns `key_id` / `key_name`.
- `db.get_audit_entries` / `count_audit_entries` are `**filters` passthroughs, so the two new query params can't `TypeError`.
- `require_admin`'s signature is unchanged, so all `Depends` sites are unaffected; all six `assert_admin` monkeypatches in the test tree were updated to `lambda user, **kw`.
- The MCP server's `OPERATOR_SCOPES` is an allowlist, so `"ops"` gets no operator tools.
- `User` is not a response model anywhere, so the new key fields don't leak to clients.
- The a2a idempotency-scope fix lands correctly via the model change alone.

obasilakis:
All six findings fixed in b996f3c7. Each was verified against the code before touching it; three of them were bigger than reported, so here is what actually changed.

### 1 (High) — admitted by the fence, refused by the gate ✅
`assert_admin(current_user, allow_scopes={"ops"})` on `GET /api/subscriptions/{id}/usage`.

Your last line is the important one: the admit-set test exercised only `_enforce_ops_key_fence`, so it asserted one half of a two-gate path and gave false assurance. Rather than extend that one test, this guards the **class** — `test_every_admin_gated_allowlisted_route_opts_ops_in` resolves every allowlisted GET from the live router modules (`inspect.getsource` on the endpoint object, so it follows a route that moves) and reds if any of them calls `assert_admin` without the opt-in. It asserts a non-empty resolved set first, so it cannot pass vacuously.

### 2 (Medium) — `/ws/events` bypassed the fence ✅ closed, not documented
`dependencies.WS_EVENT_STREAM_SCOPES = {None, "user", "agent", "system"}`, checked right after `validate_mcp_api_key` in the handler; a bounded scope gets close code 4003.

Closed rather than documented for the reason you gave — the fence is this PR's value proposition, and a docstring claiming "enforced at the single auth entry point" while a second entry point exists is worse than no claim. The pre-existing `connector` / `portal_delegate` hole is closed by the same allowlist: both are fenced to one or two routes everywhere else, so nothing legitimate reaches this one, and an unknown future scope is refused by construction.

### 3 (Medium) — forgeable header reaching durable storage ✅ and it was wider than the three lines
You were right, and the surface was larger than `chat_execution_service`. `X-MCP-Key-Id` was still declared as `Header(None)` on **five** routers and written into durable provenance in four places:

| Writer | Column |
|---|---|
| `routers/chat.py` (`/chat`, `/task`) | `schedule_executions.source_mcp_key_id` |
| `routers/schedules.py` (manual trigger) | `schedule_executions.source_mcp_key_id` |
| `routers/loops.py` | `agent_loops.source_mcp_key_id` |
| `routers/reminders.py` | `agent_reminders.source_mcp_key_id` |
| `routers/fan_out.py` | fan-out execution rows |

plus `backlog_service` persisting both names into `backlog_metadata` — and `_spawn_drain` never reads either key back, so it was a forgeable value stored, never reconstructed, in the one blob canary G-04 scans and #1449 scrubs.

All five now derive from `current_user.mcp_key_id`/`mcp_key_name`, and the fix is asserted in the strongest available form: **no router declares either header at all**, checked by a scan over the whole router tree. Grepping for the *use* is what let this survive #2323; a value nothing declares cannot be forged, threaded, or picked up by a new consumer.

The dead parameters are removed from the entire `chat_execution_service` → `capacity_manager` → `backlog_service` chain (finding 5's rule, applied at scale — 17 signatures), not accepted-and-ignored. A pre-existing queued row still carrying the two blob keys drains unchanged, since the drain reads the blob key-by-key with `.get()`.

The MCP server still *sends* the headers. Left alone deliberately: nothing on the backend reads them, and the value it sends is the same key the bearer already identifies — so removing it is an MCP rebuild for no behaviour change, and keeping it is one less rolling-deploy edge.

### 4 (Low) — the guard that could never fire ✅
Correct, and the right fix was deletion. `getattr(key_data, "agent_name", None)` on a model that declares no such field is permanently falsy — protection in appearance only, which is worse than none. The guard is gone; `_AGENTLESS_SCOPES` stays as the documented invariant, and `test_ops_keys_must_never_carry_an_agent_name` (which asserted a tuple contains a string) is replaced by a test that drives the real creator and `SELECT`s the written row.

### 5 (Low) — dead parameters on `admit_chat_request` ✅
Removed, along with the router's pass-through.

### 6 (Low) — `_user(mcp_scope=...)` swallowed ✅
Forwarded. Same defect in `test_connector_auth.py::_user`, fixed with it.

---

**Verification:** full unit suite **12383 passed / 21 failed**, against a baseline run of the branch point in a clean worktree at **12373 passed / 21 failed** — byte-identical failure set (IPv6/SSRF, environment-dependent). Zero regressions, +10 new passing. Docs updated in `architecture.md` (scope table + the `/ws/events` second-entry-point note) and `requirements/security.md`.

dolho:
## Code review — re-review at `5bbec15`

Traced the `ops` scope end to end: fence → admin gate → each allowlisted route's own gate → MCP-server scope handling → `/ws/events` → audit derivation → backlog payload drain.

The previous round's items are fixed — `/api/subscriptions/{id}/usage` now opts in (the fence-admits-gate-rejects gap), `/ws/events` is scope-allowlisted, and the forgeable `X-MCP-Key-Id` no longer reaches `source_mcp_key_id`. Verified rather than assumed: the fence is anchored and method-belted, every allowlisted route resolves to a real handler and is reachable by an ops key, the dropped `PersistentTaskPayload` fields are safe because `_spawn_drain` reads the blob with `.get()`, and `_resolve_actor` short-circuits on `actor_user` so the new `mcp_scope` derivation cannot flip `actor_type` across ~70 call sites.

Four findings.

### Medium

**1. `dependencies.py:823` (and `docs/memory/requirements/security.md:314`) — the documented "survives the minting admin being offboarded" property is not implemented.**

Both the code comment and the requirements doc claim authority comes from *being an ops key*, not from who owns it. It does not: `assert_admin(..., allow_scopes={"ops"})` runs `_reject_scope_at_admin_gate` and **then still** enforces `current_user.role != "admin"`, and `get_current_user` rejects the principal outright when the owner carries `suspended_at`.

So: admin Alice mints an ops key for the monitoring dashboard and later leaves. Offboarding demotes or suspends her account, and the ops key immediately 403s on all six `/api/ops/*` reads and on `/api/subscriptions/{id}/usage` — precisely the "every ops integration dies when that admin is offboarded" outcome the doc says this design prevents. An operator will act on that sentence during an offboarding. Either drop the role check for the opted-in scope, or correct both claims.

**2. `dependencies.py:1001` — `require_admin` has no `allow_scopes`, and the new CI guard cannot see it.**

`require_admin` is the `Depends` form and the dominant admin-gate spelling platform-wide, but it gained no `allow_scopes` parameter — so an ops-allowlisted route gated that way is permanently dead to ops keys with no opt-in available. And `tests/unit/test_2323_machine_identities.py:365` only regex-matches `assert_admin\(` inside handler source, so it cannot see `Depends(require_admin)` in a signature.

Someone converts `/api/ops/fleet/status` to `Depends(require_admin)` for consistency with the enterprise routers: the ops dashboard silently 403s, and CI stays green. That is exactly the two-gate false-assurance class the guard was written for after the `/subscriptions/{id}/usage` bug — the guard closed the instance and left the shape open. Either give `require_admin` a parameterised form, or have the guard reject any allowlisted handler whose source mentions `require_admin`.

### Low

**3. `dependencies.py:879` — `WS_EVENT_STREAM_SCOPES` admits `"agent"` wholesale, including ephemeral ghost keys.**

`/ws/events` authenticates the key itself and never calls `get_current_user`, so `_enforce_ephemeral_key_fence` — whose entire purpose is that a ghost's key on an untrusted workspace must not be a fleet skeleton key — does not reach it. A ghost's own `TRINITY_MCP_API_KEY` therefore opens the fleet-wide event stream, scoped by `get_accessible_agent_names(owner_email, is_admin)`, i.e. everything on a default admin-owned install.

Pre-existing, but this PR is the one that enumerated the permitted scopes on that surface and closed the same hole for `connector` and `portal_delegate`; ghost keys are the one bounded principal left in. Gating `"agent"` on the row's `is_ephemeral` (the predicate the fence already uses) would finish it.

**4. `settings/McpKeysTab.vue:496` — the UI advertises the tier but cannot mint it.** The tab renders an "Ops (read-only)" badge, but the create modal only ever sends `scope: 'portal_delegate'` — there is no ops option in `newKey`. An operator reads the release note, opens Settings → MCP Keys, and finds no way to create the credential the badge describes. The tier is API-only.

### Minor, no action needed

- The four provenance comments in `routers/{fan_out,loops,reminders,schedules}.py` each start with a doubled marker: `# # #2389:`.
- `db/mcp_keys.py::_AGENTLESS_SCOPES` is defined and referenced nowhere — deliberate per its own comment, noted here only so the next reader does not mistake it for an enforced guard.
- `src/mcp-server/src/client.ts` still sends the now-ignored `X-MCP-Key-ID` / `X-MCP-Key-Name` headers. Harmless, but dead.

src/backend/dependencies.py
```diff
@@ -784,6 +798,111 @@ def _enforce_ephemeral_key_fence(request: Request, agent_name: str) -> None:
...
+#   1. This route allowlist, enforced HERE at the single auth entry point rather
+#      than per-endpoint — the ent#293/#297 lesson (five occurrences meant the
+#      GATE was wrong, not the endpoints), and the same placement as the
+#      connector, ephemeral and portal_delegate fences.
+#   2. `ADMIN_GATE_SCOPES` above, which keeps `ops` OUT of the admin gates by
+#      default; an admin-gated ops route opts in explicitly with
+#      `assert_admin(..., allow_scopes={"ops"})`.
+#
+# Layer 2 is what makes this a machine identity rather than a human's proxy:
+# authority comes from being an ops key, not from who owns it. It also flips the
```
dolho: **MEDIUM — the documented "survives the minting admin being offboarded" property is not implemented.**

Both this comment and `docs/memory/requirements/security.md:314` claim authority comes from *being an ops key*, not from who owns it. It does not: `assert_admin(..., allow_scopes={"ops"})` runs `_reject_scope_at_admin_gate` and **then still** enforces `current_user.role != "admin"`, and `get_current_user` rejects the principal outright when the owner carries `suspended_at`.

So: admin Alice mints an ops key for the monitoring dashboard and later leaves. Offboarding demotes or suspends her account, and the ops key immediately 403s on all six `/api/ops/*` reads and on `/api/subscriptions/{id}/usage` — precisely the "every ops integration dies when that admin is offboarded" outcome the doc says this design prevents.

An operator will act on that sentence during an offboarding. Either drop the role check for the opted-in scope, or correct both claims.

src/backend/dependencies.py
```diff
@@ -1212,6 +1413,7 @@ def assert_admin(current_user: User, *, detail: str = "Admin access required") -
     """
     _reject_connector_principal(current_user)
     reject_agent_principal(current_user)
+    _reject_scope_at_admin_gate(current_user, allow_scopes)
     if current_user.role != "admin":
         raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail=detail)
```
dolho, line 1522: _(anchored to the nearest line in this PR's diff; the finding is at `src/backend/dependencies.py:1001`)_

**MEDIUM — `require_admin` has no `allow_scopes`, and the new CI guard cannot see it.**

`require_admin` is the `Depends` form and the dominant admin-gate spelling platform-wide, but it gained no `allow_scopes` parameter — so an ops-allowlisted route gated that way is permanently dead to ops keys with no opt-in available. And `tests/unit/test_2323_machine_identities.py:365` only regex-matches `assert_admin\(` inside handler source, so it cannot see `Depends(require_admin)` in a signature.

Someone converts `/api/ops/fleet/status` to `Depends(require_admin)` for consistency with the enterprise routers: the ops dashboard silently 403s, and CI stays green.

That is exactly the two-gate false-assurance class the guard was written for after the `/subscriptions/{id}/usage` bug — the guard closed the instance and left the shape open. Either give `require_admin` a parameterised form, or have the guard reject any allowlisted handler whose source mentions `require_admin`.

src/backend/dependencies.py
```diff
@@ -784,6 +798,111 @@ def _enforce_ephemeral_key_fence(request: Request, agent_name: str) -> None:
...
+# activity and execution events scoped by the OWNER's accessible agents, which
+# for an admin owner is everything, so it is exactly the kind of broad read the
+# bounded scopes exist to exclude.
+#
+# `ops` is absent because the stream is not in its read allowlist. `connector`
+# and `portal_delegate` are absent for the same reason their route fences deny
+# them everything else — this closes a hole that predates #2323 rather than
+# preserving it out of politeness; both are fenced to one or two routes, so
+# nothing legitimate reaches this one.
+WS_EVENT_STREAM_SCOPES = frozenset({None, "user", "agent", "system"})
```
dolho, line 914: **LOW — `WS_EVENT_STREAM_SCOPES` admits `"agent"` wholesale, including ephemeral ghost keys.**

`/ws/events` authenticates the key itself and never calls `get_current_user`, so `_enforce_ephemeral_key_fence` — whose entire purpose is that a ghost's key on an untrusted workspace must not be a fleet skeleton key — does not reach it. A ghost's own `TRINITY_MCP_API_KEY` therefore opens the fleet-wide event stream, scoped by `get_accessible_agent_names(owner_email, is_admin)`, i.e. everything on a default admin-owned install.

Pre-existing, but this PR is the one that enumerated the permitted scopes on that surface and closed the same hole for `connector` and `portal_delegate`; ghost keys are the one bounded principal left in. Gating `"agent"` on the row's `is_ephemeral` (the predicate the fence already uses) would finish it.

src/frontend/src/components/settings/McpKeysTab.vue
```diff
@@ -88,6 +88,15 @@
...
+                       unbounded personal key here, which defeats the point of
+                       having a bounded tier at all. -->
+                  <span v-else-if="key.scope === 'ops'"
+                        class="ml-2 inline-flex items-center px-2 py-0.5 rounded-full text-xs font-medium bg-status-info-100 dark:bg-status-info-900/50 text-status-info-800 dark:text-status-info-300"
+                        title="Read-only machine credential — fleet health, telemetry, execution and subscription reads only">
+                    Ops (read-only)
+                  </span>
                   <span v-else-if="key.scope === 'system'" class="ml-2 inline-flex items-center px-2 py-0.5 rounded-full text-xs font-medium bg-status-urgent-100 dark:bg-status-urgent-900/50 text-status-urgent-800 dark:text-status-urgent-300">
                     System
                   </span>
```
dolho, line 102: _(anchored to the nearest line in this PR's diff; the finding is at `src/frontend/src/components/settings/McpKeysTab.vue:496`)_

**LOW — the UI advertises the tier but cannot mint it.**

The tab renders an "Ops (read-only)" badge, but the create modal only ever sends `scope: 'portal_delegate'` — there is no ops option in `newKey`. An operator reads the release note, opens Settings → MCP Keys, and finds no way to create the credential the badge describes. The tier is API-only.

src/backend/routers/loops.py
```diff
@@ -176,8 +174,9 @@ async def start_loop(
...
         started_by_user_email=current_user.email,
         source_agent_name=x_source_agent,
-        source_mcp_key_id=x_mcp_key_id,
-        source_mcp_key_name=x_mcp_key_name,
+        # # #2389: the credential actually presented, never the forgeable X-MCP-Key-* headers.
+        source_mcp_key_id=getattr(current_user, "mcp_key_id", None),
+        source_mcp_key_name=getattr(current_user, "mcp_key_name", None),
     )
     return StartLoopResponse(
         loop_id=loop_row["id"],
```
dolho, line 182: _(anchored to the nearest line in this PR's diff; the finding is at `src/backend/routers/loops.py:1`)_

**MINOR — doubled comment marker.**

The provenance comment here (and the matching ones in `routers/fan_out.py`, `routers/reminders.py` and `routers/schedules.py`) starts `# # #2389:`.

src/backend/db/mcp_keys.py
```diff
@@ -106,7 +106,26 @@ def _row_to_mcp_api_key(row) -> McpApiKey:
...
+    #
+    # The invariant holds HERE BY CONSTRUCTION: this creator writes the literal
+    # `agent_name=None` below, and `McpApiKeyCreate` declares no `agent_name` for
+    # a caller to supply. An earlier revision guarded on
+    # `getattr(key_data, "agent_name", None)` — which Pydantic makes permanently
+    # falsy, so it could never fire while reading as protection. Removed in
+    # review; the constant stays as the documented invariant and
+    # `test_2323_machine_identities.py` asserts the WRITTEN ROW rather than
+    # membership in this tuple.
+    _AGENTLESS_SCOPES = ("user", "portal_delegate", "ops")
```
dolho: **MINOR, no action — `_AGENTLESS_SCOPES` is defined and referenced nowhere.**

Deliberate per its own comment ("holds HERE BY CONSTRUCTION"). Noted only so the next reader does not mistake it for an enforced guard — a second minting path could write an `agent_name` for a non-agent scope with nothing to stop it.

obasilakis:
All four fixed in 1f4b3ea7, plus the three minors. One of them I fixed by **correcting the claim rather than implementing it** — reasoning below, because it is a design decision and not a typo.

### 1 (Medium) — the offboarding property ✅ corrected, deliberately not implemented

You are right on the facts: `assert_admin(..., allow_scopes={"ops"})` admits the scope and *then* still enforces `role == "admin"`, and `get_current_user` rejects a suspended owner one layer above. The sentence was mine, it was wrong, and "an operator will act on it during an offboarding" is exactly the damage.

I took the second of your two options. Dropping the role check for the opted-in scope was considered and refused for two reasons:

- **It would make the bounded tier harder to revoke than the unbounded one it exists to displace.** A `user`-scoped key loses admin the moment its owner is demoted. An ops key that ignored the owner's role would not — so the *narrower* credential would survive a demotion the *wider* one does not. Wrong direction for a tier whose whole pitch is bounds.
- **It still could not deliver the claim.** Suspension is the usual offboarding action and it kills the key at the auth entry point (#995), before any of this runs. So the change would have half-delivered a property while leaving the same misleading sentence approximately true.

What the opt-in actually buys stands unchanged and is now stated as such: the grant is **per route**, so a new ops route is inaccessible until someone adds it. The opt-in is an *additional* gate, never a substitute one — an ops key is a narrowing of its owner, not a decoupling from them.

The operator consequence is now written where it will be read (code comment, `requirements/security.md`, `architecture.md` scope table): **an ops key is bound to the account that minted it — mint it under a service admin account that is not offboarded with people, and put revoke-and-re-mint in the offboarding runbook.** `test_an_opted_in_ops_key_still_needs_its_owner_to_be_admin` pins both directions (demoted owner 403s; admin owner passes) so the claim cannot drift back into the docs.

Worth noting the scope of what #2323 asked for: a credential that survives enforced **2FA** — which it does, key validation never passing through the MFA gate. Surviving its *owner* was an embellishment I added, not a requirement.

### 2 (Medium) — `require_admin` had no opt-in, and the guard could not see it ✅

Both halves, since the guard closing only the instance was your point:

- `require_admin_allowing("ops")` is the `Depends` form. It delegates the **entire** ladder to `assert_admin` rather than restating it — two admin gates that must stay identical is precisely how the `require_role("admin")` third spelling happened — and it matches `require_role`'s factory shape exactly.
- `test_no_allowlisted_route_uses_the_unparameterised_require_admin` resolves the same live handler set as the existing guard and reds on `require_admin(?!_allowing)` in any allowlisted handler's source. Your `/api/ops/fleet/status` conversion scenario now fails CI instead of silently 403ing the dashboard.

### 3 (Low) — ghost keys on the event stream ✅ closed

`scope_may_open_event_stream(scope, agent_name)` now refuses an `is_ephemeral` row, using `_enforce_ephemeral_key_fence`'s own predicate; `/ws/events` passes `key_info["agent_name"]`.

Two things I want on the record because they are judgement calls, not mechanics:

- **The sub-check fails CLOSED**, inverting the ephemeral fence's fail-open. That fence guards heartbeats and result callbacks, where a DB blip must not take the fleet down; losing this stream costs an observability client a reconnect. Different blast radius, different direction.
- **Presence of the `is_ephemeral` key is what makes the answer real.** `get_agent_ephemeral_info` coalesces the column for every live row, so a dict without it is *no row at all* — and a bare `.get()` maps that onto the same falsy value a genuine durable agent gives. Tested with `{}` explicitly, alongside `None`, a non-dict, a raising lookup, and a nameless `agent`-scoped key. An ordinary agent key still opens the stream, so the gate is not satisfied by denying everyone.

### 4 (Low) — the UI advertised a tier it could not mint ✅

The create modal is now a three-option scope selector (Standard / Ops / Portal delegate), admin-only, with the ops copy stating the bound honestly — including that the key stops working if the account's admin role is removed or the account is suspended, per finding 1.

A radio group rather than a second checkbox: the scopes are mutually exclusive because the column holds one value, so independent booleans can express a state the backend cannot store. The shared chrome is hoisted into two scoped classes so the file's raw-color ratchet count stays **flat at its baseline of 97** — adding a three-option control would otherwise have pushed it to 107, and the ratchet may only shrink.

### Minors

- The four `# # #2389:` markers are single again.
- `_AGENTLESS_SCOPES` left as-is; its own comment already says it is a documented invariant and not an enforced guard, which is what you were guarding the next reader against.
- The MCP client's `X-MCP-Key-*` sends stay, but each of the four sites now carries a note saying they are **inert** (no router declares the headers, asserted by a scan), why they are kept — the value sent is the same key the bearer already identifies, so dropping it is an MCP rebuild for zero behaviour change — and that a backend reader must never return. Dead-but-explained beats dead-and-puzzling.

---

**Verification:** full unit suite **12392 passed / 21 failed**, against **12383 / 21** at `5bbec159`. Identical failure set (IPv6/SSRF, environment-dependent); +9 net new passing (10 added, one parametrize case moved to its own tests). Docs updated in `architecture.md` (scope table + the `/ws/events` block) and `requirements/security.md`.

dolho:
## Code review — re-review at `1f4b3ea`

The core holds up: the admin gate is an allowlist, audit attribution comes from the presented bearer rather than a forgeable header, and the `ops` fence is enforced at the single auth entry point. Verified rather than assumed — every fenced route resolves to a real handler; every allowlisted route's own gate now admits `ops` (the `/api/subscriptions/{id}/usage` opt-in was a real bug and the fix is right); the backlog drain reads its blob with `.get()`, so dropping the two payload fields is upgrade-safe; `_resolve_actor` short-circuits on `actor_user`, so `actor_type` genuinely stays `user`; and `_AGENTLESS_SCOPES` / `_USER_CREATABLE_SCOPES` cannot let `ops` carry an `agent_name`.

Six findings.

### Medium-high

**1. `main.py:1294` — the new `/ws/events` allowlist still admits ephemeral ghost keys, bypassing the ent#69 fence.**

`WS_EVENT_STREAM_SCOPES` includes `"agent"`, and this handler never calls `get_current_user`, so `_enforce_ephemeral_key_fence` is not reached either. A prompt-injected ghost — whose entire containment story is that its key reaches heartbeat, result callback, reports and self-info and nothing else — can run `ws://backend:8000/ws/events?token=$TRINITY_MCP_API_KEY` and receive fleet-wide `agent_activity` / `schedule_execution_completed` / `agent_collaboration` events, scoped to its **owner's** accessible agents: everything, on a default admin-owned install.

This PR is the one that added a gate here, and its own comment argues that a bounded credential must not read fleet-wide activity through an unlisted surface. The ephemeral bound is exactly such a credential and is the one left out. Reject when the key's `agent_name` resolves to an `is_ephemeral` agent — the predicate `_enforce_ephemeral_key_fence` already uses — or route this handler through a shared fence helper.

### Medium

**3. `settings/McpKeysTab.vue:496` — the ops scope is badged but cannot be created from the UI.** `createKey` only ever sends `{ scope: 'portal_delegate' }`, and the create modal has no ops control; the only new UI is the read-only badge at :95. An admin who wants the credential this PR exists for has to `curl POST /api/mcp/keys -d '{"scope":"ops"}'`. The verb the feature implies is not reachable end to end.

**4. `routers/mcp_keys.py:52` — `portal_delegate` still doesn't require an interactive session, while the strictly weaker `ops` now does.**

The new comment argues correctly that `reject_agent_principal` + `assert_admin` are no-ops for a scope that sets neither `agent_name` nor `connector_agent`. That reasoning applies identically to `portal_delegate` — which is the *more* dangerous credential, since it impersonates any portal user. An admin's `user`-scoped MCP key can still mint one; an ops key cannot reach the route only because of the route fence, which is a different mechanism doing accidental duty. Moving `reject_non_interactive_principal(current_user)` above the `if requested_scope == OPS_SCOPE` branch covers both.

### Low-medium

**2. `dependencies.py:1417` (and `docs/memory/feature-flows/mcp-api-keys.md`, the `routers/ops.py` comments) — the offboarding claim is still not what the code does.**

"Authority comes from being an ops key … which is why it keeps working when the owning admin is offboarded." But `assert_admin` runs `_reject_scope_at_admin_gate(..., allow_scopes={"ops"})` and *then* still `if current_user.role != "admin": raise`, and `get_current_user` separately rejects a suspended owner. Demoting or suspending the minting admin silently 403s every `/api/ops/*` read — the exact scenario the docs promise survives.

Raised last round and worth repeating because of who acts on it: an operator following the documented offboarding procedure takes the dashboard down. Either drop the role check for an opted-in bounded scope, or correct the claim in all three places.

### Low

**5. `dependencies.py:898` — `HEAD` requests to allowlisted routes are refused.** Starlette auto-registers `HEAD` for every `GET` route, so a `HEAD /api/version` — a common liveness-probe shape — reaches `get_current_user`, and `method == allowed_method` never matches because every entry is literally `"GET"`. The same URL succeeds under `GET`, so this reads as a flaky fence rather than a policy. Accept `HEAD` wherever `GET` is allowed, or say so.

**6. `tests/unit/test_2323_machine_identities.py:365` — the parity guard still only inspects `assert_admin(...)` call text.** `require_admin` takes no `allow_scopes`, so converting any allowlisted route to the `Depends` form — the more common spelling in this codebase — silently loses ops keys while `test_every_admin_gated_allowlisted_route_opts_ops_in` stays green. That is the same "guard that asserts only one half" failure this PR describes finding for the fence-vs-gate bug. Extend the scan to the route's `dependant` chain, or make `require_admin` refuse to be used on an allowlisted path.

### Minor

- `db/mcp_keys.py:128` — `_AGENTLESS_SCOPES` is referenced by nothing. The comment says it holds by construction, but a reader will assume it is live.
- `src/mcp-server/src/client.ts:727,892,1069,1273` — the MCP server still sends `X-MCP-Key-ID` / `X-MCP-Key-Name` on four call paths the backend no longer declares or reads. Harmless today (the bearer is always `authContext.mcpApiKey`, so attribution matches), but it leaves the forgeable header alive on the third surface after the backend guard banning it landed OSS-side only.

dolho:
## Correction — my previous comment reviewed a stale tree

Apologies: [my re-review above](https://github.com/Abilityai/trinity/pull/2389#issuecomment-5412189266) was produced against a checkout that predates `1f4b3ea`, even though the PR head was already `1f4b3ea` when it ran. Four of its six findings are things that commit fixes. I have since checked each one by hand against `origin/feature/2323-machine-identities`.

**Withdrawn — already fixed at `1f4b3ea`:**

- **Ghost keys on `/ws/events`.** Fixed. `scope_may_open_event_stream(scope, agent_name)` gates the `agent` arm on `db.get_agent_ephemeral_info`, and `main.py:1299` passes `key_info.get("agent_name")` precisely because the missing `get_current_user` also skips `_enforce_ephemeral_key_fence`. The `not isinstance(info, dict) or "is_ephemeral" not in info → False` branch is the right call: it refuses to let a stand-in row read as a genuine non-ephemeral one.
- **`require_admin` has no `allow_scopes`.** Fixed — `require_admin_allowing(*allow_scopes)` (`dependencies.py:1113`) is the `Depends` form.
- **The ops scope can't be minted from the UI.** Fixed — `McpKeysTab.vue:225` has the radio, with the account-binding caveat in the helper text.
- **The offboarding claim in `dependencies.py`.** Fixed, and better than I asked for: :831-850 now states plainly that it does *not* survive offboarding, gives the reason (suspension kills the key via #995), and prescribes minting under a dedicated service admin with revoke-and-re-mint in the runbook.

**Still open, verified at `1f4b3ea`:**

1. **`docs/memory/feature-flows/mcp-api-keys.md:40` still carries the retracted claim** — *"Authority comes from being an ops key — which is why it keeps working when the owning admin is offboarded"*. The code comment was corrected and the flow doc was not, so the two now contradict each other, and the doc is the surface an operator reads while writing an offboarding runbook. This is the one I would fix before merge; it is one sentence.

2. **The parity guard is now blind to the form this PR just added.** `test_2323_machine_identities.py:375` regexes `assert_admin\(([^)]*)\)` over handler source. `require_admin_allowing` appears in a signature, not a call in the body — so an allowlisted route written in the new `Depends` style passes the guard vacuously and would silently lose ops keys. That is the same one-half-asserted shape the guard exists to catch, and it got slightly more likely the moment the second spelling shipped.

3. **`HEAD` on an allowlisted route is refused** (`dependencies.py`, `_OPS_ALLOWED_ROUTES` entries are literally `"GET"` and the loop is `if method != allowed_method`). Starlette auto-registers `HEAD` for every `GET`, so `HEAD /api/version` — a common liveness-probe shape — 403s while `GET` on the same URL succeeds. Low, but it reads as a flaky fence rather than a policy.

4. Minor, unchanged: `_AGENTLESS_SCOPES` (`db/mcp_keys.py:128`) is referenced by nothing, and `src/mcp-server/src/client.ts` still sends `X-MCP-Key-ID` / `X-MCP-Key-Name` on four paths the backend no longer reads.

Sorry for the noise — the withdrawn four were real work already done, and the comment should not have implied otherwise.

dolho (approved):

obasilakis:
Thanks for going back and checking by hand — the correction is more useful than the review.

`1dc0607a` closes three of the four. The fourth is refuted with a measurement rather than fixed.

### 1 — the flow doc carried the retracted claim ✅
Right, and it was the worse half of the pair: the code comment was corrected and the *doc* is what someone reads while writing an offboarding runbook. `docs/memory/feature-flows/mcp-api-keys.md` now says the same thing as `dependencies.py:831-850` — does not survive the owner, both reasons (demotion 403s at the gate; `suspended_at` kills the key one layer up per #995), why dropping the role check was refused, the 2FA-vs-owner distinction, and mint-under-a-service-admin as the operator consequence.

### 2 — the guard's blind spot ✅, though smaller than reported
`1f4b3ea` had already added `test_no_allowlisted_route_uses_the_unparameterised_require_admin`, which covers the `Depends(require_admin)` case you named — worth pointing out since your comment reads as though nothing did.

What neither guard caught was `require_admin_allowing("telemetry")` — the opt-in form, present, parameterised with the wrong scope. All three spellings now resolve through one `_ops_gate_offences` predicate:

```
assert_admin(user, allow_scopes={"ops"})      imperative
Depends(require_admin_allowing("ops"))        declarative
Depends(require_admin)                        declarative, no opt-in exists
```

The sibling test delegates to that predicate rather than keeping a second copy of the regex — two guards over one policy, each with its own pattern, is how the spellings diverged in the first place. `test_the_ops_gate_guard_detects_every_spelling` pins each offending form *and* each correct one, so it can neither go blind again nor be satisfied by rejecting everything. Live mutation check: reverting the `/api/subscriptions/{id}/usage` opt-in reds it by name.

### 3 — `HEAD` on an allowlisted route ❌ not a defect
The premise is Starlette's `Route`, which does add `HEAD` alongside `GET`. FastAPI's `APIRoute` does not. Measured on this stack:

```
GET  /api/version -> 200
HEAD /api/version -> 405
```

405 for **every** principal, admin JWT included. Routing rejects the method before dependencies are solved, so `_enforce_ops_key_fence` never runs and cannot be what refuses it — there is no ops-key-specific behaviour to be flaky. A liveness probe using `HEAD` is already broken against this API for everyone, which is a different (and pre-existing) question.

Adding `"HEAD"` to `_OPS_ALLOWED_ROUTES` would therefore fix nothing and widen the method belt for a method the app does not serve — and that belt is the thing standing between an allowlisted prefix and a future `POST /api/ops/*`, where `emergency-stop` and `fleet/stop` live. Left alone deliberately.

### 4 — the two minors
- **`_AGENTLESS_SCOPES` deleted.** You were right that it is unreferenced, and the fix is deletion rather than a reader: removing the guard that consumed it left a tuple nothing consults, which reads as enforcement to the next person — the same defect one step smaller. The invariant holds by construction (this creator writes the literal `agent_name=None`; `McpApiKeyCreate` declares no field for a caller to supply) and is now carried by the comment plus the written-row assertion.
- **`client.ts` still sends `X-MCP-Key-ID`/`-Name`.** Deliberate, and recorded in `requirements/security.md` rather than left implicit: nothing on the backend reads them, and the value it sends is the same key the bearer already identifies — so removing it is an MCP rebuild for zero behaviour change and one more rolling-deploy edge (MCP updated before backend would drop attribution on the old backend, which still reads the header). Happy to strip it if you'd rather not carry the dead field.

CI green on `1f4b3ea` (all six pytest shards); re-running on `1dc0607a`.

Issue #2283, opened 2026-08-18
feat: community one-click paths — "Deploy on Hostinger" button + Dokploy template (Coolify queued at 1,000 stars)


vybe:
## Summary

Add the two zero-cost community one-click paths that only need a pull-only compose file: a **"Deploy on Hostinger" button** (compose-URL + referral code) in the README/docs, and a **Dokploy template** PR to `Dokploy/templates`. Queue the equivalent **Coolify** service template for when the repo passes Coolify's 1,000-star contribution gate.

## Context

Hostinger has no vendor program — its 1,000+ VPS/Docker templates are Hostinger-authored — but it publishes an open "Deploy on Hostinger" button: `hostinger.com/docker-hosting?compose_url=<raw compose URL>&REFERRALCODE=<code>` drops the user into Docker Manager on a KVM VPS with our compose pre-loaded (referral: 20% off for them, 20% of the sale back). Dokploy's catalog accepts PRs (compose + `template.toml` + logo; recent adds merged in 1–3 days) and already has 29 Docker-socket templates including OpenHands, the closest analog to Trinity. Coolify's catalog is the same idea but requires ≥1,000 GitHub stars (Trinity: ~460 at time of filing).

Both put Trinity one click away for anyone with a €8 Hetzner-class VPS — exactly the self-hoster the OSS front door serves. Depends on the pull-only hosted compose (#2280).

## Acceptance Criteria

- [ ] "Deploy on Hostinger" button in the README + `docs/DEPLOYMENT.md`, pointing at the raw hosted compose URL on `main`; copy states KVM2 (8 GB) or larger — KVM1 is under the floor and the button cannot pin a plan.
- [ ] Referral code created and used in the button (owner: operator account); recorded where the team can find it.
- [ ] Dokploy PR opened and merged: `blueprints/trinity/{docker-compose.yml, template.toml, meta.json, logo}`; the docker-socket bind mount follows the OpenHands blueprint precedent; template verified from a Dokploy instance preview.
- [ ] Coolify service template drafted (`templates/compose/trinity.yaml` format, `SERVICE_*` magic vars) and parked with a note to submit at ≥1,000 stars.
- [ ] Both external listings added to a recurring audit (staleness check on the compose URL / pinned tags — buttons decay: Langflow, Cal.com, Dify all had dead deploy paths within a year).

## Technical Notes

[...]

Issue #2282, opened 2026-08-18
feat: Vultr Marketplace imageless listing for Trinity (cloud-init Vendor Data)


vybe:
## Summary

List Trinity on the **Vultr Marketplace** as an **imageless** app — a clean-OS + cloud-init "Vendor Data" provisioning script (the form Vultr itself recommends) — so Vultr users get a running Trinity in one click.

## Context

Vultr's vendor program is free, explicitly takes no revenue share, and the imageless path is essentially the existing `trinity-ops-public/provision/cloud-init.sh` plus the unattended start: no snapshot to maintain, script updated in place. Coolify and OpenClaw are already listed there (Docker-socket workloads accepted). Cheapest listing in the field once the pull-only images exist (#2280).

## Acceptance Criteria

- [ ] "Become a Verified Vendor" application submitted and approved; Vendor Portal app profile created (icons, categories, support contact, gallery 800×500).
- [ ] Imageless provisioning script: installs Docker, pulls the hosted compose, reads app variables (auto-generated admin password etc.) from the Vultr metadata API, starts Trinity unattended, writes post-deploy notes.
- [ ] Private test deploy passes on a ≥8 GB plan; listing text recommends 8 GB minimum (Vultr cannot enforce plan size — say it loudly).
- [ ] Made public and approved by Vultr's review; "Deploy on Vultr" badge added to the README once live.
- [ ] Update runbook: how the script tracks Trinity releases (pinned image tag bump per release).

## Technical Notes

- Vultr docs: docs.vultr.com/vultr-marketplace (build guide + reference). Build/test on the dedicated `marketplace-2c-2gb` plan only if going the snapshot route; imageless avoids it.
- Their vendor legal page was not readable during research (403 to fetchers) — confirm obligations in a browser before submitting.

obasilakis:
## The gate is met, and most of this ticket already exists as DigitalOcean code (survey, 2026-09-11)

#2280 has landed on `dev`: `docker-compose.hosted.yml` pulls only, `start.sh --hosted --unattended` never prompts, and the five images are public on GHCR (anonymous pull verified; measurements on #2280). The DigitalOcean bundle in `packer/digitalocean/` is, in effect, a working reference implementation of this ticket's provisioning script — built for a snapshot, but the per-instance half is almost entirely provider-neutral.

### What the backend already accepts

`vultr-marketplace` is a recognised install source today — no backend work needed for provenance:

[...]

PR #2390, opened 2026-08-24
feat(hosting): publish prebuilt images and install from them (#2280)

obasilakis:
## Summary

A fresh VM was spending 5-10 minutes compiling the agent base image (Python + Node + Go + Claude Code, ~1.9 GB) before it served anything. That fails the "one click" bar for every marketplace channel, and managed hosts and template catalogues (Elestio, Dokploy, Coolify, Hostinger Docker Manager) reject a compose file that builds at all. This is the gate the whole hosting arc — #2281 DigitalOcean, #2282 Vultr, #2283 Hostinger/Dokploy, #835 Packer — sits behind.

Three pieces:

- **`.github/workflows/publish-images.yml`** pushes five images to GHCR on every `v*` tag: the four platform services **and the agent base image**. `latest` lands only on a real release tag, so a `workflow_dispatch` smoke build can never silently become what every hosted install pulls. The checkout sets `submodules: false`, so the published backend is OSS-only *by construction* rather than by a `.dockerignore` rule someone can edit.
- **`docker-compose.hosted.yml`** — `docker-compose.prod.yml` with the `build:` blocks replaced by GHCR `image:` refs and nothing else changed. Generated from prod, not hand-written.
- **`start.sh --hosted`** installs from those images. A flag, not a second script: secret generation, the `ADMIN_PASSWORD` contract #2381 just made honest, `DOCKER_GID` detection, the serving health poll and the next-steps card are all shared. A parallel installer would be the same bug one layer up.

## The agent base image is not a compose service, and never will be

The backend creates agent containers through the Docker SDK from the literal local tag `trinity-agent-base:latest` — hardcoded in `services/agent_service/lifecycle.py`, allowlisted as `trinity-agent-base:*` by SEC-172 — and compose cannot retag. So `--hosted` pulls the GHCR copy and `docker tag`s it locally before bringing the stack up.

Retagging rather than making the reference configurable is deliberate: it keeps the SEC-172 allowlist and the #1809/#1816 drift check untouched, since both read the container's own `Config.Image`, which stays `trinity-agent-base:latest` either way. The alternative needed a widened allowlist for no gain.

The consequence is worth knowing before someone writes a runbook: **a bare `docker compose -f docker-compose.hosted.yml up -d` starts a platform that cannot create a single agent**, and it fails later at agent-create time rather than at install time. That is why the upgrade instruction everywhere is "re-run `start.sh --hosted`", never "pull".

## Why the parity test exists

Two compose files describing one platform is exactly the shape of a bug this repo has now shipped five times and named: `LOG_*` (#1039), the VoIP master switch (#1056), `AGENT_AUTH_SECRET` (#1707), the container log caps (#1871), and — as recently as #2381 — `ADMIN_USERNAME`, present in the dev compose and absent from prod, so an `ADMIN_USERNAME=root` install silently kept provisioning `admin`.

`tests/unit/test_2280_hosted_compose_parity.py` fails CI when hosted and prod disagree on the service set, any third-party image pin, the top-level volumes/networks, or **any** per-service `environment` / `ports` / `volumes` / `networks` / `depends_on` / `container_name` / `restart` / `cap_*` / `security_opt` / `healthcheck` / `user` / `group_add` value.

It compares the **raw** YAML rather than `docker compose config` output on purpose: the raw form still carries the unexpanded `${VAR:-default}` strings, so a changed default (`VOIP_ENABLED:-false` → `:-true`) is a diff, where the resolved form would silently agree whenever the local `.env` happened to match.

I hit the class while building this — I diffed against a stale copy of prod and the guard immediately flagged `ADMIN_USERNAME` as missing. It works.

## Also in scope

- **Build args are per image.** Only `docker/backend/Dockerfile` declares the six provenance ARGs; the scheduler and mcp-server declare none, and an undeclared `--build-arg` warns on every build, which trains everyone to ignore the warnings that matter. The frontend takes **none** deliberately: every `VITE_*` var is statically inlined into the world-readable bundle, so a published image may carry only values correct for *every* install — possible at all only because `VITE_API_URL` is inert (`api.js` uses `baseURL: ''`, nginx proxies same-origin; #722 is an explicit warning against wiring it up).
- **The `docker/*` actions are SHA-pinned**, matching this repo's posture for every other third-party action (tailscale, pypa, appleboy, google-github-actions). One of them holds a `packages:write` token.
- **Provenance reaches the build-args step through `env:`, never `${{ }}` inside a shell body.** The commit *subject* is chosen by whoever lands the commit, so interpolating it into a `run:` script is a command-injection sink on a runner holding that token. Found in my own draft during self-review.
- **Hosted-mode side effects are suppressed rather than left to no-op silently:** an explicit `-f` disables compose's override auto-merge, so the Docker Desktop Vector fix (#1432) is skipped instead of written as a file that looks like a fix and does nothing; the #926 git-provenance export is skipped too, since there is no `build:` block to consume it.
- **A failed base-image pull is fatal**, never a silent fallback to building — that would reintroduce the 5-10 minute first boot on the install least able to absorb it.
- **TLS on a bare VM is decided and written down** (AC5): tunnel / private network / operator-run reverse proxy, with plain HTTP on a public IPv4 called out as the one combination to avoid. The managed fleet's HTTP-over-WireGuard is documented as a *finished* posture, not a compromise.
- **The 8 GB floor** is stated in the compose header, `docs/DEPLOYMENT.md` and `docs/AGENT_INSTALL_GUIDE.md`, and pinned by the parity test so it cannot quietly drop out of the compose file.

## Acceptance criteria

| AC | State |
|---|---|
| 1 — release workflow publishes all five images | ✅ (amd64; arm64 deliberately deferred, see below) |
| 2 — pull-only hosted compose, same `.env` contract | ✅ |
| 3 — `--unattended` never prompts; setup form unreachable on a public IP | ✅ already, by #2385 (ticked on the issue) |
| 4 — cloud-init user-data example | ❌ lives in `trinity-ops-public`, cross-repo |
| 5 — TLS story decided and documented | ✅ |
| 6 — 8 GB minimum stated | ✅ |
| 7 — both install docs lead with pull-only for servers | ✅ |

**#2280 stays open** for AC4 — hence `Refs`, not `Fixes`.

**arm64 is deliberately not here.** amd64 covers DigitalOcean Droplets and Vultr Cloud Compute, i.e. every marketplace lane; cross-building the ~1.9 GB agent base under QEMU is slow and flaky. Gate it on Hetzner CAX / Umbrel actually needing it.

## Test plan

- [x] `pytest tests/unit/test_2280_hosted_compose_parity.py` — 25 pass.
- [x] Negative test: added a var to prod only → `test_service_key_parity[environment]` fails; reverted → 25 pass.
- [x] `docker compose -f docker-compose.hosted.yml config` resolves clean; `diff` of resolved hosted vs prod is **exactly** the build blocks and image refs — every env key, port, volume and network identical.
- [x] Existing compose/env packaging guards: `test_1810_1814_version_truth`, `test_848_inline_auth_compose_wiring`, `test_ent237_skill_source_env_packaging`, `test_1076_voice_model_config`, `test_1987_instance_label`, `test_canary_env_prod_parity`, `test_ent14_registry_env_packaging`, `test_1489_vite_bug_build_args`, `test_1871_log_config_parity` — 148 pass.
- [x] `pytest tests/unit -k "parity or packaging or guard or wiring or compose or env"` — 1518 pass. The 5 `test_mcp_validator.py::TestCgnatSsrfGuard` failures reproduce identically on an untouched checkout (macOS IPv6/DNS environment), verified this session.
- [x] `bash -n scripts/deploy/start.sh`.
- [x] End-to-end smoke of **both** install paths with a stubbed `docker` on `PATH`. Hosted issues exactly `pull ghcr.io/abilityai/trinity-agent-base:latest` → `tag … trinity-agent-base:latest` → `compose -f docker-compose.hosted.yml pull` → `compose -f docker-compose.hosted.yml up -d`; dev issues `compose up -d` with no `-f` and no pull, and prints the unchanged dev next-steps card.
- [ ] Not verifiable pre-merge: the GHCR images do not exist until this workflow first runs on a tag. The first release cut is the real test of the publish job.

## Review notes

The one thing worth a second opinion is **HOST-007** (pull-and-retag vs. making the base-image reference configurable). I went with retagging to avoid widening the SEC-172 allowlist; if anyone would rather have `AGENT_BASE_IMAGE` as real config, that is a different PR and a security review, not a tweak to this one.

Refs #2280


dolho:
## Code review — `feature/2280-hosted-images-unattended-boot`

Reviewed the prebuilt-image publish workflow, `docker-compose.hosted.yml`, the `--hosted` path in `start.sh`, the parity guard, and the deployment docs. Nine findings.

### High

**1. `scripts/deploy/start.sh:43` — a pinned `TRINITY_IMAGE_TAG` in `.env` is silently overridden.**
`TRINITY_IMAGE_TAG="${TRINITY_IMAGE_TAG:-latest}"; export TRINITY_IMAGE_TAG` unconditionally exports a value without ever consulting `.env` — the script never `source`s it, and unlike `FRONTEND_PORT` (grepped out of `.env` at line 324) nothing reads it from there. Compose gives the shell environment precedence over `.env`, so an operator who pins `TRINITY_IMAGE_TAG=0.9.0` in `.env` — the file every other knob in this install lives in, and the only place an `--unattended` / marketplace install can persist config — has it replaced with `latest` on every `start.sh --hosted` run, for both `docker compose pull` and the agent-base `docker pull`.

That is exactly the unscheduled upgrade this PR warns against, and the summary then prints `Currently pinned to: TRINITY_IMAGE_TAG=latest` as though the operator had chosen it.

**2. `docs/DEPLOYMENT.md:48` (also `docs/AGENT_INSTALL_GUIDE.md:53`, `scripts/deploy/start.sh:496`, `docker-compose.hosted.yml:23`) — every documented pin example is a tag that is never published.**
The examples all say `TRINITY_IMAGE_TAG=v0.9.0`, but `docker/metadata-action`'s `type=semver,pattern={{version}}` (publish-images.yml:117) strips the leading `v`: pushing tag `v0.9.0` publishes `0.9.0`, `0.9`, `latest`, `sha-…` — never `v0.9.0`.

An operator following the documented server-install path gets `manifest unknown` on `docker pull ghcr.io/abilityai/trinity-agent-base:v0.9.0`, which `start.sh` treats as fatal (`exit 1`, line 311) with an error that blames "image-tag spelling" rather than the docs. Either add `type=semver,pattern=v{{version}}` to the metadata step, or fix the four examples.

### Medium

**3. `scripts/deploy/start.sh:75` — `FRONTEND_PORT` is inert in hosted mode.**
It is honoured only by `docker-compose.yml` (`"${FRONTEND_PORT:-80}:80"`); the hosted file inherits prod's hardcoded `"80:8080"`. So the port-conflict advice at line 75 ("or set `FRONTEND_PORT=` in .env for the Web UI") does nothing under `--hosted` — `up -d` still tries to bind 80 and fails. Worse, if `FRONTEND_PORT=8081` is already in `.env` from a prior dev install, lines 435-439 print `http://localhost:8081` as the access point while the UI is actually served on `:80`.

**4. `docs/DEPLOYMENT.md:87` — the documented Cloudflare Tunnel path never starts the tunnel.**
The TLS table presents Cloudflare Tunnel as "the default for a public instance" and says `cloudflared` "is already in the compose file, set `TUNNEL_TOKEN`". But `cloudflared` carries `profiles: ["tunnel"]`, and neither `start.sh --hosted` nor the documented commands pass `--profile tunnel` / `COMPOSE_PROFILES=tunnel`. An operator who sets `TUNNEL_TOKEN` and runs `./scripts/deploy/start.sh --hosted` gets no tunnel container, no error, and lands in the plain-HTTP-on-a-public-IPv4 state the same table says to avoid.

**5. `scripts/deploy/start.sh:36` — converting an existing source install to `--hosted` in place silently starts from an empty database.**
Hosted and dev share a compose project name (the directory basename) but not a `/data` source: dev mounts the named volume `trinity-data:/data`, hosted inherits prod's `${TRINITY_DATA_PATH:-./trinity-data}:/data` bind mount. Re-running `start.sh --hosted` in the same checkout brings the platform up against an empty `./trinity-data` (fresh `trinity.db`, migrations from zero) while the old database sits untouched in `<project>_trinity-data` — and Redis (`redis-data`, a shared named volume) is *not* reset, so the result is a half-migrated install. The docs frame `--hosted` as "the same script… only the image source differs", with no warning about this.

### Low

**6. `docker-compose.hosted.yml:32` — the header points twice at a script that does not exist.**
"Compose cannot retag, so `scripts/deploy/hosted-up.sh` pulls…" and "Use that script". There is no such file in this PR or the repo; the entry point is `start.sh --hosted`. This is the file managed hosts and template catalogues ingest, so it is the copy most likely to be read standalone.

**7. `docker-compose.hosted.yml:57` — not standalone, despite being written for catalogues that consume a compose file directly.**
Elestio/Dokploy/Coolify are named in the header, but the file inherits four relative host bind mounts from prod — `./config/agent-templates`, `./config/process-templates`, `./config/process-docs`, `./config/manifests` — plus `${TRINITY_DATA_PATH:-./trinity-data}`. Ingested without the repo, Docker creates those as empty directories: the local template catalog is empty and `config/manifests` is empty, so the ent#124 first-run system seed finds nothing to deploy, with no error at install time.

**8. `.github/workflows/publish-images.yml:120` — a `workflow_dispatch` build stamps provenance it was not built from.**
`type=sha,format=short` resolves from `github.sha`, and `git_branch` (line 94) from `GITHUB_REF_NAME` — neither follows the `inputs.ref` used for checkout (line 80). A dispatch with `ref: v0.9.0` from `dev` builds the tag's tree but publishes it tagged `sha-<dev-head>` and stamps `GIT_BRANCH=dev` into `GET /api/version`, on the exact path intended for out-of-band smoke builds.

**9. `tests/unit/test_2280_hosted_compose_parity.py:117` — the parity guard is narrower than it claims.**
The docstring and the PR describe it as catching drift on "any" per-service value, but `test_service_key_parity` is a fixed 18-key allowlist that omits `profiles`, `env_file`, `logging`, `labels`, `deploy` and `stop_grace_period`. `profiles` is live today (`cloudflared`), so a future prod change that profile-gates or un-gates a service, or adds an `env_file:`, drifts silently — the packaging-gap class the file exists to prevent. Consider inverting it: compare `prod_svc` minus `{build, image}` against `hosted_svc` minus `{image}` wholesale.

**10. `scripts/deploy/start.sh:360` — hosted mode disables the #1432 Docker Desktop log fix.**
`--hosted` forces `_log_src=docker`, so on Docker Desktop / any VM-backed runtime a hosted install ships the `docker_logs` source that #1432 exists to avoid, and Vector busy-loops and pegs the Docker VM. The comment is right that writing `docker-compose.override.yml` would be a no-op under an explicit `-f`, but the fix is available: append the override to `COMPOSE_FILES` when the runtime is Docker Desktop, rather than disabling the fix outright.

obasilakis:
Thanks — all ten verified against the tree before touching anything, all ten real, all ten fixed in 7c5b880. Notes where I did something other than the literal suggestion:

**2 (unpublished tag).** Took the first option — added `type=semver,pattern=v{{version}}` rather than editing the four examples. Both `v0.9.0` and `0.9.0` now resolve to one digest, which costs a tag ref and makes the release name and the image tag the same string. Fixing only the docs would have left anyone typing the release name from the releases page hitting `manifest unknown`.

**3 (`FRONTEND_PORT`).** Fixed in `docker-compose.prod.yml` too, not just hosted. It is the same inert knob there — prod inherited the hardcoded `"80:8080"` — and changing only hosted would have put the two files out of parity, which the guard in this PR exists to prevent. Default is unchanged at 80, so no existing deployment moves.

**4 (tunnel).** Made it work rather than documenting the flag. HOST-010 names the tunnel as *the default posture for a public instance*, and a default that silently no-ops is the finding — `--hosted` now appends `--profile tunnel` when `TUNNEL_TOKEN` is non-empty. Every other invocation still needs it by hand, which `.env.example` and DEPLOYMENT.md now say, along with `docker ps | grep cloudflared` to confirm.

**5 (dev → hosted).** Refuses and exits rather than warning. The failure is silent, and by the time anyone notices, the empty DB may already have been written to. It prints the volume-copy command; `TRINITY_DATA_PATH` is the escape for a deliberate fresh start.

**9 (parity allowlist).** Took your inversion — `prod` minus `{build, image}` vs `hosted` minus `{image}`, wholesale. You were right that an allowlist only watches the keys someone remembered. It then earned itself twice inside this change: it caught a `profiles` key I planted to test it, and it caught me reverting the prod half of finding 3 with a stray `git checkout --` during that same test.

**10 (Vector).** Took your suggestion exactly — the override is appended to `COMPOSE_FILES` by name instead of the fix being switched off. The `vector` service is byte-identical between `docker-compose.yml` and the hosted file, so it merges onto either.

**7 (not standalone)** is the one I have documented rather than solved. The file still carries six relative bind mounts, so it needs the repo checked out beside it; packaging the config into the images is a real change and a separate PR. The header now says so plainly instead of implying a catalogue can ingest it bare.

Behaviour verified with a stubbed `docker` on `PATH`, since the GHCR images do not exist until this workflow first runs on a tag: `.env` pin honoured (`v0.9.0`) and an explicit shell value still overriding it (`0.8.1`); `--profile tunnel` present only with a token set and absent when blank; the override file appended only on a VM-backed runtime; the dev→hosted guard exiting 1; and the dev path unchanged — bare `compose up -d`, no `-f`, no pull, no profile.

`§8.9` in `docs/memory/requirements/infrastructure.md` absorbed the findings as HOST-013 through HOST-016 plus edits to HOST-001/004/005/008, so the reasoning is in the requirement rather than only in this thread.

dolho:
## Code review — re-review at `7c5b880`

The ten findings from the previous round are addressed. This pass concentrated on the new code that addressed them — particularly the dev→hosted data guard, which is the piece doing the most work — and on the paths an unattended install actually takes.

Verified clean rather than assumed: the hosted↔prod compose diff really is only the `build:`→`image:` swap (direct file diff), the vector service is byte-identical between dev and hosted so the #1432 override merges cleanly under the explicit `-f`, the parity test's wholesale comparison is sound, and `build_args: version` for the base image matches `build-base-image.sh`.

Nine findings.

### High

**1. `start.sh:348` — the dev→hosted data guard fails open on any checkout directory containing `-` or `_`.**

```sh
_project=$(basename "$(pwd)" | tr '[:upper:]' '[:lower:]' | tr -cd '[:alnum:]')
```

That strips hyphens and underscores; Compose's own project-name normalisation **keeps** them (verified empirically: dir `proj_Trin-ity` → project `proj_trin-ity`). So in a checkout named `project_trinity` — the layout CLAUDE.md itself documents — or `trinity-dev`, or a worktree like `trinity-140`, `docker volume inspect "${_project}_trinity-data"` looks up `projecttrinity_trinity-data`, misses, and the guard silently passes. That produces exactly the empty-DB-plus-shared-Redis half-migration it was written to prevent, on the directory names most likely to be in use.

`COMPOSE_PROJECT_NAME` is also ignored, and this is now a *third* project-name derivation in the file, disagreeing both with Compose and with the pre-existing `volume_exists()` (`basename "$PWD"`, no lowercase). Use `tr -cd '[:alnum:]_-'` plus `${COMPOSE_PROJECT_NAME:-…}`, or read the real name from `docker compose config --format json`.

### Medium

**2. `docs/AGENT_INSTALL_GUIDE.md:51-54` — the documented hosted install creates a one-line `.env` and thereby suppresses the `.env.example` copy.**

The snippet is `git clone … && cd trinity && echo 'TRINITY_IMAGE_TAG=v0.9.0' >> .env && ./scripts/deploy/start.sh --hosted --unattended`. `start.sh` copies the template only under `if [ ! -f .env ]`, which is now false — so the install proceeds with a `.env` holding a single key: no `ANTHROPIC_API_KEY`, no `FRONTEND_URL`, no `EMAIL_*`, no `TRINITY_DATA_PATH`. `docs/DEPLOYMENT.md` gets this right (`cp .env.example .env` first); this guide, which an installing agent follows verbatim, omits it.

**3. `start.sh:306` — `resolve_image_tag`'s naive `.env` parse is exported, so it overrides Compose's own correct parse.**

`cut -d'=' -f2- | tr -d '[:space:]'` keeps surrounding quotes and swallows inline comments. `TRINITY_IMAGE_TAG="v0.9.0"` yields `"v0.9.0"`; `TRINITY_IMAGE_TAG=v0.9.0  # pinned` yields `v0.9.0#pinned`. Compose would have parsed both correctly — but because the value is `export`ed, the shell wins over `.env` for every service image ref too. The run then dies at the agent-base pull with the tailored message blaming "image-tag spelling", pointing the operator at a `.env` line that is in fact valid. Strip surrounding quotes and a trailing ` #…` before exporting.

**4. `start.sh:340-372` — the reverse switch (hosted → dev) is unguarded, and is the likelier mistake.**

The new refusal covers "dev volume exists, hosted bind DB missing". An operator who installed with `--hosted` and later runs `./scripts/deploy/start.sh` *without* the flag — forgetting it, or following older docs — gets `docker-compose.yml`, an empty `trinity-data` named volume and the same shared `redis-data`: identical half-migrated state, with no warning at all. A symmetric check (`${_data_path}/trinity.db` exists && `HOSTED != 1` && no dev volume) closes it.

### Low

**5. `start.sh:363` — the guard's suggested copy command is wrong for an absolute `TRINITY_DATA_PATH`.** `"$(pwd)/${_data_path#./}"` strips only a leading `./`, so `TRINITY_DATA_PATH=/srv/trinity-data` emits `-v "/path/to/repo//srv/trinity-data":/to` and silently copies the database into the wrong place. (That value was also run through `tr -d '[:space:]'` on read, so a path containing spaces is already corrupted before it reaches here.)

**6. `start.sh:575` — the printed hosted commands drop `--profile tunnel`, so they cannot manage the container the script just started.** With `TUNNEL_TOKEN` set the script activates the profile (:327), but the summary prints `docker compose -f docker-compose.hosted.yml stop` / `logs -f`. Compose acts only on services in the active profile set, so `stop` leaves `trinity-cloudflared` running and the tunnel live after the operator believes the stack is down. Persisting `COMPOSE_PROFILES=tunnel` in `.env` — which every later bare compose command inherits — makes the printed commands correct.

**7. `start.sh:470` — a failed platform-image pull dies bare on `set -e`, unlike the base-image pull one screen above.** The agent-base pull has a tailored fatal message naming the tag and the `--hosted` fallback; `docker compose … pull` immediately after has none. The likeliest failure — a tag not yet published, or GHCR denying an anonymous pull — surfaces as a raw Compose error on precisely the install least equipped to interpret it.

**8. `.github/workflows/publish-images.yml` — nothing makes the published GHCR packages anonymously pullable, and the failure lands on the operator.** Container packages first created by `GITHUB_TOKEN` are not reliably public; if any of the five lands private, `docker pull ghcr.io/abilityai/trinity-agent-base:latest` on a fresh VM fails with `denied`, and `start.sh` reports it as a network or tag-spelling problem. This is the one path the PR itself notes is untestable pre-merge, so a release-checklist line ("set package visibility to public on first publish") or an explicit post-push verification step is worth having.

**9. `start.sh:69` — the port pre-flight still hardcodes `80` after `FRONTEND_PORT` was made live.** Both prod and hosted now honour `${FRONTEND_PORT:-80}`, but the loop still probes `80 8000 8080 6379`. An operator who took the message's own advice and set `FRONTEND_PORT=8081` keeps being warned about 80 and is never warned about a conflict on 8081.

scripts/deploy/start.sh
```diff
@@ -263,8 +288,112 @@ ensure_docker_gid() {
...
+# volume both files share) is NOT reset, so the result is a half-migrated
+# install with live session/lock state pointing at rows that no longer exist.
+#
+# Nothing about "the same script, only the image source differs" prepares an
+# operator for that, so refuse rather than warn: the failure is silent, and by
+# the time it is noticed the fresh DB may have been written to.
+if [ "$HOSTED" = "1" ]; then
+    _data_path="${TRINITY_DATA_PATH:-$(grep -E '^TRINITY_DATA_PATH=' .env 2>/dev/null | tail -1 | cut -d'=' -f2- | tr -d '[:space:]')}"
+    _data_path="${_data_path:-./trinity-data}"
+    _project=$(basename "$(pwd)" | tr '[:upper:]' '[:lower:]' | tr -cd '[:alnum:]')
```
dolho: **HIGH — the dev→hosted data guard fails open on any checkout directory containing `-` or `_`.**

```sh
_project=$(basename "$(pwd)" | tr '[:upper:]' '[:lower:]' | tr -cd '[:alnum:]')
```

That strips hyphens and underscores; Compose's own project-name normalisation **keeps** them (verified empirically: dir `proj_Trin-ity` → project `proj_trin-ity`).

So in a checkout named `project_trinity` — the layout CLAUDE.md itself documents — or `trinity-dev`, or a worktree like `trinity-140`, `docker volume inspect "${_project}_trinity-data"` looks up `projecttrinity_trinity-data`, misses, and the guard silently passes. That produces exactly the empty-DB-plus-shared-Redis half-migration it was written to prevent, on the directory names most likely to be in use.

`COMPOSE_PROJECT_NAME` is also ignored, and this is now a *third* project-name derivation in the file, disagreeing both with Compose and with the pre-existing `volume_exists()` (`basename "$PWD"`, no lowercase).

Fix: `tr -cd '[:alnum:]_-'` plus `${COMPOSE_PROJECT_NAME:-...}`, or read the real name from `docker compose config --format json`.

docs/AGENT_INSTALL_GUIDE.md
```diff
@@ -45,7 +45,31 @@ docker compose version >/dev/null 2>&1 && echo "compose: ok" || echo "compose: M
 
 ## Step 1 — Install (one shot, unattended)
 
-**Fresh machine (no clone yet):**
+**On a server — pull prebuilt images (`--hosted`). This is the default you want (#2280):**
+
+```bash
+git clone https://github.com/abilityai/trinity.git
```
dolho, line 51: **MEDIUM — the documented hosted install creates a one-line `.env` and thereby suppresses the `.env.example` copy.**

The snippet is `git clone ... && cd trinity && echo 'TRINITY_IMAGE_TAG=v0.9.0' >> .env && ./scripts/deploy/start.sh --hosted --unattended`.

`start.sh` copies the template only under `if [ ! -f .env ]`, which is now false — so the install proceeds with a `.env` holding a single key: no `ANTHROPIC_API_KEY`, no `FRONTEND_URL`, no `EMAIL_*`, no `TRINITY_DATA_PATH`.

`docs/DEPLOYMENT.md` gets this right (`cp .env.example .env` first); this guide, which an installing agent follows verbatim, omits it.

scripts/deploy/start.sh
```diff
@@ -263,8 +288,112 @@ ensure_docker_gid() {
...
+# precedence over `.env`. The summary then printed `Currently pinned to: latest`
+# as though they had chosen it, which is exactly the unscheduled upgrade the
+# rest of this script warns about.
+#
+# Mirrors the FRONTEND_PORT read further down; `cut -d'=' -f2-` keeps a value
+# containing '=' intact, and `tail -1` matches shell/compose last-wins semantics
+# for a duplicated key.
+resolve_image_tag() {
+    if [ -z "${TRINITY_IMAGE_TAG:-}" ]; then
+        TRINITY_IMAGE_TAG=$(grep -E '^TRINITY_IMAGE_TAG=' .env 2>/dev/null | tail -1 | cut -d'=' -f2- | tr -d '[:space:]')
```
dolho: **MEDIUM — `resolve_image_tag`'s naive `.env` parse is exported, so it overrides Compose's own correct parse.**

`cut -d'=' -f2- | tr -d '[:space:]'` keeps surrounding quotes and swallows inline comments. `TRINITY_IMAGE_TAG="v0.9.0"` yields `"v0.9.0"`; `TRINITY_IMAGE_TAG=v0.9.0  # pinned` yields `v0.9.0#pinned`.

Compose would have parsed both correctly — but because the value is `export`ed, the shell wins over `.env` for every service image ref too. The run then dies at the agent-base pull with the tailored message blaming "image-tag spelling", pointing the operator at a `.env` line that is in fact valid.

Strip surrounding quotes and a trailing ` #...` before exporting.

scripts/deploy/start.sh
```diff
@@ -263,8 +288,112 @@ ensure_docker_gid() {
...
+
+# --- Refuse a silent dev -> hosted data switch (#2280) ------------------------
+# Dev and hosted share a compose project name (the directory basename) but NOT a
+# /data source: docker-compose.yml mounts the named volume `trinity-data`, while
+# hosted (inheriting prod) binds ${TRINITY_DATA_PATH:-./trinity-data}. So
+# re-running this script with --hosted in a checkout that has been running the
+# dev stack comes up against an EMPTY database and migrates from zero, while the
+# real one sits untouched in the named volume — and Redis (`redis-data`, a named
+# volume both files share) is NOT reset, so the result is a half-migrated
+# install with live session/lock state pointing at rows that no longer exist.
```
dolho: **MEDIUM — the reverse switch (hosted → dev) is unguarded, and is the likelier mistake.**

The new refusal covers "dev volume exists, hosted bind DB missing". An operator who installed with `--hosted` and later runs `./scripts/deploy/start.sh` *without* the flag — forgetting it, or following older docs — gets `docker-compose.yml`, an empty `trinity-data` named volume and the same shared `redis-data`: identical half-migrated state, with no warning at all.

A symmetric check (`${_data_path}/trinity.db` exists && `HOSTED != 1` && no dev volume) closes it.

scripts/deploy/start.sh
```diff
@@ -263,8 +288,112 @@ ensure_docker_gid() {
...
+        echo "   Wanted: ${_data_path}/trinity.db (the bind mount docker-compose.hosted.yml uses)" >&2
+        echo "" >&2
+        echo "   Starting anyway would migrate a NEW empty database from zero while the real" >&2
+        echo "   one stays in the volume — and Redis is shared between the two stacks, so you" >&2
+        echo "   would get a half-migrated install rather than a clean one." >&2
+        echo "" >&2
+        echo "   Copy the data across first (with the stack stopped):" >&2
+        echo "       docker compose stop" >&2
+        echo "       mkdir -p ${_data_path}" >&2
+        echo "       docker run --rm -v ${_project}_trinity-data:/from -v \"\$(pwd)/${_data_path#./}\":/to \\" >&2
```
dolho: **LOW — the guard's suggested copy command is wrong for an absolute `TRINITY_DATA_PATH`.**

`"$(pwd)/${_data_path#./}"` strips only a leading `./`, so `TRINITY_DATA_PATH=/srv/trinity-data` emits `-v "/path/to/repo//srv/trinity-data":/to` and silently copies the database into the wrong place.

(That value was also run through `tr -d '[:space:]'` on read, so a path containing spaces is already corrupted before it reaches here.)

scripts/deploy/start.sh
```diff
@@ -407,13 +565,39 @@ fi
...
-echo "(See docs/DEPLOYMENT.md → Troubleshooting → Stale platform images.)"
-echo ""
+# #2280: hosted installs have no build context, so the stale-image remedy is a
+# pull, not a build — and every compose command needs the explicit -f, since
+# hosted deliberately opts out of the default file merge. Printing the dev
+# commands to a hosted operator sends them at a `build` that cannot work.
+if [ "$HOSTED" = "1" ]; then
+    _cf="-f docker-compose.hosted.yml"
+    echo "To view logs:      docker compose ${_cf} logs -f"
+    echo "To stop services:  docker compose ${_cf} stop"
```
dolho, line 734: **LOW — the printed hosted commands drop `--profile tunnel`, so they cannot manage the container the script just started.**

With `TUNNEL_TOKEN` set the script activates the profile (:327), but the summary prints `docker compose -f docker-compose.hosted.yml stop` / `logs -f`. Compose acts only on services in the active profile set, so `stop` leaves `trinity-cloudflared` running and the tunnel live after the operator believes the stack is down.

Persisting `COMPOSE_PROFILES=tunnel` in `.env` — which every later bare compose command inherits — makes the printed commands correct.

scripts/deploy/start.sh
```diff
@@ -314,11 +456,23 @@ if [ "$_log_src" = "file" ]; then
...
+    # after the base file it patches.
+    if [ "$HOSTED" = "1" ]; then
+        COMPOSE_FILES+=(-f docker-compose.override.yml)
+    fi
 fi
 # -----------------------------------------------------------------------------
 
+if [ "$HOSTED" = "1" ]; then
+    echo "Pulling platform images (tag: ${TRINITY_IMAGE_TAG})..."
+    docker compose "${COMPOSE_FILES[@]}" pull
```
dolho: **LOW — a failed platform-image pull dies bare on `set -e`, unlike the base-image pull one screen above.**

The agent-base pull has a tailored fatal message naming the tag and the `--hosted` fallback; `docker compose ... pull` immediately after has none. The likeliest failure — a tag not yet published, or GHCR denying an anonymous pull — surfaces as a raw Compose error on precisely the install least equipped to interpret it.

.github/workflows/publish-images.yml
```diff
@@ -0,0 +1,206 @@
+name: Publish container images
```
dolho, line 1: **LOW — nothing makes the published GHCR packages anonymously pullable, and the failure lands on the operator.**

Container packages first created by `GITHUB_TOKEN` are not reliably public. If any of the five lands private, `docker pull ghcr.io/abilityai/trinity-agent-base:latest` on a fresh VM fails with `denied`, and `start.sh` reports it as a network or tag-spelling problem.

This is the one path the PR itself notes is untestable pre-merge, so a release-checklist line ("set package visibility to public on first publish") or an explicit post-push verification step is worth having.

scripts/deploy/start.sh
```diff
@@ -407,13 +565,39 @@ fi
...
+    echo "To view logs:      docker compose logs -f"
+    echo "To stop services:  docker compose stop"
+    echo "  (use 'stop', NOT 'down' — 'down' destroys agent containers)"
+    echo ""
+    echo "Just pulled new code? If services fail with ModuleNotFoundError or"
+    echo "the UI shows 'Disconnected', the platform images may be stale —"
+    echo "rebuild with:  docker compose build && docker compose up -d"
+    echo "(See docs/DEPLOYMENT.md → Troubleshooting → Stale platform images.)"
+    echo ""
+fi
```
dolho, line 762: _(anchored to the nearest line in this PR's diff; the finding is at `scripts/deploy/start.sh:69`)_

**LOW — the port pre-flight still hardcodes `80` after `FRONTEND_PORT` was made live.**

Both prod and hosted now honour `${FRONTEND_PORT:-80}`, but the loop still probes `80 8000 8080 6379`. An operator who took the message's own advice and set `FRONTEND_PORT=8081` keeps being warned about 80 and is never warned about a conflict on 8081.

scripts/deploy/start.sh
obasilakis: Confirmed and fixed — you were right about both halves, and the empirical check was decisive.

`docker compose config` on this repo's Compose version:

```
dir project_trinity  -> name: project_trinity
dir trinity-dev      -> name: trinity-dev
dir proj_Trin-ity!x  -> name: proj_trin-ityx
dir --_Foo.Bar       -> name: foobar
```

So the rule is: lowercase, keep `[a-z0-9_-]`, drop everything else, trim leading `_`/`-`. `tr -cd '[:alnum:]'` disagreed on exactly the two characters that appear in every real checkout name.

There is now one `compose_project_name()` at the top of the file, and `volume_exists()` uses it too — so the fourth spelling (`basename "$PWD"`, unlowercased) is gone with it. `COMPOSE_PROJECT_NAME` is honoured with Compose's own precedence (shell over `.env`), also verified rather than assumed.

I did not read the name back out of `docker compose config --format json`: it would need a JSON parse with no guaranteed `jq` on the host, and the guard runs before anything else has proven the compose file resolves. A helper whose output is pinned equal to Compose's gets the same correctness without that dependency.

`tests/unit/test_2390_start_sh_env_and_project_name.py` executes the helper under bash for every case above, and end-to-end smoke confirms the guard now refuses in both `trinity-dev` and `project_trinity` (it passed silently in both before).

docs/AGENT_INSTALL_GUIDE.md
obasilakis: Fixed. Added `cp .env.example .env` as the first line, plus a note saying why the order matters, since the failure is invisible at install time:

> **Copy the template before you append to it.** `start.sh` seeds `.env` from `.env.example` only `if [ ! -f .env ]`. Writing the pin first *creates* the file, so the seed is skipped and the install proceeds on a `.env` holding one key [...] The stack still boots, which is what makes it worth calling out: you find out later, one missing knob at a time.

Good catch that this is the guide an agent follows verbatim — `DEPLOYMENT.md` had it right and this one is the copy that actually gets executed.

scripts/deploy/start.sh
obasilakis: Fixed, and generalised — this was four call sites, not one (`TRINITY_IMAGE_TAG`, `TUNNEL_TOKEN`, and `TRINITY_DATA_PATH` in two places), each with its own spelling.

There is now a single `env_value()` that mirrors compose-go's dotenv rules, with the expected values taken from `docker compose config` rather than from reading the source:

```
X="v0.9.0"        -> v0.9.0        (quotes stripped)
X=v0.9.0  # pin   -> v0.9.0        (inline comment stripped)
X=/srv/my dir     -> /srv/my dir   (internal spaces preserved)
X='has#hash'      -> has#hash      ('#' inside quotes is literal)
```

Your point about the export is the reason this mattered more than it looked, and it is now written into the comment: the value wins over `.env` for every service ref, so a mis-parse here *overrode* a valid operator line and then blamed them for it.

Smoke-verified: `TRINITY_IMAGE_TAG="v0.9.0"  # pinned` now pulls `ghcr.io/abilityai/trinity-agent-base:v0.9.0`.

scripts/deploy/start.sh
obasilakis: Agreed, and fixed — and I think you are right that it is the likelier of the two, because it needs no new flag, just a forgotten one.

Rather than bolt on a second independent check, the state is read once and both refusals branch off it:

```sh
_dev_volume=0; docker volume inspect "${_project}_trinity-data" >/dev/null 2>&1 && _dev_volume=1
_hosted_db=0;  [ -f "${_data_path}/trinity.db" ] && _hosted_db=1
```

so the pair cannot drift into disagreeing about what it found. The reverse message points at the `--hosted` flag first (the usual cause) and gives the copy-into-the-volume path second.

Smoke-verified: a checkout with `trinity-data/trinity.db` and no dev volume now refuses a bare `./scripts/deploy/start.sh` and prints both routes.

scripts/deploy/start.sh
obasilakis: Both fixed.

The path is absolutised by case rather than by prefix-stripping:

```sh
case "$_data_path" in
    /*) _data_abs="$_data_path" ;;
    *)  _data_abs="$(pwd)/${_data_path#./}" ;;
esac
```

and `_data_abs` is what the `mkdir -p` / `docker run -v` lines print, quoted.

The parenthetical was the more interesting half: `tr -d '[:space:]'` is gone from every `.env` read, so a path containing spaces now survives — that is pinned by a test case (`X=/srv/my data/dir -> /srv/my data/dir`), matching what `docker compose config` does with the same line.

Smoke-verified with an absolute `TRINITY_DATA_PATH`: the printed command carries the path as-is, with no `$(pwd)` prefix.

scripts/deploy/start.sh
obasilakis: Fixed, and I took the `COMPOSE_PROFILES` route you suggested — verified first that Compose reads it from `.env`, and that a duplicated key there resolves last-wins.

`persist_compose_profile tunnel` appends rather than rewriting in place. Appending is enough given last-wins, and it keeps an operator-authored value out of a `sed` replacement, where an `&` or the delimiter would be substitution syntax rather than text. It is additive (`COMPOSE_PROFILES=obs` becomes `obs,tunnel`) and a re-run is a no-op, both smoke-verified.

Your framing of the consequence is now the comment on it: `stop` left the tunnel container running and the instance publicly reachable *after the operator had been told the stack was down*.

scripts/deploy/start.sh
obasilakis: Fixed — it now has a tailored fatal matching the agent-base one, and it names the two causes you identified:

```
ERROR: could not pull the platform images at tag 'latest'.
       Most likely one of:
         • the tag does not exist — check TRINITY_IMAGE_TAG against
           https://github.com/abilityai/trinity/releases (both 'v0.9.0'
           and '0.9.0' are published for the same digest);
         • 'denied' / 'unauthorized' — the GHCR package is not public;
           report it, it is a publishing fault, not yours;
         • no network route to ghcr.io.
```

The middle line is deliberate after your next comment: an operator hitting a private package should be told it is not theirs to fix, rather than be sent looking for a login.

.github/workflows/publish-images.yml
obasilakis: Good catch, and this is the one I could not test any other way — so I made CI test it instead of writing it on a checklist.

The workflow now has a post-push `Verify anonymous pull` step: it fetches the manifest with a scratch `DOCKER_CONFIG`, so the runner's own `ghcr.io` login cannot mask a private package, and it retries a few times because visibility/propagation is not instant on a first publish. On failure it emits a GitHub error annotation carrying the fix:

> GitHub → Packages → <image> → Package settings → Change visibility → Public. This is normally needed only on the first publish of a package.

It runs after the push and fails the job on purpose: the image is already published at that point, so a red step is the signal to flip visibility once, not a lost build.

Paired with a line in `DEPLOYMENT.md` telling an operator that `denied` (as opposed to `manifest unknown`) is a publishing fault and points at that step — so nobody works around it with a login.

I preferred this to querying the packages API for `visibility`: it exercises the operator's literal path, and needs no permission `GITHUB_TOKEN` may not have on an org-owned package.

scripts/deploy/start.sh
obasilakis: Fixed. The loop now probes the configured port:

```sh
_preflight_frontend_port="${FRONTEND_PORT:-$(env_value FRONTEND_PORT)}"
_preflight_frontend_port="${_preflight_frontend_port:-80}"
for _pp in "$_preflight_frontend_port" 8000 8080 6379; do ...
```

On a first install there is no `.env` to read, so it falls back to 80 — which is what will be bound.

Smoke-verified on a host where 80 *is* listening: with `FRONTEND_PORT=9481` and 9481 bound, the message reads `Ports already in use: 9481 8000 8080 6379` and says nothing about 80. Previously it was the exact inverse.

dolho:
## /review — re-review after `7c5b8809` + `cc6c9769`

**Branch**: `feature/2280-hosted-images-unattended-boot` → `dev` @ `cc6c9769` · **+2107/−22** · CI **21 pass / 4 skipping / 0 fail**

**All nine findings are fixed.** I verified each against the current code rather than against the replies, and two of them are better fixed than I asked for.

---

### The one worth showing evidence for

The project-name fix isn't just "a more careful regex" — it closes a guard that **failed open on the most likely directory names in use**. Your `compose_project_name` reproduces compose-go's `NormalizeProjectName` exactly (lowercase → keep `[a-z0-9_-]` → TrimLeft `_-`). Run against the old derivation:

```
  project_trinity    -> project_trinity     (new, matches the real volume)
  trinity-dev        -> trinity-dev
  --_Foo.Bar         -> foobar
  proj_Trin-ity!x    -> proj_trin-ityx

  project_trinity    -> projecttrinity      (old — matched no volume, guard fell open)
  trinity-dev        -> trinitydev          (old — same)
```

`project_trinity` is the layout **CLAUDE.md itself documents**, and `trinity-dev` is the submodule checkout name. So the pre-fix guard was inert on precisely the installs it was written for. Consolidating three disagreeing derivations into one was the right call.

### The rest, each checked

- **`env_value()`** — mirrors compose-go's dotenv rules where it matters: quotes stripped, inline comment stripped **only when preceded by whitespace** (so `TAG=v1#c` stays literal, which is correct and easy to get wrong in the other direction), internal spaces preserved, `export ` prefix handled, `${line#*=}` keeps `=` inside values, last-wins via `tail -1`, and a commented-out key can't match. Generalising from one call site to four was the right read — the export-beats-compose interaction made every one of them a silent override.
- **Bidirectional data guard** — state read once into `_dev_volume`/`_hosted_db`, both refusals branch off it. The absolute path is resolved by `case` rather than prefix-stripping, so `TRINITY_DATA_PATH=/srv/...` no longer prints a `docker run -v` that would copy the DB somewhere nothing reads. I checked the cwd assumption the whole guard rests on: `cd "$(dirname "$0")/../.."` at line 5, pre-existing — so the relative `_data_path` probe is anchored.
- **`persist_compose_profile`** — idempotent (`,cur,` containment), additive, append-not-`sed` for the reason you give. Verified the reasoning holds: compose does resolve a duplicated `.env` key last-wins.
- **Both pull failures** now carry tailored fatals naming the two real causes; **port pre-flight** probes the configured port.
- **`Verify anonymous pull`** — the scratch `DOCKER_CONFIG` is the detail that makes it a real test rather than a green rubber stamp, since the runner is logged in to the very registry being probed. Making CI prove the operator's exact path was the right answer to a finding neither of us could otherwise test.

### The guard I'd call out as the best thing in the PR

`test_service_parity_wholesale` inverts the comparison — prod minus `{build,image}` **must equal** hosted minus `{image}` — so a compose key nobody thought of is covered the day it appears. Your note that the 18-key allowlist had already missed `profiles`, which is live on `cloudflared` today, is the whole argument for the shape. This is the correct answer to the class the PR body itself cites (#1039/#1056/#1707/#1871), and it's rare to see it written as an inversion rather than a longer list.

`33 passed` across both new test files in a clean worktree.

---

### Plan completion

| AC | | |
|---|---|---|
| 1 — publish five images | ✅ | amd64; arm64 deferred with a stated reason |
| 2 — pull-only hosted compose, same `.env` contract | ✅ | and guarded wholesale |
| 3 — unattended never prompts / form unreachable | ✅ | via #2381 |
| 4 — cloud-init user-data example | ❌ | cross-repo |
| 5 — TLS decided + documented | ✅ | `DEPLOYMENT.md:102` — a real decision table, not a shrug |
| 6 — 8 GB floor stated | ✅ | pinned by a test |
| 7 — both guides reference the pull-only path | ✅ | |

**AC4's handling is exemplary and I want to say so plainly**: reported ❌ in the PR body, `Refs` not `Fixes` so #2280 stays open, and recorded in `infrastructure.md` as 🟡 Partial with the outstanding item named. That is the honest classification, and the requirements file carrying it is what stops "partial" decaying into "done" three weeks from now.

---

### New findings — 2, both minor, neither blocking

**[I1] The verify step interpolates `${{ matrix.image }}` inside a `run:` body** (Confidence: 7/10)

Two steps above, the PR states the rule and explains it well:

> Every value arrives through `env:`, never through `${{ }}` inside the shell body … a command-injection sink.

Then the `::error title=` line in **Verify anonymous pull** does exactly that. It is **not exploitable** — `matrix.image` is one of five literals in this file, not attacker-controlled — so this is consistency, not a vulnerability. But the value of a rule like that is being unconditional: the next person copying this step as a template won't know which interpolations were audited. `IMAGE` is already in `env:`; the message can use `${IMAGE##*/}`.

**[I2] A fifth `>> .env` appender, and none of them guard the trailing newline** (Confidence: 6/10 — pre-existing)

`persist_compose_profile` joins four appenders that already exist on `dev`. If `.env`'s last line lacks a trailing newline, `echo "K=V" >> .env` silently concatenates onto it. `.env.example` ends with `\n` and the documented install uses `echo … >> .env` (which adds one), so the shipped paths are safe — this only bites a hand-authored file. **Not this PR's to fix**, and I'd rather it not grow scope; noting it because the fix is one shared helper and this PR is the natural place someone would later put it.

---

### Summary

- **Prior findings**: 9/9 fixed, all verified — one demonstrated empirically
- **New**: 2 informational, 0 blocking
- **Plan**: 6/7 ACs met, the 7th honestly deferred cross-repo with the issue kept open
- **Tests**: 33 passed; CI 21 pass / 0 fail

Nothing here blocks. The two things I'd carry into other work from this PR are the inverted parity guard and making CI prove the anonymous-pull path — both are the shape where "someone will remember to check" gets replaced by something that fails loudly.

dolho (approved):

AndriiPasternak31:

**/review: needs-changes (one confirmed-by-execution defect) · /validate-pr: request-changes (docs).** Code, security posture and tests are above bar — 33/33 tests re-run green locally, resolved-config parity independently re-proven via `docker compose config` (the diff is exactly the four `build:` blocks + image refs), workflow secrets/injection hygiene and action pinning clean. These are new findings, not repeats of the absorbed review round:

**C1 — the anonymous-pull verify step fails on every publish, before any network call.** `.github/workflows/publish-images.yml:221`: `IMAGE: ghcr.io/${{ github.repository_owner }}/${{ matrix.image }}` interpolates `Abilityai` (uppercase — verified via `gh repo view`), and docker rejects uppercase repo names outright (reproduced locally: `docker manifest inspect ghcr.io/Abilityai/trinity-backend:latest` → "repository name … must be lowercase", instant). On the first release cut all five matrix jobs burn 5×15s retries then exit 1 with `::error title=GHCR package is not public` — even when the packages *are* public — destroying exactly the signal the step exists for ("red = flip visibility once") and, per the workflow's own comment, training everyone to ignore the warnings that matter. The push itself is unaffected (`docker/metadata-action` documents lowercasing its `images:` input) and `start.sh` hardcodes lowercase. Fix: lowercase in the shell body (`tr '[:upper:]' '[:lower:]'`) or hardcode `ghcr.io/abilityai/` as start.sh already does.

**I1 — a `workflow_dispatch` from a tag ref re-publishes that tag's semver tags AND repoints `latest` backwards.** `publish-images.yml:146`'s `enable=${{ startsWith(github.ref, 'refs/tags/v') }}` distinguishes tag vs branch refs, not push vs dispatch — dispatching on an old tag (the natural "smoke-test the workflow" choice) rebuilds at a new digest (the agent base bakes "Claude Code (latest)", so materially different content) and moves `latest` backwards, silently downgrading every unpinned hosted install on its next `start.sh --hosted`. Note metadata-action's default `flavor: latest=auto` *also* applies latest on any semver tag ref, so tightening only the `enable=` is insufficient: add `flavor: latest=false` and gate latest on `github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')`.

**I2 — `stop.sh` is not hosted-aware, and runs the command start.sh's own summary forbids.** `scripts/deploy/stop.sh:12` bare `docker compose down` → on a hosted install it loads the dev file (same directory ⇒ same project), tears down and *removes* the platform containers, and — dev compose defines no `cloudflared` service — leaves `trinity-cloudflared` publicly serving after "✅ All services stopped". This is the hazard `persist_compose_profile` was added to close, re-entering through the one entry point the PR didn't touch. Fine as a fast-follow: hosted-aware file selection + `stop` not `down` (or an explicit warning in the hosted summary).

**Docs (the validate-pr request-changes):** this is a new capability with no feature flow — suggest `docs/memory/feature-flows/hosted-install.md` (publish workflow → GHCR → `start.sh --hosted` pull/retag → data-switch guards → upgrade path, with a Testing section pointing at the two test files; precedent: `database-backup.md` for #2216) plus the `docs/memory/feature-flows.md` index row. And `docs/memory/architecture.md` now carries stale "both compose files" claims at `:407` (#2216 DB Backup), `:2731` (#1159), and the #1871 x-logging block — there are three compose files now. Trivial: requirements §8.9's `Tests:` row omits `tests/unit/test_2390_start_sh_env_and_project_name.py`.

Process note: `Refs #2280` (not `Fixes`) is correct — AC4 cloud-init lives in trinity-ops-public and the issue must stay open — but whoever tracks the epic should know this PR doesn't move the status label.

obasilakis:
Agreed on all four — thanks, C1 in particular would have burned on the first release cut and told us the packages were private when they were not. Reproduced it locally before fixing: `docker manifest inspect ghcr.io/Abilityai/trinity-backend:latest` → `invalid reference format: repository name must be lowercase`, instantly, no network.

Fixed in 1f3ac10:

**C1** — the verify step now builds its reference in the shell body and lowercases it (`tr '[:upper:]' '[:lower:]'`). Kept it derived rather than hardcoding `ghcr.io/abilityai/` so it still follows a fork's owner, but the rule is now enforced instead of assumed.

**I1** — took both halves of your note. `flavor: latest=false` (without it the `latest=auto` default applies `latest` on any semver tag ref, so gating the explicit `type=raw` line alone is inert), and then went one step further than the recommendation: **every** mutable/version tag is gated on `github.event_name == 'push'`, not just `latest`. Your finding names the semver re-publish too, and republishing `0.9.0` at a new digest is arguably worse than `latest` moving — it silently changes what an operator's `TRINITY_IMAGE_TAG=v0.9.0` pin resolves to. So a dispatch now publishes `sha-<short>` and nothing else, which is exactly what a smoke build wants; re-cutting a release is re-pushing the git tag, which fires `push` and takes the full set. Flagging the trade explicitly since it goes past what you asked for: a legitimate re-publish of the current release now needs a tag re-push rather than a dispatch.

**I2** — took it now rather than as a fast-follow. The dangerous half is introduced by this PR (a hosted `cloudflared` left serving publicly after "All services stopped"), and the script is 15 lines. Detection is the running stack's own `com.docker.compose.project.config_files` label rather than a marker file or a `trinity.db`-at-the-bind-path heuristic — compose's own record of what created the project can't drift; missing container or label degrades to today's dev default. Also switched `down` → `stop`, since start.sh's summary tells operators not to run `down` in *both* branches and a script named `stop.sh` running it was a standing contradiction.

**Docs** — added `docs/memory/feature-flows/hosted-install.md` on the `database-backup.md` shape (publish workflow → GHCR → `--hosted` pull/retag → data-switch guards → upgrade path, Testing section pointing at all three test files), plus the index row and a Recent Updates entry. Fixed the three stale "both compose files" claims in `architecture.md` (#2216, #1159, #1871). §8.9 gains HOST-017/HOST-018 for the two defects above, a corrected HOST-001 tag-gating rule, and the two test files its `Tests:` row was missing.

New guard `tests/unit/test_2280_publish_workflow_and_stop.py` (7 tests) pins all three: no raw mixed-case owner in an image reference and the built one is lowercased; `latest=false` present; every non-sha tag gated on push and the sha tag gated on nothing; `stop.sh` never runs `compose down` and selects its file from the label. 40 passed across the three #2280/#2390 files.

Process note taken — `Refs #2280` stays, AC4 (cloud-init) is in trinity-ops-public, so the issue and its status label don't move on this merge.

Issue #679, opened 2026-05-06
feat(executions): plumb cancel signal into agent task-runner reply (#671 defense-in-depth)


dolho:
## Summary

Follow-up to #671 (DB-level CAS guard). The minimum fix in #671 ensures a user cancel is never overwritten by a late SUCCESS at the database boundary. This ticket tackks the **defense-in-depth** layer the original report asked for: make the container task-runner aware that a cancel happened and surface it in the response.

## Problem (residual after #671)

When the operator cancels mid-run, Claude Code currently catches the signal, emits a graceful final message, exits with code 0, and the agent task-runner reports:

\`\`\`
INFO:agent_server.routers.chat:[Task] Task <id> completed successfully
INFO: 172.28.0.5:37822 - "POST /api/task HTTP/1.1" 200 OK
\`\`\`

The DB-level fix in #671 stops this from poisoning the schedule_executions row. But:

- Container logs still say "completed successfully" — confusing during incident triage.
- Agent has no signal that its work was repudiated; for agents with side effects (Slack post, sheet rows, CRM), the agent might still finish a half-baked deliverable on the next turn.
- The HTTP response carries no \`status=cancelled\` marker — only the gateway knows.

## Asks

- [ ] Plumb cancel signal into \`docker/base-image/agent_server/routers/chat.py\` so the task-runner knows when the parent gateway issued a cancel.
- [ ] On detected cancel, log \`[Task] Task <id> cancelled by user\` instead of \`completed successfully\`.
- [ ] Include \`status: \"cancelled\"\` in the \`POST /api/task\` reply payload so the gateway can cross-validate the DB CAS write.
- [ ] Decide whether agent should suppress the final assistant-message side effects (Slack post, sheet write) on cancel — likely yes, but needs design.
- [ ] Base-image rebuild + rollout after the change.

## Related

- #671 — minimum fix landed (DB CAS protects the row)
- \`docker/base-image/agent_server/routers/chat.py\` — task endpoint + cancel handling
- \`src/backend/services/task_execution_service.py\` — gateway-side execution

AndriiPasternak31:
Follow-up from the #679 review: the cancel signal is now correct on the `schedule_executions` row, but a cancelled execution's **activity** row is still closed as `ActivityState.FAILED` (no `CANCELLED` activity state). Observability-only, deferred from #679 — tracked in #1332.

AndriiPasternak31:
Reopening — the implementation was never actually merged.

[...]

Issue #271, opened 2026-04-06
feat(scheduler): retry mechanism for failed scheduled executions


vybe:
## Summary

Add a configurable retry mechanism for scheduled executions that fail. When a scheduled task fails (timeout, agent error, rate limit), automatically retry it after a configurable delay.

**Default: 1 retry, ON by default** — most failures are transient (agent not ready, 429 rate limits, brief network issues). Users who set up schedules expect resilience without needing to discover a retry toggle.

### `agent_schedules` — 2 new columns
```sql
max_retries INTEGER DEFAULT 1,          -- 0 = no retry, 1-5 range
retry_delay_seconds INTEGER DEFAULT 60   -- seconds between retries (30-600)
```

### `schedule_executions` — 2 new columns
```sql
attempt_number INTEGER DEFAULT 1,        -- which attempt this is (1 = first try)
retry_of_execution_id TEXT,              -- links retry to original execution
```

## Execution Flow

```
Cron fires → Execute → Failed?
  ├─ No  → Done (status=success)
  └─ Yes → attempt_number < max_retries + 1?
       ├─ No  → Done (status=failed, final)
       └─ Yes → Schedule delayed retry (retry_delay_seconds)
                → New execution record with:
                    attempt_number = previous + 1
                    retry_of_execution_id = original execution id
                    triggered_by = "retry"
```

## Design Decisions

| Decision | Choice | Rationale |
|----------|--------|-----------|
| Default retries | 1 (ON) | Most failures are transient; autonomous agents should be resilient by default |
| Retry = new execution record | Yes | Full observability — each attempt has own logs, cost, duration |
| Link retries | `retry_of_execution_id` → first attempt | UI can group attempts together |
| Delay mechanism | APScheduler `date` trigger | Survives scheduler restart (vs asyncio.sleep) |
| Which failures retry | All except `skipped` and `cancelled` | Skipped = capacity issue. Cancelled = intentional. |
| Rate limit (429) | Retry with 2x delay | Already detected in task_execution_service |
| Max retries cap | Hard cap at 5 | Prevents misconfiguration |
| Delay range | 30–600 seconds | Too short = same issue. Too long = defeats purpose. |

### Backend
- [ ] Migration: 4 new columns across 2 tables
- [ ] `db/schedules.py`: new fields in create/update/query operations
- [ ] `models.py`: new fields in schedule and execution request/response models
- [ ] Scheduler service: retry scheduling logic after execution failure (~50 lines)
[...]

PR #2393, opened 2026-08-25
feat(chat): Escape and Stop cancel the in-flight turn, and the words come back (ent#155)

dolho:
Stop a message that is still processing, from every surface where people talk to agents, and get the text back so it can be edited and resent.

## What was actually missing

The cancel machinery has existed end to end for a long time and is already in production use by `TasksPanel.vue`: terminate route → agent-server process registry → SIGINT → a CAS-guarded **CANCELLED** terminal (#679/#1332 — deliberately not FAILED, so it stays neutral for the dispatch breaker). None of that is new or re-implemented here.

What was missing: a trigger on the conversation surfaces, a way to reach it from the two credentials that are not a JWT, and a rule for what happens to the words.

## Scope — three surfaces, not four

The AC lists Session mode as a fourth. ent#358 retired it (`SessionPanel.vue` is deleted, `?tab=session` redirects to `/workspace`), so its successor takes its place rather than the AC being quietly dropped.

| Surface | Credential | Route |
|---|---|---|
| Agent Detail → Chat | operator JWT | existing `POST /api/agents/{name}/executions/{id}/terminate` |
| Public link | the link token | **new** `POST /api/public/executions/{token}/{id}/terminate` |
| Workspace | portal token / platform JWT | **new** `POST /api/enterprise/client-portal/agents/{name}/executions/{id}/terminate` |

**The public link scopes per LINK.** Its token is the credential — the same one required to start the turn — and the route sits beside `status` and `stream`, scoped identically. There is deliberately no per-visitor gate: a public link has no per-visitor identity to check against. Anyone holding the link can already read a turn's stream; stopping one is the same authority, not a wider one.

**The Workspace scopes per CALLER**, behind the same three gates as its stream route and using the *same* `execution_belongs_to_caller` function, not a second copy: roster → execution-belongs-to-agent → **started-by-this-caller**. The third is load-bearing — executions are agent-scoped, so without it any client of a shared agent could stop another client's turn by guessing an id, which is strictly worse than the read it already prevents. A miss is a uniform 404.

## `terminate_execution` is now principal-agnostic

It used `current_user` for exactly one thing: `user_id=` on the activity row. A public-link visitor and a Workspace client are real people with no `users` row, so `current_user` is optional and a NULL `user_id` is correct — with an `actor_kind` (`operator` / `public_link` / `workspace_client`) so the blank is legible rather than mysterious.

Neither new route writes a status, runs a CAS, or touches the breaker. A test pins that: a surface with its own cancel write would be a second cancel semantics for one product.

## Escape, and everything Escape must not break

Escape closes modals, dismisses the composer typeahead, leaves the voice loop. The rule is a conjunction and deliberately conservative — a missed cancel costs one click on Stop, a wrong one destroys a turn the user is still waiting for. It fires only when: the key is Escape, a turn is in flight, no cancel is already running, the event is not a composed IME candidate (Escape abandons a candidate — taking it would cancel the turn instead of the character), nothing else called `preventDefault`, and no declared overlay owns Escape. With nothing in flight it never touches the input.

## The words

Prepended, not substituted: the cancelled text is what the user is returning to edit, so it goes where the caret starts, and anything typed while waiting follows it. Nothing is dropped, and the merge is **idempotent** — Escape then Stop cannot stack two copies.

## Honest status

- Successful cancel → renders as **cancelled**, not an error.
- Lost the race → `already_terminal`, and the UI says **nothing at all**: the reply is on screen and there is no action to offer. Not a 4xx.
- **Refused** terminate → the input is left untouched and the message says the turn is still running. Restoring there would imply a stop that did not happen, and the turn is still spending. In the Workspace this gets its own dismissible line rather than `markFailed`, because the message has not failed.

## Shape

Rules are pure and shared (`utils/turnCancel.js`) — `vitest.config.js` runs `environment: 'node'` with no mount harness, so a rule decided inside an SFC is one no test can reach. The Stop control lives in the shared `components/chat/ChatInput.vue` for both chat surfaces (one control, not a second hand-built copy — #2370's lesson) and **replaces** Send rather than sitting beside it: Send is disabled for the whole turn, so a permanent second button would be a dead control most of the time.

## Tests

- `tests/unit/test_ent155_chat_cancel.py` (14) — optional principal + `actor_kind`; that neither new route re-implements cancellation; the public route's link+agent scoping; the Workspace route's caller scoping and its reuse of the stream route's gate; `already_terminal` across all four terminal statuses and real termination for `running`/`queued`; missing execution → 404; an agent-side failure reworded for a client without naming the agent host.
- `src/frontend/tests/unit/turnCancel.spec.js` (30) — every arm of the Escape rule incl. IME and `defaultPrevented`; restore/merge incl. idempotence and whitespace-only drafts; the three outcome shapes; and from source, that all three surfaces import the shared rules, capture and clear the id and text together, register/remove the listener, guard a double cancel, and do **not** restore the words on a refusal.

Full frontend suite green (58 files / 1262). Backend: 30 passed across the new file and the auth/model guards.

## Known gaps, stated

- Stop is offered only once the dispatch returns an execution id (~the first second). A control offered earlier would be a lie — there is no turn yet. The Workspace's synchronous fallback path never learns an id and correctly offers nothing.
- A SIGINT'd turn may report partial cost; the display follows the terminal envelope, as it already does for every cancelled execution.
- Rooms are out of scope: a room turn dispatches to N participants, so "stop this turn" has no single referent.

## Gating

**OSS-core by decision — deliberately ungated.** Recorded explicitly because CLAUDE.md's default for an enterprise-tracker issue is *gated unless ruled otherwise*. Two of the three surfaces are OSS-core chat, the cancel machinery is OSS-core, and Workspace ships OSS-core throughout.

Docs: requirements §5.14 (`core-agent.md`), new flow `chat-turn-cancellation.md` + index row, and an `architecture.md` endpoint row.

Related to Abilityai/trinity-enterprise#155


dolho:
## Code review — `d9d6d1a`

Reviewed against `dev`. Seven findings, the first of which makes the Workspace half of this PR non-functional. Reported as found rather than quietly fixed, since the PR is already open for review.

### Critical

**1. `PortalConversation.vue:947` — `voiceConvLive` is not defined in this file, so Escape throws on every keypress.**

```js
overlays: [typeaheadOpen.value, voiceConvLive.value],
```

Grepping the branch, `voiceConvLive` appears on exactly this line — it is not a `ref`, not a `computed`, and not imported. It belongs to ent#440 (`feature/ent440-workspace-voice-conversation`), which is not in `dev`; it was carried over by mistake.

The options object is built *before* `shouldCancelOnEscape` is called, so this raises `ReferenceError: voiceConvLive is not defined` on **every** Escape keydown in the Workspace conversation — in flight or not. Escape-to-cancel is therefore entirely dead on that surface, and the console fills with errors whenever anyone presses Escape to dismiss anything.

Worth noting *why the tests missed it*: `turnCancel.spec.js`'s surface assertions are source greps (imports present, listener registered, restore inside the success branch) and `vitest` has no mount harness, so nothing ever evaluates this handler. That is the known cost of the pure-module split, and this is exactly the class it leaves uncovered. The right substitute is the existing `voiceMode` / voice-loop ref, or dropping that overlay entry until ent#440 lands.

### High

**2. `utils/turnCancel.js:74` and all three call sites — the agent's `already_finished` is treated as a successful cancel.**

`docker/base-image/agent_server/routers/chat.py:340` returns **HTTP 200** with `{"status": "already_finished", …}` when the process registry finds the turn already gone, and `_proxy_terminate_and_finalize` passes that dict straight back through all three routes. Every surface tests only `status === 'already_terminal'` (`ChatPanel.vue:743`, `PublicChat.vue:865`, `PortalConversation.vue:964`).

So `already_finished` lands in `cancelOutcome({ok: true, alreadyTerminal: false})` → `kind: 'cancelled'` → `restoreDraft()` runs. The user clicks Stop just as the turn completes; the assistant's real reply arrives **and** their message is prepended back into the composer as though it had been stopped. The PR's "already_terminal → the UI says nothing at all" contract holds for the DB-level pre-check only, not for the agent-level race it exists to cover.

**3. `ChatPanel.vue:739` — Agent Detail has no `already_terminal` path, so a lost race reports "it's still running" for a finished turn.**

The two new routes carry a DB pre-check; ChatPanel calls the **pre-existing** operator route (`routers/chat.py:849`), which has none. When the turn has already finished, `_proxy_terminate_and_finalize` gets a 404 from the agent and raises `ChatDispatchError(404, "Execution not found in agent")` → 404 → `cancelTurn`'s `catch` sets "Couldn't stop the turn — it's still running." It is not still running; the reply is on screen. And since `cancelTurn` races `sendMessage`, that line can overwrite the correct banner `sendMessage`'s `finally` just set. The same window exists on the two new routes (DB row still `running`, agent already done → 404).

### Medium

**4. `client_portal/service.py:2086` — a deliberately cancelled Workspace turn is rendered to the client as a failure.**

`terminate_portal_turn` writes the CANCELLED terminal, but the still-running background `_run()` in `start_portal_turn` then finishes: `portal_chat` hits `if status in ("failed", "cancelled")` (:1467), no `error_code` branch matches, and it raises the generic `ClientPortalError(502, …, category="agent_error")`. `_run`'s handler calls `record_turn_outcome`, `awaitPersistedReply` returns `{failed: true, …}`, and `send()` calls `markFailed`.

So the Workspace user presses Stop, correctly gets their text back — and the thread shows their message struck out in red with "Something went wrong while the agent was working on this." There is no `cancelled` branch anywhere in the portal outcome path, so "a successful cancel renders as cancelled, not an error" does not hold on this surface.

**5. `routers/public.py:1081` — public terminate is scoped only to the agent, so a link holder can cancel other principals' executions.**

The checks are the link token and `execution.agent_name == agent_name`. Unlike the `status` / `stream` routes it is modelled on, this is a **destructive write**: any execution on that agent is cancellable given its id — the owner's own Agent Detail turn, a scheduled run, a loop iteration, another visitor's turn. Ids are 128-bit (`secrets.token_urlsafe(16)`) so this is not blind-guessable, but the route does not restrict `triggered_by`, so a leaked or shared id (a screenshot, a log, a shared browser session) lets a public visitor kill the owner's scheduled work. A `triggered_by == "public"` check would make the documented "same authority as reading the stream" claim actually true — that symmetry argument is sound for a read and does not carry to a write.

**6. `ChatPanel.vue:834` — the document-level Escape listener survives navigation, because AgentDetail is KeepAlive-cached.**

`App.vue:29` wraps the router view in `<KeepAlive :include="['AgentDetail']">`, so `onUnmounted` does not fire on nav-away — only `onDeactivated` would. The `keydown` listener stays bound to `document` while a turn is still in flight (`loading` and `activeExecutionId` remain set; the send closure outlives the navigation). The user sends a long turn, goes to the Dashboard, presses Escape to dismiss something that doesn't `preventDefault` — and the in-flight turn is cancelled silently from a page showing no chat, with none of the declared overlays visible to that handler.

### Low

**7. `PortalConversation.vue:378` — `cancelError` is never cleared when a new turn starts.** `cancelTurn` clears it at entry (:957) and Dismiss clears it, but `send()` / `deliver()` do not. A refused cancel therefore stays pinned above the composer across every later turn until manually dismissed, describing a turn that ended long ago.

obasilakis (changes requested):
## Re-review — the branch has not moved since the last review

Before looking at code I checked what changed. Nothing did. `feature/ent155-chat-cancel` is still at `d9d6d1a`, the exact commit the previous review was written against:

- `gh api repos/abilityai/trinity/pulls/2393/commits` returns a single commit, `d9d6d1a`.
- The PR timeline contains no `head_ref_force_pushed` event.
- A local `git fetch` plus `git log origin/dev..FETCH_HEAD` gives the same one commit, and `git rev-parse FETCH_HEAD` is `d9d6d1a727fdc761334e48c15e38799e250a4520`.

So the tree under review is byte-identical to the one that produced the earlier seven findings. If work was done in response to that review, it has not been pushed.

I did not take the earlier review's word for anything — I re-derived each finding from the current source. All seven reproduce. Below are the verdicts with the evidence, then three things the earlier pass did not cover, then the verdict on merging.

## Prior findings — per-item verdict

**1. `PortalConversation.vue:947` — `voiceConvLive` is undefined. NOT ADDRESSED. Blocking.**

`git grep voiceConvLive` over the branch returns exactly one hit, line 947, inside the options object passed to `shouldCancelOnEscape`. There is no `ref`, no `computed` and no import for it anywhere in the frontend. The ref that does exist is `voiceMode` (declared at :1205). Because the options object is evaluated as an argument, `onEscapeKeydown` raises `ReferenceError` before `shouldCancelOnEscape` is ever entered — on every Escape keydown in the Workspace conversation, whether or not a turn is in flight. Escape-to-cancel is dead on that surface, and Escape used for anything else there now throws.

**2. `already_finished` is treated as a successful cancel. NOT ADDRESSED. High.**

Confirmed end to end. `docker/base-image/agent_server/routers/chat.py:340` returns **HTTP 200** with `{"status": "already_finished", …}` when the process registry finds the turn already gone. `chat_execution_service._proxy_terminate_and_finalize` returns that dict verbatim (`return result`, after the `status in ["terminated", "already_finished"]` branch at :1709), and correctly does **not** write a CANCELLED terminal for it (:1721-1724). All three surfaces test only for `already_terminal` — `ChatPanel.vue:750`, `PublicChat.vue:865`, `PortalConversation.vue:964` — so `already_finished` reaches `cancelOutcome({ok: true, alreadyTerminal: false})` (`turnCancel.js:74-78`), yields `kind: 'cancelled'`, and runs `restoreDraft`. The user clicks Stop as the turn lands: the real reply appears **and** their message is pushed back into the composer as though it had been stopped.

This is the exact case the PR body promises to handle ("Lost the race → `already_terminal`, and the UI says nothing at all"). The promise holds for the DB-level pre-check the two new routes carry; it does not hold for the agent-level race, which is the narrower and more likely one.

**3. `ChatPanel.vue:739` — Agent Detail has no already-finished path at all. NOT ADDRESSED. High.**

ChatPanel calls the pre-existing operator route, which carries no DB pre-check (unlike `public.py:1117` and `service.py`'s `terminate_portal_turn`). When the row is already terminal the agent answers 404, `_proxy_terminate_and_finalize` raises `ChatDispatchError(404, "Execution not found in agent")`, and `cancelTurn`'s `catch` at :758 sets "Couldn't stop the turn — it's still running." It is not still running. Note also that `res.data?.status === 'already_terminal'` at :750 is unreachable on this surface — the operator route never emits that value — so the only two outcomes Agent Detail can produce are "cancelled" and a wrong "still running".

**4. `client_portal/service.py` — a deliberately cancelled Workspace turn is reported to the client as a failure. NOT ADDRESSED. Medium.**

`portal_chat`'s terminal classifier at :1467 is `if status in ("failed", "cancelled")`, and none of the branches below it (`AUTH`/`BILLING`, `CAPACITY`, `TIMEOUT`) match a cancellation, so it falls through to the generic `ClientPortalError(502, _turn_failed_detail(...), category="agent_error", retryable=False)` at :1507. `start_portal_turn._run`'s `except ClientPortalError` then calls `record_turn_outcome(...)`, the client's reattach poll reads that verdict, and `send()` calls `markFailed`. There is no `cancelled` branch anywhere in the portal outcome path.

One correction to the earlier report, in the PR's favour: the sibling `_fail_unstarted_execution` call on that path is safe. It goes through `update_execution_status`, which guards non-success terminals against overwriting an already-terminal row (RELIABILITY-005), so the DB row stays CANCELLED and there is no double-terminal write. The defect is confined to the client-facing verdict — which is still user-visible, and still contradicts the PR's "a successful cancel renders as cancelled, not an error".

**5. `routers/public.py:1081` — public terminate is scoped to the agent, not to the turn. NOT ADDRESSED. Medium, see also New finding 1.**

The only checks are the link token and `execution.agent_name != agent_name` (:1111). `triggered_by` is unrestricted, so any execution on that agent is cancellable given its id: the owner's own Agent Detail turn, a scheduled run, a loop iteration, another visitor's turn. Ids are 128-bit so this is not blind-guessable, but the route makes a leaked or shared id (a screenshot, a log line, a shared browser) sufficient to kill the owner's scheduled work.

The symmetry argument in the docstring — "anyone holding the link can already read a turn's stream, so stopping one is the same authority" — is sound for a read and does not carry to a destructive write. Restricting to `triggered_by == "public"` would make the claim true as stated.

**6. `ChatPanel.vue:834` — the document-level Escape listener outlives navigation. NOT ADDRESSED. Medium.**

Confirmed, and the situation is a little worse than reported. `App.vue:29` is `<KeepAlive :include="['AgentDetail']">` and `AgentDetail.vue:341` declares `name: 'AgentDetail'`, so `onUnmounted` does not fire on nav-away and the `keydown` listener registered at :834 is never removed by :844. On top of that, `AgentDetail.vue:150` renders ChatPanel under `v-show`, not `v-if` ("keeps the surface mounted so state/polling survives tab switches"), so the handler is live while the user is on the Tasks tab, the Files tab, or on another page entirely. Its overlay list is `[voice.isActive.value, showSessionDropdown.value]` (:775) — neither of which knows anything about the page actually on screen. Send a long turn, navigate to the Dashboard, press Escape to dismiss something that does not `preventDefault`, and the turn is cancelled silently. `onDeactivated` is the hook this needs.

**7. `PortalConversation.vue:378` — `cancelError` is never cleared on a new turn. NOT ADDRESSED. Low.**

Set at :968, cleared only at :957 (on re-entering `cancelTurn`) and by the Dismiss button at :183. Neither `send()` nor `deliver()` clears it, so a refused cancel stays pinned above the composer across every later turn, describing a turn that ended long ago.

## New findings

**N1. The cancel path force-releases capacity for the whole agent, and that is now reachable from an anonymous surface. Medium.**

`_proxy_terminate_and_finalize` calls `capacity.force_release(name)` (`chat_execution_service.py:1712`) on both `terminated` and `already_finished`. `CapacityManager.force_release` (`capacity_manager.py:609`) calls `slot_service.force_clear_slots`, which `ZRANGE`s and then `DEL`s `agent:slots:{name}` wholesale plus every per-slot metadata key (`slot_service.py:414-436`), and then deletes the agent's in-memory overflow queue. It is documented in-tree as "emergency use".

That was tolerable while the only caller was an operator terminate from the Tasks panel. It is much less so now: on an agent with `max_parallel_tasks > 1`, one public-link visitor cancelling **their own** turn drops slot accounting for every other in-flight execution on that agent and discards the queued overflow. Combined with finding 5 it also means any link holder can trigger an agent-wide capacity wipe. This is not something either new route can fix locally — it needs either a per-execution release on this path or an explicit decision that the blast radius is acceptable, written down.

**N2. The two new endpoints are missing from the architecture endpoint catalog. Low.**

`docs/memory/architecture.md` gains exactly one row, and it documents the **pre-existing** operator route. Neither `POST /api/public/executions/{token}/{id}/terminate` nor `POST /api/enterprise/client-portal/agents/{name}/executions/{id}/terminate` appears in the catalog, although both are new public API surface and the file has tables that would host them.

**N3. Test coverage does not reach the cases above.** The backend file is largely `inspect.getsource` substring assertions, and `test_a_live_turn_is_actually_terminated` monkeypatches `terminate_execution` itself, so no test exercises the real proxy path. There is no test for `already_finished`, none for cross-`triggered_by` cancellation on the public route, and none for the portal cancelled-renders-as-failed path. The frontend file's surface assertions are source greps, which is the acknowledged cost of the pure-module split — and finding 1 is precisely the class it leaves open.

## Things I checked and found fine

- `_require_roster` (`client_portal/router.py:110`) and the execution check in `portal_terminate_execution` both return a uniform 404, so the Workspace route is enumeration-safe. The public route's `if not execution or execution.agent_name != agent_name` is likewise uniform.
- The Workspace route's three gates and its reuse of `execution_belongs_to_caller` are exactly as described; the rate limiter is applied after the access gates, so it cannot be used to mint unbounded limiter keys.
- No schema change, so the dual-track migration rule does not apply.
- `bg-status-danger-600` / `bg-action-primary-600` in `ChatInput.vue` are semantic tokens (`tailwind.config.js:28`), not raw palette classes.
- `_fail_unstarted_execution` is CAS-guarded, so the portal path writes no second terminal (see finding 4).
- The in-flight marker and resume lock are released by `_run`'s `finally`, as the route docstring claims.
- `ScheduleExecution.status` is a plain `str`, so the `not in ("running", "queued")` pre-checks compare correctly.
- The playbook typeahead calls `preventDefault()` on Escape (`ChatInput.vue:414-418`), so the `defaultPrevented` guard does cover it. `PublicChat.vue:876` passes no `overlays` at all, but I could not find another Escape owner on that view, so I am recording this as an observation rather than a defect — worth a second look from you, since it is the one surface with no list.

## Verdict

Requesting changes. The tree is unchanged from the last review, and finding 1 alone is a hard `ReferenceError` on every Escape keypress in the Workspace, which makes one of the three surfaces this PR exists to deliver non-functional. Findings 2, 3 and 4 are each a direct contradiction of a race-safety or honest-status guarantee the PR body makes in writing.

The design is right — one shared pure rule module, one cancel semantics, per-surface auth chosen for the principal — and none of the above requires rethinking it. Suggested order: 1 (one-line), then 2 and 3 together (`already_finished` and the operator route's 404 are the same "we lost the race" outcome and should map to the same silent no-op), then 4, 6, 5, N1, 7.

Happy to re-review as soon as something lands on the branch.

src/frontend/src/components/portal/PortalConversation.vue
```diff
@@ -874,10 +925,51 @@ async function deliver(text) {
...
+// there is nothing to unwind here.
+// Escape stops the turn — but only when nothing else owns Escape. The
+// typeahead uses it to dismiss (`resolveComposerKey` → 'dismiss'/'close') and
+// the voice loop uses it to leave, so both are listed as overlays; the rule
+// itself is pure and lives in `utils/turnCancel.js`.
+function onEscapeKeydown(event) {
+  if (!shouldCancelOnEscape(event, {
+    inFlight: canCancelTurn.value,
+    cancelling: cancelling.value,
+    overlays: [typeaheadOpen.value, voiceConvLive.value],
```
dolho: **CRITICAL — `voiceConvLive` is not defined in this file, so Escape throws on every keypress.**

Grepping the branch, `voiceConvLive` appears on exactly this line — not a `ref`, not a `computed`, not imported. It belongs to ent#440 (`feature/ent440-workspace-voice-conversation`), which is not in `dev`; it was carried over by mistake.

The options object is built *before* `shouldCancelOnEscape` is called, so this raises `ReferenceError: voiceConvLive is not defined` on **every** Escape keydown in the Workspace conversation — in flight or not. Escape-to-cancel is entirely dead on that surface, and the console fills with errors whenever anyone presses Escape to dismiss anything.

Why the tests missed it: `turnCancel.spec.js`'s surface assertions are source greps, and vitest has no mount harness, so nothing ever evaluates this handler. That is the known cost of the pure-module split and this is exactly the class it leaves uncovered.

Fix: use the existing `voiceMode` / voice-loop ref, or drop that overlay entry until ent#440 lands.

src/frontend/src/utils/turnCancel.js
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * What the user is told after a cancel attempt.
+ *
+ * `cancelled` is not an error and must not render as one (AC: "honest status");
+ * a failed terminate is, because the turn is still running and still spending.
+ * `already_terminal` is neither — the reply arrived, so there is nothing to
+ * say at all.
+ */
+export function cancelOutcome({ ok, alreadyTerminal }) {
```
dolho, line 94: **HIGH — the agent's `already_finished` is treated as a successful cancel.**

`docker/base-image/agent_server/routers/chat.py:340` returns **HTTP 200** with `{"status": "already_finished", ...}` when the process registry finds the turn already gone, and `_proxy_terminate_and_finalize` passes that dict straight back through all three routes. Every surface tests only `status === 'already_terminal'` (`ChatPanel.vue:743`, `PublicChat.vue:865`, `PortalConversation.vue:964`).

So `already_finished` lands in `cancelOutcome({ok: true, alreadyTerminal: false})` → `kind: 'cancelled'` → `restoreDraft()` runs. The user clicks Stop just as the turn completes; the assistant's real reply arrives **and** their message is prepended back into the composer as though it had been stopped.

The "already_terminal → the UI says nothing at all" contract holds for the DB-level pre-check only, not for the agent-level race it exists to cover.

src/frontend/src/components/ChatPanel.vue
```diff
@@ -708,9 +726,58 @@ const sendMessage = async (userMessage, files = []) => {
...
+    // a later Escape terminate an execution that already ended, and a stale
+    // pending message would be restored over a draft the user has moved on to.
+    activeExecutionId.value = null
+    pendingUserMessage.value = ''
+    cancelling.value = false
   }
 }
 
+// ent#155: stop the turn, give the words back.
+const cancelTurn = async () => {
```
dolho, line 739: **HIGH — Agent Detail has no `already_terminal` path, so a lost race reports "it's still running" for a finished turn.**

The two new routes carry a DB pre-check; this one calls the **pre-existing** operator route (`routers/chat.py:849`), which has none. When the turn has already finished, `_proxy_terminate_and_finalize` gets a 404 from the agent and raises `ChatDispatchError(404, "Execution not found in agent")` → 404 → this `catch` sets "Couldn't stop the turn — it's still running."

It is not still running; the reply is on screen. And since `cancelTurn` races `sendMessage`, that line can overwrite the correct banner `sendMessage`'s `finally` just set. The same window exists on the two new routes (DB row still `running`, agent already done → 404).

src/backend/client_portal/service.py
```diff
@@ -2083,6 +2083,50 @@ async def _run() -> None:
     }
 
 
+async def terminate_portal_turn(agent_name: str, execution_id: str) -> dict:
```
dolho, line 2137: **MEDIUM — a deliberately cancelled Workspace turn is rendered to the client as a failure.**

`terminate_portal_turn` writes the CANCELLED terminal, but the still-running background `_run()` in `start_portal_turn` then finishes: `portal_chat` hits `if status in ("failed", "cancelled")` (:1467), no `error_code` branch matches, and it raises the generic `ClientPortalError(502, ..., category="agent_error")`. `_run`'s handler calls `record_turn_outcome`, `awaitPersistedReply` returns `{failed: true, ...}`, and `send()` calls `markFailed`.

So the user presses Stop, correctly gets their text back — and the thread shows their message struck out in red with "Something went wrong while the agent was working on this." There is no `cancelled` branch anywhere in the portal outcome path, so "a successful cancel renders as cancelled, not an error" does not hold here.

src/backend/routers/public.py
```diff
@@ -1075,6 +1077,58 @@ async def public_execution_status(
     }
 
 
+@router.post("/executions/{token}/{execution_id}/terminate")
+async def public_terminate_execution(
```
dolho, line 1081: **MEDIUM — public terminate is scoped only to the agent, so a link holder can cancel other principals' executions.**

The checks are the link token and `execution.agent_name == agent_name`. Unlike the `status` / `stream` routes this is modelled on, this is a **destructive write**: any execution on that agent is cancellable given its id — the owner's own Agent Detail turn, a scheduled run, a loop iteration, another visitor's turn.

Ids are 128-bit (`secrets.token_urlsafe(16)`) so this is not blind-guessable, but the route does not restrict `triggered_by`, so a leaked or shared id (a screenshot, a log, a shared browser session) lets a public visitor kill the owner's scheduled work.

A `triggered_by == "public"` check would make the documented "same authority as reading the stream" claim actually true — that symmetry argument is sound for a read and does not carry to a write.

src/frontend/src/components/ChatPanel.vue
```diff
@@ -764,6 +831,7 @@ watch(() => props.agentName, () => {
 // Initialize
 onMounted(() => {
   document.addEventListener('click', handleClickOutside)
+  document.addEventListener('keydown', onEscapeKeydown)
```
dolho: **MEDIUM — the document-level Escape listener survives navigation, because AgentDetail is KeepAlive-cached.**

`App.vue:29` wraps the router view in `<KeepAlive :include="['AgentDetail']">`, so `onUnmounted` does not fire on nav-away — only `onDeactivated` would. The `keydown` listener stays bound to `document` while a turn is still in flight (`loading` and `activeExecutionId` remain set; the send closure outlives the navigation).

The user sends a long turn, goes to the Dashboard, presses Escape to dismiss something that doesn't `preventDefault` — and the in-flight turn is cancelled silently from a page showing no chat, with none of the declared overlays visible to that handler.

src/frontend/src/components/portal/PortalConversation.vue
```diff
@@ -332,6 +369,14 @@ const currentSessionId = ref(props.sessionId)
 const loadingHistory = ref(false)
 const input = ref('')
 const sending = ref(false)
+// ent#155 — stopping an in-flight turn. The id arrives with the 202, so Stop is
+// offered only once there is something to stop; a turn that fell back to the
+// synchronous send has no id and correctly offers nothing.
+const activeExecutionId = ref(null)
+const pendingUserText = ref('')
+const cancelling = ref(false)
+const cancelError = ref('')
```
dolho, line 478: **LOW — `cancelError` is never cleared when a new turn starts.**

`cancelTurn` clears it at entry (:957) and Dismiss clears it, but `send()` / `deliver()` do not. A refused cancel therefore stays pinned above the composer across every later turn until manually dismissed, describing a turn that ended long ago.

dolho:
## Code review — re-review at `864bc75`

The seven findings from the last round are fixed and hold up under re-reading: `voiceConvLive` is gone, `isNoopCancel` covers both spellings at all three call sites, a 404 reads as the lost race, the public route is narrowed to `triggered_by == "public"` with the same uniform 404, and the Escape listener binds on activate.

Eight findings this round — mostly the same *classes* landing one caller, one tab, or one branch over from where they were fixed.

### Medium

**1. `PortalConversation.vue:1222` — `retry()` skips the cancelled-turn guard that `send()` got.**

`send()` now consults `cancelledExecutionIds` before `markFailed`; `retry(i)` still calls it unconditionally. Stop is offered during a retry (`canCancelTurn` = `sending && activeExecutionId`, both true). So: a message fails → Retry → Escape/Stop → the cancel succeeds, the words are restored to the composer, and the retried message is *still* struck out in red with a Retry button. That is the exact bug this PR describes fixing, on the other of the two `deliver()` callers.

**2. `PortalConversation.vue:998` — the cancelled id is recorded AFTER the terminate round-trip, so the guard is a race.**

`cancelledExecutionIds.value.add(executionId)` runs only once `store.cancelPortalTurn()` resolves. Meanwhile `deliver()` is polling, and `terminate_execution` writes the CANCELLED CAS *before* it returns its HTTP response — so a poll landing in that window sees the outcome, `awaitPersistedReply` returns `{failed:true}`, and `send()` evaluates `cancelledExecutionIds.has(...)` against a set that is still empty → `markFailed`. Record the id before awaiting, and remove it on a refusal.

**3. `ChatPanel.vue:779` — Escape cancels the turn from every other AgentDetail tab.**

`AgentDetail.vue:150` renders the chat tab with `v-show` while every other tab is `v-if`, so ChatPanel stays mounted with its document listener bound while the user is on Files, Schedules, Settings, Git. The `overlays` list covers only the voice overlay and the session dropdown. User sends a message, switches to Schedules, presses Escape to dismiss something that doesn't `preventDefault` — the in-flight turn is silently cancelled from a tab showing no chat.

This is the same failure the `onActivated`/`onDeactivated` fix in this very hunk addresses across *pages*; the in-page case needs `activeTab === 'chat'` (or an `isVisible` prop) in the conjunction.

**4. `ChatPanel.vue:701` and `PublicChat.vue:825` — a successful cancel renders as a red error.**

The poll's terminal branch sets `error.value = 'Task was cancelled'` / `chatError.value = 'Request was cancelled.'`, both rendered in the `status-danger` banner. So pressing Stop produces an error box — contradicting the AC and this PR's own flow doc, which says a successful cancel renders as cancelled, "not as an error… nothing there changed". Nothing there *did* change, and that was the mistake: the branch predates the feature, when `cancelled` only ever arrived from an operator killing someone's turn. Now that it is user-initiated it needs a neutral rendering, or suppression while `cancelling` is set.

**5. `routers/public.py:1141` — the public terminate route leaks the internal agent slug to anonymous visitors.**

`raise HTTPException(status_code=e.status_code, detail=e.detail, …)` passes `ChatDispatchError` details straight through: `"Failed to connect to agent 'foo-bar'"`, `"Timeout connecting to agent 'foo-bar'"`. `GET /api/public/link/{token}` deliberately returns only `agent_display_name` and never the slug, and no other route in `public.py` puts an agent name in a `detail`. This PR's own portal sibling rewords these for exactly this reason ("Couldn't stop the turn — the agent didn't respond."); the fully-anonymous surface should at least match it.

### Low

**6. `ChatPanel.vue:742` (also `PublicChat.vue:859`, `PortalConversation.vue:991`) — `cancelling` is never cleared on the success path.**

The refusal branches reset it; the `try` branch relies on the send/deliver `finally`. When the response is `already_terminal` / `already_finished` but the DB row is still `running` — the agent lost the turn from its process registry, which `_proxy_terminate_and_finalize` handles without writing CANCELLED — the turn keeps polling to `maxAttempts` while Stop is a permanently disabled spinner, Escape is inert (`cancelling` short-circuits `shouldCancelOnEscape`), the words are never restored, and the user ends at "Request timed out."

**7. `PortalConversation.vue:403` — `lastDeliveredExecutionId` is never reset, so a stale id can swallow a real failure.**

It is assigned only inside `if (started)` and never cleared. If a cancelled turn's `deliver` happens to return `true`, its id stays in `cancelledExecutionIds`; a later turn whose *dispatch* throws (a 429, say) never reaches `started`, so `lastDeliveredExecutionId` still names the old id, the guard matches, and `markFailed` is skipped — the user's message sits in the transcript with no error and no Retry. Clear both at the top of `deliver`.

**8. `docs/memory/requirements/core-agent.md:686` — duplicate section number.** The new `### 5.14 Stopping an in-flight turn` sits immediately above the existing `### 5.14 Workspace deliverables` (:750). Two §5.14 in one file makes the `architecture.md` and flow-doc cross-references ambiguous; this one should be 5.15.

### Verified sound, no finding

The `triggered_by == "public"` narrowing and its uniform 404; reuse of `execution_belongs_to_caller` on the portal route; `terminate_execution`'s optional principal (keyword-only, and neither helper has another `current_user` use); `rate_limiter.enforce`'s `detail=` kwarg; route ordering in both routers; `ChatDispatchError.headers`; no import cycle from `routers/public.py` → `chat_execution_service`; the `str`-Enum status comparisons; `restoreDraft` idempotence.

obasilakis (changes requested):
## Re-review at `864bc753`

I diffed the previously-reviewed tip (`d9d6d1a7`, base `36877bd5`) against the rebased head (base `88efbe7d`): stripping index and hunk-header lines, the two PR diffs are byte-identical at 1197 added/removed lines each. The rebase carried no content change, so everything below is derived from `864bc753` and the current source. All 26 checks pass.

Your seven inline replies are restatements of the previous findings rather than counter-arguments, so there is nothing to adjudicate there. Your self-review comment landed while I was reading — findings 1, 3 and 7 in it match what I had independently reached, so I am not restating those below beyond noting agreement.

## Prior findings — verdicts

**1. `voiceConvLive` undefined — FIXED.** `PortalConversation.vue:982` now passes `overlays: [typeaheadOpen.value]`, and the identifier is gone from source (it survives only in `turnCancel.spec.js:169` as an absence assertion). The semantics check out: the Workspace has no voice *overlay* — `voiceMode` (:1260) is a speak-replies TTS toggle and `listening` (:1267) is dictation, neither of which owns Escape — while the typeahead does, via `onComposerKeydown` → `'dismiss'` calling `preventDefault`, which the rule reads through `defaultPrevented`. See NEW-3 for what the list still omits.

**2. `already_finished` treated as a successful cancel — FIXED.** `turnCancel.js:34-38` defines `NOOP_CANCEL_STATUSES = ['already_terminal','already_finished']`, consumed at `ChatPanel.vue:750`, `PublicChat.vue:865` and `PortalConversation.vue:996`. I verified those two spellings are the complete set of lost-race statuses across `agent_server/routers/chat.py:340`, `routers/public.py:1131` and `client_portal/service.py:2137`, and that `cancelled_while_queued` (`chat_execution_service.py:1646`) correctly remains a real cancel.

**3. Agent Detail had no already-finished path — FIXED.** `ChatPanel.vue:763-766` returns silently on `err?.response?.status === 404`, matching `PublicChat.vue:872` and `PortalConversation.vue:1005`. One residual worth knowing: `terminate_execution` also raises 404 for "Agent not found" (`chat_execution_service.py:1786`), and the portal and public routes 404 on an authorization miss, so all three surfaces now swallow those too. That is defensible under the uniform-404 contract, but a Stop pressed against a vanished container is now completely silent.

**4. A cancelled Workspace turn reported to the client as a failure — PARTIAL, effectively still open.** See NEW-1.

**5. Public terminate scoped to the agent, not the turn — FIXED.** `routers/public.py:1124` gates on `getattr(execution, "triggered_by", None) != "public"` with the same uniform 404, placed before the `already_terminal` early return. I checked the direction rather than the presence: public turns are stamped `triggered_by="public"` at `:582`, `:613` and `:945`, and the Stop control exists only on the async path that creates the `:582` row, so the guard cannot break the feature it gates. Two links on one agent can still cross-cancel; that is the narrower residual and I am not asking for it.

**6. Escape listener outliving navigation — FIXED.** `ChatPanel.vue:175` imports `onActivated`/`onDeactivated` and `:847-851` binds and unbinds through them. `App.vue:29`'s `<KeepAlive :include="['AgentDetail']">` propagates both hooks to descendants, so nav-away now unbinds, and the shared function reference means `addEventListener` cannot double-bind. The in-page half is your finding 3 and I agree with it.

**7. `cancelError` never cleared on a new turn — PARTIAL.** `PortalConversation.vue:1204` clears it in `send()`. `deliver()` does not, and `retry(i)` (`:1226-1229`) calls `deliver` directly, so a refused cancel still outlives a Retry.

## Blocking

### NEW-1 — the cancelled Workspace turn is marked failed again on the next load, so finding 4 is deferred rather than fixed

The suppression added at `PortalConversation.vue:400` and `:1214-1216` is client-side and in-memory. The server classifier is unchanged: `client_portal/service.py:1467` still falls through to `ClientPortalError(502, category="agent_error")` for a `cancelled` status, and `_run`'s handler at `:2071` **durably records** that verdict through `record_turn_outcome`. `fetchHistory` returns it as `last_turn_outcome` (`:2345`), and `PortalConversation.vue:461` calls `markLastUserTurnFailed(outcome)` on every `loadThread`; `reattach()` does the same at `:512`.

So: press Stop, get a clean thread in that tab, then switch conversation and back — or simply reload — and the user's own message is struck out in red with "Something went wrong while the agent was working on this." plus a Retry button. A client who refreshes mid-turn gets it immediately. The `cancelledExecutionIds` set shields exactly one tab until its next load.

Your finding 2 sharpens this further: because the id is recorded only after `cancelPortalTurn()` resolves, and `terminate_execution` writes the CANCELLED CAS before returning, a poll landing in that window reaches `markFailed` against a still-empty set. Both point at the same conclusion — the durable fix belongs in the classifier or in `record_turn_outcome`, not in client memory.

### N1 — the cancel path still force-releases capacity for the whole agent, now from an anonymous surface

`chat_execution_service.py:1709-1716` is unchanged: `capacity.force_release(name)` → `capacity_manager.py:609-622`, documented in-tree as "Emergency: clear all running slots and the in-memory queue". `force_clear_slots` DELs `agent:slots:{name}` wholesale plus every per-slot metadata key and drops the overflow LIST.

That was tolerable while the only caller was an operator terminate. On an agent with `max_parallel_tasks > 1`, one public-link visitor cancelling their own turn now drops slot accounting for every other in-flight execution on that agent and discards the queued overflow — and it fires on the `already_finished` no-op branch as well, where nothing was cancelled at all. I grepped `chat-turn-cancellation.md` and `core-agent.md` for `force_release`: no written decision anywhere. This needs either a per-execution release or an explicit statement that the blast radius is accepted.

## Non-blocking

**N2 — both new endpoints are missing from the architecture endpoint catalog.** The PR adds exactly one row (`architecture.md:1088`), and it documents the pre-existing operator route. Grepping `terminate` in that file returns only lines 179, 396, 492, 854 and 1088; neither `POST /api/public/executions/{token}/{id}/terminate` nor `POST /api/enterprise/client-portal/agents/{name}/executions/{id}/terminate` appears, although both are new public API surface.

**NEW-5 — the docs now contradict the shipped code, and the fix commit updated none of them.**
- `chat-turn-cancellation.md:32-38` and `core-agent.md:708-710` still assert public cancel is scoped per *link* and is "the same authority, not a wider one" — the code's own new comment at `public.py:1118-1123` says that argument does not carry, and the guard narrows it.
- `chat-turn-cancellation.md:95` and `:139`, and `core-agent.md:731`, still name `already_terminal` alone rather than the shipped two-spelling predicate.
- `core-agent.md:725` and `chat-turn-cancellation.md:70` still list "the voice loop" as a Workspace Escape overlay.
- `core-agent.md:731` and `chat-turn-cancellation.md:91-93` still claim a successful cancel renders as cancelled, which NEW-1 shows is untrue for the Workspace.

**NEW-6 — in-code comments contradicting the fix in the same function.** `PortalConversation.vue:969-972` still says the voice loop uses Escape to leave "so both are listed as overlays", three lines above the comment explaining why it is not. `public.py:1086-1098`'s docstring still argues for no gate beyond the link, immediately above the gate.

**NEW-7 — the only test for the security narrowing passes on an inverted guard.** `tests/unit/test_ent155_chat_cancel.py:220-226` is `inspect.getsource` plus `assert 'triggered_by' in src` and `assert '"public"' in src`. Flipping the comparison to `== "public"` → 404 satisfies every assertion. Given this is a security narrowing, it should call the route (or at minimum the predicate) with a `triggered_by="schedule"` row and assert 404. More broadly on coverage: `isNoopCancel` gets a genuine unit test at `turnCancel.spec.js:133-142`, but the rest of the added frontend assertions (`:145-193`) are source greps, no test invokes `public_terminate_execution`, nothing covers `already_finished` through the backend, and nothing covers the portal cancelled-render path.

**NEW-3 — the Workspace overlay list is narrower than ChatPanel's, asymmetrically.** `ChatPanel.vue:783` guards `[voice.isActive, showSessionDropdown]` while `PortalConversation.vue:982` guards only `[typeaheadOpen]`. The agent picker (`pickerOpen`, `:412`, opened at `:17`) closes on outside-click only (`:579-580`), so pressing Escape to dismiss it during an in-flight turn cancels the turn. Same for dictation: the mic is disabled only on `transcribing` (`:235`), so `listening` can be true mid-turn and Escape kills the turn instead of the recording. This is the "a wrong cancel destroys a turn the user is still waiting for" case the rule exists to avoid. Related to your finding 3, one surface over.

**NEW-8 — `cancelledExecutionIds` (`:400`) is never pruned on the `deliver() === true` path**, so entries leak for the session. Harmless alone; it is the precondition for your finding 7.

## Verdict

Requesting changes. Seven of ten prior findings are genuinely closed, and the `triggered_by` narrowing in particular is the right shape and verified against all three dispatch sites.

Blocking: NEW-1 (fix the verdict server-side, not in client memory — and your finding 2 says the client-side guard is racy even for the tab that pressed Stop) and N1 (per-execution release or a written decision). Your finding 1 is a one-line asymmetry on the other `deliver()` caller and reproduces the defect this PR targets, so it belongs in the same pass.

Cheap to clear alongside: N2, prior finding 7 on the retry path, NEW-5 and NEW-6 doc and comment drift, and NEW-7. NEW-3 deserves a decision but need not block.

dolho:
## Re-review — ent#155, fresh pass over `08072582…06b40538`

Reviewed the diff against the **merge-base** (`08072582`), not the base tip — 18 files, +1562/−24 — read the code rather than the commit messages, and ran the suites locally.

**Tests, run against the branch tree:**
- backend `-k 'ent155 or 679 or cancel or 2320'` → **197 passed, 1 skipped**
- backend `-k 'portal or capacity or terminate or public'` → **549 passed, 1 skipped** (regression sweep for the `force_release` change)
- frontend `npx vitest run` → **60 files / 1327 tests passed**

_(Context, not a finding: CI's `pytest (base, seed 67890)` is cancelled at ~25m. That is the known base-side infra timeout hitting many PRs; #2395 raises the cap.)_

---

## Verdict on the two prior blocking findings

### NEW-1 — a cancelled Workspace turn durably recorded as `agent_error` — **FIXED, verified**

The fix is in the classifier, not in the browser, and the whole durable chain holds:

- `client_portal/service.py` — the arm is genuinely *ahead* of the ladder:
  ```python
  if status == "cancelled":
      ...
      raise ClientPortalError(
          409, "You stopped this message before the agent finished.",
          category="cancelled", retryable=False,
      )
  if status == "failed":
  ```
- The token survives persistence: `PORTAL_FAILURE_CATEGORIES` gained `"cancelled",  # the PERSON stopped it — not a failure at all`, and `record_turn_outcome` allowlists against exactly that tuple (`"category": category if category in PORTAL_FAILURE_CATEGORIES else "internal"`) — so without the tuple edit the verdict would have been silently coerced back to `internal`. `get_turn_outcome` returns the parsed dict whole, so `category` reaches `last_turn_outcome`.
- The client keys on it on **both** replay surfaces — `loadThread` → `if (outcome) markLastUserTurnFailed(outcome)` and `reattach` → `markLastUserTurnFailed(data.outcome)` — and the guard sits at the top of that one function: `if (outcome.category === 'cancelled') return`.
- I checked the one thing the commit message doesn't claim: `_run`'s except branch still calls `_fail_unstarted_execution(execution_id, e.detail)` on this path, which writes `core_db.update_execution_status(execution_id, "failed", …)`. That is **not** a durable regression — `CANCELLED` is in `update_execution_status`'s `_TERMINAL` tuple, and its contract is *"Non-success terminal writes (FAILED, CANCELLED) are guarded against overwriting any already-terminal status"*. The CANCELLED terminal is already on the row by then (the agent labels a SIGINT'd turn `status: "cancelled"` in its 200, and `execute_task`'s #679 cross-validation finalizes CANCELLED through `apply_result`). So the row stays CANCELLED and the FAILED write is a no-op.
- Guarded, not just implemented: the ordering is pinned (`assert cancelled_at < failed_at`), the `test_2320` terminal-site table row moved `502/agent_error` → `409/cancelled`, and `test_every_declared_category_is_actually_raised_somewhere` enforces the taxonomy in **both** directions — all green in my run.

### N1 — `force_release(name)` reachable from public/Workspace — **FIXED, verified**

- The emergency clear is gone from this path. Grepping `force_release` across `src/backend` leaves exactly one live caller, `services/agent_service/queue.py:96`, behind the dedicated operator force-release endpoint (`routers/agents.py:961`). Not dead code, not reachable from a visitor.
- The replacement is genuinely per-execution — `capacity_manager.py:449`:
  ```python
  slots_key = f"{self._slots.slots_prefix}{agent_name}"
  if self._redis.zscore(slots_key, execution_id) is None:
      return False
  await self._slots.release_slot(agent_name, execution_id)
  self._mem_pop(agent_name)
  ```
  i.e. exactly what `release` does, with a membership pre-check. Nothing else's accounting is touched.
- `already_finished` releases nothing: the gate is now `if result.get("status") == "terminated" and task_execution_id:`. I checked `task_execution_id` can't be falsy at any call site — the single caller (`routers/chat.py`) does `if not task_execution_id: task_execution_id = execution_id` — so folding it into the outer gate changes no reachable behaviour.
- Guarded behaviourally, not textually: `test_679_terminate_already_finished.py` now asserts `capacity.force_release.assert_not_awaited()` on **both** branches plus `release_if_matches.assert_awaited_once()` / `assert_not_awaited()`. That test fails if either fix is reverted.

Both prior findings hold up.

---

## New findings

### 1. [Medium] The public terminate route is weaker than the route that creates what it destroys — on `require_email` links

`POST /api/public/chat/{token}` refuses outright without a verified session on an email-gated link (`routers/public.py`):

```python
require_email = _agent_requires_email(agent_name)
if require_email:
    if not chat_request.session_token:
        raise HTTPException(status_code=401, detail="Session token required for this link")
    session_valid, email = db.validate_session(link["id"], chat_request.session_token)
    ...
    verified_email = (email or "").strip().lower()
```

The new terminate route asks for nothing of the sort — link token, agent match, `triggered_by`, and that's all:

```python
link = _validate_public_link(token)
agent_name = link["agent_name"]
execution = db.get_execution(execution_id)
if not execution or execution.agent_name != agent_name:
    raise HTTPException(status_code=404, detail="Execution not found")
if getattr(execution, "triggered_by", None) != "public":
    raise HTTPException(status_code=404, detail="Execution not found")
```

So on an email-gated link, someone holding only the URL — who **cannot send a message at all** — can stop turns.

The route's own comment says the identity isn't there:

> Deliberately no cross-visitor gate beyond the link: a public link has no per-visitor identity to check against (`source_user_email` is only set when the visitor verified an email)

but that parenthetical describes precisely the case where it *is* there. `routers/public.py` populates it: `source_email = verified_email or f"anonymous ({client_ip})"`, threaded into `db.create_task_execution(..., source_user_email=source_email)`. And the Workspace arm in this same PR proves the predicate already exists — `execution_belongs_to_caller` ends on `owner == (email or "").lower()`.

Honest about reachability: exploitation needs a *known* execution id, and ids aren't enumerable. But that is the identical bound the author already accepted as worth defending against when adding the `triggered_by` gate two lines up — *"one leaked id (a screenshot, a log, a shared browser) would let a visitor stop the owner's scheduled work"* — and the same sentence applies to another verified visitor's turn on the same link. Having written *"The symmetry-with-reading argument is sound for a read and does not carry"*, the consistent conclusion is that it doesn't carry here either.

Suggested shape (small): accept an optional `session_token` on the terminate route; when `_agent_requires_email(agent_name)` is true, require `db.validate_session(link["id"], session_token)` and `execution.source_user_email == verified_email`; anonymous links keep today's per-link scope unchanged. **Confidence: high** on the asymmetry, **medium** on whether to close it now vs. record it beside the cross-link residual already in `architecture.md`.

### 2. [Low] The Workspace records a durable "cancelled" verdict that nothing ever renders

The category is used purely as a suppressor:

```js
if (outcome.category === 'cancelled') return
```

and grepping `cancelled` across `src/frontend/src/components/portal/` returns only that guard, the in-memory `cancelledExecutionIds` set, and comments — no render path. The two sibling surfaces *do* say it:

- `components/ChatPanel.vue:701` → `error.value = 'Task was cancelled'`
- `views/PublicChat.vue:825` → `chatError.value = 'Request was cancelled.'`

So in the Workspace a stopped turn leaves the user's message with no reply and no explanation — on this tab and after a reload. Strictly better than the red "Something went wrong", but the PR's own *"Successful cancel → renders as **cancelled**, not an error"* is only half-delivered on the one surface the fix was written for. Two strings are dead as a result: the server's `"You stopped this message before the agent finished."` and `cancelOutcome`'s `{ kind: 'cancelled', message: 'Stopped.' }` (all three surfaces branch on `outcome.kind` and never read `outcome.message`). **Confidence: high.**

### 3. [Low] Test comment contradicts the row it annotates

`tests/unit/test_2320_portal_failed_turn_visibility.py`:

```python
    # ... The classifier now answers it ahead of
    # the failure arms, and retryable is TRUE: re-asking is exactly what someone
    # who changed their mind may want.
    ("generic_cancelled", {"error_code": None, "error": None, "status": "cancelled"},      409,  "cancelled",   False),
```

The last column is `retryable` and it is `False` — matching the service, which spends a paragraph explaining exactly why False is right (*"the only retryable verdicts are the ones where nothing reached the agent"*). The comment reads like a leftover from an earlier draft and now argues against the code it sits on. **Confidence: high.**

### 4. [Low] The newest security gate is the one with no behavioural test

`test_a_public_link_can_only_cancel_what_a_public_link_produced` asserts only:

```python
assert 'triggered_by' in src
assert '"public"' in src
assert src.count('detail="Execution not found"') >= 2
```

All three survive an inverted comparison (`!=` → `==`, which would 404 *every* legitimate public cancel), and all three survive the gate being moved below the `_terminate_execution(...)` call. Contrast the portal arm, which has real route-level behaviour tests (`test_a_live_turn_is_actually_terminated`, `test_a_missing_execution_is_a_404_not_a_silent_success`, `test_an_agent_side_failure_is_worded_for_a_client`). One stubbed-`db.get_execution` test — `triggered_by="schedule"` → 404, `triggered_by="public"` → delegates with `actor_kind="public_link"` — would put the gate's *behaviour* under guard rather than its spelling. **Confidence: high.**

### 5. [nit] Double `get_execution` on the portal cancel path

`portal_terminate_execution` → `service.execution_belongs_to_caller(...)` reads the row, then `service.terminate_portal_turn(...)` reads it again (`execution = core_db.get_execution(execution_id)`). Harmless — the second read is the one that decides `already_terminal`, so the gap can't produce a wrong authorization — but it's two round-trips where one would do.

---

## Checked and clean (with what proved it)

- **The agent-level `not_found` race.** `agent_server/routers/chat.py` answers **404** — not `already_finished` — once `registry.unregister()` has run (`terminate()` → `{"success": False, "reason": "not_found"}` when the entry is gone), and both new routes surface that as a 404. I expected this to render as *"Couldn't stop the turn — it's still running"*. It doesn't: all three surfaces early-return silently — `ChatPanel.vue:763`, `PublicChat.vue:872`, `PortalConversation.vue:1024`, each `if (err?.response?.status === 404) { cancelling.value = false; return }`. Correct, and the reasoning is written down at each site.
- **Enum/value completeness, producer side.** `NOOP_CANCEL_STATUSES = ['already_terminal', 'already_finished']` is exhaustive against the actual producers: the agent emits `terminated` / `already_finished` / 404 / 500 (`agent_server/routers/chat.py`), and the backend adds `already_terminal` (both new routes) and `cancelled_while_queued` (`_cancel_queued_if_queued`) — the latter correctly *not* in the no-op set, since it is a real cancel.
- **`triggered_by == "public"` is the right literal.** All three public dispatch sites use it (`routers/public.py` lines 582, 613, 945), including the async submit that mints the id the browser cancels with (`db.create_task_execution(..., triggered_by="public")`). The gate is not dead and not inverted.
- **`PENDING_RETRY` isn't a hole.** I checked whether the `status not in ("running", "queued")` pre-check would report `already_terminal` for a turn that is about to retry and bill. `TaskExecutionStatus`'s own state machine names the writer: `RUNNING → PENDING_RETRY  writer: scheduler retry handler (#271)` — scheduler-only, unreachable for a portal or public chat turn.
- **Escape overloading.** The conjunction holds even where the overlay lists are short. `ChatInput.vue` calls `event.preventDefault()` on its own Escape (playbook typeahead dismiss), and `shouldCancelOnEscape` refuses on `event.defaultPrevented` — so the surfaces that don't enumerate the typeahead (`PublicChat` passes no `overlays` at all) are still covered. Grepping `Escape` across `src/frontend/src/{views,components}` shows no other Escape consumer mounted alongside these three surfaces (`Portal.vue`'s `escapeStage` is a route helper, not a key handler).
- **KeepAlive listener leak.** Real fix, correctly done: `onActivated(bindEscape)` / `onDeactivated(unbindEscape)` alongside mount/unmount, and the double `addEventListener` with the same function reference dedupes per spec. Guarded by a spec assertion on `onDeactivated`.
- **Principal-agnostic `terminate_execution`.** `user_id=getattr(current_user, "id", None)` on both activity-tracking sites, `actor_kind` recorded on both, default `"operator"` — so the existing operator route's rows are byte-identical to before.
- **Docs.** The three `architecture.md` endpoint rows match the code, including the honest residual (*"Two links on one agent can still cross-cancel"*), and the feature-flow index row is present.

---

## Summary

The two blocking findings are genuinely fixed — I traced NEW-1 through the classifier, the Redis allowlist, the CAS guard on `_fail_unstarted_execution`, and both client replay paths, and N1 through `capacity_manager` and the updated `test_679` assertions. Nothing I found is critical: no data loss, no unauthenticated escalation, no broken cancel path.

Finding **1** is the only one I'd want an answer on before merge, and "record it as a residual" is a legitimate answer — it's a consistency gap with the route's own stated principle rather than a break-glass hole, and it needs a leaked execution id to matter. **3** and **4** are cheap (a comment and one test). **2** is a follow-up unless "renders as cancelled" is load-bearing in the AC.

Nothing pushed to the branch — none of this met the bar for editing someone else's PR, and finding 1 is an authorization call that belongs to the author.

dolho:
## CI red is infrastructure, not this PR

`pytest (head, seed 12345)` and `(head, seed 67890)` were **cancelled at exactly 25m18s / 25m15s** — the job's `timeout-minutes: 25`, not a test failure. The log tail shows unbroken dots to 73% with multi-minute gaps between progress lines (12:19:59 → 12:22:13 for one 72-test chunk), then `##[error]The operation was canceled`. No F, no E.

Evidence that the diff is not the cause — same workflow, same hour, other branches:

| run | shard | result |
|---|---|---|
| `AndriiPasternak31/issue-ent279` | head 67890, head 99999 | **cancelled 25m** (unrelated branch) |
| `feature/ent253-refresh-treatment` | head 67890 | success **23m** — cleared the cap by 2 min |
| `feature/2338-journey-catalog` | **base** 99999 | success **24m** |

That last row is the one that settles it: a `base` shard runs **plain `dev`**, byte-identical for every PR. A 24-minute base proves the variance is runner-side. Median is ~14m, so the cap sits at ~1.7× the median and the slow tail lands on it.

I also checked this PR's own tests for the obvious cause — a cancellation feature is exactly where a real wait would hide — and there are **zero** `sleep` / `wait_for` / `timeout=` occurrences across `test_ent155_chat_cancel.py` and `test_679_terminate_already_finished.py`.

**The fix is #2395** (`timeout-minutes` 25 → 45), already open and `MERGEABLE`, blocked only on review. Its rationale matches today's numbers exactly. Merging it unblocks this PR and every other one currently eating spurious reds.

I've re-run the two killed shards here in the meantime — one of the three head shards passed at 15m14s, so a re-run has a real chance, but it is a dice roll until #2395 lands.

dolho:
## Review findings addressed — `6e0c7d5d`

NEW-1, N1, N2 and the NEW-5 doc drift landed in the commits pushed after your `864bc753` re-review (`cf84448a`, `06b40538`, `503391ca`). This commit clears the four that were left. Verdicts against the current head:

| # | Status |
|---|---|
| NEW-1 — cancelled turn durably recorded as failed | **fixed** — `client_portal/service.py:1468` classifies `cancelled` before the failure ladder, `category="cancelled"`, so the verdict `record_turn_outcome` persists is correct rather than shielded in client memory |
| N1 — agent-wide `force_release` from an anonymous surface | **fixed** — per-execution release (`chat_execution_service.py:1699`) |
| N2 — both new endpoints missing from the catalog | **fixed** — `architecture.md:1091`, `:1092` |
| NEW-5 — doc drift | **fixed** — the "same authority" argument, the two-spelling predicate, and "renders as cancelled" all corrected; the overlay sentences are updated here |
| your finding 1 — the other `deliver()` caller | **fixed here** |
| NEW-6 — comment contradicting the fix below it | **fixed here** (`public.py`'s docstring was already corrected in `503391ca`) |
| NEW-7 — the security test passes on an inverted guard | **fixed here** |
| NEW-3 — overlay list asymmetry | **decided and fixed here** |
| NEW-8 — `cancelledExecutionIds` never pruned | **fixed** (`cf84448a`) |

### Your finding 1 — `retry()` was the other `deliver()` caller

You called it a one-line asymmetry; it was worth slightly more than a line, because the asymmetry is only *possible* while the rules live at the call site. `send()` had grown both the cancel-aware settling and the stale-error clear and `retry()` had neither, so cancelling a **retried** turn still struck the user's own message out in red with a Retry button — this feature's own defect, reproduced one caller over — and a refused cancel's message stayed pinned above the composer across every later turn.

Both callers now settle through one `settleDelivery(index, text, res)`, and `retry()` clears `cancelError` for the same reason `send()` does. A spec case pins that there are exactly two `deliver()` call sites and both go through it, so the next branch cannot re-open it.

### NEW-3 — the overlay list, decided rather than deferred

Widened to `[typeaheadOpen, pickerOpen, listening]`. Your two cases are both reachable: the agent picker closes on outside-click only, so Escape is how a user dismisses it, and the mic is disabled only while `transcribing`, so `listening` can be true mid-turn.

The decision rule is the module's own stated bias — a missed cancel costs one click on Stop, a wrong one destroys work the user is still waiting for — so an overlay is now listed whenever it *plausibly* owns Escape on that surface, not only when it provably does. The voice loop stays out and the spec keeps pinning that identifier's absence, since naming a ref that only exists on the ent#440 branch is what killed the feature the first time. Recorded in `chat-turn-cancellation.md` and `core-agent.md` as a per-surface decision rather than left implicit.

### NEW-6

The comment above `onEscapeKeydown` still described the voice loop as a listed overlay, three lines above the comment explaining why it is not. Rewritten.

### NEW-7 — exercised, and checked against the inverted guard

You were right that `assert 'triggered_by' in src` is satisfied by `== "public"`. The test now drives the route with a `triggered_by="schedule"` row and asserts both the uniform 404 and that the terminator is never reached, plus the positive case so it cannot be satisfied by refusing everything. Verified by flipping the comparison in the source:

```
--- guard inverted (!= → ==) ---
FAILED test_a_public_link_can_only_cancel_what_a_public_link_produced
--- restored ---
1 passed
```

20 backend, 40 frontend, all passing. Docs carry the corrected test count (30 → 40).

obasilakis (approved):
Approving. Every previously-blocking finding is fixed, and I re-derived each one from the current source rather than reading the commit messages.

## Prior blockers — verdicts

**NEW-1 — a cancelled Workspace turn durably recorded as a failure. FIXED.**
`client_portal/service.py:1468` now classifies `status == "cancelled"` *before* the failure ladder, so it never reaches the AUTH/BILLING/agent_error branches. It raises `409 / category="cancelled"`, `"cancelled"` is a declared member of `PORTAL_FAILURE_CATEGORIES` (`:96`), and `_run`'s handler records it durably via `record_turn_outcome(..., category=e.category)` (`:2097`). The client side matches: `PortalConversation.vue:487` returns early on `outcome.category === 'cancelled'` inside `markLastUserTurnFailed`, so `loadThread` and `reattach` no longer strike the message out on the next load. This is the durable fix in the classifier that the previous round asked for, not the in-memory tab-scoped suppression.

**N1 — cancel force-released capacity for the whole agent. FIXED.**
`chat_execution_service.py:1699-1725`: `capacity.force_release(name)` is replaced by `capacity.release_if_matches(name, task_execution_id)` (confirmed present at `capacity_manager.py:448`), so a public-link visitor stopping their own turn no longer DELs `agent:slots:{name}` wholesale or discards the overflow queue. The release is also now gated on `status == "terminated"`, so the `already_finished` no-op branch releases nothing at all — which was the part with no defensible reading.

**Public terminate weaker than the route that creates the turn. FIXED, and better than I asked for.**
`routers/public.py:1085` takes `session_token`, and on a `require_email` link demands both the token its `status`/`stream` siblings require *and* that the turn is the caller's own via `source_user_email`. Uniform 404 on a mismatch, so the refusal cannot confirm an execution id exists on the link. Open links are unchanged, which is the honest answer there — there genuinely is no visitor identity. The commit message is right that the old docstring asserted something false.

**N2 — endpoints missing from the architecture catalog. FIXED.**
`architecture.md:1090-1092` now carries all three rows, and the public one records the `triggered_by == "public"` gate plus the two-links-on-one-agent residual by name.

## Non-blocking — one edit pass, worth taking

Three of these four are comments that now assert the opposite of the code they sit beside, which is the exact class this review thread has been about.

1. **`chat_execution_service.py:1677`** — the `_proxy_terminate_and_finalize` docstring still says "force-release capacity on a terminal outcome". It does not force-release any more.

2. **`chat_execution_service.py:1730`** — "Capacity is force-released in both cases." This is false in both halves: it is a per-execution release, and it happens in one case only. It also directly contradicts the comment twenty lines above it in the same block, which correctly explains that `already_finished` no longer releases.

3. **`chat_execution_service.py:1732`** — `if task_execution_id and result.get("status") == "terminated":` is nested inside an outer `if` with the identical condition. Dead.

4. **`client_portal/service.py:1485`** — the comment reads "`retryable=True` because re-asking is exactly what a person who changed their mind may want", and the code two lines down is `retryable=False`, with a second comment arguing correctly for False. The first sentence is a leftover from an earlier draft.

5. **`routers/public.py:1085`** — `session_token: str = None` is an implicit Optional; `Optional[str] = None` (or an explicit `Query(None)`) matches the rest of the file.

## Residual worth knowing, not asking for

All three surfaces now swallow a 404 silently, and `terminate_execution` also raises 404 for "Agent not found" (`chat_execution_service.py:1786`) while the portal and public routes 404 on an authorization miss. That is defensible under the uniform-404 contract, but a Stop pressed against a vanished container is now completely silent. Recording it so it is a known consequence rather than a surprise.

Issue #2387, opened 2026-08-24
bug(ci): backend-unit-test still cancels at 25m — not a hang this time, the suite no longer fits the budget on a slow runner


obasilakis:
## Summary

`backend-unit-test` can still exhaust its 25-minute budget and cancel, reddening an arbitrary PR — but by a **different mechanism** than #2019, whose fix is present and working.

#2019 was a single test *hanging*. Its fix (`--timeout=300 --timeout-method=signal`, `if: always()` on the JUnit upload) landed and is in the workflow today. This is the suite making **steady progress the whole time** and simply not finishing inside 25 minutes — so no per-test timeout ever fires, and there is nothing for the timeout to catch.

## Evidence — PR #2385, run 32726975203

| Job | Side/seed | Duration | Outcome |
|---|---|---|---|
| [97430359824](https://github.com/Abilityai/trinity/actions/runs/32726975203/job/97430359824) | **base, 67890** | 25m16s | **cancelled at 77%** |
| 97430359776 | head, 67890 | 13m19s | pass |
| 97430359791 | base, 12345 | 14m03s | pass |
| 97430359925 | head, 12345 | 13m50s | pass |
| 97430359843 | base, 99999 | 14m52s | pass |
| 97430359870 | head, 99999 | 14m54s | pass |

Same seed, same suite: 13m19s on head, >25m on base. Five of six shards finish in 13–15m.

The tail of the cancelled job shows continuous output right up to the cancellation:

```
12:48:58  ........ [ 75%]
12:49:04  ........ [ 75%]
12:49:07  ........ [ 76%]
12:49:08  ........ [ 76%]
12:49:16  ........ [ 77%]
12:49:27  ........ [ 77%]
12:50:15  ##[error]The operation was canceled.
```

Gaps of 6–11 seconds, then cancellation 48s after the last progress line. Nothing approaches the 300s per-test timeout — which is exactly why `--timeout` did not fire. Contrast #2019's signature, which was a 3m37s gap with no output at all.

(Caveat on the evidence: I read the tail of the log but could not re-fetch it afterwards to measure every inter-line gap across the full run, so "no stall anywhere" is not proven — only that the visible tail was progressing normally and no test hit 300s.)

## Two distinct problems

**1. The suite is at ~60% of the budget on a good day.** 13–15m against a 25m ceiling leaves no headroom for a slow runner. Any runner ~1.8x slower than baseline cancels. That is not a rare draw.

[...]

PR #2395, opened 2026-08-25
ci: restore the unit suite's timeout headroom — 25 → 45 minutes

dolho:
The cap's own comment reasons from a baseline that no longer exists:

> The suite runs ~7-8 min on a normal runner, but GitHub's shared runners vary ~2-2.5x […] 25 gives real headroom for the slow tail

25 was ~3× that baseline when it was chosen. Measured across three consecutive runs of this workflow, same six-shard matrix:

| shard | duration | result |
|---|---|---|
| base, seed 67890 | 13m | ✅ |
| head, seed 12345 | 13m | ✅ |
| head, seed 67890 | 14m | ✅ |
| head, seed 99999 | 14m | ✅ |
| base, seed 99999 | 16m | ✅ |
| base, seed 12345 | **23m** | ✅ |

plus repeated `The operation was canceled` at exactly 25m on both sides. The suite roughly doubled while the cap stood still: 25 is now ~1.8× the baseline, and a 23-minute shard clears it by two minutes. That is not headroom.

## Why it matters beyond noise

The matrix runs **both sides**, so the `base` shard is plain `dev` — it reddens a PR that cannot have caused it. That is the #2019 symptom unchanged; its fix added the per-test timeout, which addresses a stalled *test* and by construction cannot address a slow *runner*.

Observed on five open PRs today, `base` and `head` roughly equally.

## Why this doesn't blind the hang detector

`--timeout=300 --timeout-method=signal` is what turns a stalled test into a **named** failure. This job cap only ever catches what that cannot: a stall the signal handler can't interrupt — a C call holding the GIL, or a process in uninterruptible I/O.

I reproduced one of the cancelled shards locally to check which of those it was. The process sat in `wchan = jbd2_log_wait_commit` (an ext4 journal commit) and was **still making progress** — slow, not hung. So these cases were anonymous cancels at 25m and will be anonymous cancels at 45m; the only thing that changes is that a merely slow runner stops producing one.

## What this is not

Buying minutes is not the fix. Making the suite faster — or sharding it with `pytest-xdist` — is, and that is a separate change. This one stops the bleeding without hiding the trend: the measured durations above are recorded in the comment, so the next reader sees the trajectory rather than just a bigger number.

Both jobs that run the full suite are raised. The 2-minute lint job and the 5-minute diff job are untouched.

### DIFF
--- a/.github/workflows/backend-unit-test.yml
+++ b/.github/workflows/backend-unit-test.yml
@@ -64,10 +64,37 @@ jobs:
     runs-on: ubuntu-latest
     # The suite runs ~7-8 min on a normal runner, but GitHub's shared runners
     # vary ~2-2.5x; a slow one was hitting the old 15-min cap at ~78% done and
-    # getting killed (a spurious red — the tests were all passing). 25 gives real
+    # getting killed (a spurious red — the tests were all passing). 25 gave real
     # headroom for the slow tail without masking a genuine hang (pytest-timeout
     # still bounds any single test).
-    timeout-minutes: 25
+    #
+    # 2026-08-25: raised 25 -> 45, because the premise above expired. The
+    # baseline is no longer 7-8 min. Measured across three consecutive runs of
+    # this workflow, same six-shard matrix:
+    #
+    #     13m  13m  14m  14m  16m  23m   (all SUCCESS)
+    #
+    # plus repeated `The operation was canceled` at exactly 25m on both sides.
+    # So the suite roughly doubled while the cap stood still: 25 was ~3x the
+    # baseline when it was chosen and is ~1.8x now, and a 23-minute shard clears
+    # it by two minutes. That is not headroom, and the resulting red lands on
+    # whichever PR is open — including the `base` side, which is plain `dev` and
+    # has nothing to do with the PR it reddens (the #2019 symptom, unchanged).
+    #
+    # 45 restores the ~3x the original comment was reasoning about. It does NOT
+    # weaken hang detection in the way the number suggests: `--timeout=300
+    # --timeout-method=signal` below is what turns a stalled TEST into a named
+    # failure, and this cap only ever catches what that cannot — a stall the
+    # signal handler cannot interrupt (a C call holding the GIL, or a process in
+    # uninterruptible I/O). Those were anonymous cancels at 25m and will be
+    # anonymous cancels at 45m; the only thing that changes is that a merely
+    # slow runner stops producing one.
+    #
+    # The honest alternative is to make the suite faster (or shard it with
+    # xdist) rather than keep buying minutes. That is worth doing and is not
+    # this change; raising the cap stops the bleeding without hiding the trend,
+    # since the durations above are recorded here for the next reader.
+    timeout-minutes: 45
     strategy:
       fail-fast: false
       matrix:
@@ -233,7 +260,9 @@ jobs:
     # job is meant to be cheap enough that nobody is tempted to delete it.
     if: github.event_name == 'push'
     runs-on: ubuntu-latest
-    timeout-minutes: 25
+    # Same suite, same runner variance, same reasoning as the matrix job above
+    # (2026-08-25: 25 -> 45).
+    timeout-minutes: 45
     steps:
       - uses: actions/checkout@v7
         with:


dolho:
## Measured the failure rate — it is ~1 in 3 PRs, not an occasional flake

Sampled the last 12 runs of this workflow, counting only shards that actually executed the suite (excluding superseded-run cancellations, which have short durations and would bias the rate down):

```
shards that actually ran the suite: 48
  succeeded: 29   hit the 25m wall: 3  (6.2%)
  duration  median 13.7m   p90 15.3m   max 25.3m
  P(at least one of 6 shards killed on a PR) = 32%
```

**The per-shard rate is only 6%, but the matrix runs six shards, so a third of all PRs get a spurious red.** That is not a hypothetical: I hit it on four open PRs simultaneously today — #2379, #2382, #2386, #2393 — each with exactly one killed shard, and three of those on the **base** side, i.e. plain `dev`, which the PR cannot have caused.

The distribution is the useful part: **median 13.7m, p90 15.3m** — the bulk is tight, and the kills are a far tail at ~1.8× the median, which is exactly the shared-runner variance this workflow's own header documents. So this is not a suite that is creeping up on the cap; it is a suite comfortably under it that loses a coin flip 6% of the time.

At 45m the cap sits at ~3× p90, which absorbs that tail with room. Re-running is the only alternative and it costs ~14 minutes of compute per attempt at the same 6% odds, on a check whose semantic verdict (`regression diff`) has usually already passed.

One thing worth noting for whoever reviews this: the four PRs above are currently blocked on nothing but this. I re-ran all four shards rather than wait, but that is treating the symptom.

AndriiPasternak31:

**/review: merge-ready.** All CI green; the 45-min value is corroborated by this PR's own 21m52s passing shard and current duration history (~14m median, 23m worst observed → restores the original ~3× design margin). Verified in the head file: both full-suite jobs raised, the 2-min lint and 5-min diff jobs untouched, and `--timeout=300 --timeout-method=signal` intact in both jobs — the #2019 per-test hang detector is genuinely unweakened.

**/validate-pr: one required item** — add `Fixes #2387` to the body. That open P2 describes exactly this defect and mechanism; without the keyword it strands open with its symptom fixed and `issue-status-on-merge.yml` never promotes it. (The bare #2019 mention is right as context — don't convert it to a closing keyword.)

**Sibling gap found during the completeness sweep (pre-existing — not a blocker here, but urgent):** `.github/workflows/backend-unit-nightly.yml:77` carries the same expired 25-min cap over six sequential full-suite runs — the last 8 nightly runs (2026-08-19 → 08-26, runs 32225399962 → 32940531599) all concluded `cancelled` at ~07:24, so the cross-PR regression sweep (#715) has been silently dead for over a week (a cancelled scheduled run reddens nothing on any dashboard). It needs a restructure (per-(side,seed) matrix like this workflow, or a ~6× cap) plus the `--timeout=300 --timeout-method=signal` guard its pytest invocations also lack — proposing that as its own issue rather than a hunk in this PR.

dolho:
Fresh data from today (2026-08-26), same six-shard matrix, confirming the trend this PR records — and it is now actively costing merges:

| branch | shard | result |
|---|---|---|
| `feature/ent155-chat-cancel` (#2393) | head 12345, head 67890 | **cancelled 25m** |
| `AndriiPasternak31/issue-ent279` | head 67890, head 99999 | **cancelled 25m** |
| `feature/ent253-refresh-treatment` | head 67890 | success **23m** |
| `feature/2338-journey-catalog` | **base** 99999 | success **24m** |

Two independent PRs killed in the same hour, and two more that cleared the cap by 1–2 minutes.

The `base` row is worth singling out, since it is the argument this PR already makes: a base shard runs **plain `dev`**, identical for every PR. A 24-minute base is not attributable to any diff — it reddens whichever PR happens to be open. Median across all shards observed today is ~14m, so the cap is at ~1.7× median, exactly the ratio this PR's comment says has expired.

#2393 is blocked on this. Worth merging ahead of the queue.

dolho:
One more data point, and it is the cleanest one available.

PR #2408 changes **six files, all of them CI scripts and templates** — no backend code, no tests outside one new unit file. Its `base` shards, which run **plain `dev`**:

| shard | duration |
|---|---|
| base, seed 67890 | **21m18s** |
| base, seed 99999 | **18m40s** |
| base, seed 12345 | 13m11s |

All passed. But 21m18s is **3m42s from the cap**, on a shard whose code is byte-identical to `dev` and has nothing to do with the PR it belongs to. Same run's head shards: 14m35s / 12m57s / 14m56s. So a 2.5m spread on head and an 8m spread on base, same commit, same hour — the variance is entirely runner-side, exactly as this PR argues.

Separately, #2393's two shards that were cancelled at 25m yesterday **passed on a plain re-run at 14m26s and 14m23s**, no code change. That is the direct measurement of what the current cap costs: a spurious red plus a full six-shard re-run (~90 runner-minutes) to recover a result the first run would have produced in 26 minutes.

obasilakis (approved):
Approving.

I checked the premise rather than the number. The cap's own comment reasons from a 7-8 min baseline that no longer exists, and the six measured durations in the diff (13/13/14/14/16/**23**) put a real shard two minutes under the old 25m cap. That is not headroom, and the `base` shard is plain `dev` — so the resulting red lands on a PR that cannot have caused it.

What I specifically checked before agreeing this does not blind anything:

- `--timeout=300 --timeout-method=signal` is what turns a stalled *test* into a named failure. This job cap only ever catches what that cannot — a stall the signal handler cannot interrupt. Raising it changes which *runner speeds* produce an anonymous cancel, not which *hangs* are detected.
- Both jobs that run the full suite are raised; the 2-minute lint job and the 5-minute diff job are untouched.
- The measured durations are recorded in the comment, so the next reader sees the trajectory rather than just a bigger number. That is the part that makes this defensible as a stop-the-bleeding change.

One non-blocking note: the honest alternative the body names — make the suite faster, or shard it with `pytest-xdist` — has no issue number attached. Worth filing, because the worst case being bought here is 6 shards x 45 minutes, and nothing in the tree will remind anyone that the trend is the actual problem.

Issue #447, opened 2026-04-22
Architecture Validation: 4 critical violations found (2026-04-22)


vybe:
## Automated Architecture Validation Report

**Date**: 2026-04-22
**Result**: 8/16 invariants FAIL — 4 are P0-P1 critical violations requiring immediate attention

---

#### #1 — Three-Layer Backend: Router → Service → DB (FAIL)
Routers contain raw SQL, bypassing the service/db layers:
- `src/backend/routers/git.py:298` — `db.execute_query("DELETE FROM agent_git_config ...")`
- `src/backend/routers/processes.py:780` — `cursor.execute("CREATE TABLE IF NOT EXISTS process_schedules ...")`
- `src/backend/routers/processes.py:819` — `cursor.execute("INSERT OR REPLACE INTO process_schedules ...")`
- `src/backend/routers/processes.py:865` — `cursor.execute("DELETE FROM process_schedules ...")`
- `src/backend/routers/triggers.py:254` — `cursor.execute("SELECT * FROM process_schedules ...")`

#### #3 — Schema in `db/schema.py`, Migrations in `db/migrations.py` (FAIL)
`CREATE TABLE` statements found outside the canonical schema files:
- `src/backend/routers/processes.py:780` — `CREATE TABLE IF NOT EXISTS process_schedules`
- `src/backend/services/process_engine/services/alerts.py:163` — `CREATE TABLE IF NOT EXISTS cost_thresholds`
- `src/backend/services/process_engine/services/alerts.py:178` — `CREATE TABLE IF NOT EXISTS cost_alerts`
- `src/backend/services/process_engine/services/templates.py:117` — `CREATE TABLE IF NOT EXISTS process_templates`
- `src/backend/services/process_engine/repositories/audit.py:62` — `CREATE TABLE IF NOT EXISTS audit_entries`
- `src/backend/services/process_engine/repositories/sqlite_definitions.py:70` — `CREATE TABLE IF NOT EXISTS process_definitions`
- `src/backend/services/process_engine/repositories/sqlite_executions.py` — process_executions and step_executions tables
- `src/backend/services/process_engine/repositories/sqlite_events.py` — `CREATE TABLE IF NOT EXISTS execution_events`

#### #8 — Auth Pattern: `Depends(get_current_user)` (FAIL)
`src/backend/routers/docs.py` exposes routes with no authentication:
- Line 33: `@router.get("/index")` — no auth dependency
[...]

PR #2397, opened 2026-08-25
fix(subscriptions): the provider's warning tier is not a rate limit (#2396)

### LINKED ISSUE #2396: bug: subscription reports "rate-limited" for the provider's own warning tier — allowed_warning fails the _headroom_indicates_limited catch-all
## Summary

`_headroom_indicates_limited` treats any window status that is not exactly `"allowed"` as a limit. Anthropic's unified rate-limit headers ship **`allowed_warning`** — a status that means *allowed, but close to the window* — so a **healthy** subscription approaching its 7-day limit is reported as rate-limited on every surface that consumes `resolve_rate_limited_now`: `GET /api/subscriptions/{id}/usage`, the per-agent auth chip, and the subscription pressure tile.

This is the third bug in the family after #2352 (auth failures counted as rate limits) and #2353 (a rejected token displayed as a rate limit). Those two split the *failure-event* side; this one is the *provider-snapshot* side, and it is the only one that mislabels a subscription that is working perfectly.

## Context

**The catch-all is deliberate — the defect is its safety assumption.** The `w.status not in ("allowed",)` comparison exists so an unknown *blocking* status is caught without anyone enumerating the provider's vocabulary, and that intent is pinned by `test_a_blocked_window_under_an_ok_status_still_reads_as_limited` in `tests/unit/test_447_subscription_reset_and_recovery.py`. What it cannot do is distinguish a non-blocking status whose name happens not to be the literal `"allowed"`. `allowed_warning` is exactly that, and it appears **nowhere else in the tree** — a grep across `src/`, `tests/` and `docs/` returns zero hits, so the value has never been considered anywhere.

**Observed live, not theorised.** A stored provider snapshot on a running instance recorded `seven_day.utilization_pct: 90.0` with `seven_day.status: "allowed_warning"` and `unified_status: "allowed_warning"`, while `overage_status` was `"allowed"` — i.e. the provider explicitly said *allowed*. Reproduced by writing a fresh snapshot at 78% / `allowed_warning`: `GET /api/subscriptions/{id}/usage` returned `rate_limited_now: true`.

**The blast radius is the worst possible window.** The mislabel fires precisely when a subscription nears its weekly limit — the moment the operator most needs an accurate reading — and persists for up to `FRESHNESS_SECONDS` (1800s) after each probe. The operator's rational response to a `LIMIT` badge is to stop work, wait out a window, or reassign agents off a subscription that was never actually blocked.

**A downstream consumer depends on the answer.** `abilityai/trinity-enterprise#434` will consume these predicates as its "is this subscription's reading assessable" gate. The natural candidate, `_headroom_indicates_healthy`, also rejects `allowed_warning` — so a subscription sitting in the warning tier reads as *not assessable* even though its reading is perfectly good. The semantics have to be settled here, once, and consumed there rather than re-derived.

## Acceptance Criteria

- [ ] A fresh snapshot whose window status is `allowed_warning` does **not** set `rate_limited_now` on any consuming surface (`/usage`, per-agent chip, pressure tile).
- [ ] Unknown and blocking statuses still fail safe: the catch-all is **narrowed to a known-good allowlist**, never replaced by an enumeration of blockers. `blocked` and any unrecognised value still read as limited, and the existing `test_a_blocked_window_under_an_ok_status_still_reads_as_limited` stays green unchanged.
- [ ] The `_headroom_indicates_healthy` side is decided **explicitly and documented in the docstring**: whether `allowed_warning` is positive proof of headroom (may clear a stale 2h db predicate) or falls through to it. This is the rule ent#434 will consume, so it must be stated rather than implied.
- [ ] The status vocabulary lives in **one named constant**, so a future provider status needs one edit, not two (#2157 one-gate discipline — the two predicates must not drift).
- [ ] Test coverage for each arm: `allowed`, `allowed_warning`, `blocked`, an unrecognised value, and the stale-snapshot case (which must keep falling through to the db predicate regardless of status).
- [ ] No new provider round-trip — this is a pure predicate change over a snapshot that already exists. No probe-volume or quota change.
- [ ] The pressure tile tells the truth about the new state: a warning tier must not silently render as "ok" now that it no longer renders as "rate-limited".

## Technical Notes

- `src/backend/services/subscription_headroom_service.py` — `_headroom_indicates_limited` (the `w.status not in ("allowed",)` comparison) and its deliberate non-negation mirror `_headroom_indicates_healthy`.
- Both feed `resolve_rate_limited_now`, THE one derivation (#2157). Do not add a second definition; consumers are `decorate_usage` and `pressure_states`.
- Tests to keep green: `tests/unit/test_447_subscription_reset_and_recovery.py`, `tests/unit/test_471_subscription_usage_observability.py`.
- Display precedence lives in `src/frontend/src/utils/subscriptionPressureTile.js` (#2353) — check whether the warning tier earns its own chip word rather than falling into `ok`.
- Siblings: #2352 (failure-kind split), #2353 (invalid_token precedence). Blocks `abilityai/trinity-enterprise#434`.

webmixgamer:
## Summary

- `_headroom_indicates_limited` judged a window by `status not in ("allowed",)`, so Anthropic's own **`allowed_warning`** tier — *allowed, nearing the window* — read as a hard rate limit on every surface consuming `resolve_rate_limited_now`: `/usage`, the per-agent chip, the pressure tile. A **healthy** subscription approaching its weekly limit wore `LIMIT`.
- Narrowed the catch-all to a named **allowlist** (`NON_BLOCKING_WINDOW_STATUSES`) rather than inverting it to a blocklist — a deny-check silently admits every future value (#848), and here the safe default runs the other way, so an unrecognised status still reads as limited. One constant, both predicates, plus a parity test pinning the frontend twin.
- `allowed_warning` now also counts as **positive proof of headroom** — deliberate, documented, and the rule `abilityai/trinity-enterprise#434` will consume as its assessability gate.

Third in the family after #2352 and #2353. Those split the *failure-event* side; this is the *provider-snapshot* side, and the only one that mislabels a subscription that is working perfectly.

## Evidence

Verified **before** the fix on a running instance: a stored snapshot recording `seven_day: 90.0 / allowed_warning` with `overage_status: allowed` sat beside **47 successful executions** in the same two hours and **zero** `subscription_rate_limit_events`. All failures in that window were `Agent unreachable — transport circuit breaker open`, i.e. container networking, nothing to do with quota.

## Changes

| File | What |
|---|---|
| `services/subscription_headroom_service.py` | `NON_BLOCKING_WINDOW_STATUSES`; both predicates rewired; the headroom decision documented in the docstring |
| `utils/subscriptionPressure.js` | Shared vocabulary + `isWindowBlocking` / `providerWarnsNearLimit` |
| `utils/subscriptionPressureTile.js` | `bindingWindow` uses the shared vocabulary; one `isNearingLimit` feeds severity/warnReason/headline; `showsReset` gains a `nearing` arm |
| `tests/unit/test_2396_provider_warning_tier.py` | 18 tests incl. the cross-language parity pin |
| `subscriptionPressureTile.spec.js` | 11 tests incl. the row-level assertion |
| `architecture.md`, 2 feature flows, `learnings.md` | 3 stale claims corrected; the vocabulary documented at its canonical home |

**Caught in review, fixed in this PR:** demoting the row from `crit` to `warn` would have *silently removed the reset time* — `crit` is what put it there, and 78% sits below the 85% high band. The row whose entire message is "this is about to matter" would have stopped saying when the limit clears. `showsReset`'s third arm repairs it; two regression tests fail without it. Captured in `learnings.md` as a durable class.

## Blast radius

Bounded by construction — the window arm is the **weakest of three independent detectors**. A real HTTP 429 sets the top-level snapshot `status` and is checked *first*; the 2h db predicate is untouched. No schema, no migration, no endpoint, no probe or quota change.

## Test Plan

- [x] Backend: 18 new tests pass; **6 fail without the fix**
- [x] Frontend: 11 new tests pass; **6 fail without the fix**
- [x] Full suites: **12356** backend / **1254** frontend green (1 pre-existing failure in the private submodule, reproduced with all changes stashed)
- [x] Live instance: 78% and 90% warning → `false`; `blocked` → `true`; healthy → `false`; **the real 2026-08-24 provider bytes replayed** → `false`. Instance restored to its real state.

Fixes #2396


AndriiPasternak31:

**Both passes: merge-ready / approve.** All 5 ACs of #2396 verified by execution, not reading: both predicates rewired through the one shared constant (`subscription_headroom_service.py:510` / `:547-549`) with the source-level guard test forbidding re-inlined literals; the #2352 display-vs-candidate split preserved by construction (zero db-layer changes, grep-verified); the #2353 stale/rejected-token/transport-error fall-throughs untouched and test-pinned; the fix's own second-order regression (severity demotion silently removing the reset time) caught and repaired in-PR with regression tests. Docs fully synced (architecture.md in both required places, both feature flows, learnings entry). 18 backend + 91 frontend tests re-run locally, green.

**The one red check is not this PR**: `pytest (head, seed 67890)` log ends at 95% progress with "The operation was canceled." at 25m16s, zero failure markers, no junit artifact — the 25-min cap that #2395 raises. The same seed passed on base (13m10s) and both other head seeds passed. Recommend landing #2395 first so this goes deterministically green rather than re-running the shard on runner-speed lottery.

Non-blocking notes:
- Agent chips now show nothing for a warning-tier subscription — the batch payload (`/api/agents/subscription-pressure`) carries `utilization_5h_pct` + `token_status` but no 7d window/status, so "near limit" is visible only on the subscription tile. Follow-up candidate if operators watched agent chips for this: add a warning bit to the batch row.
- Pre-existing, not this PR: `docs/memory/requirements/security.md:192` still describes `rate_limited_now` with the pre-#447 "2h … OR fresh provider status" wording — worth a one-line touch-up in a follow-up.

dolho (approved):

PR #2404, opened 2026-08-26
DO NOT MERGE — regression replay to prove the journey gate goes red (#2336 AC 3)

dolho:
Throwaway. Reverts `ecf1327f` (the #2092 fix) on top of `dev` so the 2026-08-14 "stopped agents will not start" regression is live again, and carries #2335/#2336's tier + gate.

**Expected: `journey-smoke` FAILS on `test_a_stopped_agent_can_be_started_again`.** If it passes, the gate does not have teeth and #2403 is not ready to be a required check.

Closing and deleting once observed.


dolho:
Verified — this is what it existed for.

`journey-smoke` went **RED** on `test_a_stopped_agent_can_be_started_again`:

```
AssertionError: starting stopped agent 'pytest-ephemeral-journey-e15347a9'
answered 500: {"detail":"Failed to start agent (ValueError — details in backend logs)"}
— this is the 2026-08-14 regression: five of six stopped agents answered 500 here
and it reached users
```

Two things this replay established that a green run never would have:

1. **The issue's replay instruction is imprecise.** `ecf1327f` (#2092) is the commit that CAUSED the regression; reverting it removes the bug. The fix to revert is `35a24d0c` (#2186). My first attempt reverted `ecf1327f` as written and the gate passed — correctly, because the code was fine.
2. **The first version of the journey had no teeth.** #2186's title says the condition exactly: the 500 happened *when a recreate is needed*. A fresh agent stopped and started immediately has no config drift, so the bug is unreachable. The journey now changes resource limits while stopped to force the recreate path. Without that step it passed against the broken code — measured, not assumed.

Closing; branch deleted. Real work is #2403.

Issue #2398, opened 2026-08-25
bug: agent-server event loop starves under sustained parallel headless_executor load — 1 core pegged, localhost:8000 stops accepting, DORMANT breaker fires


oleksandr-korin:
## Summary

Under sustained parallel `headless_executor` load, a single Python thread inside `agent-server` burns 100% of one CPU core, the asyncio event loop stops accepting new connections (`curl localhost:8000` from *inside* the container times out at 133s+), backend heartbeats/health probes fail, and after 10 consecutive probe failures the platform files a `circuit_breaker_dormant` alert on the agent. Sanitizer regex is verified fixed (PR #1670 present) — so this is a distinct root cause from #1661 / #728 / #407, all confirmed closed.

Repeated ≥3× on the same production agent within 48h. Cleared each time by API stop→start + circuit-breaker reset (state-clearing workaround, not a fix). Recurs under load.

## Component

Agent Runtime (`agent-server`) — specifically the async task/execution management path when multiple long-running headless tasks run in parallel.

## Priority

**P1** — major feature broken under real workload (public-link chat / concurrent scheduled tasks). No auto-recovery: `HEALTHCHECK` isn't defined on the agent image, so Docker won't restart the container; the circuit breaker's own "DORMANT" state is filed as an operator-queue alert (silent until a human acknowledges).

## Theme

reliability

## Environment

- Trinity version: **v0.9.0** (`0cb3a4e2`) — updated 2026-08-25 13:19 UTC, ~5h before the py-spy dumps were captured. **Original issue text said v0.8.5 — that was correct at filing but stale by the time the overnight captures fired. Corrected 2026-08-26.**
- Agent base image: `trinity-agent-base:latest` = `dbf75ba2affe` (built 2026-08-25 13:16 UTC as part of the v0.9.0 upgrade)
- Python 3.13.15 (agent), Python 3.13 (backend)
- Docker 24+, Ubuntu 22.04 host
- Postgres backend (enterprise edition), Redis auth enabled

## Observed symptoms (captured live)

At the moment the platform's `circuit_breaker_dormant` alert fired, the affected agent container showed:

| Signal | Observed |
|---|---|
| Container CPU (`docker stats`) | **101.25%** — one core fully pegged |
| Container memory | 110 MB / 4 GB — **not OOM**, `oom_kill=0`, `RestartCount=0` |
| Container PIDs | 13 (baseline healthy peer: 6) |
| `agent-server` process | 1 process, 9 Python threads |
[...]

PR #2405, opened 2026-08-26
fix(sanitizer): the key test was still quadratic — one call below the #1670 fix (#2398)

dolho:
## Summary

**This is my regression.** #1670 fixed the catastrophic backtracking in the credential sanitizer's *line* scan (#1661) and left the identical shape one call deeper, in `_is_sensitive_kv_key`. Every one of the 14 py-spy dumps is on that stack.

```
<genexpr> (credential_sanitizer.py)      <- the any(...) in this function
_is_sensitive_kv_key
_redact_kv_match  ->  sanitize_text  ->  sanitize_dict (recursive)
sanitize_subprocess_line  ->  read_stdout  (headless_executor)
```

## The cost

`re.search(r'(?:.*TOKEN.*)\Z', key)` retries the leading `.*` from every offset — O(n²) in key length. Measured on 3.13:

| key length | old | new | |
|---|---|---|---|
| 4 KB | 0.32 s | — | |
| 16 KB | 5.31 s | — | |
| 44 KB | **39.8 s** | **0.00035 s** | 112,000× |
| 44 KB (agent copy, 30 patterns) | **43.9 s** | **0.00066 s** | 67,000× |

Forty seconds of CPU **for one key**, per match, per line, on a thread holding the GIL. That is the pegged core, and it is why the event loop stopped accepting.

## Why a quadratic test looked harmless

A wrong docstring. It described `key` as *"a short `KEY=` name"*. It is not — the key is whatever preceded an `=` in `_KV_LINE_RE`'s **unbounded** `([^\s"'=]+)`. And the stack above reaches it from **stream-json tool results** through recursive `sanitize_dict`, so multi-KB "keys" are the normal case, not an adversarial one. That single wrong assumption is what let the quadratic survive the #1670 review.

## The fix

Every pattern is `.*LITERAL.*` or `LITERAL.*`. Under `.search()` with a trailing `\Z`, **both mean exactly "the key contains this literal"** — so the test is substring containment: linear, C-level, no backtracking.

**Verdicts are unchanged, and that was the part to get right — a faster check that redacts *less* is a credential leak.** Pinned against the old implementation as an oracle: **5,420 keys on the backend copy and 9,607 on the agent copy, zero disagreements**, including the deliberate #1670 suffix semantics (`DB_` still redacts `MY_DB_PASSWORD`).

The literals are **derived** from the patterns and the derivation **refuses** anything that is not a plain literal — so a future pattern needing real regex fails loudly at import instead of being silently reduced to a check that redacts less. That is the guard #1670 should have had.

## Both copies — and the agent one is the one that matters

Invariant #5. The two files are vendored in *shape*, not byte-identically: the agent list is `SENSITIVE_VAR_PATTERNS` with 30 entries to the backend's 13. **Every py-spy dump was taken in the agent copy**, so a backend-only fix would have left all 14 burns exactly where they were. A test asserts the agent copy carries the fix.

## What this does and does not close

- **Closes** the burn: the hot path is now linear, so the arrival of a large tool result cannot peg a core.
- **Does not** address the zombie `schedule_executions` rows or the missing `HEALTHCHECK` on the agent image (no auto-recovery). Those are real and separate; the burn was making them visible.
- The mitigations in place (`max_parallel_tasks` 10→5, `execution_timeout` 3600→1800) cap blast radius and are still worth keeping until this ships to the fleet — **note the agent copy means a base-image rebuild is required for existing agents.**

## Test Plan

- [x] `tests/unit/test_2398_sanitizer_key_redos.py` — verdict equivalence against the old implementation, the #1670 suffix semantics, linearity at 64 KB, end-to-end `sanitize_text` on a 44 KB token, the derivation's refusal of a non-literal pattern, and the agent-copy parity assertion
- [x] Equivalence oracle: 5,420 backend / 9,607 agent keys, **0 disagreements**
- [x] `727 passed` across the sanitizer, #1661 and #1670 suites
- [ ] Base-image rebuild + fleet rollout (the agent copy is where the burn is)
- [ ] Watchdog confirmation over a full night with the fix deployed

Related to #2398 · follow-up to #1670 / #1661

### DIFF
--- a/docker/base-image/agent_server/utils/credential_sanitizer.py
+++ b/docker/base-image/agent_server/utils/credential_sanitizer.py
@@ -144,14 +144,65 @@ _KV_LINE_RE = re.compile(r'(?<![^\s"\'=])([^\s"\'=]+)=(["\']?)([^\s"\']+)\2')
 # composed regex could start matching mid-token, so `DB_.*` redacted
 # `MY_DB_PASS=x`. Anchoring with \Z reproduces exactly that (a `fullmatch`
 # rewrite would silently redact LESS — a leak, not a speedup).
-_SENSITIVE_KEY_SUFFIX_RE = [
-    re.compile(r'(?:' + p + r')\Z', re.IGNORECASE) for p in SENSITIVE_VAR_PATTERNS
-]
+#
+# --- #2398: the same quadratic shape, one call deeper -----------------------
+# #1670 made the LINE scan linear (#1661) and left THIS test quadratic:
+# `re.search(r'(?:.*TOKEN.*)\Z', key)` retries `.*` from every offset, so cost
+# is O(len(key)^2). Measured on 3.13: 4 KB key = 0.32s, 44 KB key = 40s — for
+# ONE key, per match, per line. py-spy caught it 14 times across three agents,
+# every dump on this stack:
+#
+#     <genexpr> (credential_sanitizer.py)      <- the `any(...)` below
+#     _is_sensitive_kv_key
+#     _redact_kv_match  ->  sanitize_text  ->  sanitize_dict (recursive)
+#     sanitize_subprocess_line  ->  read_stdout  (headless_executor)
+#
+# The old docstring called `key` "a short KEY= name". It is not. The key is
+# whatever preceded an `=` in `_KV_LINE_RE`'s `([^\s"\'=]+)`, which is
+# UNBOUNDED — and the stack above reaches here from stream-json tool RESULTS,
+# so multi-KB "keys" are the normal case, not an adversarial one. That wrong
+# assumption is what made a quadratic test look harmless.
+#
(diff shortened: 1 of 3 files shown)


dolho:
Superseded by #2406, which targets **main**.

The burn is on client instances and they run `main`. Both sanitizer copies on main carry the vulnerable code, and the agent copy's `_is_sensitive_kv_key` is at **line 150 there — exactly the line in every py-spy dump**, which confirms which tree the dumps came from.

The two files have diverged between main and dev, so #2406 is written against main's tree and measured there (7,566 + 9,606 keys, 0 disagreements; 704 tests) rather than cherry-picked from this branch.

`dev` has the identical bug and gets it via the main→dev back-merge, which will likely need a manual resolution on these two files. Branch kept in case we'd rather have a separate dev PR.

PR #2406, opened 2026-08-26
hotfix(sanitizer): the key test was still quadratic — one call below the #1670 fix (#2398)

dolho:
> **Targeting `main`, deliberately.** The burn is on client instances, which run `main`. Both sanitizer copies on `main` carry the vulnerable code, and the agent copy's `_is_sensitive_kv_key` sits at **line 150 there — exactly the line in every py-spy dump**. The two files differ between `main` and `dev`, so this is written against `main`'s tree rather than cherry-picked; everything below is measured on `main`.

## Summary

**This is my regression.** #1670 fixed the catastrophic backtracking in the credential sanitizer's *line* scan (#1661) and left the identical shape one call deeper, in `_is_sensitive_kv_key`. Every one of the 14 py-spy dumps is on that stack.

```
<genexpr> (credential_sanitizer.py)      <- the any(...) in this function
_is_sensitive_kv_key
_redact_kv_match  ->  sanitize_text  ->  sanitize_dict (recursive)
sanitize_subprocess_line  ->  read_stdout  (headless_executor)
```

## The cost

`re.search(r'(?:.*TOKEN.*)\Z', key)` retries the leading `.*` from every offset — O(n²) in key length. Measured on 3.13:

| key length | old | new | |
|---|---|---|---|
| 4 KB | 0.32 s | — | |
| 16 KB | 5.31 s | — | |
| 44 KB (backend, 13 patterns) | **40.9 s** | **0.00035 s** | 116,000× |
| 44 KB (agent copy, 30 patterns) | **44.4 s** | **0.00072 s** | 62,000× |

Forty seconds of CPU **for one key**, per match, per line, on a thread holding the GIL. That is the pegged core, and it is why the event loop stopped accepting.

## Why a quadratic test looked harmless

A wrong docstring. It described `key` as *"a short `KEY=` name"*. It is not — the key is whatever preceded an `=` in `_KV_LINE_RE`'s **unbounded** `([^\s"'=]+)`. And the stack above reaches it from **stream-json tool results** through recursive `sanitize_dict`, so multi-KB "keys" are the normal case, not an adversarial one. That single wrong assumption is what let the quadratic survive the #1670 review.

## The fix

Every pattern is `.*LITERAL.*` or `LITERAL.*`. Under `.search()` with a trailing `\Z`, **both mean exactly "the key contains this literal"** — so the test is substring containment: linear, C-level, no backtracking.

**Verdicts are unchanged, and that was the part to get right — a faster check that redacts *less* is a credential leak.** Pinned against the old implementation as an oracle: **17,172 keys across both copies (7,566 + 9,606), zero disagreements**, including the deliberate #1670 suffix semantics (`DB_` still redacts `MY_DB_PASSWORD`).

The literals are **derived** from the patterns and the derivation **refuses** anything that is not a plain literal — so a future pattern needing real regex fails loudly at import instead of being silently reduced to a check that redacts less. That is the guard #1670 should have had.

## Both copies — and the agent one is the one that matters

Invariant #5. The two files are vendored in *shape*, not byte-identically: the agent list is `SENSITIVE_VAR_PATTERNS` with 30 entries to the backend's 13. **Every py-spy dump was taken in the agent copy**, so a backend-only fix would have left all 14 burns exactly where they were. A test asserts the agent copy carries the fix.

## What this does and does not close

- **Closes** the burn: the hot path is now linear, so the arrival of a large tool result cannot peg a core.
- **Does not** address the zombie `schedule_executions` rows or the missing `HEALTHCHECK` on the agent image (no auto-recovery). Those are real and separate; the burn was making them visible.
- The mitigations in place (`max_parallel_tasks` 10→5, `execution_timeout` 3600→1800) cap blast radius and are still worth keeping until this ships to the fleet — **note the agent copy means a base-image rebuild is required for existing agents.**

## Test Plan

- [x] `tests/unit/test_2398_sanitizer_key_redos.py` — verdict equivalence against the old implementation, the #1670 suffix semantics, linearity at 64 KB, end-to-end `sanitize_text` on a 44 KB token, the derivation's refusal of a non-literal pattern, and the agent-copy parity assertion
- [x] Equivalence oracle **on `main`'s tree**: 7,566 backend + 9,606 agent keys, **0 disagreements**
- [x] `704 passed` across the sanitizer, #1661 and #1670 suites on `main`
- [ ] Base-image rebuild + fleet rollout (the agent copy is where the burn is)
- [ ] Watchdog confirmation over a full night with the fix deployed

Related to #2398 · follow-up to #1670 / #1661

## Getting it to `dev`

`dev` has the identical bug in files that have since diverged from `main`, so the standard main→dev back-merge is the route — and it will likely need a manual resolution on these two files. I have a dev-based version of the same change ready if you would rather have it as its own PR; say which you prefer and I will close the other.

### DIFF
--- a/docker/base-image/agent_server/utils/credential_sanitizer.py
+++ b/docker/base-image/agent_server/utils/credential_sanitizer.py
@@ -140,14 +140,68 @@ _KV_LINE_RE = re.compile(r'(?<![^\s"\'=])([^\s"\'=]+)=(["\']?)([^\s"\']+)\2')
 # composed regex could start matching mid-token, so `DB_.*` redacted
 # `MY_DB_PASS=x`. Anchoring with \Z reproduces exactly that (a `fullmatch`
 # rewrite would silently redact LESS — a leak, not a speedup).
-_SENSITIVE_KEY_SUFFIX_RE = [
-    re.compile(r'(?:' + p + r')\Z', re.IGNORECASE) for p in SENSITIVE_VAR_PATTERNS
-]
+#
+# --- #2398: the same quadratic shape, one call deeper -----------------------
+# #1670 made the LINE scan linear (#1661) and left THIS test quadratic:
+# `re.search(r'(?:.*TOKEN.*)\Z', key)` retries the leading `.*` from every
+# offset, so cost is O(len(key)^2). Measured on 3.13: 4 KB key = 0.32s,
+# 44 KB key = 40s — for ONE key, per match, per line, on a thread holding the
+# GIL. A py-spy watchdog caught it 14 times across three agents, every dump on
+# this stack:
+#
+#     <genexpr> (credential_sanitizer.py)      <- the `any(...)` below
+#     _is_sensitive_kv_key
+#     _redact_kv_match  ->  sanitize_text  ->  sanitize_dict (recursive)
+#     sanitize_subprocess_line  ->  read_stdout  (headless_executor)
+#
+# The old docstring called `key` "a short KEY= name". It is not: the key is
+# whatever preceded an `=` in `_KV_LINE_RE`'s `([^\s"\'=]+)`, which is
+# UNBOUNDED — and the stack above reaches here from stream-json tool RESULTS
+# via recursive `sanitize_dict`, so multi-KB "keys" are the normal case rather
+# than an adversarial one. That wrong assumption is what made a quadratic test
+# look harmless.
+#
+# Every pattern is `.*LITERAL.*` or `LITERAL.*`, and under `.search()` with a
+# trailing `\Z` both mean exactly "the key CONTAINS this literal". So the test
+# is substring containment: linear, C-level, no backtracking, and byte-for-byte
+# the same verdicts (pinned against the old implementation as an oracle over
+# thousands of keys in tests/unit/test_2398_sanitizer_key_redos.py).
+#
+# The literals are DERIVED from the patterns rather than restated, and the
+# derivation REFUSES anything that is not a plain literal. A future pattern
+# needing real regex semantics therefore fails loudly at import instead of
+# being silently reduced to a check that redacts less — the guard #1670 should
+# have carried.
+def _literal_from_pattern(pattern: str) -> str:
+    body = pattern
+    if body.startswith(".*"):
+        body = body[2:]
+    if body.endswith(".*"):
+        body = body[:-2]
+    if not body or re.escape(body) != body:
+        raise ValueError(
+            f"sensitive-key pattern {pattern!r} is not a plain `.*LITERAL.*` / "
+            f"`LITERAL.*` form. #2398 replaced the quadratic regex test with "
+            f"substring containment; a pattern needing real regex semantics "
+            f"must be handled explicitly rather than silently reduced to "
+            f"{body!r}."
+        )
+    return body.upper()
+
+
+_SENSITIVE_KEY_LITERALS = tuple(
+    _literal_from_pattern(p) for p in SENSITIVE_VAR_PATTERNS
+)
 
 
 def _is_sensitive_kv_key(key: str) -> bool:
-    """True if `key` (a short `KEY=` name) names a credential."""
-    return any(r.search(key) for r in _SENSITIVE_KEY_SUFFIX_RE)
+    """True if `key` names a credential.
+
+    `key` is whatever preceded an `=` in the scanned text — NOT necessarily a
+    short env-var name (#2398). Containment, not regex: see the note above.
+    """
+    upper = key.upper()
+    return any(lit in upper for lit in _SENSITIVE_KEY_LITERALS)
 
 
 def _redact_kv_match(match: "re.Match") -> str:
--- a/docker/base-image/agent_server/utils/credential_sanitizer.py
+++ b/docker/base-image/agent_server/utils/credential_sanitizer.py
@@ -181,8 +181,25 @@ def _literal_from_pattern(pattern: str) -> str:
     body = pattern
     if body.startswith(".*"):
         body = body[2:]
-    if body.endswith(".*"):
-        body = body[:-2]
+    # The TRAILING `.*` is what makes containment equivalent, so its absence is
+    # refused rather than tolerated (self-review, second pass). Under
+    # `.search()` the old form was `(?:BODY)\Z` — anchored at the END — so a
+    # pattern with no trailing `.*` meant "the key ENDS WITH this", and
+    # containment is wider: a bare `TOKEN` would stop matching only `MYTOKEN`
+    # and start matching `TOKEN_SUFFIX` too. That direction redacts MORE, so it
+    # is not a leak — but this function exists to refuse SILENT changes of
+    # meaning, and accepting one would be the same internal inconsistency that
+    # produced #2398: a stated contract that no longer described the code
+    # beneath it.
+    if not body.endswith(".*"):
+        raise ValueError(
+            f"sensitive-key pattern {pattern!r} has no trailing `.*`, so it "
+            f"meant 'the key ENDS WITH {body!r}'. #2398's containment test is "
+            f"WIDER than that — it would also match keys carrying {body!r} in "
+            f"the middle. Write it as `{body}.*` if containment is what you "
+            f"want, or handle the suffix case explicitly."
+        )
+    body = body[:-2]
     # An explicit safe set, NOT `re.escape(body) != body`. That was the first
     # form and it is wrong in the dangerous direction for a module imported
     # everywhere: `re.escape` also escapes `-`, which is not a metacharacter
--- a/docker/base-image/agent_server/utils/credential_sanitizer.py
+++ b/docker/base-image/agent_server/utils/credential_sanitizer.py
@@ -172,13 +172,24 @@ _KV_LINE_RE = re.compile(r'(?<![^\s"\'=])([^\s"\'=]+)=(["\']?)([^\s"\']+)\2')
 # needing real regex semantics therefore fails loudly at import instead of
 # being silently reduced to a check that redacts less — the guard #1670 should
 # have carried.
+# Characters that carry no regex meaning, so a body made only of these is a
+# literal and containment is exactly equivalent to the old anchored search.
+_PLAIN_LITERAL_RE = re.compile(r'[A-Za-z0-9_-]+')
+
+
 def _literal_from_pattern(pattern: str) -> str:
     body = pattern
     if body.startswith(".*"):
         body = body[2:]
     if body.endswith(".*"):
         body = body[:-2]
-    if not body or re.escape(body) != body:
+    # An explicit safe set, NOT `re.escape(body) != body`. That was the first
+    # form and it is wrong in the dangerous direction for a module imported
+    # everywhere: `re.escape` also escapes `-`, which is not a metacharacter
+    # outside a character class — so adding a perfectly literal `.*API-KEY.*`
+    # would have raised at import and taken the whole process down at boot.
+    # `.` IS special, so `A.B` is still (correctly) refused.
+    if not body or not _PLAIN_LITERAL_RE.fullmatch(body):
         raise ValueError(
             f"sensitive-key pattern {pattern!r} is not a plain `.*LITERAL.*` / "
             f"`LITERAL.*` form. #2398 replaced the quadratic regex test with "
--- a/src/backend/utils/credential_sanitizer.py
+++ b/src/backend/utils/credential_sanitizer.py
@@ -81,14 +81,68 @@ _KV_LINE_RE = re.compile(r'(?<![^\s"\'=])([^\s"\'=]+)=(["\']?)([^\s"\']+)\2')
 # composed regex could start matching mid-token, so `DB_.*` redacted
 # `MY_DB_PASS=x`. Anchoring with \Z reproduces exactly that (a `fullmatch`
 # rewrite would silently redact LESS — a leak, not a speedup).
-_SENSITIVE_KEY_SUFFIX_RE = [
-    re.compile(r'(?:' + p + r')\Z', re.IGNORECASE) for p in SENSITIVE_KEY_PATTERNS
-]
+#
+# --- #2398: the same quadratic shape, one call deeper -----------------------
+# #1670 made the LINE scan linear (#1661) and left THIS test quadratic:
+# `re.search(r'(?:.*TOKEN.*)\Z', key)` retries the leading `.*` from every
+# offset, so cost is O(len(key)^2). Measured on 3.13: 4 KB key = 0.32s,
+# 44 KB key = 40s — for ONE key, per match, per line, on a thread holding the
+# GIL. A py-spy watchdog caught it 14 times across three agents, every dump on
+# this stack:
+#
+#     <genexpr> (credential_sanitizer.py)      <- the `any(...)` below
+#     _is_sensitive_kv_key
+#     _redact_kv_match  ->  sanitize_text  ->  sanitize_dict (recursive)
+#     sanitize_subprocess_line  ->  read_stdout  (headless_executor)
+#
+# The old docstring called `key` "a short KEY= name". It is not: the key is
+# whatever preceded an `=` in `_KV_LINE_RE`'s `([^\s"\'=]+)`, which is
+# UNBOUNDED — and the stack above reaches here from stream-json tool RESULTS
+# via recursive `sanitize_dict`, so multi-KB "keys" are the normal case rather
+# than an adversarial one. That wrong assumption is what made a quadratic test
+# look harmless.
+#
+# Every pattern is `.*LITERAL.*` or `LITERAL.*`, and under `.search()` with a
+# trailing `\Z` both mean exactly "the key CONTAINS this literal". So the test
+# is substring containment: linear, C-level, no backtracking, and byte-for-byte
+# the same verdicts (pinned against the old implementation as an oracle over
+# thousands of keys in tests/unit/test_2398_sanitizer_key_redos.py).
+#
+# The literals are DERIVED from the patterns rather than restated, and the
+# derivation REFUSES anything that is not a plain literal. A future pattern
+# needing real regex semantics therefore fails loudly at import instead of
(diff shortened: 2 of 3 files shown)


dolho:
## Findings from building this against `main`

Notes that came out of the work rather than out of the issue — recording them here because two of them change what should happen next.

### 1. The dumps came from `main`, and the line numbers prove it

`_is_sensitive_kv_key` sits at **line 150 of `agent_server/utils/credential_sanitizer.py` on `main`**, and `<genexpr>` at 150 with `_redact_kv_match` at 161 is exactly what all 14 py-spy stacks show. That is a stronger confirmation than "the code looks like the bug": the *offsets* match the tree the clients run.

It also means the fix could not be a cherry-pick — **the two sanitizer files have diverged between `main` and `dev`** (different constant names, and 30 patterns vs 13). Everything below is measured on `main`'s tree.

### 2. The two copies are NOT byte-identical vendoring

Invariant #5 describes `credential_sanitizer.py` as a vendored mirror. In practice the agent copy uses `SENSITIVE_VAR_PATTERNS` with **30** entries; the backend uses `SENSITIVE_KEY_PATTERNS` with **13**. What is shared is the *policy shape*, not the file.

That matters here beyond tidiness: I initially patched the agent copy by splicing the backend's block into it, and it failed at import with `NameError: SENSITIVE_KEY_PATTERNS`. A fix written once and pasted twice would have shipped a broken agent image. The derivation in this PR is written against each file's own constant, and a test asserts the agent copy carries it — because **every dump was taken in that copy**, so a backend-only fix would have left all 14 burns untouched.

### 3. The measurements, on `main`

| | old | new | |
|---|---|---|---|
| backend copy (13 patterns), 44 KB key | **40.9 s** | 0.00035 s | 116,000× |
| agent copy (30 patterns), 44 KB key | **44.4 s** | 0.00072 s | 62,000× |

Growth is textbook quadratic on the way up — 500 ch: 0.003 s · 1 K: 0.013 s · 2 K: 0.063 s · 4 K: 0.32 s. Forty seconds for **one key**, per match, per line, on a thread holding the GIL.

### 4. Why my #1670 review missed it — a docstring, not the regex

The function said `key` was *"a short `KEY=` name"*. It is whatever preceded an `=` in `_KV_LINE_RE`'s **unbounded** `([^\s"'=]+)`, and the dumps show it arriving from **stream-json tool results** through recursive `sanitize_dict`. So multi-KB "keys" are the normal case, not an adversarial one. I measured the cost of the wrong input in #1670 and concluded the key test was cheap.

The guard added here is the one that change should have carried: the literals are **derived** from the patterns, and the derivation **refuses** anything that is not a plain literal, so a future pattern needing real regex fails at import rather than being silently reduced to a check that redacts less.

### 5. Equivalence was the expensive half, and worth it

A faster check that redacts *less* is a credential leak, so the old implementation is used as an oracle: **17,172 keys across both copies (7,566 + 9,606), zero disagreements** — random keys plus targeted cases built from every pattern fragment, and the deliberate #1670 suffix semantics (`DB_` still redacts `MY_DB_PASSWORD`). `704 passed` across the sanitizer, #1661 and #1670 suites on `main`.

---

## What this does not close

Stated because the issue is right to separate them, and because **merging this will make them less visible, not fixed**:

- **Zombie `schedule_executions` rows** — 3 rows `running` with `claim_token=NULL` and no worker.
- **No `HEALTHCHECK` on the agent image**, so Docker will not restart a wedged container; the DORMANT breaker files an operator-queue alert that is silent until a human acknowledges it. The burn is what made both observable.

## Next steps

1. **A base-image rebuild is required** before the fleet sees any of this — the burn is in the agent copy. Keep `max_parallel_tasks 10→5` and `execution_timeout 3600→1800` until the rebuilt image is out.
2. **`dev` has the identical bug** and gets it via the main→dev back-merge, which will likely need a manual resolution on these two diverged files. A dev-based branch already exists (`fix/2398-sanitizer-key-redos`, PR #2405 closed as superseded) if a separate PR is preferred.
3. **Keep the py-spy watchdog running** through the first night on the rebuilt image — 14 dumps on one stack is what turned four plausible hypotheses into one answer, and it is the only thing that will confirm the burn is actually gone rather than moved.

dolho:
## /review Report — self-review

**Branch**: `hotfix/2398-sanitizer-key-redos` → `main` @ `1f4a6d0`
**Files Changed**: 3 (+250/−10 at review start)
**Scope**: CLEAN — two sanitizer copies and one test file; nothing else touched on a hotfix branch
**Plan**: the issue's fix is delivered; the two residuals it names (zombie rows, missing `HEALTHCHECK`) are explicitly *not* in scope and are stated in the PR body

---

### Critical Findings — 1 found, fixed in `1f4a6d0`

**[C1] The guard I added to prevent a silent failure could crash boot instead** (Confidence: 9/10)

`_literal_from_pattern` decided "is this a plain literal" with:

```python
if not body or re.escape(body) != body:
    raise ValueError(...)
```

`re.escape` escapes `-`, which is **not** a metacharacter outside a character class. So a future entry like `.*API-KEY.*` — or `X-AUTH-TOKEN.*`, a header-shaped name, which is exactly the sort of thing that gets added to this list — would raise at **import**. `credential_sanitizer` is imported by the backend and the agent server, so that is not a failed check: it is the whole platform failing to start.

Measured rather than reasoned about:

```
API_KEY      escape='API_KEY'      changed=False
FOO-BAR      escape='FOO\-BAR'     changed=True     <- false rejection
A.B          escape='A\.B'         changed=True     <- correct rejection
```

Now an explicit safe set, `[A-Za-z0-9_-]+`, which accepts every current pattern and any plausible env-var/header name while still refusing real metacharacters — `A.B` and `TOKEN+` stay rejected, pinned by tests.

Worth naming the shape: **a guard added to stop a silent failure, which itself fails in a louder-than-intended way.** It was fail-safe in direction (nothing starts, rather than credentials leaking) but the blast radius was wrong.

---

### Informational

**[I1] Dead code, and not the harmless kind** (Confidence: 8/10) — fixed in `1f4a6d0`

`_sensitive_key_re` in the backend copy compiled the very quadratic regexes this change stopped using, and nothing referenced it any more (grepped `src/`, `tests/`, `docker/` — zero external references).

Anywhere else that is a style note. Here it is precisely the artifact that let this bug survive #1670: a module-level name that *looks like* the live check, sitting next to the one that is. Removed, with a test asserting neither it nor `_SENSITIVE_KEY_SUFFIX_RE` returns.

---

### Checked and found clean — the ones that mattered

**Incomplete-fix completeness (4.14) — is there a THIRD instance?** This is the second recurrence of one shape, so the question is the important one. The agent copy has two further uses of the pattern list (`_sensitive_var_re` in `get_credential_values`, lines 238 and 264). Both run `pattern.match(var_name)` against **environment variable names** — anchored at offset 0, and genuinely short. `.match` does not retry from every offset, so this is **not** the `.search`-every-offset shape. Verified by reading both call sites, not by assuming.

**Unicode equivalence.** The oracle is ASCII. `upper()` and `re.IGNORECASE` can disagree on some Unicode (ligatures, dotless i), so I checked the plausible cases directly — `toﬀken`, `ıTOKEN`, `tokeKn` — and old and new agree on all of them. No divergence observed; had there been one, the `upper()` direction redacts *more*, which is the safe side.

**Redaction semantics.** The #1670 suffix behaviour is preserved (`DB_` still redacts `MY_DB_PASSWORD`) and pinned. Equivalence re-run after the C1 fix: **7,566 backend + 9,606 agent keys, zero disagreements.**

**Both copies.** Vendored in shape, not byte-identically (30 patterns vs 13, different constant names). Every py-spy dump was in the agent copy; a test asserts it carries the fix.

**Performance, after the change:** backend 44 KB key **40.3 s → 0.00035 s**; agent **43.4 s → 0.00067 s**.

**Scope discipline.** Nothing beyond the three files. On a hotfix to `main` that matters more than usual.

---

### Summary

- **Critical**: 1 found, 1 fixed — 0 outstanding
- **Informational**: 1 found, 1 fixed
- **Tests**: 706 passed; equivalence 17,172 keys / 0 disagreements
- **Still required before this helps anyone**: the base-image rebuild (the burn is in the agent copy), and the main→dev back-merge on two diverged files.

The self-review earned its keep here: the C1 guard bug was introduced *by this PR*, in the code whose whole purpose is to stop the next person making the mistake I made in #1670.

dolho:
## /review — pass 2

**Branch**: `hotfix/2398-sanitizer-key-redos` → `main` @ `4c1e2b9` · 3 files, +310/−11 · **708 passed**

Pass 1's two findings are fixed and verified. This pass found **two more**, both in code this PR introduced, and one of them is the third consecutive finding inside the same twelve-line guard.

---

### [C2] The stray submodule bump — third time this session, so the cause is mine and mechanical

The diff carried `src/backend/enterprise | 2 +-`, moving `main`'s pointer from `2a5def35` to `e2ef835c` — an unrelated enterprise change riding a security hotfix onto `main`, invisible in a file list as a one-line gitlink.

Cause is not carelessness in the abstract, it is one command: `git add -u src/backend` stages the gitlink along with everything else. I have a learnings entry telling me to check for exactly this, and it still happened, because the entry says "check" and the fix is "don't use that command". Stripped with `git update-index --cacheinfo`; the pointer now matches `main` exactly, and this and every later commit stage **explicit paths only**.

### [C3] The guard validated the LITERAL but not the FORM

`_literal_from_pattern` checked that the body is a plain literal and never that the pattern kept its **trailing `.*`** — which is precisely what makes containment equivalent. Under `.search()` the old form was `(?:BODY)\Z`, anchored at the END, so a pattern without it meant *"the key ENDS WITH this"*. Containment is wider:

```
pattern "TOKEN"   key TOKEN_SUFFIX    old(suffix)=False   new(contains)=True
```

No shipped pattern is affected — all 13 backend and 30 agent patterns end in `.*`, now asserted — and the drift redacts *more*, so it was never a leak. Refused anyway, because refusing **silent changes of meaning** is this function's entire job. Tolerating one inside it would be the same internal inconsistency that produced #2398 in the first place: a stated contract that no longer describes the code beneath it.

---

### Checked clean this pass

- **The enterprise submodule does not reference what I deleted.** Pass 1 grepped `src/`, `tests/`, `docker/` — which does **not** cover `src/backend/enterprise`, a separate tree that imports from the OSS one. Re-grepped it properly: no references to `credential_sanitizer` internals, so removing `_sensitive_key_re` is safe there too. That was a real gap in pass 1's evidence, not just an omission from the report.
- **Equivalence re-verified after every change**: 7,566 backend + 9,606 agent keys, **0 disagreements**.
- **Performance holds**: 44 KB key — backend 40.4 s → 0.00034 s, agent 44.3 s → 0.00066 s.
- **Scope**: three files, all `#2398`. Nothing else on a hotfix to `main`.

---

### Summary

| pass | found | fixed |
|---|---|---|
| 1 | C1 (guard crashes boot on a hyphen), I1 (dead quadratic regexes left compiled) | both |
| 2 | C2 (stray submodule bump), C3 (guard accepts a silent semantic change) | both |
| | **outstanding** | **0** |

**The honest read on this PR:** the sanitizer fix itself has been right since the first commit and is well covered — 17,172-key equivalence oracle, both copies, measured 60,000–117,000×. **Every finding across both passes has been in the scaffolding I added around it**, and three of the four were in the twelve-line guard whose purpose is to stop the next person repeating my #1670 mistake. A guard is code, it gets the same scrutiny as the fix, and on this evidence it deserved more of it than the fix did.

Unchanged and still required before any of this helps a client: the **base-image rebuild** (the burn is in the agent copy) and the **main→dev back-merge** across two diverged files.

dolho:
## After this merges — what the operator of an affected instance has to do

Written for whoever runs the burning instance. **Deploying the backend alone changes nothing for them**, so the order below matters more than usual.

### The trap

Every py-spy dump was in `agent_server/utils/credential_sanitizer.py` — the **agent base image**, not the backend. A deploy that skips the rebuild leaves the burn exactly where it is.

And a rebuilt image is **not adopted by a restart**. On `main` the image-drift check sits behind `not was_already_running` (`lifecycle.py`), so it is only evaluated on a **cold** start. Each agent must be **stopped, then started**. A `docker restart`, or a `start` on an already-running agent, adopts nothing.

*(Checked while writing this: the stop→start itself is safe on `main` — #2186's `require_running=False` is present. It reached `main` squashed into the v0.9.0 release under a different SHA, so an ancestry check on `35a24d0c` reports it missing. Read the code, not the SHA.)*

### Steps

1. **Deploy the backend** as normal. Its copy sanitizes execution logs before the DB write, so it burns backend CPU on the same input — worth having, just not the fix that matters here.

2. **Rebuild the agent base image** — `./scripts/deploy/build-base-image.sh`. **Nothing below matters without this.**

3. **Stop → start every agent, one at a time.** The cold start is what triggers `image_drift` → recreate; look for `recreate_reason: "image_drift"` in the start response.
   - **`trinity-system` is skipped while running** — #1816 returns `recreate_deferred: "system_agent_running"` rather than replacing the orchestrator mid-operation. It needs its own deliberate stop→start.
   - A stop→start kills any in-flight execution on that agent, so sequence this in a quiet window rather than agent-by-agent at peak.

4. **Clear the wreckage the burn left**, none of which this PR touches:
   - the zombie `schedule_executions` rows (`status='running'`, `claim_token=NULL`, no worker in the container)
   - the DORMANT breaker: `POST /api/agents/{name}/circuit-breaker/reset` (admin)

5. **Confirm the new code is actually in the container** before trusting it:
   ```bash
   docker exec <agent> grep -c _SENSITIVE_KEY_LITERALS \
       /app/agent_server/utils/credential_sanitizer.py
   ```
   Non-zero means the fix is live. This is a stronger check than the image tag, which can be stale after a partial rebuild.

6. **Keep the mitigations** (`max_parallel_tasks 10→5`, `execution_timeout 3600→1800`) through at least one full night with the py-spy watchdog running, then revert. **The watchdog is the only thing that distinguishes "gone" from "moved"** — 14 dumps on one stack is what turned four hypotheses into one answer, and it is the same instrument that should confirm the answer was right.

### What this does NOT fix

Stated plainly because merging it makes both **less visible, not resolved**:

- **Zombie `schedule_executions` rows** — the burn produced them; nothing here reconciles them.
- **No `HEALTHCHECK` on the agent image** — so if a wedge recurs from any other cause, Docker still will not restart the container, and the DORMANT breaker still files an operator-queue alert that stays silent until a human acknowledges it. There is no auto-recovery either way.

Both deserve their own issues; they are not this PR's scope, and they were the reason a CPU burn took 48 hours and a human poking around to notice.

webmixgamer (approved):
## Approved — verification notes

Reviewed against `main` (`0cb3a4e2`, the branch's exact base). Everything below was re-derived independently rather than taken from the PR body.

### Verified

- **Verdicts unchanged.** Old implementation as oracle over ~27.5k keys across both copies: the only divergences are newline-bearing keys, which now redact *more* — and they are unreachable, since `_KV_LINE_RE`'s `[^\s"'=]+` cannot capture a newline. End-to-end `sanitize_text` differential (old vs new) over 1,510 generated lines × both copies: **0 output differences**. Same result on Python 3.9, 3.12; the PR's own oracle test ran green on 3.13 in CI.
- **Quadratic reproduced, linear confirmed.** 1 KB → 4 KB → 16 KB: 0.022s → 0.36s → 5.8s (×16 per ×4). New path: 8.5 → 85 µs.
- **Import-safe.** All 13 backend + 30 agent patterns pass the derivation; both modules import clean. No third `.search`-per-offset instance — the remaining `_sensitive_var_re` uses are anchored `.match()` on env-var names.
- **Nothing dangling.** Zero references to the removed names at the PR head; `main`'s enterprise submodule pin does not import the sanitizer at all.

### Back-merge to `dev` — easier than the body says

Trial-merged this branch into `origin/dev`: **both sanitizer files auto-merge cleanly**, `dev` keeps its #2208 `sk-` consolidation and gains the fix, and the full sanitizer/credential suite (498 tests) passes on the merged tree. The only conflict is the routine enterprise gitlink. The standard `main → dev` back-merge is sufficient; no separate dev PR needed.

### CI: the cancelled `pytest (head, seed 99999)` shard is not a failure

It hit the 25-minute job ceiling at 54% while still progressing — no test failed, no per-test timeout fired, no JUnit was produced. Timing profile vs the three healthy shards (base and head alike: ~740s, 2 stalls >30s, max ~122s — the pre-existing 60s / 40s waits in `test_start_agent_skip_inject` and `test_1083_result_callback::TestPersistResend`, identical on both sides): the cancelled shard hit **14 stalls >30s starting at 1%**, before the suite had meaningfully run. The 25 new tests total 0.25s in CI. That is an order-sensitive stall in pre-existing tests on paths this diff does not touch, surfaced by the shuffle over a 25-item-larger collection. `regression diff` handled it as designed ("Degraded — proceeding on surviving seeds"): **no new failures**. Locally, head and main at seeds 12345 and 99999 are all clean, delta exactly the 25 new tests. Worth a `gh run rerun --failed` for a clean record.

### Before merging (process, not code)

1. **Issue disposition.** Merging to `main` runs no status automation (`issue-status-on-merge.yml` is `dev`-only), and the body says `Related to #2398`, not `Fixes`. As written, #2398 stays open and `status-in-progress` with no record the fix shipped. Either `Fixes #2398` + file the two residuals (zombie `running` rows; missing agent `HEALTHCHECK`) as their own issues, or keep it open deliberately and set the label by hand at the tag.
2. **Priority.** P-07 hotfix-to-`main` is scoped to P0; the issue is labelled P1. Relabel, or note the exception so this isn't read later as a process bypass.
3. **Rollout is what actually fixes the fleet.** The burn is in the agent copy, baked into the base image; nothing rebuilds it on merge. Tag → base-image rebuild → deploy → *recreate* agents (cold start adopts the new image; `POST /api/ops/fleet/restart` covers a running fleet). Keep the `max_parallel_tasks` / `execution_timeout` mitigations until that lands, and keep the py-spy watchdog running through the first night.

### Minor

- Squash title carries `hotfix(` while the commits use `fix(`; conventional prefix is `fix(`.
- The agent-copy parity test is a source grep, not an import — it would not catch the `NameError` described in the write-up. It is covered today only because `test_1661` `exec_module`s the copy; worth strengthening on the `dev` side.

Issue #2338, opened 2026-08-20
feat(tests): journey catalog (intent only) + generated JOURNEYS.md


vybe:
**Rail R4** of the journey-coverage deck.

## Summary

`tests/journeys/catalog.yaml` — the durable record of what the platform promises — plus a generated `JOURNEYS.md`, with run status read from CI artifacts and **never committed**.

## Context

An epic is a delivery vehicle that closes; the promise list must outlive it. The catalog is that durable home, and it is what makes coverage debt **queryable rather than discovered in an exploratory sweep**.

Two hard constraints, both learned from an independent critique of the proposal:

1. **Status never lives in git.** `last_green` / `green|red` in a committed YAML either rot (humans forget to update) or spam history and cause merge conflicts (bots do update). Git holds intent; CI holds state.
2. **One invariant namespace** — records reference IDs into the invariant catalog (R3), never restate them.

`built: false` is legal and visible. That is the point.

## Acceptance Criteria

- [ ] `tests/journeys/catalog.yaml` exists with all 10 journeys declared and `built: false` where true
- [ ] Each record carries: id, promise in the actor's words, lanes, invariant ID references, harness path, tier, owner
- [ ] **Zero status fields are committed** — no `last_green`, no `green|red`; a committed status field fails validation
- [ ] Coverage % (journeys built and green ÷ declared) is computed at report time from CI artifacts
- [ ] `JOURNEYS.md` is generated from the catalog, not hand-written and not from docstrings
- [ ] The catalog links to `tests/registry.json` rather than merging with it — registry indexes tests, catalog indexes promises
- [ ] Every referenced invariant ID resolves in `docs/testing/orchestration-invariant-catalog.md`

## Technical Notes

Depends on R3 for the invariant IDs. This is the artifact the PM layer reads; harnesses are Eng's.

---
*Journey-coverage process, filed 2026-08-20 from the operator session of 08-19/20. Epic #1850 (rechartered). **R1–R5 ship before any journey issue moves to `status-in-progress`.** Journey records reference invariant IDs in `docs/testing/orchestration-invariant-catalog.md` — never restate them.*

vybe:
[...]

Issue #2349, opened 2026-08-20
J10: Agent calls agent — journey harness


vybe:
## Summary

**Promise:** Playbook call → `chat_with_agent` / A2A / `fan_out` → the result returns, the permission boundary is enforced, and loop and budget guards hold.

## Context

Protocol-level units exist; **no two-agent round trip is asserted anywhere**. Inter-agent work is the platform's differentiator and its least-tested promise.

Filed `status-ready` because it is cheap against a prebuilt stack and it guards the boundary — a permission regression here is silent and severe.

## Acceptance Criteria

- [ ] A two-agent round trip returns a real result to the caller
- [ ] A call the permission model forbids is **denied**, and the denial is observable
- [ ] Loop guards stop runaway inter-agent recursion
- [ ] Budget guards bound the total work of a fan-out
- [ ] A dead or stopped callee fails fast with an honest reason rather than hanging to timeout
- [ ] Harness is named for the journey (`test_j10_*.py`), never for a ticket
- [ ] Every finding ships as a `strict=True` xfail plus its own issue — no silent product fixes inside the test task
- [ ] A record exists in `tests/journeys/catalog.yaml` referencing invariant IDs

## Technical Notes

Spans A2A, `fan_out`, permissions, loop service.

---
*Journey-coverage process, filed 2026-08-20 from the operator session of 08-19/20. Epic #1850 (rechartered). **R1–R5 ship before any journey issue moves to `status-in-progress`.** Journey records reference invariant IDs in `docs/testing/orchestration-invariant-catalog.md` — never restate them.*

egoriklok:
@vybe — #2349 defines exactly the boundary my Fiscal fixed-scope audit covers: one real A2A/fan-out journey across denied permissions, aggregate budget, loop halt, and dead-callee recovery.

I can independently audit one Trinity workflow for 99 Base USDC, delivered in 72h. If useful, reply with the exact workflow and payer entity/approver: https://egoriklok.github.io/agent-fiscal-autonomy-pack/

Issue #2340, opened 2026-08-20
J01: Cold install to first login — journey harness


vybe:
## Summary

**Promise:** Fresh host → `install.sh` / `quickstart.sh --defaults` → stack healthy → admin logs in → dashboard renders.

## Context

**Coverage today: none.** `install.sh` (264 lines) and `quickstart.sh` (175) have no automated test; `tests/deploy/` holds a single docker-GID shell script. This is the first thing every OSS adopter touches.

**This journey carries the hosted-provider lanes.** Per the 2026-08-20 lane ruling: **DigitalOcean is a supported lane — it must be green to cut a release**; Vultr and Hostinger are best-effort (tracked, non-blocking). If a provider's lane cannot satisfy the shared invariants, *that is the finding*, and it becomes a product decision about what "supported" means — surfaced early rather than at launch.

## Acceptance Criteria

- [ ] Lanes: bare-host, compose, digitalocean-1click (supported) · vultr, hostinger (best-effort)
- [ ] Start state is a genuinely fresh host; the harness provisions and destroys it
- [ ] Done state is asserted from outside the box: an admin session authenticates and the dashboard renders
- [ ] Migrations run for real (`alembic upgrade head`), not mocked
- [ ] A failing lane names which invariant broke and on which substrate
- [ ] Best-effort lane failures are reported but do not fail the tier
- [ ] Harness is named for the journey (`test_j01_*.py`), never for a ticket
- [ ] Every finding ships as a `strict=True` xfail plus its own issue — no silent product fixes inside the test task
- [ ] A record exists in `tests/journeys/catalog.yaml` referencing invariant IDs

## Technical Notes

Release-cut tier, not per-PR — it builds images. Spans installer, compose, migrations, auth.

---
*Journey-coverage process, filed 2026-08-20 from the operator session of 08-19/20. Epic #1850 (rechartered). **R1–R5 ship before any journey issue moves to `status-in-progress`.** Journey records reference invariant IDs in `docs/testing/orchestration-invariant-catalog.md` — never restate them.*

Issue #2337, opened 2026-08-20
refactor(tests): extend the orchestration invariant catalog to journey-level invariants


vybe:
**Rail R3** of the journey-coverage deck.

## Summary

Extend `docs/testing/orchestration-invariant-catalog.md` to cover journey-level invariants so there is exactly **one** invariant namespace.

## Context

The catalog already has the right method — invariants first, Tier A (always) / Tier B (eventually ≤ T), severity-ranked, each a runnable SQL/Redis/HTTP predicate — and `src/backend/canary/` implements Phase 1 in code.

The failure mode this prevents: journey records inventing free-string invariants (`invariants: ["stack healthy"]`) alongside the real catalog. **A second list of invariants is how the first one dies.** Journey records reference IDs; they never restate.

## Acceptance Criteria

- [ ] The catalog gains journey-level invariants with the same tiering and severity scheme as the existing entries
- [ ] Every invariant is a runnable predicate (SQL / Redis / HTTP / file), not prose
- [ ] Journey catalog records reference invariant IDs only — a free-string invariant fails validation
- [ ] Existing orchestration invariant IDs are unchanged; this extends, it does not renumber
- [ ] The relationship to `src/backend/canary/` is documented: which invariants the canary evaluates live vs which are journey-only

## Technical Notes

Cheap but ordering-critical: land it before the catalog (R4) so records have IDs to point at.

---
*Journey-coverage process, filed 2026-08-20 from the operator session of 08-19/20. Epic #1850 (rechartered). **R1–R5 ship before any journey issue moves to `status-in-progress`.** Journey records reference invariant IDs in `docs/testing/orchestration-invariant-catalog.md` — never restate them.*

webmixgamer:
## Implementation note — what the plan review found, and one AC relaxed by ruling

**The defect was resolution, not authoring.** The catalog defined **71** invariants as `**X-NN**` body entries, but `tests/unit/test_2338_journey_catalog.py` resolved journey `invariants:` from **markdown table rows only** — and the only table was the 12-row *Recommended starting subset* — so a journey record could cite **11 of 71**. Citing `P-02` (§7, precisely J10's permission promise) failed the validator as "coverage theatre". The canary evaluates **16** live. PR #2407's body ("the catalog references only the 11 ids that resolve today") shows the 11 was believed to be the universe, not a curated gate.

[...]

PR #2407, opened 2026-08-26
feat(tests): journey catalog (intent only) + generated JOURNEYS.md (#2338)

dolho:
## Summary

**Rail R4.** `tests/journeys/catalog.yaml` — the durable record of what Trinity promises, in the words of whoever it is promised to — plus a generated `docs/testing/JOURNEYS.md`.

All ten journeys (J01–J10, #2340–#2349) declared, all `built: false`. That is the point: the list exists so coverage debt is **queryable** rather than rediscovered in an exploratory sweep six months later.

## The two hard rules are enforced, not documented

**1. No status is committed.** A `last_green` in git has exactly two futures — a human owns it and it rots, or a bot owns it and every CI run writes a commit (history spam plus merge conflicts on a file everybody edits). Git holds **intent**; CI holds **state**. A status-shaped key at *any depth* fails the test.

**2. One invariant namespace.** Records **reference** ids in `docs/testing/orchestration-invariant-catalog.md` and never restate them, and every id is checked to resolve. A journey whose invariants don't exist yet carries an **empty list rather than an invented id** — the gap is the information, and #2337 (R3) is what fills it. An id that resolves nowhere reads as coverage that does not exist, which is worse than a blank.

Both are **mutation-checked**: injecting a `last_green` and an invented `Z-99` fails exactly the three tests that should fail, and restoring returns 11/11.

## Coverage is computed at report time

```bash
# committed form — intent only
python scripts/ci/generate_journeys_md.py --out docs/testing/JOURNEYS.md

# report form — joins CI artifacts, for a step summary
python scripts/ci/generate_journeys_md.py --junit <artifact-dir>
```

Verified against a **real downloaded artifact set** from a green `backend-unit-test` run, not a fixture. A harness with no artifact counts as **"no evidence"**, never as green — absence of evidence isn't a pass.

The catalog **links to** `tests/registry.json` rather than merging with it: the registry indexes *tests*, the catalog indexes *promises*. Merging would make every new test file a catalog edit and bury ten promises in a 240 KB index; the generator joins them at report time instead.

## The validator earned its keep on its first run

It caught the catalog naming J03's harness at `tests/journeys/test_first_turn_journey.py` — a path that only exists on the **unmerged #2403 branch**. Fixed by correcting the catalog, not by relaxing the check; the field gains the path in the commit that merges it. That is the same failure mode as an invented invariant id, one field over.

## On the R3 dependency — handled, not blocked

R4 declares a dependency on R3 for invariant ids. Rather than stall, the catalog references only the **11 ids that resolve today** and leaves the rest empty. So it is correct as of this commit and gains references as R3 lands — which is exactly the `built: false` philosophy applied one level down.

## One judgement recorded

The agent-lifecycle journey #2403 ships is **not** a journey of its own; it is J03's first half. Minting a "J11: an agent can be restarted" would turn the catalog from a list of promises into a list of steps — nobody asks for a restart, they ask for a useful answer. Noted in the file so the next person doesn't re-litigate it.

## Test Plan

- [x] `tests/unit/test_2338_journey_catalog.py` — **11 tests**: no committed status (any depth), J01–J10 present and ordered, required fields, promises phrased in the actor's voice, lanes defined, **every invariant id resolves**, harness paths exist in both built and unbuilt states, registry linked not merged, `JOURNEYS.md` not stale, committed doc carries no status
- [x] Both hard rules mutation-checked
- [x] Generator verified against real CI JUnit artifacts
- [x] No workflow change needed — the staleness check rides the unit suite, which already runs on every PR

Related to #2338

## Staging note — `--junit` is not wired anywhere yet (review finding 5)

Nothing in this PR, or in any workflow, calls the generator's report form. So
the coverage line and the whole `green_harnesses` path — including the two
findings fixed in it — are **unreachable in CI today**. That is deliberate
staging, not an oversight: the catalog and the staleness gate stand on their
own, and the report form activates when a workflow starts passing `--junit` at
a directory of downloaded artifacts. Said out loud so the next reader does not
go looking for a summary that does not exist yet.


AndriiPasternak31:

**Both passes: merge-ready / approve — 0 critical.** CI fully green; the drift guard follows the #2086 byte-parity convention and was verified working locally (`--check` → `PARITY OK`); the PR's mutation checks were re-run rather than trusted (injected `last_green` → exactly the status test fails; injected `Z-99` → exactly the invariant-resolution + staleness tests fail); all 11 referenced invariant ids resolve against the R3 catalog.

Small items, none blocking:

1. **Coverage counter counts "green" without requiring "built"** — `scripts/ci/generate_journeys_md.py:96-98` tests `j["harness"] in greens` while the line at `:122` reads "{green}/{total} journeys built and green". Once #2403 backfills J03's harness with `built: false` (the exact intermediate state the catalog documents), a report renders the self-contradiction "1/10 built and green. 0 declared built." One-line fix: `if j["built"] and j["harness"] in greens`. Inert today (no workflow invokes `--junit`).
2. **Merge-order coupling with #2403**: the J03 comment promises the harness field "gains the path in the commit that merges" #2403 — but #2403's file list doesn't touch `catalog.yaml`, and `harness: null` passes every validator, so forgetting the backfill fails nothing and J03's harness join stays silently inert. Whichever PR lands second should carry the one-line `harness: tests/journeys/test_first_turn_journey.py` **and regenerate JOURNEYS.md in the same commit** (else the staleness guard trips).
3. **No closing keyword** — "Related to #2338" won't auto-promote; either `Closes #2338` in the body or a manual label bump after merge.
4. (fyi) Once any `harness` is non-null, the committed doc's bytes also depend on `tests/registry.json` (`generate_journeys_md.py:132-137`) — a later registry refresh can flip the footer and turn the staleness unit test red on an unrelated PR. Consider excluding the footer from the committed (no-`--junit`) form.

obasilakis (changes requested):
Requesting changes on two findings, both in the enforcement layer rather than in the catalog itself.

The catalog is the right artefact and the two hard rules are the right two. What I am flagging is that the code enforcing them has a false-positive landmine, and the registry link the PR advertises does not actually work against the registry's real shape.

## Blocking

### 1. `tests/unit/test_2338_journey_catalog.py` — `_STATUS_KEY_RE` matches ordinary words

```python
_STATUS_KEY_RE = re.compile(
    r"(last_green|last_run|green|red|passing|failing|status|result|"
    r"coverage|pass_rate|last_seen|flaky)",
    re.IGNORECASE,
)
```

Unanchored alternation, applied with `.search()` to every key at every depth. `red` is a substring of **`credential`** and of **`required`**.

J06 is *"I can give an agent a credential and it can use the tools that credential unlocks."* The moment any record gains a `credentials:` key — or anyone adds a `required:` field to the schema — `test_no_status_field_is_committed` fails with a message accusing the author of committing run status. That is a confusing failure on a rule whose whole value is that it fails clearly, and it will land on someone who did nothing wrong.

`green` has the same shape (`evergreen`, `greenfield`), `result` matches `results`/`resultant`. Fix is word boundaries — `\b(...)\b` — or, better for a fixed vocabulary, match against an explicit key allowlist and drop the substring search entirely.

### 2. `scripts/ci/generate_journeys_md.py:harness_is_indexed` — the registry join cannot match

```python
files = data.get("test_files")
if isinstance(files, list):
    return any(harness in str(f) or Path(harness).name == Path(str(f)).name
               for f in files)
```

`tests/registry.json` has `test_files` as a **list of dicts**, not of strings:

```json
{"file": "unit/test_157_a2a_inbound_server.py", "feature": "ent#157", ...}
```

So `str(f)` is the repr of a dict and `Path(str(f)).name` is nonsense. The `harness in str(f)` arm cannot match either, because of the second half of the problem: registry paths are **`tests/`-relative** (`unit/test_x.py`) while catalog harnesses are **repo-relative** (`tests/journeys/test_first_turn_journey.py`).

Consequence: as soon as any journey names a harness, the generated doc will carry the footer *"Harnesses not present in `tests/registry.json`"* for every one of them, permanently and wrongly. AC #6 — link to the registry rather than merge with it — is the AC this implements, and the link is currently inert. It is invisible today only because all ten harnesses are `null`.

Fix is `f.get("file")` on the dict shape plus one normalisation of the `tests/` prefix.

## Non-blocking, same function family

### 3. `green_harnesses()` — the same prefix mismatch, and it is the one that was never exercised

JUnit `file` attributes are rootdir-relative, so they will read `journeys/test_x.py` while `render()` compares against the catalog's `tests/journeys/test_x.py`. Every journey would report `no evidence` forever.

Worth being explicit about why this survived verification: the PR says the generator was verified against a real downloaded artifact set from a green `backend-unit-test` run. That artifact set contains no journey harness at all, so the run exercised the *parse* path and never the *match* path. The verification was real; it just could not reach the code the finding is in.

### 4. `green_harnesses()` — an all-skipped harness reports green

```python
bad = any(case.find(t) is not None for t in ("failure", "error"))
seen[f] = seen.get(f, True) and not bad
```

`skipped` is neither, so a file whose every test skipped is recorded as green. J03's harness is precisely the credential-bound one that skips on every PR-triggered run — so the first journey to carry a path is also the first to be misreported. That contradicts the docstring directly above it ("Absence of evidence is not green"), which is right about the *missing-artifact* case and silent about this one.

### 5. Nothing wires `--junit`

No workflow in the diff calls the report form, so the coverage line and everything findings 3 and 4 touch is unreachable today. Fine as staging, but worth saying in the PR body so the next reader does not assume a summary exists somewhere.

## What I checked and found fine

- The no-status rule is genuinely enforced at every depth via `_walk_keys`, not just at the top level, and both hard rules are mutation-checked as claimed.
- `test_every_invariant_id_resolves` parses the invariant table by row anchor and asserts the parse itself is non-empty first, so a format change fails loudly instead of vacuously passing. That is the right shape and it is the check that would otherwise rot silently.
- The empty-list-rather-than-invented-id decision is correct and is the more useful half of the one-namespace rule. Same for keeping the J03 harness `null` while #2403 is unmerged — that is the validator doing its job on its own author, which is the best evidence it works.
- `--check` compares the full rendered body and writes nothing, so the staleness gate cannot mask a drift by fixing it.
- Not minting a J11 for "an agent can be restarted" is the right call, and recording the reasoning in the file so it is not re-litigated is worth more than the decision itself.

## Verdict

Requesting changes on 1 and 2. Finding 1 is a false positive waiting for J06's own vocabulary; finding 2 makes an advertised AC inert. Both are small, and neither implies anything about the catalog's content, which I have no objection to.

dolho:
## All five addressed — `689029c3`

### 1. `_STATUS_KEY_RE` matched ordinary words

Confirmed against the shipped regex before changing it — every one of these matched:

```
credential  True    required  True    evergreen   True
credentials True    hundred   True    greenfield  True    resultant True
```

`hundred` is the one that makes the point: nothing about the vocabulary was doing the work, `red` was just a substring.

Your suggested `\b(...)\b` turned out not to be enough on its own, and it fails in the more dangerous direction: `_` is a word character, so `\bstatus\b` does **not** match `journey_status` — the rule would have gone quiet on exactly the keys it exists to catch while looking fixed. So the key is split into tokens and matched against a fixed vocabulary, with an exact-match set for multi-token names whose parts are individually innocent (`last_run`, `pass_rate`). Both directions are pinned by parametrized cases now — the false positives you named, and compound spellings like `journey_status` / `ci_coverage`.

### 2. The registry join could not match

Fixed as you diagnosed: `test_files[].file` on the dict shape, plus one shared `_norm_test_path` that reconciles the three spellings (catalog repo-relative, registry tests-relative, JUnit rootdir-relative). Proven against the real `registry.json` rather than a fixture.

### 3 + 4. `green_harnesses`

Same normalisation fixes the prefix mismatch. All-skipped no longer counts as green — green now requires at least one real pass with no failure or error, so a mixed skip+pass file is still green and an all-skipped one is not. The docstring said "absence of evidence is not green"; it now covers both shapes of absence rather than one.

Behaviour, measured:

```
green set: ['journeys/test_first_turn_journey.py', 'journeys/test_mixed.py']
all-skipped green?   False      failing green?  False
mixed skip+pass?     True       repo-relative catalog match?  True
```

### Why these were invisible, and what stops it recurring

Your explanation was the useful part of the review: the generator was verified against a real artifact set that contains no journey harness, so the run exercised the parse path and never the match path. Verification that cannot reach the code is worth naming.

Three tests now drive the match path directly, including the registry join against a genuine `registry.json` entry. Each was mutation-checked rather than assumed — reverting `f.get("file")` and reverting the all-skipped rule each turn the corresponding test red.

### 5. `--junit` is unwired

Added to the PR body as an explicit staging note, so nobody goes looking for a summary that does not exist: nothing calls the report form, so the coverage line and everything findings 3 and 4 touch is unreachable in CI today. It activates when a workflow starts passing `--junit` at a directory of downloaded artifacts.

`docs/testing/JOURNEYS.md` is byte-identical after all of this (every harness is still `null`) and `--check` passes; 50 tests green.

dolho:

**What landed:** the status regex now matches by token (the shipped one flagged `credential`, `required` and even `hundred` — verified against the old pattern), the registry join reads `test_files[].file` through one shared path normaliser, and `green_harnesses` no longer calls an all-skipped harness green. Three tests now drive the match path your review identified as unreachable by the original verification.

obasilakis (approved):

I re-derived all four findings against the current branch source rather than reading the commit subjects. All four are fixed, and the first one is fixed better than the word-boundary patch I suggested.

### 1. The status-key rule false-firing on `credential` — fixed by tokenising

I suggested `\b(...)\b`. What landed is a token split plus an explicit multi-token key set:

```python
_STATUS_TOKENS = frozenset({"green", "red", "passing", "failing", "status",
                            "result", "results", "coverage", "flaky"})
_STATUS_KEYS = frozenset({"last_green", "last_run", "last_seen", "pass_rate", "passrate"})
```

with the comment noting that `\bstatus\b` would *not* have matched `journey_status` — so the word-boundary fix I proposed would have gone quiet on the exact keys the rule exists to catch, while fixing the false positive. That is a genuine correction to my finding, not just a compliance with it. `credential` and `required` are single tokens and no longer match; `journey_status` and `last_green` still do.

### 2. The registry join — fixed on both halves

`harness_is_indexed` now reads `f.get("file")` off the dict shape and compares through `_norm_test_path`, so the `tests/`-relative-vs-repo-relative mismatch is gone too. The dict-vs-list branch also tolerates a mapping shape. AC #6's link is live rather than inert, and the generated doc will stop claiming every harness is missing from the registry the moment one carries a path.

### 3. The JUnit prefix mismatch — fixed

`green_harnesses` normalises through the same `_norm_test_path`, so rootdir-relative JUnit `file` attributes match repo-relative catalog harnesses. This is the one that had never been exercised — the verification artifact set contained no journey harness, so the run reached the parse path and not the match path — and it is now correct by construction rather than by luck.

### 4. All-skipped reporting green — fixed, and the docstring now says both shapes

```python
skipped = case.find("skipped") is not None
seen[f] = seen.get(f, True) and not bad
if not bad and not skipped:
    passed.add(f)
return {f for f, ok in seen.items() if ok and f in passed}
```

Requiring at least one genuinely passing case is the right rule, and the docstring now states both shapes of "absence of evidence is not green" instead of the one. J03's credential-bound harness — the first journey to carry a path and the one that skips on every PR-triggered run — would have been the first misreported, so this mattered more than its size suggests.

---

The catalog itself I had no findings on and still don't. The two hard rules are the right two, and enforcing them rather than documenting them is what makes the file worth committing. Declaring all ten journeys `built: false` is the point rather than an embarrassment — coverage debt you can query beats coverage debt you rediscover in a sweep six months later. Carrying an empty `invariants` list rather than an invented id is the same discipline: an id that resolves nowhere reads as coverage that does not exist, which is worse than a blank.

Checklist: base `dev`, `#2338` resolves with correct labels, security scan clean, no packaging, schema or backend surface, generated `JOURNEYS.md` committed alongside its generator. CI green.

One process note, non-blocking: the PR references `#2338` without a closing keyword, so `issue-status-on-merge.yml` will not promote it to `status-in-dev` on merge. Rephrasing to `Closes #2338` before merge would save `/groom` a cleanup pass.

Issue #2350, opened 2026-08-20
feat(process): Journey Impact declaration on epic + PR templates, enforced at merge


vybe:
**Gate G1** of the journey-coverage deck.

## Summary

Every `type-epic` and user-facing `type-feature` declares **Journey Impact**: `new: <id>` / `extends: <id>` / `none: <reason>` — and declaring `new:` obliges the epic's first PR to include a journey skeleton.

## Context

Ownership alone collapses into "PM nags, Eng defers." **The merge gate is the load-bearing part** — a `new:` declaration that arrives without a skeleton blocks the epic's own first PR, so the PM does not chase; the gate does.

The anti-gaming design matters: an intake field alone would default to `none: <reason>` for everything. Moving the burden onto the `new:` path — where the skeleton is required — means `none:` stops being the cheap escape, because choosing it is a visible claim that the change touches no promise.

A `strict=True` xfail asserting the invariants is a sufficient skeleton.

## Acceptance Criteria

- [ ] Epic and PR templates carry the Journey Impact field
- [ ] A PR under an epic declaring `new:` cannot merge without a journey skeleton present
- [ ] A `strict=True` xfail asserting the journey's invariants satisfies the gate
- [ ] `none: <reason>` requires a reason string — a bare `none` fails
- [ ] The check reports which declaration it read and why it passed or failed

## Technical Notes

Depends on R1 (a skeleton needs a tier to live in). The workspace half — the field in the PM drafting and filing skills, and the grooming check for it — is tracked outside this tracker.

---
*Journey-coverage process, filed 2026-08-20 from the operator session of 08-19/20. Epic #1850 (rechartered). **R1–R5 ship before any journey issue moves to `status-in-progress`.** Journey records reference invariant IDs in `docs/testing/orchestration-invariant-catalog.md` — never restate them.*

dolho:
Gate G1 implemented in #2408.

Shape: the declaration parses out of the PR **or** any `type-epic` issue the PR references; the decision is a pure function (`scripts/ci/journey_impact.py`, 26 unit tests) rather than a workflow shell body, so the rule is testable instead of discovered in a CI log.

All five ACs land. Two things worth flagging here rather than only in the PR:

1. **The skeleton is bound to the declared id.** Without that, `new: J11` is satisfied by touching *any* file in the journey tier — including an unrelated edit to an existing journey — which hands the obligation back to the compliance theatre the issue explicitly warns about. The id must now appear in the skeleton's path or body.

[...]

PR #2408, opened 2026-08-26
feat(ci): Journey Impact declaration, enforced at merge (#2350)

dolho:
## Description

**Gate G1** of the journey-coverage deck. Every epic and user-facing feature declares what it does to the promise list, and the declaration is **enforced at merge**:

```
Journey Impact: new: J11
Journey Impact: extends: J03
Journey Impact: none: <why this touches no promise>
```

The declaration is cheap; **the gate is the load-bearing part**. Ownership alone collapses into "PM nags, Eng defers", and an intake field on its own is worse than nothing — everything is `none:` inside a month and the field reads as compliance. So the burden sits on the `new:` path: declaring a new promise obliges the same PR to carry a skeleton for it. That is what stops `none:` being the cheap escape — choosing it becomes a visible claim, in writing, next to a diff a reviewer can check it against.

## Related Issue

Related to #2350

## Journey Impact

Journey Impact: none: CI/process gate, no user-facing promise

## What fails, what passes

| Input | Verdict |
|---|---|
| `none` (bare) | **REJECTED** — AC 4 |
| `new: J11` with no skeleton naming J11 | **REJECTED** |
| epic referenced by this PR declares `new: J11`, PR carries nothing | **REJECTED** |
| `new: J11` + `strict=True` xfail naming J11 under `tests/journeys/` | passes — AC 3 |
| `extends: J03` | passes (that promise already has a harness) |
| `none: <reason>` | passes |
| **no declaration at all** | **passes, with a loud note** |

That last row is deliberate: adopting the field must not red-X every in-flight PR and dependabot bump on day one. Tightening it is a follow-up once the templates have been in use.

## Two design points

**The skeleton is bound to the declared id.** Without that, `new: J11` is satisfied by any edit anywhere in the journey tier — including an unrelated tweak to an existing journey — which hands the obligation straight back to the theatre the gate exists to prevent. The id must appear in the skeleton's path or body.

The match deliberately does **not** use `\b`. Ids are routinely embedded underscore-delimited (`test_j07_journey.py`) and `_` is a word character, so `\bJ07\b` never matches the very filename the naming convention produces — it would have rejected correctly-named skeletons. It delimits on letters and digits instead, so `J07` still rejects `J071` and `XJ07`.

**The decision is a pure function**, in `scripts/ci/journey_impact.py`, not a workflow shell body — a rule that only runs in CI is a rule nobody can test. 26 unit tests cover the parser and the decision; removing the id binding fails three of them (mutation-checked).

## Security

A PR body is attacker-authored text, so it reaches the checker through `env:` and never through `${{ }}` inside a `run:` — pinned by a test that greps every `run:` body in the workflow.

## Acceptance criteria

- [x] Epic and PR templates carry the Journey Impact field
- [x] A PR under an epic declaring `new:` cannot merge without a journey skeleton present
- [x] A `strict=True` xfail asserting the journey's invariants satisfies the gate
- [x] `none: <reason>` requires a reason string — a bare `none` fails
- [x] The check reports which declaration it read and why it passed or failed

## Verification

```
26 passed   tests/unit/test_2350_journey_impact.py
32 passed   + test_1891_python_version_parity.py (new workflow's pin)
```

End-to-end against a real diff (the commit that added the journey tier):

```
$ PR_BODY="Journey Impact: new: J01" BASE_SHA=571bb964^ HEAD_SHA=571bb964 \
    python3 scripts/ci/check_journey_impact.py
PR declares Journey Impact: 'new: J01'
PASSED — skeleton present for J01 (…)
```

and the rejection path:

```
$ PR_BODY="Journey Impact: new: J11" … 
REJECTED — this PR declares `new: J11`, which obliges this PR to carry a
journey skeleton under tests/journeys/. None found in 0 changed file(s).
A `strict=True` xfail naming J11 is enough:
    @pytest.mark.xfail(strict=True, reason='J11 not built yet')
    def test_j11_promise(): ...
```

## Notes for the reviewer

- `tests/journeys/` lands with #2403 (R1). The gate is correct either way — no PR declares `new:` today, and a `new:` before the tier exists would simply create the first file there.
- **Residual, same as #2336 AC 4:** making this a *required* status check needs repo-admin settings and cannot be done from a PR.

## Type of Change

- [x] CI / process

### DIFF
--- a/.github/ISSUE_TEMPLATE/epic.md
+++ b/.github/ISSUE_TEMPLATE/epic.md
@@ -1,0 +1,39 @@
+---
+name: Epic
+about: A delivery vehicle spanning several PRs, with named sub-issues
+title: 'Epic: '
+labels: type-epic
+assignees: ''
+---
+
+## Charter
+
+What this epic delivers, and what it deliberately does not.
+
+## Journey Impact
+
+<!-- #2350. One line, exactly one of:
+
+       Journey Impact: new: J11
+       Journey Impact: extends: J03
+       Journey Impact: none: <why this touches no promise>
+
+     `none` REQUIRES a reason — a bare `none` fails the check.
+
+     Declaring `new:` obliges this epic's first PR to carry a journey skeleton
+     under tests/journeys/. That is the point: a new promise that nothing
+     asserts is how coverage debt is created silently. A `strict=True` xfail
+     asserting the journey's invariants satisfies it.
+
+     The promise list lives in tests/journeys/catalog.yaml; invariant ids
+     resolve in docs/testing/orchestration-invariant-catalog.md. -->
+
+Journey Impact: 
+
+## Sub-issues
+
+- [ ] 
+
+## Acceptance
+
+How we know the epic is done — and what its rollup % does and does not mean.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,6 +6,22 @@ Brief description of changes.
 
 Fixes #(issue number)
 
+## Journey Impact
+
+<!-- #2350. One line, exactly one of:
+
+       Journey Impact: new: J11
+       Journey Impact: extends: J03
+       Journey Impact: none: <why this touches no promise>
+
+     `none` REQUIRES a reason — a bare `none` fails the check. Declaring `new:`
+     obliges this PR to carry a journey skeleton under tests/journeys/; a
+     `strict=True` xfail asserting the journey's invariants is enough.
+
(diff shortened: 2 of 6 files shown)


obasilakis (approved):
Approving.

CI is fully green (22 checks) and the one thread — CodeQL `py/bad-tag-filter` on the pair-matching regex — is genuinely fixed in `a6b1a7da`: `raw.split("<!--", 1)[0].strip()` also closes the unclosed-`<!--` case the alert documented.

One non-blocking note for the merge: the body says "Related to #2350" with no closing keyword, so #2350 won't auto-close.

Issue #2409, opened 2026-08-26
auto-switch picks an alternative subscription without checking headroom — it can move agents onto a nearly-exhausted one


webmixgamer:
## Summary

`select_best_alternative_subscription` (`src/backend/db/subscriptions.py`) returns the **first** candidate with no recent failure rows and **never reads headroom**. SUB-003 auto-switch therefore relocates agents onto a subscription that may be at 99% of its weekly window — the switch reports success, and the agent hits the wall again shortly afterwards. No human is in the loop, and nothing surfaces that the destination was a bad choice.

## Context

Surfaced while reviewing how subscription headroom is consumed across the platform. The selector's candidate filter is `has_recent_subscription_failures(...)` and nothing else, so headroom never enters the decision.

That filter is **correct and must stay first** — a subscription whose token just failed is unusable no matter how much quota it has, which is what #444 established ("see #444 for what a forgetful candidate filter does" is in the code today). The gap is that among *usable* candidates there is no preference at all: first match wins.

Wrong in both directions:
- a subscription at 99% utilization with no failure rows is picked as if it were a good target;
- a subscription at 5% with one auth failure 90 minutes ago is skipped — correctly, but headroom played no part either way.

The two signals answer different questions and should compose rather than compete: recent failures answer *"can this be used at all?"*, headroom answers *"how long will it last?"*.

## Acceptance Criteria

- [ ] Recent-failure filtering is unchanged and still runs **first** — a candidate with recent failures is never selected regardless of its headroom (#444 must not regress)
- [ ] Among candidates passing that filter, selection prefers the **most headroom** (lowest 7d utilization) from a fresh provider reading
- [ ] **Never returns "no viable target" where it previously returned a candidate.** If every survivor is saturated, the least-saturated one is still selected — leaving an agent on a definitely-limited subscription is strictly worse than moving it to a nearly-full one
- [ ] Missing, stale, or unreadable headroom **never blocks a switch**: candidates with no usable reading stay eligible and sort after those with one, degrading to today's first-match behaviour
- [ ] The selector adds **no probes** — it reads the existing snapshot only, never forcing a provider call on the auto-switch path
- [ ] Tests cover: ranking correctness, all-saturated fallback, no-headroom-data fallback, and that the failure filter still short-circuits ahead of any headroom logic

## Technical Notes

- Selector: `src/backend/db/subscriptions.py::select_best_alternative_subscription`
[...]

PR #2410, opened 2026-08-26
feat(subscriptions): warn before the weekly limit, not after (abilityai/trinity-enterprise#434)

webmixgamer:
## Summary

- Alerts the operator when a subscription's **7-day window** crosses a configurable threshold (default 75%), with a fleet escalation when every subscription is saturated at once.
- **Measuring the provider first deleted most of the specified design.** One query against ent#433's history table shows `seven_day_resets_at` holding *constant* across five days of probes while utilization climbed 36→90, then stepping exactly **+7 days**. The weekly window is **fixed-with-reset, not rolling** — so utilization is monotonic within a window, the specified hysteresis floor is dead code, and `resets_at` **is** the window's identity.
- That makes the alert id the entire state machine — `sub-headroom-{sid}-{reset-day}-{tier}` — with edge-trigger, cross-worker dedup and re-arm falling out of the sink's existing `ON CONFLICT DO NOTHING`. **No durable memo** to leak, race, or clean up on `delete_subscription`.
- **The threshold fires; the projection ranks.** `projected_end = utilization_pct / fraction_of_window_elapsed` sets urgency from the operator's own burn rate rather than a configured time gap that is wrong for everyone: 75% at day 3 projects to 175% (`high`), 75% with hours left projects to ~77% (`low`). The alert is never withheld at the threshold.

## Evidence

Live `subscription_headroom_history`, two windows:

| `seven_day_resets_at` | samples | utilization |
|---|---|---|
| `2026-08-25T00:00:00Z` | 8 over 5 days | 36 → 40 → 41 → 45 → 46 → 47 → **90** |
| `2026-09-01T00:00:00Z` (+7d exactly) | 5 | 17 → 17 → 17 → 18 → 25 |

Under a rolling window with continuous usage `resets_at` would have advanced. It did not move for five days.

## Three states, because two cannot express "we could not tell"

`classify_headroom` → `saturated | has_headroom | unassessable`.

`_headroom_indicates_healthy` returns `False` for stale, rejected-token, transport error **and** a genuinely saturated subscription — so #2396's docstring naming ent#434 as its consumer was wrong: the fleet escalation would have been blocked by exactly the condition it exists to report. **Docstring corrected, body untouched** — it feeds `resolve_rate_limited_now`, so a behavioural edit there moves every `LIMIT` badge (learnings 2026-08-25).

The classifier keys on **utilization, not window status**: `allowed_warning` is deliberately non-blocking (#2396), so a status-driven classifier would file a live 90% reading as `has_headroom`.

## Where it runs

Inside the existing #447 sweep — **not** a second loop, which would mean a second lease over the same probe budget. Per subscription: `recover_probe` **first**, then `ensure_reading` + `classify_headroom`, in **sibling** `try/except` blocks.

Both orderings are load-bearing. `_probe_floor_ok` bounds probes at 60s/subscription, so whichever consumer probes first floors the other out — this way recovery's zero-age snapshot is reused rather than re-fetched; reversed, `recover_probe` answers `"floored"` and #447 quietly stops working. The sibling split is the same argument in the failure direction.

## Two false claims corrected on the way

- The auto-refresh toggle said probing happens **"only while a dashboard is open"** — untrue since #447, more so now. Label and caption rewritten (no per-day figure, per operator decision; it goes in the release note).
- The generic settings catch-all now runs `validate_ops_setting`. **20 ops keys previously accepted `"abc"` or `"-40"` verbatim** there while `/ops/config` validated them.

## What this deliberately does NOT claim

The fleet alert says what was *measured*, never "auto-switch has nowhere to go" — `select_best_alternative_subscription` reads **no headroom at all** and will relocate agents onto a 99% subscription today. That is a real, separate bug: **abilityai/trinity#2409** (P1), which consumes this classifier.

## Changes

| File | What |
|---|---|
| `services/subscription_headroom_alerts.py` | **new** — projection, tier, deterministic id, emitter; no durable state |
| `services/subscription_headroom_service.py` | `classify_headroom` (3-state), `ensure_reading` (one probe decision/sub/cycle), `SAMPLE_INTERVAL_SECONDS`; #2396 docstring corrected |
| `services/subscription_recovery_service.py` | sampling + alert evaluation, sibling guards, bounded concurrency, in-cycle lease refresh, re-priced fail-open rationale |
| `routers/subscriptions.py` | `PUT /settings/headroom-alert-threshold`; auto-refresh GET carries the honest `weekly_alert` status |
| `routers/settings.py` | 422-block the key + `validate_ops_setting` on the catch-all |
| `canary/snapshot.py`, `operator_queue_service.py`, `test_1677_*` | the three registries a platform emitter must be in |
| `SubscriptionsPanel.vue` | toggle copy corrected |
| 5 docs + `learnings.md` (×2) | incl. the rolling-window correction |

## Test Plan

- [x] **59 new tests; 55 of 56 fail without the fix** (verified by reverting the source and keeping the tests — the 1 that passes is the guard asserting #447's `{probed, outcomes}` return shape survives, which must pass in both states)
- [x] 176 pass across the new suite + every neighbouring suite (#447, #471, #2396, ent#433, #1677)
- [x] Full unit suite: **12488 passed / 2 failed**, against a clean-tree baseline of **12431 / 3**. Failure sets differ in *both* directions and each failure passes in isolation — order-dependent, pre-existing (learnings 2026-08-13). The only failure common to both is the known `test_1920` private-submodule scan.
- [x] `/cso --diff`: **no findings at the 8/10 gate**; report at `docs/security-reports/cso-2026-08-26-ent434.md`

## Cost

~24 probes/day/subscription (`max_tokens=1` Haiku) — **4× less than a watched instance already spends** through the 900s ambient refresh. Interval is a constant, not a knob (#1644); inherits #471's 60s floor and single-flight; stops entirely at `threshold_pct = 0`.

## Residual

`create_item` has no UPDATE path, so a 75% row still reads 75% at 92% — the same residual `retention_guard` documents. The escalation is a separate, self-contained item.

Refs abilityai/trinity-enterprise#434


dolho:
## `/review` Report — PR #2410

**Branch**: `feature/ent434-headroom-alert` → `dev` (merge-base `775fab41`)
**Files**: 23 (+2004/-46) · **Scope**: CLEAN (one deliberate adjacent fix, argued in the body)
**Plan completion** (ent#434 AC): 5 done · 1 **partial** · 1 **not done**

Measuring the provider before building the specified state machine is the right call, and the two learnings entries earn their place. The projection-as-urgency design, the id-as-state-machine, the sibling `try/except` isolating #447, and the three registries are all correct and well argued. One blocking defect below, in the one place the PR's own reasoning says it must not be.

---

### Critical

**[C1] The fleet alert can claim "All N registered subscriptions" over a partial denominator** (Confidence: 9/10)
`services/subscription_recovery_service.py` — `_evaluate_alerts`

```python
classifications = {
    r["sid"]: r["classification"] for r in results if r.get("classification")
}
verdict = alerts.fleet_verdict(classifications)
```

`fleet_verdict` is correct in isolation: an `UNASSESSABLE` member blocks the claim. But the dict it is handed is built **only from results that produced a classification**, and it is never reconciled against `subs`. A subscription that was never classified is not `UNASSESSABLE` — it is *absent*, and absent members cannot block anything.

Two reachable paths drop a member:

1. **A sampling exception.** `_sweep_one` deliberately swallows it and leaves `classification: None` — the isolation property your own `test_an_alert_bug_does_not_stop_recovery_detection` asserts (`assert result["classification"] is None`). That `None` is then filtered out by the comprehension above.
2. **Mid-cycle lease yield.** The chunk loop `break`s on `_try_acquire_leadership` returning False, so `results` is shorter than `subs`, and `_evaluate_alerts` is still called with the partial list.

Repro (3 subscriptions, `s3`'s sampling raises, `s1`/`s2` at 88%):

```
SUMMARY: {'probed': 3, 'outcomes': {'not_limited': 3}, 'subscriptions': 3,
          'alerts': {'fleet': True, 'subscriptions': 0}, 'fleet_blocked_reason': None}
EMITTED: sub-headroom-fleet-2026-09-01
  title:    All 2 subscriptions are near their weekly limit
  question: All 2 registered subscriptions are at or past 75% of their weekly limit: one, two.
```

The instance has **three** subscriptions. `s3` was never assessed and may have full headroom. The operator gets a `high` item asserting fleet-wide saturation on two-thirds of the evidence — and because the fleet branch returns early, the per-subscription alerts are suppressed too (`'subscriptions': 0`), so the wrong claim is the *only* thing emitted.

This is exactly AC #3 ("an unassessable subscription blocks the ALL-saturated claim and is named in the alert instead") and the ent#100 rule the module docstring invokes. The pure function honours it; the wiring does not. Note `MIN_FLEET_SIZE = 2` means it needs ≥3 registered subscriptions to bite — real on any multi-subscription instance.

**Fix**: build the classification map from `subs`, not from `results` — a missing or `None` entry is `UNASSESSABLE`, not a non-member:

```python
by_result = {r["sid"]: r.get("classification") for r in results}
classifications = {
    s.id: by_result.get(s.id) or alerts.UNASSESSABLE for s in subs
}
```

That also makes the lease-yield path self-correcting: a partial sweep can no longer produce a positive fleet claim. Worth a `learnings.md` entry with the fix — the class is "the pure predicate was right and the caller shrank its input", which is a sibling of the 2026-08-25 entry already in the ledger.

---

### Informational

**[I1] The toggle copy points at a control that does not exist** (Confidence: 9/10)
`src/frontend/src/components/settings/SubscriptionsPanel.vue:460`

```
rate-limited subscription recovering, and to watch weekly limits for the alert below.
```

There is no alert control below. `grep` finds nothing in `src/frontend/` referencing `weekly_alert`, `headroom-alert-threshold`, or the threshold at all — the backend computes `_weekly_alert_status` and no client renders it. Consequences: AC #1's "Settings-surfaced value" is API-only, and AC #4's "reports itself inactive **on the Settings surface**" (the #2217 lesson the docstring cites) is not delivered — the honest status exists in the payload and reaches nobody. The PR body doesn't claim the UI, but the copy does, which is the same over-claiming class the body quotes from learnings 2026-08-25.

Either ship the threshold + `inactive_reason` control, or drop "the alert below" and file the UI. It should not merge promising a control it doesn't render.

**[I2] `TIER_WARN` is unreachable for any threshold ≥ 90** (Confidence: 8/10)
`escalation_pct` returns `max(threshold_pct, 90)`, and `decide_tier` tests `>= escalate_at` first. At a legal `threshold_pct` of 90–99 the two bounds coincide, so every crossing files as `crit`/`high` and the warning tier is dead. The derived-single-knob argument is right; this consequence just isn't stated anywhere. A line in the endpoint docstring is enough.

**[I3] Two justifications for the same allowlist entry contradict each other** (Confidence: 8/10)
`tests/unit/test_1677_operator_alert_emitters.py`: *"no agent input reaches any field"*. `subscription_headroom_alerts._emit` docstring: *"agent-chosen names **do** reach the body (an agent may spawn children and name them), but they arrive sanitized"*. The docstring is the accurate one. The allowlist reason is the living caller inventory, so it should carry the true claim.

**[I4] Unguarded `int()` on a newly-documented env knob, at import time** (Confidence: 7/10)
`_MAX_CONCURRENT_PROBES = max(1, int(os.getenv("SUBSCRIPTION_SWEEP_CONCURRENCY", "4")))` — module scope, so `SUBSCRIPTION_SWEEP_CONCURRENCY=eight` raises `ValueError` during import and takes the backend down rather than degrading. `${VAR:-4}` in both composes covers unset *and* empty, so only a non-numeric value bites — but this is now a knob advertised in `.env.example`. Same pattern as the pre-existing `RECOVERY_PROBE_SECONDS` / `_DISABLED_POLL_SECONDS` in this file, so it is a house pattern rather than something you introduced; flagging it because #1197 is in the ledger for this shape.

**[I5] Two blocking calls added to an async endpoint's request path** (Confidence: 7/10)
`_weekly_alert_status` runs `db.list_subscriptions()` and `client.ping()` synchronously inside `async def get_headroom_auto_refresh_setting`. Admin-only and low-traffic, and the endpoint already did one sync settings read, so this is small — but the sweep took care to `asyncio.to_thread` the same DB call and this path doesn't.

**[I6] The status can report `active: true` on an unreadable subscription list** (Confidence: 7/10)
On exception `subscription_count = None`; `None == 0` is False, so it falls through every reason arm and returns `active: true, subscription_count: null`. Everything else in this function is scrupulous about naming why it is off — this one path claims health from a failure.

---

### Verified clean

- **Auth**: both new/changed write paths use `assert_admin`, which since #2323 rejects agent principals itself; the new route is two-segment and collides with no `/{subscription_id}` pattern (Invariant #4).
- **Alert content**: names, percentages, counts, timestamps only — no tokens, no provider error bodies (AC #6). `_SID_SAFE` sanitizes the id; `expires_at: None` is correct against the 5s expiry sweep.
- **Registries**: reserved id prefix, canary `_PLATFORM_ALARM_SENTINELS`, and the #1677 allowlist all present (content of the last: see I3).
- **Empty fleet**: `all([]) is True` handled by the explicit short-circuit before `_evaluate_alerts` — and tested.
- **`_earliest_reset` string `min()`**: safe — `_parse_reset` emits a single canonical `%Y-%m-%dT%H:%M:%SZ`, so lexicographic order is chronological.
- **Leadership**: `set(nx)` then own-token `GET`-compare + `expire`; the mid-chunk re-assert yields correctly to a sibling.
- **`validate_ops_setting` on the settings catch-all**: returns `str`, so `db.set_setting` is unaffected; the retention keys were already 422-blocked above it.
- **Suites**: `test_ent434` + `test_1677` = 72 passed; `test_447` + `test_2396` = 39 passed; settings/ops/retention selection = 308 passed, 1 skipped.
- **Docs**: `architecture.md`, `operating-room.md`, the `dashboard-grid-view.md` rolling→fixed-window correction, and two `learnings.md` entries are all accurate against the code.

---

### Summary

- **Critical: 1** — C1 must be fixed before merge; it emits a false fleet-wide claim in the one scenario the design says it must refuse.
- **Informational: 6** — I1 is user-visible and worth resolving in this PR (either the control or the copy).
- **Scope**: clean.

Everything else here is careful work — the measurement that deleted the specified state machine, and the floored-resource learning in particular, are the kind of thing the ledger exists for.

dolho (approved):
## Re-review at `a19dafe7`

Independent pass: I re-derived each finding from the current source and re-ran my original repro rather than reading the commit subjects. **C1 is fixed on both trigger paths, all six informational findings are addressed, and the new surface holds up.**

### C1 — the fleet claim over a partial denominator

Fixed at the right place. `classifications` is now built from `subs` — the roster — with a missing or `None` entry becoming `UNASSESSABLE`, so `fleet_verdict` gets the full membership it was always written to expect.

I re-ran the exact repro from my review (3 subscriptions, `s3`'s sampling raises):

```
{'subscriptions': 3, 'alerts': {'fleet': False, 'subscriptions': 2},
 'fleet_blocked_reason': 'unassessable_members'}
EMITTED: sub-headroom-s1-... | Subscription 'one' passed 75% of its weekly limit
EMITTED: sub-headroom-s2-... | Subscription 'two' passed 75% of its weekly limit
```

Better than the minimum I asked for: the false fleet claim is gone **and** the two genuine per-subscription alerts now fire, where before the early return suppressed them and the wrong claim was the only thing emitted.

I also checked the **second** trigger I named — the mid-cycle lease yield, which I could not exercise in the first review. Forcing the lease to drop after the first chunk:

```
{'probed': 4, 'subscriptions': 6, 'alerts': {'fleet': False, 'subscriptions': 4},
 'fleet_blocked_reason': 'unassessable_members'}
```

Four of six swept, fleet claim blocked, the four real ones still alerted. Closed by construction rather than by a second patch, which is the right shape.

### The informational six

| # | Verdict |
|---|---|
| I1 — copy promised a control that did not exist | **fixed, properly.** A real threshold control, a `weekly_alert` consumer, a store action, and the decidable parts in `utils/headroomAlertSettings.js`. "The alert below" is now true. |
| I2 — `TIER_WARN` unreachable at threshold ≥ 90 | **documented**, and argued rather than just noted — at 90%+ of a weekly window there is no gentle tier. Fair. |
| I3 — allowlist reason contradicted the module docstring | **fixed** — it now says agent-chosen names *do* reach the body but arrive sanitized and capped, and that it is the VOLUME an agent cannot drive. That is the accurate claim. |
| I4 — unguarded `int()` at import | **fixed** via `_env_int`, and the docstring names the exact reason `${VAR:-4}` does not cover it (unset and empty, but not `eight`). |
| I5 — blocking calls on the event loop | **fixed** — the whole status build moved to `asyncio.to_thread`. |
| I6 — `active: true` on an unreadable subscription list | **fixed** with its own `count_unavailable` arm, and the comment says exactly why `None == 0` let it fall through. |

### The new surface

`describeSaveFailure` is a genuinely good addition and not one I asked for: the first version blamed the network for *any* thrown value, which sent an operator hunting a connection problem when the real cause was a stale store after HMR. Claiming an unobserved cause is worse than being vague, and the fix says so.

The `SUBSCRIPTION_SWEEP_CONCURRENCY` packaging catch is worth noting too — there are **three** deployment composes now, not two, because #2280 added the hosted twin while this branch was open. The parity guard caught it, which is the guard doing its job on a file that did not exist when the check was written.

### One low finding, not blocking

**`INACTIVE_COPY` has no guard tying it to the backend's reason set.** All five reasons are covered today — I checked each one — but the parity test iterates `Object.keys(INACTIVE_COPY)`, which is self-referential: it proves every key renders, not that every backend reason has a key. A sixth `reason = "..."` added to `_weekly_alert_status_blocking` would fail nothing.

The blast radius is small because the fallback is honest (`'Inactive — this backend did not report a reason.'`) rather than silent or invented, which is why this is not blocking. If you want it closed, `tests/unit/test_2086_model_catalog_parity.py` is the in-repo precedent for a Python test validating a checked-in JS mirror.

### Verification

118 pass across `test_ent434`, `test_447`, `test_2396` and `test_1677`; 26 pass on the new `headroomAlertSettings` spec.

### Verdict

Approving. The measurement that deleted the specified state machine was the right instinct in the first place, and the follow-up work fixed the one defect that mattered at its source rather than at the symptom.

PR #2412, opened 2026-08-27
fix(ci): serialize the dev deploy build, and let a green deploy re-arm its own alert

dolho:
## Why

`Deploy to Dev` has failed **ten times in a row** since 2026-08-24 — the dev instance has not been running `dev` for three days — and nobody was told, for a second, separate reason.

## The build failure

The stage ran:

```
docker compose build --no-cache backend frontend mcp-server scheduler
```

compose runs those in **parallel**, and `--no-cache` means nothing is reused — four simultaneous dependency downloads on one small VM: two `npm ci` and two `apt-get install`, the backend's pulling ffmpeg and `postgresql-client-17`.

Measured on the failing run, not inferred:

| observation | value |
|---|---|
| `mcp-server` `npm ci` | 142s, then `npm error Exit handler never called!` |
| `frontend` `npm ci` | 144s, never finished (cancelled when mcp-server died) |
| `apt` InRelease fetch | `Ign:` after **61s**, second after **121s** |

Three independent package fetches degrading at once, on one host.

**It is not the code.** No change to `docker/` or `src/mcp-server/` between the last green deploy (`b7bac09d`, 08-21) and the first red one (`c7320a13`, 08-24) — the Dockerfile and that `npm ci` line are byte-identical to when they last worked. And the same `npm ci`, same lockfile, same base-image digest (`node:22-alpine` @ `c610fcdf`, node 22.23.2 / npm 10.9.8) completes in **2 seconds** off that host. The images are not the regression; the concurrent burst is what the host can no longer absorb.

So the build now runs one service at a time. Identical images, one download at a time, and a failure names the service instead of whichever of four lost the race. `--no-cache` stays — it is the clean-build guarantee and it is not what broke. `set -e` already governs the loop body, so a failed build still aborts rather than restarting on a stale image.

Both timeouts move with it **and keep their ordering**: a green deploy ran ~8 min when builds overlapped, so serially `command_timeout: 25m` would just become the new way it dies. It goes to 35m and the job's `timeout-minutes` to 45 — the inner one must fire first, because it ends the SSH command and reports the stage, where the job timeout cancels with no message at all.

## The alerting failure

`notify-failure` dedups on ONE open `deploy-failure` issue: while one is open, every further failure is a deliberate no-op. **Nothing ever closed it** — the reset was a human step the issue body asks for in prose.

It did not happen. #2267 opened 08-17; the deploy recovered on 08-21 and nobody closed it; when `dev` broke again on 08-24 the next ten failures each logged `already tracked at #2267 -- not re-filing` and notified no one. An alerting rule whose reset is a human habit silences itself on its second outage.

`notify-recovery` closes it on a green deploy so the dedup re-arms itself. It closes *every* open one (two can exist if the label was applied by hand, and leaving one behind restores the suppression), and every step tolerates its own failure — the deploy already succeeded, and a red run over bookkeeping would be its own false alarm.

## Honest scope

Serializing is a **mitigation for a host-side constraint**, not a repair of it. If the dev VM is short on memory, disk or egress, that is still true after this merge — this only stops the deploy from being the thing that tips it over. Worth checking on the host directly:

```
free -h; df -h /var/lib/docker; docker system df
journalctl -u docker --since '2026-08-24' | grep -i -e oom -e killed
```

If `docker system df` shows a large reclaimable figure, `docker builder prune` is likely overdue — `--no-cache` on every deploy since 08-24 has been writing build layers that nothing reuses.

## Verification

- `yaml.safe_load` parses; jobs resolve to `deploy`, `notify-recovery`, `notify-failure` with `success()` / `failure()` guards.
- `bash -n` clean on the ssh-action `script:` and on the new job's `run:`.
- `command_timeout` (35m) < job `timeout-minutes` (45), the required ordering.
- Cannot be verified end to end without merging: the workflow only runs on push to `dev`.

Related to #2267.

### DIFF
--- a/.github/workflows/deploy-dev.yml
+++ b/.github/workflows/deploy-dev.yml
@@ -28,7 +28,12 @@ concurrency:
 jobs:
   deploy:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    # MUST stay above the ssh-action's `command_timeout` (35m). The inner
+    # timeout ends the SSH command and reports which stage was running; this
+    # one cancels the job with no such message, so if it fires first every
+    # slow-build failure becomes unreadable. Raised alongside the serialized
+    # build, keeping that ordering.
+    timeout-minutes: 45
 
     steps:
       - name: Connect to Tailscale
@@ -46,7 +51,13 @@ jobs:
           host: ${{ secrets.DEV_HOST }}
           username: ${{ secrets.DEV_USER }}
           key: ${{ secrets.DEV_SSH_KEY }}
-          command_timeout: 25m
+          # 25m until the build was serialized. A green deploy ran ~8 min
+          # end to end while the four images built in PARALLEL; built one at a
+          # time the build stage is the sum rather than the max, and this
+          # bound would become the new way the deploy dies — on the same slow
+          # host that made serializing necessary. Raised with margin; it is a
+          # ceiling on a hung SSH command, not a target.
+          command_timeout: 35m
           script: |
             set -e
             cd ~/trinity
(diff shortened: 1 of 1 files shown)


obasilakis (approved):
`/validate-pr` — approving.

Two separate failures correctly separated, which is what makes this reviewable at all. The build contention and the silenced alert are unrelated causes of the same visible symptom, and treating them as one bug would have fixed whichever was noticed first.

**On the build.** The evidence is the right kind: no change to `docker/` or `src/mcp-server/` between the last green deploy and the first red one, same lockfile, same base-image digest, and the same `npm ci` completing in 2 seconds off that host. That rules out the images and leaves the concurrent burst, which is what the fix addresses. Keeping `--no-cache` is the right call — it is the clean-build guarantee and it is not what broke. Serial buys determinism and a failure that names the service that failed rather than whichever of four lost the race.

Raising both timeouts together, and keeping `command_timeout` below the job timeout so the inner one fires first and reports which stage was running, is the detail that stops the next slow build from failing unreadably.

**On the alert.** This is the more important half. A one-way dedup whose reset is a human habit is a rule that silences itself on its second outage, and it did: ten consecutive failed deploys each logged "already tracked at #2267 — not re-filing" and told nobody. Making the recovery close the issue turns the dedup self-resetting, which is the only version of this that survives contact with a team.

Closing *every* open `deploy-failure` issue rather than the first is right — leaving one behind restores the suppression the job exists to remove. `set -uo pipefail` without `-e`, plus `|| true` on both `gh` calls, is deliberate and correct: the deploy already succeeded, and reporting a red run for a bookkeeping error would be its own false alarm.

Checklist: base `dev`, security scan clean, no packaging surface, workflow-only change so there is nothing to unit-test. CI green.

One process note, non-blocking: the PR references `#2267` without a closing keyword, so `issue-status-on-merge.yml` will not promote it. In this case that is nearly moot — the new `notify-recovery` job closes `#2267` itself on the first green deploy after merge, which is arguably the more honest signal since it closes on evidence rather than on merge. Flagging it only so the absence reads as intentional rather than as an oversight.

PR #2413, opened 2026-08-27
fix(ui): the SSH toggle read the ops payload one level too shallow, so it inverted the operator's intent (#2411)

### LINKED ISSUE #2411: bug: the SSH access toggle reads the wrong payload shape — it always renders OFF and the first click inverts the operator's intent
## Summary

The **SSH access** toggle in Settings reads the ops-config response one level too shallow, so it is always `undefined`. The switch therefore **renders OFF regardless of the stored value**, and because the click handler computes the new value by negating that always-false state, the **first click always sends `true`** — an operator who has ephemeral SSH enabled and clicks intending to disable it re-enables it instead, then sees it render as off.

The write path is correct; only the read is broken. That is why it has gone unnoticed: a change appears to stick within the session and silently reverts to "off" on reload.

## Context

`GET /api/settings/ops/config` returns the settings nested:

```python
# src/backend/routers/settings.py
return {
    "settings": ops_config
}
```

`src/frontend/src/views/Settings.vue` reads them flat:

```js
const response = await axios.get('/api/settings/ops/config', { headers: authStore.authHeader })
sshAccessEnabled.value = response.data.ssh_access_enabled === 'true'
//                                     ^^^ always undefined — the value is under .settings
```

`undefined === 'true'` is `false`, so `sshAccessEnabled` is pinned to `false` on every load. The toggle then does:

```js
const newValue = !sshAccessEnabled.value   // always true on the first click
await axios.put('/api/settings/ops/config', { settings: { ssh_access_enabled: newValue ? 'true' : 'false' } })
```

Note the PUT sends the correct **nested** shape — the mismatch is one-sided.

**Why P1 rather than cosmetic.** This is not an enforcement bypass — the backend honours `ssh_access_enabled` correctly, and SSH credential minting is admin-only. It is a *control that misreports its own state in the unsafe direction*: the operator believes ephemeral SSH access is disabled while it may be enabled, and their attempt to disable it turns it on. A security control that lies about itself is worth fixing before a cosmetic one.

Found while mapping the settings surface during unrelated work (`abilityai/trinity-enterprise#434`); deliberately not fixed in that PR, which is a subscription-alerting change — bundling a frontend fix there is the scope drift `/review` exists to catch.

## Acceptance Criteria

- [ ] The toggle reflects the **stored** value on load (read `response.data.settings?.ssh_access_enabled`), verified with the setting both enabled and disabled
- [ ] With SSH enabled, one click **disables** it — the first click no longer inverts the operator's intent
- [ ] A regression test naming this issue that fails without the fix: given a `{"settings": {"ssh_access_enabled": "true"}}` response, the toggle renders ON
- [ ] Optional-chaining or an equivalent guard, so a malformed/absent payload degrades to `false` (the safe direction) rather than throwing
- [ ] Confirm no sibling binding in `Settings.vue` shares the mismatch — at time of filing `/api/settings/ops/config` is read in exactly one place, so this is the only occurrence; re-verify if other ops settings have since been surfaced

## Technical Notes

- Read site: `src/frontend/src/views/Settings.vue` (the `/api/settings/ops/config` GET handler)
- Endpoint: `src/backend/routers/settings.py`, `GET /ops/config` — returns `{"settings": ops_config}`
- The one-line fix is obvious; the value is in the **test**, since nothing currently pins the payload shape across this boundary and the same class can reappear the next time an ops setting is surfaced in this view.
- Do **not** "fix" it by flattening the endpoint response: `PUT /ops/config` takes the nested shape, `GET /api/settings/retention` and the other ops readers depend on the current contract, and the endpoint is the side that is right.

dolho:
## Summary

The SSH access toggle read `GET /api/settings/ops/config` flat while the endpoint nests settings twice, so it was `undefined === 'true'` on every load: the switch rendered **OFF whatever was stored**, and since the click handler negates that state, the **first click always sent `true`** — an operator with ephemeral SSH enabled, clicking to disable it, re-enabled it and then watched it render as off.

The write path was always correct. The mismatch was one-sided, which is why it looked like a change stuck for the session and silently reverted on reload.

## The issue's own suggested fix is not sufficient

#2411 proposes `response.data.settings?.ssh_access_enabled`. That is **still wrong**, for the same underlying reason. `get_ops_settings` builds a *descriptor* per key:

```python
ops_config[key] = {"value": current_value, "default": default_value,
                   "description": ..., "is_default": ...}
```

So `settings.ssh_access_enabled` is an **object**, and `object === 'true'` is `false` — the toggle would still have rendered OFF while looking fixed. The `.value` hop is the one that matters, and a test pins it so the half-fix cannot come back as a "simplification".

## Why a pure module rather than a one-line edit

The issue says the value is in the test rather than the fix, and that is only achievable with a seam: `Settings.vue` cannot be mounted here — `@vue/test-utils` is not a dependency and vitest runs `environment: 'node'` — so a rule kept inline is a rule **no test can reach**. That is exactly how a one-line read bug survived inside a security control.

`utils/opsSettings.js` owns both spellings — `readOpsBool` for the read, `opsBoolValue` for the write — in one file rather than at opposite ends of a 3500-line SFC, which is what let them drift.

## Acceptance criteria

| AC | Status |
|---|---|
| Toggle reflects the stored value, verified both enabled and disabled | ✅ |
| With SSH enabled, one click **disables** it | ✅ |
| Regression test naming the issue that fails without the fix | ✅ mutation-checked, below |
| Malformed/absent payload degrades to `false` (safe direction) | ✅ 9 cases |
| No sibling binding shares the mismatch | ✅ and now **pinned by a test**, not eyeballed once |

The endpoint is untouched, per the issue's explicit instruction — `PUT /ops/config` takes the nested shape and the other ops readers depend on the current contract.

`false` is the safe degrade direction on purpose: `ssh_access_enabled` defaults to `"false"` server-side, and a control that cannot read its own state must not claim the permissive one.

## Verification

Each side of the fix was mutation-checked rather than assumed:

```
reader reverted to payload?.[key]           → 4 failed
SFC reverted to the flat read that shipped  → 2 failed
restored                                     → 18 passed
```

Full frontend suite: **64 files / 1454 tests**, green. `vite build` clean. Raw-color ratchet unaffected (no new colour usage).

Closes #2411

### DIFF
--- a/src/frontend/src/utils/opsSettings.js
+++ b/src/frontend/src/utils/opsSettings.js
@@ -1,0 +1,60 @@
+/**
+ * Reading `GET /api/settings/ops/config` (#2411).
+ *
+ * The payload is nested TWICE, and the toggle in `Settings.vue` read it flat:
+ *
+ *     { settings: { ssh_access_enabled: { value: "true", default: "false",
+ *                                         description: "...", is_default: false } } }
+ *
+ * `response.data.ssh_access_enabled` is therefore `undefined`, `undefined ===
+ * 'true'` is `false`, and the SSH toggle rendered OFF no matter what was
+ * stored. Worse, the click handler computes `!sshAccessEnabled.value`, so with
+ * the state pinned to `false` the FIRST click always sent `true` — an operator
+ * with ephemeral SSH enabled, clicking to disable it, re-enabled it and then
+ * watched it render as off. The write path was always correct; only the read
+ * was broken, which is why a change appeared to stick for the session and
+ * silently reverted on reload.
+ *
+ * The rule lives here rather than inline because `Settings.vue` cannot be
+ * mounted in this project's test setup (`@vue/test-utils` is not a dependency
+ * and vitest runs `environment: 'node'`), so a rule kept in the SFC is a rule
+ * no test can reach — which is exactly how a one-line read bug survived in a
+ * security control. Issue #2411's own suggested fix,
+ * `response.data.settings?.ssh_access_enabled`, is still wrong for the same
+ * underlying reason: that resolves to the descriptor OBJECT, and
+ * `object === 'true'` is `false`. The `.value` hop is the one that matters.
+ */
+
+/**
+ * Read one ops setting as a boolean, from either payload shape.
+ *
+ * Accepts the descriptor form the endpoint actually returns
+ * (`{ value: "true", … }`) and a bare `"true"`, because the only thing that
+ * distinguishes them is a wrapper this reader does not need — and being
+ * tolerant here costs nothing while removing a whole class of one-sided
+ * mismatch.
+ *
+ * Everything unreadable degrades to `false`: absent key, absent `settings`,
+ * a null payload, a number, an unexpected object. `false` is the SAFE
+ * direction for every flag in this group — `ssh_access_enabled` defaults to
+ * `"false"` server-side, and a security control that cannot read its own state
+ * must not claim the permissive one.
+ */
+export function readOpsBool(payload, key) {
+  const entry = payload?.settings?.[key]
+  const raw = entry !== null && typeof entry === 'object' ? entry.value : entry
+  // Only a string can be `"true"`; a real boolean `true` is accepted too, since
+  // the column is TEXT today but nothing here should break if that changes.
+  if (raw === true) return true
+  return typeof raw === 'string' && raw.trim().toLowerCase() === 'true'
+}
+
+/**
+ * The value to SEND for a boolean ops setting. `PUT /ops/config` takes strings
+ * (`validate_ops_setting` requires `'true'`/`'false'` for a `bool` key), and
+ * pairing the writer with the reader here keeps the two spellings in one file
+ * rather than at opposite ends of a 3500-line SFC.
+ */
+export function opsBoolValue(enabled) {
+  return enabled ? 'true' : 'false'
+}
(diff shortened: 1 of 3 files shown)


obasilakis (approved):
`/validate-pr` — approving.

Correct diagnosis and the fix goes to the right place. Worth stating plainly what this was: a security control that rendered OFF regardless of what was stored, on a UI whose click handler negates the rendered state — so an operator with ephemeral SSH enabled, clicking to *disable* it, re-enabled it and then watched it render as off. The write path was always right, which is exactly why it looked like it stuck for the session and reverted on reload.

The detail that makes this more than a one-line patch is noticing that the fix the issue itself proposes — `response.data.settings?.ssh_access_enabled` — is still wrong: it resolves to the descriptor object, and `object === 'true'` is false. The toggle would have stayed broken while looking fixed, and the test pinning that is the most valuable assertion in the file.

Degrading everything unreadable to `false` is the right direction for this flag specifically, and the comment says why rather than asserting it.

The `axios.get('/api/settings/ops/config')` call-count assertion is a good structural pin: it makes a second reader added later come through the same rule instead of re-deriving the shape.

Checklist: base `dev`, `Closes #2411`, security scan clean, no packaging or infra surface, no schema change. Docs commit-message-only per the tiered rule. CI green including e2e.

One note, non-blocking: `readOpsBool` accepts `'TRUE'` and `' true '` while the server's `validate_ops_setting` requires exactly `'true'`/`'false'`. The comment justifies it (a hand-edited DB row should not silently read as false) and the asymmetry is in the tolerant direction on the read side only, so it is fine — just flagging that reader and writer no longer agree on the accepted vocabulary.

`theme-security` on a P1: this is a frontend read fix and the backend gate on ephemeral SSH minting is untouched, so I did not run `/cso --diff` over it.

PR #2414, opened 2026-08-27
fix(operator-queue): an approval's decision must be one the agent offered (#2376)

### LINKED ISSUE #2376: bug: POST /api/operator-queue/{id}/respond accepts any string as an approval decision — no layer checks response ∈ options

dolho:
## Summary

`OperatorResponse.response` is a bare `str`, and the router, the DB write and the write-back into the agent's own queue file all passed it through verbatim. No layer checked that an **approval** item's recorded decision was one of the options the agent authored — so #2370 shipped `response: "approved"` against `options: ["Approve", "Deny"]` for five months without a single 4xx, and the agent read back a decision string it never offered.

## At the sink, not in a router

This is the approval channel for irreversible actions (#1402 poison-park, ent#329 respond→resume, the TARGET_ARCHITECTURE v2 human gate), with four producers today. A rule living in whichever router is next re-ships the class the first time someone adds a channel button or an ops agent — so `services/operator_queue_choices.py` owns it and both writers call it.

A **leaf** module rather than `operator_queue_service`: that module instantiates its sync-service singleton and imports `AgentClient` at module scope, and the Workspace asks path must not drag either in to answer a question about a list of strings.

## The rule, and what it deliberately does not do

- **Exact match.** The options are agent-authored, so only the agent knows whether `approve` and `Approve` mean the same thing to it. Normalising would answer that on its behalf, when the point is that the recorded decision is one the agent can compare against its own list.
- **The 422 carries the offered options.** A refusal saying only "invalid" leaves the operator guessing at strings they never wrote.
- **An empty `response` is not rejected.** Both entry points allow an answer carried entirely by `response_text`; a 422 there would break a working path to validate a field nobody filled in.
- **`acknowledged` never reaches the rule** because both writers refuse a non-pending item first — structural, and pinned by asserting the ordering rather than by comment.

Exempt cases are all spelled as the same absence rather than as special cases at the call site: questions, alerts, approvals offering nothing, a non-list or non-string blob, and — worth naming — one whose options are the **#1632 size-cap marker**. That placeholder *records that the choices were dropped*, so treating it as offered would let it be recorded as a decision, and treating the item as constrained would make it unanswerable.

`_OPTIONS_DROPPED_MARKER` now imports from the leaf: two literals drift the day either side is reworded, and the sink would then accept the placeholder as a decision.

## Acceptance criteria

| AC | Status |
|---|---|
| Named 422 listing offered options, exact match, real-options only | ✅ |
| Questions / alerts / no-usable-options / `acknowledged` unaffected | ✅ 7 exempt shapes |
| Check in ONE place reached by both entry points | ✅ and **enforced**, not just done |
| MCP surfaces the 422 structurally | ✅ pinned by test |
| Tests on both entry points incl. the marker-list case | ✅ 18 |

## Verification

AC #3 is the one worth calling out: rather than asserting "both call it", an **AST caller-parity guard** (the #1677 shape) pins the writer set of `db.respond_to_operator_queue_item` to exactly the two known callers and requires each to validate. A third writer fails CI.

```
MUTANT: asks path stops validating   → 2 failed
restored                              → 18 passed
```

244 pass across the operator-queue, asks, #1632, #1677 and ent#329 suites.

MCP needed no change — `ApiError` embeds the raw body in `.message`, so the structured detail already reaches `respond_to_operator_queue`'s `{error: msg}`. Pinned by test rather than assumed.

Closes #2376

### DIFF
--- a/docs/memory/feature-flows/operating-room.md
+++ b/docs/memory/feature-flows/operating-room.md
@@ -191,6 +191,30 @@ Event handling in the store (`handleWebSocketEvent`, line 177-200):
 
 The type pill text comes from `utils/operatorQueue.js::queueTypeLabel` (`Needs approval` / `Question` / `Heads up`), shared with `/m` (#2370). **The `/m` mobile admin mirrors this table** — approval: select → restated consequence → optional note → `Cancel` + `Send: <option>` (nothing sends on one tap); question: text + Send; alert: `Got it` — and builds the identical body via `buildQueueResponse` (see [mobile-admin-pwa.md](mobile-admin-pwa.md)). The Workspace asks panel posts to its own endpoint/model and is **not** yet a consumer of the shared builder (#2375).
 
+**An approval's decision must be one it OFFERED (#2376).** `OperatorResponse.response`
+is a bare `str` and every layer passed it through verbatim — router, DB write,
+and the write-back into the agent's own queue file — so no layer checked
+membership. #2370 recorded `response: "approved"` against
+`options: ["Approve", "Deny"]` for five months without a single 4xx, and the
+agent read back a decision string it never offered.
+`services/operator_queue_choices.py` is the one rule: for `type == "approval"`
+with a usable options list, a non-empty `response` outside it is a named **422**
+`response_not_an_offered_option` **carrying the offered list** — a refusal that
+does not name agent-authored strings leaves the operator guessing. Matching is
+**exact**: only the agent knows whether `approve` and `Approve` mean the same
+thing to it, so normalising here would answer that on its behalf. Exempt, all
+spelled as the same absence rather than as special cases at the call site:
+questions, alerts, approvals that offered nothing, a non-list or non-string
+blob, and one whose options were replaced by the #1632 size-cap marker (that
+placeholder RECORDS that the choices were dropped, so it is neither binding nor
+selectable). A text-only answer is untouched — an empty `response` means the
+field was never filled in, not that it is wrong. The rule sits at the SINK
+rather than in a router because this is the approval channel for irreversible
+actions (#1402, ent#329) with four producers today; both writers
+(`routers/operator_queue.py`, `client_portal/asks/service.py`) call it, and an
+AST caller-parity guard (`tests/unit/test_2376_approval_response_membership.py`,
+the #1677 shape) fails CI if a third writer appears without it.
+
 ### NavBar Badge
 
 `src/frontend/src/components/NavBar.vue`:
--- a/src/backend/client_portal/asks/router.py
+++ b/src/backend/client_portal/asks/router.py
@@ -30,8 +30,10 @@ router = APIRouter(
 
 
 def _raise(e: AskError):
+    # #2376: merge any structured fields the refusal carries (the offered
+    # options, today) beside the code and message rather than only in prose.
     raise HTTPException(status_code=e.status_code,
-                        detail={"code": e.code, "message": e.detail})
+                        detail={"code": e.code, "message": e.detail, **e.data})
 
 
 @router.get("", response_model=List[WorkspaceAsk])
--- a/src/backend/client_portal/asks/service.py
+++ b/src/backend/client_portal/asks/service.py
@@ -12,6 +12,10 @@ import logging
 from typing import List, Optional
 
 from database import db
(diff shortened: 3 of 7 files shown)


obasilakis (approved):
`/validate-pr` — approving.

The placement is the whole value here, and it is right. This is the approval channel for irreversible actions with four producers today, so a check in whichever router happened to be next would have been the fifth place to forget it. Putting it at the sink and pinning that with an AST caller-parity guard is the #1677 shape applied correctly.

Things I checked and agree with:

- **Exact match, deliberately.** The options are agent-authored, so normalising `approve` to `Approve` would answer on the agent's behalf a question only the agent can answer. The docstring says this rather than leaving it to be rediscovered.
- **The dropped-options marker is now one definition, imported.** `operator_queue_service` holding its own literal while the validator held a second copy is exactly how the sink would have ended up accepting a size-cap placeholder as a decision. Good catch to fold it.
- **The exemptions are spelled as one absence.** Questions, alerts, an approval that offered nothing, a non-list blob, and a marker-only list all return `None` from `usable_options` rather than becoming five special cases at two call sites.
- **An empty `response` is not rejected.** Turning a text-only answer into a 422 would have broken a working path in the name of validating a field that was never filled in; the Workspace path keeps its own `empty_answer` refusal for a genuinely empty one.
- **The 422 carries `offered_options`.** A refusal that does not name agent-authored strings leaves the operator guessing, and re-deriving them by parsing the message is what breaks the day the wording changes. Threading `AskError.data` through `_raise` rather than stuffing it into prose is the right plumbing.

That `#2370` recorded `"approved"` against `["Approve", "Deny"]` for five months without a single 4xx is a good piece of evidence to have put in the PR body.

Checklist: base `dev`, `Closes #2376`, feature flow updated with the rule and its exemptions, security scan clean, no packaging or schema surface, both writers reach the validator, CI green.

One non-blocking note: `src/mcp-server/src/tools/operator_queue.ts` is the fourth producer and routes through this endpoint, so it is covered — but its inline comment still says only *"Backend 400s on a non-pending item"* and the tool description does not mention that an approval's response must be one of the offered options. An agent will now hit a 422 the tool never told it about. Worth a sentence there when convenient.

PR #2416, opened 2026-08-27
fix(ui): emotion avatars were cache-busted by a constant, so a regenerated avatar kept its old face for 24h (#2374)

### LINKED ISSUE #2374: bug: stale avatar shown when cycling emotion modes after avatar change/regeneration
## Summary

After an agent's avatar is changed or regenerated, cycling through emotion modes can still display emotion variants of the **old** avatar. The header swaps to the new base avatar, but the emotion cycle (AVATAR-002) intermittently shows stale images until a hard refresh or until the browser cache expires.

## Context

Reported from live use: change/regenerate an avatar, then click the "Next emotion" button (or let the 30s auto-cycle run) — the previous avatar's face reappears.

The moving parts:

- `src/backend/routers/avatar.py` serves emotion variants at a **stable URL** (`GET /api/agents/{name}/avatar/emotion/{emotion}`) with `Cache-Control: public, max-age=86400` (24h browser cache).
- `src/frontend/src/views/AgentDetail.vue::cycleEmotion()` cache-busts with `?v=${avatarVersion}`, where `avatarVersion = agent.avatar_url?.split('v=')[1] || '1'`.

Two ways this goes stale:

1. **Version derived from possibly-stale state**: if the in-memory `agent.avatar_url` isn't refreshed after regeneration (or the avatar was regenerated in another tab/session), the `?v=` value doesn't change, so every emotion URL keys the browser's day-old cache entry for the previous avatar.
2. **Static fallback**: when `avatar_url` carries no `v=` query param, the fallback collapses every version to `'1'` — emotion URLs are then permanently `?v=1`, and any regeneration serves cached old variants for up to 24h.

A secondary race exists during regeneration: the upload path deletes old emotion files and regenerates 8 variants one-by-one in the background (~15s each), while the frontend polls `/avatar/emotions` and starts cycling as soon as the first variant appears — worth verifying the polling can't pick up a not-yet-replaced file, but the cache-version issue above is the primary suspect.

## Acceptance Criteria

- [ ] After changing or regenerating an avatar, cycling emotion modes never displays a variant belonging to the previous avatar (same session, no hard refresh required)
- [ ] Emotion variant URLs are cache-busted by a version tied to the actual avatar generation (e.g., derived from `avatar_updated_at`), not parsed out of the display URL with a static `'1'` fallback
- [ ] An avatar regenerated in another session/tab is reflected for other viewers once the agent payload refreshes — no residual 24h stale window from the `max-age=86400` cache
- [ ] Regression test covering the emotion-URL version derivation / cache-busting path

## Technical Notes

- Backend: `src/backend/routers/avatar.py` — emotion serving (`/avatar/emotion/{emotion}`, `/avatar/emotions`), background generation `_generate_emotions_background`, and the regenerate path that unlinks old `_emotion_*` files.
- Frontend: `src/frontend/src/views/AgentDetail.vue` (`cycleEmotion`, `loadAvailableEmotions`, the post-regenerate poll) and `src/frontend/src/components/AgentHeader.vue` (renders `emotionAvatarUrl || agent.avatar_url`, so a stale emotion URL overrides even a fresh base avatar).
- Simplest likely fix: have the backend return a stamp with `/avatar/emotions` (or use `agent.avatar_updated_at` from the agent payload) and build emotion URLs from that, refreshing the agent object after regeneration completes.

dolho:
## Summary

Emotion variants are served from a **stable** URL under `Cache-Control: public, max-age=86400`, so the `?v=` the client appends is the only thing separating a fresh variant from a day-old cached one. It was:

```js
const avatarVersion = agent.value.avatar_url?.split('v=')[1] || '1'
```

Two failures, and the second is what made it permanent:

1. **Stale source.** Read out of possibly-stale in-memory state — an avatar regenerated in another tab, or before `agent` refreshed, leaves `avatar_url` unchanged, so every emotion URL keys the same old cache entry.
2. **A constant is not a cache-buster.** With no `v=` on `avatar_url` the fallback collapses to `'1'`. Every emotion URL is then permanently `?v=1`, and *no regeneration can ever change it* — the browser serves the previous face until the entry expires 24 hours later.

## The stamp tracks the variants, not the base avatar

`GET /avatar/emotions` now returns a `version`: the newest mtime among the files it is listing.

Deliberately **not** `avatar_updated_at`, though the issue suggests it. These URLs serve the *variants*, and regeneration replaces them one at a time in the background — so a stamp taken from the files moves as the new set lands, and a client polling this endpoint re-keys its URLs progressively instead of holding whatever it cached on the first poll. It also doesn't depend on the DB column being populated. This addresses the secondary race the issue flags, not just the primary suspect.

`"0"` (no variants) is treated as *absent* rather than as a new constant. A file that vanishes between `exists()` and `stat()` is mid-regeneration — skipping its mtime only understates the version and the next poll picks it up, so it never fails the listing.

## The fallback ladder never ends in a constant

endpoint stamp → the `v=` already on `avatar_url` (correct whenever the payload is fresh; the path for a backend predating the stamp) → a per-load value.

The last rung costs one refetch per page load and *cannot go stale*, which is the safe direction for a control whose failure mode is showing the wrong person's face. A failed poll **clears** the stamp rather than reusing a stale one — keeping it would be the same bug by a slower route.

The derivation lives in `utils/avatarEmotion.js` because `AgentDetail.vue` cannot be mounted in this project's test setup (`@vue/test-utils` isn't a dependency, vitest runs `environment: 'node'`), so a rule kept inline is a rule no test can reach.

## Acceptance criteria

| AC | Status |
|---|---|
| Cycling never shows a previous avatar's variant, same session | ✅ |
| Version tied to actual generation, not parsed with a `'1'` fallback | ✅ tied to the variant files themselves |
| Regeneration elsewhere reflected without a 24h stale window | ✅ stamp moves per poll |
| Regression test covering the version derivation | ✅ 17, mutation-checked |

## Verification

```
MUTANT: constant '1' fallback restored → 7 failed
restored                                → 17 passed
```

Full frontend suite 64 files / 1453 tests; `vite build` clean; backend avatar tests pass.

Closes #2374

### DIFF
--- a/src/backend/routers/avatar.py
+++ b/src/backend/routers/avatar.py
@@ -365,14 +365,46 @@ async def _generate_emotions_background(
 
 @router.get("/{agent_name}/avatar/emotions")
 async def get_avatar_emotions(agent_name: str):
-    """Return which emotion variant PNGs exist on disk for an agent. No auth required."""
+    """Which emotion variant images exist for an agent, and their version.
+
+    #2374: the `version` is the cache key the client MUST build emotion URLs
+    with. Variants are served from a stable URL under
+    `Cache-Control: public, max-age=86400`, so without a version tied to the
+    actual files a regenerated avatar keeps showing the OLD face for up to 24
+    hours. The client used to parse `?v=` out of `avatar_url` and fall back to
+    the constant `'1'` — which pins every emotion URL to one cache entry
+    forever, so regeneration was invisible until the entry expired.
+
+    The stamp is the newest mtime among the files being returned, not
+    `avatar_updated_at`. It tracks the VARIANTS rather than the base avatar,
+    which is what these URLs actually serve: regeneration replaces them one at a
+    time in the background, and each replacement moves the stamp, so a client
+    polling this endpoint re-keys its URLs as the new set lands instead of
+    holding whatever it cached on the first poll.
+
+    `0` means no variants exist — there is nothing to cache-bust.
+    """
     available = []
+    newest = 0.0
     for emotion in AVATAR_EMOTIONS:
-        if (AVATAR_DIR / f"{agent_name}_emotion_{emotion}.webp").exists():
+        for path in (AVATAR_DIR / f"{agent_name}_emotion_{emotion}.webp",
+                     AVATAR_DIR / f"{agent_name}_emotion_{emotion}.png"):
+            if not path.exists():
+                continue
             available.append(emotion)
-        elif (AVATAR_DIR / f"{agent_name}_emotion_{emotion}.png").exists():
-            available.append(emotion)
-    return {"agent_name": agent_name, "emotions": available}
+            try:
+                newest = max(newest, path.stat().st_mtime)
+            except OSError:
+                # A file that vanished between exists() and stat() is mid-
+                # regeneration. Skipping its mtime only understates the version,
+                # and the next poll picks it up — never fail the listing over it.
+                pass
+            break
+    return {
+        "agent_name": agent_name,
+        "emotions": available,
+        "version": str(int(newest)),
+    }
 
 
 @router.get("/{agent_name}/avatar/emotion/{emotion}")
--- a/src/frontend/src/utils/avatarEmotion.js
+++ b/src/frontend/src/utils/avatarEmotion.js
@@ -1,0 +1,62 @@
+/**
+ * Cache-busting for emotion avatar variants (#2374).
+ *
+ * Variants are served from a STABLE url — `/avatar/emotion/{emotion}` — under
+ * `Cache-Control: public, max-age=86400`. The only thing separating a fresh
+ * variant from a day-old cached one is the `?v=` the client appends, so getting
+ * that version wrong shows the PREVIOUS avatar's face after a regeneration.
+ *
+ * It was `agent.avatar_url?.split('v=')[1] || '1'`, which fails two ways:
+ *
+ *   1. It reads the version out of possibly-stale in-memory state. An avatar
+ *      regenerated in another tab, or before `agent` was refreshed, leaves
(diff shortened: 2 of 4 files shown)


obasilakis (approved):
`/validate-pr` — approving.

Clean fix with the right shape. The version is derived from the files these URLs actually serve rather than from the base avatar's `avatar_updated_at`, which is what makes it move as background regeneration replaces variants one at a time. Extracting the derivation into `utils/avatarEmotion.js` because `AgentDetail.vue` cannot be mounted in this test setup is the correct call, and the reason is stated in the module rather than left implicit.

The part I want to single out: refusing to let the fallback be a constant. A constant is the actual defect — `|| '1'` pinned every emotion URL to one cache entry for 24 hours — and `String(Date.now())` costs one refetch per page load while being structurally incapable of going stale. The `"0"` case falling through rather than becoming a new constant is the same instinct applied once more.

Test coverage is genuinely non-happy-path: empty `v=`, `v=` followed by another param, null and undefined URLs, and a source-level assertion that the failed-poll branch clears the stamp rather than reusing a stale one. The last one is the case that would have re-created the bug by a slower route.

Checklist: base `dev`, `Closes #2374` present so the issue auto-promotes, security scan clean, no infra or packaging surface, no new top-level backend module. Docs are commit-message-only per the tiered rule for a bug fix. CI green.

Two non-blocking notes, neither worth another round:

1. `str(int(newest))` is second-resolution, so two regenerations landing inside the same second produce the same version. Vanishingly unlikely given the background generator's pacing, and the next poll fixes it anyway.
2. `GET /{agent}/avatar/emotions` is unauthenticated, so the stamp discloses when an agent's avatar was last regenerated to anyone who can guess a name. The endpoint already leaked which emotions exist for a given name, so this widens an existing surface rather than opening one, and the value is an mtime rather than anything sensitive.

Issue #1593, opened 2026-07-13
operator intake: non-2xx delivery failure is logged at debug — endpoint outage was invisible for 3 weeks


vybe:
## Problem

`services/operator_intake_service.py` logs a non-2xx response from the hosted intake endpoint at **debug** level (`logger.debug("Operator intake POST returned HTTP %s", ...)`) and burns the at-most-once `operator_intake_submitted` marker **before** the POST.

Real-world impact: the hosted `/v1/operator-intake` endpoint didn't exist between 2026-06-22 (feature ship) and 2026-07-13 (endpoint deployed) — every consent-checked opt-in from fresh installs hit a 404 and was silently dropped, and nothing in any log surfaced it. Because the marker is claimed pre-POST, those installs will never re-submit.

## Proposed fix (small)

- Log the non-2xx branch at **WARNING** with the status code (never the email — existing no-PII discipline stays).
- Optionally include the response body's `error` field (the Worker returns structured `{ok:false, error}`).
- Keep at-most-once semantics as-is (duplicate-lead avoidance is a deliberate design choice, documented in the module docstring) — this issue is observability only.

## Context

- Sender: `src/backend/services/operator_intake_service.py` (`submit_operator_intake`, non-2xx branch)
- Receiver contract: `docs/memory/feature-flows/first-time-setup.md` → Flow 3 (hosted intake endpoint)
- Receiver-side fix (endpoint now live): trinity-ops-agent `ff1e8fd`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

AndriiPasternak31:
/claim — picking this up as part of the reliability/security open-ticket push.

Issue #1018, opened 2026-06-02
Potential settlement-ordering issue: paid resource can be released without completed settlement


chenshj73:
# Potential settlement-ordering issue: paid resource can be released without completed settlement

Hi, I noticed a possible payment-flow issue while reviewing the current repository state. This is a conservative report based on the visible code path, and I may be missing deployment-specific guards outside this repository.

Reviewed HEAD: `742bfcc2d41a`

## What I observed

In the visible payment path, I observed:

> Nevermined paid endpoint verifies payment-signature, executes the paid agent task, then settles. Verify does not burn credits, and settlement failure still returns the agent response with status success.

Relevant code locations:

- `src/backend/routers/paid.py:72-80`
- `src/backend/services/nevermined_payment_service.py:86-89`

## Relevant code excerpts

`src/backend/routers/paid.py:72-80`

```text
  72 @router.post("/{agent_name}/chat")
  73 async def paid_chat(agent_name: str, request_body: PaidChatRequest, request: Request):
  74     """Main paid chat endpoint using x402 payment protocol.
  75 
  76     Flow:
  77     1. No payment-signature header → 402 Payment Required
  78     2. Invalid/insufficient token → 403 Forbidden
  79     3. Valid token → verify → execute → settle → return response + receipt
  80     """
```

`src/backend/services/nevermined_payment_service.py:86-89`

```text
  86         """Verify a payment token before processing a request.
  87 
  88         Does NOT burn credits — only checks validity and balance.
  89         Timeout: 15 seconds.
```

## Why this may matter

If the paid resource is returned after verification but before settlement is completed or enforced, a client may receive the protected result without a confirmed transfer.

## Suggested check

Consider making the paid-resource path depend on server-trusted payment requirements and a completed payment state. In particular, re-check recipient/payTo, amount, asset, network, nonce/idempotency, resource binding, and settlement result at the exact point where the protected API/tool/content is released.

## Conservative caveat

I only reviewed the code visible in this repository at the HEAD above. If deployment-specific middleware or an upstream service enforces the missing binding/settlement invariant, this may already be mitigated there.

maxim-ultra:
Hi - I looked at Abilityai/trinity#1018 from the payment-reliability angle. One risk worth isolating is: Public issue mentions payment, x402, receipt, idempotency; Potential settlement-ordering issue: paid resource can be released without completed settlement.

[...]

PR #2418, opened 2026-08-27
feat(settings): operator-intake Settings home (ent#463)

oleksandr-korin:
## Summary

Adds a durable admin-only Settings surface for the operator-intake opt-in
(identified contact record — email + optional company/name/role/use_case),
beside the existing telemetry-sharing panel.

**Why now.** #2385 (merged to dev, awaiting v0.9.5 cut) refuses the first-run
welcome form on any install with a pre-provisioned admin. Without this
Settings home, the intake opt-in silently disappears for the majority of
installs post-v0.9.5. Fixes [ent#463](https://github.com/Abilityai/trinity-enterprise/issues/463).

Both surfaces converge on `submit_operator_intake` — no second intake client,
no forked payload — and the at-most-once marker is preserved across both
producers.

## Design decisions (recorded in ent#463 comment; codified in code)

- **Resubmit = no-op** (AC #4). At-most-once is the intake's designed contract;
  a duplicate lead is worse than a missed one. Panel shows terminal
  "already submitted" state with the timestamp when known, "date unknown"
  otherwise (for pre-ent#463 markers).
- **Opt-out = durable decline, does NOT roll back the marker** (AC #5). The
  record has already been sent; the panel is honest that record deletion
  requires a support request to the hosted endpoint (no local delete
  authority).
- **`OPERATOR_INTAKE_ENABLED` / `DO_NOT_TRACK` win over the Settings control**
  (AC #6). A fresh-install submit attempt while hard-disabled returns 409;
  pure consent flips (no email, or opt-out) still record durable state.
- **Admin AND human-only** (`reject_agent_principal`), matching the
  telemetry-sharing gate (AC #7). An admin-owned agent-scoped key must not
  be able to write consent (trinity-ops-agent#232 class).
- **Distinct audit action** `operator_intake_consent` (AC #8) so a
  Settings-driven consent is distinguishable from a first-run one in the
  audit log. Email + profile values never logged.

## Changes

- **Backend service** (`src/backend/services/operator_intake_service.py`):
  added `is_consent_enabled`, `is_hard_disabled`, `is_already_submitted`,
  `get_status`, `set_consent`, `submit_from_settings`. `submit_operator_intake`
  now also stamps `operator_intake_submitted_at` so the panel can render
  honest state on new submissions (legacy markers report NULL and the UI
  shows "date unknown").
- **Router** (`src/backend/routers/settings.py`): `GET`/`PUT /api/settings/operator-intake`
  (registered before `/{key}` — Invariant #4). Generic catch-all guard added
  for the `operator_intake_*` key family (mirror of the ent#12
  `telemetry_sharing_*` guard).
- **Model** (`src/backend/models.py`): `OperatorIntakeUpdate`.
- **Frontend**: `OperatorIntakePanel.vue` — three states (fresh form /
  terminal already-submitted / hard-disabled banner) + `operatorIntake`
  Pinia store, mounted on the General tab beside `TelemetrySharingPanel`.
- **Tests**: `tests/unit/test_ent463_operator_intake_settings.py` — 14 unit
  tests, all pass locally. Covers state axes, at-most-once, opt-out
  no-rollback, hard-disabled 409, agent-principal rejection, audit no-PII,
  catch-all guard for all four `operator_intake_*` keys.
- **Docs**: `docs/memory/architecture.md` § `operator_intake_service.py`
  updated with the Settings-home addition.

## Acceptance criteria coverage (from ent#463)

- [x] AC 1 — admin can opt in from Settings on any install
- [x] AC 2 — panel shows honest state (not-yet / already / disabled)
- [x] AC 3 — reaches same `operator_intake_service` path (`submit_from_settings`
      delegates to `submit_operator_intake`)
- [x] AC 4 — at-most-once preserved; explicit no-op on resubmit
- [x] AC 5 — opt-out stops future; retraction language points at support
      (no local delete authority)
- [x] AC 6 — `DO_NOT_TRACK` / `OPERATOR_INTAKE_ENABLED` win over Settings (409)
- [x] AC 7 — admin + `reject_agent_principal`
- [x] AC 8 — audit-logged (`operator_intake_consent` action, no PII in details)

## Test plan

- [x] `pytest tests/unit/test_ent463_operator_intake_settings.py` — 14 pass
- [x] Sibling suites unaffected: `test_operator_intake.py` (14 pass),
      `test_ent12_telemetry_sharing.py` (10 pass), `test_setup_operator_profile.py`
      (24 pass), `test_1987_instance_label.py` (28 pass)
- [ ] Manual verification: hit `GET /api/settings/operator-intake` as admin
      → fresh install returns unsubmitted state; submit form; refresh →
      terminal state with timestamp; toggle off → durable decline;
      `OPERATOR_INTAKE_ENABLED=false` restart → hard-disabled banner
- [ ] Manual verification: agent-scoped MCP key attempting `PUT` → 403
- [ ] Manual verification: raw `PUT /api/settings/operator_intake_consent_enabled`
      → 422 with pointer to `/api/settings/operator-intake`

Fixes Abilityai/trinity-enterprise#463


dolho (approved):
## `/review` Report — PR #2418

**Branch**: `feature/ent463-operator-intake-settings` → `dev` (merge-base `d3405c8e`)
**Files**: 8 (+960/-10) · **Scope**: CLEAN · **Plan completion**: 8 AC — 7 done, 1 partial (AC #2)

Good PR. The convergence decision is right, the state model is genuinely three-axis rather than a boolean pretending, and the design decisions are argued in code rather than only in the description. One finding on the state the panel *cannot* tell the truth about — which is the one AC #2 promises.

---

### Informational

**[I1] The Settings surface reports success for a submission that did not land — and the marker makes it terminal** (Confidence: 9/10)

`services/operator_intake_service.py` — the marker is written *before* the POST and deliberately never rolled back:

```python
db.set_setting(_INTAKE_SUBMITTED_KEY, "true")
db.set_setting(KEY_SUBMITTED_AT, submitted_at)
...
resp = await client.post(OPERATOR_INTAKE_URL, json=payload)
# The marker is NOT rolled back (re-send risks a duplicate lead ...)
```

and `submit_from_settings` returns success unconditionally after awaiting it:

```python
await submit_operator_intake(...)
return SettingsIntakeResult.SUBMITTED
```

`submit_operator_intake` never raises and returns nothing, so a 500 from the hosted endpoint, a redirect, a DNS failure or a read timeout all reach the caller as `SUBMITTED`. The router then refreshes status, the panel switches to its **terminal** "already submitted on {date}" state, and the marker forecloses any retry.

So an operator who opts in while the intake endpoint is down gets: a panel confidently asserting the record was submitted, a durable marker preventing a second attempt, and **no record at Ability.ai** — recoverable only by hand-editing `system_settings`. The sole trace is a WARNING in the backend log.

This is not a criticism of the at-most-once design, which is right, and the pre-#1593 behaviour is fine for what it was: a *background, best-effort* POST at first-run, where nobody is watching and nothing claims otherwise. What changes here is the context — an interactive control a human just clicked, whose stated job (AC #2) is to *show honest state*. "Submitted on 2026-08-27" when the POST 500'd is the one state it renders dishonestly, and it is also the one that cannot be corrected afterwards.

The repo already has the shape for this: #1018's `success_unsettled` on the x402 path, which returns 200 with the response but refuses to call an unsettled payment `"success"`. Same medicine here — have `submit_operator_intake` report whether it landed, and add a `SUBMITTED_UNCONFIRMED` result the panel renders as *"recorded — delivery not confirmed; see logs"*. The audit `submit_outcome` then also stops recording `SUBMITTED` for a failed send.

Minimum version if you would rather not touch the plumbing: change the panel's wording so it asserts *consent recorded*, not *delivery*. What must not stay is a terminal, unrecoverable UI claiming an external write happened when the code cannot know that.

**[I2] `get_status()` degrades inconsistently — three fields defensive, two not** (Confidence: 8/10)

```python
"submitted_at": db.get_setting_value(KEY_SUBMITTED_AT, None),
"consent_at": db.get_setting_value(KEY_CONSENT_AT, None),
```

`is_consent_enabled()` and `is_already_submitted()` each wrap their read in `try/except → safe default`, with a comment saying a read failure must degrade. These two do not, so a settings-read failure raises out of `get_status()` and the GET route 500s — the whole panel disappears rather than rendering the three axes it *could* still answer. Either wrap them the same way, or drop the try/except from the siblings so the degradation policy is one decision rather than three.

**[I3] No test drives the failed-submission path** (Confidence: 8/10)

The 14 tests cover the state axes, at-most-once, opt-out no-rollback, the hard-disabled 409, agent-principal rejection and the catch-all guard — but nothing exercises a non-2xx or a raising POST. That is precisely the path I1 is about, and it is why a green suite is compatible with the panel lying. One test asserting the outcome when `client.post` returns 500 (and another when it raises) would pin whichever resolution you choose for I1.

---

### Verified clean

- **Auth (4.3).** Both routes are `assert_admin`, which since #1890 rejects agent *and* connector principals itself — I read the body to confirm rather than trusting the name. The explicit `reject_agent_principal` on the PUT is belt-and-braces, consistent with how the repo describes that pairing. No path-param/dependency mismatch (these are keyless routes).
- **Invariant #4.** `/operator-intake` is registered at lines 330/344, well above `/{key}` at 2861.
- **Catch-all guard.** `key.startswith("operator_intake_")` → 422 in `PUT /{key}`, and it deliberately covers the pre-ent#463 `operator_intake_submitted` key too — the ent#12 mirror, done properly.
- **Credential exposure (4.4).** The email and profile fields never reach the audit `details` (only `has_email`), and the service logs the install-id prefix rather than the address. Checked both.
- **Hard-disable precedence.** `submit_intent` is computed before the 409 so a pure consent flip (no email, or an opt-out) still records durable state while a submit attempt is refused — that distinction is subtle and correct.
- **Frontend.** Store surfaces `e.response.data.detail`, so the 409's message reaches the operator rather than a generic failure. No `v-html`.
- **Tests.** 38 pass across this suite plus `test_operator_intake` and `test_ent12_telemetry_sharing`.

---

### Summary

- **Critical: 0**
- **Informational: 3** — I1 is the one I would take before merge
- **Scope: clean** — 8 files, all of them the surface described; nothing incidental

Not blocking on I1: the consent record itself is correct, the at-most-once semantics it inherits are pre-existing and deliberate, and the fix is additive. But I would not leave a terminal, unrecoverable panel asserting an external write the code cannot confirm — and the wording change alone is a one-line version of the fix.

Worth a `learnings.md` entry alongside whichever fix you pick, because the class is durable and not obvious: **a fire-and-forget background contract reused by an interactive surface inherits its silent-failure semantics, and "nobody is watching" stops being true at exactly that moment.**

**Approving.**

Issue #2421, opened 2026-08-27
bug: subscription register/delete/assign and the auto-switch settings are not audit-logged — a wiped subscription leaves no trace (SEC-001 gap)


webmixgamer:
## Summary

None of the subscription lifecycle endpoints write an `audit_log` entry: `POST /api/subscriptions` (register), `DELETE /api/subscriptions/{id}` (delete — which also cascades every assigned agent to API-key auth), `PUT`/`DELETE /api/subscriptions/agents/{name}` (assign / clear), and the settings toggles under `/api/subscriptions/settings/*`. `src/backend/routers/subscriptions.py` contains no `platform_audit_service.log(...)` call at all, whereas sibling credential/config surfaces (e.g. `routers/agent_config.py`) audit every mutation.

## Context

A subscription token is a credential and its assignment decides which account pays for an agent's turns — SEC-001 lists credential operations and settings among the audited event families. On 2026-08-27 every subscription on a developer instance was deleted by a test suite (companion issue), dropping all agents to API-key auth and disabling auto-switch; the audit log held only the agent-lifecycle rows of an unrelated test agent, so the deletion had to be reconstructed from `system_settings.updated_at` and backups. The same blindness applies to a malicious or accidental deletion on a production instance.

## Acceptance Criteria

- [ ] Register, delete, assign, clear-assignment and the `/settings/*` toggles each write an `audit_log` entry with actor, source, endpoint, target subscription id + name, and (for assign/clear/delete-cascade) the affected agent names
- [ ] Token values never appear in any entry — masked/absent, per the existing credential-audit convention
- [ ] A delete records how many agents were cascaded to API-key auth
- [ ] SEC-001 tests extended to cover the new event actions; `docs/memory/architecture.md` SEC-001 coverage line updated

## Technical Notes

- Precedent: `platform_audit_service.log(...)` calls in `src/backend/routers/agent_config.py`
- The delete cascade lives in `src/backend/db/subscriptions.py::delete_subscription` (returns the cleared-agent count already)

webmixgamer:
Companion: #2420 (the test suite that performed the untraced wipe).

Issue #2420, opened 2026-08-27
bug: tests/test_subscriptions.py wipes every subscription on its target instance and leaves auto-switch OFF — the API integration suite is destructive to a real Trinity


webmixgamer:
## Summary

The API integration suites under `tests/` are destructive to whatever instance they are pointed at, and they point at `http://localhost:8000` by default (`TRINITY_API_URL`, see `tests/conftest.py`) — i.e. a developer's live local Trinity, not only CI's ephemeral stack:

- `tests/test_subscriptions.py` — `test_create_agent_no_subscriptions` does `GET /api/subscriptions` and then `DELETE`s **every** subscription on the instance ("Clear ALL subscriptions first to ensure clean state"), not only the ones the suite created. `delete_subscription` cascades `agent_ownership.subscription_id → NULL`, so every agent on the instance silently falls back to API-key auth.
- `tests/test_subscription_auto_switch.py` — `test_auto_switch_setting_toggle_cycle` cycles the setting `[True, False, True, False]` and `test_…_can_disable` ends on `enabled=false`; neither restores the pre-test value, so the instance is left with auto-switch **OFF**.

## Context

Running these three suites (`test_subscriptions.py`, `test_subscription_auto_switch.py`, `test_subscription_usage.py`) against a developer instance on 2026-08-27 deleted the developer's real subscriptions mid-session, dropped all four agents to API-key auth, and disabled auto-switch — with no trace in the audit log, because subscription mutations are not audited (companion issue filed separately). Recovery required copying the `subscription_credentials` rows back from `/data/backups/`; one subscription registered after the last backup was unrecoverable.

The suites are correct for the disposable stack CI boots and for `/verify-local`'s isolated sibling stack; the hazard is that nothing distinguishes a disposable target from a real one, and the default target is the developer's own instance.

## Acceptance Criteria

- [ ] No test deletes a subscription it did not create: the "clean state" step deletes only subscriptions whose name carries the suite's own prefix (e.g. `test-`), or creates its precondition differently (a dedicated fixture instead of a wipe)
- [ ] Every test that toggles `auto_switch_subscriptions` (or any other platform setting) reads the value first and restores it in a `finally`/fixture teardown, so the target's configuration is unchanged after the run
- [ ] The shared test client refuses to run mutating suites unless `TRINITY_TEST_ALLOW_DESTRUCTIVE=1` is set (skip with a message naming the variable), so pointing the suite at a real instance is an explicit act
[...]

Issue #2419, opened 2026-08-27
`_parse_utilization` reads a >1.0 fraction as an already-percent value — an overage subscription would display and rank as ~1%


webmixgamer:
## Summary

`src/backend/services/subscription_headroom_service.py::_parse_utilization` converts the provider's `anthropic-ratelimit-unified-{5h,7d}-utilization` header with:

```python
return round(v * 100.0, 1) if v <= 1.0 else round(v, 1)
```

The header carries a **fraction** (`0.39` = 39%, verified on live probes). The `else` branch exists to "tolerate an already-percent value", but it makes every fraction above `1.0` — i.e. an **overage** reading such as `1.2` (120%) — come out as `1.2%`. Every consumer of the number then sees a nearly-empty subscription where the provider reported an exhausted one:

- Settings → Subscriptions and the Dashboard pressure tile render ~1%.
- The ent#434 weekly alert classifies it `has_headroom` and never fires — in exactly the case it exists for.
- Once #2409 lands, the auto-switch ranker would rate it the **best** destination in the fleet.

## Context

Surfaced during the #2409 plan review (independent engineering voice). The snapshot also records `overage_status`, so the provider evidently can report usage past the limit; whether it expresses that as a fraction `> 1.0` is **unconfirmed** — no overage header sample has been captured yet. The fix should be made against a real sample, not a guess.

## Acceptance Criteria

- [ ] Capture one real overage header set (or the provider's documented format) and pin the expected shape in a test
- [ ] `_parse_utilization` maps a fraction `> 1.0` to `> 100%` (never to a single-digit percent); if an already-percent value must still be tolerated, the disambiguation rule is explicit and tested at the boundary
- [ ] Non-finite inputs (`nan`, `inf`) parse to `None`, never to a number
- [ ] Existing #471 / ent#433 / ent#434 tests unchanged

## Technical Notes

- `src/backend/services/subscription_headroom_service.py` — `_parse_utilization`, `parse_unified_headers`
- Related: #2409 (ranker consumes the same figure), abilityai/trinity-enterprise#434 (alert threshold compare)

webmixgamer:
Fix is up as #2864. Three rulings from the plan gate, recorded here so the acceptance criteria read correctly against the PR:

[...]

PR #2422, opened 2026-08-27
fix(subscriptions): auto-switch ranks alternatives by cached headroom, never by load alone (#2409)

### LINKED ISSUE #2409: auto-switch picks an alternative subscription without checking headroom — it can move agents onto a nearly-exhausted one

webmixgamer:
## Summary
- `select_best_alternative_subscription` returned the **first** survivor of the 2h failure filter in `agent_count ASC` order and read no headroom — SUB-003 could move an agent onto a subscription at 99% of its weekly window, and an *unused dead-token* subscription (no agents ⇒ no failure rows) sorted **first**.
- Now: **filter in the db, rank in the service, never a probe.** The db lists survivors (kind-blind 2h filter unchanged and first, #444/#2352, `agent_count ASC, name ASC`); the service ranks them over the cached provider snapshot (one `MGET`) furthest-from-the-nearest-wall first (the fuller of the 5h/7d windows — the #792 retry lands on the destination immediately), in 10-point bands so load still spreads a storm; a **fresh** provider refusal is dropped; anything unusable sorts in today's order; any failure of the ranking half falls back to today's pick **with a warning**.
- `classify_headroom` (ent#434) and the ranker share one usability gate (`headroom_reading`) — verdicts byte-identical, pinned by a differential test against a frozen copy. New-agent auto-assign (#74) rides the same ranker. The switch now records **why** (`destination_headroom` + one notification clause).
- Approved deviations from the literal AC, recorded on the issue: nearest-wall key instead of 7d-only; fresh refusals filtered instead of ranked last.

## Changes
- `src/backend/services/subscription_headroom_service.py` — gate, MGET reader, threshold-free ranker, `MAX_READING_AGE_SECONDS` (owned here now); `classify_headroom` becomes policy over the gate
- `src/backend/services/subscription_auto_switch.py` — service-layer selector (`asyncio.to_thread` under the agent lock), `destination_headroom` on activity / notification / result
- `src/backend/services/subscription_service.py` — `select_subscription_for_new_agent`
- `src/backend/db/subscriptions.py` + `database.py` — `list_viable_alternative_subscriptions` / `list_assignable_subscriptions` (filter only); first-match selectors retired
- `src/backend/services/agent_service/crud.py` — call site; `subscription_headroom_alerts.py` — constant re-export + docstring
- Tests: new `tests/unit/test_2409_headroom_ranked_switch.py`; pingpong / 2352 / concurrency / 1484 / 1759 adapted to the list form (assertions kept)
- Docs: `architecture.md`, `subscription-auto-switch.md` (+ management, usage-tracking), requirements §20.4, `learnings.md` (2 entries), CSO diff report

## Test Plan
- [x] New suite: `pytest tests/unit/test_2409_headroom_ranked_switch.py` — 83 passed; **80/81 fail on the unmodified source**
- [x] Full `tests/unit`: 12,718 passed / 30 skipped / 1 pre-existing failure (`test_1920`, private submodule, untouched)
- [x] API integration (`test_subscription_auto_switch`, `test_subscriptions`, `test_subscription_usage`): 36 passed
- [x] Live: a real switch chose the 18%/9% subscription over the 0-agent 88%/60% one; every-survivor-refused → no switch + WARNING; no snapshot → today's order
- [x] `/review` clean (informational findings fixed in-review); `/cso --diff` no findings
- Follow-ups filed while testing: #2419 (parser overage), #2420 (destructive integration suite), #2421 (subscription audit gap)

Fixes #2409


dolho (approved):
## `/review` Report — PR #2422

**Branch**: `feature/2409-headroom-ranked-switch` → `dev` (merge-base `908f8713`, the ent#434 PR)
**Files**: 23 (+1768/-169) · **Scope**: CLEAN · **Plan completion**: AC met, with 2 deviations declared and argued

I flagged this bug in the #2410 review — *"`select_best_alternative_subscription` reads no headroom at all and will relocate agents onto a 99% subscription today"* — so I read this one looking for whether the fix actually closes both halves, and whether the new ranking can make things worse than the load-only order it replaces. It does close both, and it cannot.

**No critical findings. One wording nit and one low-confidence appendix item.**

---

### Verified against the source, not the description

**Both halves of the reported bug are closed.** The 99%-destination half is the ranker. The subtler half — *"an unused dead-token subscription (no agents ⇒ no failure rows) sorted first"* — is closed by `HeadroomReading.refusing`, which is `provider_status != "ok"` **or** any blocked window. `invalid_token` therefore counts as refusing, so a fresh dead-token candidate is dropped rather than sorted to the front by its zero agent count. That is the half that would have been easy to miss.

**The failure filter is still first and still kind-blind.** `list_viable_alternative_subscriptions` runs before anything headroom-related, so a candidate that just failed is never even read. Narrowing that filter is what would have re-opened #444's ping-pong, and it is untouched — the pingpong and #2352 suites still pass with their assertions intact.

**Nothing in the selection path probes.** `cached_headroom_readings` is a single `MGET`; I grepped its body for `_locked_probe` / `_probe_and_store` / `await` and got zero. Auto-switch runs on a failure, which is exactly when spending provider quota to decide where to go would be worst.

**The fail-open is loud and correctly tiered.** Any exception in the ranking half falls back to the db's own order — the literal pre-#2409 pick — with a WARNING, and the surfaced tier becomes `SELECTION_UNRANKED` rather than a fabricated reading. The comment names the right class (learnings 2026-08-12: *an inert ranker must not look like a working one*). Same shape on the `#74` new-agent path, which was the other consumer and is easy to forget.

**The `importlib` note is a real point, not defensive noise.** Service tests stub `services` as a bare module at load; `import_module` answers from `sys.modules` rather than from a package attribute a previous test may have left behind. That is a genuine hazard in this codebase and worth the line explaining it.

### The two declared deviations both hold up

**Nearest-wall instead of 7d-only.** The argument is right and I would not have accepted 7d-only: the #792 retry re-issues the failed turn on the destination *immediately*, so a 7d-20% / 5h-98% subscription fails within the minute. And the implementation detail matters — using `max` rather than a lexicographic 7d-then-5h key, because with 1-decimal utilization the second term never gets reached. Ranking by the wall you hit first is the correct reading of "furthest from the wall".

The 5h figure is only consulted while `age <= FAST_WINDOW_FRESHNESS_SECONDS`, separately from the overall gate. That distinction is easy to overlook and is correct: a 5h window decays far faster than a weekly one, so a stale 5h figure is worse than no 5h figure.

**Fresh refusals dropped rather than ranked last.** Ranking them last still lets them be chosen when everything else is refused too, and the consequence is a guaranteed failed turn plus a fresh failure row — with several such subscriptions the agent walks through all of them, which is #444's class arriving by a new route. Dropping is right. And `refusing` is gated on *freshness* (`age <= REFUSAL_FRESHNESS_SECONDS`), so a stale refusal degrades to UNKNOWN and is ranked rather than dropped — a stale refusal is not a refusal.

### The parity claim is the strongest part

"Verdicts byte-identical" is the sort of claim I would normally push back on, because rewriting a classifier as policy over a new gate is exactly where a subtle change hides. Here it is a genuine differential test: a **verbatim frozen copy** of the ent#434 function with constants inlined, as an independent oracle, over the full product of 6 ages × 6 statuses × 10 weekly shapes × 2 five-hour shapes × 4 thresholds × 2 max-ages — and it asserts the case count, so a quietly shrunken matrix cannot pass.

The interesting case is preserved correctly: the gate now admits `invalid_token` (the probe *did* answer), and the classifier's policy layer maps it back to `unassessable` via `provider_status != "ok"`. Old behaviour, new structure, and the matrix proves it rather than the comment asserting it.

### Ordering edges I checked and found right

- Missing "other" window sorts with `math.inf`, not `100` — because an overage window legitimately reads past 100%, so `100` would rank a real overage *after* an unknown.
- The UNKNOWN tier key is `(1, 0, agents, 0.0, 0.0, name)` — exactly today's `agent_count ASC, name ASC` — and the sort is stable, so a candidate set with no readings behaves precisely as it does today.
- Within a band the agent count still decides, so the storm-spreading property of the old order survives banding.

### Behaviour when nothing is choosable

All alternatives freshly refused → no switch, WARNING, `None` — which is the same shape as "no alternatives at all", a path the caller already handles. Leaving the agent on a failing subscription is better than moving it onto one that provably cannot serve, and ent#434's fleet alert covers the operator-visibility side of that state.

Nice touch: "no reading for ANY alternative" logs INFO when ambient refresh is on and WARNING when it is off — because with refresh off, no readings is the expected consequence of an operator's own choice, not a fault.

---

### Nit

**`rank_subscriptions`' docstring contradicts itself within the paragraph.**

> Never fewer than the input minus fresh refusals — ranking IMPROVES a choice, it never prevents one.

When every survivor is freshly refused it *does* prevent one, and the next sentence explains why that is deliberate. The clause is self-correcting for a careful reader and misleading for a skimming one. "…it never prevents one **except when every candidate is refusing**" costs four words.

### Low confidence (appendix)

**`_finite_pct` is a one-sided sanitiser** (Confidence: 4/10). It rejects `NaN`/`inf` but not sign, so a negative utilization would band to `-1` and sort **best** — ranked ahead of a genuinely empty subscription. This needs malformed provider data to reach (a parser bug upstream, not an input anyone controls), so it is theoretical; I raise it only because the function already exists to sanitise this exact field and the missing lower bound is an asymmetry in it, not an absent guard.

---

### Summary

- **Critical: 0**
- **Informational: 1 nit + 1 low-confidence**
- **Scope: clean** — 23 files, all in the described surface; the adapted sibling suites keep their assertions rather than being loosened

Verified locally: 203 pass across the new suite plus ent#434 / pingpong / #2352 / concurrency; 396 pass across every subscription and headroom suite. I spot-checked the "fails on the unmodified source" claim by neutering `rank_subscriptions` to a passthrough — 16 of the new tests go red on that alone, so the suite is genuinely coupled to the behaviour rather than to its shape.

**Approving.**

PR #2427, opened 2026-08-28
fix(workspace): the sync portal turn never carried its session, so report-back could not fire (#2426)

### LINKED ISSUE #2426: Workspace report-back can't fire on the synchronous chat path — the pre-created row has no session binding
## Summary

ent#457 gave the Workspace a report-back: an agent that delegates work during a chat turn gets the completion posted into that thread. It cannot fire on the **synchronous** chat path, because the parent execution never receives the session binding the report needs.

`report_completion` gates on `if not source_channel_chat_id`, and on that path the field is NULL.

## Measured

Eight portal executions on a dev instance, split cleanly by path:

```
portal rows total: 8
missing chat_id : 5

07:55 → 09:04   chat=7d27744d…   ← browser, streaming path (start_portal_turn)
09:06 → 09:09   chat=NULL        ← POST .../chat, synchronous path
```

A child delegated from one of the NULL rows inherits `source_channel='portal'` with no `chat_id`, so its terminal is suppressed at the gate.

## Root cause — two correct changes that collide

ent#457 passes the binding into the turn:

```python
# client_portal/service.py:1532
source_channel_chat_id=session_id,
source_channel_client=email,
```

Those reach `execute_task`, which persists them **only when it creates the row**:

```python
# services/task_execution_service.py:1095
if not execution_id:
    execution = db.create_task_execution(
        ...,
        source_channel_chat_id=source_channel_chat_id,
        source_channel_client=source_channel_client,
    )
```

But ent#365 pre-creates the row and hands the id down, so `execute_task` **adopts** it and skips that branch entirely. And the pre-create stamps only the surface:

```python
# client_portal/service.py:1271
core_db.create_task_execution(
    agent_name=agent_name,
    message=message,
    triggered_by="public",
    source_user_email=email,
    subscription_id=subscription_id,
    source_channel=PORTAL_SOURCE_CHANNEL,   # ← no chat_id, no client
)
```

Its own docstring states the invariant it breaks:

> Mirrors `start_portal_turn`'s creation exactly — same trigger, same `source_channel` stamp — so the two paths produce **indistinguishable rows** and a report published from either can be joined back to its chat.

They are distinguishable, and the sync one cannot be joined back to its chat. `start_portal_turn` (the streaming path) stamps both fields, which is why the browser path works and this one does not.

## Why it went unnoticed

Neither feature is wrong alone, and each has passing tests:

- ent#457's tests drive `portal_chat` with the engine mocked, so they assert the kwargs are *passed* — which they are.
- ent#365's pre-create is covered for the property it was built for (no orphan `running` row), which still holds.

Nothing asserts the **persisted row**, which is the only place the two meet. Same shape as `test_ent457_portal_turn_kwargs.py`'s own stated lesson: *"the only place the two signatures meet is here."*

## Reproduce

```bash
curl -s -X POST -H "Authorization: Bearer $JWT" -H "Content-Type: application/json" \
  -d '{"message":"hello"}' \
  http://localhost:8000/api/enterprise/client-portal/agents/<agent>/chat

docker exec trinity-postgres psql -U trinity -d trinity -tAc \
 "select id, source_channel, source_channel_chat_id, source_channel_client
  from schedule_executions where source_channel='portal' order by started_at desc limit 1;"
# → portal | NULL | NULL
```

Drive the same agent from the Workspace UI instead and the row carries both.

## Fix

Stamp the two fields in `_precreate_sync_execution`, making its docstring's claim true. The session id and email are both in scope at the call site (`service.py:1510`).

Rejected alternative: teaching `execute_task` to UPDATE an adopted row. That widens a hot path used by every trigger to repair one caller's omission, and the adopted row is the caller's to get right.

## Notes

- Related: ent#457 (report-back), ent#365 (pre-create), ent#265 (inheritance), #2157 (the `source_channel` stamp).
- Not a regression in either feature — the gap is between them.
- Client-facing impact is a missing notification, not a wrong one: a delegated job finishes and nobody is told, which is the exact scenario ent#457 was written for.

dolho:
Journey Impact: none: repairs an existing surface's report-back on one code path — no user-facing promise is added or extended.

Closes #2426

## The bug

ent#457 gave the Workspace a report-back: an agent that delegates during a chat turn gets the completion posted into that thread. **It cannot fire on the synchronous chat path** — the parent execution never receives the session binding, so `report_completion` drops every report at its own `if not source_channel_chat_id` gate.

Measured on a dev instance, split cleanly by path:

```
portal rows total: 8
missing chat_id : 5

07:55 → 09:04   chat=7d27744d…   browser, streaming path
09:06 → 09:09   chat=NULL        POST .../chat, synchronous path
```

## Two correct changes that collide

ent#457 passes the binding down, and `execute_task` persists it — but only when it **creates** the row:

```python
# task_execution_service.py:1095
if not execution_id:
    execution = db.create_task_execution(..., source_channel_chat_id=..., source_channel_client=...)
```

ent#365's `_precreate_sync_execution` has already created the row and handed the id over, so `execute_task` **adopts** it and skips that branch. And the pre-create stamped only the surface:

```python
core_db.create_task_execution(..., source_channel=PORTAL_SOURCE_CHANNEL)   # no chat_id, no client
```

Its own docstring named the invariant it broke:

> Mirrors `start_portal_turn`'s creation exactly … so the two paths produce **indistinguishable rows** and a report published from either can be joined back to its chat.

And `start_portal_turn`'s sibling comment says *"both creation sites or the stamp is a coin flip depending on which path made the row."* ent#457 covered the two sites that existed when it was written; ent#365 had added a third.

## The fix

Stamp both fields in the pre-create. `session_id` is a **required** parameter, not optional — the value is in scope at the only call site, and a default would let a future caller silently reintroduce the inert row.

**Rejected:** teaching `execute_task` to UPDATE an adopted row. That widens a hot path used by every trigger to repair one caller's omission; the adopted row is the caller's to get right.

## Two guards that were red on `dev`

`backend-unit-test` is **failing on `dev` right now**. Both failures are in this feature area and both are guards that had gone inert, so they're repaired here rather than left for the next PR to trip over. Frontend-only PRs stay green because the `changes` job path-filters the backend suite away — which is why nobody saw it.

**`test_both_portal_row_creation_sites_name_the_chat`** asserted a literal census, `== 2` sites. It went red the moment the third site appeared — *the guard working* — and the bug it names shipped anyway. It now asserts the rule instead of the count:

```python
assert destination == surface, "…a row created by the unstamped one can never be joined back to its chat"
```

**`test_portal_turn_kwargs_bind_against_execute_task`** parsed `portal_chat` for a literal `run_resumable_turn(...)` call. That call had moved into `_run_sync_turn_and_clear_marker`, where it is `run_resumable_turn(**kwargs)` — a splat, which names nothing — so the walk found no keywords and the guard asserted itself dead. It now reads the keywords where they're actually named (the wrapper's call site), scans both entry names, and subtracts the wrapper's own consumed parameters.

Neither rewrite loses coverage; both now fail for the reason their docstring gives rather than because a number or a call site moved. Say so if you disagree — I edited guards I didn't write.

## Why the bug survived its own tests

Both features have passing tests:

- ent#457's mock the engine and assert the kwargs are **passed** — they are.
- ent#365's assert no orphan `running` row — still true.

Nothing asserted the **persisted row**, the only place the two meet. That is verbatim the lesson `test_ent457_portal_turn_kwargs.py` states about itself: *"the only place the two signatures meet is here."* The new suite asserts at that layer and adds a derived parity check, so a fourth channel field added to one writer and forgotten in another fails here rather than shipping as another silently-inert report path.

## Verification

```
pytest -k "portal or ent457 or ent365 or 2426"
  before this branch:   2 failed, 400 passed     ← both pre-existing on dev
  after:              402 passed, 1 skipped
```

Mutation-checked:

| Mutation | Result |
|---|---|
| Remove the new stamp | 4 failed (new suite + the repaired census guard) |
| Feed `execute_task` an unknown kwarg | repaired binding guard fires |

## Impact

Client-facing effect is a **missing** notification, not a wrong one: a delegated job finishes and nobody is told — precisely the scenario ent#457 was written for. No schema change, no migration, no API change.

### DIFF
--- a/src/backend/client_portal/service.py
+++ b/src/backend/client_portal/service.py
@@ -1244,12 +1244,30 @@ async def _run_sync_turn_and_clear_marker(owns_marker: bool, marker_session_id: 
             clear_turn_inflight(marker_session_id, marker_execution_id)
 
 
-def _precreate_sync_execution(agent_name: str, message: str, email: str) -> str | None:
+def _precreate_sync_execution(
+    agent_name: str, message: str, email: str, session_id: str,
+) -> str | None:
     """Create the execution row for a synchronous portal turn (ent#365 review).
 
     Mirrors `start_portal_turn`'s creation exactly — same trigger, same
-    `source_channel` stamp — so the two paths produce indistinguishable rows and
-    a report published from either can be joined back to its chat.
+    `source_channel` stamp, same destination — so the two paths produce
+    indistinguishable rows and a report published from either can be joined back
+    to its chat.
+
+    #2426: that claim used to be false, and it read as true. This site stamped
+    the SURFACE and not the DESTINATION, so a synchronous turn landed with
+    `source_channel='portal'` and a NULL `source_channel_chat_id`. ent#457 does
+    pass the binding down, but `execute_task` persists it only inside
+    `if not execution_id:` — and this function has already made the row and
+    handed the id over, so that branch never runs. A child delegated during such
+    a turn inherited a portal context with nowhere to go, and
+    `report_completion` dropped it at its own `if not source_channel_chat_id`
+    gate. Measured before the fix: 5 of 8 portal rows NULL, split exactly by
+    path — the browser (streaming) stamped, `POST .../chat` did not.
+
+    `session_id` is a required parameter rather than an optional one: the value
+    is in scope at the only call site, and a default would let a future caller
+    reintroduce the silent-inert row this fixes.
 
     Fail-soft to None: this exists so an addressed report finds its session, and
     a turn must not be refused because that bookkeeping could not be done. ONLY
@@ -1275,6 +1293,13 @@ def _precreate_sync_execution(agent_name: str, message: str, email: str) -> str 
             source_user_email=email,
             subscription_id=subscription_id,
             source_channel=PORTAL_SOURCE_CHANNEL,
+            # #2426: the destination, not just the surface. The sibling comment
+            # in `start_portal_turn` says "both creation sites or the stamp is a
+            # coin flip depending on which path made the row" — ent#457 covered
+            # the two sites that existed when it was written; ent#365 had added
+            # this third one.
+            source_channel_chat_id=session_id,
+            source_channel_client=email,
         )
         return execution.id if execution else None
     except Exception:  # noqa: BLE001
@@ -1507,7 +1532,7 @@ async def portal_chat(agent_name: str, message: str, email: str,
         # than this one growing its own turn machinery.
         owns_marker = False
         if not execution_id:
-            execution_id = _precreate_sync_execution(agent_name, message, email)
+            execution_id = _precreate_sync_execution(agent_name, message, email, session_id)
             if execution_id:
                 mark_turn_inflight(session_id, execution_id, turn_timeout + 60)
                 owns_marker = True
--- a/tests/unit/test_2426_portal_sync_session_binding.py
+++ b/tests/unit/test_2426_portal_sync_session_binding.py
@@ -1,0 +1,184 @@
+"""#2426 — the synchronous Workspace turn must carry its session binding.
+
+ent#457 gave the Workspace a report-back: an agent that delegates during a chat
+turn gets the completion posted into that thread. `report_completion` gates on
+``if not source_channel_chat_id``, so the parent row MUST carry the session or
+the feature is silently inert.
+
+It was inert on the synchronous path. ent#457 passes the binding to
+`execute_task`, which persists it only inside ``if not execution_id:`` — and
+ent#365's `_precreate_sync_execution` had already created the row and handed the
+id down, so that branch never ran, and the pre-create stamped only
+`source_channel`. Measured on a dev instance: 5 of 8 portal rows had a NULL
+`chat_id`, split exactly by path (browser/streaming stamped, `POST /chat` not).
+
+WHY THIS FILE EXISTS AT THIS LAYER. Neither feature is wrong alone and both have
+passing tests — ent#457's mock the engine and assert the kwargs are *passed*
+(they are), ent#365's assert no orphan `running` row (still true). Nothing
+asserted the PERSISTED ROW, which is the only place the two meet. That is the
+same lesson `test_ent457_portal_turn_kwargs.py` states about itself: "the only
+place the two signatures meet is here."
+"""
+from __future__ import annotations
+
+import ast
+import inspect
+
+import pytest
+
+pytestmark = pytest.mark.unit
+
+
+@pytest.fixture()
+def portal_service():
+    from client_portal import service as m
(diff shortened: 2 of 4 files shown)


obasilakis (approved):
Approving. No code blockers on any axis, and I confirmed the headline claim empirically rather than by reading.

## Verified

Exported the trees read-only via `git archive` and ran the suites:

| tree | selection | result |
|---|---|---|
| PR base `8d9af18b` | the two files | 2 failed, 27 passed |
| PR base `8d9af18b` | `-k "portal or ent457 or ent365"` | 2 failed, 395 passed |
| `origin/dev` `8df57509` | same + `2426` | 2 failed, 395 passed |
| base + this diff | same | 402 passed |

The failures are exactly `assert 3 == 2` and "portal_chat no longer calls run_resumable_turn by that name / assert set()". `dev` is red right now, and this PR is what makes it green — the claimed 402 matches exactly.

Non-vacuity checks out too: applying only the test half of the diff to the unfixed base leaves the census test and all four new `test_2426` tests red, so it is green for the right reason. The wrapper-parameter subtraction is not over-broad — `wrapper ∩ run_resumable_turn` and `wrapper ∩ execute_task` are both empty, the masked set is empty, and `passed - rrt` and `passed - (rrt | wrapper)` are identical. Post-change, `forwarded` is 8 kwargs including `source_channel_client`, i.e. the exact ent#457 regression, all accepted by `execute_task`. Neither rewritten guard can silently pass over a genuine defect.

## Comments (non-blocking)

- The census test's premise is wrong, not merely loosely worded. `surface = 3` counts `service.py:1523/1548` — `portal_chat`'s kwarg forwarded to `_run_sync_turn_and_clear_marker` — which is not a row-creation site, while the actual third creation site (`task_execution_service.py:1103`) lives in another file and is never counted. So `test_both_portal_row_creation_sites_name_the_chat` counts two creation sites plus one forwarding site. The pairwise equality it enforces is a real property; the name and comment assert something the test does not check.
- `test_portal_turn_kwargs_bind_against_execute_task` was dead, not wrong — it passes at base with only the test rewrite applied, so its repair is a resurrection rather than a second live defect fixed. The body's "two guards that were red" is accurate but shouldn't be read as two bugs.
- `== 2` → `>= 2` trades away a tripwire. The original census loudly announced "a new creation site appeared, go audit it", which is exactly what it did when ent#365 landed — nobody acted, and the bug shipped. The replacement enforces the pairing invariant, which is net stronger, but the "count moved" signal is gone.
- The string-literal census stays brittle: a future site stamping from a variable not named `session_id` goes red spuriously. Fail-loud direction, so acceptable.

Note for sequencing: #2430 edits the same `client_portal/service.py` regions (`_precreate_sync_execution`, `start_portal_turn`). Land this one first.

Issue #2281, opened 2026-08-18
feat: DigitalOcean Marketplace 1-Click Droplet listing for Trinity (vendor application + Packer snapshot, AI Agents category)


vybe:
## Summary

List Trinity as a **DigitalOcean Marketplace 1-Click Droplet** — vendor application, Packer-built snapshot via DigitalOcean's `marketplace-partners` tooling, listing copy, and a referral link — placed in the Marketplace's **AI Agents** category.

## Context

DigitalOcean is the only major cloud with an open, documented vendor path for OSS projects (vendor form → Vendor Portal → snapshot review), it is free to list, and it already carries Docker-socket-mounting 1-clicks (Coolify, CapRover, Dokku, Cloudron, Easypanel) plus OpenClaw as a Staff Pick in a new "AI Agents" category — so Trinity's runtime shape (sibling agent containers over the host Docker socket) is a known quantity there. Ghost's arrangement (DigitalOcean named "official hosting partner", $100-credit signup link in the docs) is the co-marketing model to ask for.

Depends on the pull-only images / unattended first-boot work (#2280) — a snapshot that builds the agent base image on first boot does not meet the one-click bar.

## Acceptance Criteria

> Revised 2026-09-10 against `digitalocean/marketplace-partners@b708788` and
> `droplet-1-clicks@master` — see the audit comment. Items marked ✅ are verified in #2522.

**Image**
- [x] Packer template (per `github.com/digitalocean/marketplace-partners`) builds an Ubuntu 24.04 snapshot, passes `cleanup.sh` + `img_check.sh` with 0 FAILED, with a first-boot script that starts Trinity unattended — verified on `trinity-v0-9-5-rc2-20260903` (#2522).
- [x] Built on the **$6 build droplet** (`s-1vcpu-1gb`), per DO's "will ensure that all Droplet types will be available" guidance — currently `s-2vcpu-4gb`, which excludes every plan under 80 GB disk. Record the measured snapshot size here; if it does not fit 25 GB, use the smallest size that does and say which. **Measured (2026-09-11):** the `v0.9.5-rc3` snapshot built on `s-1vcpu-1gb` is 9.3 GB with a 25 GB minimum disk, so it fits every plan.
- [x] The build installs security updates (`apt full-upgrade`, as DO's reference template does). `img_check` scores pending security updates as **FAIL**, so without this a rebuild fails on any day Ubuntu has published one that `unattended-upgrades` has not yet applied.
[...]

PR #2431, opened 2026-08-28
feat(hosting): install provenance + a first-run hardening guide gated on it (#2380)

### LINKED ISSUE #2380: feat: record install provenance (DO Marketplace) + first-run HTTPS/VPN hardening guide
## Summary

Record **how a Trinity instance was installed** (DigitalOcean Marketplace 1-Click, Vultr Marketplace, script, manual), surface that provenance to the frontend, and — on a marketplace-installed instance only — show a first-run **hardening guide** covering HTTPS (bring a domain) and non-exposed public access (Cloudflare Tunnel).

The guide must be invisible on every install that did not come from a marketplace, including the existing managed fleet.

## Context

Child of #2332 (one-click hosted install). Sits on top of #2280 (prebuilt images + pull-only hosted compose + unattended first boot) and beside #2281 (DO Marketplace listing).

### Why provenance is needed rather than inferring from the environment

A marketplace instance is the only kind where Trinity knows, at boot, that:

- there is a **public IPv4 and no domain**,
- the operator has done **zero** network configuration,
- nobody has read `docs/DEPLOYMENT.md`.

None of that is true for the managed fleet, and there is no environmental signal that separates the two reliably. Measured on the current fleet (16 instance dirs in `trinity-ops-agent/instances/`): **every instance has `FRONTEND_PORT=80`, no `DOMAIN`, no `HTTPS_ENABLED`, and an `SSH_HOST` on `100.x.y.z`** — Tailscale CGNAT (`100.64.0.0/10`). That includes the two marked `ACCESS_MODE=public` (`us14`, `us15`); only `ability-services` has tunnel ports set.

So a naive "no TLS configured → show the hardening guide" rule would fire permanently on all of them — paying clients included (`paynearme`, `paradigmlife`, `tapestry`, `runproper`, `luminous`, `ceoci`, …). Those instances are already correct: HTTP over a WireGuard tunnel is encrypted transport. **Provenance, not TLS state, is the correct gate.**

### What the TLS situation actually is (settled 2026-08-21, both facts verified first-hand)

1. **Let's Encrypt IP-address certificates are generally available** (GA 2026-01-15). ACME `shortlived` profile, **~6-day validity**, `http-01` / `tls-alpn-01` only — no DNS-01 for IP certs.
   Source: https://letsencrypt.org/2025/07/01/issuing-our-first-ip-address-certificate/
2. **DigitalOcean builds its own 1-Clicks this way.** `digitalocean/droplet-1-clicks`, file `.github/agents/1-click-agent.md`, rule 10, verbatim:

   > If the installed software has an http interface, the 1-click should install caddy and letsencrypt and use on of the new IP certs like this:
   > ```
   > PLACEHOLDER_DOMAIN {
   >     tls { issuer acme { dir https://acme-v02.api.letsencrypt.org/directory
   >                         profile shortlived } }
   >     reverse_proxy localhost:18789
   > }
   > ```

Consequence: a marketplace droplet **can** come up on real, browser-trusted HTTPS at its bare IP with zero user input. That removes the "plaintext first login" problem, but it is **not a finished posture** — an IP cert renews every ~6 days, the URL is an unmemorable IP, and the instance is still exposed to the public internet. That gap is what this guide addresses: it is an *upgrade* prompt, not a *warning* about something broken.

### Decision recorded: DigitalOcean 1-Clicks have no deploy-time input form

Verified against `digitalocean/marketplace-partners`: the only vendor-configurable prompt at Droplet creation is the optional Managed Database checkbox. There is **no field for a domain or any app variable** — unlike Vultr, which does have real app variables (15-char snake_case names, auto-generated password variables, metadata API at `http://169.254.169.254/v1/internal/app-{name}` with header `METADATA-TOKEN: vultr`).

So on DO the domain **cannot** be a setup input. It has to be collected after first login, from inside Trinity — which is precisely what this guide is for.

### Decision recorded: Cloudflare Tunnel replaces the VPN path (2026-09-01)

The second path was originally "put it behind a VPN (Tailscale)". That is withdrawn, for a reason that only became clear once the port surface was mapped: **a VPN does not merely fail to add functionality, it removes most of the channel surface**, because those integrations call *us*.

| feature | behind a VPN | behind a Cloudflare Tunnel |
|---|---|---|
| Telegram | broken — `setWebhook` needs a public URL | works |
| WhatsApp / Twilio | broken — webhook callbacks | works |
| VoIP Media Streams | broken — inbound WSS from Twilio | works |
| Public agent links, x402, inbound A2A | broken | works |
| Webhook schedule triggers | broken | works |
| Slack | works (Socket Mode is an outbound WebSocket) | works |
| Web UI, Workspace, MCP | tailnet members only | works |

For a personal instance a VPN is a fine deployment mode. For a marketplace user who may want a Telegram bot working tomorrow, offering it as a co-equal first-run path is actively harmful — and it cannot be automated anyway, since it requires a Tailscale account and a pre-generated auth key before the droplet exists.

A tunnel reaches the same security outcome — **nothing listening on the public interface** — while keeping every inbound integration working, and Trinity already carries the moving parts.

**Honest limitation, recorded so it is not rediscovered in review:** the setup is not fully collectable in-app. `TUNNEL_TOKEN` is a credential that must reach `.env`, and the tunnel then starts under a compose profile — Trinity runs in a container with no host privileges and cannot restart its own stack with a new profile. So this path's card is *guidance plus a place to record the token*, and the last step is an operator action on the host. Do not design the card as if it can finish the job; say what remains. (Same constraint that ruled out a "connect to Tailscale" button.)

VPN is not deleted from the world — it stays a documented deployment mode in `docs/DEPLOYMENT.md`, with the channel tradeoff above stated. It is removed from the first-run card.

## Acceptance Criteria

- [ ] Install provenance is persisted at first boot as a durable, machine-readable value (e.g. `install_source` in `system_settings`, seeded from a marker the provisioning script writes). Values at minimum: `do-marketplace`, `vultr-marketplace`, `script`, `unknown`. Absent/unreadable resolves to `unknown`, never to a marketplace value.
- [ ] Provenance is surfaced to the frontend over an existing channel — `GET /api/settings/feature-flags` is the established home for UI-gating flags (`session_tab_enabled`, `voice_available`, `canary_enabled`, …). **No new endpoint for UI state.** _Amended 2026-09-11 (@obasilakis):_ this still holds for everything the browser reads. `GET /api/public/tls-allowed` (#2707) is not a UI channel — it is the gate Caddy calls during a TLS handshake before issuing a certificate for a saved domain, unauthenticated because Caddy holds no credential, and fail-closed to the exact configured host. Adding a domain was otherwise a root shell on the host.
- [ ] A first-run hardening guide renders **only** on a provenance that is known to be a public cloud VM at a bare IP. On `unknown` / `script` it renders nothing at all — verified against the existing fleet's shape (no `DOMAIN`, no `HTTPS_ENABLED`, `100.x` access) so no managed instance ever sees it. _Amended 2026-09-11 (@obasilakis):_ the gate is `hardening_guide_eligible` — the marketplace sources **plus `do-script`**, the provenance a droplet installed by following the DigitalOcean deploy doc records. Deliberately a separate set from `marketplace_install`, which keeps answering only "did this come from a vendor listing". Still provenance, never TLS state.
- [ ] The guide covers two paths, and is explicit that the second builds on the first rather than replacing it:
  - **HTTPS with a real name** — point a domain at the droplet IP, set it in Trinity, Caddy obtains a normal 90-day certificate and the ~6-day IP cert stops being load-bearing.
  - **Serve it without exposing it — Cloudflare Tunnel.** `cloudflared` connects OUTBOUND to Cloudflare and traffic arrives through that connection, so the droplet can stop listening on the public interface entirely while remaining publicly reachable. Trinity already ships this: a `cloudflared` service behind `--profile tunnel`, driven by `TUNNEL_TOKEN` (`.env.example` §CLOUDFLARE TUNNEL). The guide's job is to walk the operator to a token and record it — not to invent a mechanism.
- [ ] The Cloudflare path states its prerequisite plainly: it needs a domain **on Cloudflare**. It is therefore not an alternative to the first path, it is a different way to serve it. A card that presents them as an either/or is wrong.
- [ ] The guide is dismissible and self-retiring: once a domain is configured, it does not come back.
- [ ] Honest state, never a fake claim: the guide reports what is actually true of this instance right now (IP cert vs domain cert vs no TLS), and never asserts "secure" on evidence it does not have.
- [ ] Provenance is visible to an operator for support purposes (surfaced on an admin settings/version surface).

### Detecting DO Marketplace provenance

Preferred: **a marker written by our own first-boot script**, because we build the snapshot (#2281) and therefore control it. Generalizes to Vultr and any future channel with no per-provider logic. Something like `/etc/trinity/install-source` or `TRINITY_INSTALL_SOURCE=do-marketplace` in `.env`, read once at boot and persisted to `system_settings`.

Rejected alternatives, with reasons:

- **DO metadata API** (`http://169.254.169.254/metadata/v1/`) answers "am I a Droplet", not "was I installed from the Marketplace". Also present on hand-built droplets. *(Whether it exposes any marketplace field is unverified — check before relying on it.)*
- **DO droplet application tags** (`common/scripts/020-application-tag.sh` in DO's own build scripts) live on the DO account side and need an API token we will not have.
- **Inferring from TLS/network state** — this is the failure mode described above; it fires on the whole managed fleet.

Note that DO's own Caddyfile convention stamps `header X-DO-MARKETPLACE "<app>"` on responses. That is an outbound response header, useful as corroboration but not as the source of truth.

### If a "your connection is not encrypted" style banner is ever built

Out of scope here, but recording the analysis so it is not re-derived: such a banner must **not** key on `location.protocol`. The correct predicate is *"is this session's transport unencrypted"* — HTTP **and** a publicly-routable client address. Excluding loopback, RFC1918, and Tailscale CGNAT (`100.64.0.0/10`) silences the entire managed fleet automatically with no config change, and TLS-terminating proxies and Cloudflare tunnels suppress for free since they connect from private addresses. `X-Forwarded-Proto` from a trusted proxy should be honored, with the existing `HTTPS_ENABLED` as an explicit override.

### Related — filed and fixed (2026-08-24)

The defect recorded here as "not yet filed" **was filed as #2381 and fixed by PR #2385** (merged 2026-08-24, `status-in-dev`). On a fresh install `setup_completed` stayed `false` while a real admin already existed, and the unauthenticated `POST /api/setup/admin-password` gated only on that flag — so it overwrote the live admin's hash and bound the caller's email as the sign-in identity.

What shipped, and what it means for this issue:

- `POST /api/setup/admin-password` now refuses whenever a usable admin exists — its own precondition, fail-closed, checked **above** the bcrypt hash. It no longer trusts the derived flag.
- `setup_completed` is reconciled at boot on **both** backends (PostgreSQL previously had no writer for the key at all). A boot-time reconciliation, not a migration — `setup_completed_backfill` is once-only and already recorded as applied on every affected install, so a new migration would have reached none of them.
- `/api/auth/email/verify` re-checks the allow-list before redeeming a code, closing the escalation that let a takeover outlive a password reset.
- `ADMIN_USERNAME` is resolved through `utils/admin_identity.admin_username()` instead of the hardcoded `"admin"`, and `docker-compose.prod.yml` now actually passes it (it was inert in production — the #1707 packaging-gap class).

**Consequence for the marketplace path:** a droplet that boots with `ADMIN_PASSWORD` set comes up with the wizard hidden and the setup endpoint already 403, before any browser touches it. This is exactly #2280's AC3 ("the setup form is never reachable unauthenticated on a public IP") for the env-provisioned case. The wizard still renders where no admin exists at boot (dev compose with a blank password, hand-rolled backend) — which is the premise trinity-enterprise#49 priced its tokenless window on, so ent#49's decision stands unchanged.

Nothing in this issue's own scope moved. Provenance + the hardening guide are still entirely unbuilt.

## Dependencies

- #2280 — prebuilt images, pull-only hosted compose, unattended first boot (the gate).
- #2281 — DO Marketplace listing; the Packer first-boot script is where the provenance marker gets written.
- #2332 — parent epic.

obasilakis:
## Summary

Records **how** an instance was installed, and shows a first-run HTTPS/VPN hardening guide on marketplace installs and nowhere else.

A marketplace droplet is the one install where Trinity knows at boot that it is on a public IPv4 with no domain, with zero network configuration, and an operator who has read no deployment docs.

**The obvious predicate is unusable, and that is the whole design.** Measured across all 16 managed instances: every one serves plain HTTP with no `DOMAIN`, no `HTTPS_ENABLED`, on a `100.x` Tailscale CGNAT address — structurally identical to an unhardened droplet, and **already correct**, because HTTP over a WireGuard tunnel is encrypted transport (HOST-010). A "no TLS configured → warn" rule fires permanently on every paying client. No environmental signal separates the two cases. The install channel does, and it is knowable only at provisioning time.

## Three properties that are security, not hygiene

- **Write-once.** Provenance is a fact about an installation *event*. If a later `.env` edit could rewrite it, it would answer "what does this box currently claim" instead, and the marketplace gate would be self-assertable by anyone who can edit a file. Enforced at the PRIMARY KEY on both backends (`INSERT OR IGNORE` / `ON CONFLICT DO NOTHING`), not by the `SELECT` before it — that is a separate statement, i.e. a check-then-act.
- **No env fallback on read.** `get_install_source` reads the row only. Without this, write-once is decorative: an unrecorded install could be talked into a marketplace verdict by exporting the variable.
- **Blocked on `PUT` *and* `DELETE`, and refused at the db sink.** The DELETE block is not symmetry — because the recorder is write-once, a delete is precisely the move that *unlocks* a rewrite. The sink guard follows `db/settings.py`'s own stated rule (boundary **and** sink, since the catch-all can write any key — the door found open by #506, #1609, ent#12, #1644, ent#14, ent#346 in turn).

**An unrecognised marker records nothing — not even `unknown`**, which would combine with write-once to freeze a typo permanently. An absent row already reads `unknown` and lets a corrected marker land next boot.

## Honest state by construction

Nothing probes a socket or reads a certificate — TLS terminates outside the backend, so no in-process check can. `install_tls_posture` is pure string parsing of the URL the instance *advertises*, resolved through the canonical `get_public_chat_url()` so it cannot disagree with public links and webhook registration about what this instance is called.

The card says "advertises", never "secure"; `https-ip` reads as working-but-upgradeable rather than broken (LE IP certs went GA 2026-01-15, DO's own 1-Click rules ship Caddy with them); and it states plainly that **Trinity issues no certificate itself** — nothing in the tree reads `public_chat_url` and reconfigures a proxy.

## Review found four defects in the first cut, all fixed here

1. The `https-ip` copy asserted *"a real, browser-trusted Let's Encrypt certificate … works today"* on the evidence that a string starts with `https://` — a direct AC violation, and the spec **required** the offending words. Copy rewritten; the honesty guard widened from grepping `secure|verified|validated` to rejecting connection-property claims generally, with a regression test that feeds it the old sentence and requires ≥4 hits.
2. The domain path promised *"a normal 90-day certificate then replaces the short-lived IP one"*. Verified false — `public_chat_url` is a display/webhook-base setting. Now attributes the effect to whatever terminates TLS.
3. The card rendered for **non-admins**, whose only action dead-ends (`general` is `adminOnly`, so the button lands them on the default tab) and whose copy discloses the box's network posture. Admin-gated, matching `AdminEmailNudge.vue`.
4. It did not retire **in-session**: `savePublicUrl` never force-refreshed the flags, so the card survived the action it asked for until a hard reload.

Also: posture resolution diverged from the canonical URL resolver; write-once rested on a check-then-act; config imports sat outside the fail-safe guards; and two doc claims of mine overstated (the parity test is prod↔hosted only, and retirement is server *state*, not verified fact — an admin can type any domain and suppress it, which is accepted and now written down rather than implied otherwise).

## Scope note

**Nothing in the OSS tree writes the marker** — #2281's Packer snapshot does. Until it lands, provenance reads `unknown` everywhere and the guide renders nowhere. That is PROV-004's contract working, and it is what makes shipping this half first safe.

## Test plan

- [x] `pytest tests/unit/test_2380_install_provenance.py` — **118 passed**. Mutation-tested: removing write-once, recording `unknown` on a bogus marker, reintroducing the env fallback, and dropping either router guard are each caught.
- [x] `pytest tests/unit/test_2217_canary_status.py test_2280_hosted_compose_parity.py` — 147 passed with the above
- [x] Broader sweep (`-k "settings or database or seed or version or ent435 …"`) — **754 passed**; the single failure (`test_736_a2a_outbound_properties`) reproduces identically on clean `dev` and is environmental (local 3.11 vs the 3.13 image pin)
- [x] `npm run test:unit` — **1528 passed**; `npm run build` and `npm run check:tokens` clean; new component contributes zero raw colors
- [ ] Manual: set `TRINITY_INSTALL_SOURCE=do-marketplace`, boot, confirm the row records once, the card renders for an admin only, and a second boot with a different marker leaves it unchanged

Fixes #2380


vybe:
Unblock status, @obasilakis:

**CodeQL** — both alerts were `js/incomplete-multi-character-sanitization` in `src/frontend/tests/unit/hardeningGuide.spec.js` (test-only). Dismissed as *used in tests*; the check should go green on the next analysis of this PR.

**regression diff** — this one is real and needs a fix. 4 new failures under HEAD that don't fail under BASE in any seed:

- `test_187_jwt_revocation::test_minted_token_carries_jti`
- `test_voice_auth.TestVoiceWebSocketAuth::test_admin_bypasses_ownership`
- `test_voice_auth.TestVoiceWebSocketAuth::test_other_user_rejected_4003`
- `test_voice_auth.TestVoiceWebSocketAuth::test_owner_passes_auth_gate`

The failure *counts* differ per head seed (6/3/2 vs base's 2/2/2), so this looks order-dependent — likely one of the PR's new tests leaking state (settings/env/module) into the JWT + voice-auth suites rather than a straight assertion break. Worth running the failing tests locally right after your new test modules to reproduce.

obasilakis:
Fixed in 412256d3.

**Root cause.** `TestMarkerNormalisation` called `importlib.reload(config)`. Reload mutates the shared module object in place, and `config.py` mints a random `SECRET_KEY` at import whenever the env var is unset — which it is under the unit suite. Modules that bound `from config import SECRET_KEY` at their own import kept the old value while anything reading `config.SECRET_KEY` per request got the new one, so a JWT signed on one side failed verification on the other. That is exactly the #1895 divergence, and the `pytest_collectstart` hook written for it structurally cannot catch this shape: it restores `sys.modules["config"]` to the *same object* a reload has already mutated.

Reproduced deterministically before the fix, which also explains the varying counts:

```
pytest tests/unit/test_2380_install_provenance.py \
       tests/unit/test_187_jwt_revocation.py \
       tests/unit/test_voice_auth.py -p no:randomly
→ 4 failed  (your four, exactly)
```

**Fix.** The intent behind the reload was right — the `marker` fixture rebinds `cfg.TRINITY_INSTALL_SOURCE` directly and so proves nothing about the real boot path — so that is kept and only the mechanism changes: `config.py` is now executed under a throwaway module name never registered in `sys.modules`, running the identical module-level statement with none of the reach. Pinning `SECRET_KEY` across the reload, or snapshot/restore of the module dict, would both have worked; each leaves a cleanup step that can be forgotten and would need revisiting the day config grows another import-time-derived value. Isolation has nothing to restore. Same command now: `143 passed`.

**Verification.** Full unit suite with the new module forced to run *first* (the ordering most likely to expose leakage from it), against the branch parent as baseline:

| | failures |
|---|---|
| base `13a3de8c^` | 23 |
| head + new module first | 23 |

`diff` of the two `FAILED` sets is empty — identical, zero new failures. The 23 are pre-existing IPv6-mapped-address failures local to this machine (`test_736_a2a_outbound_edges`, `test_ent14_registry_url_ssrf`, `test_mcp_validator::TestCgnatSsrfGuard`, …), red on both sides and untouched by this branch.

**CodeQL** — thanks, and confirmed independently: the `replace()` calls are only in `hardeningGuide.spec.js` (comment-stripping so the spec can assert on prose). `hardeningGuide.js` and `HardeningGuide.vue` contain no `replace()` and no sanitization, so there is no production twin of the flagged pattern.

vybe (approved):
Validated via /validate-pr: settings-row-only (no schema change on either track), write-once enforced by PK, both catch-all routes 422 the key after the admin gate, TRINITY_INSTALL_SOURCE wired into all three compose files + .env.example (hosted parity test green), 28/28 checks green incl. six-seed pytest + regression diff + frontend-e2e, CodeQL alerts test-only. Locally: 147 + 525 backend tests and 31 vitest pass on the head. Approving.

PR #2432, opened 2026-08-28
fix(retention): every install writes its own retention rows, not just fresh ones (#2085)

### LINKED ISSUE #2085: Seed retention rows on existing installs, not just fresh ones — the remaining half of #1638
## Summary

#1645 fixed #1638 by reverting `OPS_SETTINGS_DEFAULTS` to the wide historical values and applying the community floor through explicit `system_settings` rows seeded on **fresh installs only** (`_seed_fresh_install_retention`, `database.py`). That closed the incident.

It left one half of #1638's preferred fix undone: **existing installs still have no rows.** Their retention windows are resolved at prune time from the code dict:

```python
# services/cleanup_service.py
for key in RETENTION_OPS_KEYS:                       # settings_service.py:68 — eight keys
    db.get_setting_value(key, OPS_SETTINGS_DEFAULTS[key])
```

So every install that predates the seed — which is every install that has ever upgraded rather than been created fresh — has its retention policy defined by a dict that ships inside the backend image and is replaced on every rebuild. The safety of those installs today rests entirely on `OPS_SETTINGS_DEFAULTS` continuing to hold safe numbers, guarded by a code comment.

That comment is doing real work and is well written. It is still the only thing standing between a future edit and the #1638 failure mode: a lowered value hard-DELETEs existing data ~20 seconds after the next boot, no error, green `/health`.

## Ask

Extend the seeding that already exists to installs that are not fresh: for each key in `RETENTION_OPS_KEYS` with no `system_settings` row, insert the value currently in `OPS_SETTINGS_DEFAULTS`.

Properties that make this cheap:

- **Behaviourally inert.** It writes the number already in force. Nothing prunes differently the day it runs.
- **Idempotent and non-destructive.** `INSERT ... ON CONFLICT DO NOTHING` / `INSERT OR IGNORE`, exactly like `_seed_fresh_install_retention`. Never clobbers an operator's explicit value.
- **Reuses the existing mechanism.** Same function shape, minus the `_is_fresh_install` guard, sourcing `OPS_SETTINGS_DEFAULTS` instead of `COMMUNITY_FRESH_INSTALL_SEED`.
- **Covers all eight windows.** The fresh-install seed covers four; the prune-time fallback covers eight. `agent_reports_retention_days`, `operator_queue_retention_days` and `agent_reminders_retention_days` have the same exposure and no seed today.

After it runs once, no install anywhere resolves a retention window from the image. A future edit to `OPS_SETTINGS_DEFAULTS` can only affect installs created after it — which is what changing a default should mean.

## Tradeoff, stated explicitly

Seeding freezes each install at the values in force when it first runs the seed. A later *widening* of a default will not reach it either. That is the intended consequence: retention becomes explicit per-install config rather than an implicit inheritance from whatever the image happens to carry. Widening for existing installs then becomes a deliberate migration, symmetric with the rule the current comment already imposes on narrowing.

Timing note: if this ships against today's `dev`, every existing install freezes at `30 / 90 / 7 / 180 / 30` plus the current values for the other three — the safe historical set.

## Where

- `src/backend/database.py` — `_seed_fresh_install_retention` / `_seed_fresh_install_retention_engine` (the shape to copy; both the SQLite and engine paths need it)
- `src/backend/services/settings_service.py:68` — `RETENTION_OPS_KEYS`, `OPS_SETTINGS_DEFAULTS`
- `src/backend/services/cleanup_service.py:266` — the prune-time resolution this removes reliance on

Same fail-safe posture as the existing seed: a failed seed must not raise, and leaves the install on the wide code defaults.

## Why it is worth doing rather than relying on the comment

Operators currently have no way to see this exposure and no supported way to fix it in bulk. `GET /api/settings/retention` reports per-key `source` (`db-row` vs `code-default`), so an operator can observe that a window is unpinned — but the only remedy is to set all eight by hand on every instance, forever, including on any DB restored from a backup predating the writes.

Ops-side this has already produced a brittle workaround: the fleet's update procedure greps backend source before every restart to guess whether the incoming image carries a dangerous default. It has fail-opened once when the value became a named constant, and currently false-positives on every instance because that constant is re-exported. Both failure modes are the same category — inferring a Python dict's value from the outside. Seeding removes the need to infer it at all.

## Acceptance

- [ ] An install with no retention rows has all eight written on its next boot, at the values then in force
- [ ] An install with some rows keeps every existing value untouched; only missing keys are filled
- [ ] Fresh installs are unaffected — `COMMUNITY_FRESH_INSTALL_SEED` still applies its four
- [ ] Seed failure is non-fatal and leaves the install on the code defaults
- [ ] Works on both SQLite and PostgreSQL

Refs: #1638, #1645, #1039, #1065, #772, #834

obasilakis:
## What

`#1645` closed `#1638` by reverting `OPS_SETTINGS_DEFAULTS` to the wide historical values and applying the `#1039` community floor through explicit `system_settings` rows seeded on **fresh installs only**. Every install that has ever *upgraded* rather than been created fresh therefore had **no rows at all**, so `cleanup_service` resolved all 11 windows at prune time from a dict that ships inside the backend image and is replaced on every rebuild.

The only thing between a future edit to that dict and the `#1638` failure mode — a silent hard-DELETE of existing data seconds after the next boot, green `/health`, no error — was a code comment. `#1638` is what happens when that comment is missed; it cost a real instance ~3 months of history.

`database._seed_retention_windows{,_engine}` now writes an explicit row for every `RETENTION_OPS_KEYS` member that has none, at **the value already in force**, on every boot regardless of install age.

## Why it is safe to ship with no operator action

**Behaviourally inert.** It writes the number the prune already used, so nothing prunes differently the day it runs. Observed on a real boot:

```
[#1638] Fresh install: seeded community retention floor (4 keys @ 5d)
[#2085] Seeded 7 retention window(s) with the value already in force
        (agent_reminders=90, agent_reports=90, agent_soft_delete=180,
         backup=14, operator_queue=90, subscription_failure_event=30,
         subscription_headroom=30)
```

4 + 7 = 11, and the `#1039` floor won the four keys both seeds write.

## Three load-bearing properties

**Key set derived from `RETENTION_OPS_KEYS`**, never a second hand-written list, so a window added later is covered the day it ships instead of quietly inheriting the image default forever. The issue text says "eight windows" — it was already 11 (`ent#433` added two, `#2216` a third).

**Ordering.** Must run *after* `_seed_fresh_install_retention`. Both writers are insert-or-ignore, so the first to reach a key wins: reversed, a fresh install silently gets the wide defaults instead of the `#1039` floor — the community floor deleted by the change meant to protect retention. Pinned behaviourally **and** by a source-order guard on both the SQLite and engine arms.

**It must actually run.** The first cut imported the constants from `services.settings_service`, which has a module-level `from database import db`. `init_database()` is called from `DatabaseManager.__init__` — while `database.py` is still executing its own module body — so `database.db` does not exist yet and that import raises `ImportError`, which this seed's fail-safe contract then **swallows**:

```
WARNING: [#2085] retention window seed skipped (cannot import name 'db' from
partially initialized module 'database' (most likely due to a circular import))
```

The feature was dead on every boot with a **fully green unit suite**, because every in-process test calls the function after `database` has finished importing. `_seed_fresh_install_retention` documents this exact trap three lines from where the seed was written.

So `RETENTION_OPS_KEYS` / `OPS_SETTINGS_DEFAULTS` / `NON_ROW_RETENTION_OPS_KEYS` move to `config.py` — a leaf, already home to `OPS_SETTINGS_VALIDATION` and `validate_ops_setting` for these same keys — and are re-exported from `settings_service`, extending the pattern `COMMUNITY_FRESH_INSTALL_SEED` already used for precisely this reason. All 23 existing import sites are unaffected. Two tests now pay for a real subprocess import; both fail if the import is reverted (verified).

## Stated tradeoff

A seeded install stops inheriting later changes to the code default in **either** direction, so *widening* a window for existing installs becomes a deliberate migration rather than something that arrives silently with an image. That is the intended consequence — retention becomes explicit per-install config instead of implicit inheritance from whatever image is running, symmetric with the rule the `OPS_SETTINGS_DEFAULTS` comment already imposes on narrowing.

## Notes

- `backup_retention_days` is seeded too, which makes `OPS_SETTINGS_DEFAULTS`' value the one that lands in the DB for a key whose private reader (`effective_backup_retention_days`, inverted coercion) falls back to its **own** module constant — the two are now parity-tested.
- **No schema change, no migration** — row inserts at boot, same as the `#1638` seed. Alembic single-head PASS.
- **Not fixed here:** generic `DELETE /api/settings/{key}` carries no `RETENTION_OPS_KEYS` guard (only `PUT` does), so an admin can still delete a window row. After this it is transient — re-seeded next boot — but the asymmetry with `PUT` remains.
- **Unblocks `ops#300` once deployed.** With rows on every install, `/update` step 8e can drop its source-text guessing for a plain assertion over stored values. The deletion must not land until this has actually *reached* the instances being updated.

## Testing

14 new tests in `tests/unit/test_2085_retention_seed_existing_installs.py` — coverage, inertness, ordering (behavioural + source-order on both arms), no-clobber, idempotence, never-raises, the `backup_retention_days` parity coupling, and two real-import subprocess guards.

Full unit suite, branch vs clean base `8df57509`:

| | base | branch |
|---|---|---|
| failed | 23 | 23 |
| passed | 12854 | 12868 (+14) |

`comm` diff both directions is empty — identical failure sets, all 23 pre-existing on `dev` (SSRF/IPv6 parsing + portal submodule guards, none near retention).

Closes #2085


vybe (approved):
Validated: closes the upgraded-install half of #1638/#2085 — retention rows seeded on every boot, not just fresh installs; fail-safe direction preserved (seed only writes missing rows). Test named for the issue; CI gate green (seed-99999 job was infra-cancelled, regression diff passed).

Issue #1850, opened 2026-07-28
Epic: CI health & gating + journey coverage — green CI, and every promise tested end to end


vybe:
Carved out of #1259 (epic-health pass 2026-07-28), **rechartered 2026-08-20** to name both halves explicitly. Its previous charter ("green CI, test infrastructure, automated validation") is why journey work had no obvious home and #2037 ended up parentless.

## Charter — two halves, one epic

**A. CI health & gating** — a trustworthy, fully-green CI that catches regressions before they ship.
- CI/CD pipeline health and coverage (#851 anchor, with #556)
- The merge-gate hole: dev's required checks run no unit tests (#1958)
- Frontend + integration test infrastructure
- Automated/scheduled validation runs on the dev instance
- Template/bundle validation guards

**B. Journey coverage** — every promise the platform makes to a user or an agent is tested end to end, and the suite grows when the product grows.
- The rails: a `tests/journeys/` tier, a per-PR journey-smoke gate, one invariant namespace, the journey catalog, consolidated testing docs
- The journeys themselves: J01–J10
- The process gates that make coverage an obligation rather than a request

## Why both halves live here

They are the same problem seen from two ends: the suite defends *past fixes* (ticket-named tests) rather than asserting *current promises*, and the gate that would catch the difference collects `tests/unit/` only. Splitting them into two epics recreates the fragmentation this recharter exists to kill.

## The durable/delivery split — read this before judging the %

An epic is a **delivery vehicle** with a rollup % and a hard 100-sub-issue cap (#1045 already hit it). The **durable** home for journey coverage is `tests/journeys/catalog.yaml` + `docs/testing/STRATEGY.md`, which never close and expand with the roadmap.

With ~40 mixed children — one-off CI bugs alongside structural journey work — **this epic's % is burn-down on the current wave, not a coverage number.** Coverage % comes from the catalog (journeys built and green ÷ journeys declared). Two numbers, two questions.

## Operator decisions (2026-08-20)

1. **Tracker: public.** Sub-issues cannot cross repos and test infra is core code.
2. **Journey cap: ~15.** Past that it is a lane or a step, not a journey.
3. **Release-blocking: advisory this cut, automatic from the next.** A suite built *during* the release it would block cannot also be that release's gate.
4. **Hosted lanes: DigitalOcean supported (must be green to cut); Vultr and Hostinger best-effort.**
5. **Owner: the `trinity` dev agent by default, with the operator named on the top three journeys.**

## Sequencing

[...]

PR #2437, opened 2026-08-31
refactor(main): lifespan is an orchestrator, not a 580-line procedure (#1028)

dolho:
## Summary

`main.py::lifespan` was **580 lines at cyclomatic complexity 109** — the longest function in the backend and the first item the 2026-06-02 refactor audit named. It is now **25 lines at CC 1**: a flat list of `await _phase()` calls over twelve startup helpers and four shutdown helpers.

Closes one of #1028's acceptance criteria (`main.py::lifespan` startup wiring grouped into helper init functions). Scope is deliberately one file, per the issue's own "one file per PR" recommendation.

| | before | after |
|---|---:|---:|
| `lifespan` lines | 580 | **25** |
| `lifespan` complexity | 109 | **1** |
| helpers over threshold | — | **none** (worst: 84 lines, CC 17) |

## Bodies are verbatim except four re-materialised imports

Extracting all non-blank lines from the sixteen helpers **in call order** and diffing against the original `lifespan` body gives **518 = 518, identical**.

That proof was true and the branch was still broken — see the section below. The bodies are now verbatim **except** four function-local imports that had to be restored, because the split removed the enclosing scope they were reaching through. The original claim is corrected rather than kept: a verbatim claim stops being true the moment a leaked name has to be re-materialised.

No logic touched, no `try/except` reshaped. Each phase keeps its own guard, because a failing phase must not take the boot down — which is what the original did.

## The test is the point, not the split

Splitting the function is easy. Keeping it split is not, and the thing worth guarding is **not the line count — it is the order**.

Boot ordering is load-bearing in ways that are invisible at the call site. A reviewer looking at sixteen `await` lines has no way to see that moving one breaks something, because the coupling lives in the bodies. Before this it was implicit in a 580-line function nobody could read in one sitting; after it, it is a list — which is an improvement **only if something enforces the list**.

`tests/unit/test_1028_lifespan_phases.py` (21 tests) pins the sequence with the reason for each constrained pair recorded beside it:

- **logging first**, so a later hang cannot swallow the boot log (#858 — `print()` is block-buffered to the Docker pipe and silently lost)
- **event bus before** any WebSocket client needs a live dispatcher to register with (#306)
- **Docker/system-agent before** the services that sweep the fleet
- **startup recovery before the channel transports**, so inbound traffic cannot create an execution that races the reconcile
- **event-bus drain LAST** on shutdown, so late-lifecycle broadcasts emitted while the other services stop still land on the stream

A reorder now fails with the reason attached, instead of surfacing weeks later as a boot-order bug nobody connects to this commit.

It also pins the `yield` split (a phase appended after it would silently become shutdown work), the per-helper thresholds the issue asked for (<100 lines, CC <20, so the split cannot rot back by accretion — which is exactly how `lifespan` reached 580 lines), and orphan/double calls.

Mutation-checked, five ways — **all caught**:

| mutation | result |
|---|---|
| recovery moved after the transports | 2 failed |
| event bus moved after Docker | 2 failed |
| a shutdown phase dropped | 3 failed |
| event-bus drain no longer last | 2 failed |
| a phase pushed after `yield` | 1 failed |

## The defect `/review` caught, and why three verifications missed it

`main.py` never imports `database` at module level. The old `lifespan` did `from database import db as _db` **once near the top**, and three blocks 100+ lines later reached it through the enclosing function scope — the first-run seed / system-agent gate, the Telegram transport, the WhatsApp transport. `message_router` was the same shape: imported in the Slack block, used by the Telegram one.

After the split all four are `NameError`. **And every use site sits inside `try/except Exception`, so the boot succeeds.** Reproduced against the pre-fix commit by actually executing those phases:

```
SWALLOWED: Error scheduling first-run seed: name '_db' is not defined
SWALLOWED: Error starting Telegram transport: name 'message_router' is not defined
```

So: first-run seeding silently never scheduled, Telegram and WhatsApp integrations silently never wired. A dead integration, not a failed boot — and the per-phase guard that makes each phase fail-open is exactly what hides it.

Three verifications had already passed on the broken branch, which is the part worth internalising:

| check | why it missed this |
|---|---|
| AST equivalence proof (518 = 518) | compares **lines**, and the lines are identical |
| 21-test structural pin | asserts order, thresholds, decorators — not **name resolution** |
| real `import main` smoke (591 routes) | never **runs** `lifespan`, so nothing resolves those names |

Fixed by re-materialising each import in the helper that needs it — the original's own idiom — each with a comment saying why, so a later reader does not tidy it back out.

Pinned by `test_no_phase_helper_depends_on_another_phases_locals`: per helper, `names_loaded − names_bound − module_globals` must be empty. Mutation-checked by deleting the restored `_db` import — reproduces the shipped bug and turns the suite red. This matters more with every future split of the same function.

`docs/memory/learnings.md` gains the class: extract-method has a failure mode the diff cannot show and an import smoke cannot reach.

## A second defect this found in itself

Recording it because it is precisely the failure this refactor's shape invites, and because it says something uncomfortable about our coverage.

The extraction moved `@asynccontextmanager` **by one definition**. It landed on the first phase helper, and `lifespan` was left a bare async generator — which FastAPI cannot use as a lifespan. **Boot-breaking.**

The entire 12,900-test unit suite stayed green, because nothing in it imports `main` and asks what shape `lifespan` is. It surfaced only from an explicit import check: `inspect.iscoroutinefunction` returned `False` for one helper, with `__code__.co_filename` pointing into `contextlib.py`.

Fixed, and pinned by `test_only_lifespan_carries_the_asynccontextmanager_decorator` — a decorator sits one line above a `def`, and this change inserts ~700 lines at exactly that seam, so it is guarded rather than remembered.

(Worth noting as a journey-shaped gap: *"the backend boots"* is a promise with no test asserting it in the merge path — #1850 / J-series territory, not this PR.)

## Test plan

- [x] `pytest tests/unit/test_1028_lifespan_phases.py` — **37 passed**
- [x] Mutation-check of the ordering pin — 5/5 caught; scope-guard mutation reproduces the shipped bug
- [x] Executed the three previously-broken phases directly — **no NameError-shaped ERROR logs**; same run on the pre-fix commit emits them
- [x] Verbatim-equivalence proof — 518 = 518 lines identical
- [x] Real import of `main` with a writable DB path: app builds, **591 routes registered**, `lifespan` is an `_AsyncGeneratorContextManager` factory, all 16 helpers are coroutine functions
- [ ] Full unit suite — running locally; CI is the gate

## Docs

`architecture.md`'s `main.py` row now records that the order is the contract, where the constraints are, and that new startup work belongs in a phase helper. Tiered rule: this is a refactor with no API or schema change, so architecture-only is the right surface — no `requirements/` change, no feature flow.

## Not in scope

The remaining #1028 acceptance criteria stay open, one PR each as the issue recommends: `routers/settings.py` (1,037 logical lines, 71 handlers — carries real Invariant #4 route-ordering risk), `routers/ops.py` (`restart_fleet` 312 lines / CC 37, carries the #1919 lease), `services/git_service.py`, `services/agent_client.py`, `routers/public.py::public_chat`.

Related to #1028


dolho (commented):
## /review — pre-landing structural review

Self-review of this branch against `dev` (merge-base `135248e9`). One critical finding, found and fixed in-branch as `cd2cc756`; recording it in full because the *way* it evaded detection is more interesting than the bug.

**Scope**: CLEAN — one file, per #1028's own "one file per PR" recommendation. The other five acceptance criteria are deliberately untouched and named in the PR body.

---

### [C1] Correctness: three phase helpers used locals the split left behind — Confidence 10/10 — **FIXED**

`src/backend/main.py:384` (bind) → `:487`, `:831`, `:874` (uses); `:766` → `:822`

```python
    from database import db as _db                       # line 384, inside phase 1
...
            if _db.get_setting_value('setup_completed', 'false') == 'true':   # 487, phase 4
            bindings = _db.get_all_telegram_bindings()                        # 831, phase 11
            bindings = _db.get_all_whatsapp_bindings()                        # 874, phase 12
        _telegram_transport = TelegramWebhookTransport(_telegram_adapter, message_router)  # 822
```

`main.py` has **no module-level `database` import**. The former `lifespan` bound `_db` once near the top and three blocks 100+ lines later reached it through the enclosing function scope. `message_router` was the same shape — imported in the Slack block, used by the Telegram one.

After the split, all four are `NameError`.

**Why this is worse than a crash.** Every one of those use sites sits inside `try/except Exception`, so the boot *succeeds*. Reproduced by actually executing the affected phases against the pre-fix commit `836412ee`:

```
SWALLOWED: Error scheduling first-run seed: name '_db' is not defined
SWALLOWED: Error starting Telegram transport: name 'message_router' is not defined
```

First-run seeding silently never scheduled. Telegram and WhatsApp integrations silently never wired. The per-phase `try/except` that makes each phase fail-open — deliberately preserved from the original — is exactly what hides the extraction bug.

**Three verifications had already passed on the broken branch.** This is the part worth internalising:

| check | why it missed this |
|---|---|
| AST equivalence proof, 518 = 518 lines | compares **lines**; the lines *were* identical |
| 21-test structural pin | asserts order, `yield` split, thresholds, decorators — not **name resolution** |
| real `import main` smoke, 591 routes | never **runs** `lifespan`, so nothing resolves those names at import time |

**Fix**: re-materialise each import in the helper that needs it — the original's own idiom — each with a comment explaining why it is there, so a later reader does not tidy it back out.

**Guard**: `test_no_phase_helper_depends_on_another_phases_locals` asserts, per helper, that `names_loaded − names_bound_locally − module_globals` is empty. Mutation-checked by deleting the restored `_db` import: reproduces the shipped bug and turns the suite red. This gets more likely with every future split of the same function, so it is a test rather than care.

**Claim corrected**: the PR body no longer says the bodies are byte-identical. They are verbatim *except* these four restored imports. A verbatim claim stops being true the moment a leaked name has to be re-materialised, and quietly keeping it would be worse than the bug.

---

### [I1] Documentation: nine docstrings claimed "phase N of 10" — Confidence 10/10 — **FIXED**

The channel transports were split into three helpers *after* that text was generated, so the count is 12.

### [I2] Consistency: `_shutdown_background_services(app)` never uses `app` — Confidence 7/10 — **not fixed, deliberate**

Uniform across all four shutdown helpers; the other three genuinely read `app.state`. Flagging so it is not later mistaken for an oversight.

---

### Clean categories

- **SQL / data safety** — no SQL, no schema change, no migration in the diff.
- **Auth boundaries** — no endpoint added, removed or re-gated; 591 routes before and after.
- **Credential exposure** — no new value logging; the `AGENT_AUTH_SECRET` warning moved verbatim and reports presence only.
- **Race conditions** — every `asyncio.create_task` call site moved unchanged. The pre-existing no-strong-reference pattern is *preserved*, not introduced: the task holds the coroutine, which holds the closure, so the helper's frame returning changes nothing.
- **Enum completeness** — no new enum, status or constant.
- **Error handling** — every `try/except` moved verbatim; none reshaped, added or removed.
- **Test gaps** — 37 tests added; ordering mutation-checked 5 ways (recovery after transports, event bus after Docker, dropped shutdown phase, drain no longer last, phase pushed past `yield` — all caught).
- **Docs staleness** — `architecture.md`'s `main.py` row updated. Refactor with no API or schema change, so architecture-only is the correct surface; no requirements or feature-flow change.

### Summary

- **Critical**: 1 — found and fixed in-branch (`cd2cc756`)
- **Informational**: 2 — one fixed, one deliberate
- **Scope**: clean
- **Learnings**: captured — `docs/memory/learnings.md` gains *"Splitting a long function breaks the shared locals nobody notices are shared"*, because extract-method has a failure mode the diff cannot show and an import smoke cannot reach. `/autoplan` reads that file.

The uncomfortable takeaway, stated plainly: this branch was verified three ways, opened as a PR, and was **broken the whole time**. It took a scope pass that asked "what names does each new function actually resolve?" — not another test of what the code does — to see it.

dolho (commented):
## /review — re-review after `cd2cc756`

Fresh structural pass against `dev` (merge-base `135248e9`), head now `c0602772`. Three commits, +526/−5 across 4 files.

### C1 is closed — verified, not assumed

All three checks that would have caught it originally, re-run against the current head:

| check | result |
|---|---|
| cross-phase scope leak (`names_loaded − names_bound − module_globals`, per helper × 16) | **clean** |
| helpers over threshold (>100 lines or CC >20) | **none** |
| `lifespan` shape | 25 lines, CC 1, `decorators=['asynccontextmanager']` |
| stale `"of 10"` docstrings | **0** (9 now correctly say `of 12`) |

And behaviourally, by executing the three previously-broken phases directly: **no `NameError`-shaped ERROR logs**. The same run against the pre-fix commit still emits them, so the guard is measuring something real rather than passing vacuously.

The fix commit's own added code is four `import` lines, nine docstring count corrections, and comments. Nothing else moved — re-checked by diffing `836412ee..cd2cc756` and stripping comments.

### New finding this pass

**[I3] Documentation: the corrected verbatim claim survived in three docstrings** — Confidence 9/10 — **FIXED** in `c0602772`

I corrected *"bodies are byte-identical"* in the PR body but left the identical claim standing in the code, where it is considerably more likely to be believed. All three helpers that gained a re-materialised import still read:

```
Extracted VERBATIM (#1028); the body below is unchanged.
```

sitting directly above the import that falsifies it. That is precisely the line a future reader checks against before concluding the import looks redundant and removing it — which reinstates C1. Now each says verbatim **except** the restored import and points at the comment explaining why.

Worth naming the pattern rather than just the instance: when a claim is corrected, correct it *everywhere it was made*. Fixing the loudest copy and leaving the one nearest the code is the wrong half.

### Re-checked and still clean

- **SQL / data safety** — no SQL, schema, or migration in the diff.
- **Auth boundaries** — no endpoint added, removed or re-gated; 591 routes.
- **Credential exposure** — the restored `from database import db as _db` imports a module-level singleton; no value is logged, no secret enters a message.
- **Race conditions** — the fix adds imports only; no new task, lock or shared state.
- **Error handling** — no `try/except` added, removed or reshaped by either follow-up commit.
- **Enum completeness** — nothing added.
- **Test gaps** — 37 tests; ordering mutation-checked 5 ways, scope guard mutation-checked against the real shipped bug.
- **Docs staleness** — `architecture.md` `main.py` row updated; `learnings.md` carries the durable class.

### Summary

- **Critical**: 0 outstanding (1 found and fixed across the two reviews)
- **Informational**: 3 total — 2 fixed, 1 deliberate (`_shutdown_background_services(app)` takes `app` it does not use, for uniformity with its three siblings that do)
- **Scope**: clean
- **Remaining gate**: full unit suite against the corrected tree, running locally; CI is the authority.

Ready for a second pair of eyes. The one thing I would ask a reviewer to look at hardest is not the split — it is whether any *other* name in those sixteen helpers was reaching through the old enclosing scope in a way both the static guard and I have missed.

vybe (approved):
Validated: pure structural refactor (lifespan 580L/CC109 → 25L orchestrator over phase helpers), closes a #1028 AC; behavior-preserving with review-round fixes already folded in and full pytest matrix green. Vas — approved to unblock the merge train; post-merge comments still welcome.

PR #2439, opened 2026-08-31
fix(files): shared links were unopenable on mobile — Range, disposition, MIME, CORP (trinity-enterprise#461)

dolho:
## Summary

A file shared via `share_file` and delivered as a link over Telegram **would not open on mobile**. Reported for a Cornelius voice note (WAV, 2.0 MB).

The bytes were never wrong. Verified from the Cloudflare edge, the object returned `HTTP 200` with correct `content-length` and correct WAV bytes for every user-agent tried, and the signature check worked. **The response shape was wrong in four ways at once**, and each is independently fatal in an iOS in-app browser:

| header | was | why it broke |
|---|---|---|
| `accept-ranges` | absent; `Range` → `200` + full 2 MB | iOS Safari / Telegram need a `206` to start audio **at all** |
| `content-disposition` | `attachment` | a forced 2 MB download in Telegram's iOS browser is a blank screen |
| `content-type` | `audio/x-wav` | unregistered; under `nosniff` a strict player declines and the browser may not guess |
| `cross-origin-resource-policy` | `same-origin` | on a link whose whole purpose is being opened from another platform |

Plus `cache-control: no-store`, which forbids the in-app browser from buffering media it will not play without buffering.

All fixes are on the **delivery layer**. The auth layer is untouched.

## The inline change is a narrowing, not a reversal

The old code forced `attachment` on everything, with the note *"defense against XSS via agent-uploaded HTML"*. **That reasoning is still correct** — this route serves agent-authored bytes from `public.*`, the same origin as public chat.

So inline is an **allowlist** (`_INLINE_SAFE_TYPES`: audio, video, image, PDF). `text/html`, `application/xhtml+xml` and **`image/svg+xml`** stay attachments. SVG is called out explicitly because it is the one a reviewer waves through — it is an image by name and a script host in fact, and "just allow `image/*`" is the tempting version of this change.

Three properties make the failure direction safe:
- the type is **python-magic-detected from the file's own bytes** at share time, never agent-supplied;
- magic's unavailable-fallback is `application/octet-stream`, which is **outside** the allowlist;
- `nosniff` is kept, and matters *more* now that anything is inline, not less.

## Two things the issue didn't ask for

Both found while implementing, both real:

**`Content-Length` came from the DB.** `size_bytes` is written at share time; the file on disk is what is actually served. Any drift is unrecoverable for the client — too small truncates the download, too large hangs it — and Range math against a wrong total emits a `Content-Range` that contradicts the body. Now `os.path.getsize`, with a WARNING on divergence.

**A ranged fetch is one play, not many downloads.** A media player fetches one file as dozens of ranged requests. Counting each would inflate `download_count` and write an audit row per chunk. The counter and the audit now fire only on the **transfer start** — a plain GET, or a range beginning at byte 0.

## Verification — against the real route, not just the parsers

```
full GET      : 200 | type audio/wav | disp inline | ranges bytes
              | corp cross-origin | cc private, max-age=3600
range 0-1023  : 206 | body 1024 | bytes 0-1023/2048000 | bytes ok
suffix -500   : 206 | bytes 2047500-2047999/2048000 | bytes ok
unsatisfiable : 416 | bytes */2048000
HEAD          : 200 | accept-ranges bytes | content-length 2048000
no sig        : 401     bad sig : 401     unknown id : 404     expired : 410
html file     : attachment          svg file : attachment
```

That is the issue's Definition of Done line by line, **including** that the signature check still rejects unsigned and expired requests.

- [x] 46 new unit tests (`test_ent461_shared_file_delivery.py`)
- [x] 175 pass across the file-sharing suites
- [x] End-to-end HTTP check above, on a 2 MB fixture matching the reported file

Tests are weighted to the allowlist and to the range parser's silent-corruption case: `bytes=-500` means the **last** 500 bytes, and reading it as `start=0` serves the wrong bytes under a `206` — which no client can detect.

## Notes

- **Tracker vs repo**: filed on the enterprise tracker, but `/api/files/{id}` is FILES-001 OSS core, so the fix lands here per *"tracker ≠ code repo"*.
- **HEAD carries `Accept-Ranges` and the true `Content-Length`** — it is the probe a player makes *before* deciding whether it can stream, so a HEAD omitting them defeats Range support even where GET implements it.
- **Multi-range** requests fall back to the full `200` (RFC 7233 §3.1 permits it). `multipart/byteranges` is a lot of machinery for a case no media player generates.
- **`Cache-Control`** is `private, max-age=min(remaining lifetime, 3600)` — never `public` (the URL is the credential), never longer than the link lives, and it falls back to `no-store` if the expiry can't be read.
- Docs: `architecture.md`'s Outbound File Sharing section records the delivery contract and the allowlist rationale. No schema change, no new endpoint, no requirements change — this is a bug fix inside an existing capability.

Related to trinity-enterprise#461


dolho:
## Tested against a live local instance — and it found a real defect

Ran this branch as a **real uvicorn server** (not a `TestClient`) against local Redis, seeded **genuine shares through `create_share_from_bytes`** so MIME came from python-magic reading actual bytes, and drove it with `curl` and `ffprobe`.

### 🔴 The fix was partly inert, and only this caught it

`main.add_security_headers` runs after every route and set `Cross-Origin-Resource-Policy` with a **plain assignment**, silently overwriting the `cross-origin` policy this route sets on its way out:

```
before: cross-origin-resource-policy: same-origin     ← my header, clobbered
after : cross-origin-resource-policy: cross-origin    ← file route
        cross-origin-resource-policy: same-origin     ← /health, unchanged
```

That is the fix deciding whether Telegram/Slack/WhatsApp can embed the link **at all** — shipped dead, with every unit test green, because a bare `FastAPI()` + router harness has no middleware. Fixed in `305477c2` with `setdefault`, so absence still resolves to the strict default and a new route must opt out deliberately rather than inherit an exception.

I'd flagged my earlier "end-to-end HTTP check" as verification. It wasn't — it exercised a bare app. Correcting that.

---

### A · The reported case — 2.0 MB WAV, `audio/x-wav` detected from real bytes

```
HTTP/1.1 200 OK
content-type: audio/wav                      ← normalized from audio/x-wav
content-disposition: inline; filename="cornelius voice thoughts.wav"
accept-ranges: bytes
cross-origin-resource-policy: cross-origin
cache-control: private, max-age=3600
x-content-type-options: nosniff
content-length: 2028644
body=2028644   byte-identical to source: YES
```

### B · Range

| request | status | `Content-Range` | body |
|---|---|---|---|
| `bytes=0-1023` | **206** | `0-1023/2028644` | 1024 |
| `bytes=1024-2047` | **206** | `1024-2047/2028644` | 1024 |
| `bytes=2000000-` | **206** | `2000000-2028643/2028644` | 28644 |
| `bytes=-500` | **206** | `2028144-2028643/2028644` | 500 |
| `bytes=0-0` | **206** | `0-0/2028644` | 1 |
| `bytes=99999999-` | **416** | `bytes */2028644` | — |

### C · Security — the allowlist, on real magic-detected types

| seeded | stored mime | served | disposition | nosniff |
|---|---|---|---|---|
| `evil.html` | `text/html` | `text/html` | **attachment** | yes |
| `evil.svg` | `image/svg+xml` | `image/svg+xml` | **attachment** | yes |
| `photo.png` | `image/png` | `image/png` | inline | yes |
| voice note | `audio/x-wav` | `audio/wav` | inline | yes |

### D · Auth — unchanged

`no sig → 401` · `wrong sig → 401` · `unknown id → 404` · `legacy ?download_token → 200` · `wrong sig + Range → 401` (auth before range) · revoked `GET`/`Range`/`HEAD` → **410** on all three

### E · Does it actually play?

```
$ ffprobe <url>
format_name=wav   duration=46.000000   size=2028644
```

A real decoder, over HTTP, through the route. Plus the iOS probe sequence: `HEAD → 200`, `Range 0-1 → 206`, `Range 0- → 206`.

### F · Reassembly

20 sequential ranges of 101,433 bytes → **2,028,644 bytes, byte-identical to source**. No off-by-one at chunk boundaries or at EOF.

### G · Counter semantics — controlled measurement

1 full GET + **10 mid-file ranges** + HEAD + 416 → `download_count = 1`. A play is one download, not thirty.

---

**Note on the 429s:** mid-run, some requests returned 429. That is the pre-existing 60/min per-IP limit after ~70 requests from one address, not a regression — confirmed by clearing the bucket and re-running, which gave the expected 410s.

**Suites:** 176 pass across the file-sharing and security-header tests; 47 in the ent#461 file, including a new pin on the `setdefault` so a future edit back to `=` can't re-break it invisibly.

vybe (approved):
Validated: root-caused delivery-shape fix (Range/disposition/MIME/CORP) for ent#461 with dedicated test file; CI gate green (the seed-99999 'fail' is an infra-cancelled job, regression diff passed). Note: cross-tracker ent#461 needs a manual status bump — same-repo automation won't promote it.

PR #2440, opened 2026-08-31
chore(enterprise): bump the submodule pointer to main (a419812 → 90f2f2c)

dolho:
## Summary

Yes — and it is a **pointer inversion**, not just staleness. `dev` pins the enterprise submodule *older than `main` does*, so the next `dev → main` release merge would have carried the pointer **backwards** and undone ent#443's enterprise-side removal.

```
OSS main -> 2a5def3   (ent#443: shared_sessions removed from enterprise)
OSS dev  -> a419812   (2026-08-19, 4 behind enterprise main)   ← older than main
ENT main -> 90f2f2c
```

`90f2f2c` is a **fast-forward from both** — `a419812...main` is ahead 0 / behind 4, `2a5def3...main` is ahead 3 / behind 0 — so nothing is rewound.

## The four commits

| | |
|---|---|
| `2a5def3` | `refactor(rooms)`: remove `shared_sessions` — it now lives in OSS core (ent#443) |
| `ff0a4f1` | `fix(security)`: guard the enterprise `system_settings` sinks against cleartext credentials (ent#435) |
| `6d82a3f` | `docs(workspace)`: agent-initiated asks — design of record |
| `90f2f2c` | `feat(credential-vault)`: governed system credential vault module (ent#279) |

`ff0a4f1` is the private twin of the OSS sink guard that `architecture.md` already records as *"the private submodule owns its twin"* — so `dev` is currently running the OSS half of a security fix without its enterprise half.

## Why it matters, rather than being housekeeping

ent#443 moved rooms into OSS core, and `dev` has that. With the stale pin, an entitled `dev` install mounts **both** the OSS rooms routers **and** the enterprise `shared_sessions` module, relying on `main.py`'s include-order (OSS before `register_enterprise`) to decide which one serves. `architecture.md` documents that ordering as the *transition* safety net. This bump is the follow-through that ends the transition rather than leaving the net permanently load-bearing.

## Verified by booting both pointers

Same box, same DB shape, real uvicorn, enterprise submodule mounted at each commit in turn:

| pointer | modules | `shared_sessions` registered | room paths | enterprise errors |
|---|---|---|---|---|
| `a419812` (today) | 17 | **True** | 6 | 0 |
| `90f2f2c` (this PR) | 17 | **False** | 6 | 0 |

Both boot clean and log `Trinity Enterprise modules registered` — the line `deploy-dev` greps. `edition: enterprise` on both.

Module count is unchanged because `shared_sessions` leaves as `credential_vault` arrives. Neither pointer produced duplicate room paths, which confirms the include-order net was holding; after this bump there is nothing left for it to net.

*(Module names withheld from this description per the standing public-repo rule — trinity-enterprise#45.)*

## Scope and risk

- **Gitlink only in the diff — but NOT inert on a live instance.** Surfaced by `/review` and stated here so it is not discovered at deploy time: the next `deploy-dev` will create `enterprise_credential_vault_entries` and `enterprise_credential_vault_grants` at module registration. Additive and idempotent (`CREATE TABLE IF NOT EXISTS` on the SQLite track, `IF NOT EXISTS` on the Alembic track), no existing row touched, and the only `DROP TABLE` statements live in `0017`'s `downgrade()`. No OSS source changes, so public CI — which never checks the submodule out (`update = none`) — is unaffected, and `backend boots without enterprise submodule` still exercises the OSS-only path.
- The real gate is `deploy-dev`'s registration boot-log grep, which is exactly what I reproduced locally.
- Both pointers are on enterprise `main`; this is not adopting a feature branch. (Note `dev`'s current pin `a419812` is reachable from `main` but its commit sits on a long-lived feature branch name, which is how the drift went unnoticed.)

Related to ent#443, ent#435, ent#279


dolho (commented):
## /review — pre-landing review

**Branch**: `chore/bump-enterprise-pointer` → `dev` (merge-base `135248e9`)
**Files changed**: 1 (+1/−1) — a gitlink and nothing else
**Scope**: CLEAN
**Plan completion**: no plan (chore)

The diff is one line, so most of the checklist is vacuous. The reviewable surface is **what is inside the four enterprise commits**, and that is where I spent the pass. Zero critical findings.

### Critical findings

**None.** The four checks that could have blocked this, each verified against the enterprise tree rather than assumed:

**1 · Does the ent#443 removal destroy live room data?** This was the highest-stakes question — the room tables moved to OSS core under the *same names* (retained history, ent#443), so a `DROP` on the enterprise side would take real transcripts with it.

`2a5def3` is **code-only**. Grepping the commit for `DROP TABLE` / `DELETE FROM` / `drop_table` returns nothing; it deletes modules and ~3,861 lines of tests. Tables survive and OSS core owns them.

**2 · Is the Alembic chain intact?** `architecture.md` states `0011_shared_sessions` must stay on its line — *"deleting it would break that chain"*. It survives at `main`, still chained `0011_shared_sessions → 0010_skill_runner_acl`.

```
alembic-heads: backend/migrations/versions — 18 revision(s), 1 head (0017_credential_vault) — PASS
```

One head on the private version-line, which is Invariant #3's requirement and the thing `check_alembic_heads.py` exists to catch.

**3 · Is the new module dual-tracked (Invariant #9)?** I flagged this as a suspected gap and it is **not one** — recording the check because any reviewer will ask it. `credential_vault` ships an Alembic revision but its `migrations/` package holds only `__init__.py`, which *looks* like a missing SQLite half. It isn't:

```python
# credential_vault/__init__.py
if is_sqlite():
    from .schema import init_credential_vault_schema
    init_credential_vault_schema()
    run_enterprise_migrations_for(credential_vault_migrations)
```

The SQLite half is a bespoke `init_..._schema()` at registration (the OSS `db/schema.py` shape); the empty `migrations/` package is for future *backfills*, not initial DDL. `schema.py`'s own docstring says so: *"SQLite track only — PostgreSQL is owned by the enterprise Alembic line."* Both tracks covered.

**4 · Is the bump destructive to an existing `dev` database?** No. Both tracks are idempotent — three `CREATE TABLE/INDEX IF NOT EXISTS` on the SQLite side, three `IF NOT EXISTS` on the Alembic side. The only `DROP TABLE IF EXISTS` statements sit in `0017`'s `downgrade()` (line 63+), never on the upgrade path.

### Informational

**[I1] The bump has a real side effect on the dev database** (confidence 9/10). Not a defect, but it should be said out loud rather than discovered: the next `deploy-dev` will create `enterprise_credential_vault_entries` and `enterprise_credential_vault_grants` at registration. Additive, idempotent, no existing row touched — but this PR is not inert on a live instance, and a one-line "gitlink only" description could read as though it were.

**[I2] Nothing prevents the next inversion** (confidence 8/10). The drift went unnoticed because `dev`'s pin sits on a long-lived *feature-branch* name (`60-brain-orb-phase-3-…`) even though the commit is reachable from `main` — so `git submodule status` looks plausible. Two cheap assertions would make this class self-reporting:

- `dev`'s pin is an ancestor of enterprise `main`
- `dev`'s pin is **not behind** OSS `main`'s pin (the inversion itself)

Deliberately **not** added here: this PR's whole value is being a one-line gitlink a reviewer can merge without reading code, and folding a CI guard in would destroy that property. Worth its own issue — happy to file it.

### Verified clean

- **Credential exposure** — `credential_vault` stores `value_encrypted` as an AES-256-GCM envelope per Invariant #12, never echoed on read; `ff0a4f1` is the *private twin* of the OSS cleartext-sink guard `architecture.md` records as *"the private submodule owns its twin"*. This bump is security-**positive**: `dev` currently runs the OSS half of ent#435 without its enterprise half.
- **OSS ↔ enterprise coupling** — `main.py` imports `shared_sessions.router` from **OSS core** (`src/backend/shared_sessions/` exists on `dev`), not from the submodule. Nothing in OSS depends on the enterprise copy, so removing it cannot break the public tree.
- **No second mount to bump** — `.gitmodules` declares one submodule (`src/backend/enterprise`); there is no separate frontend gitlink on `dev`.
- **Fast-forward, not a rewind** — `a419812...main` = ahead 0 / behind 4; `2a5def3...main` = ahead 3 / behind 0. Both current pins are strict ancestors.
- **Public-repo hygiene** — no paid-module names in the commit message or PR body (trinity-enterprise#45); `enterprise-docs-guard` unaffected, no OSS seam file touched.
- **Public CI** — no OSS source changed, and CI never checks the submodule out (`update = none`), so `backend boots without enterprise submodule` still exercises the OSS-only path unchanged.

### Runtime evidence

Both pointers booted against `dev` on the same box, real uvicorn, submodule mounted at each commit in turn:

| pointer | modules | `shared_sessions` registered | room paths | enterprise errors |
|---|---|---|---|---|
| `a419812` (today) | 17 | **True** | 6 | 0 |
| `90f2f2c` (this PR) | 17 | **False** | 6 | 0 |

Both log `Trinity Enterprise modules registered` and report `edition: enterprise`. Count is unchanged because `shared_sessions` leaves as `credential_vault` arrives.

### Summary

- **Critical: 0**
- **Informational: 2** — one is a disclosure about live effect, one is a follow-up suggestion
- **Scope: clean** — gitlink only, no drift

Recommend merge. The real gate remains `deploy-dev`'s registration boot-log grep, which is what I reproduced locally.

vybe (approved):
Validated via /validate-pr: pointer target 90f2f2c matches enterprise main tip; fixes the dev<main pointer inversion before it could ship backwards in a release. CI green.

PR #2451, opened 2026-08-31
fix(workspace): the agent page rendered every ask twice, and one copy could not show expiry (#2449)

### LINKED ISSUE #2449: Workspace agent page renders every ask twice, and the second copy cannot show expiry
## Summary

The Workspace agent page renders the same ask **twice**, from two different projections of one `operator_queue` row. The second rendering is the pre-ent#428 one: it carries no `status` and no `expires_at`, so it cannot show expiry, and its only action is a "Reply in chat →" link justified by a comment that is no longer true.

Found by testing ent#364/#428/#429 against a live instance on `dev` (`9b0ed63b`).

## The two renderings

`src/frontend/src/components/portal/PortalAgentPage.vue`:

```
line  42:  <div v-if="store.asksForAgent(agentName).length">
             <PortalAsks :agent-name="agentName" ... />     <- /asks store (ent#428/#429)

line 212:  <section v-if="asks.length">                     <- page.asks (ent#360)
             <h2>Waiting on you</h2>
```

`const asks = computed(() => page.value?.asks || [])` (line 461) is the agent-page payload; `PortalAsks` reads `store.asksForAgent(...)`. Both are fed by the same `operator_queue` row, so **both guards are true at the same time** — measured, not inferred:

```
GET /asks                     -> ['2acea3eeb49c']
GET /agents/weather-watch/page -> asks: ['2acea3eeb49c']
```

## The shapes disagree

Same ask, same instant:

| | `/asks` (ent#428) | `page.asks` (ent#360) |
|---|---|---|
| kind | `"kind": "question"` | `"type": "question"` |
| status | `"status": "pending"` | **absent** |
| expiry | `"expires_at": null` | **absent** |
| chat | `"chat_id": "98464e98…"` | **absent** |
| agent | `"agent_name": "weather-watch"` | **absent** |

`PortalAsks.vue` reads `ask.kind`, `ask.status`, `ask.expires_at`, `ask.agent_name` — none of which exist on `page.asks`.

## Why it matters

1. **ent#429 AC is not met on that surface** — *"An expired ask is visibly expired in all three surfaces, never silently absent."* `agent_page._asks()` queries `status="pending"`, so an expired-but-pending ask **is** returned, and with no `expires_at` in the projection the "Waiting on you" block renders it identically to a live one. The `PortalAsks` block above it does show expiry, so the page shows the same ask in two states at once.

2. **The comment justifying the second block is obsolete.** It reads *"Answering an approval writes to the operator queue, an operator surface with its own auth… rather than render a control that 403s for a client, the answer path is the conversation."* That was true before ent#428. A client answer now works:

   ```
   POST /api/enterprise/client-portal/asks/{id}/answer  -> 200
   ```
   with `responded_by_email` set and `responded_by_id` NULL, exactly as ent#428 specifies. So the page offers "Reply in chat →" for something the surface directly above it can answer in one click.

3. **ent#364's "one ask entity, three renderings"** is contradicted on the page that is one of the three: two renderings, two shapes, two different affordances.

Not a staleness bug — both sources filter on `pending` and both cleared when I answered, so answering in one does clear the other.

## Reproduce

1. Share an agent with a Workspace client.
2. Write an ask addressed to that client into the agent's `~/.trinity/operator-queue.json` and let the 5s sync ingest it.
3. Open that agent's page in the Workspace as the client.
4. The ask appears twice — once with answer controls, once under "Waiting on you" with "Reply in chat →".

## Suggested fix

Delete the `page.asks` block (lines ~212-243) and the `asks` computed, and keep `PortalAsks` as the single rendering. That also drops `_asks()`'s DB query from every agent-page load and removes `PortalAgentAsk` from the response model.

Two details for whoever takes it:
- The Overview tab badge (`asksBadge`, line 108) also reads `page.asks`; repoint it at `store.asksForAgent(agentName)` so the count and the list share a source.
- If the payload field is kept for an API consumer, it should be re-projected through the ent#428 shape rather than left at `type`/no-status — but nothing currently reads it except this component.

## Acceptance criteria

- [ ] An ask appears exactly once on the agent page
- [ ] An expired ask is visibly expired there, matching `/asks` and the chat surface
- [ ] The answer control on the page is the one that works (no "Reply in chat →" detour for an answerable ask)
- [ ] The Overview badge count and the rendered list come from one source
- [ ] A regression test asserts the page cannot render asks from two sources

## Environment

Backend built from `origin/dev` `9b0ed63b`, enterprise edition, PostgreSQL. Verified with a real portal client session against a real shared agent.

Related to abilityai/trinity-enterprise#364, abilityai/trinity-enterprise#428, abilityai/trinity-enterprise#429

dolho:
Closes #2449

## What was wrong

`PortalAgentPage.vue` mounted asks from **two sources whose guards were true at the same time**. Measured against a live instance — same ask id in both:

```
GET /asks                       -> ['2acea3eeb49c']
GET /agents/weather-watch/page  -> asks: ['2acea3eeb49c']
```

```
line  42   <PortalAsks>        <- the /asks store (ent#428/#429)
line 212   "Waiting on you"    <- page.asks, the page's own _asks() reader
```

A client saw every ask twice: once with answer controls, once with a "Reply in chat →" link. That link's justifying comment — *"rather than render a control that 403s for a client"* — predated ent#428's client answer route and was no longer true; `POST /asks/{id}/answer` returns **200**, verified.

## Why deleting one was the fix, not styling around it

The halves disagreed:

| | `/asks` | `page.asks` |
|---|---|---|
| cap | 200 | `MAX_ASKS` = 20 |
| `status` | ✅ | **absent** |
| `expires_at` | ✅ | **absent** |
| `chat_id` | ✅ | **absent** |
| field name | `kind` | `type` |

`_asks()` queries `status="pending"`, so an **expired-but-pending** ask was returned and rendered identically to a live one — ent#429's *"visibly expired in all three surfaces"* could not hold there. It also fed the Overview badge, which therefore counted expired asks while the sidebar's `askCount` — pending-only by design, *"the person did not fail to answer something they can still answer"* — did not.

## What changed

The badge now counts the same store list the section renders, pending-only, so the two agree. `_asks()`, `ASK_TYPES`, `MAX_ASKS`, `PortalAgentAsk` and the payload field are gone — one DB query per page load with them.

**Nothing is lost, and I checked rather than assumed.** `list_asks` covers strictly more than the removed reader: revoked-share re-check, unreadable-roster fail-closed, expiry, context-never-forwarded. The two `_asks` tests I dropped assert rules the surviving path already tests; the two ent#364 ACs are **repointed** at `list_asks`, which is the path the page actually uses.

One behavioural difference is real and **was already live**: `_asks` excluded `alert` items while `/asks` includes them deliberately. The page already showed them through `<PortalAsks>`; only the duplicate hid them.

## Residual recorded, not resolved — please read this one

**#2169's instructed "asks sit below the top row" no longer describes the page — and it was already void before this PR.** ent#428 mounted `<PortalAsks>` above the stats strip, so the page led with asks *and* repeated them lower down. The two ordering tests kept passing only because the lower copy existed.

Removing a duplicate is the wrong place to settle where asks belong, so the rule is recorded as retired in the flow doc and inline in the spec, with the note that the fix is to **move the survivor**, never to restore a second one. If "below the top row" still stands, that is a one-line move and I'll do it.

## Tests

New `portalAskSingleSource.spec.js` guards the mechanism — one `<PortalAsks>` mount, no reader of the removed field anywhere under `src/`, badge from the store, pending-only. **Mutation-verified three ways**, each fails:

| mutation | result |
|---|---|
| re-introduce a `page.asks` reader | **FAIL** |
| mount `<PortalAsks>` a second time | **FAIL** |
| drop the pending filter from the badge | **FAIL** |

Its first draft was itself blind — the scan matched the string inside my own explanatory **comment**, the identical defect the ent#430 CAS guard hit this morning. It now uses the shared `stripComments` helper (#2161), which handles unterminated and nested HTML comments that the hand-rolled stripper I started with did not.

Of the six tests pinning the deleted block: **three repointed** at the surviving rendering, **three retired** with their reasons stated inline rather than silently dropped.

Frontend **69 files / 1530 tests**; backend **521 passed, 1 skipped** across the portal/asks/agent-page suites.


vybe (approved):
Validated via /validate-pr: Closes #2449 ✅, single-source ask render with FE+BE regression tests, feature-flow doc updated, build + e2e green.

obasilakis (commented):
Validated against the Trinity methodology — no blockers, CI green, all five ACs met, and deletion-as-fix is the right call here. Holding off on approving for two things I'd like confirmed first, both about what the PR says versus what it does.

**1. An undisclosed test removal.** `portalAgentPageUx.spec.js` drops `keeps the top row two columns whether or not there are asks`, which is not named in the PR's "RETIRED BY #2449" block — contradicting the claim that retired tests had their reasons stated inline. That test was #2169's only guard on the real Overview row being a two-column grid. The sole surviving `xl:grid-cols-2` assertion (`:326`) is scoped to the **loading skeleton**, which the deleted test's own comment explicitly said cannot stand in for it ("passed on the skeleton, and the headline AC failed with a green suite"). Production classes at `PortalAgentPage.vue:161` are untouched so there's no behaviour regression, but #2169's layout AC is now unguarded. Either restore a real-row assertion, or add it to the retired list with a reason.

**2. A response-shape change.** `client_portal/models.py` drops `asks` from `PortalAgentPage` — a breaking change on `GET /agents/{name}/page`. The issue asserts this component is the only consumer, which I can verify inside the repo but not for external headless clients. Worth a release-note line at minimum.

I did verify the "dropped tests are already covered" claim rather than trusting it, and it **holds**: `tests/unit/test_ent428_workspace_asks.py:206` (`test_the_projection_never_forwards_agent_authored_context`) genuinely covers the dropped context test, and the alert-exclusion difference is honestly disclosed as intentional rather than silently lost. The `/asks` versus `page.asks` divergence table in the body is accurate — deleting one half rather than styling around it was correct.

Cosmetic leftovers, no action needed unless you're already in the files:

- `portalAgentPageUx.spec.js:71` — `overviewBlock` is now dead code (its only consumer was the deleted test), and the file-header comment at `:6` still describes "a two-column grid conditional on `asks.length`".
- `:260` and `:314` — duplicate test, same name *and* identical assertion, both added or repointed in this PR. Both pass; pure redundancy.
- `PortalAgentPage.vue:377-383` — two orphaned comment blocks describing the deleted `ASKS_PREVIEW` / `ASKS_CAP` constants, one still explaining why they were "named here rather than inlined so the duplication across the API boundary is visible".

Security and packaging clean: no secrets, emails or IPs; no new top-level `src/backend/*.py`; no new `os.getenv()`.

PR #2453, opened 2026-08-31
feat(ci): publish the five Trinity images to GHCR on release (#2452)

### LINKED ISSUE #2452: feat(ci): publish the five Trinity images to GHCR on release — the unshipped half of the #2280 hosted-install gate
## Summary

Add a release workflow that builds and publishes the five Trinity images to GHCR on every release tag, with public (anonymous) pull access. This is **#2280's AC1**, carved out because it is the single unchecked item blocking every marketplace and one-click channel, and it was invisible inside an issue labelled `status-in-dev`.

## Context

`docker-compose.hosted.yml`, `scripts/deploy/start.sh --hosted`, and the #2281 Packer bundle all shipped against a registry contract that nothing satisfies. There is no publish workflow in the repo:

```
$ grep -rln "ghcr.io" .github/workflows/     → (nothing)
$ curl -s -o /dev/null -w "%{http_code}\n" \
    https://ghcr.io/v2/abilityai/trinity-{backend,frontend,mcp-server,scheduler,agent-base}/tags/list
  401  401  401  401  401
```

Consequence: `packer build` for the DigitalOcean 1-Click (#2281) fails at its first `docker pull`, and any operator following the documented `--hosted` path fails at `start.sh:524`. #2280 is labelled `status-in-dev` with this AC unchecked, so the blocker reads as done.

Downstream: #2281 (DigitalOcean), #2282 (Vultr), #2283 (Hostinger/Dokploy), #835 (GCP/Hetzner Packer). Epic #2332 names #2280 "the P1 gate; everything else depends on it" — this is the half of that gate which has not landed.

## The contract (already fixed by three consumers)

| requirement | set by |
|---|---|
| repos `ghcr.io/abilityai/trinity-{backend,frontend,scheduler,mcp-server,agent-base}` | `docker-compose.hosted.yml`, five `image:` lines |
| **both** a version tag and `latest` per image | compose defaults `${TRINITY_IMAGE_TAG:-latest}`; `packer/digitalocean/trinity.pkr.hcl` requires an explicit tag and rejects `latest` |
| anonymously pullable | `start.sh:524` and `packer/digitalocean/scripts/01-provision.sh:50` both `docker pull` with no `docker login` |
| linux/amd64 | DigitalOcean droplets |

GHCR packages default to **private**. Publishing alone does not clear the 401s — each package's visibility has to be set public.

## Acceptance Criteria

- [ ] `.github/workflows/publish-images.yml` builds and pushes all five images on a `v*` tag push to `main`, plus `workflow_dispatch` for re-cuts. `latest` is never published from `dev`.
- [ ] Checkout is explicitly `submodules: false` — these are public images and must not carry the private enterprise submodule. (`build-without-submodule.yml` already proves the backend builds without it; this workflow must *choose* it rather than inherit it.)
- [ ] Each image is pushed with two tags: the release version (e.g. `v0.9.0`) and `latest`.
- [ ] The backend image receives its six provenance build args (`VERSION`, `GIT_COMMIT`, `GIT_COMMIT_SUBJECT`, `GIT_COMMIT_TIMESTAMP`, `GIT_BRANCH`, `BUILD_DATE`, #926) — verified by asserting `GET /api/version` returns real values, not `unknown`.
- [ ] All five GHCR packages are public; an anonymous `docker pull` (no `docker login`) succeeds from a clean machine.
- [ ] `TRINITY_IMAGE_TAG=<version> ./scripts/deploy/start.sh --hosted` comes up on a clean VM using only pulled images — no local build, and `trinity-agent-base:latest` is present after the retag at `start.sh:533`.
- [ ] #2280's AC1 is ticked, or #2280 is closed if this was its last open item.

## Technical Notes

- **Build contexts differ per image**: backend `docker/backend/Dockerfile`; frontend `docker/frontend/Dockerfile.prod`; scheduler `docker/scheduler/Dockerfile`; mcp-server `src/mcp-server/Dockerfile` with context `src/mcp-server`; agent base `docker/base-image/Dockerfile`.
- **The frontend image is droplet-agnostic** — `src/frontend/src/api.js` sets `baseURL: ''`, so the SPA calls its own origin and `VITE_API_URL` is vestigial on this path. Nothing per-instance needs baking.
- **linux/amd64 only for v1.** #2280 lists arm64 as desirable, not required; it roughly doubles agent-base build time (Go, Node, Claude Code, plugin preinstall) for no marketplace target today. Track arm64 separately if Hetzner CAX or Umbrel is wanted.
- **`TRINITY_PREINSTALL_PLUGINS=1` makes network calls at build time** (`claude plugin marketplace add`) and is deliberately non-fatal, so a registry flake ships an image with the platform plugin silently absent — the #2305/#2308 symptom, baked in and fleet-wide. Decide explicitly whether the publish build fails on that step rather than inheriting the local-build tolerance.
- Precedent for a publish workflow with tag-or-dispatch triggering: `.github/workflows/publish-cli.yml`.
- Retention/cleanup for old tags is out of scope; GHCR is free for public packages.

obasilakis:
Closes #2452.

## What

Adds `.github/workflows/publish-images.yml` — builds and publishes the five Trinity images to GHCR on a `v*` release tag. This is **#2280's AC1**, the one unchecked item on an issue labelled `status-in-dev`, and the blocker under every marketplace channel (#2281 DigitalOcean, #2282 Vultr, #2283 Hostinger/Dokploy, #835 GCP/Hetzner).

Before this there was no publish workflow in the repo at all:

```
$ grep -rln "ghcr.io" .github/workflows/     → (nothing)
$ curl -s -o /dev/null -w "%{http_code}\n" \
    https://ghcr.io/v2/abilityai/trinity-{backend,frontend,mcp-server,scheduler,agent-base}/tags/list
  401  401  401  401  401
```

So `start.sh --hosted` failed at line 524 and `packer build` failed at its first `docker pull`.

## Shape

Two phases. Phase 1 pushes only the immutable version tag for all five images; phase 2 moves `latest`, and only once every image has landed, by re-tagging the pushed digests (`buildx imagetools create`, no rebuild). A single-phase matrix that tagged `latest` per image would, on a partial failure, leave `latest` pointing at four new images and one old one — the set every hosted install pulls by default, and one no error message would explain.

The registry contract was not a design choice here; three consumers had already fixed it:

| requirement | set by |
|---|---|
| the five repository names | `docker-compose.hosted.yml`'s `image:` lines |
| both a version tag **and** `latest` | compose defaults `${TRINITY_IMAGE_TAG:-latest}`; `trinity.pkr.hcl` requires an explicit tag and rejects `latest` |
| anonymous pull | `start.sh:524` and `01-provision.sh:50` — neither runs `docker login` |
| image tag is also a git ref | `01-provision.sh` clones `--branch "${TRINITY_IMAGE_TAG}"` |

## Load-bearing choices

- **`submodules: false` is explicit, not inherited.** These images are world-pullable; a runner with enterprise access checking out recursively would bake private code into a public artifact. `build-without-submodule.yml` proves the backend builds without it — this workflow has to *choose* it.
- **Release tags must be ancestors of `main`**, and the tag's implied version must match the `VERSION` file. Without the first, a tag pushed from a feature branch publishes `latest` from unreleased code, which #2280's AC forbids. Without the second, a release that forgot a bump ships an image that misreports itself to every hosted install.
- **The backend's six provenance args (#926) are asserted**, by reading the baked ENV rather than booting the backend — the claim under test is only that the args reached the image `/api/version` reads.
- **A withheld platform plugin fails the build.** ent#411's pre-install is deliberately non-fatal so a local build survives an unreachable marketplace; on a published artifact that same tolerance bakes the #2305/#2308 symptom into every agent, fleet-wide. The assertion reads `installed_plugins.json` and confirms the recorded `installPath` exists — *not* that `~/.claude/plugins` is non-empty, because a `marketplace add` that succeeds followed by an `install` that fails leaves that directory full of marketplace metadata and no plugin.
- **A final job pulls all five with no credentials.** GHCR packages are private by default and publishing does not change that, so without this a green run still leaves every marketplace install on a 401 — the exact state #2452 describes.

linux/amd64 only: #2280 lists arm64 as desirable, not required, and there is no arm64 marketplace target today.

## Verification

Run locally against this branch:

- All five images build with the exact contexts the matrix declares. Three of the five are **not** the repo root, and the backend's `COPY ../../src/backend/...` paths resolve because BuildKit clamps them to the context root — confirmed by a real build, not inspection.
- Provenance assertion against a built backend image: `VERSION=0.9.0 GIT_COMMIT=deadbeef …`, no `unknown`.
- Plugin assertion **both directions** — passes on a normal build (`trinity@abilityai 2.8.2 at /home/developer/.claude/plugins/cache/abilityai/trinity/2.8.2`), and fails on one built with `TRINITY_PREINSTALL_PLUGINS=0` (`no installed_plugins.json — the pre-install was withheld entirely`). A check that cannot fail is worse than no check.
- `actionlint` clean.

Not verified: an actual push to GHCR, which needs a real release tag on `main`.

## One manual step after merge

The workflow cannot flip package visibility. On the first publish, set each of the five packages to **Public** once at `https://github.com/orgs/abilityai/packages`. The `verify-anonymous-pull` job fails with that instruction until it is done.

## Follow-ups, deliberately not here

- **#2280's other unchecked AC** — the cloud-init user-data example — is untouched.
- **#2381** (unauthenticated `POST /api/setup/admin-password` overwrites the admin password) reaches a marketplace droplet through Caddy at `https://<ip>/api/setup/admin-password`, since `nginx.conf:39` proxies `/api/` to the backend. The firewall rules in #2281 do not close it because 443 is the product. That is the other hard blocker on the DigitalOcean listing, and it is its own issue.
- **`docker-compose.hosted.yml` publishes 8000/8080/8686 and the OTel ports on `0.0.0.0`**, and Docker's iptables run ahead of ufw. #2281's `DOCKER-USER` rule covers new droplets; already-deployed hosted installs on public IPs are unprotected. Worth its own issue.

https://claude.ai/code/session_01218nVkqK7vU2jy9R73s4jT

### DIFF
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -1,269 +1,326 @@
-name: Publish container images
+name: publish-images
 
-# Publishes the five Trinity-built images to GHCR on every release tag, so a
-# hosted install pulls instead of building (#2280). Without this, a fresh VM
-# spends 5-10 minutes compiling the agent base image on first boot, which
-# fails the one-click bar for every marketplace channel (#2281, #2282).
+# Publish the five Trinity images to GHCR on a release tag (#2452).
 #
-# The checkout deliberately does NOT init submodules: `src/backend/enterprise`
-# is private, and these images are PUBLIC. A plain checkout leaves that path
-# empty, so the published backend is OSS-only by construction rather than by
-# a .dockerignore rule someone can edit (build-without-submodule.yml already
-# proves the OSS tree builds standalone).
+# WHY THIS EXISTS
+# `docker-compose.hosted.yml`, `scripts/deploy/start.sh --hosted` and the #2281
+# DigitalOcean Packer bundle all shipped against a registry contract that
+# nothing satisfied: there was no publish workflow in the repo and all five
+# GHCR repositories answered 401. So the documented pull-only install path
+# failed at `start.sh:524` and `packer build` failed at its first `docker pull`.
+# This is #2280's AC1 — the unshipped half of the one-click hosted gate.
+#
+# THE CONTRACT IS NOT A CHOICE — three consumers already fixed it:
+#   * repository names come from docker-compose.hosted.yml's five `image:` lines
+#   * EVERY image needs BOTH tags: compose defaults to `${TRINITY_IMAGE_TAG:-latest}`,
+#     while trinity.pkr.hcl requires an explicit tag and REJECTS `latest`
+#   * pulls are anonymous — neither start.sh nor 01-provision.sh runs `docker login`
+#   * the git tag IS the image tag: 01-provision.sh does
+#     `git clone --branch "${TRINITY_IMAGE_TAG}"`, so a tag that is not also a
+#     git ref produces a snapshot whose checkout and images cannot agree.
+#
+# TWO PHASES, DELIBERATELY
+# Phase 1 pushes only the immutable version tag, for all five images. Phase 2
+# moves `latest` — and only after all five have succeeded. A single-phase matrix
+# that tagged `latest` per image would, on a partial failure, leave `latest`
+# pointing at four new images and one old one: a mixed set that every hosted
+# install pulls by default and that no error message anywhere would explain.
+# Phase 2 re-tags the already-pushed digests via `buildx imagetools create`,
+# so it never rebuilds and cannot publish anything phase 1 did not verify.
+
 on:
   push:
     tags:
       - 'v*'
   workflow_dispatch:
     inputs:
-      ref:
-        description: 'Tag or SHA to build (defaults to the triggering ref)'
-        required: false
+      tag:
+        description: 'Existing git tag to build (e.g. v0.9.0)'
+        required: true
         type: string
+      publish_latest:
+        description: 'Also move the `latest` tag'
+        required: false
+        default: true
+        type: boolean
 
 permissions:
   contents: read
   packages: write
 
 concurrency:
-  group: publish-images-${{ github.ref }}
+  # Never two publishes racing for the same `latest`.
+  group: publish-images
   cancel-in-progress: false
 
+env:
+  REGISTRY: ghcr.io
+  # Lowercased deliberately: GHCR rejects an uppercase path component, and the
+  # org is spelled `Abilityai` in git remotes.
+  OWNER: abilityai
+
 jobs:
-  publish:
-    name: ${{ matrix.image }}
+  # ---------------------------------------------------------------------------
+  # Resolve and vet the release before anything is built.
+  # ---------------------------------------------------------------------------
+  resolve:
+    name: Resolve release
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          # `build_args` is per image, NOT a shared block. Only the backend
-          # declares the six provenance ARGs (docker/backend/Dockerfile:37-42)
-          # → ENV → `GET /api/version`; the scheduler and mcp-server Dockerfiles
-          # declare none at all, and passing an undeclared --build-arg makes
-          # BuildKit warn on every build, which trains everyone to ignore the
-          # warnings that matter.
-          - image: trinity-backend
-            context: .
-            dockerfile: docker/backend/Dockerfile
-            build_args: provenance
-          - image: trinity-frontend
-            context: ./src/frontend
-            dockerfile: docker/frontend/Dockerfile.prod
-            # Deliberately none. Every VITE_* var is statically inlined into the
-            # world-readable bundle at build time, so a published image can only
-            # carry values that are correct for EVERY install. The two that are
-            # actually read (VITE_BUG_REPORTING_ENABLED, VITE_BUG_INTAKE_URL)
-            # already default to exactly that in the Dockerfile. VITE_API_URL is
-            # inert — api.js uses `baseURL: ''` and nginx proxies same-origin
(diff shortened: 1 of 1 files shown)


obasilakis:
Closing — I built this on a false premise and it would have been a regression.

`.github/workflows/publish-images.yml` **already exists on `dev`**, added by #2390 (#2280) on 2026-08-27, and `tests/unit/test_2280_publish_workflow_and_stop.py` is its reviewed spec. My local checkout of `~/trinity` was 40-odd commits behind `origin/dev`, so `ls .github/workflows/` and `grep -rl ghcr.io` both came back empty and I read that as 'no publish workflow exists'. This branch then **overwrote** the real one, and the five `regression diff` failures on this PR are that reviewed spec correctly rejecting the replacement.

The real workflow is better specified than mine on every point the review round caught — `flavor: latest=false`, per-tag `github.event_name == 'push'` gating, the deliberate rejection of `type=sha` because `github.sha` resolves to the dispatching branch head, and the lowercased reference in `Verify anonymous pull` (this org is `Abilityai`; docker rejects a mixed-case reference locally, before any network call).

What is actually true, and all that remains: the workflow **has never run** — `gh run list --workflow publish-images.yml` returns zero runs, because the newest tag is `v0.9.0` and the workflow merged after it. So GHCR still answers 401 on all five repositories, and #2280's AC1 stays unticked, but the missing thing is a release cut, not a workflow.

#2452 was filed on the same false premise and is being closed with it.

